PR #1083, opened 2015-03-08
 Add const to count

loverszhaokai:
2. Thus makes it clear that the variable 'count' will never be changed in crypt_all()


magnumripper:
I won't merge this with that local merge of yours present. You need to rebase.

loverszhaokai:
OK, I will rebase it.

PR #1085, opened 2015-03-08
Add const to the variable: count in *_fmt*.c::crypt_all()

loverszhaokai:
2. Add const to make it clear that the variable can be changed in the function crypt_all()


magnumripper:
Sorry, I still can't accept this patch. It destroys indentation. You can't just change tabs to spaces and think noone will care about it. I, for one, care a lot about that.

loverszhaokai:
OK, I will change it.

PR #1092, opened 2015-03-10
change 'int count = *pcount;' to 'const int count = *pcount;'

loverszhaokai:
2. Add const to make it clear that the variable can be changed in the function crypt_all()


magnumripper:
> to make it clear that the variable can be changed

You mean can't be changed ;-)

OK, I'm merging this. I hope you tested it...

loverszhaokai:
Some format has a very long crypt_all() and they never change the variable count, So I think change it to const to make it clear. I think it will never lead to a bug. Thank you.

PR #1100, opened 2015-03-12
Building for MIC

zzlei:
There're some minor modifications to autoconf scripts and a source file.
A tutorial is also added, named README-MIC.

### DIFF
--- a/doc/README-MIC
+++ b/doc/README-MIC
@@ -1,0 +1,74 @@
+
+Intel MIC (Xeon Phi) is a coprocessor computer architecture developed by Intel.
+Program built for MIC cannot run on normal Intel processors. So you need to 
+have a MIC coprocessor and related enviroments setup before trying to build 
+John the Ripper (JtR) for MIC. Tutorial on how to setup the software 
+environments for MIC can be found on Intel's website.
+
+
+-----------------------
+Library Denpendencies:
+-----------------------
+
+JtR requires some libraries that are not available on MIC, which means you'll 
+need to built them for MIC by yourself.
+These libraries are:
+    Zlib (libz)
+    GMP (libgmp)
+    OpenSSL (libssl & libcrypto)
+
+They can be downloaded from their websites. But building them require some 
+effort, and only the mentioned versions are guaranteed to work.  
+Assuming those libraries are to be install under path $MIC and the path to
+JtR is $JOHN, then follow the steps below.
+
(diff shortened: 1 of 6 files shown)


doc/README-MIC
```diff
@@ -0,0 +1,74 @@
...
+    OpenSSL (libssl & libcrypto)
+
+They can be downloaded from their websites. But building them require some 
+effort, and only the mentioned versions are guaranteed to work.  
+Assuming those libraries are to be install under path $MIC and the path to
+JtR is $JOHN, then follow the steps below.
+
+Build Zlib (version 1.2.8):
+$ cd */zlib-1.2.8
+$ CC=icc CFLAGS="-mmic" ./configure --prefix=$MIC
```
magnumripper: The normal way to pass things to configure is on the right side, as in
```
./configure CC=icc CFLAGS="-mmic" --prefix=$MIC
```
Even if having them on the left side "works" there are subtle differences that should be avoided.

magnumripper:
Looks good to me. I think putting those external patches in unused/ was the right thing for now. Please correct the `./configure` syntax and I'll merge this.

BTW we have never had to patch config.sub before. The one we had was verbatim from autoconf. Did you do any research leading up to this solution? I'm not saying you did wrong at all, I'm just not sure.

frank-dittrich:
> BTW we have never had to patch config.sub before. The one we had was verbatim from autoconf. Did you do any research leading up to this solution? I'm not saying you did wrong at all, I'm just not sure.

I don't think it is wrong. If you look at the patches in unused, you'll see the same change for those libraries autoconf.
I guess it is just lacking mic support in the particular autoconf versions.
May be something to report upstream?

magnumripper:
> May be something to report upstream?

That's part of what I meant. Maybe there *is* a newer version available upstream.

jfoug:
NOTE, we are 9 months to a year out of date on the 'core' autoconf stuff.  mic*) may already be in this file, we just have not check for updates ourselves.

zzlei:
I checked out autoconf's latest source code. It still doesn't recognise MIC. Actually config.sub was last updated in 2013.
We may consider submitting a patch to their community.

doc/README-MIC
zzlei: That's fixed now. Thank you.
BTW, I'm still working on removing the "-lssl -lcrypto" flags from LDFLAGS when building JtR for MIC.

PR #1102, opened 2015-03-14
fix valid bug for 7z_fmt_plug.c

loverszhaokai:
The result when p points to different string.

String p points to | Result    
---------------------- | -------------
"0"                       | PASS
""                         | PASS
"."                       | PASS
"1"                       | FAILED

Should "" and "." pass the valid() check?
If they are should not, I guess **the author wants only "0"**?
Am I right?  @kholia

### DIFF
--- a/src/7z_fmt_plug.c
+++ b/src/7z_fmt_plug.c
@@ -100,7 +100,7 @@ static void init(struct fmt_main *self)
 static int valid(char *ciphertext, struct fmt_main *self)
 {
 	char *ctcopy, *keeptr, *p;
-	int len, type, NumCyclesPower;
+	int len, NumCyclesPower;
 
 	if (strncmp(ciphertext, FORMAT_TAG, TAG_LENGTH) != 0)
 		return 0;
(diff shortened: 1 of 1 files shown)


magnumripper:
Have a look at 7z2john.py for clues.

loverszhaokai:
Thanks. I found it maybe $7z$0$... by 7z2john.py:786.

```
        while True:
            ...
            data = bytes('', 'ascii')
            for folder in streams.unpackinfo.folders:
                file.seek(streams.packinfo.packpos, 1)
                props = folder.coders[0]['properties']
                # decode properties
                if SetDecoderProperties2(props):
                    # derive keys
                    # password = "password".encode('utf-16-le')
                    # print NumCyclesPower, Salt, password
                    # key = pylzma.calculate_key(password, NumCyclesPower, salt=Salt)
                    # cipher = pylzma.AESDecrypt(key, iv=str(iv))
                    global Salt
                    if len(Salt) == 0:
                        Salt = "\x11\x22"  # fake salt
                    for idx in range(len(streams.packinfo.packsizes)):
                        tmp = file.read(streams.packinfo.packsizes[idx])
                        fname = os.path.basename(self._file.name)
786:                    print "%s:$7z$0$%s$%s$%s$%s$%s$%s$%s$%s$%s" % (fname,
                            NumCyclesPower, SaltSize, binascii.hexlify(Salt),
                            ivSize, binascii.hexlify(iv), folder.crc, len(tmp),
                            folder.unpacksizes[idx], binascii.hexlify(tmp))
                        size = folder.unpacksizes[idx]  # for checksum check
                        if len(folder.unpacksizes) > 1:
                            sys.stderr.write("%s : multiple unpacksizes found, not supported fully yet!\n" % fname)
                        return
```

magnumripper:
Wow so it's actually hard coded to 0.

Issue #1120, opened 2015-03-19
Segfault in config file 


loverszhaokai:
This segment fault is found by AFL fuzzing

# 1. How to reproduce this segment fault

## 1.1 Change config file

Add '?' before List.Rules in line 739

```
 738# Default Loopback mode rules.
 739[?List.Rules:Loopback]
 740.include [List.Rules:NT]
 741.include [List.Rules:Split]
```

## 1.2 Run it with anything

./john ....

# 2. How to fuzz config 

```shell
$ cd JohnTheRipper/run
$ mkdir config_test
$ cd config_test
$ mkdir test_cases
$ cp ../john.conf  test_cases/
$ afl-fuzz -m none -t 1500+ -i test_cases/ -o out/  ../john  7z_fmt  --config=@@  --skip-self-test
```
# 3. Why segment fault

file: **src/config.c**, list is **NULL** 

```
static void cfg_add_line(char *line, int number)
{
	struct cfg_list *list;
	struct cfg_line *entry;

	entry = mem_alloc_tiny(sizeof(struct cfg_line), MEM_ALIGN_WORD);
	entry->next = NULL;

	entry->data = str_alloc_copy(line);
	entry->number = number;
	entry->cfg_name = cfg_name;

	list = cfg_database->list;
	if (list->tail) {
		entry->id = list->tail->id + 1;
		list->tail = list->tail->next = entry;
	}
	else {
		entry->id = 0;
		list->tail = list->head = entry;
	}
}
```

PR #1121, opened 2015-03-19
Must check cfg_database->list before cfg_add_line()

loverszhaokai:
Fix bug: #1120

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -417,10 +417,13 @@ static int cfg_process_directive_include_section(char *line, int number)
 	strnzcpy(&Section[1], p2, 254);
 	if ((newsection = cfg_get_section(p, Section))) {
 		if (newsection->list) {
-			struct cfg_line *pLine = newsection->list->head;
-			while (pLine) {
-				cfg_add_line(pLine->data, number);
-				pLine = pLine->next;
+			// Must check cfg_database->list before cfg_add_line()
+			if (NULL != cfg_database->list) {
+				struct cfg_line *pLine = newsection->list->head;
+				while (pLine) {
+					cfg_add_line(pLine->data, number);
+					pLine = pLine->next;
+				}
 			}
 			return 0;
 		}


magnumripper:
Thanks. Please note that if you had written eg. "Closes #1120" or "Fixes #1120" instead of what you wrote, the issue would be automatically closed when I merge, which is very convenient. See https://help.github.com/articles/closing-issues-via-commit-messages

loverszhaokai:
@magnumripper  Thanks. I get it.

frank-dittrich:
This fixes the segfault, but I think we need a better fix.

If I change john.conf line 419 in a similar way:

```
-[List.Rules:Loopback]
+[?List.Rules:Loopback]
```
and run john, I get
```
(bleeding-jumbo)run $ ./john hashes.descrypt 
Error in ./john.conf at line 420
```
Looks like reporting the error in line 420 instead of line 419 is an error in john core.
(I checked out 1.8.0 and made a similar change to the `[ List.Rules:Wordlist]` section.)

But john at least tells me there's an error in the config file, while the similar change in line 739
```
-[List.Rules:Loopback]
+[?List.Rules:Loopback]
```
is not detected as an error.
Instead, the line is just ignored and the siubsequent lines will silently be added to the previous section `[List.Rules:oi]`.

loverszhaokai:
@frank-dittrich  
> This fixes the segfault, but I think we need a better fix.
I agree with you

loverszhaokai:
@frank-dittrich  

> Looks like reporting the error in line 420 instead of line 419 is an error in john core.

I think this error can be accepted.

> Instead, the line is just ignored and the subsequent lines will silently be added to the previous section

I am not sure about this, so I tested. The result is that **both the line and the subsequent lines are ignored**, I think it's ok!

magnumripper:
If they are **silently** ignored, it's not OK. This should be improved somehow.

loverszhaokai:
@magnumripper  
Case1: currently ignored
Case2: stop and print error message
Case3: go on and print error message
We should choose Case2 ?

magnumripper:
Option 1 with no warning is dangerous. Option 2 or 3 is fine with me: I'm not sure any of them is better. Pick whatever works best for you.

loverszhaokai:
@magnumripper 
**config_1**
```
# Default Loopback mode rules.
[?List.Rules:Loopback]
.include [List.Rules:NT]
.include [List.Rules:Split]
# For Single Mode against fast hashes
[List.Rules:Single-Extra]
```

**config_2**
```
# Default Loopback mode rules.
[?List.Rules:Loopback]
# For Single Mode against fast hashes
[List.Rules:Single-Extra]
```

config_1: should report error msg for we try to add [List.Rules:NT] to [?List.Rules:Loopback]->list while  [?List.Rules:Loopback]->list is **NULL** 

config_2: **Should we report error msg** when there is **only** a section name [**?**List.Rules:Loopback] ?
thanks

loverszhaokai:
@magnumripper 
> config_1: we should report error msg for we try to add [List.Rules:NT] to [?List.Rules:Loopback]->list while [?List.Rules:Loopback]->list is NULL

Now, I think we should **not** report error msg in config_1. Since cfg_section has **params** and **list**, if current section does not have **list**, then we can include children's **params**

magnumripper:
> config_2: Should we report error msg when there is only a section name [?List.Rules:Loopback] ?
thanks

Either it's an allowed section name, and then all is fine (although we may want to warn about an empty section) or it's a disallowed name, and if so we need to bail out with error. Actually I'm not sure which it is.

magnumripper:
> Now, I think we should not report error msg in config_1. Since cfg_section has params and list, if current section is not a list section [list.XXX], then we can include children's params

> When we find a section, and children are includes, we try to include children's list, If children do not have list, then we try to include their params

In a sense it makes sense. The includes are just like #include in a C file. It's like that section was literally there.

However, if we for some reason disallow `[?List.Rules:Loopback]` as a list, we need to bail out with error regardless of what comes after it.

magnumripper:
I take it `[?List.Rules:Loopback]` is not really disallowed per se, but it's not parsed as a `List.(...)`. So it's like we had this
```
[Label:whatever]    <--    [?List.Rules:Loopback]

[other stuff]    <--    the .include stuff
```
So we should ideally react just like that was literally in the file. Whatever that should be...

loverszhaokai:
@magnumripper 
> However, if we for some reason disallow [?List.Rules:Loopback] as a list, we need to bail out with error regardless of what comes after it.

Do you mean that only the sections which are **[list.XXX]** can have **.include XXX** ?

loverszhaokai:
@magnumripper  I will be offline in two days

loverszhaokai:
@magnumripper 
> However, if we for some reason disallow [?List.Rules:Loopback] as a list, we need to bail out with error regardless of what comes after it.

Do you mean that only the sections which are **[list.XXX]** can have children like **.include XXX** ?

magnumripper:
I think an include section can be anywhere, but the *included* stuff need to be (originally) a section on its own. Not just a list, but any kind of section.

jfoug:
Everything in the conf is/has to be in a section of some kind.  But the start of one section closes the last section

magnumripper:
And the *contents* of an .include is always a full section (of any kind) right? This was some time ago, I can't really remember.

loverszhaokai:
@magnumripper 
**config_1**
```shell
# Default Loopback mode rules.
[?List.Rules:Loopback]
.include [List.Rules:NT]
.include [List.Rules:Split]
# For Single Mode against fast hashes
[List.Rules:Single-Extra]
```

**config_2**
```shell
# Default Loopback mode rules.
[?List.Rules:Loopback]
# For Single Mode against fast hashes
[List.Rules:Single-Extra]
```

In conclusion, should we report any error message about config_1 and config_2?
thanks

magnumripper:
Actually, I'm not sure what **is** the error: Is it that '?' is a disallowed character? Not sure it is.

loverszhaokai:
@magnumripper 

Should I ask the question in john-dev mail list ?

magnumripper:
Yes unless you can find the answers in the docs, that's probably a good idea. Actually for a question about config syntax, john-users may be the better list.

loverszhaokai:
@magnumripper 

After discussed in john-users and john-dev, I think currently this patch is enough. Maybe we can change the way of parsing the john.conf in the future.

magnumripper:
I still wonder if we shouldn't bail out with error if we hit that NULL.

PR #1129, opened 2015-03-20
some bug fixes

zzlei:
Also remove the use of bzero, which is deprecated.

### DIFF
--- a/src/jumbo.c
+++ b/src/jumbo.c
@@ -262,7 +262,7 @@ int strcasecmp(char *dst, char *src) {
 		if ( ((l = (unsigned char)(*(src++))) >= 'A') && (l <= 'Z') )
 			l -= 'A' - 'a';
 	} while (f && (f==l));
-	return return(f - l);
+	return(f - l);
 }
 #endif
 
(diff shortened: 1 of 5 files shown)


src/jumbo.c
```diff
@@ -262,7 +262,7 @@ int strcasecmp(char *dst, char *src) {
 		if ( ((l = (unsigned char)(*(src++))) >= 'A') && (l <= 'Z') )
 			l -= 'A' - 'a';
 	} while (f && (f==l));
-	return return(f - l);
```
magnumripper: Wow. How did that pass?

magnumripper:
Was there a bug fix in there besides the `return return`? I'm not saying you did anything wrong, just trying to understand. The bzero crap was already not used anymore, right?

zzlei:
Yeah, I found no bug with bzero indeed. I just think it's deprecated and removing them should cause no trouble.

src/jumbo.c
zzlei: I was amazed, too. At first I even thought it was some arcane syntax.

Issue #1037, opened 2015-01-28
GPU mask mode, general discussion


magnumripper:
For future experimental/incomplete commits (eg. when taking a stab at some UTF-16 format), please use the topic branch again so we keep the amount of likely-broken code in bleeding-jumbo as low as possible. But let's merge things as soon as they seem stable so it doesn't diverge too much. Also, we often don't find out about problems until we actually merge (#1036 is a good example).

Let's use this issue for general discussion. For specific problems we should create separate issues.

magnumripper:
I saw 6-7G c/s on a GTX 980 with current code. My laptop GT 650M does 3-400M c/s.

magnumripper:
FWIW I saw this *once* while running TS
```
form=raw-md4-opencl               guesses:    0 -show=   0 unk unk : Expected count(s) (1500)  [!!!FAILED!!!  return code 256]
Self test failed (cmp_one(1))
```
...but I can't reproduce it. This was with CPU device and TS sets LWS=8 and GWS=64 (although this device will force LWS down to 1).

Sayantan2048:
I getting some issues with cpu devices too. Apparently GWS=512 and LWS=64 seems to be the minimum limit.

magnumripper:
But there is no particular reason there would be such limit, right? Maybe it's just another barrier needed somewhere or something like that.

magnumripper:
@Sayantan2048 I think this patch fixes the performance counters for good, please review
```
diff --git a/src/cracker.c b/src/cracker.c
index 45fb099..4aadc71 100644
--- a/src/cracker.c
+++ b/src/cracker.c
@@ -47,6 +47,7 @@
 #include "recovery.h"
 #include "external.h"
 #include "options.h"
+#include "mask_ext.h"
 #include "mask.h"
 #include "unicode.h"
 #include "john.h"
@@ -747,7 +748,8 @@ static int crk_salt_loop(void)
 	} while ((salt = salt->next));
 
 	if (done >= 0)
-		add32to64(&status.cands, crk_key_index);
+		add32to64(&status.cands, crk_key_index *
+		          mask_int_cand.num_int_cand);
 
 	if (salt)
 		return 1;
```
This code assumes `mask_int_cand.num_int_cand` is *always* 1 unless GPU generation is active. In particular, it has to be 1 even if mask mode was not used at all.

magnumripper:
[...]

PR #1131, opened 2015-03-21
Mask mode implementation in raw-sha256-opencl

claudioandre:
Discussion opened:
* there are 2 'remove me' that need some analysis before merge.

### DIFF
--- a/src/common-gpu.c
+++ b/src/common-gpu.c
@@ -7,7 +7,7 @@
  * Copyright (c) 2010-2012 Samuele Giovanni Tonon <samu at linuxasylum dot net>
  * Copyright (c) 2010-2013 Lukas Odzioba <ukasz@openwall.net>
  * Copyright (c) 2010-2013 magnum
- * Copyright (c) 2012-2013 Claudio André <claudioandre.br at gmail.com>
+ * Copyright (c) 2012-2015 Claudio André <claudioandre.br at gmail.com>
  * and is hereby released to the general public under the following terms:
  *    Redistribution and use in source and binary forms, with or without
  *    modifications, are permitted.
--- a/src/common-opencl.c
+++ b/src/common-opencl.c
@@ -7,7 +7,7 @@
  * Copyright (c) 2010-2012 Samuele Giovanni Tonon <samu at linuxasylum dot net>
  * Copyright (c) 2010-2013 Lukas Odzioba <ukasz@openwall.net>
  * Copyright (c) 2010-2013 magnum
- * Copyright (c) 2012-2013 Claudio André <claudioandre.br at gmail.com>
+ * Copyright (c) 2012-2015 Claudio André <claudioandre.br at gmail.com>
  * and is hereby released to the general public under the following terms:
  *    Redistribution and use in source and binary forms, with or without
  *    modifications, are permitted.
--- a/src/opencl/sha256_kernel.cl
+++ b/src/opencl/sha256_kernel.cl
@@ -12,6 +12,7 @@
  */
 
 #include "opencl_rawsha256.h"
+#include "opencl_mask_extras.h"
 
 inline void _memcpy(               uint32_t * dest,
                     __global const uint32_t * src,
@@ -21,7 +22,7 @@ inline void _memcpy(               uint32_t * dest,
         *dest++ = *src++;
 }
 
-inline uint32_t sha256_block(uint32_t * buffer, int total) {
+inline void sha256_block(uint32_t * buffer, int total, uint32_t * H) {
     uint32_t a = H0;
     uint32_t b = H1;
     uint32_t c = H2;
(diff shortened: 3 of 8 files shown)


claudioandre:
Tests (on well):
* ../tests/jtrts.pl  raw-sha256-opencl raw-sha512-opencl xsha512-opencl
* ../run/john ~/testhashes -form=raw-sha256-opencl -wo:/home/claudio/solucao.txt
* ../run/john ~/testhashes -form=raw-sha512-opencl -wo:/home/claudio/solucao.txt
* ../run/john ~/testhashes -form=xsha512-opencl    -wo:/home/claudio/solucao.txt
* rm ../run/*.pot
* ../run/john ~/testhashes -form=raw-sha256-opencl --mask=cla_?a?a?l?d?d?d
* ../run/john ~/testhashes -form=raw-sha256-opencl --wo:/home/claudio/wl.txt --mask=?w?d?d
* ../run/john ~/testhashes -form=raw-sha256-opencl --mask=pass -dev=0 --skip
* ../run/john ~/testhashes -form=raw-sha256-opencl --mask=?d?d?u?l?s?l?l 
* ../run/john ~/testhashes -form=raw-sha256-opencl --mask=[Pp][Aa@][Ss5][Ss5][Ww][Oo0][Rr][Dd]
* ../run/john ~/testhashes -form=raw-sha256-opencl --mask=?u?u?uBERT__17
* ../run/john ~/testhashes -form=raw-sha256-opencl -mask:mot?Arhead
* ../run/john ~/testhashes -form=raw-sha256-opencl -mask:?d
* ../run/john ~/testhashes -form=raw-sha256-opencl -mask:?l?l?l?l?l?l?l
* GWS=262144 ../run/john ~/testhashes -form=raw-sha256-opencl -mask:?u?u?u?u?u?u?u
* ../run/john ~/testhashes -form=raw-sha256-opencl -mask:?a?a?a?a?a?a?a?a (still running)

Sayantan2048:
How much performance benefit does using mask mode yield ?

claudioandre:
>How much performance benefit does using mask mode yield ?

A lot. For example: 
* here Radeon 6770 -> 2G
* well Radeon 7770 -> 6G

A 'regular' run cannot be much better than 60M.

PS: and the tuning is not made considering a mask run.

magnumripper:
Cool! That "get key error" stuff will be discussed in #1037.

magnumripper:
I see 833Mp/s at best on Super. That's with a manual GWS of 64K or higher. How do you achieve 6G? BTW oclHashcat can only do 1.1Gp/s on HD7970 which makes 6G sound a bit too good to be correct.

claudioandre:
This is because I got the first number I see in the screen and I get the
C/S. I suppose you are interested only in the c/s numbers.

2015-03-21 20:44 GMT-03:00 magnum <notifications@github.com>:

magnumripper:
How is the C/s number the first you see? It's written last. It doesn't tell squat about speed. I can get a CPU format to report tens of GC/s.

claudioandre:
I usually read the status from right to left. 
***
Well, the C/S is important to me because it is able to measure how fast the kernel is discarding candidates (and this is an important detail in the implementation). But, I should paid more attention to what you guys are expecting.

claudioandre:
> I would love to see a PoC commit for any format with the bare minimum of changes to get GPU-side 

This is almost that (host code changes a lot). Anyway, sooner or latter, sha512 will be the next victim.

claudioandre:
>  @Sayantan2048's code (eg. in md4_kernel.cl) is pretty much uncommented and you just sorted it out?

He did a good job solving the problem, I only had to document it.

claudioandre:
> GWS of 64K or higher. [...] BTW oclHashcat can only do 1.1Gp/s

In order to get the best we can, we probably need to tune mask_int_cand_target (since this is the important number in a pure mask run).

claudioandre:
> How hard would it be to make mask mode crack partial hashes

Not sure if I get your point. Well, thinking about, for example, LinkedIn zeros in (part of) the hash, it is basically the same problem solved by Sayantan's mask mode using a few cycles less. 
* You are going to compute almost a full raw-md? or sha??? or whatever. 
* Then
  * Compare, for example, if H[0] has the value you are expecting.

It is less than the usual work, but, not much less.

So, I don't expect it is going to be much faster than a 'normal' run on a masked GPU format.

claudioandre:
>> How hard would it be to make mask mode crack partial hashes

>So, I don't expect it to be much faster than a 'normal' run on a masked GPU format.

If you are thinking about false positives and --keep-guessing, it will be more complicated then this because you will need to accept and go on cracking. More buffers and host code, but my feeling is that Sayantan did the hard work for you.

magnumripper:
> Well, the C/S is important to me because it is able to measure how fast the kernel is discarding candidates (and this is an important detail in the implementation). But, I should paid more attention to what you guys are expecting.

The C/s figure is fine to use, but then it need to be accompanied by what number of hashes were loaded. Otherwise it doesn't say much.

PR #1161, opened 2015-03-26
report error message and exit when --input-encoding= a long string

### LINKED ISSUE #1157: --input-encoding problem
# 1. Compare the two cases

Both Case 1 and Case 2 are invalid.

## 1.1 Case 1
$ ./john  7z_fmt --input-encoding=11

**Invalid encoding**. Supported encodings:
ASCII, UTF-8, ISO-8859-1 (or latin1), ISO-8859-2, ISO-8859-7, ISO-8859-15, KOI8-R,
CP437, CP720, CP737, CP850, CP852, CP858, CP866, CP868,
CP1250, CP1251, CP1252, CP1253, CP1254, CP1255, CP1256

## 1.2 Case 2
$ ./john 7z_fmt --input-encoding=11111111111111111111111111111111111

Loaded 1 password hash (7z, 7-Zip [SHA256 AES 32/64])
Will run 8 OpenMP threads
Note: This format may emit false positives, so it will keep trying even after
finding a possible candidate.
password         (?)

# 2. Debug

I think john should **report error message** and **exit** in Case 2.

## 2.1 unicode::cp_name2id()

Location:  unicode.c    int  cp_name2id(char *encoding)

```C
 int cp_name2id(char *encoding)
 {
         char enc[16] = "";
         char *d = enc;
 
         if (!encoding || strlen(encoding) > sizeof(enc))
                return CP_UNDEF;
         ...

         if (!strcmp(enc, "raw") || !strcmp(enc, "ascii"))
               return ASCII;
         else

         fprintf(stderr, "Invalid encoding. Supported encodings:\n");
         listEncodings(stderr);
         exit(EXIT_FAILURE);
}
```

16 == sizeof(enc), so the cp_name2id() will return **CP_UNDEF** for **long encoding** in Case 2. 
Why not print error message?

loverszhaokai:
closes #1157.

### DIFF
--- a/src/unicode.c
+++ b/src/unicode.c
@@ -958,7 +958,9 @@ int cp_name2id(char *encoding)
 	char enc[16] = "";
 	char *d = enc;
 
-	if (!encoding || strlen(encoding) > sizeof(enc))
+	if (!encoding)
+		return CP_UNDEF;
+	if (strlen(encoding) > sizeof(enc))
 		goto err;
 
 	/* Strip iso prefix */
--- a/src/unicode.c
+++ b/src/unicode.c
@@ -959,7 +959,7 @@ int cp_name2id(char *encoding)
 	char *d = enc;
 
 	if (!encoding || strlen(encoding) > sizeof(enc))
-		return CP_UNDEF;
+		goto err;
 
 	/* Strip iso prefix */
 	if (!strncasecmp(encoding, "iso-", 4))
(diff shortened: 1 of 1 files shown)


src/unicode.c
```diff
@@ -959,7 +959,7 @@ int cp_name2id(char *encoding)
 	char *d = enc;
 
 	if (!encoding || strlen(encoding) > sizeof(enc))
-		return CP_UNDEF;
+		goto err;
```
magnumripper: This might lead to regression problems. If encoding == NULL we should still return CP_UNDEF.

PR #1171, opened 2015-04-01
Uses less VGPRs on GCN

claudioandre:


claudioandre:
Are you aware that Travis is failing?

magnumripper:
Yep, fixed now. Two problems, neither seen here. Those bots do a good job.

PR #1174, opened 2015-04-02
Pre GSoC robustness work

claudioandre:
- It removes the static code from headers.
- Only one set of tests for all formats.


magnumripper:
I cherry-picked 7cd2439. I guess 6b2b733 is the one you sent me. The two middle ones need manual merging, I don't have time right now.

PR #1175, opened 2015-04-02
Used shared code between CPU and GPU formats for raw-SHA512.

claudioandre:
(No static methods in header file).

Let's keep it around before merging.


magnumripper:
> Let's keep it around before merging.

I don't dare doing that :)

PR #1177, opened 2015-04-03
Use pseudo-intrinsics for sse-intrinsics.c

magnumripper:
...making nearly all remaining SIMD formats AVX2/AVX512 capable.

Except it doesn't work quite yet ;-)

TO DO: Find all instances of hard-coded vector sizes, and
replace them with SIMD_COEF_32/64 (or the shorter names
VS32 and VS64 within sse-intrinsics.c).


magnumripper:
Created this PR to document progress. We can't merge this until it builds fine on at least AVX2 and Xeon Phi, and we should regression test older SIMD (eg. just SSE2, AVX or XOP) as well as 32-bit and non-SIMD before merging.

I just rebased the avx2 branch onto latest bleeding-jumbo.

src/dynamic_types.h
```diff
@@ -183,7 +183,7 @@ typedef struct private_subformat_data
 # endif
 # endif
 # define LOOP_STR
-# if SIMD_COEF_32 == 4
+# if SIMD_COEF_32 >= 4
 #  ifdef MD5_SSE_PARA
 #   define ALGORITHM_NAME		"128/128 " MD5_SSE_type  " " STRINGIZE(BY_X) "x4x" STRINGIZE(MD5_SSE_PARA)
 #   define BSD_BLKS (MD5_SSE_PARA)
```
magnumripper: All the above should probably have their own clauses for AVX2 and for MIC, where the latter might need lower OMP_SCALE.

src/nt2_fmt_plug.c
```diff
@@ -508,7 +508,7 @@ static void set_key_utf8(char *_key, int index)
 		keybuf_word += SIMD_COEF_32;
 	}
 
-	((unsigned int *)saved_key)[14*SIMD_COEF_32 + (index&3) + (index>>2)*16*SIMD_COEF_32] = len << 4;
+	((unsigned int *)saved_key)[14*SIMD_COEF_32 + (index&3) + (index>>SIMD_COEF32_BITS)*16*SIMD_COEF_32] = len << 4;
```
magnumripper: I bet fixes like the ones above are needed in a LOT of formats. @jfoug already did many of them (in GETPOS macros and the like) but not all.

src/pseudo_intrinsics.h
```diff
@@ -68,12 +77,9 @@ typedef __m512i vtype;
 #define vunpacklo_epi32         _mm512_unpacklo_epi32
 #define vunpacklo_epi64         _mm512_unpacklo_epi64
 #define vxor                    _mm512_xor_si512
-#define vpermute4x64_epi64      _mm512_permute4x64_epi64
-#define vpermute2x128           _mm512_permute2x128_si512
-#define vset_epi32              _mm512_set_epi32
 
 // MIC doesn't support _mm512_shuffle_epi8
-#if __MIC__   
+#if __MIC__
```
magnumripper: @zzlei please don't commit trailing whitespace

magnumripper:
BTW just as a curious exercise I'm tempted to support MMX to, using intrinsics. But I guess it will need a lot of emulation for unsupported intrinsics. And it will only ever build on 32-bit so is hard for me to pull off.

magnumripper:
At this point the majority of formats work just fine - except for dynamic, which only works for a few subformats. @jfoug you have a lot to do... I can't believe how much stuff you have hard coded. But most of it are systematic bugs so fixes progress pretty quick (I have fixed about 180 formats last couple of days).

magnumripper:
BTW @zzlei if you test this with the mic, use `-form:cpu-dynamic` - this will skip all dynamic formats. Some of them segfaults.

magnumripper:
#1189 #1190 #1192 open. The rest of the formats pass not only self-test with AVX2 but also the Test Suite.

magnumripper:
@zzlei from now on, please use the "intrinsics" branch instead. It's currently rebased onto latest bleeding jumbo and lots of commits were squashed into fewer ones.

Please take care not to push local merges to that (or any) branch.

PR #1186, opened 2015-04-06
Fix for the intermitent mask error on CPU.

claudioandre:
- see: https://github.com/magnumripper/JohnTheRipper/issues/1037#issuecomment-71936052

1. There are other ways to go. I saw some performance penalty trying the obvious one.
2. md4 is up to you
3. Tests on more hardware is still pending.


magnumripper:
This looks like a good solution. I suppose this finishes quicker than transfer, even in mask mode.

claudioandre:
> I suppose this finishes quicker than transfer, even in mask mode.

This was what happened here.

PR #1194, opened 2015-04-09
Use pseudo-intrinsics for sse-intrinsics.c, with AVX2 and AVX512 (MIC) support.

magnumripper:
As of this writing, everything works on AVX2 except dynamic (well dynamic 0 and 30 does work - the rest are currently disabled in john.conf).


magnumripper:
I'm in process of rebasing this onto latest bleeding-jumbo and will then merge. The Dynamic format will be disabled for AVX2 builds, unless `./configure --enable-experimental-code`

PR #1198, opened 2015-04-09
Add eCryptfs file version 2

sylvainpelissier:
Since version 105 (See See https://launchpad.net/ecryptfs/+milestone/105) eCryptfs uses a new file version with the salt included in the file. I had the new file version parsing capability to JTR.


magnumripper:
Do we need to change the crack format too or is this change all that's needed?

magnumripper:
Thanks!

sylvainpelissier:
You're welcome, the crack format has not changed. The salt is now stored inside the file by default with a prefix to indicate the version.

PR #1211, opened 2015-04-19
Refactor code that uses __AVX512__

zzlei:
also remove some non-existent intrinsics in AVX512 or MIC


magnumripper:
I missed that you seem to have added a `.pseudo_intrinsics.h.swp` by accident (right?)

magnumripper:
Dropped it in 072cc30

PR #1212, opened 2015-04-20
Some optimization with gather/scatter

zzlei:


src/pseudo_intrinsics.h
```diff
@@ -45,20 +45,12 @@ typedef __m512i vtype;
 #define vand                    _mm512_and_si512
 #define vandnot                 _mm512_andnot_si512
 #define vcmov(y, z, x)          vxor(z, vand(x, vxor(y, z)))
-#if __MIC__
 /*
  * NOTE: AVX2 has it as (base, index, scale) while MIC and AVX512 are
- * different (even from each other).
+ * different.
  */
 #define vgather_epi32(b, i, s)  _mm512_i32gather_epi32(i, (void*)(b), s)
```
zzlei: This syntax works fine for AVX512, too. Tested.

src/pseudo_intrinsics.h
```diff
@@ -110,10 +102,12 @@ typedef __m512i vtype;
...
 }
 
-#define GATHER64(x, y, z)                                \
-{                                                        \
-    x = vset_epi64x(y[7][z], y[6][z], y[5][z], y[4][z],  \
-                    y[3][z], y[2][z], y[1][z], y[0][z]); \
+#define GATHER64(x, y, z)                                               \
+{                                                                       \
+    uint64_t stride = sizeof(*y);                                       \
+    vtype indices = vset_epi64x(7*stride, 6*stride, 5*stride, 4*stride, \
```
zzlei: Those multiplications should be pre-computed in compile time (?)

src/pseudo_intrinsics.h
```diff
@@ -133,10 +127,10 @@ typedef __m512i vtype;
                                     0x0c0d0e0f, 0x08090a0b,     \
                                     0x04050607, 0x00010203))
 #define vswap64(n) \
-    n = vshuffle_epi8(n, vset_epi64(0x38393a3b3c3d3e3f, 0x3031323334353637, \
-                                    0x28292a2b2c2d2e2f, 0x2021222324252627, \
-                                    0x18191a1b1c1d1e1f, 0x1011121314151617, \
-                                    0x08090a0b0c0d0e0f, 0x0001020304050607))
+    n = vshuffle_epi8(n, vset_epi64x(0x38393a3b3c3d3e3f, 0x3031323334353637, \
```
zzlei: Yet another confusing intrinsic name.

src/pseudo_intrinsics.h
magnumripper: I'm pretty sure we can safely assume that.

src/pseudo_intrinsics.h
```diff
@@ -133,10 +127,10 @@ typedef __m512i vtype;
...
+                                     0x28292a2b2c2d2e2f, 0x2021222324252627, \
+                                     0x18191a1b1c1d1e1f, 0x1011121314151617, \
+                                     0x08090a0b0c0d0e0f, 0x0001020304050607))
```
magnumripper: Not that a 32-bit AVX512 build makes much sense, but we should probably hang ULL tags on 64-bit constants.

PR #1213, opened 2015-04-21
append 'ULL' to some int64 literals

zzlei:


src/pseudo_intrinsics.h
```diff
@@ -127,10 +127,14 @@ typedef __m512i vtype;
                                     0x0c0d0e0f, 0x08090a0b,     \
                                     0x04050607, 0x00010203))
 #define vswap64(n) \
-    n = vshuffle_epi8(n, vset_epi64x(0x38393a3b3c3d3e3f, 0x3031323334353637, \
-                                     0x28292a2b2c2d2e2f, 0x2021222324252627, \
-                                     0x18191a1b1c1d1e1f, 0x1011121314151617, \
-                                     0x08090a0b0c0d0e0f, 0x0001020304050607))
```
zzlei: I don't have a 32-bit platform to experiment with. But the absence of `ULL` causes no trouble when using the '-m32' flag on a 64-bit platform.

magnumripper: I'm not sure when it's really needed but better safe than sorry

PR #1221, opened 2015-04-24
unify the interface of vcmpeq_epi8

zzlei:


src/pseudo_intrinsics.h
```diff
@@ -45,6 +45,7 @@ typedef __m512i vtype;
 #define vand                    _mm512_and_si512
 #define vandnot                 _mm512_andnot_si512
 #define vcmov(y, z, x)          vxor(z, vand(x, vxor(y, z)))
+#define vcmpeq_epi8_mask        (uint64_t)_mm512_cmpeq_epi8_mask
```
zzlei: I found the following typedefs in icc's headers:
```
typedef unsigned char       __mmask8;
typedef unsigned short      __mmask16;
typedef unsigned int        __mmask32;
typedef unsigned __int64    __mmask64;
```
So there's actually no need for something like `vmask2int`.

src/rawSHA1_ng_fmt_plug.c
```diff
@@ -330,12 +330,10 @@ static void sha1_fmt_set_key(char *key, int index)
 	vtype  Z   = vsetzero();
 	vtype  X   = vloadu(key);
 	vtype  B;
-#if __AVX512F__ || __MIC__
-	uint32_t len = strlen(key);
-#else
-	uint32_t mask = vmovemask_epi8(vcmpeq_epi8(X, Z));
+
+	// FIXME: even uint64_t won't be long enough for AVX-1024
+	uint64_t mask = vcmpeq_epi8_mask(X, Z);
```
zzlei: This replacement might have worse performance than `strlen` if `vcmpeq_epi8_mask` is emulated. But the code gets clearer. I'm not sure which is more appropriate.

magnumripper: Maybe you should benchmark the impact on MIC, and re-add an `#if __MIC__` if needed. Try benchmarking wordlist.c too (eg. just exhausting some huge wordlist without rules against a dummy hash). Perhaps MIC should not use intriniscs at all in wordlist.c (there's a 64-bit alternative below it which is almost as fast).

zzlei: I experimented that with raw-sha1-ng, and got some strange result.

Using `strlen`:
```
Will run 240 OpenMP threads
Benchmarking: Raw-SHA1-ng, (pwlen <= 55) [SHA1 512/512 MIC 16x]... (240xOMP) DONE
Raw:	4869K c/s real, 1408K c/s virtual
```
Using `vcmpeq_epi8_mask`:
```
Will run 240 OpenMP threads
Benchmarking: Raw-SHA1-ng, (pwlen <= 55) [SHA1 512/512 MIC 16x]... (240xOMP) DONE
Raw:	1137K c/s real, 723488 c/s virtual
```

The "real" is slower, but "virtual" gets significantly higher. 

Hints? 
(I still have a poor understanding of "real"/"virtual"...)

magnumripper: No, virtual is also slower (from 1408K ro 723K).

Consider this:
```
$ time sleep 1

real	0m1.010s
user	0m0.002s
sys	0m0.003s
```
Here, "real" is wall clock while "user + sys" is how much CPU time you actually used.

Raw shows `crypts / wall clock time` while virtual shows `crypts / system time used` where "system time" is (I think) same as "user + sys" above. So if you measure on a host that is otherwise heavily loaded by other users, the "virtual" will reflect what speed you'd get if the system had not been overloaded.

When using GPU, you often get very high virtual figures because the host CPU was practically not involved. On the other hand when using OpenMP and say 4 threads, you may spend 4 CPU seconds in one "wall clock second" so the virtual figure will be lower than the real.

frank-dittrich: Without OMP (non-OMP builds or formats or OMP_NUM_THREADS set to 1), `virtual` reports what you could expect as the `real` c/s rate if there were no other load, i.e., if your system has no other load, the reported `virtual` c/s rate should be exactly the same as the `real` c/s rate, or only slightly higher.
With OMP, a system without any other load should report `real` c/s rates of about `virtual * OMP_NUM_THREADS`.

May be we should print `1408 Kc/s` instead of `1408K c/s`, to make the difference more obvious, but that's for Solar to decide. (Such a change would require either `benchmark-unify` or `relbench` and possibly other tools parsing `john --test` output to be adjusted.)

PR #1234, opened 2015-04-27
A few fixes which might need discussion

frank-dittrich:
1. report glibc version with --list=build-info
2. Add _BSD_SOURCE and _DEFAULT_SOURCE definitions to listconf.c for HAVE_OPENCL (We use setenv() here, but didn't have the feature request definitions...)
3. Fix some memory leaks for ./john --status
4. Report build timestamp with --list=build-info
5. Report MS compiler version with --list=build-info

### DIFF
--- a/src/listconf.c
+++ b/src/listconf.c
@@ -203,6 +203,24 @@ static void listconf_list_build_info(void)
 #if defined(__clang_version__) && !__INTEL_COMPILER
 	printf("clang version: %s\n", __clang_version__);
 #endif
+
+#ifdef _MSC_VER
+/*
+ * See https://msdn.microsoft.com/en-us/library/b0084kay.aspx
+ * Currently, _MSC_BUILD is not reported, but we could convert
+ * _MSC_FULL_VER 150020706 and _MSC_BUILD 1 into a string
+ * "15.00.20706.01".
+ */
+#ifdef _MSC_FULL_VER
+	printf("Microsoft compiler version: %d\n", _MSC_FULL_VER);
+#else
+	printf("Microsoft compiler version: %d\n", _MSC_VER);
+#endif
+#ifdef __CLR_VER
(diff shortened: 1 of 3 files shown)


src/listconf.c
```diff
@@ -63,6 +67,8 @@ extern void cuda_device_list();
 #endif
 #if HAVE_OPENCL
 #include "common-opencl.h"
+#define _BSD_SOURCE 1
+#define _DEFAULT_SOURCE 1
```
magnumripper: Feature macros are supposed to be defined prior to including *any* system header so these should be moved up above the `#define NEED_OS_FLOCK`.

magnumripper:
I'm not overly excited about recovery.c; It's an error condition, no need to bother. But that's just me. The rest is fine (once you move the feature macros).

frank-dittrich:
@magnum I moved the _DEFAULT_SOURCE and _BSD_SOURCE definition up, and I added two more commits.
The build time stamp might be helpful to let users know how old their john version is (and that it might be a good idea to get a newer version).
I had no way of testing the "Add Microsoft compiler version info..." commit.
Who can test this, now that Jim is inactive?

magnumripper:
On Yosemite using a real gcc:
```
listconf.c:57:30: fatal error: gnu/libc-version.h: No such file or directory
 #include <gnu/libc-version.h>
                              ^
compilation terminated.
make[1]: *** [listconf.o] Error 1
make: *** [default] Error 2
```
I'll #ifdef that code away for now.

magnumripper:
Temporary band-aid in 3b54b18. Using __GNUC__ was not a good idea, I should have noticed that.

frank-dittrich:
Sorry for that.
Meanwhile, I also reverted the second part of that commit, to avoid a --list=build-info segfault under Linux.

PR #1240, opened 2015-04-28
Report glibc version with --list=build-info

frank-dittrich:
another try, this time hopefully correct

### DIFF
--- a/src/listconf.c
+++ b/src/listconf.c
@@ -53,11 +53,9 @@
 #endif
 #endif
 
-/* fails when building on gcc on non-GNU systems (eg. OSX)
-#if __GNUC__
+#if __GLIBC__
 #include <gnu/libc-version.h>
 #endif
-*/
 
 #include "regex.h"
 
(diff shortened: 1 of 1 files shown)


magnumripper:
So where does this `__GLIBC__` define come from? I thought we had a hen-or-egg situation.

magnumripper:
This works fine (does not include the sections) on my Mac.

frank-dittrich:
What libc do you have on Mac? Do you know how to report that libc version?
I tried to google that info, but didn't succeed.

frank-dittrich:
What if I have a libc at build time, but not at run time? Can this ever happen (and work)?

magnumripper:
I have absolutely no idea but on Macs it may simply be tied to the general OS X version (as in 10.10.3). Why would we care about libc version anyway? Is it really needed or are we bloating build-info?

frank-dittrich:
I added that glibc version info because the new version on Fedora 22 caused that compile time warning.
So I thought it might be helpful to diagnose problems.

magnumripper:
> What if I have a libc at build time, but not at run time? Can this ever happen (and work)?

You mean another version at run time. I'm pretty sure it can happen.

magnumripper:
> on Fedora 22 caused that compile time warning.

I think we should limit build-info to things that might help with run-time problems, or we'll end up with an amount of code/output that can't be justified.

frank-dittrich:
I didn't just mean another version.
I mean, you build with glibc, however, you don't have glibc at run time,but another libc implementation.
Unlikely, but possible.
May be we really should revert reporting of glibc version, or just drop reporting the runtime libc version.

magnumripper:
> I mean, you build with glibc, however, you don't have glibc at run time,but another libc implementation.

That would not execute at all. The dynamic linker wouldn't even try to resolve it.

frank-dittrich:
Then there is no problem (except bloat). Even a very old libc should have  gnu_get_libc_version().

PR #1241, opened 2015-04-29
remove the emulation of _mm512_cmp_epi8_mask

zzlei:
it's too slow under benchmarking


zzlei:
Somehow the use of `vcmpeq_epi8_mask` in wordlist.c causes no compile error when building for MIC. I left it intact for now.

PR #1249, opened 2015-05-01
MainProcess forward sigint/sigterm to children.

math07:
When I started doing the --fork option on Johnny, it seems to me that John only kills its children when ascii code 3 ( CTRL-c) or 'q'. is pushed on the keyboard, not with signals. I looked at the code in "signal.c" of JtR and it seems to be consistent with my tests. So now when main process receive sigterm/sigint, I forward it to its children.

If you want more details, here are the most important messages on the list where Shinnok, Frank and me talked about it:
http://www.openwall.com/lists/john-dev/2015/04/28/16
http://www.openwall.com/lists/john-dev/2015/04/30/4
http://www.openwall.com/lists/john-dev/2015/04/30/5
http://www.openwall.com/lists/john-dev/2015/04/30/6

I tested it but I'm not a john pro yet so I'm not sure to know if placing the forwarding at that place doesn't have unwished behaviour. So I'll take your comments.

### DIFF
--- a/src/signals.c
+++ b/src/signals.c
@@ -146,6 +146,16 @@ void check_abort(int be_async_signal_safe)
 
 static void sig_install_abort(void);
 
+#if OS_FORK
+static void signal_children(int signum)
+{
+	int i;
+	for (i = 0; i < john_child_count; i++)
+		if (john_child_pids[i])
+			kill(john_child_pids[i], signum);
+}
+#endif
+
 static void sig_handle_abort(int signum)
 {
 	int saved_errno = errno;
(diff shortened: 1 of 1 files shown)


magnumripper:
Thanks, I see no obvious problem with this but I'll think it over for a while.

magnumripper:
Oh, but we can't merge stuff to the master branch here. That branch is merely a one-way mirror of Solar's CVS repo for convenience. We need Solar to apply the patch to core and commit it to CVS.

I could apply it to bleeding-jumbo though. But if Solar is willing to add this to core that's much better, and it'd be merged to jumbo.

math07:
Update : it has now been fixed in CVS.

magnumripper:
OK, will merge when I get the time.

magnumripper:
Done

magnumripper:
> Done

It seems I forgot to push it to GitHub %-). But NOW it's done.

math07:
Thanks :)

PR #1252, opened 2015-05-02
mic.h: tune interleaving factors

zzlei:
MIC model: 5110P
icc version: 15.0.2

### DIFF
--- a/src/mic.h
+++ b/src/mic.h
@@ -60,8 +60,8 @@
 
 #define MD5_SSE_PARA		1
 #define MD5_N_STR			"16x"
-#define MD4_SSE_PARA		1
-#define MD4_N_STR			"16x"
+#define MD4_SSE_PARA		6
+#define MD4_N_STR			"16x6"
 #define SHA1_SSE_PARA		1
 #define SHA1_N_STR			"16x"
 


zzlei:
For MIC, performance is optimal when `MD5_SSE_PARA` and `SHA1_SSE_PARA` stay set to 1.

magnumripper:
Was this with 240 threads or (60 forks x 4 threads), or something else?

magnumripper:
I think the MD4 para showed too little improvement to be conclusive, we probably have some bottleneck scewing the figures. Also, MD4 and MD5 usually ends up with the same para.

zzlei:
> Was this with 240 threads or (60 forks x 4 threads), or something else?

240 threads. 
BTW, what is the proper way to run in forked mode? The following command doesn't work:
```
$ jumbo/john --fork=2 --test --format=md5crypt
Invalid options combination or duplicate option: "--fork=2"
```

frank-dittrich:
So far, --fork doesn't work with --test.
For benchmarks, you can extract the top two test hashes into a file

```
./john --list=format-tests --format=md5crypt | cut -f 3 | head -n 2 > hashes.md5crypt
```
And then either test some wordlist which doesn't include the passwords for those hashes, "0123456789ABCDE", and "test":

```
./john --format=md5crypt hashes.md5crypt --fork=2 --wordlist=mywordlist
```

Or, use an internal mode and avoid length switching /which would impact performance), and interrupt that after some time.

```
./john --format=md5crypt hashes.md5crypt --incremental=upper --min-length=8 --max-length=8
```

For this md5crypt example, you wouldn't need to specify `=upper`, but for most other formats `=upper` will make it less likely that you crack some of the test hashes in your performance test.

magnumripper:
Right. I use to run eg. `-mask=?l?l?l?l?l?l?l?l -max-run=10` for benchmarking a real hash.

Issue #1250, opened 2015-05-01
Cygwin build problem


magnumripper:
Reported by Rich Rumble
```
./configure --enable-mpi &&  make clean && make
...
Configured for building John the Ripper jumbo:

Target CPU .................................. i686 AVX, 32-bit LE
AES-NI support .............................. depends on OpenSSL
Target OS ................................... cygwin
Cross compiling ............................. no
Legacy arch header .......................... x86-sse.h

Optional libraries/features found:
Experimental code ........................... no
OpenMPI support (default disabled) .......... yes
Fork support ................................ yes
OpenMP support .............................. yes
OpenCL support .............................. no
CUDA support ................................ no
Generic crypt(3) format ..................... yes
Rexgen (extra cracking mode) ................ no
GMP (PRINCE mode and faster SRP formats) .... yes
128-bit integer (faster PRINCE mode) ........ no
PCAP (vncpcap2john and SIPdump) ............. no
BZ2 (gpg2john extra decompression logic) .... yes
Memory map (share/page large files) ......... yes

Development options (these may hurt performance when enabled):
Memdbg memory debugging settings ............ disabled
AddressSanitizer ("ASan") ................... disabled
....

ERROR:

d-declarations -Wno-format-extra-args -std=gnu89 -D_GNU_SOURCE
-fopenmp -DHAVE_MPI    -funroll-loops -DJTR_MODE
-Wno-declaration-after-statement -std=c99 -c pp.c
pp.c: In function ‘do_prince_crack’:
pp.c:1466:22: warning: implicit declaration of function ‘fileno’
[-Wimplicit-function-declaration]
                      fileno(read_fp), 0);
                      ^
mpicc -DAC_BUILT -march=native -mavx -c -g -O2 -DARCH_LITTLE_ENDIAN=1
 -Wall -Wdeclaration-after-statement -fomit-frame-pointer --param
allow-store-data-races=0 -Wno-deprecated-declarations
-Wno-format-extra-args -std=gnu89 -D_GNU_SOURCE    -fopenmp -DHAVE_MPI
   -funroll-loops c3_fmt.c -o c3_fmt.o
mpicc -DAC_BUILT -march=native -mavx -c -g -O2 -DARCH_LITTLE_ENDIAN=1
 -Wall -Wdeclaration-after-statement -fomit-frame-pointer --param
allow-store-data-races=0 -Wno-deprecated-declarations
[...]

PR #1332, opened 2015-05-18
Define only _GNU_SOURCE for Cygwin builds.

shinnok:
See issue #1250 for details:
https://github.com/magnumripper/JohnTheRipper/issues/1250

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -13057,22 +13057,6 @@ fi
     ;;
 
   linux*|cygwin*)
-    # For exposing fileno()
-
-   for i in -D_POSIX_SOURCE; do
-      jtr_list_add_dupe=0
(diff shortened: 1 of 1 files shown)


magnumripper:
This is not the correct way to edit configure. The source is configure.ac - edit that one, then build configure using eg. `autoreconf`. I can do it if you don't have autoconf installed.

shinnok:
@magnumripper I'm not sure these options were in the .ac in the first place? Can you take a look?

magnumripper:
Actually you'd edit `m4/jtr_systems_specific_logic.m4` which is sourced by configure.ac.

shinnok:
Got it. I'll do that properly tomorrow.

magnumripper:
Jim fixed it

PR #1338, opened 2015-05-21
Fix: wpapcap2john.c buffer overflow

loverszhaokai:
1. Change full_packet to malloc(). The packet size may be bigger than 65535.

2. `pkt_hdr.incl_len <= pkt_hdr.orig_len`,

```C
typedef struct pcaprec_hdr_s {
        uint32 ts_sec;         /* timestamp seconds */
        uint32 ts_usec;        /* timestamp microseconds */
        uint32 incl_len;       /* number of octets of packet saved in file */
        uint32 orig_len;       /* actual length of packet */
} pcaprec_hdr_t;
```

The incl_len is the packet size, so use incl_len when copy packet.

### DIFF
--- a/src/wpapcap2john.c
+++ b/src/wpapcap2john.c
@@ -28,7 +28,7 @@ static void DumpKey(int idx, int one_three, int bIsQOS);
 
 static uint32 start_t, start_u, cur_t, cur_u;
 static pcaprec_hdr_t pkt_hdr;
-static uint8 full_packet[65535];
+static uint8 *full_packet;
 static uint8 *packet;
 static int bROT;
 static WPA4way_t wpa[MAX_ESSIDS];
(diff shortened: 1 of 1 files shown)


magnumripper:
If you mention `__LINE__` you'd better mention `__FILE__` rather than `__FUNCTION__` (or both). Also, I would print what size was attempted. If it says it tried to allocate 27 GB you'd know the problem is not your available memory but a corrupt file...
```diff
 	if (NULL == full_packet) {
-		fprintf(stderr, "%s:%d: malloc failed\n", __FUNCTION__, __LINE__);
+		fprintf(stderr, "%s:%d: malloc failed allocating %zu bytes\n", __FILE__, __LINE__, pkt_hdr.incl_len);
 		exit(EXIT_FAILURE);
 	}
```

magnumripper:
BTW that was just my spontaneous comments/suggestions, not a requirement. Do as you wish!

magnumripper:
Oh, I see now there were similar prints before this commit. Never mind, I will do some changes!

magnumripper:
Tweaked it in b2d2c97

PR #1349, opened 2015-05-23
add interleaving to SHA256 & SHA512 (experimental)

zzlei:


magnumripper:
Thanks. I'd rather not have that local merge of yours committed upstream so I rebased it away. You'll need to `pull --rebase` now when updating your fork from upstream.

PR #1371, opened 2015-05-29
keepass plaintext length

smarteam:
as per http://openwall.com/lists/john-users/2015/05/29/3 plaintext length is set to 32, but it works fine with 100


magnumripper:
The actual limit (in this implementation) is 125, so I set that in c02e04c instead. But thanks anyway!

PR #1373, opened 2015-06-01
use scatter instruction in sse-intrinsics

zzlei:
tested on MIC, no noticeable performance improvement

### DIFF
--- a/src/pseudo_intrinsics.h
+++ b/src/pseudo_intrinsics.h
@@ -54,6 +54,8 @@ typedef __m512i vtype;
 #define vload(x)                _mm512_load_si512((void*)(x))
 #define vloadu(x)               _mm512_loadu_si512((void*)(x))
 #define vor                     _mm512_or_si512
+#define vscatter_epi32          _mm512_i32scatter_epi32
+#define vscatter_epi64          _mm512_i64scatter_epi64
 #define vset1_epi8              _mm512_set1_epi8
 #define vset1_epi32             _mm512_set1_epi32
 #define vset1_epi64x            _mm512_set1_epi64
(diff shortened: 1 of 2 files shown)


magnumripper:
Right. Are there more instances of vswap we need to fix?

magnumripper:
OK we have one
```C
#define vswap64(n)                                          \
  {                                                         \
    n = vshufflehi_epi16(vshufflelo_epi16(n, 0xb1), 0xb1);  \
    n = vxor(vslli_epi16(n, 8), vsrli_epi16(n, 8));         \
    n = vshuffle_epi32(n, 0xb1);                            \
  }
```
Will this not "return" the right thing though? I think in gcc it will. The last statement will be the value.

zzlei:
This works:
```c
int a = 0;
int b = ({a += 1, a += 2, a += 3});
```
while this doesn't:
```c
int a = 0;
int b = {a += 1, a += 2, a += 3};
```
The parentheses around the curly braces are necessary. So every time we use `vswap64` defined as above, we must guarantee it's enclosed by parentheses.

magnumripper:
Cool. Is this guaranteed by the C standards? I have a feeling we're on somewhat thin ice here.

We should add the parens to the definitions.

zzlei:
Well, I made a mistake in my last comment. 
I intended to write `{a += 1; a += 2; a += 3;}` instead of `{a += 1, a += 2, a += 3}`. 
And actually `(a += 1, a + 2, a += 3)` would work, which is a standard use of comma operator.

magnumripper:
I have some faint memory we're not guaranteed execution order when using a comma. I might be wrong.

magnumripper:
It seems I was wrong, commas should be fine. The whole thing will evaluate to whatever is after the last comma.

magnumripper:
Fixed the SSE2 version of vswap64 in a0407fe

PR #1382, opened 2015-06-05
Fix misspelled words

loverszhaokai:
Fix misspelled words found by [checkpatch.pl](https://github.com/loverszhaokai/JohnTheRipper/blob/fuzzing/run/checkpatch.pl)

### DIFF
--- a/src/gost.c
+++ b/src/gost.c
@@ -24,7 +24,7 @@ extern unsigned rhash_gost_sbox_cryptpro[4][256];
  * Initialize algorithm context before calculaing hash
  * with test parameters set.
  *
- * @param ctx context to initalize
+ * @param ctx context to initailize
  */
 void john_gost_init(gost_ctx *ctx)
 {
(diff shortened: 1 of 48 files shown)


src/gost.c
```diff
@@ -24,7 +24,7 @@ extern unsigned rhash_gost_sbox_cryptpro[4][256];
  * Initialize algorithm context before calculaing hash
  * with test parameters set.
  *
- * @param ctx context to initalize
+ * @param ctx context to initailize
```
magnumripper: Hey, these replace one typo with another!

loverszhaokai: Sorry, my mistake.

magnumripper:
This was an astonishing amount of typos, even in official AMD header stuff!

Fun fact of the day: You do get google hits for ADL_ERR_INVALID_DIPLAY_IDX but not for ADL_ERR_INVALID_DISPLAY_IDX! We should change it anyway of course. Just fix that "initailize" typo.

loverszhaokai:
@magnumripper  Thanks for reminding. I should have **rechecked** again. :+1:

PR #1440, opened 2015-06-16
MinGW64 Warning Muting

jfoug:
This also converts all %lld (x u) into LLd (or LLx LLu).  Also all %zu and %zd are converted into Zu and Zd

This allows compilers to have these defined properly and drastically reduce the number of warnings.

NOTE, for MingW64, in the *printf function, it is not legal to use %I64d.  If that is done, the compiler (says) that it is only handling it as %64d (which is bad, and if in a sscanf, will likely core).  But MingW throws warnings in other user var args functions, like log_alert(...).  Those warnings are still there, but I have yet to find a way to fix that.


jfoug:
I have to try to figure out why CircleCI fails. It says it can not find the CircleCI-MinGW-sh file (but I can see it).

magnumripper:
Note: 7be84bf and 85bd245 are local merges, you must not merge this branch into bleeding-jumbo without rebasing them away first.

Never merge topic branches with latest bleeding-jumbo. There is no good in doing that. Just let them lag behind, or rebase.

magnumripper:
> I have to try to figure out why CircleCI fails

It might be due to merging magnum-jumbo, if that is what you did. That must have caused lots of weirdness.

magnumripper:
Hmmm and even stranger: The first commit of this PR is from December 2011. How did that happen?!

jfoug:
Dhiru thinks the circlci fail is due to the fork name being wrong. I will work on it in the morning.

jfoug:
I will not pull from here. I would rather just push the changes. I'm,using a fork so I can see circleci build to make sure it's ok. The only mingw64 I have is fedora.

jfoug:
The rename of my local repo to JohnTheRipper (it was magnum-jumbo) has gotten past the problem on CircleCI.  Now if this builds properly, I will close out this pull request (it was only made to get the build working on CircleCI), and I will drop the changes into magnumripper/JohnTheRipper.   I just wanted a MinGW64 auto-test environment other than Fedora, which I know is happy with the changes.

jfoug:
The build gets the same warnings as I see on Fedora.  The log_event(...) has warnning issues with %lld %zu while the *printf() works fine.  NOTE, the warnings received on the *printf() for %I64* are very bad, and I am pretty sure using those would have serious issues.

I will look closer.  There may be some pragma which causes the *printf() class of functions to 'work' properly warning wise, that we may have to somehow add to our function declaration for log_event(...) function on mingw64 to remove the warnings.

However, the mingw64 build on circleCI was happy with my changes, and looks the same.

PR #1443, opened 2015-06-17
applied changes to match version 1.2.4 of rexgen

teeshop:


magnumripper:
Oh, I should read all my mail before committing code :laughing: 

I already applied this. Thanks!

PR #1465, opened 2015-06-24
tune several formats' OMP_SCALE for MIC

zzlei:


frank-dittrich:
I think you need to `git pull --rebase` locally instead of pushing your local merges.

zzlei:
Yeah, it was a mistake. I already closed that pull request and created a new one.

PR #1487, opened 2015-07-02
Add --enable-tsan

frank-dittrich:
TODO: avoid trying to combine --enable-asan and --enable-tsan,
since -fsanitize=address and -fsanitize=thread cannot be combined

TODO: -enable-tsan shouldn't be allowed for --disable-openmp or
when the compiler doesn't support -fopenmp

@magnumripper please also see the comments in the pull request I closed:
https://github.com/magnumripper/JohnTheRipper/pull/1486

### DIFF
--- a/src/bench.c
+++ b/src/bench.c
@@ -524,7 +524,7 @@ int benchmark_all(void)
 AGAIN:
 #endif
 	total = failed = 0;
-#if defined(WITH_ASAN) || defined(WITH_UBSAN) || defined(DEBUG)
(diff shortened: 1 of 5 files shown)


frank-dittrich:
May be TSan isn't that useful for john.
It produces lots of messages, but john --test still succeeds, albeit it takes an extremely long time.
I also tried a TSan core build. lots of TSan messages, no problems, not with --test, not with real cracking runs.

So, I'm in doubt whether we should att --enable-tsan.

frank-dittrich:
I think that --enable-tsan is currently of no use for john, so I'll close the pull request.

PR #1524, opened 2015-07-09
add AltiVec intrinsics

zzlei:
There're still a bunch of temporary workarounds, which will be fixed later.

### DIFF
--- a/src/ppc64.h
+++ b/src/ppc64.h
@@ -52,4 +52,21 @@
 #define BF_SCALE			0
 #define BF_X2				0
 
+#define SIMD_COEF_32		4
+#define SIMD_COEF_64		2
+
+#define SIMD_PARA_MD4		1
+#define SIMD_PARA_MD5		1
+#define SIMD_PARA_SHA1		1
+#define SIMD_PARA_SHA256	1
+#define SIMD_PARA_SHA512	1
+
+#define MD4_N_STR			"4x"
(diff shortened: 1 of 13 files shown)


src/ppc64.h
```diff
@@ -52,4 +52,21 @@
...
 #define BF_X2				0
 
+#define SIMD_COEF_32		4
+#define SIMD_COEF_64		2
+
+#define SIMD_PARA_MD4		1
+#define SIMD_PARA_MD5		1
+#define SIMD_PARA_SHA1		1
+#define SIMD_PARA_SHA256	1
+#define SIMD_PARA_SHA512	1
```
magnumripper: These should be wrapped in `#ifndef SIMD_PARA_xx`, remember?

src/ppc64.h
```diff
@@ -52,4 +52,21 @@
...
+
+#define MD4_N_STR			"4x"
+#define MD5_N_STR			"4x"
+#define SHA1_N_STR			"4x"
+#define SHA256_N_STR		"4x"
+#define SHA512_N_STR		"4x"
```
magnumripper: And these should be using the macros resulting in eg. "4x3" for para 3

zzlei: Oh, it's kind of mess up there. I'll clean up the code a bit and send another pull request.

PR #1528, opened 2015-07-09
Add AltiVec intrinsics

zzlei:


src/ppc64.h
```diff
@@ -52,4 +52,55 @@
...
+#define SIMD_PARA_SHA1		1
+#endif
+#ifndef SIMD_PARA_SHA256
+#define SIMD_PARA_SHA256	1
+#endif
+#ifndef SIMD_PARA_SHA512
+#define SIMD_PARA_SHA512	1
+#endif
+
+#define STR_VALUE(arg)			#arg
```
zzlei: I think the code from line 74 to line 102 should be somehow reused for different archs, instead of being copied & pasted whenever a new arch is supported.

magnumripper: Yeah we could move it into sse-intrinsics.h

PR #1529, opened 2015-07-09
Explain DBError exception.

rbrenton:
Expose the underlying exception in the error log, causing the open to fail.


magnumripper:
Thanks!

rbrenton:
You're fast! Thanks as well.

PR #1542, opened 2015-07-12
Just to check if it will hurt on CI

claudioandre:
I'm not confident yet!

### DIFF
--- a/src/common-opencl.c
+++ b/src/common-opencl.c
@@ -267,6 +267,8 @@ static char *opencl_driver_info(int sequential_id)
 		{1526, 3},
 		{1573, 4},
 		{1642, 5},
+		{1702, 3},
+		{1729, 3},
 		{0, 0}
 	};
 
@@ -281,6 +283,8 @@ static char *opencl_driver_info(int sequential_id)
 		"14.6 beta (Mantle)",
 		"14.9 (Mantle) [recommended]",
 		"14.12 (Omega) [supported]",
+		"15.05 [buggy]",
+		"15.07 [recommended]",
 		""
 	};
 	clGetDeviceInfo(devices[sequential_id], CL_DRIVER_VERSION,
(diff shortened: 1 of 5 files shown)


claudioandre:
This warning about UTF-8 is new:
```
../run/john ~/testhashes -form=raw-sha512-opencl -mask:mot?Arhead
../run/john: /usr/local/cuda/lib64/libOpenCL.so.1: no version information available (required by ../run/john)
Loaded 16 password hashes with no different salts (Raw-SHA512-opencl [SHA512 OpenCL])
Warning: invalid UTF-8 seen reading $JOHN/john.pot
Remaining 1 password hash
Device 0: Tahiti [AMD Radeon HD 7900 Series]
Using Mask Mode with internal candidate generation, global worksize(GWS) set to 16384
mot�rhead           (?)
1g 0:00:00:00  2.380g/s 228.5p/s 228.5c/s 228.5C/s GPU:46°C util:59% fan:20% mot#rhead
```

magnumripper:
You get that warning as a result of running non-UTF8 without telling john about it. You should use `--enc:latin1` or whatever applies to your hashes' encoding. I bet that would also fix the output for that test hash: The password is very likely "motörhead".

frank-dittrich:
@claudioandre That warning started with  commit 90e21c5bae90258f7b6420cda8cc4a72b725c8ac, about two months ago.
It changed default encoding settings in john.conf

claudioandre:
Oh, I can see the warning is harmless.

claudioandre:
This seems good enough to be merge **after** the contest.

magnumripper:
For faster testing, I'm merging now. The TS get seriously slowed down from `--list=format-details` being so slow without this.

magnumripper:
See https://github.com/magnumripper/jtrTestSuite/issues/58#issuecomment-127463701

PR #1544, opened 2015-07-12
pseudo_intrinsics.h: optimize vtesteq & vtestz

zzlei:


src/pseudo_intrinsics.h
```diff
@@ -289,6 +284,9 @@ typedef __m256i vtype;
 #define vunpacklo_epi64         _mm256_unpacklo_epi64
 #define vxor                    _mm256_xor_si256
 
+#define vtesteq_epi32(x, y)     vmovemask_epi8(vcmpeq_epi32(x, y))
```
zzlei: `vtesteq_epi32` returns a non-zero value when it evaluates to true, but the value is not necessarily 1. This might be a potential problem (for user code that relies on it to return either 0 or 1), but it works fine so far.

magnumripper: > The new version of vtesteq_epi32 returns a non-zero value when it evaluates to true, but the value is not necessarily 1

I've seen code that goes around this using double bang, as in
```
#define vtesteq_epi32(x, y)     !!vmovemask_epi8(vcmpeq_epi32(x, y))
```
But I'm not sure how good that is

PR #1565, opened 2015-07-19
rename sse-intrinsics to simd-intrinsics

zzlei:
also renamed all related functions, e.g. SSESHA1body => SIMDSHA1body


zzlei:
I think it's probably time to change those names, as many non-SSE archs are involved already.

PR #1566, opened 2015-07-20
Implement --fuzz options to fuzz formats' functions

loverszhaokai:
Implement --fuzz options to fuzz formats' functions: prepare(), valid(), init(), split(), binary(), salt(), salt_hash()


src/loader.c
```diff
@@ -301,7 +301,7 @@ static void ldr_set_encoding(struct fmt_main *format)
 
 	/* john.conf alternative for --internal-encoding */
 	if (options.flags &
-	    (FLG_RULES | FLG_SINGLE_CHK | FLG_BATCH_CHK | FLG_MASK_CHK))
+	    (FLG_RULES | FLG_SINGLE_CHK | FLG_BATCH_CHK | FLG_MASK_CHK | FLG_FUZZ_CHK))
```
magnumripper: I'm not sure the above makes any sense. It's supposed to happen when we use rules or mask mode.

magnumripper:
Perhaps this feature should be optional, eg. using --with-fuzzer option to configure, and #ifdef'd out otherwise?

src/loader.c
loverszhaokai: I add this because there is a bug. But I can not reproduce now. I will remove this change. If the bug  comes back, I will fix it then.

loverszhaokai:
I created a new pr for the fuzz option which is optional and have #ifdef HAVE_FUZZ

PR #1591, opened 2015-07-29
Add NEON intrinsics

zzlei:
Please **DON'T** merge this pull request. I'm bringing here for some discussions.

### DIFF
--- a/src/pseudo_intrinsics.h
+++ b/src/pseudo_intrinsics.h
@@ -46,9 +46,13 @@ typedef vtype64 vtype;
 #define vadd_epi64              vaddq_u64
 #define vand                    vandq_u32
 #define vandnot                 vbicq_u32
-#define vcmov(x, y, z)          vxor(y, vand(z, vxor(x, y)))
+#define vcmov                   vcmov_emu
 #define vload(m)                vld1q_u32((uint32_t*)(m))
+#define vloadu                  vloadu_emu
 #define vor                     vorrq_u32
+#define vroti_epi32             vroti_epi32_emu
+#define vroti_epi64             vroti_epi64_emu
+#define vroti16_epi32           vroti_epi32
 #define vset1_epi32(x)          vdupq_n_u32(x)
 #define vset1_epi64(x)          vdupq_n_u64(x)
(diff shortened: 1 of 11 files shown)


src/pseudo_intrinsics.h
```diff
@@ -490,35 +555,14 @@ typedef __m64i vtype;
 #endif /* __SSE3__ || __MIC__ */
```
zzlei: Does this kind of optimization really make sense?

src/pseudo_intrinsics.h
```diff
@@ -34,8 +34,57 @@
...
+#ifdef __ARM_NEON__
+#include <arm_neon.h>
+
+typedef uint32x4_t vtype32;
+typedef uint64x2_t vtype64;
+typedef vtype64 vtype;
+
+#define vadd_epi32              vaddq_u32
+#define vadd_epi64              vaddq_u64
+#define vand                    vandq_u32
```
zzlei: This is troublesome, just like in AltiVec. Although `vandq_u32` and `vandq_u64` translate to the same instruction, they have different function signatures, thus are not interchangeable.

I'm thinking maybe we should have different versions of `vand`, `vload`, etc, which would make it more natural to write AltiVec and NEON intrinsics. So far only x86 uses the type-agnostic form of those intrinsics. 

We could also distinguish between `vtype32` and `vtype64`, which may improve readability (knowing what exactly type of data we are dealing with).

zzlei:
I refactored pseudo_intrinsics.h a bit. Those `#ifdef`s for CPU dispatching make me kind of dizzy...

I think there's still space for improvement of readability. I'll try to refactor it further after NEON support is completed.

src/pseudo_intrinsics.h
magnumripper: You mean the one "for really old CPUs"? I'm not sure of its origin but I guess it makes sense. Anything that can save a single CPU cycle is...

magnumripper: BTW please not that the ternary function is supposed to be optimized away completely, so the code will be just either of the functions in each place this macro is used. Maybe you didn't realize that.

zzlei: So the argument `s` is always a constant in the code. That makes sense.

PR #1593, opened 2015-07-30
Improvements to the .gitignore file

shinnok:


shinnok:
Sort was done per group and with the sort utility, hope it's what you hinted at.

PR #1665, opened 2015-08-15
Add SIMD support to episerver

zzlei:


magnumripper:
You need to rebase, we don't want that local merge

magnumripper:
Oh I see you already did. No need to close this PR and open another one though.

PR #1671, opened 2015-08-16
Merge of raw-sha1 and nsldap (and start of common code for GPU)

jfoug:
This is a merge of raw-sha1 and nsldap.  The 'default' value stored into the .pot file is now the {SHA}base-64 type (nsldap), and not the $dyanmic_26$ which used to be default within raw-sha1.

However, the format properly handles {SHA}, $dynamc-26$ and 'raw' 40 byte hex (upper and mixed case), in .pot files, and also in input files.

I have moved some code over into a new common file (and header). The same merging needs to be done on the OpenCL side (there is raw-sha1-opencl and raw-sha1-nsldap formats).  However, I do not know which format would be the best to use as the 'new' format, to merge them into 1.  I will leave that task to someone else.  However, they should use the common raw-sha1 code, so that any changes to valid, prepare/split, etc only have to be done one time.


jfoug:
Not sure why checks are failing after changing the attributes to 644?  Looking at travis, it claims the header file is simply not there ???

magnumripper:
The file is named rawSHA1_common.h and you try to include rawsha1_common.h.

Please squash these two commits (as well as this new fix) into a single commit. And perhaps fix your umask once and for all?

PR #1674, opened 2015-08-16
Merge of raw-sha1 and nsldap, and start of common raw-sha1 code for GPU

jfoug:
This is a merge of raw-sha1 and nsldap. The 'default' value stored into the .pot file is now the {SHA}base-64 type (nsldap), and not the $dyanmic_26$ which used to be default within raw-sha1.

However, the format properly handles {SHA}, $dynamc-26$ and 'raw' 40 byte hex (upper and mixed case), in .pot files, and also in input files.

I have moved some code over into a new common file (and header). The same merging needs to be done on the OpenCL side (there is raw-sha1-opencl and raw-sha1-nsldap formats). However, I do not know which format would be the best to use as the 'new' format, to merge them into 1. I will leave that task to someone else. However, they should use the common raw-sha1 code, so that any changes to valid, prepare/split, etc only have to be done one time.


jfoug:
This code is ready, I just did it this way so that @magnumripper and @frank-dittrich could have a look to make sure they were ok with it.

Prior:

raw-sha1   Handled raw and dyna_26, output to .pot dyna_26.  Only handled dyna_26 from pot
nsldap       Handles {SHA} only

New:
nsldap (gone)
raw-sha1  handles raw, dyna_26, and {SHA} from input and .pot file.  Writes ONLY {SHA} to the .pot file.

magnumripper:
Looks good to me, though I guess we should modify the raw-sha1-ng format too, and perhaps even the linkedin format.

jfoug:
We can add those 2 to the list of items still left open on the issue.  I really wonder if we should drop the LI format (or put it into unused) ?

magnumripper:
People are still using it for research and stuff. I even did so myself not long ago: It's kind of cool when you find a new wordlist/rule combo (or mode, like --prince) that cracks a bunch of LinkedIn stuff never revealed before.

jfoug:
If that is the case, then we probably should make an issue to get the LI format 'fixed', and working properly like it is with raw-sha1.

magnumripper:
I tried to add reversing steps to LinkedIn but that made it "impossible" (well not impossible, but harder) to use source() and we really want it for that format. So it stayed without reversing steps.

jfoug:
I was not talking reversing. I thought there were problems for some PARA values.

magnumripper:
That issue was closed long ago, cff90698

PR #1682, opened 2015-08-18
Boost 7z formats' many-salts speed a *lot*

### LINKED ISSUE #1679: 7-zip doesn't use any salt within its KDF
...so we can do the same trick we do with WPA-PSK, boosting the "many salts" speed infinitely.

However, let's wait until we have merged @zzlei's SIMD code first (or maybe he implements this too)

magnumripper:
It's KDF is effectively unsalted as long as the iteration count doesn't change. Closes #1679.

To see the boost for OpenCL in benchmark, you currently need to use a longer test time, eg. --test=5 (or more, depending on GPU). This is actually needed to see the full boost even with the CPU format.

@zzlei I made this a PR so not to disturb your SIMD work. You get to decide if you want to merge it before your SIMD changes (and rebase your SIMD branch), or after (and we'll manually sort out the conflicts).

### DIFF
--- a/src/7z_fmt_plug.c
+++ b/src/7z_fmt_plug.c
@@ -40,7 +40,7 @@ john_register_one(&fmt_sevenzip);
 #define TAG_LENGTH		4
 #define ALGORITHM_NAME		"SHA256 AES 32/" ARCH_BITS_STR
 #define BENCHMARK_COMMENT	" (512K iterations)"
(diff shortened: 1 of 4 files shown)


zzlei:
Sorry, I kind of messed up the workflow. First I sent a pull request with my updates on 7z, and then sent another after merging your code with mine. This is a bit cumbersome, but should work...

So I think this PR is not needed now.

PR #1695, opened 2015-08-21
Avoid strcpy if key is already aligned in set_key()

zzlei:
as discussed in https://github.com/magnumripper/JohnTheRipper/pull/1693


magnumripper:
After e1d4dd3, benchmark speeds might improve since we're always using an aligned buffer when benchmarking.

PR #1703, opened 2015-08-24
Add SIMD support to RAR

zzlei:
and work around the alignment issue of stack arrays


zzlei:
The test suite is passed.

PR #1736, opened 2015-09-08
Detect FMT_SPLIT_UNIFIES_CASE error

loverszhaokai:


loverszhaokai:
@jfoug 

Thanks for your patches for detecting FMT_SPLIT error. Should I change the **"PASS" to "FAILED"** when
there is FMT_SPLIT warning ?

magnumripper:
If this had been the normal self-test it should definitely be just a non-fatal warning. But maybe nothing in --test-full should be mere warnings? I'm not sure.

magnumripper:
We're not counting/summarizing warnings. That is an argument for actually give FAILED for *any* problem in `-test-full`. An alternative would be to start counting and summarizing warnings but that may be overkill.

loverszhaokai:
@magnumripper  Thanks. I think I should change it to "FAILED".

loverszhaokai:
@jfoug @magnumripper  

When there is a FMT_SPLIT warning or error, change to "FAILED" by: https://github.com/magnumripper/JohnTheRipper/pull/1741

jfoug:
Good, I agree that in -test=full, it should not be a warning, but an error.  In that mode, the user is certainly doing an expensive test.  I really think however that -test-full probably should be removed from the normal options, and placed into the 'hidden' set.

magnumripper:
I think you are now failing but still printing "Warning" for the two latter cases? I didn't try it though.

loverszhaokai:
@magnumripper 

> I think you are now failing but still printing "Warning" for the two latter cases? I didn't try it though.

Yes.

```
[...]
Testing: dynamic=md5($p) [256/256 AVX2 8x3]... FAILED (Warning: dynamic=md5($p) should not set FMT_SPLIT_UNIFIES_CASE)
[...]
1 out of 335 tests have FAILED
```

magnumripper:
> I really think however that -test-full probably should be removed from the normal options, and placed into the 'hidden' set.

I also think we should put most of this in `#if !JTR_RELEASE_BUILD` or someting like that. Maybe Solar disagrees, and for that matter this is not a lot of code.

magnumripper:
>> I think you are now failing but still printing "Warning" for the two latter cases? I didn't try it though.

> Yes.

And actually I think for the first case, FAILED will end up printed twice, won't it? I think you need to fix all three. Just give a description of the problem and let the caller add FAILED to it.

loverszhaokai:
@magnumripper 

> And actually I think for the first case, FAILED will end up printed twice, won't it? I think you need to fix all three. Just give a description of the problem and let the caller add FAILED to it.

OK.

magnumripper:
Now that I got the attention, it's like this:
```
Benchmarking: dynamic=md5($p) [128/128 AVX 4x3]... FAILED (dynamic=md5($p) should not set FMT_SPLIT_UNIFIES_CASE)
```
I think the repetition of the format name is redundant and should be dropped. This probably affects most of your fails.

magnumripper:
I think it should be
```
Benchmarking: dynamic=md5($p) [128/128 AVX 4x3]... FAILED (should not set FMT_SPLIT_UNIFIES_CASE)
```

loverszhaokai:
@magnumripper  

I will change it soon. : )

jfoug:
It's kind of funny.  These have been problems for a LONG time, and no one has complained a bit.  Toss out a warning, and they get fixed in hours.

PR #1741, opened 2015-09-10
Change to 'FAILED' when there is a FMT_SPLIT warning or error

loverszhaokai:


jfoug:
While this was passing the CI's it would be failing (I have not looked at logic, but 'assume' it works), if a -test-full was done, which IS in the end, how we should setup the CI's to run.

However, with 724f9f3 I have short term eliminated split_case from compiled dynamics.  I still do believe that compiled dynamics 'should' have casing, so I am really thinking we are missing the deep test for when a format SHOULD have casing, but does not.

I am going to dig into the test-full code, because I think I know how to make this test find the formats which SHOULD either set casing or change their valids.

BUT as it stands, I think this PR is ready to merge.  Since @magnumripper is not on holiday, I will let him do it.

jfoug:
Actually, since I will be working with this code right now, I am going to merge.

I am going to add a 4th case, and this is the one I think Solar has been saying we should have.  It will detect times when there is no split, no bit set, BUT the format is behaving in a manner where there should be one.

loverszhaokai:
@jfoug  

> I am going to add a 4th case

It's great. :+1:  Thanks.

jfoug:
the 'proto' code for this case is:

```C
f=split(c)
f=prepare(c)
if (valid(f)) {
    f2 = strdup(f)
    b=bin(f2)
    h=tohex(b);
    if(strstr(f2,h)) {
      strupr(h)
      overwrite(f2,h) // only the matching part.
      if (split(f2)!=f && valid(f2)) BUG=4;
    } else {
       strupr(h)
       if(strstr(f2,h)) {
          strupr(h)
          overwrite(f2,h) // only the matching part.
          if (split(f2)!=f && valid(f2)) BUG=4;
        }
   }
}
```

Or something kind of like that (I am not done, so not 100% sure).

What this does is find what I consider to be the 'normal' bug.  That is where there is a hex string for the binary. Since a hex string is valid with abcdef or ABCDEF (or a mix), then when that is wrong, it causes the hashes to be viewed as 'independant' items, and stored twice, not good, and why this flag was made.   So what I am doing, is trying to force this bug to happen, in a controlled manner.

loverszhaokai:
@jfoug 

> Since a hex string is valid with abcdef or ABCDEF (or a mix), then when that is wrong, it causes the hashes to be viewed as 'independant' items, and stored twice, not good, and why this flag was made.

```C
/* Splits a ciphertext into several pieces and returns the piece with given
 * index, starting from 0 (will usually return the ciphertext unchanged).
 * For hex-encoded hashes which are compared by the target system/application
 * irrespective of the case of characters (upper/lower/mixed) used in their
 * encoding, split() must unify the case (e.g., convert to all-lowercase)
 * and FMT_SPLIT_UNIFIES_CASE must be set. */
    char *(*split)(char *ciphertext, int index, struct fmt_main *self);
```

From the comments of split(), I think whether split() should unify the case is determined by the **target system/application**.

Is the **split()** has any relation with **whether the binary() is case-sensitive** ?

jfoug:
I am not sure you are getting the point.  Here is the problem:

hash = md5
password = password

inputfile

5f4dcc3b5aa765d61d8327deb882cf99
5f4DCC3b5aa765d61d8327deb882cf99
5f4DCc3b5aa765d61d8327deb882cf99
5f4Dcc3b5aa765d61d8327deb882cf99

How many hashes are in the input file?

If you say 1, you are correct.  But if a format is 'broken' it will say 4. That is what this flag was added for.

Now, a format can handle this in 2 ways.  One, it can unify the case within split().  The other is it can reject hashes in valid()   The first way is better IMHO.  Since you may have some hashes that are in up and some that are in low case, but none that are duplicated like my test shows above.

So the test is this.  See if I can find the hex representation of a call to binary() within the hash. If so, change the case of THAT section.  Then either split should change it back, OR valid should say the hash is not valid. Any other result, and john core code will treat this as 2 separate hashes.

loverszhaokai:
@jfoug 

> hash = md5
password = password

>inputfile

>5f4dcc3b5aa765d61d8327deb882cf99
5f4DCC3b5aa765d61d8327deb882cf99
5f4DCc3b5aa765d61d8327deb882cf99
5f4Dcc3b5aa765d61d8327deb882cf99

>How many hashes are in the input file?

>If you say 1, you are correct. But if a format is 'broken' it will say 4. That is what this flag was added for.

**All** the formats with **hex-encoded hashes** should add this flag ? I think not all. Is that right ?

frank-dittrich:
If the system which creates these hashes is using either upper case or lower case, the format should not accept the other case. (There are already too many formats which would grab the same raw hashes as valid.)
The same applies if the hashes a format accepts have been generated by one of out *2john tools.
In that case, we know exactly what to expect.
So, I don't agree that accepting any mix of upper and lower case (and using split() to unify case) is always better than just rejecting hashes in valid().

jfoug:
I also agree.  If there IS a defined way the hash comes in, then valid() should reject anything NOT meeting that criteria.

It is the formats which do not have one or the other methods.

Again, if a format was run on my 'sample' data I show above (well real data that has been 'morphed' in the same way), and it says that there are more than 1 hash, then the format has the bug.

All I am trying to do, is to see if I can find a hex string that matches what the binary() function returns, and if so, flip the case of that string, and then make sure that either split() fixes it back to what we started with OR the valid() function says this is not a valid hash.  Either method shows the format is working properly, and will not load duplicate hashes with mixed case.

loverszhaokai:
@jfoug 

1. Why split() before prepare() ?
2. I add some comments in the following code. Are they right ?
```C
f=split(c)
f=prepare(f)  // in an array.
if (valid(f)) {
    // e.g. f="5f4dcc"
    f2 = strdup(f)
    //     f2="5f4dcc"
    b=bin(f2)
    h=tohex(b);
    if(strstr(f2,h)) {
      strupr(h)
      overwrite(f2,h) // only the matching part.
      // f2="5F4DCC"
      if (split(f2)!=f2 && valid(f2)) {
          // There is #4 ERROR if (split(f2) = "5f4dcc" != "5F4DCC") and valid("5F4DCC") == 1
          BUG=4;
      }
    } else {
       [...]
   }
   mem_free(f2);
}
```

jfoug:
This was part of what I as asking for help for.  I wanted to make this 
code 100% compatible with the file loader code.  I believe this is the 
order of operations in the file loaded.   It does a split(), then 
prepare() then if valid() is yes, it uses the hash.  It runs this was 
for input, but for john.pot it only uses split.

Now, I did this, because I wanted to KNOW that valid is not rejecting 
the hash. If valid rejects the hash, then we should not be calling 
binary() on it.

 From that point on, in the code, I am only calling split and valid, so 
I am still following assertions required by the functions being used.  
The assertion for binary() is that the hash string has been passed 
through valid() and valid was happy with it.

frank-dittrich:
prepare() is executed before valid(), and split is called after valid().

jfoug:
thank you, then that order should be preserved here.

jfoug:
Looking at things, it may be that this new test method is really the only thing that should be tested.

@frank-dittrich this bug has been driven by you for the last decade. I think you are probably the person who is most intimately aware of this problem and all ramifications. 

Are there problems which my method of attack would not turn up?  If so, it may be best to simply use this validation in the end, and simply ignore the flag totally.  The format is either working properly (if this test is fully comprehensive), or it is NOT working properly. Telling the user what is wrong is not required.  Simply listing that the case duplication bug is present should suffice.

jfoug:
This is the code.  I have run it and watched on the formspring.  Now, dynamic only returns 16 bytes for binary (all dyna work that way), so the entire sha256 hash is not returned, BUT the first 16 byte are, and the code finds it just fine, then changes case and smashes the buffer, then runs it.  Split() cleans things up just fine.

Now, also with the testing section, you will also notice that if split leaves the string alone, that the code also calls valid. ONLY if valid returns true, do we treat this as a bug.  

```C
// check for common case problem.  hash is HEX, so find it, change it,
// and is there is no split, or split() does not fix it, then check
// valid().  If valid allows the hash, THEN we have a #4 problem.
flds[1] = cipher_copy;
ret = format->methods.prepare(flds, format);
if (format->methods.valid(ret, format)) {
	char *cp;
	int do_test=0;
	ret = format->methods.split(ret, 0, format);
	ret_copy = strdup(ret);
	bin = format->methods.binary(ret_copy);
	bin_hex = mem_alloc(format->params.binary_size*2 + 4);
	base64_convert(bin, e_b64_raw, format->params.binary_size, bin_hex, e_b64_hex, format->params.binary_size*2+1, 0);
	cp = strstr(ret_copy, bin_hex);
	if (cp) {
		do_test = 1;
		strupr(bin_hex);
		memcpy(cp, bin_hex, strlen(bin_hex));
	} else {
		strupr(bin_hex);
		cp = strstr(ret_copy, bin_hex);
		if (cp) {
			do_test = 1;
			strlwr(bin_hex);
			memcpy(cp, bin_hex, strlen(bin_hex));
		}
	}
	if (do_test) {
		ret = format->methods.split(ret_copy, 0, format);
		if (!strcmp(ret_copy, ret)) {
			if (format->methods.valid(ret_copy, format)) {
				// we have the bug!
				MEM_FREE(ret_copy);
				MEM_FREE(cipher_copy);
				*is_split_unifies_case = 4;
				return;
			}
		}
	}
}
```

jfoug:
NOW, we have results :+1: 

I will post what I see, and let @frank-dittrich validate that yes, these really are the problem items (lots of them)

jfoug:
NOTE, I will have to figure these 2 out.  They did not show up prior to this adjustment.
It could be something I am doing that 'smashes' the format (possibly not calling init properly
or not clearing).  Working with dyna can be a nightmare.  It's usually trivial things, but 
finding out what they are is NOT trivial.

Testing: dynamic_40 [sha1($s.pad20($p)) (net-sha1) 32/32]... FAILED (cmp_all(1))
Testing: dynamic_1016 [md5($p.$s) (long salt) 32/32 (MD5_body)]... FAILED (cmp_all(1))

Testing: OpenBSD-SoftRAID (8192 iterations) [PBKDF2-SHA1 32/32]... FAILED (split() case, or valid() should fail)
Testing: dynamic=md5($p) 32/32 (MD5_body) [dynamic_6000]... FAILED (split() case, or valid() should fail)
Testing: PBKDF2-HMAC-MD5 [PBKDF2-MD5 32/32]... FAILED (split() case, or valid() should fail)
Testing: Oracle12C [PBKDF2-SHA512 32/32]... FAILED (split() case, or valid() should fail)
Testing: has-160 [HAS-160 32/32]... FAILED (split() case, or valid() should fail)
Testing: pomelo [POMELO 32/32]... FAILED (split() case, or valid() should fail)
Testing: tc_aes_xts, TrueCrypt AES256_XTS [SHA512 32/32 /RIPEMD160/WHIRLPOOL]... FAILED (split() case, or valid() should fail)
Testing: tc_ripemd160, TrueCrypt AES256_XTS [RIPEMD160 32/32]... FAILED (split() case, or valid() should fail)
Testing: tc_sha512, TrueCrypt AES256_XTS [SHA512 32/32]... FAILED (split() case, or valid() should fail)
Testing: tc_whirlpool, TrueCrypt AES256_XTS [WHIRLPOOL 32/32]... FAILED (split() case, or valid() should fail)
Testing: PFX, PKCS12 (.pfx, .p12) [32/32]... FAILED (split() case, or valid() should fail)
Testing: SSH (one 2048-bit RSA and one 1024-bit DSA key) [RSA/DSA 32/32]... FAILED (split() case, or valid() should fail)
Testing: rsvp, HMAC-MD5 / HMAC-SHA1, RSVP, IS-IS [MD5 32/32]... FAILED (split() case, or valid() should fail)
Testing: eigrp, EIGRP MD5 / HMAC-SHA-256 authentication [MD5 32/32]... FAILED (split() case, or valid() should fail)
Testing: Mozilla, Mozilla key3.db [SHA1 3DES 32/32]... FAILED (split() case, or valid() should fail)
Testing: vtp, "MD5 based authentication" VTP [MD5 32/32]... FAILED (split() case, or valid() should fail)
Testing: dmd5, DIGEST-MD5 C/R [MD5 32/32]... FAILED (split() case, or valid() should fail)
Testing: EPI, EPiServer SID [SHA1 32/32]... FAILED (split() case, or valid() should fail)
Testing: ipb2, Invision Power Board 2.x [MD5 32/32]... FAILED (split() case, or valid() should fail)
Testing: krb4, Kerberos v4 TGT [DES 32/32]... FAILED (split() case, or valid() should fail)
Testing: krb5, Kerberos v5 TGT [3DES 32/32]... FAILED (split() case, or valid() should fail)
Testing: KeePass [SHA256 AES 32/32 OpenSSL]... FAILED (split() case, or valid() should fail)
Testing: skey, S/Key [MD4/MD5/SHA1/RMD160 32/32]... FAILED (split() case, or valid() should fail)
Testing: agilekeychain, 1Password Agile Keychain [PBKDF2-SHA1 AES 32/32]... FAILED (split() case, or valid() should fail)
Testing: Blockchain, My Wallet (x10) [PBKDF2-SHA1 AES 32/32]... FAILED (split() case, or valid() should fail)
Testing: chap, iSCSI CHAP authentication [MD5 32/32]... FAILED (split() case, or valid() should fail)
Testing: cloudkeychain, 1Password Cloud Keychain [PBKDF2-SHA512 32/32]... FAILED (split() case, or valid() should fail)
Testing: dmg, Apple DMG [PBKDF2-SHA1 3DES/AES 32/32]... FAILED (split() case, or valid() should fail)
Testing: eCryptfs (65536x) [SHA512 32/32 OpenSSL]... FAILED (split() case, or valid() should fail)
Testing: EFS [PBKDF2-SHA1-efs-variant 3DES 32/32]... FAILED (split() case, or valid() should fail)
Testing: EncFS [PBKDF2-SHA1 AES/Blowfish 32/32]... FAILED (split() case, or valid() should fail)
Testing: fde, Android FDE [PBKDF2-SHA1 SHA256/AES 32/32]... FAILED (split() case, or valid() should fail)
Testing: gpg, OpenPGP / GnuPG Secret Key [32/32]... FAILED (split() case, or valid() should fail)
Testing: keychain, Mac OS X Keychain [PBKDF2-SHA1 3DES 32/32]... FAILED (split() case, or valid() should fail)
Testing: keyring, GNOME Keyring [SHA256 AES 32/32 OpenSSL]... FAILED (split() case, or valid() should fail)
Testing: keystore, Java KeyStore [SHA1 32/32]... FAILED (split() case, or valid() should fail)
Testing: kwallet, KDE KWallet [SHA1 32/32]... FAILED (split() case, or valid() should fail)
Testing: lotus5, Lotus Notes/Domino 5 [8/32 X2]... FAILED (split() case, or valid() should fail)
Testing: lotus85, Lotus Notes/Domino 8.5 [8/32]... FAILED (split() case, or valid() should fail)
Testing: MD2 [MD2 32/32]... FAILED (split() case, or valid() should fail)
Testing: MongoDB, system / network [MD5 32/32]... FAILED (split() case, or valid() should fail)
Testing: mssql, MS SQL [SHA1 32/32]... FAILED (split() case, or valid() should fail)
Testing: mssql05, MS SQL 2005 [SHA1 32/32]... FAILED (split() case, or valid() should fail)
Testing: mssql12, MS SQL 2012/2014 [SHA512 32/32 OpenSSL]... FAILED (split() case, or valid() should fail)
Testing: nk, Nuked-Klan CMS [SHA1 MD5 32/32]... FAILED (split() case, or valid() should fail)
Testing: o5logon, Oracle O5LOGON protocol [SHA1 AES 32/32]... FAILED (split() case, or valid() should fail)
Testing: ODF [SHA1/SHA256 BF/AES 32/32 OpenSSL]... FAILED (split() case, or valid() should fail)
Testing: Office, 2007/2010/2013 [SHA1 32/32 / SHA512 32/32 OpenSSL AES]... FAILED (split() case, or valid() should fail)
Testing: oldoffice, MS Office <= 2003 [MD5/SHA1 RC4 32/32]... FAILED (split() case, or valid() should fail)
Testing: openssl-enc, OpenSSL "enc" encryption [32/32]... FAILED (split() case, or valid() should fail)
Testing: oracle11, Oracle 11g [SHA1 32/32]... FAILED (split() case, or valid() should fail)
Testing: Panama [Panama 32/32]... FAILED (split() case, or valid() should fail)
Testing: PDF [MD5 SHA2 RC4/AES 32/32]... FAILED (split() case, or valid() should fail)
Testing: PKZIP [32/32]... FAILED (split() case, or valid() should fail)
Testing: postgres, PostgreSQL C/R [MD5 32/32]... FAILED (split() case, or valid() should fail)
Testing: PST, custom CRC-32 [32/32]... FAILED (split() case, or valid() should fail)
Testing: PuTTY, Private Key [SHA1/AES 32/32]... FAILED (split() case, or valid() should fail)
Testing: RACF [DES 32/32]... FAILED (split() case, or valid() should fail)
Testing: RAdmin, v2.x [MD5 32/32]... FAILED (split() case, or valid() should fail)
Testing: rar, RAR3 (4 characters) [SHA1 AES 32/32]... FAILED (split() case, or valid() should fail)
Testing: RAR5 [PBKDF2-SHA256 32/32 OpenSSL]... FAILED (split() case, or valid() should fail)
Testing: 7z, 7-Zip (512K iterations) [SHA256 32/32 AES]... FAILED (split() case, or valid() should fail)
Testing: Snefru-128 [32/32]... FAILED (split() case, or valid() should fail)
Testing: Snefru-256 [32/32]... FAILED (split() case, or valid() should fail)
Testing: SSH-ng [RSA/DSA 32/32]... FAILED (split() case, or valid() should fail)
Testing: STRIP, Password Manager [PBKDF2-SHA1 32/32]... FAILED (split() case, or valid() should fail)
Testing: sxc, StarOffice .sxc [SHA1 Blowfish 32/32]... FAILED (split() case, or valid() should fail)
Testing: Sybase-PROP [salted FEAL-8 32/32]... FAILED (split() case, or valid() should fail)
Testing: VNC [DES 32/32]... FAILED (split() case, or valid() should fail)
Testing: wbb3, WoltLab BB3 [SHA1 32/32]... FAILED (split() case, or valid() should fail)

jfoug:
If all of these are truly buggy (I think they are), then lets keep this code outside of core john, fix all of these, and ONLY place it into the jumbo tree (and core is solar wants it), once we get a 'clean' run.

That way, people will NOT be able to easily add formats that do not properly correct for this bug.

NOTE, many of these will need to have their valids changed, simply to return false on wrong case.  The rest must return only 1 case within the split.

jfoug:
Here is the patch (warning, it is reversed) 

```diff
--- 1/src/formats.c	Thu Sep 10 06:01:15 2015
+++ 2/src/formats.c	Thu Sep 10 10:39:59 2015
@@ -12,6 +12,7 @@
 #include "dyna_salt.h"
 #include "misc.h"
 #include "unicode.h"
+#include "base64_convert.h"
 #ifndef BENCH_BUILD
 #include "options.h"
 #else
@@ -863,13 +864,70 @@
 	*second_index = sp - max + 1;
 }
 
-static int test_fmt_split_unifies_case(struct fmt_main *format, char *ciphertext)
+/* settings for is_split_unifies_case
+ * case -1: casing happening, but every hash has been correct.
+ * case 0: no error found (or casing not being used)
+ * case 1: "should set FMT_SPLIT_UNIFIES_CASE"
+ * case 2: "should not set FMT_SPLIT_UNIFIES_CASE"
+ * case 3: "split() is only casing sometimes"
+ * case 4: "split() case, or valid() should fail"
+ */
+static void test_fmt_split_unifies_case(struct fmt_main *format, char *ciphertext, int *is_split_unifies_case, int call_cnt)
 {
-	char *cipher_copy, *ret;
+	char *cipher_copy, *ret, *bin_hex, *ret_copy;
+	void *bin;
 	int first_index, second_index, size, index;
 	int change_count = 0;
+	char fake_user[5] = "john";
+	char *flds[10] = {0,0,0,0,0,0,0,0,0,0};
+
+	if (*is_split_unifies_case>2) return; // already know all we need to know.
 
 	cipher_copy = strdup(ciphertext);
+
+	// check for common case problem.  hash is HEX, so find it, change it,
+	// and is there is no split, or split() does not fix it, then check
+	// valid().  If valid allows the hash, THEN we have a #4 problem.
+	flds[0] = fake_user;
+	flds[1] = cipher_copy;
+	ret = format->methods.prepare(flds, format);
+	if (format->methods.valid(ret, format)) {
+		char *cp;
+		int do_test=0;
+		ret = format->methods.split(ret, 0, format);
+		ret_copy = strdup(ret);
+		bin = format->methods.binary(ret_copy);
+		bin_hex = mem_alloc(format->params.binary_size*2 + 4);
+		base64_convert(bin, e_b64_raw, format->params.binary_size, bin_hex, e_b64_hex, format->params.binary_size*2+1, 0);
+		cp = strstr(ret_copy, bin_hex);
+		if (cp) {
+			do_test = 1;
+			strupr(bin_hex);
+			memcpy(cp, bin_hex, strlen(bin_hex));
+		} else {
+			strupr(bin_hex);
+			cp = strstr(ret_copy, bin_hex);
+			if (cp) {
+				do_test = 1;
+				strlwr(bin_hex);
+				memcpy(cp, bin_hex, strlen(bin_hex));
+			}
+		}
+		if (do_test) {
+			ret = format->methods.split(ret_copy, 0, format);
+			if (!strcmp(ret_copy, ret)) {
+				if (format->methods.valid(ret_copy, format)) {
+					// we have the bug!
+					MEM_FREE(ret_copy);
+					MEM_FREE(cipher_copy);
+					*is_split_unifies_case = 4;
+					return;
+				}
+			}
+		}
+
+	}
+
 	ret = format->methods.split(cipher_copy, 0, format);
 	if (strcmp(cipher_copy, ret)) {
 		get_longest_common_string(cipher_copy, ret, &first_index,
@@ -920,12 +978,18 @@
 
 	MEM_FREE(cipher_copy);
 	if (!change_count)
-		return -1;
-	return 0;
+		*is_split_unifies_case = 2;
+	else
+		*is_split_unifies_case = 0;
+	return;
 
 change_case:
+	if (call_cnt == 0)
+		*is_split_unifies_case = -1;
+	else if (*is_split_unifies_case != -1)
+		*is_split_unifies_case = 3;
 	MEM_FREE(cipher_copy);
-	return 1;
+	return;
 }
 
 static char *fmt_self_test_full_body(struct fmt_main *format,
@@ -1129,19 +1193,9 @@
 			}
 		}
 #endif
-		++cnt_split_unifies_case;
-		// here we find cases where the unify_case flag is not set
-		// but should be, is set but should not be, and where the
-		// case unification code only 'sometimes' works.
-		if (is_split_unifies_case == 0 &&
-		    test_fmt_split_unifies_case(format, ciphertext) == 1) {
-			if (cnt_split_unifies_case > 1)
-				is_split_unifies_case = -1;
-			else
-				is_split_unifies_case = 1;
-		} else if (is_split_unifies_case == 1 &&
-		           !test_fmt_split_unifies_case(format, ciphertext))
-			is_split_unifies_case = -1;
+		test_fmt_split_unifies_case(format, ciphertext,
+		                            &is_split_unifies_case,
+		                            cnt_split_unifies_case);
 
 		ciphertext = format->methods.split(ciphertext, 0, format);
 
@@ -1418,16 +1472,23 @@
 		}
 	}
 
-	if (is_split_unifies_case == 1 && !(format->params.flags & FMT_SPLIT_UNIFIES_CASE)) {
-		snprintf(err_buf, sizeof(err_buf), "should set FMT_SPLIT_UNIFIES_CASE");
-		return err_buf;
-	} else if (!is_split_unifies_case && (format->params.flags & FMT_SPLIT_UNIFIES_CASE)) {
-		snprintf(err_buf, sizeof(err_buf), "should not set FMT_SPLIT_UNIFIES_CASE");
-		return err_buf;
-	} else if (is_split_unifies_case == -1) {
-		snprintf(err_buf, sizeof(err_buf),
-			"FMT_SPLIT_UNIFIES_CASE is only working for some cases");
-		return err_buf;
+	switch (is_split_unifies_case) {
+		case 1:
+			snprintf(err_buf, sizeof(err_buf), "should set FMT_SPLIT_UNIFIES_CASE");
+			return err_buf;
+		case 2:
+			snprintf(err_buf, sizeof(err_buf), "should not set FMT_SPLIT_UNIFIES_CASE");
+			return err_buf;
+		case 3:
+			snprintf(err_buf, sizeof(err_buf), "split() is only casing sometimes");
+			return err_buf;
+		case 4:
+			snprintf(err_buf, sizeof(err_buf), "split() case, or valid() should fail");
+			return err_buf;
+		case 0:
+		case -1:
+		default:
+			break;
 	}
 
 	format->methods.clear_keys();

```

jfoug:
I am pretty darn sure the above list IS a list of formats with problems.  here is an simple example. I pulled the data from nk (nuked klan ??) these are the test hashes out using -list=format-tests

```
$ cat in

openwall:$nk$*379637B4fcde21b2c5fbc9a00af505e997443267*#17737d3661312121d5ae7d5c6156c0298::
password:$nk$*379637B4fcde21b2c5fbc9a00af505e997443267*#5c20384512ee36590f5f0ab38a46c6ced::
test1:$nk$*503476424C5362476f36463630796a6e6c656165*#2f27c20e65b88b76c913115cdec3d9a18::
thatsworking:$nk$*7A317a71794339586c434d50506b6e4356626a67*#b62a615f605c2fd520edde76577d30f90::
test3:$nk$*796b7375666d7545695032413769443977644132*#4aec90bd9a930faaa42a0d7d40056132e::

openwall:$nk$*379637b4fcde21b2c5fbc9a00af505e997443267*#17737D3661312121d5ae7d5c6156c0298::
password:$nk$*379637b4fcde21b2c5fbc9a00af505e997443267*#5C20384512ee36590f5f0ab38a46c6ced::
test1:$nk$*503476424c5362476f36463630796a6e6c656165*#2f27C20e65b88b76c913115cdec3d9a18::
thatsworking:$nk$*7a317a71794339586c434d50506b6e4356626a67*#b62A615f605c2fd520edde76577d30f90::
test3:$nk$*796b7375666d7545695032413769443977644132*#4aec90Bd9a930faaa42a0d7d40056132e::

openwall:$nk$*379637b4fcde21b2c5fbc9a00af505e997443267*#17737d3661312121d5ae7d5c6156c0298::
password:$nk$*379637b4fcde21b2c5fbc9a00af505e997443267*#5c20384512ee36590f5f0ab38a46c6ced::
test1:$nk$*503476424c5362476f36463630796a6e6c656165*#2f27c20e65b88b76c913115cdec3d9a18::
thatsworking:$nk$*7a317a71794339586c434d50506b6e4356626a67*#b62a615f605c2fd520edde76577d30f90::
test3:$nk$*796b7375666d7545695032413769443977644132*#4aec90bd9a930faaa42a0d7d40056132e::

$ ../run/john in
Loaded 15 password hashes with 5 different salts (nk, Nuked-Klan CMS [SHA1 MD5 32/64])
Will run 8 OpenMP threads
test3            (test3)
test3            (test3)
test3            (test3)
thatsworking     (thatsworking)
thatsworking     (thatsworking)
thatsworking     (thatsworking)
test1            (test1)
test1            (test1)
test1            (test1)
password         (password)
password         (password)
password         (password)
openwall         (openwall)
openwall         (openwall)
openwall         (openwall)
15g 0:00:00:00 DONE 1/3 (2015-09-10 13:07) 1500g/s 1300p/s 2300c/s 6900C/s openwall..password
Session completed

$ cat ../run/john.pot
$nk$*796b7375666d7545695032413769443977644132*#4aec90bd9a930faaa42a0d7d40056132e:test3
$nk$*7A317a71794339586c434d50506b6e4356626a67*#b62a615f605c2fd520edde76577d30f90:thatsworking
$nk$*503476424C5362476f36463630796a6e6c656165*#2f27c20e65b88b76c913115cdec3d9a18:test1
$nk$*379637B4fcde21b2c5fbc9a00af505e997443267*#5c20384512ee36590f5f0ab38a46c6ced:password
$nk$*379637B4fcde21b2c5fbc9a00af505e997443267*#17737d3661312121d5ae7d5c6156c0298:openwall

$ ../run/john in
Loaded 15 password hashes with 5 different salts (nk, Nuked-Klan CMS [SHA1 MD5 32/64])
Remaining 9 password hashes with 5 different salts
Will run 8 OpenMP threads
thatsworking     (thatsworking)
thatsworking     (thatsworking)
test1            (test1)
test1            (test1)
password         (password)
password         (password)
openwall         (openwall)
openwall         (openwall)
test3            (test3)
9g 0:00:00:00 DONE 1/3 (2015-09-10 13:08) 9000g/s 12000p/s 22000c/s 37000C/s test3..openwall
Session completed

$ cat ../run/john.pot
$nk$*796b7375666d7545695032413769443977644132*#4aec90bd9a930faaa42a0d7d40056132e:test3
$nk$*7A317a71794339586c434d50506b6e4356626a67*#b62a615f605c2fd520edde76577d30f90:thatsworking
$nk$*503476424C5362476f36463630796a6e6c656165*#2f27c20e65b88b76c913115cdec3d9a18:test1
$nk$*379637B4fcde21b2c5fbc9a00af505e997443267*#5c20384512ee36590f5f0ab38a46c6ced:password
$nk$*379637B4fcde21b2c5fbc9a00af505e997443267*#17737d3661312121d5ae7d5c6156c0298:openwall
$nk$*7a317a71794339586c434d50506b6e4356626a67*#b62a615f605c2fd520edde76577d30f90:thatsworking
$nk$*503476424c5362476f36463630796a6e6c656165*#2f27c20e65b88b76c913115cdec3d9a18:test1
$nk$*379637b4fcde21b2c5fbc9a00af505e997443267*#5c20384512ee36590f5f0ab38a46c6ced:password
$nk$*379637b4fcde21b2c5fbc9a00af505e997443267*#17737d3661312121d5ae7d5c6156c0298:openwall
$nk$*796b7375666d7545695032413769443977644132*#4aec90Bd9a930faaa42a0d7d40056132e:test3

$ ../run/john in
Loaded 15 password hashes with 5 different salts (nk, Nuked-Klan CMS [SHA1 MD5 32/64])
Remaining 4 password hashes with 4 different salts
Will run 8 OpenMP threads
thatsworking     (thatsworking)
test1            (test1)
password         (password)
openwall         (openwall)
4g 0:00:00:00 DONE 1/3 (2015-09-10 13:08) 400.0g/s 900.0p/s 1500c/s 1500C/s openwall..password
Session completed

$ tail -15 ../run/john.pot
$2a$12$Qgj/IxYT7JZCbmeeAJvuceD6Vgxk0ENDm8nN7rwMmsizLekK8bbdy:12345
$nk$*796b7375666d7545695032413769443977644132*#4aec90bd9a930faaa42a0d7d40056132e:test3
$nk$*7A317a71794339586c434d50506b6e4356626a67*#b62a615f605c2fd520edde76577d30f90:thatsworking
$nk$*503476424C5362476f36463630796a6e6c656165*#2f27c20e65b88b76c913115cdec3d9a18:test1
$nk$*379637B4fcde21b2c5fbc9a00af505e997443267*#5c20384512ee36590f5f0ab38a46c6ced:password
$nk$*379637B4fcde21b2c5fbc9a00af505e997443267*#17737d3661312121d5ae7d5c6156c0298:openwall
$nk$*7a317a71794339586c434d50506b6e4356626a67*#b62a615f605c2fd520edde76577d30f90:thatsworking
$nk$*503476424c5362476f36463630796a6e6c656165*#2f27c20e65b88b76c913115cdec3d9a18:test1
$nk$*379637b4fcde21b2c5fbc9a00af505e997443267*#5c20384512ee36590f5f0ab38a46c6ced:password
$nk$*379637b4fcde21b2c5fbc9a00af505e997443267*#17737d3661312121d5ae7d5c6156c0298:openwall
$nk$*796b7375666d7545695032413769443977644132*#4aec90Bd9a930faaa42a0d7d40056132e:test3
$nk$*7a317a71794339586c434d50506b6e4356626a67*#b62A615f605c2fd520edde76577d30f90:thatsworking
$nk$*503476424c5362476f36463630796a6e6c656165*#2f27C20e65b88b76c913115cdec3d9a18:test1
$nk$*379637b4fcde21b2c5fbc9a00af505e997443267*#5C20384512ee36590f5f0ab38a46c6ced:password
$nk$*379637b4fcde21b2c5fbc9a00af505e997443267*#17737D3661312121d5ae7d5c6156c0298:openwall
```

The first block of 5 hashes, I simply changed 1 byte of each salt to make it upper cased. in the 2nd block, I upcase a letter of the hash which is the same as is being done in the test I made.  then in the 3rd test, it is the lines right from the -list=format-test -format=nk

Look at the hash count, 15, then 10 then 5 (actually 9 then 4 because I think I missed something somewhere).   This is a perfect demonstration of the problem.

jfoug:
NOTE, as this code shows, this format also has a the bug spill over into the hex encoded salt. HOWEVER the test code should not make assumptions here.  If we just upcase the whole thing past the format signature, we make totally invalid hashes.  That is why I 'picked' finding the hash as a hex string (or even just part of the hash).  If I find that, i KNOW it is a hex representation that must be properly cased or rejected as valid.  But  when we fix these, the binary hex blob MAY NOT be the only thing that salt must make sure is fixed.  If the salt had been actual text, then upcasing would be perfectly fine. The hash would not find the password, BUT it should be treated as 2 separate hashes (actually 2 separate salts).  But not for NK.  NK if the case of the salt is different, it needs to be fixed, or rejected.

frank-dittrich:
@jfoug how do you avoid false positives for hashes which use base64, but  for sume readon happen to just use 0-9, a-f, A-F?

jfoug:
```C
+       bin = format->methods.binary(ret_copy);
+       bin_hex = mem_alloc(format->params.binary_size*2 + 4);
+       base64_convert(bin, e_b64_raw, format->params.binary_size, bin_hex, e_b64_hex, format->params.binary_size*2+1, 0);
+       cp = strstr(ret_copy, bin_hex);
+       if (cp) {
+     
```

it will ONLY call this part of the function if somewhere I can find the hex representation of binary,  somewhere in the format string.  I check both lower and upper case.  if I find the string, I change the case of the string, then overwrite that part of the hash with the changed case.  It ONLY finds hex encoded strings.  It is not 100% bulletproof, BUT it should not false positive, ever (I hope)

jfoug:
Keep in mind, I may not even catch the entire binary hex encoded hash.  The example I showed in this thread was formspring. That is an sha256 thin dynamic format.  But the binary size within dynamic is 16 bytes (all dynamic ONLY use 16 byte binaries).  So the code only smashes half of the hash. But unless that half hash is 100% numeric and EVERY test hash had that first 16 byte hash being composed of only numeric data, then it would false pass this format.  It is not going to happen very often, even if there is only 1 test hash.  There may be some formats that are only 4 byte binaries.  Then we are only dealing with 8 bytes of hex. Then there is more chance of the first 4 bytes being all numeric, and we may want to check to make 'sure'.

But overall for a automatic testing code, I think it caught all of them, or at least almost all of them.

jfoug:
Like I also said, it will not catch the case where a format DOES have a split() case, but that only does the binary hash. Take the nk format I just did.  If we only cased the binary, this full ST code would say things are great, when they really are not.  If we have any formats which there are casing items not being caught, this is likely the reason why.  I guess I could do the same thing for salts.  Change the case only on the salt (after calling salt), and if I can find it IN hex in the original hash string, change the case and overwrite that section of hash buffer, and then make sure that either split fixes it, OR valid now fails.  Should actually be easy to do that.

jfoug:
@frank-dittrich thanks for joining the thread, since you are the absolute expert here, and have been tracking/pushing for this for a decade.  Lets get it cleaned up.  I know that no-one has complained (too loudly), but this bug is there 100% sure, even though there has been no complaints.

jfoug:
I am checking to see if case smashing of found hex encoded salts adds any 'new' broken hashes. I will update the 'full' list and #1742 with any new hashes. found.

NOTE, I have not performed ANY tests of the GPU formats.  I will leave that to others.  But I would be really surprised if the same opencl formats as the formats that show the bug on CPU side, are also not broken.

frank-dittrich:
@jfoug I am no more an expert here than anybody else. I just stumbled across the problem with dupe suppression for hashes with different hash representations, about 10 years ago. (I wasn't even interested in upper vs. lower case hex strings, I just used that as an example, and Solar decided to address this special case of hashes with different representations.)
And currently, I am even less of an expert. I caught a cold, and my head feels like it might explode any time. So, thinking is really hard for me right now.

PR #1755, opened 2015-09-13
Move Claudio's SHA2 file to be a '-ng' format.

claudioandre:
Step 1:
- To revert to Myrice version.


magnumripper:
Does your second commit screw up the filename changes, or am I just confused?

claudioandre:
The second commit is there to give you some time.
- raw-sha512-opencl now can handle a lot of hashes without speed regressions (so, people will stop complaining for a while).

claudioandre:
You can, when you have it done, replace these two files.
- src/opencl/sha512_kernel.cl
- src/opencl_rawsha512_fmt_plug.c

magnumripper:
I think I get it :)

PR #1805, opened 2015-10-08
Previous fix for sha256crypt is necessary for Catalyst 15.7 too.

claudioandre:
Sorry, I wasn't aware of it. 
*****

BTW: on super (not sure in which openwall machine you guys are using TS).
```
$ ./jtrts.pl raw-sha256-opencl 
Can't locate JSON/XS.pm in @INC (@INC contains: /usr/local/lib64/perl5 /usr/local/share/perl5 /usr/lib64/perl5/vendor_perl /usr/share/perl5/vendor_perl /usr/lib64/perl5 /usr/share/perl5 .) at ./jtrts.pl line 8.
BEGIN failed--compilation aborted at ./jtrts.pl line 8.
```

### DIFF
--- a/src/opencl/cryptsha256_kernel_GPU.cl
+++ b/src/opencl/cryptsha256_kernel_GPU.cl
@@ -411,7 +411,7 @@ inline void sha256_prepare(__constant sha256_salt	 * salt_data,
 
 	sha256_digest(ctx);
 
-#if __GPU__ && DEV_VER_MAJOR == 1729
+#if __GPU__ && (DEV_VER_MAJOR == 1729 || DEV_VER_MAJOR == 1800)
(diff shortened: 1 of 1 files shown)


magnumripper:
Yeah you need to install JSON::XS (locally) from CPAN. But I have yet to succeed in that (cpan self-tests fail) so I'd rather ditch that dependency.

jfoug:
I will work on ditching it and writing my own native JSON_PP functions. I think there is only 2 function calls I use it for.

jfoug:
Are you able to install JSON  (vs JSON:XS) ?

magnumripper:
I can't seem to install any damn thing on that host. Maybe it's me, I'm not used to do user-local installs.

PR #1825, opened 2015-10-13
Dynamic type for PostgreSQL stored MD5 hashes

reveliant:
Seems nobody published such a type already, whereas JtR cracks mysql(-sha1) and oracle hashes.
I've successfully used that new type today on real life databases hashes.
Might it be usefull for anybody else...


magnumripper:
Cool, thanks.

jfoug:
Glad you caught this, I was going to list that it needed done. (saltlen)

PR #1853, opened 2015-10-26
Use a "fake db" for self-test and benchmark.

magnumripper:
WIP, do not merge.

### DIFF
--- a/src/bench.c
+++ b/src/bench.c
@@ -678,7 +678,11 @@ AGAIN:
 		}
 
 		total++;
+
+		/* (Ab)used to mute some messages from source() */
+		bench_running = 1;
 		test_db = ldr_init_test_db(format, NULL);
+		bench_running = 0;
 
 		if ((result = benchmark_format(format,
 		    format->params.salt_size ? BENCHMARK_MANY : 1,
--- a/src/bench.c
+++ b/src/bench.c
@@ -175,7 +175,7 @@ static unsigned int get_cost(struct fmt_main *format, int index, int cost_idx)
 #endif
 
 char *benchmark_format(struct fmt_main *format, int salts,
-	struct bench_results *results)
+	struct bench_results *results, struct db_main *test_db)
 {
 	static void *binary = NULL;
 	static int binary_size = 0;
@@ -252,7 +252,7 @@ char *benchmark_format(struct fmt_main *format, int salts,
 #endif
 	if (!(current = format->params.tests) || !current->ciphertext)
 		return "FAILED (no data)";
-	if ((where = fmt_self_test(format, NULL))) {
+	if ((where = fmt_self_test(format, test_db))) {
 		sprintf(s_error, "FAILED (%s)\n", where);
 		return s_error;
 	}
@@ -413,7 +413,7 @@ char *benchmark_format(struct fmt_main *format, int salts,
 
 		if (salts > 1) format->methods.set_salt(two_salts[index & 1]);
 		format->methods.cmp_all(binary,
-		    format->methods.crypt_all(&count, NULL));
+		    format->methods.crypt_all(&count, test_db->salts));
 
 		add32to64(&crypts, count);
 #if !OS_TIMER
@@ -529,7 +529,7 @@ int benchmark_all(void)
 #endif
 	unsigned int total, failed;
 	MEMDBG_HANDLE memHand;
-
+	struct db_main *test_db;
 #ifdef _OPENMP
 	int ompt;
 	int ompt_start = omp_get_max_threads();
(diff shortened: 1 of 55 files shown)


magnumripper:
mschap-v2 doesn't seem to like this, very interesting since it uses neither of reset's db nor crypt_all's salt...

claudioandre:
FYI: I'm having problems with free() inside done(). Please try your patch using memory debug on.

This works: ```git checkout f6d0ba353c787d718975d0606748b5d3a634d66b```
This segfaults: ```git checkout bleeding-jumbo```

I'm offline now, can't do more than warn.

jfoug:
Do not EVER (there are caveates) use free().

Use MEM_FREE()

The caveats are if there are library calls that require the system free, then use free.  Actually, it should be libc_free() I think.  One lib call to NOT use free on is strdup().  the memdbg code handles these items:   alloc() calloc() and strdup()  (free of course).

I will look at the code from that commit and see if I can spot the problem.

jfoug:
Ok, here are all of the free() calls:

```
$ egrep -v '(BN)|(PKCS12)|(BIO)|(OPENSSL)|(rarvm)|(MEMDBG)|(libc)|(bt)|(RSA)|(DSA)|(db_main)|(EVP_PKEY)'_free *.[ch] | grep -v 2john | grep -v memdbg\. | grep -v memory\.c | grep 'free('

c3_fmt.c: * support and error checking.  We do not need to free() this memory anyway.
DES_bs.c: * for its alignment support and error checking.  We do not need to free() this
MD5_std.c: * for its alignment support and error checking.  We do not need to free() this
memory.h:               _aligned_free((ptr)); \
memory.h:               free((ptr)); \
mkvcalcproba.c:         free(ligne);
mkvcalcproba.c:         free(ligne);
SIPdump.c:              free(pcap_file);
SIPdump.c:              free(dev);
SIPdump.c:      free(dump_file);
SIPdump.c:      free(dev);
SIPdump.c:      free(pcap_file);
SIPdump.c:              free(filter);
SIPdump.c:                              free(payload_buffer);
SIPdump.c:              free(lines[i]);
SIPdump.c:              free(lines[num_lines - 1]);
SIPdump.c:              free(lines[num_lines - 1]);
SIPdump.c:              free(lines[num_lines - 1]);
SIPdump.c:      free(lines[num_lines - 1]);
SIPdump.h:              free(tempfile);
SIPdump.h:              free(tempfile);
SIPdump.h:              free(tempfile);
SIPdump.h:      free(tempfile);
SIPdump.h:      free(tempbuf);
unrarfilter.c:  free(filter);
```

The one I am concerned about is the one in unrarfilter.c

magnumripper:
IIRC the unrarfilter one is actually handled by a macro, or should be left as free for some other reason. I'm sure we looked at it before.

I suppose @claudioandre meant `MEM_FREE()` in some OpenCL format. I will look into it.

magnumripper:
unrarfilter was changed back to `free()` in af239d47. It's unclear why.

magnumripper:
Using a memdbg build, I ran a full -test=0 with all OpenCL formats, no problem seen. Will try ASan instead.

magnumripper:
Ah, no. What I just did was ASan. So I'll test memdbg now instead. :dizzy:

magnumripper:
The memdbg build is broken?
```
Development options (these may hurt performance when enabled):
Memdbg memory debugging settings ........... -DMEMDBG_ON
AddressSanitizer ("ASan") .................. disabled
UndefinedBehaviorSanitizer ("UbSan") ....... disabled

Configure finished.  Now 'make clean && make -s' to compile.
DES_std.c: In function 'DES_std_set_key':
DES_std.c:639:17: warning: array subscript is above array bounds [-Warray-bounds]
   while (DES_key[i++]) k += 2;
                 ^
gcc: error: memdbg.o: No such file or directory
make[1]: *** [../run/mkvcalcproba] Error 1
ar: creating aes.a
make: *** [default] Error 2
```

magnumripper:
Anyway I can't reproduce - tried real crack as well. Please open an issue with more details.

magnumripper:
Found it, opening an issue.
```
$ ../run/john -test -form:sha256crypt-opencl -dev=1
Device 1: HD Graphics 4000
Benchmarking: sha256crypt-opencl, crypt(3) $5$ (rounds=5000) [SHA256 OpenCL]... DONE
Raw:	1420 c/s real, 546133 c/s virtual

Mem leak: 65536 bytes, alloc_num 239, file opencl_cryptsha256_fmt_plug.c, line 561
```
Only seen with that very format as of yet

magnumripper:
Nah, that was #1945 and fixed now. You must mean something else (this wouldn't segfault). I guess you'll need to give more details.

claudioandre:
See below (it segfauts using HEAD). But I believe db is no longer NULL (or something like that). So, only 'my' format code has to change, nor core or shared code.

```
git checkout f6d0ba353c787d718975d0606748b5d3a634d66b
make -s distclean ; ./configure --enable-asan && make -s -j 4
$ ../run/john -form:raw-sha512-ng-opencl --mask --test
Device 0: Juniper [AMD Radeon HD 6700 Series]
Benchmarking: Raw-SHA512-ng-opencl [SHA512 OpenCL]... DONE, GPU util:84%
Raw:	49257K c/s real, 2285M c/s virtual
```

magnumripper:
db is never null now.

* If running -test we get a db built from test vectors.
* If running a real crack, before self-tests we first a db built from test vectors (also pointing to the real db in `db->real`) and after self-tests we get the read db (and `db->real` points back to `db`).

magnumripper:
OK, using `-test -mask` I do get a segfault. I should have found that during my regression testing, sorry.

magnumripper:
I think the reset "scheme" now used in nt-opencl and mscash-opencl is correct.

PR #1864, opened 2015-10-29
Rename vtesteq_epi32 to vanyeq_epi32

zzlei:
since the latter is more intuitive (borrowed from AltiVec's vec_any_eq);
and drop vtestz_epi32 as it's not used anywhere.


zzlei:
vtesteq_epi32 might be mis-interpreted as test_all_eq, while it actually means test_any_eq (I've made that mistake before). So it's better to clear that out.

PR #1918, opened 2015-11-26
Update driver detection to new Crimson.

claudioandre:


magnumripper:
Cool. Did they come up with any new ridiculous bugs?

claudioandre:
You know AMD! Of course there are ridiculous bugs, e.g., the compilation of a simple kernel segfaults. 
After some tests on GCN this is a serious candidate to be "not recommended". 
****

Poor magnum, a man without faith.

BTW: The 2D driver seems good enough, maybe really better.

magnumripper:
> compilation of a simple kernel segfaults. 

D'oh, sounds like Apple. Their drivers started crappy, then got better and better until a point where *all* our formats worked perfectly fine on all my devices. Then right after that, it went totally downhill. Now I can't even run the NT kernel on nvidia :cry:

PR #1922, opened 2015-11-29
Improve recommendations about OpenCL drivers.

claudioandre:
If changed this way, we can control better the recommendations. 
* E.g., legacy can have a set of recommended drivers. GCN a different one.

### DIFF
--- a/src/common-opencl.c
+++ b/src/common-opencl.c
@@ -244,7 +244,7 @@ static char *opencl_driver_ver(int sequential_id)
 static char *opencl_driver_info(int sequential_id)
 {
 	static char ret[64];
-	char dname[MAX_OCLINFO_STRING_LEN];
+	char dname[MAX_OCLINFO_STRING_LEN], tmp[64];
 	int major = 0, minor = 0, i = 0;
 
 	int known_drivers[][2] = {
@@ -276,13 +276,13 @@ static char *opencl_driver_info(int sequential_id)
 		"13.12",
 		"14.4 (Mantle)",
 		"14.6 beta (Mantle)",
-		"14.9 (Mantle) [recommended]",
-		"14.12 (Omega) [recommended]",
+		"14.9 (Mantle)",
+		"14.12 (Omega)",
 		"15.5 beta [not recommended]",
 		"15.5",
-		"15.7 [recommended]",
+		"15.7",
 		"15.7.1",
-		"15.9 [recommended]",
+		"15.9",
 		"15.11",
 		""
 	};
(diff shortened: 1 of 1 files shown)


magnumripper:
I've had some vague idea of moving the definition of "recommended" to john.conf. Perhaps also the translation from reported version number to eg. a Catalyst version. Then we wouldn't need to change our executable just for adding a new version.

claudioandre:
Seems good. Not sure what do you want listed, I would like something like this:

Driver Version| Description | Recomendation
------------ | ------------- | -------------
938 , 2  | "12.8"  |
1084, 4  | "13.1"  |
1124, 2  | "13.4"  |
1214, 3  | "13.6 beta"  |
1311, 2  | "13.11 beta-1"  |
1348, 5  | "13.12"  |
1445, 5  | "14.4 (Mantle)"  |
1526, 3  | "14.6 beta (Mantle)"  |
1573, 4  | "14.9 (Mantle)"  | VLS
1642, 5  | "14.12 (Omega)"  | VLS
1702, 3  | "15.5 beta"  | **T
1729, 3  | "15.5"  |
1800, 5  | "15.7"  | **R
1800, 8  | "15.7.1"  | *WR
1800, 11 | "15.9"  |
1912, 5  | "15.11"  |
0, 0     | ""  |

---- Label ----
\* -> all hardware/software
G -> GCN
V -> VLIW4 and VLIW5
W -> Windows
L -> Linux
R -> recommended
S -> supported
T -> not recommended: really bad software. I mean "trash".

claudioandre:
Is it possible to deal with an array like this using john.conf and JP helper functions?

--

Notice (opinion):
1. to me GCN requires a new driver. Today it is 15.7.
2. legacy (but very good) hardware has more options and 15.7 is good enough. 13.9 can be recommended too.
3. GCN recommendations will move ahead (in the future). Legacy won't.
4. I would say [update needed] for a bunch of listed drivers

claudioandre:
> I would say [update needed] for a bunch of the listed drivers.

Ok, not exactly "[update needed]", but seriosly, if JtR is running on the new and unusable Apple driver let the user knows about it.

magnumripper:
> if JtR is running on the new and unusable Apple driver let the user knows about it.

Yeah it could say "known bad" or something like that.

> Is it possible to deal with an array like this using john.conf and JP helper functions?

I'm not quite sure. Maybe we'd use a list like `[List.Drivers:OpenCL]` or `[List.OpenCL:Drivers]` but each entry in that list will be a string so we might need to handle a field delimiter within that string on our own. Or maybe there are some better thing to use, I can't recall the alternatives right now.

claudioandre:
Something like this is going to be neeeded (I guess it is not complete):

```
# This patch file was generated by NetBeans IDE
# It uses platform neutral UTF-8 encoding and \n newlines.
--- HEAD
+++ Modificado na Árvore de Trabalho
@@ -337,6 +337,11 @@
 #if HAVE_OPENCL
 	if (!strcasecmp(options.listconf, "opencl-devices"))
 	{
+	    	//path_init_ex(options.config);
+		//cfg_init(options.config, 0);
+		cfg_init(CFG_FULL_NAME, 1);
+		cfg_init(CFG_ALT_NAME, 0);
+			
 		opencl_preinit();
 		opencl_list_devices();
 		exit(EXIT_SUCCESS);
```

magnumripper:
If that's needed I presume (without looking) that the proper fix would be to move --list-devices to `listconf_parse_late()`.

PR #1925, opened 2015-11-29
Use a proper max_run_time while auto tuning.

claudioandre:
Please review and check if it could be an improvement.

### DIFF
--- a/src/opencl_cryptsha256_fmt_plug.c
+++ b/src/opencl_cryptsha256_fmt_plug.c
@@ -348,6 +348,7 @@ static void reset(struct db_main *db)
 		char *tmp_value;
 		char *task = "$JOHN/kernels/cryptsha256_kernel_DEFAULT.cl";
 		int major, minor;
+		unsigned long long int max_run_time;
 
 		opencl_prepare_dev(gpu_id);
 		source_in_use = device_info[gpu_id];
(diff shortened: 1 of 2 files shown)


magnumripper:
Does it change the outcome of the auto tune? Ideally none of this would be needed for a split kernel format - it should find optimal speed in a split second and if that happens to be 99 or 301 ms, so be it.

One could opt to change number of iterations per kernel call instead, that's on my to-do list to try out. Eg. a very fast GPU may do the full WPAPSK in a single call to the loop kernel while a weak device may do it in, say, 10 calls.

claudioandre:
> Does it change the outcome of the auto tune?

* A weak GPU needs 300ms to get the optimal GWS.
* A powerful GPU waste time probing up to 300ms since optimal is found at 90ms.
* Sometimes a powerful GPU selects a huge GWS and this is undesiderable.

So, this is not really important but can offer a better user experience.

Probably tune iterations to 200ms - 300ms is a better choice. In this case, I would use something like this.

```
if (speed() < 1M)  
    300ms
else
    200ms //a magic number we saw in the past.
```

I mean, the spirit of this patch will remain.

magnumripper:
Still, if a fast device completes a split kernel in 90 ms, it's likely more effective to double the number of iterations per loop kernel call (so half the number of calls, for less overhead) for it. And a slow device that currently needs 300ms would get *fewer* number of iterations per call. Given that, all devices could aim for 200ms. I think that's the ultimate way to go but I'm not sure our current "speed index" is good enough. It's a very very rough figure.

PR #1927, opened 2015-12-01
Use john.conf to keep calibration values to specific GPUs.

claudioandre:
* This allows us to be able to tune to a broader range of hardware, for each board (board name).


src/opencl_sha512.h
```diff
@@ -25,6 +25,8 @@
         ((ulong)amd_bitalign((uint)(x), (uint)((x) >> 32), (uint)(n)) << 32)) : \
         (amd_bitalign((uint)(x), (uint)((x) >> 32), (uint)(n) - 32) |           \
         ((ulong)amd_bitalign((uint)((x) >> 32), (uint)(x), (uint)(n) - 32) << 32)))
+#elif cpu(DEVICE_INFO)
+#define ror64(x, n)             ((x >> n) | (x << (64UL-n)))
```
magnumripper: This is faster on CPU? It's truly amazing how all vendors damn near always fail at implementing the high level functions. I mean, WTF, how hard can it be?!

claudioandre: > This is faster on CPU?

Yes. Impressive, I agree!!

PR #1933, opened 2015-12-02
Select the most powerful GPU as the default one.

claudioandre:


claudioandre:
On super, it is a bad idea to select the accelerator Phi. It is useless in OpenCL. So:

It selects the most powerful GPU. 
1. If none, select the most powerful accelerator.
2. If note, select -dev=0.

magnumripper:
Works great, thanks!

PR #1975, opened 2015-12-31
compiles cleanly with rexgen 1.3.1, contains version check for rexgen

teeshop:
I added a version check, which should prevent linking john against an incompatible version of rexgen, In addition, the same changes as for rexgen 1.2.8 are included.

Regards, Jan

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -10136,20 +10136,14 @@ int main() { return ! (rexgen_version_int() >= 0x010300); }
 _ACEOF
 if ac_fn_c_try_run "$LINENO"; then :
   using_rexgen=yes
-else
-  as_fn_error $? "incompatible version of librexgen found :-( you need at least version 1.3.0" "$LINENO" 5
 fi
 rm -f core *.core core.conftest.* gmon.out bb.out conftest$ac_exeext \
   conftest.$ac_objext conftest.beam conftest.$ac_ext
 fi
 
-else
-  as_fn_error $? "unable to find librexgen" "$LINENO" 5
 
 fi
 
-else
-  as_fn_error $? "unable to find librexgen/version.h" "$LINENO" 5
 
 fi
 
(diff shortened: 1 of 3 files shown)


frank-dittrich:
I tested several --regex options (standalone and combined with --wordlist) using --stdout.
Looks good. Thanks!
(Didin't yet check why circleci isn't happy.)

frank-dittrich:
circleci tried to build with rexgen support, even though there is no librexgen.

I guess it is the same for travis-ci.

frank-dittrich:
If librexgen isn't found (or the version is wrong), the build shouldn't fail.
Instead, john should be built without the --regex option.

teeshop:
Looks better now. Happy new year ;-)

magnumripper:
Works like a champ, thank you!

frank-dittrich:
> Works like a champ, thank you!

Yes, except that librexgen writes tons of debug output to stderr:

```
(bleeding-jumbo)run $ echo test > test.lst
(bleeding-jumbo)run $ ./john --stdout --wordlist=test.lst --regex='1\02'
APPEND
0074
0065
0073
0074
1test2
1p 0:00:00:00 100.00% (2016-01-03 11:00) 25.00p/s 1test2
```

I reported that as rexgen issue 25:
https://github.com/teeshop/rexgen/issues/25

magnumripper:
There were also thrashed whitespace, I missed that. Fixed in bfbfd49.

frank-dittrich:
I had a similar commit (whitespace fix) in my local repository.
I noticed these when I started to look into a few remaining memleak issues.

I am still fighting with a `‘retval’ may be used uninitialized` warning which IMHO is invalid.
(Even including misc.h to let the compiler know that error() doesn't return didn't help.)
But I don't want to initialize that variable when there is no good reason.
May be I need more coffee before I am able to figure this out.

PR #1982, opened 2016-01-07
AxCrypt file format + parser

Fist0urs:
Two files added:
   - axcrypt2john.py
   - axcrypt_fmt_plug.c

The first one's outputs needed data for the cracker. It actually supports two formats : file.axx and file.axx + key-file

There is no way in the file.axx to be sure that user's provided key-file is the right one as any file can be used to proceed encryption. Nevertheless, AxCrypt provides a key-file generator that outputs a .txt file containing 53 bytes in base64 format, so if you manage to find that kind of file, it is likely to be the good one.

I tried to follow currently used coding-style (except for tabs)

TODO:
  *add SIMD support for SHA1 part
  *add doc/AxCrypt-Auditing-HOWTO file, as there are many things to say :)

Happy new year john community!

### DIFF
--- a/src/axcrypt_fmt_plug.c
+++ b/src/axcrypt_fmt_plug.c
@@ -175,7 +175,7 @@ static void *get_salt(char *ciphertext)
 
     /* if key-file present */
     if ((p = strtokm(NULL, "*")) != NULL){
-        cs.keyfile = (char*) mem_calloc(strlen(p)/2, sizeof(char));
+        cs.keyfile = (char*) mem_calloc_tiny(strlen(p)/2 + 1, sizeof(char));
         for (i = 0; i < strlen(p)/2; i++)
             cs.keyfile[i] = atoi16[ARCH_INDEX(p[i * 2])] * 16
                 + atoi16[ARCH_INDEX(p[i * 2 + 1])];
@@ -263,8 +263,6 @@ static int crypt_all(int *pcount, struct db_salt *salt)
             memcpy(lsb, cipher + 8, 8);
         }
         if (!memcmp(cipher, AES_WRAPPING_IV, 8)) {
-            if (cur_salt->keyfile != NULL)
-                MEM_FREE(cur_salt->keyfile);
             cracked[index] = 1;
 #ifdef _OPENMP
 #pragma omp atomic
--- a/src/axcrypt_fmt_plug.c
+++ b/src/axcrypt_fmt_plug.c
@@ -52,7 +52,7 @@ john_register_one(&fmt_axcrypt);
 		(cp)[0] ^= (unsigned char)((value)), \
 		(cp)[1] ^= (unsigned char)((value) >> 8), \
 		(cp)[2] ^= (unsigned char)((value) >> 16), \
-		(cp)[3] ^= (unsigned char)(value) >> 24 )
+		(cp)[3] ^= (unsigned char)((value) >> 24 ) )
 
 static struct fmt_tests axcrypt_tests[] = {
 	/*
@@ -131,11 +131,11 @@ static int valid(char *ciphertext, struct fmt_main *self)
 		goto err;
 	if ((p = strtokm(NULL, "*")) == NULL)	/* salt */
 		goto err;
-	if (strlen(p) != 32)
+	if (strlen(p) != 32 || !ishexlc(p))
 		goto err;
 	if ((p = strtokm(NULL, "*")) == NULL)	/* wrappedkey */
 		goto err;
-	if (strlen(p) != 48 )
+	if (strlen(p) != 48 || !ishexlc(p))
 		goto err;
 	/* optional key-file following */
 
--- a/src/axcrypt_fmt_plug.c
+++ b/src/axcrypt_fmt_plug.c
@@ -28,46 +28,46 @@ john_register_one(&fmt_axcrypt);
 #ifdef _OPENMP
 #include <omp.h>
 #ifndef OMP_SCALE
-#define OMP_SCALE             1
+#define OMP_SCALE		1
 #endif
 #endif
 #include "memdbg.h"
 
-#define FORMAT_LABEL          "axcrypt"
-#define FORMAT_NAME           "AxCrypt"
-#define ALGORITHM_NAME        "SHA1 AES 32/" ARCH_BITS_STR
-#define BENCHMARK_COMMENT     ""
-#define BENCHMARK_LENGTH      -1
-#define PLAINTEXT_LENGTH      125 /* actual max is 250 */
-#define BINARY_SIZE           0
-#define SALT_SIZE             sizeof(struct custom_salt)
-#define BINARY_ALIGN          MEM_ALIGN_NONE
-#define SALT_ALIGN            sizeof(int)
+#define FORMAT_LABEL		"axcrypt"
+#define FORMAT_NAME		"AxCrypt"
+#define ALGORITHM_NAME		"SHA1 AES 32/" ARCH_BITS_STR
+#define BENCHMARK_COMMENT	""
+#define BENCHMARK_LENGTH	-1
+#define PLAINTEXT_LENGTH	125 /* actual max is 250 */
+#define BINARY_SIZE		0
+#define SALT_SIZE		sizeof(struct custom_salt)
+#define BINARY_ALIGN		MEM_ALIGN_NONE
+#define SALT_ALIGN		sizeof(int)
 /* constant value recommended by FIPS */
-#define AES_WRAPPING_IV       "\xA6\xA6\xA6\xA6\xA6\xA6\xA6\xA6"
-#define MIN_KEYS_PER_CRYPT    1
-#define MAX_KEYS_PER_CRYPT    1
+#define AES_WRAPPING_IV		"\xA6\xA6\xA6\xA6\xA6\xA6\xA6\xA6"
+#define MIN_KEYS_PER_CRYPT	1
+#define MAX_KEYS_PER_CRYPT	1
 
 #define PUT_64BITS_XOR_MSB(cp, value) ( \
-        (cp)[0] ^= (unsigned char)((value)), \
-        (cp)[1] ^= (unsigned char)((value) >> 8), \
-        (cp)[2] ^= (unsigned char)((value) >> 16), \
-        (cp)[3] ^= (unsigned char)(value) >> 24 )
+		(cp)[0] ^= (unsigned char)((value)), \
+		(cp)[1] ^= (unsigned char)((value) >> 8), \
+		(cp)[2] ^= (unsigned char)((value) >> 16), \
+		(cp)[3] ^= (unsigned char)(value) >> 24 )
 
 static struct fmt_tests axcrypt_tests[] = {
-    /*
-        formats can be:
-        $axcrypt$*version*iterations*salt*wrappedkey
-        $axcrypt$*version*iterations*salt*wrappedkey*key-file
-    */
-    {"$axcrypt$*1*1337*0fd9e7e2f907f480f8af162564f8f94b*af10c88878ba4e2c89b12586f93b7802453121ee702bc362", "Bab00nmoNCo|\\|2$inge"},
-    {"$axcrypt$*1*60000*7522aa07694d441e47f8faad8a8cb984*95e02b7ccbdc27c227a80d1307505d8b769e87b32f312aa1", "nuNuche<3rewshauv"},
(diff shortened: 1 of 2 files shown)


frank-dittrich:
TODO:
make sure the self test passes ;)

CircleCI
Testing: axcrypt, AxCrypt [SHA1 AES 32/64]... (2xOMP) FAILED (cmp_all(2))

Travis gcc:
Testing: axcrypt, AxCrypt [SHA1 AES 32/64]... (4xOMP) FAILED (cmp_all(4))

Travis clang:
Testing: axcrypt, AxCrypt [SHA1 AES 32/64]... FAILED (cmp_all(1))

Fist0urs:
oh damn, you're right! I knew I was forgetting some tests... I'll fix that, thanks :)

magnumripper:
What exactly does "except for tabs" mean? I'd really like at least the .c file to be indented with tabs.

jfoug:
This -test failure may be an OMP issue.

$ ../run/john -test=0 -form=axcrypt
Will run 2 OpenMP threads
Testing: axcrypt, AxCrypt [SHA1 AES 32/64]... (2xOMP) FAILED (cmp_all(2))

$ OMP_NUM_THREADS=1 ../run/john -test=0 -form=axcrypt
Testing: axcrypt, AxCrypt [SHA1 AES 32/64]... PASS

jfoug:
Could this OMP issue be due to using OSSL for the AES ?  Also, wouldn't it be better to use JtR AES code?

magnumripper:
It does use our local aes.h already

magnumripper:
Code looks thread-safe to me. Perhaps it's an issue with keys_per_crypt > 1 even without OMP? Try bumping OMP_SCALE to 8 and run with OMP_NUM_THREADS=1.

magnumripper:
Also, Travis clang fails - it's non-OMP.

src/axcrypt_fmt_plug.c
```diff
@@ -0,0 +1,354 @@
...
+
+            /* 2nd block treatment */
+            PUT_64BITS_XOR_MSB(cipher, 2 * j + 1);
+            memcpy(cipher + 8, lsb, 8);
+            AES_decrypt(cipher, cipher, &akey);
+            memcpy(lsb, cipher + 8, 8);
+        }
+        if (!memcmp(cipher, AES_WRAPPING_IV, 8)) {
+            if (cur_salt->keyfile != NULL)
+                MEM_FREE(cur_salt->keyfile);
```
magnumripper: This use of free() on a global pointer within a parallel block is probably dangerous, but isn't the cause of the failures we're seeing.

Fist0urs:
> What exactly does "except for tabs" mean? I'd really like at least the .c file to be indented with tabs.

Well, depending on text editor, I had some alignment problems using tabs (had some display changes using gitlab for example). But I'll fix that :).



> This -test failure may be an OMP issue.

I also thought OMP was the issue (see below)

> This use of free() on a global pointer within a parallel block is dangerous, but isn't the cause of the failures we're seeing (or is it? Maybe it actually is).

You're right, I'll have to fix it, but I think the issue does not come from this

In fact I noticed quite odd behaviors. I didn't look in details at self-tests code (I guess I'll have to) as I thought it was handled the same way that cracking mode in testing provided cipher/plain. Here is what I observed:

At the begining I did not include self-tests and as I provided the self-tests ciphers, actually used here, as to be cracking, everything worked fine (with the 2 different mode: cipher and cipher + key-file). At that moment, no dyna_salt was used.

Then I added self-tests without providing key-file tests, everything worked.
Then I added key-file tests and I encountered strange behaviors:
   - I got alternatively "Warning: No dupe-salt detection" and "Self test failed (Could not find salt in db - salt() return inconsistent?)"
   - if I already had a john.pot containing results, I did not have these problems... (once again, I did not look at selt-test code)
   - if self-tests contained "cipher-only" tests, it worked (even if in "tocrack" ciphers there were key-file cases) but if it contained key-file tests (with or without "cipher-only") I got failures indicated previously

I thought that using dyna_salt would fix the problem, it did on my local machine and some others I tested on (that's why I pull-requested), but I forgot to try tests procedures.

As jfoug, I noticed that disabling OMP resolved the problems (before using dyna_salt) and with OMP there seemed to be some concurencial accesses (only on self-test cases containing key-file, whereas it didn't seem to be some on provided "tocrack" ciphers (cipher and cipher + key-file combined), but couldn't find where this was...
I also noticed that using "--fork" while cracking (beside self-tests), even if everything was cracked, it took quite some time to finish (I mean, there should be a delta on forked jobs, but not that long: ~10sec). That's why I think that the problem comes from how my implementation handles parrallel computation.

I guess I have to debug a little more when I have time. Anyway, thanks for reviews guys

magnumripper:
I can't reproduce any problems myself though:
```
$ ../run/john -test -form:axcrypt 
Will run 8 OpenMP threads
Benchmarking: axcrypt, AxCrypt [SHA1 AES 32/64]... (8xOMP) DONE
Raw:	368 c/s real, 53.8 c/s virtual

$ cd ../test
$ ./jtrts.pl -internal axcrypt
-------------------------------------------------------------------------------
- JtR-TestSuite (jtrts). Version 1.13, Dec 21, 2014.  By, Jim Fougeron & others
- Testing:  John the Ripper v3-1884-g6c6e1ac OMP [darwin15.2.0 64-bit AVX-ac]

John Jumbo build detected.
Running JTRTS in -internal mode

form=axcrypt                      guesses:    5 0:00:00:00 DONE  [PASSED]
.pot CHK:axcrypt                  guesses:    5 0:00:00:00 DONE  [PASSED] (5 val-pwd)
All tests passed without error.  Performed 1 tests.  Time used was 5 seconds
```
That's on OSX.

src/axcrypt_fmt_plug.c
```diff
@@ -0,0 +1,354 @@
...
+    if ((p = strtokm(NULL, "*")) != NULL){
+        cs.keyfile = (char*) mem_calloc(strlen(p)/2, sizeof(char));
+        for (i = 0; i < strlen(p)/2; i++)
+            cs.keyfile[i] = atoi16[ARCH_INDEX(p[i * 2])] * 16
+                + atoi16[ARCH_INDEX(p[i * 2 + 1])];
+    }
+    MEM_FREE(keeptr);
+
+    cs.dsalt.salt_cmp_offset = SALT_CMP_OFF(struct custom_salt, salt);
+    cs.dsalt.salt_cmp_size = SALT_CMP_SIZE(struct custom_salt, salt, wrappedkey, 0);
```
magnumripper: I think this should be `SALT_CMP_SIZE(struct custom_salt, salt, keyfile, size_of_keyfile);` for really proper salt compares... @jfoug?

However, that would not be the cause of problem seen.

magnumripper: Ah, but then you need to allocate all of the dyna salt in one go, not just keyfile. That's also needed for being able to free the salt with `alloc_needs_free`.

jfoug:
The reason dyna-salt was needed was due to the pointer.  If you add a hash with the same salt twice, that pointer would be different for both of them.  Thus, you have to use the dyna code to skip the pointer in the salt, but compare the actual 'key file' data.

But that has nothing to do with the ST fails.  I will look a bit and see if I can spot the problem.  But yes, the code 'appears' thread safe to me also .......

jfoug:
The bug only hits on the keyed file hashes.  I have put break points and seen this, and then if you comment them out, things run and pass just fine.

Fist0urs:
I am doing some tests too, changed code to be able to use salt_alloc_needs_free set to 1 and got a strange behaviour: 
  - if there's just one test provided, it crashes invoking double free or corrupted memory (key-file case or not);
  - if there is more than one, it goes well (and valgrind says no memory leak).

jfoug:
In get_salt.  The key file needs null:

  if ((p = strtokm(NULL, "*")) != NULL){
-         cs.keyfile = (char*) mem_calloc(strlen(p)/2, sizeof(char));
+        cs.keyfile = (char*) mem_calloc(strlen(p)/2+1, sizeof(char));

jfoug:
Here are the 'bugs' I found in the original source.  It now runs fine.  NOTE, to gets ST to work, you can NOT free the key_file pointer.  That really may have to simply wait until program exit (or use the mem_calloc_tiny function.

```diff
--- axcrypt_fmt_plug.c~	2016-01-08 16:25:02.255719400 -0600
+++ axcrypt_fmt_plug.c	2016-01-08 16:22:19.240701100 -0600
@@ -52,7 +52,7 @@ john_register_one(&fmt_axcrypt);
         (cp)[0] ^= (unsigned char)((value)), \
         (cp)[1] ^= (unsigned char)((value) >> 8), \
         (cp)[2] ^= (unsigned char)((value) >> 16), \
-        (cp)[3] ^= (unsigned char)(value) >> 24 )
+        (cp)[3] ^= (unsigned char)((value) >> 24) )
 
 static struct fmt_tests axcrypt_tests[] = {
     /*
@@ -131,11 +131,11 @@ static int valid(char *ciphertext, struc
         goto err;
     if ((p = strtokm(NULL, "*")) == NULL)    /* salt */
         goto err;
-    if (strlen(p) != 32)
+    if (strlen(p) != 32 || !ishexlc(p))
         goto err;
     if ((p = strtokm(NULL, "*")) == NULL)    /* wrappedkey */
         goto err;
-    if (strlen(p) != 48 )
+    if (strlen(p) != 48 || !ishexlc(p))
         goto err;
     /* optional key-file following */
 
@@ -175,7 +175,7 @@ static void *get_salt(char *ciphertext)
 
     /* if key-file present */
     if ((p = strtokm(NULL, "*")) != NULL){
-        cs.keyfile = (char*) mem_calloc(strlen(p)/2, sizeof(char));
+        cs.keyfile = (char*) mem_calloc(strlen(p)/2+1, sizeof(char));
         for (i = 0; i < strlen(p)/2; i++)
             cs.keyfile[i] = atoi16[ARCH_INDEX(p[i * 2])] * 16
                 + atoi16[ARCH_INDEX(p[i * 2 + 1])];
@@ -263,14 +263,14 @@ static int crypt_all(int *pcount, struct
             memcpy(lsb, cipher + 8, 8);
         }
         if (!memcmp(cipher, AES_WRAPPING_IV, 8)) {
-            if (cur_salt->keyfile != NULL)
-                MEM_FREE(cur_salt->keyfile);
             cracked[index] = 1;
 #ifdef _OPENMP
 #pragma omp atomic
 #endif
             any_cracked |= 1;
         }
+//	if (any_cracked && cur_salt->keyfile)
+//	   MEM_FREE(cur_salt->keyfile);
     }
     return count;
 }
```

jfoug:
Do the doc, and I think this one is ready to pull.  Going to SIMD is probably not going to speed things up any noticeable amount.  I guess that would depend upon what that 'normal' iteration counts are.

However, if the keyfile is 'normal', and larger, then moving to SIMD SHA code, 'could' provide some benefit.

magnumripper:
I'm merging this, but yes please add AxCrypt-Auditing-HOWTO as well.

Fist0urs:
> Do the doc, and I think this one is ready to pull. Going to SIMD is probably not going to speed things up any noticeable amount. I guess that would depend upon what that 'normal' iteration counts are.
However, if the keyfile is 'normal', and larger, then moving to SIMD SHA code, 'could' provide some benefit.

I was also wondering if it worths implementing SIMD SHA1 as it would need to retransform it afterward for AES. Also, default iteration number is 10000 (the number of iterations depends on the computer using AxCrypt: the more powerfull the highest number... Which is quite odd for a tool mainly used to share ciphered files...), which can be modified in the registry.

> I'm merging this, but yes please add AxCrypt-Auditing-HOWTO as well.

Yes I will add it as soon as I have time :)

I'll have to modify axcrypt2john.py too, to handle a case I didn't consider: AxCrypt gives the ability to create a self-decrypting archive, embedded in a PE. I guess I'll have to parse the PE to exactract needed data (cracker format won't change).

Fist0urs:
@jfoug @magnumripper 

AxCrypt-Auditing-HOWTO is ready to PR, but there is a part concerning auditing the SHA1 truncated to 16bytes. 
So I made a new "rawSHA1_axcrypt_fmt_plug.c" to be able to crack these hashes. Does this worth to PR too? (otherwise I'll remove the mention to "use --format=raw-sha1-axcrypt" in the AxCrypt-Auditing-HOWTO :) )

magnumripper:
Sure, I think it's worth a PR. Please go ahead.

PR #1997, opened 2016-01-18
AxCrypt auditing documentation + new format to crack in-memory secrets

Fist0urs:


Fist0urs:
Ok, fixing the no-sse case :)

magnumripper:
Thanks

Issue #1964, opened 2015-12-21
Keystore-opencl Format


sambwest:
It'd be great if there was a keystore-opencl format.

magnumripper:
Looks trivial enough and would not need GPU-side mask.

magnumripper:
Fixed in 4aca35ce

PR #2001, opened 2016-01-19
keystore-opencl format

terrybwest:
Further to Issue #1964 ... an OpenCL implementation for finding Java Keystore 'outer' password(s) (makes no attempt to implement finding password(s) for 'inner' private-key-based alias password(s)).
Appears to work well on AMD GPUs, but not tested on nVidia GPUs apart from those in 2 relatively ancient MacBook Pro's in which self-testing fails (as it does for the majority of SHA-based OpenCL formats).
In essence, the kernel implements generic OpenCL SHA-1 hashing for <password + very long salt> where the salt is almost as long as the keystore file itself. Current maximum salt length is hard-coded as 4096 bytes but this may turn out to be insufficient for keystores with many contained aliases. In theory, the kernel may be useful for hashing of other formats with long salts.
OpenMP is supported.
A disclaimer: I am new to this game and to GitHub ... please forgive any errors in the process. I've also not been able to find your 'coding standards' advice so may not have conformed to them!
I hope you'll find the format useful ...
Terry West

### DIFF
--- a/src/opencl_keystore_fmt_plug.c
+++ b/src/opencl_keystore_fmt_plug.c
@@ -1,0 +1,587 @@
+/* Java KeyStore password cracker for JtR.
+ * (will NOT address password(s) for alias(es) within keystore).
+ *
+ * OpenCL plugin by Terry West.
+ * Derived from keystore_fmt_plug.c,
+ * written by Dhiru Kholia <dhiru at openwall.com> and
+ * Narendra Kangralkar <narendrakangralkar at gmail.com>.
+ *
+ * Input Format: $keystore$target$salt_length$salt$hash$nkeys$keylength$keydata$keylength$keydata...
+ *
+ * This software is Copyright (c) 2015, Terry West <terrybwest at gmail dot com>
+ * and it is hereby released to the general public under the following terms:
+ * Redistribution and use in source and binary forms, with or without modification,
+ * are permitted.
+ */
+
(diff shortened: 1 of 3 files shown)


src/opencl_keystore_fmt_plug.c
```diff
@@ -0,0 +1,587 @@
...
+/*
+ Has this now become redundant? It looks like
+ autotuning does the same or similar?
+#ifdef _OPENMP
+	omp_t = omp_get_max_threads();
+	self->params.min_keys_per_crypt *= omp_t;
+	omp_t *= OMP_SCALE;
+	self->params.max_keys_per_crypt *= omp_t;
+#endif
+*/
```
magnumripper: For OpenCL formats, OpenMP is only needed if we do some post-processing on host side. For this format, all `#if _OPENMP` stuff, as well as the FMT_OMP format flag, can be dropped.

magnumripper:
Thanks! Since this seems to pass our build bots, I'll just merge it and test it more thoroughly later.

terrybwest:
That seems like good news - thank you. As I said above, I'm new to his game ... so does that mean that I can now abandon the code and leave it entirely in your capable hands?

magnumripper:
If you abandon it we will take care of things as long as we can. Failing that, we will drop it. Doesn't happen very often. All formats we've dropped that I can recall has been ones that are somehow superseded by other formats.

PR #2015, opened 2016-01-29
Improve find_valid_opencl_device() detection.

claudioandre:
This seems to be better.


magnumripper:
Just looking at the code I think it was a good catch.

PR #2026, opened 2016-02-07
Improvments of krb5tgs

Fist0urs:
Add support to different ciphertexts input format
Get rid of max data len concerning edata
Get rid of RC4 rounds
Get rid of last hmac-md5
Performances increase

I normalized as suggested by @magnumripper, alos changed displayed name to fit "etype 23" corresponding to arc4 algorithm.

Concerning the performances increase, this is what I get:

before:

    ./john --test --format=krb5tgs
    Will run 2 OpenMP threads
    Benchmarking: krb5tgs, Kerberos 5 TGS [MD4 HMAC-MD5 RC4]... (2xOMP) DONE
    Many salts:	117607 c/s real, 59991 c/s virtual
    Only one salt:	107520 c/s real, 54303 c/s virtual

after:

    ./john --test --format=krb5tgs
    Will run 2 OpenMP threads
    Benchmarking: krb5tgs, Kerberos 5 TGS [MD4 HMAC-MD5 RC4]... (2xOMP) DONE
    Many salts:	863232 c/s real, 542913 c/s virtual
    Only one salt:	405037 c/s real, 265641 c/s virtual

So an increase of multiple factors. Could you please provide me your benchmarks too, to see if you have such an improvment?

This increase was possible by switching to known-plain attack and thus getting rid of arc4'ing all edata2 but only 32 bytes. It also permitted to avoid last hmac-md5.
The know-plain attack is based on well-know ASN1 structures, but I am quite sure it is possible to still increase performances by relying on 8 bytes Nonce (I'll have to do some tests to check if it is the case).

I've tested all provided samples and also 6 others samples (tickets) from 6 differents domains (that I can't provide here as they are from work)

I had also to change the base code a bit because I was facing OpenMP problems


Fist0urs:
Right, I fix that :)

magnumripper:
For good measure I think you should add a full test when the known-plain indicates a match. The current code might give false positives (albeit rare).

magnumripper:
```
$ ../run/john-b4 -test -form:krb5tgs 
Will run 8 OpenMP threads
Benchmarking: krb5tgs, Kerberos 5 TGS [MD4 HMAC-MD5 RC4]... (8xOMP) DONE
Many salts:	843776 c/s real, 119177 c/s virtual
Only one salt:	792839 c/s real, 108065 c/s virtual

$ ../run/john -test -form:krb5tgs 
Will run 8 OpenMP threads
Benchmarking: krb5tgs, Kerberos 5 TGS etype 23 [MD4 HMAC-MD5 RC4]... (8xOMP) DONE
Many salts:	3856K c/s real, 620270 c/s virtual
Only one salt:	2609K c/s real, 398343 c/s virtual
```
:+1:

Fist0urs:
Nice bench!

you're right, I'll add the checksum part to be sure. To have something straitforward I'll change the format so that checksum is at the end. The immediat consequence is that actual format won't work anymore (I'll try to look at the "Nonce" case, tomorrow. This would resolve all problems)

PR #2030, opened 2016-02-08
Add verification of checksum if potential pass were to be found

Fist0urs:
Changed input format from edata1$edata2 to edata2$edata1
Thus had to use dyna_salt (and put back edata max len definition)

Performance should decrease a bit looking at --test, but this is not relevant as "success test" is to be done everytime in test cases. Thus we should expect quite as much improvment as previously posted

EDIT: Ok I've found that part that could be fixed (flags) is after some ASN1 structures, so we will face the same situation. I consider this implementation to be the final one as for the known-plain trick


magnumripper:
Technically we don't need any max data size as long as we're using dyna salt. But the current max should be fine for most situations anyway.

Also, I believe setting BENCHMARK_LENGTH to -1000 will work around the benchmark speed problem (this will cause benchmark [but not self-test] to use mangled plains) but I have no time to try it right now.

Fist0urs:
@magnumripper I've implemented the format in oclhashcat (still some little things to adjust before PR) thus I've been wondering if it would not be better to use format : checksum$edata2 (as previously) instead of edata2$checksum, as you pointed out that using dyna_salt would permit to get rid of max_data...

atm, my implementation on oclhashcat works for format: $krb5tgs$23$accountstuff$checksum$edata2 and this is more natural as checksum is at the beginning of edata. 

Should I rebase?

magnumripper:
For krb5tgs I think the most important thing, by far, is to ensure HC and JtR ends up supporting the same format. Second most important is that kirbi2john.py outputs it that way OR is changed to do so (committed upstream to whatever that project was... kerberoast?).

The only reason I'm aware of (off the top of my head) to place checksum last, is that the krb5pa-md5 format has it that way and it's a little confusing to differ from that (but definitely less important than the above). I'm not quite sure why that format was made like that - the original "mskrb5" format had checksum first and this is still supported by krb5pa-md5, but with another "legacy" tag.

Fist0urs:
Great, I'll make the changes then (and it will fix the issue noticed by @frank-dittrich that legacy version does not work anymore)

magnumripper:
:+1: and please try the `BENCHMARK_LENGTH = -1000` when you're at it!

PR #2039, opened 2016-02-13
Shrink the hash/spread function. 

claudioandre:
False positives < 0.05% (for 100,000 and 3,000,000 test cases).

- [x] CI
- [x] Tests


claudioandre:
- Fails on keystore.

claudioandre:
Merge-able

PR #2054, opened 2016-02-15
Trigger command when password found

akos-sereg:
Adding new configuration key in john.conf: ExecOnCrackedPassword
If this configuration is enabled, the value of this parameter (it should be a command) will be executed when a password is found, passing the login name and the cracked password as parameters.

### DIFF
--- a/run/john.conf
+++ b/run/john.conf
@@ -32,11 +32,11 @@ Save = 60
 # Beep when a password is found (who needs this anyway?)
 Beep = N
 
-# Shell script to be executed when a password is cracked.
+# Command to be executed when a password is found
 # If set, the following parameters will be passed when executing command: 
 #   $1 - Login
 #   $2 - Plain password
-ExecOnCrackedPassword = /home/akoss/dev/c/JohnTheRipper-bleeding-jumbo/run/success.sh
+#ExecOnCrackedPassword = ./password_found.sh
 
 # if set to Y then dynamic format will always work with bare hashes. Normally
 # dynamic only uses bare hashes if a single dynamic type is selected with
--- a/run/john.conf
+++ b/run/john.conf
@@ -31,6 +31,13 @@ Idle = Y
 Save = 60
 # Beep when a password is found (who needs this anyway?)
 Beep = N
+
+# Shell script to be executed when a password is cracked.
+# If set, the following parameters will be passed when executing command: 
+#   $1 - Login
+#   $2 - Plain password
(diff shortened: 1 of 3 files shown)


magnumripper:
Cool idea, and KISS implementation. I'll tweak it just a little.

frank-dittrich:
Passing Login and clear text password on the command line should at least be optional (i.e., an additional config option) and disabled in a default config.

akos-sereg:
It is disabled by default (config option commented out in john.conf)

frank-dittrich:
I meant you should be able to trigger a command without passing the login and/or clear text password on the command line. Otherwise, unauthorized users might see this sensitive information (e.g., using the ps command), especially if that script takes some time to complete.
(You couls always enable logging of user name and cracked password, make sure the file permissions prevent unprivileged users from reading that information, and parse the .log file contents in your script.
(May be the script would need the session name as a command line parameter to do that.)

Other possible problems you need to be aware of:
If the same password gets cracked by different processes / threads, you would execute the script several times.
For formats which split the hash (e.g., LM), you'd run that script even for partially cracked passwords.

If you can relax the requirement to immediately run you script whenever a password is cracked, you could use or adjust the run/mailer script.

What is your use case for this new feature?

magnumripper:
I can see several use cases. I can't see a desperate need to make login/password optional but sure, it wouldn't hurt.

The run/mailer script is a different beast, it's not run automatically at all.

magnumripper:
I made some tweaks in 4fb8c31

magnumripper:
I can see use for a corresponding command-line option (for executing script only for certain jobs, or different scripts for different jobs) but OTOH I don't want to add too much complexity or bloat.

frank-dittrich:
More or less I saw the "use case" of trying to take over accounts with cracked passwords as fast as possible. Hardly legal.

frank-dittrich:
Should we discuss this on john-users?

magnumripper:
Yeah why not.

magnumripper:
Another use is for formats that emit false positives.

akos-sereg:
I agree that there is a security concern about this feature (ps output was a good example), but I think the severity of the risk is up to the user: whether the target script is long-running, or the target script is writeable.

This would be a general exit point for john, and the purpose of this surface is to make it possible to notify external systems that a password got cracked.

Parameterless script execution sounds good, I am going to create a follow-up PR with that.

PR #2055, opened 2016-02-15
Making login/plain password argument passing optional (when executing password_found.sh)

akos-sereg:
For security purposes, default settings are harmless: trigger is switched off + plain password passing is switched off.

### DIFF
--- a/run/john.conf
+++ b/run/john.conf
@@ -34,9 +34,16 @@ Beep = N
 
 # Command to be executed when a password is found
 # If set, the following parameters will be passed when executing command:
-#   $1 - Login
-#   $2 - Plain password
+#   Login - if "PassArgLogin" config parameter is Y
+#   Plain password - if "PassArgPassword" config parameter is Y
+#
+# Security notes:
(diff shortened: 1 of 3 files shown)


magnumripper:
Not quite OK.

- You need to get the cfg_get_bool stuff out of the inner loops. They should be in log_init().
- I think the supplied sample script should print $2 if given (so it works just by changing john.conf)
- The sprintf logic is busted. It writes things that are then overwritten. It will never write both login and password as parameters, as currently written.

akos-sereg:
Please see my changes in 6251f96

magnumripper:
Did you test it thoroughly, all combinations? You should, we are not supposed to do it for you. Just from looking at the code I believe you now don't allocate space for a final null.

magnumripper:
Never mind, it looks like we're reverting this whole thing for now. See http://www.openwall.com/lists/john-dev/2016/02/15/1

akos-sereg:
ah okay, just added more space for the terminating char. 
decline it then

PR #2057, opened 2016-02-16
set ARCH_LITTLE_ENDIAN for powerpc64 little endian in src/ppc64.h

michelmno:


frank-dittrich:
Solar should apply that change as well.

Issue #2019, opened 2016-02-03
Build a tool to create and execute test cases


claudioandre:
Get some code from TS and build a tool that parses and execute test cases: 
1. read the test cases;
2. for each test case, execute the commands requested;
3. validate as TS does now: # of guesses ok, exit()=0, etc.

Examples:
```
<Prepare>
	../run/john -form:cpu --list=format-tests | cut -f3 >alltests.in
	../run/john --format=cpu --list=format-tests | cut -f4 > solution.dic
</Prepare>

<Clear>
	rm ../run/*.pot
</Clear>

<TestCases format=raw-SHA512-opencl>

<Case1 guesses=1>
../run/john alltests.in -form=raw-sha512-ng-opencl -mask=?u?u?uCAPS
</Case1>

<Case2 guesses=10>
../run/john alltests.in -form=raw-sha512-ng-opencl --wo:solution.dic
</Case2>
<TestCases>

<Execution>
  <Name=T1>
 	DO=Prepare()
	DO=Clear()
	DO=Case1()
	DO=Case2()
  </Name>
</Execution>
```

jfoug:
and how is this different/better than ```./jtrts.pl -internal``` ?

claudioandre:
I have to admit that I do not have a clear view of -internal. Anyway, I want to create test cases, to:
- automate the check of reported errors/avoid regressions.
- I would like to test if mask works: none, few and all cracked.
- I would like to test --skip-self-test 
- I would like to test --min-len and --max-len
- Etc.

The point is: I need to test. And I have to write these tests all the time.

magnumripper:
The best would be to add features to the TS.

claudioandre:
I probably will mess up things inside TS. So, I created a script just to keep tests saved.

I added it to /unused and will PR (to keep it saved and under VC)

claudioandre:
3e455de886d610520250fff3714134e4989e8692

PR #2079, opened 2016-03-07
Tool to save and run test cases. Address #2019.

claudioandre:


magnumripper:
Perhaps we should put this in TS directory instead?

claudioandre:
We could make the change, but I would like to add (at least) one new *to test* hash file.
- it reveals a bug in previous raw-sha256 [OpenCL].
- so, imagine more test cases can be put there. The recent zip regression, for example.

In order to avoid to mess things, should I create a sub-directory? 
TS
-> dic_make/
-> [claudio] OR [sha2-opencl]/ **(to be created)**
.Ts file
.Ts file
.Ts file

jfoug:
Put in TS, this PR is not needed.

Issue #570, opened 2014-04-19
Add session save/restore for --regex mode


magnumripper:
- [ ] on its own
- [x] in combination with --wordlist

jfoug:
I have been working on this, and it is NOT fun.

First, there are issues with rexgen save/resume. I am trying to work through all of those first.  I 'think' I have the save/resume code in JtR working semi OK, I just need to fix rexgen (with as small as possible of a  patch), and then remove a bunch of work around 'hacks' in JtR  The hacks could not get things done, because there is a latent bug that I could not hack my way around.

jfoug:
This CR was given to the rexgen page.  I think there will be another before I am done.

https://github.com/teeshop/rexgen/issues/28

NOTE, I can not incorporate my changes until the above issue is addressed.  We use rexgen in hybrid mode 1 word at a time.  For save/restore to work, we have to not create a new session for each word,but create the session 1 time, and then re-use it. But rexgen is in charge of reading words, AND there were real problems restarting a session with new words (until the above patch is done).

NOTE, there is also still a bug IN the resume.  Here are examples:

```
001_88           (?)
001_98           (?)
001_09           (?)
001_19           (?)
192g 0:00:00:04 0.07% (ETA: 00:57:09) 39.99g/s 39.99p/s 39.99c/s 39.99C/s 001_29..002_32
Session aborted

$ OMP_NUM_THREADS=2 ../run/john -res=tst-x
Loaded 1 password hash (restore_tester [None])
Will run 2 OpenMP threads
Note: This format may emit false positives, so it will keep trying even after
finding a possible candidate.
Press 'q' or Ctrl-C to abort, almost any other key for status
192g 0:00:00:05 0.05% (ETA: 02:17:23) 38.40g/s 38.40p/s 38.40c/s 38.40C/s 001_20
001_20           (?)
001_30           (?)
```

the regex is \0[0-9][0-9]    The 'proper' resume should be either 001_19 or 001_29 (my choice would be 001_19)

If you see where the session left off, when reloaded, it incremented BOTH of the sets (1 went to 2 and 9 went to restart or 0).  So from this regex, we are missing 10 candidates (well, in the above example, we actually are re-doing 90 candidates.

This happens all the time, here we see how for this regex we miss 10 items:

```
001_16           (?)
001_26           (?)
001_36           (?)
001_46           (?)
001_56           (?)
[...]

PR #2084, opened 2016-03-10
rexgen save-restore

jfoug:
This is not ready yet, but I wanted the pull request, so that others could see the changes also.

NOTE, this requires v1.4.x or greater of librexgen https://github.com/teeshop/rexgen/pulls

### DIFF
--- a/.pre-commit.sh
+++ b/.pre-commit.sh
@@ -36,7 +36,7 @@ fi
 ########################################################
 # Tests 2/3/4/5. Make sure source does not have execute bit set
 ########################################################
-TXT_FILE_PATTERN='\.(cl|c|h|txt|S|s|in|chr|conf|txt|md|asm|cin|dep|gost|legacy|macros|stub|m4|cu|bash_completion|zsh_completion|lst)(\..+)?$'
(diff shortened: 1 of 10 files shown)


jfoug:
Issue #570

magnumripper:
Can we merge this right now or should we wait for something in rexgen?

jfoug:
We have to wait. This will break RUNTIME of rexgen in john if merged now. Well, actually, what it will do it 100% disable rexgen until it is updated to v1.04.   Now, someone can pull the 1.04 from my git repo, but that is not the official release.

The issue is that we went from creating a new iterator for each word (so the bug in rexgen was not apparent), to using a single iterator over and over again as words are tested.  Rexgen requires the saved iterator to be able to resume, BUT it has a bug that its state is not reset properly when the callback function lists that the run for that word is done.

magnumripper:
So once this is fixed in rexgen we should probably bump our version requirement for that lib in configure, in order to "accept" it.

jfoug:
Check out the first patch. Already done that, BUT we really have to wait and see what actual version gets put into the lib.  From past history, he has accepted no changes from me, and has redone them himself (often wrong), and always changes things, thus the code done will have to adapt.

jfoug:
Created new PR, based upon current code. Closing this one out.

jfoug:
Completed with PR #2199

Issue #2082, opened 2016-03-08
Use restore_tester_fmt.c to find and correct restore issues


jfoug:
Things I know about.  Add more as they are seen/fixed, since some of these should get done in core also.

- [ ] rules.  When restarting, the rules are restarted for whatever words were being worked on.
- [ ] rexgen does not resume  #570 
- [ ] Multi-salt mode restarts all salt work over again

Some of these could be handled as 'hybrid' type blocks added to the end of a .rec file (I think). Things like the rules and multi-salts redoing work, could be addressed in that manner.  The 'normal' V4 processing of the rec file gets us back to the right words, and then the hybrid extra block appended provides additional information on how far to skip ahead inside the subwork done on the word or set of words, so that prior completed work is not replicated.

The restore_tester_fmt is great for determining exactly what was processed, written to .pot files, etc.  So that we know 100% that we restart at proper location, not before, and CERTAINLY not later in the word stream.

jfoug:
Here is what rules does (currently)

```
27e_01           (?)
27f_01           (?)
5904g 0:00:01:49 1.19% (ETA: 11:45:04) 54.16g/s 54.16p/s 54.16c/s 54.16C/s 280_01..2ff_01
Session aborted

$ ../run/john -res=tst-x
Loaded 1 password hash (restore_tester [None])
5904g 0:00:01:50 1.00% (ETA: 12:14:52) 53.67g/s 53.67p/s 53.67c/s 53.67C/s 000_01
000_01           (?)
001_01           (?)
002_01           (?)
003_01           (?)
```

here is the rec file:

```
Jim.Fougeron@1PQHGQ1 /c/projects/phpbb/johnripper/john-1.7.9/bleed64/test
$ cat *.rec
REC4
9
--rules=appendnumnum
--wordlist=tst-x.dic
tst-xx.in
--session=tst-x
--pot=tst-x.pot
--format=restore_tester
--input-encoding=ASCII
--target-encoding=ASCII
112
5936
1730
0
0
1730
0
1730
0
0
0
1
0
1
0
32
```

So it appears that under rules, only the rule# gets restored, and the run starts at the first input word on that rule.

So we need to save not only the rule# but also the word# for restore to work 'correctly'.

jfoug:
Rules bug (listed above), found and fixed.  Still more testing to do, but that one WAS a bug, and is corrected.

jfoug:
[...]

PR #2087, opened 2016-03-12
Multisalt resume

magnumripper:
Not ready for merging yet, but this PR can be used to look at the "sum" of diffs. See #2082

### DIFF
--- a/src/status.h
+++ b/src/status.h
@@ -41,9 +41,8 @@ struct status_main {
 	int compat;
 	int pass;
 	int progress;
-	int salt_idx;
-	int total_salts;
-	unsigned char salts_sha1[20];
+	int resume_salt;
+	uint32_t *resume_salt_md5;
 };
 
 extern struct status_main status;
--- a/src/status.h
+++ b/src/status.h
@@ -41,6 +41,9 @@ struct status_main {
 	int compat;
 	int pass;
 	int progress;
+	int salt_idx;
+	int total_salts;
+	unsigned char salts_sha1[20];
 };
 
 extern struct status_main status;
--- a/run/john.conf
+++ b/run/john.conf
@@ -119,11 +119,6 @@ SingleWordsPairMax = 6
 # with all other salts.
 #SingleRetestGuessed = N
 
-# If this is true, formats not having a "salt sort method" will
-# get their salts sorted "most used first". For most uses it should not
-# matter at all.
-AlwaysSortSalts = Y
-
 # Protect the restore files (*.rec) from being overwritten. The default
 # mode is "Disabled". This mode will provide no protection, but has been
 # the default mode in JtR forever, so to not change behavior, that mode
--- a/src/7z_fmt_plug.c
+++ b/src/7z_fmt_plug.c
@@ -311,10 +311,16 @@ static void set_salt(void *salt)
 
 static int salt_compare(const void *x, const void *y)
 {
+	int c;
 	const struct custom_salt *s1 = x;
 	const struct custom_salt *s2 = y;
 
-	return (s1->NumCyclesPower - s2->NumCyclesPower);
+	// we had to make the salt order deterministic, so that intersalt-restore works
+	if (s1->NumCyclesPower != s2->NumCyclesPower)
+		return (s1->NumCyclesPower - s2->NumCyclesPower);
+	c = memcmp(s1->salt, s2->salt, 16);
+	if (c) return c;
+	return memcmp(s1->iv, s2->iv, 16);
 }
 
 // XXX port Python code to C *OR* use code from LZMA SDK
--- a/src/cracker.c
+++ b/src/cracker.c
@@ -138,25 +138,6 @@ static void crk_help(void)
 	printed = 1;
 }
 
-int crk_update_cur_salt_checksum(unsigned char buf[20])
-{
-	SHA_CTX ctx;
-	struct db_salt *salt;
-	int total = 0;
-
-	SHA1_Init(&ctx);
-	/* does the salt list EVER change during runtime?  note, if so, */
-	/* then we 'could' simply recompute this info when it changes */
-	salt = crk_db->salts;
-	while (salt) {
-		++total;
-		SHA1_Update(&ctx, salt->salt, crk_params.salt_size);
-		salt = salt->next;
-	}
-	SHA1_Final(buf, &ctx);
-	return total;
-}
-
 void crk_init(struct db_main *db, void (*fix_state)(void),
 	struct db_keys *guesses)
 {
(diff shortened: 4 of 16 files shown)


magnumripper:
Apparently the latest commit fails bot testing

jfoug:
This part?

```C
/*here's also an experimental john.conf setting AlwaysSortSalts that,
 * if true, will fallback to sort "most used first" if the format does
 * not have a salt_compare method defined.
 */
```

magnumripper:
Yes, I wrote it [in this context](https://github.com/magnumripper/JohnTheRipper/commit/5015bc66ae400094d3ef563088ea077c4a395151#commitcomment-16664708)

jfoug:
Now I just have to spend some quality time testing 'REAL' work.  Stuff where i have millions of salts, almost all are singletons, running the stuff, resuming, having other processes also removing them, etc.

I also need to build some real world tests where there are large salt-hash data.  I am still not 100% sure that I have all that logic right.

But this is getting very close.  It does NOT change the RECOVER version at all, it optionally appends a salt recovery section.  But just because there is a salt recovery section does not mean we 'can' recover.  If the salt is not found, then we start over.

One thing that may impact recovery, is getting a .rec file built on a different system. It will md5 differently if there are BE/LE differences and ints in the salt structure. Also, alignment issues will make the md5's not match.  But that is OK, and there is no way around that. The amount of risk to having failed restore due to running on a different system is very small anyway, AND the resume will work, it will just redo salts.

jfoug:
Once the salts were in deterministic sorted order, this actually was not all that hard to pull off.  Easier than I thought. Since we already had fought through the salt sorting, it made that part SOOOO much simpler to get right.

jfoug:
Just did a compare against master.  an even 10,000 commits (exactly) !!!

magnumripper:
>> What was the reason for the sort by number?

> If one salt has 10 hashes, all the others just have one hash, your chance of a correct guess is 10 times as high for the first salt, compared to all the others.
That's the reason I think.

Yes. But it's really really not important at all. I just did it because I could. We don't need that functionality at all.

> IF there is a format which will not end up sorting salts in a deterministic manner across runs. i.e. if it HAS to put thing in order based upon what salts are still left. In that situation, where we would never know proper order from run to run ahead of time, then salt-resume is not possible, and should be flagged at the format level.

Why? It will work fine, reverting to legacy behavior, so that flag doesn't add much value. Also, we currently don't even have any such format, right?

jfoug:
> Why? It will work fine, reverting to legacy behavior, so that flag doesn't add much value. Also, we currently don't even have any such format, right?

It would not auto revert. The data would be loaded, the md5 checksums of each salt would be generated, and then resume would find the salt. However, since order was NOT the same as the prior run, finding the salt does not mean that all salts prior to it are done, nor all salts following it are not done. That is why the flag may (in the future) be needed. Yes at this time, there is no format that can not be made to store salts in a deterministic manner. 

But there could be, and if/when, then we will need a flag.  Right now, all salt sorting is done based upon the salt and ONLY the salt.  If there were some other criteria (like count of matching salts), then order (as @frank-dittrich first pointed out), will not be assured over runs.

However, we can skip the flag, until we need it.

src/status.h
```diff
@@ -41,6 +41,8 @@ struct status_main {
 	int compat;
 	int pass;
 	int progress;
+	int resume_salt;
+	uint32_t *resume_salt_md5;
```
magnumripper: Is the status struct really the correct place for this? The database seems like a better place for it.

jfoug: It was put there because of this (from recovery.c)

```C
	fprintf(rec_file, "%u\n%u\n%x\n%x\n%x\n%x\n%x\n%x\n%x\n"
	    "%d\n%d\n%d\n%x\n",
	    status_get_time() + 1,
	    status.guess_count,
	    status.combs.lo,
	    status.combs.hi,
	    status.combs_ehi,
	    status.crypts.lo,
	    status.crypts.hi,
	    status.cands.lo,
	    status.cands.hi,
	    status.compat,
	    status.pass,
	    status_get_progress ? (int)status_get_progress() : -1,
	    rec_check);
```

Everything else about the status of the current run's state is stored there.  The only thing that is questionable is the resume_salt flag (the trigger to cracker.c).  I placed that flag here out of convenience.  The pointer certainly belongs here.

magnumripper: OK, fine by me

jfoug:
Where do we stand on this pull?  The logic is right, and should be universal, now that salt sorting is correct for all formats.   The only question was how it was hooked into the .rec file.  I did it by leaving REC4 format intact, and appending a block to the end (like what was done for rexgen and exterenal.  All 3 (including multi-salt), allows john to restore to the proper 'word' correctly, and then these appended recovery 'states' adjust where WITHIN that word the work starts from. That is why I think the multi-salt resume should be handled just like the other 'sub' word restores.

magnumripper:
If all known issues are resolved I guess we can merge it now. Is that the case? Are all Frank's concerns dealt with? Is there any other thing we can think of that should be tested?

magnumripper:
Are all formats that have their own salt sorting dealt with? For example, WPA-PSK doesn't seem to be touched - don't we need new logic to guarantee stable order for same-SSID salts?

frank-dittrich:
Do we need to improve format self tests, i.e., for format implementing their own sort, make sure they never return 0 unless both salts are indeed identical?
This may be tricky.

frank-dittrich:
Also, once you add, remove or change a format's sort, you'll break restoring, i.e., you risk skipping some salts...

jfoug:
The return of zero means nothing frank.  what is supposed to happen, is that correct behavior for qsort is returned.  If the 2 salts ARE the same, then zero should be returned from a comp function.  However, we were returning 0 for some functions if something as trivial as the length of salt was the same.  I had to 'fix' that.  Also 100% of the value of the return MUST be from the salt, and not from other outside data (like count of same salts in the input file)

Here is from wpapsk (and why I did NOT need to change it)

```C
static int salt_compare(const void *x, const void *y)
{
	return strncmp((const char*)x, (const char*)y, 36);
}
```

That already is 100% perfect order, every time.

jfoug:
> Also, once you add, remove or change a format's sort, you'll break restoring, i.e., you risk skipping some salts... 

That is a non-issue. Yes, it should  be documented, but that is it.  It will only impact one restart, AND if so, then the user can be informed to edit the .rec file, removing the slt-v0 data and the entire salt group will be started over.

jfoug:
>  - don't we need new logic to guarantee stable order for same-SSID salts?

I do not believe so.  If there are 10 SSID's of NETGEAR then they all have the same salt, AND when resume goes, it simply does a linear search, finding the first one, and starting from there (which is the right salt).  If the order of hashes with the same salt is changed, that should not make any difference.  This can easily be tested, by building just a couple salts but having dozens of same ones for different hashes (using the restore_tester format). Then simply making sure it properly restarts.  I did this, but only with a few, and reran tests until the salt before, the salt that was duplicated, and the salt following the duplication were all the restart data.  Things worked properly.

magnumripper:
> Here is from wpapsk (and why I did NOT need to change it)

> ```C
static int salt_compare(const void *x, const void *y)
{
	return strncmp((const char*)x, (const char*)y, 36);
}
```

> That already is 100% perfect order, every time.

Did you test it? This sorts by SSID, and NOT the complete salt. So if you have a bunch of salts with the same SSID, their order will be random, will it not?

jfoug:
> Did you test it? 

No I did not test it.  I thought that was the whole salt.  Let me look further.

jfoug:
Do not merge yet....

Still to test,  multi-salt resume logic using rules.

I will update this message listing it is tested, and add a new message once I know the behavior.  I need to run a full suite of testing that I have already done, but make sure that rules were also used.

jfoug:
It looks like multi-salt with rules will get all candidates, BUT it is rerunning some words.  It looks like the multi-salt part is working, but the rules may not be restarting where it is expected to start.

I will do a bit more investigation.  It may be that this patch is fine, and there are issues in other parts of the code and the restore_tester format is simply showing other issues.

jfoug:
Ok, once the right build of john was used, multi-salt with rules tests 100% successful.  Exactly where things left off, the restore started at, every time.  No missed words.  No extra work.

I am going to merge this PR.  It is ready.

jfoug:
Should the multi-salt resume branch be dropped, now that it is merged?

magnumripper:
Yep, normally we always do so.

PR #2092, opened 2016-03-16
Do not use a #define as a function.

claudioandre:
- [x] CI
- [x] Testing


claudioandre:
I have to admit, I'm pretty bad with rules.

PR #2100, opened 2016-03-26
Spelling fix in a comment: equivelent to equivalent

solardiz:
Spelling fix in JimF's recent comment: equivelent to equivalent


frank-dittrich:
@magnumripper It is just a pull request.
You can ignore it, or apply that change, specifying Solar as the author.

But there's the same spelling error in other places.

```
 $ git grep -n equivelent
run/korelogic.conf:436:#these 2 are replaced by the equivelent above 1 rule.
run/korelogic.conf:439:#these are replaced by the equivelent above 2 rule lines.
src/jumbo.h:156: * proper working equivelent.  The _r version is thread safe. In the
src/sunmd5_fmt_plug.c:835:                       * this is the equivelent of the original code:
```

magnumripper:
We can't merge this, all those local merges has to be fixed. But we could cherry-pick 260d906.

PR #2101, opened 2016-03-27
Pushing Solar's recent commit.

claudioandre:
Spelling fix in a comment: equivelent to equivalent


claudioandre:
For the record:
- it is only Solar's recent commit (git am solar.patch) on a new and clean PR. 

*Safe to merge*.

PR #2112, opened 2016-04-06
Added Keepass 2.x keyfile format + debug Keepass 1.x

Fist0urs:
All cases within Keepass 2.x and Keepass 1.x should be treated


magnumripper:
Thanks!

PR #2114, opened 2016-04-12
Resume fixes

jfoug:
Fixes needed for many resume issues, also cleans up (significantly) the resume and stacked resume logic.

NOTE, needs more testing,  The PR is made, but do NOT merge just yet.


magnumripper:
I can't see how this would end up correctly but I'll be happy as a clam if you prove me wrong.

magnumripper:
I had a look at this but it's out of sync with HEAD. Some of it is already in there.

magnumripper:
@jfoug I think we should try to revive this PR and rebase it on current tree.

magnumripper:
#2308 replaces this one

PR #2126, opened 2016-04-26
Generic big hash helper

jfoug:
Handled super large lines in input/.pot files.

!!! Need to drop the 'sample' data before we merge !!!


jfoug:
!!! Need to drop the 'sample' data before we merge !!!

src/loader.c
```diff
@@ -157,7 +150,7 @@ static void read_file(struct db_main *db, char *name, int flags,
 			}
 		}
 		process_line(db, line);
-		if (ex_size_line)
+		if (ex_size_line!=line_buf)
```
magnumripper: Code style: Please add some air around the operator here.

magnumripper: Same for line 124

src/loader.c
```diff
@@ -1953,7 +1959,12 @@ static void ldr_show_pw_line(struct db_main *db, char *line)
 
 	if (found && show) {
 		if (source[0])
+		{
 			printf("%c%s", db->options->field_sep_char, source);
+			/* with change to fgetll() vs fgets() source may no longer be \n term */
```
magnumripper: Isn't it a latent bug in fgetll() that should be fixed, instead of worked around here?

PR #2129, opened 2016-05-04
Always inline PoC

magnumripper:
NOT READY FOR MERGE

Here's a PoC for "always inline" without intrusive changes to loader/logger. Needs more testing and only changes RAR as of now. ~~It also does not yet use Base64 (it definitely should before merge)~~. See #2117


magnumripper:
`./jtrts.pl --internal rar` fails the pot-check but that is simply due to how that check works (it tries to use the truncated pot entry as input). This should be worked around in TS somehow, possibly simply by disabling that specific test for affected formats.

jfoug:
Have fun.  It's yours now.  The code you have started on now, is about where I was 2 weeks ago, with non of the road blocks looked at.  But it's yours now.

magnumripper:
Please show me where you had code anywhere near this one? I only saw core hacks of varying merits and wanted to try attacking it from the opposite angle.

Also please tell me why this experiment of mine *in a topic branch* hurt your feelings? I'm just tossing ideas and trying out solutions just like you. In the end we'll obviously pick whatever ends up best.

jfoug:
Its not that I had 'hurt feelings'.  It was that I had the work done, and working fine (in core, I have not yet started on ANY *2john changes), only to have that idea poopoo'd to move the work over into the *2johns,and have all of their signatures reworked.  Ok, I can handle re-work and more time spent.  started making the core changes to allow for this, and again got things ready to roll, but again, still NO changes to any of the *2john code, since doing that for all the 1'offs simply does not make sense until the core changes are fully functional and 'set in stone'.

Well it seems to me that again, that work and time put in has been ignored with a 'here, this is what I mean' set of work.  Well, that's fine, what you have done and your ideas on this may be great.  But I am not going to continue to spend a lot of time on this at this time.  I simply do not have time to spend if what is done is not going to end up used.  

Now, in the end, the direction you are going may be great, and encompass all that is needed.  I have no heartburn what so ever either way.  If however, this does not pan out in the end, or if there are problems which creep in causing your streamlined ideas to need to expand, and expand again, etc, then I will pick back up where I left off.  

I still feel my original methods were the way to go.  Nothing changed in signatures,  Nothing changed in any of the *2john files. Things just worked properly.  But I will simply wait and see where your testing code goes.

jfoug:
Just keep in mind with all the changes you are looking  to do for the *2johns, that you will also have to likely update things like:

pass_gen.pl
jtrts.pl
relbench (or whatever tool is used to unify pre/post hash changed).

magnumripper:
I cherry-picked some of your recent commits as a base for this branch. That work is not lost.

> It was that I had the work done, and working fine (in core, I have not yet started on ANY *2john changes), only to have that idea poopoo'd to move the work over into the *2johns,and have all of their signatures reworked.

If you hate when your work end up not used, I suggest you take time to **discuss** intrusive core changes before implementing them. Your work started out as very dubious truncation hacks in loader/logger. That was clearly a bad idea so I raised concerns and tried to discuss better solutions. Instead of discussing, you kept hacking away in core between each time I logged in, band-aiding on the poor start.

I'm showing you a KISS solution here that doesn't rely on core changes. Most of my changes is actually just dropping the code that handled external files/pointers.

jfoug:
The original KISS solution, is what you are using.  That is where i started from.

Where things deviated from the simple KISS solution (which is simply overcoming the line_buffer length) came when I realized that putting these monster lines into the .pot file is absolutely asinine.  As input data, yes these huge lines make sense.  As a 'key' to tell us which password was found, that is absolutely not a good deal.

So where this is at now, is back to the original easy to do, avoidance of the max line buffer size. That is a 'good thing', but we are now back to how to deal with the .pot file.

jfoug:
You may not have liked how things were done, but I still think the best solution so far, was the hash builder within logger code.  If you followed along, you will see all of the nuanced gotchas that were seen along the way. One of the larger of these is the usage of the max_line_size in logger, and the locking.  Especially now that max len is back to 1k, there are a LOT of lines of data which greatly exceed that, even if they were happy before.  Well, try and work around that in a KISS manner.  

Already with this KISS code here in this patch, I will assure you that the code in logger will have ALL kinds of problems, and none of them are addressed.  You may still be under the belief that this is easy to achieve but you will find it is far from that, and problems from this simple change (actually  from JUST the simple change of reducing max line length back to 1k) will spider into numerous places within core.   

I too agree that the divergence from cores max line length was a monumental BAD IDEA. But reverting it back out is NOT going to be trivial. We have SO many formats which are much longer than 1k all the time.  super john does not have to deal with any of these 'real' world hashes. it is so limited in its scope, that these type issues can be ignored.  Well in jumbo that's not the case.

So in this direction this patch is going, how are you proposing handling all of the input data that is between 1k and 0x30000 bytes in length, which jumbo will have to handle. Numerous of these DO NOT have *2john's to deal with.  Also, are you going to change signatures of 30 to 50 formats?  Pretty big change, and certainly not KISS IMHO.  It may seem very simple when you are only focusing on 1 format. Yes making the changes, and hooking it in work pretty easily.  But in the holistic view, where all formats are in play, might this be a much bigger and more error prone task, than simply handling this properly and automatically within the running code, and within non-hot areas?

If your direction pans out all the way to completion, I will be very happy. But for now, I will wait and watch, and provide help if you want some, but hold off on any work on the direction I was heading.

magnumripper:
> So where this is at now, is back to the original easy to do, avoidance of the max line buffer size. That is a 'good thing', but we are now back to how to deal with the .pot file.

Not sure what you mean. My patch fixes that (but only for RAR in this PoC).

Input lines are
```
$RAR3$*2*hex(salt)*hex(crc)*PACK_SIZE*UNP_SIZE*method*hash(blob)*base64(possibly HUGE full encrypted file blob)
```
(The currently published version use hex instead of Base64 but final version would use Base64).

Pot lines are
```
$RAR3$*2*hex(salt)*hex(crc)*PACK_SIZE*UNP_SIZE*method*hash(blob)
```

And this all happens within the format, basically in split(). Yes, we'd need to fix all data-inlining formats one by one. But we'd need to fix each of them anyway, removing old pointer-to-file code and so on. And since we no longer inline files or even support that, obviously the signature **already** changed.

**However**, the above is, and I was, totally focusing on the very few formats that *inline - or point to - files*. Like RAR. Not formats that simply use a very large datablob but never pointed to external files (because they never exceed that 0x30000).

So I'm not sure, you may be right. I'll try and assess the situation in a few days. If we can avoid those core hacks, I sure would like to do so. If we can't, well, then my RAR PoC will still be "usable" it wont be needed.

magnumripper:
I would really like @solardiz to chime in here, in case he has any ideas or plans for core that will support or disqualify one or both of our solutions.

solardiz:
I currently have no relevant plans for core. So I think this is for the two of you (JimF and magnum) to decide. Thank you for working on it!

What might or might not be relevant is that yescrypt will provide a generic string encoding, and may export generic encoding/decoding functions, which could be reused. However, those emphasize compact encoding (of both integer parameters and binary data) rather than human friendliness of the encoding. For encoded strings that are pretty long either way, your current plan of including decimal/hex/Base64 components may be more appropriate.

PR #2130, opened 2016-05-08
Generic big hash helper

magnumripper:
DO NOT MERGE YET

This is @jfoug's code for #2117, I'm adding this PR just to find it and be able to compare stuff with ease.


magnumripper:
Rebased onto current bleeding, and added some fixes (CRLF endings). Hacked rar2john so it always inlines.

I also had to add a speed hack in 6e62291 because fgetll() was WAY to slow. We may want to solve that in a different way, I just made a quick hack doubling alloc size each time. While it's orders of magnitudes faster for large files, it may end up allocating nearly twice needed memory.

magnumripper:
I noticed your truncation ends up at exactly LINE_BUFFER_SIZE **before** adding the field separator and the cracked password, instead of after. Is this intended? I just have a feeling it adds a vector for bugs to attack us later.

src/logger.c
```diff
@@ -103,7 +103,10 @@ static void log_file_init(struct log_file *f, char *name, int size)
 	    O_WRONLY | O_CREAT | O_APPEND, S_IRUSR | S_IWUSR)) < 0)
 		pexit("open: %s", path_expand(name));
 
-	f->ptr = f->buffer = mem_alloc(size + LINE_BUFFER_SIZE);
+	/* plain will now always be < LINE_BUFFER_SIZE. We add some extra bytes
+	 * so that there is ALWAYS enough buffer to write our line, and we no
+	 * longer have to check length before a write (.pot or .log file) */
+	f->ptr = f->buffer = mem_alloc(size + LINE_BUFFER_SIZE + 512);
```
magnumripper: What are this "some extra bytes" for exactly, and what makes 512 a good figure? Is this actually the headroom for adding a field separator plus a cleartext to the ciphertext (that was truncated to LINE_BUFFER_SIZE)? In that case we might want to simply use PLAINTEXT_BUFFER_SIZE instead of 512.

src/loader.c
```diff
@@ -95,12 +96,96 @@ static char *skip_bom(char *string)
...
+int ldr_isa_pot_source(const char *ciphertext) {
+	if (strlen(ciphertext) <= LINE_BUFFER_SIZE) {
+		const char *cp = strstr(ciphertext, "$SOURCE_HASH$");
+		if (!cp)
+			return 0;
+		return !!strstr(cp, "$SOURCE_BIN$");
+	}
+	return 0;
+}
+
```
magnumripper: The way this function is implemented above paints us into a corner if we ever want to change LINE_BUFFER_SIZE. So despite the performance impact, we might want to always do the `strstr(ciphertext, "$SOURCE_HASH$")`.

But this function won't ever be called unless we're actually using a format with potentially long lines, right? I certainly hope it will not.

src/krb5_tgs_fmt_plug.c
```diff
@@ -122,15 +122,37 @@ static int valid(char *ciphertext, struct fmt_main *self)
...
+		if (!ldr_in_pot || strlen(ciphertext) != LINE_BUFFER_SIZE)
+			goto err;
+		len = hexlen(p);
+		if (len < LINE_BUFFER_SIZE - 100 || len > LINE_BUFFER_SIZE)
+			goto err;
+		p += len;
+		if (*p != '$' || !strstr(p, LDR_TRIMMED_POT_BIN_SIG))
+			goto err;
+		return 1;
+	}
```
magnumripper: There's some hard-coded arbitrary (?) figures here we should probably avoid (or at least add a FIXME comment for later review) or document. How were "500" and "100" chosen?

src/loader.c
magnumripper: BTW if we always did the strstr, would that mean we can use this function in krb5_tgs too?

src/loader.c
```diff
@@ -1102,9 +1189,12 @@ struct db_main *ldr_init_test_db(struct fmt_main *format, struct db_main *real)
...
 	bench_running++;
 	while (current->ciphertext) {
-		char line[LINE_BUFFER_SIZE];
+		char *ex_len_line = NULL, _line[LINE_BUFFER_SIZE], *line = _line;
 		int i, pos = 0;
-
+		if (strlen(current->ciphertext) > LINE_BUFFER_SIZE-200) {
+			ex_len_line = mem_alloc(strlen(current->ciphertext)+300);
+			line = ex_len_line;
+		}
```
magnumripper: What are "200" and "300" here? Should we derive them from some macro constants instead?

src/logger.c
jfoug: The 512 was way more than needed.  It allows all these 'checks' to be ignored (sep char, password, etc). It is not used in ANY computation, it is just extra waste bytes allocated.

Doing that allowed quite a lot of worthless bullet-proof checking to simply go away.

magnumripper:
Rebased again since I really needed 7342aaffc to be in there. Also squashed many of @jfoug's commits, they are exactly the same but fewer. Added some tweaks and fixes from testing.

All formats seem to pass self-tests as well as the Test Suite (--internal).

Also made a number of performance tests (which is how I found 7342aaffc which was introduced in 9aaafc7d and unrelated to this new stuff). Everything looks good.

jfoug:
I have not had a chance to look yet, but you did mention one thing that needs to be addressed. That is the slow growth I used in fgetll() 

Is that change in the latest stuff?  What I had was adequate up to a point, but then quickly showed slowness due to too many reads.  But in hind site, how often is one going to read in HUGE lines.  There usually will be only 1 file.  We need to be careful in this case, in that we do not cause the load to fail, because our buffer allocation increases were too aggressive.  But I will leave that with you for now.  It is not  something I think will be used all that much, BUT it is something we should think through a bit from both directions.

src/misc.c
```diff
@@ -162,15 +162,13 @@ char *fgetll(char *s, size_t size, FILE *stream)
...
+	cp = strdup(s);
+
 	while (1) {
-		cp2 = cp;
-		cp = mem_alloc(len+60000);
-		strcpy(cp, cp2);
-		if (cp2 != s)
-			MEM_FREE(cp2);
+		cp = realloc(cp, 2 * len);
+
```
jfoug: cp = realloc(cp,...)  is BAD practice.  Read the info about realloc. It is REALLY a hairy function.  You should never realloc a pointer using the same pointer.  Always use a different pointer (or save the original one off).

magnumripper:
Yeah my current code in fgetll() is just temporary. Although I can't see it actually hurt. Unless there really is some really huge line somewhere, it will normally me a lot LESS than the previous 0x30000.

Even if we end up reading a gigabyte line and have maximum worst case so we allocate 2 GB, that is only virtual memory. Half of those pages will never be touched so it's really not an issue other than a large figure in top(1).

src/misc.c
magnumripper: Yeah, I will change that. I just needed a quick fix for testing huge (really huge) RAR files. The original code took several minutes to reach a couple hundred megabytes (out of 2 GB) and just kept getting exponentially slower so it would never finish...

jfoug: Yea.  The original code itself was just a hack. It 'works', but was not optimal. For the initial size I was working with, it was fine. But for having to deal with data in the 100k to multiple GB in size, it was not adequate.

src/loader.c
```diff
@@ -1954,8 +2059,7 @@ static void ldr_show_pw_line(struct db_main *db, char *line)
 	if (found && show) {
 		if (source[0])
 			printf("%c%s", db->options->field_sep_char, source);
-		else
-			putchar('\n');
+		putchar('\n');
 	}
```
magnumripper: This change still bugs me: Not that I mind it, I just want to understand just WHAT changed that made this change necessary. And I simply can't see what it is! Actually, I'm more confused with the OLD version. Where the heck did it get a newline from?

I tried putting fgets() back (and just running short hashes) and sure as hell, the above change need to be reverted or there will be double linefeeds. But WHY?!

jfoug: The change is in behavior of fgets vs fgetll   fgetll removes the \n just like fgetl did

magnumripper: Oh right, I thought it changed from fgetl to fgetll. But it actually used fgets before. That explains it! Hard to see with naked eye when mind was decided.

src/misc.c
magnumripper: Did some changes in 6127c1f.

magnumripper:
@jfoug when/why is the SOURCE_BINARY needed? I don't quite understand why we would ever need it. Is it used with --show?

magnumripper:
> when/why is the SOURCE_BIN needed?

OK I see it's used by ldr_init_password_hash(). I wonder if we could get away with storing the hash instead of the full binary.

magnumripper:
I think I'll have enough time this week to test this and merge. If at all possible, I will first get rid of the SOURCE_BIN but it might not be possible (or end up too ugly to be worth it). Failing that, I will merge more or less what we have here right now and add fixes for any *2john tools that needs tweaks (all of them should now always inline, etc.).

jfoug:
The source_bin was a bit ugly, BUT it was so trivial to add, since the binary function is a required function for all hashes.  Yes, it is only (AFAIK) used during loading and for a hash.  But providing the binary allows for it to work properly vs trying to store some hash, which may or may not change in the future.

magnumripper:
My ultimate goal is to have NO pot line ever exceed LINE_BUFFER_SIZE - if possible. The current code always blows that size after truncation, by design. This isn't a problem in JtR itself, but I have external things in mind.

So ideally we should only add `$SOURCE_HASH$<md5 hash>` and the truncation should give room for that **within** LINE_BUFFER_SIZE. If I can't find a way to do it without blowing performance when removing already cracked hashes, I will just give in. But I want to give it a shot.

magnumripper:
Rebased upon current bleeding-jumbo. Then added d094b75 which seems to do what I wanted: A full pot line will never exceed LINE_BUFFER_SIZE now, even including the field separator and the cleartext. It seems to work fine and was easier than I feared. So far I've tested some manually plus --test, --test-full and a full Test Suite run. Will continue testing.

jfoug:
In all, I was kind of surprised at how easy it was to use the 'big buffer fixes'   But as other things were tossed into the mix, it did grow in complexity.

I am still 'recovering' from the trip, and will be out on another this weekend (it is summer btw, lol).  I have not looked at the code you have done, just followed along with the email chain as it has grown.

magnumripper:
Yeah the sun wins over the laptop backlight for sure, and no GPU can beat a v-twin hotrod bike. No problem, we'll just proceed slowly. I think this is it now, just a tad more testing needed. Whenever it rains...

jfoug:
rain??  Hell, a tornado did not stop me when I was on the iron butt run (no shit).

Issue #2118, opened 2016-04-17
Create a Snappy Ubuntu package for JtR


claudioandre:
* Snappy works on:  Raspberry Pi 2, Intel NUC, Azure,  Google Compute Engine cloud,  Amazon Elastic Compute Cloud, Beagle, Duovero, ...
* Snappy packages are going to be supported on Ubuntu 16.04 LTS [1].
* It provides transactional updates with rigorous application isolation.
* It has an improved security model.

So, when we have this available, users can install JtR via Ubuntu Store.

[1] Ubuntu 16.04 LTS introduces the snappy Ubuntu Core experience to the desktop by allowing you to create, install and distribute snaps (snappy apps).

claudioandre:
Thinking loud:
- one package can handle (JtR already does this) AVX, SSE, XOP, ...
- not sure: one package with MPI and one without it?
- improve OpenCL runtime detection to use only one package?

claudioandre:
Ubuntu Core (150MB download) running inside kvm.

$ ./john -test=0
[...]
All 343 formats passed self-tests!
**********

The basic stuff works fine, but the overall quality is low. Anyway, when the environment matures, it might be very useful.

$ kvm -m 512 -redir :8090::80 -redir :8022::22 ubuntu-15.04-snappy-amd64-generic.img &
$ scp -CP 8022 JtR/*.snap ubuntu@localhost:~
$ ssh -p 8022 ubuntu@localhost
$ sudo snappy install john-the-ripper_1.8.0-Jumbo-2.dev_amd64.snap

claudioandre:
$ john-the-ripper.john-the-ripper  --list=build-info
Version: 6421abe
Build: linux-gnu 64-bit SSE2-ac OMP
SIMD: SSE2, interleaving: MD4:3 MD5:3 SHA1:1 SHA256:1 SHA512:1
$JOHN is /snaps/john-the-ripper.sideload/LWYlFDKCcTRp/bin/
Format interface version: 13
Max. number of reported tunable costs: 3
Rec file version: REC4
Charset file version: CHR3
CHARSET_MIN: 1 (0x01)
CHARSET_MAX: 255 (0xff)
CHARSET_LENGTH: 24
SALT_HASH_SIZE: 1048576
Max. Markov mode level: 400
Max. Markov mode password length: 30
gcc version: 5.3.1
GNU libc version: 2.23 (loaded: 2.21)
Crypto library: OpenSSL
[...]

PR #2137, opened 2016-05-14
OpenCL build works without any installed device (WIP).

claudioandre:
* It allows a user to run an OpenCL build of JtR without any driver installed or GPU connected (still requires the OpenCL ICD Loader).
* It handles the removal of all GPUs without requiring a JtR rebuild.

@magnumripper, your opinion about merge-ability of the patch:
* In the past we were, actually, opening the requested --device on john_init(). So, a 100% CPU format will be using resources on GPU.
* Now, we are only querying what is available (as we could do about memory, files, ...).
* Even the -format=CLASS can be solved.

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -350,8 +350,11 @@ static void print_usage(char *name)
 		exit(0);
 
 	printf(JOHN_USAGE, name);
-#if defined(HAVE_OPENCL) || defined(HAVE_CUDA)
+#if defined(HAVE_CUDA)
 	printf("%s", JOHN_USAGE_GPU);
+#elif defined(HAVE_OPENCL)
+	if (!none_opencl_device)
+		printf("%s", JOHN_USAGE_GPU);
 #endif
 	printf("%s", JOHN_USAGE_FORMAT);
 	exit(0);
@@ -370,7 +373,8 @@ void opt_print_hidden_usage(void)
 	printf(", cuda");
 #endif
 #ifdef HAVE_OPENCL
-	printf(", opencl");
+	if (!none_opencl_device)
+		printf(", opencl");
 #endif
 #endif
 #ifdef _OPENMP
@@ -409,8 +413,10 @@ void opt_print_hidden_usage(void)
 	puts("--internal-codepage=NAME  codepage used in rules/masks (see doc/ENCODING)");
 	puts("--target-encoding=NAME    output encoding (used by format, see doc/ENCODING)");
 #ifdef HAVE_OPENCL
-	puts("--force-scalar            (OpenCL) force scalar mode");
-	puts("--force-vector-width=N    (OpenCL) force vector width N");
+	if (!none_opencl_device) {
+		puts("--force-scalar            (OpenCL) force scalar mode");
+		puts("--force-vector-width=N    (OpenCL) force vector width N");
+	}
 #endif
 #if HAVE_LIBGMP || HAVE_INT128 || HAVE___INT128 || HAVE___INT128_T
 	puts("\nPRINCE mode options:");
--- a/src/common-opencl.c
+++ b/src/common-opencl.c
@@ -113,7 +113,7 @@ typedef struct {
 	int num_devices;
 } cl_platform;
 static cl_platform platforms[MAX_PLATFORMS];
-
+int none_opencl_device = 1;
 
 cl_device_id devices[MAX_GPU_DEVICES];
 cl_context context[MAX_GPU_DEVICES];
@@ -416,7 +416,7 @@ static int get_if_device_is_in_use(int sequential_id)
 	return found;
 }
 
-static void start_opencl_environment()
+void start_opencl_environment()
 {
 	cl_platform_id platform_list[MAX_PLATFORMS];
 	char opencl_data[LOG_SIZE];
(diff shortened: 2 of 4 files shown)


src/options.c
```diff
@@ -351,7 +351,8 @@ static void print_usage(char *name)
 
 	printf(JOHN_USAGE, name);
 #if defined(HAVE_OPENCL) || defined(HAVE_CUDA)
-	printf("%s", JOHN_USAGE_GPU);
+	if (!none_opencl_device)
+		printf("%s", JOHN_USAGE_GPU);
```
magnumripper: In case someone has CUDA but (for whatever reason) not OpenCL, we need to handle that.

magnumripper: BTW does this work now?
```
./configure --enable-cuda --disable-opencl
```

magnumripper: Perhaps the `none_opencl_device` should be in common-gpu.[ch] so it's there ever for a CUDA-only build. Or we need to have different logic, like
```C
#if defined(HAVE_OPENCL)
	if (!none_opencl_device)
		printf("%s", JOHN_USAGE_GPU);
#elif defined(HAVE_CUDA)
	printf("%s", JOHN_USAGE_GPU);
#endif
```

claudioandre: > BTW does this work now at all?

Oops, I introduced a bug.

claudioandre: I did it.

magnumripper:
Perhaps we should refactor the variable from `none_opencl_device` to `ocl_no_device` to keep our traditions. All variables and stuff exported from common-opencl.c should be prefixed with `ocl_`. We do break that rule for some other names for historical reasons but we shouldn't introduce new ones.

claudioandre:
Fails on --fork=n.

When the library is loaded before fork it shares more memory between forked processes than it can, and it behaves badly.
- I need to provide a way to "restart" the library;
- or give up the idea;
- or ...

claudioandre:
Maybe?!

```
typedef cl_int (*CLGETPLATFORMIDS) (cl_uint, cl_platform_id *, cl_uint *);
typedef cl_int (*CLGETDEVICEIDS) (cl_platform_id, cl_device_type, cl_uint, cl_device_id *, cl_uint *);

#if __linux__ && HAVE_LIBDL
CLGETPLATFORMIDS cl_GetPlatformIDs = NULL;
CLGETDEVICEIDS cl_GetDeviceIDs = NULL;
#endif /* __linux__ && HAVE_LIBDL */

cl_uint opencl_get_num_devices()
{
	cl_platform_id platform_list[MAX_PLATFORMS];
	cl_uint num_platforms, num_device, total_devices = 0;
	void *cl_lib;
	int i;

	if (!(cl_lib = dlopen("libOpenCL.so", RTLD_LAZY)))
		return 0;

	cl_GetPlatformIDs = dlsym(cl_lib, "clGetPlatformIDs");
	cl_GetDeviceIDs = dlsym(cl_lib, "clGetDeviceIDs");

	if (!cl_GetPlatformIDs || !cl_GetDeviceIDs)
		return 0;

	/* Find OpenCL enabled devices. We ignore error here, in case
	 * there is no platform and we'd like to run a non-OpenCL format. */
	cl_GetPlatformIDs(MAX_PLATFORMS, platform_list, &num_platforms);

	for (i = 0; i < num_platforms; i++) {
		cl_GetDeviceIDs(platform_list[i], CL_DEVICE_TYPE_ALL,
			MAX_GPU_DEVICES, NULL, &num_device);

		// Point to the end of the list
		total_devices += num_device;

#ifdef OCL_DEBUG
		fprintf(stderr, "OpenCL platform %d: %d device(s).\n",
		        i, num_device);
#endif
	}
	dlclose(cl_lib);
	ocl_no_device = (total_devices == 0);
	return total_devices;
}
```

claudioandre:
As soon as I call clGetPlatformIDs, it fails for a --fork run. dlclose() does nothing.

So, I guess I won't be able to work around this:
- best is to avoid to register an OpenCL format if none OpenCL dev exists. But I need to do some OpenCL stuff before forking.
- my original approach works with --fork but has a lot of downsides: see #2118.

magnumripper:
Too bad. Well it may not be worth spending too much time on it.

claudioandre:
Using latest bleeding source I tried to move OpenCL initialization inside john_fork() and the result was the same: no way out of the trap.

claudioandre:
Checking if a new link-map could help. Verbose debug with dlmopen(LM_ID_NEWLM, ...):
```
     27695:	symbol=XcmsCIEXYZToCIExyY;  lookup in file=/usr/lib64/libX11.so.6 [1]
     27695:	binding file /usr/lib64/libX11.so.6 [1] to /usr/lib64/libX11.so.6 [1]: normal symbol `XcmsCIEXYZToCIExyY'
     27695:	
     27695:	calling init: /usr/lib64/libXau.so.6
     27695:	
     27695:	
     27695:	calling init: /usr/lib64/libxcb.so.1
     27695:	
     27695:	
     27695:	calling init: /usr/lib64/libX11.so.6
     27695:	

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7def964 in add_to_global () from /lib64/ld-linux-x86-64.so.2
```

Issue #2132, opened 2016-05-09
All our base64_convert.[ch] helper functions use int instead of size_t


magnumripper:
We will need to rewrite all those helper functions to use size_t for sizes/lengths in order to be usable for things like rar2john with files larger than a gigabyte or so (depending on encoding - a larger than 1 GB binary will become larger than 2 GB of hex, overflowing signed integer).

I started fixing it but ended up scrapping it and post-poning because some functions currently return negative for error, which will not be possible with size_t. So it's not quite as simple as changing some "int" to "size_t".

This is probably not a very urgent issue and it's not needed at all until we merge the "arbitrary length input lines" code discussed elsewhere. And anyway it should be fairly uncommon to end up needing to inline that much data. OTOH it's an easy fix once we figure out how to change that error return (perhaps simply return `__SIZE_MAX__`).

jfoug:
Which helpers were you talking about?

base64_convert.[ch]   ?
misc.[ch] ?
common.[ch] ?


Somewhere else.

magnumripper:
Like I said, base64_convert.[ch]. Most other function will never see really large data but personally I will probably try to use size_t more from now on. Things change rapidly.

jfoug:
5b59b08

magnumripper:
That commit has got to be broken. Like I wrote in the OP:

*I started fixing it but ended up scrapping it and post-poning** ***because some functions currently return negative for error, which will not be possible with size_t.*** *So it's not quite as simple as changing some "int" to "size_t".*

magnumripper:
I guess you should revert this until you fix it properly.

magnumripper:
No, looks like I'm wrong. Those functions still return int and the internals look good.

magnumripper:
No I was right (lol). Here's one
```c
/*
 * if full string is HEX, then return is positive. If there is something
 * other than hex characters, then the return is negative but is the length
 * of 'valid' hex characters that start the string.
 *
 * NOTE, if the length is odd, then the 'full' length will not be returned.
 * the length returned will be length-1 since it would not be proper to try
 * to hex convert the last odd byte.
 */
static MAYBE_INLINE size_t _hexlen(const char *q, unsigned char dic[0x100])
{
	const char *s = q;
	size_t len = strlen(q);
	if (len&1) --len;

	while (dic[ARCH_INDEX(*q)] != 0x7F)
		++q;
	if ((size_t)(q - s)&1) --q;
[...]

PR #2168, opened 2016-07-07
base64: changed int into size_t

jfoug:


magnumripper:
#2132

jfoug:
This one is ready.  All tests pass.  ASAN has no problems.  using base64_convert.exe in test mode (with my testing script in jtrts directory) all works fine.

Yes, a lot of format changes, but it was simply adding the new 'null' variable listing we do not care about the error return.

magnumripper:
Did you try base64_convert.exe with, say, a 3 GB file? I'm thinking I'll try to produce a 3 GB (compressed) RAR file and see if it goes through rar2john and then gets cracked. But I didn't get around to it yet. However, I think we can merge this for now anyway.

jfoug:
/c/phpbb/johnripper/bleed-64/src $ ../run/rar2john p1.rar > in1
file name: x1.rar
mem_calloc(): Cannot allocate memory trying to allocate 18446744070987741568 bytes
/c/phpbb/johnripper/bleed-64/src $ ls -l p1.rar
-rwxr-x---+ 1 barb barb 2.8G Jul  9 09:21 p1.rar*

jfoug:
$ ../run/rar2john -v p1.rar
! HEAD_SIZE: 59, PACK_SIZE: 18446744072348646080, UNP_SIZE: 18446744072348524060
! HIGH_PACK_SIZE present
! HIGH_UNP_SIZE present
file name size: 6 bytes
file name: x1.rar
! EXT_TIME present with size 5
! Dictionary size: 4096 KB
mem_calloc(): Cannot allocate memory trying to allocate 18446744070987741568 bytes

jfoug:
```diff
/c/phpbb/johnripper/bleed-64/src $ git diff rar2john.c
diff --git a/src/rar2john.c b/src/rar2john.c
index 1339bee..8060983 100644
--- a/src/rar2john.c
+++ b/src/rar2john.c
@@ -332,20 +332,35 @@ next_file_header:

                file_header_head_size =
                    file_header_block[6] << 8 | file_header_block[5];
-               file_header_pack_size = file_header_block[7] +
-                       (file_header_block[8] << 8) +
-                       (file_header_block[9] << 16) +
-                       (file_header_block[10] << 24);
-               file_header_unp_size = file_header_block[11] +
-                       (file_header_block[12] << 8) +
-                       (file_header_block[13] << 16) +
-                       (file_header_block[14] << 24);
+               //file_header_pack_size = file_header_block[7] +
+               //      (file_header_block[8] << 8) +
+               //      (file_header_block[9] << 16) +
+               //      (file_header_block[10] << 24);
+               file_header_pack_size = file_header_block[10];
+               file_header_pack_size <<= 8; file_header_pack_size += file_header_block[9];
+               file_header_pack_size <<= 8; file_header_pack_size += file_header_block[8];
+               file_header_pack_size <<= 8; file_header_pack_size += file_header_block[7];
+
+               //file_header_unp_size = file_header_block[11] +
+               //      (file_header_block[12] << 8) +
+               //      (file_header_block[13] << 16) +
+               //      (file_header_block[14] << 24);
+               file_header_unp_size = file_header_block[14];
+               file_header_unp_size <<= 8; file_header_unp_size += file_header_block[13];
+               file_header_unp_size <<= 8; file_header_unp_size += file_header_block[12];
+               file_header_unp_size <<= 8; file_header_unp_size += file_header_block[11];
+
                if (verbose) {
+                       int i;
                        fprintf(stderr,
                                "! HEAD_SIZE: %d, PACK_SIZE: "LLu", UNP_SIZE: "LLu"\n",
                                file_header_head_size,
                                (unsigned long long)file_header_pack_size,
                                (unsigned long long)file_header_unp_size);
+                       fprintf(stderr, "! file_header_block: ");
+                       for (i = 0; i < 15; ++i)
+                               fprintf(stderr, " %02x", file_header_block[i]);
+                       fprintf(stderr, "\n");
                }
                /* calculate EXT_TIME size */
                ext_time_size = file_header_head_size - 32;
```

magnumripper:
I guess this is a problem of rar2john.c and not the base64 functions. We probably have a lot of such "problems" but RAR is one of the very few that is a problem IRL, if that. This should be a new issue.

magnumripper:
```c
		file_header_pack_size = file_header_block[7] +
			(file_header_block[8] << 8) +
			(file_header_block[9] << 16) +
			(file_header_block[10] << 24);
```
This must be incomplete. There's got to be more that 32 bits in the RAR header, for larger sizes.

magnumripper:
Here it is: http://www.forensicswiki.org/wiki/RAR#File_Header_.28File_in_Archive.29

There's another 32 bits later on **if** bit 0x100 in HEAD_FLAGS is set.

magnumripper:
I opened an issue for this, #2170

magnumripper:
I wonder why it ended up as 18446744070987741568 bytes though. Since we neglected to read the high bits, we should just end up with a small positive size.

magnumripper:
And BTW since size is less than 4 GB, it should actually work with just those 32 bits. We ARE using size_t so there should be no problem.

solardiz:
On Sat, Jul 09, 2016 at 08:24:02AM -0700, magnum wrote:
> I wonder why it ended up as 18446744070987741568 bytes though. Since we neglected to read the high bits, we should just end up with a small positive size.

This is FFFFFFFF5DC47980.  Sign extension occurred somewhere.

magnumripper:
Yeah the code @jfoug posted fixes it. I'm not quite sure why, perhaps writing it like eg. `(file_header_block[8] << 8ULL)` would also fix it.

jfoug:
Not sure where/why sign extension (it may be a bug or something else).  The conversion has size_t result, unsigned char for the buffer, signed int for the shift constants. But somewhere in that major conversion sign extension happens (the byte[7] which will be MSB) is sign extended.  But the way I 'see' the code, there should be no sign.  All I know is adding 1 byte at a time to the size_t and then shifting before next byte, made the problem go away.  But I still 'think' that must be some compiler error.   It may be how the size_t is usually a typedef or something  But it 'should' be unsigned, AFAIK on all systems.

I was also surprised at 4 bytes for rar length.  Do we account for the additional 32 bits of size?  Also, is it 56 bit number, or a full 64  bit number?

Issue #2140, opened 2016-05-18
"Log rule"


magnumripper:
Here's an idea I got. I'm just documenting it for now, I'm not sure I want more bloat.

Log verbosity for rules has decreased in Jumbo and this was more or less needed (the "all" ruleset is over seven million rules so was producing several hundreds of MB of log entries), but now you have no idea how far we've got among the rules. We could add some kind of directive that does nothing (it's not an actual rule) except outputs a log and/or screen message. Example:
```
# For Wordlist mode and very fast hashes
[List.Rules:Jumbo]
.log "Wordlist rules"
.include [List.Rules:Wordlist]
.log "Single rules"
.include [List.Rules:Single]
.log "Extra single rules"
.include [List.Rules:Extra]
.log "Case-toggling rules"
.include [List.Rules:NT]
```

Perhaps we could have `.logscreen "message"` for outputting to screen, `.log "message"` for outputting to log file, and something like `.logboth "message"` for doing both.

jfoug:
I really like this idea (yea, I am back from my 11k mile ride across western US, WOW, blast).

We could even add this as the default in john.conf, putting a .log as the first item right after each rule section. That way, running something like all, we would know what part of the 'all' rule we are at.

If we do this, then lets enhance this a bit.

The .log will be the pragma.

Then we can have other operators to the .log

so:

.log msg "Wordlist rules"
.log level 3

etc.  the msg could be 'default' so that 

.log "Wordlist Rules"     would be same as .log msg "Wordlist Rules"

But having an optional operator allow us to 'enhance' this feature.  

Yup, it is 'bloat', but as long as it does not impact the core running hot thread, who cares.  Just more bells (or whistles) and I think it is useful.

But instead of doing:
```
# For Wordlist mode and very fast hashes
[List.Rules:Jumbo]
.log "Wordlist rules"
.include [List.Rules:Wordlist]
.log "Single rules"
.include [List.Rules:Single]
.log "Extra single rules"
.include [List.Rules:Extra]
.log "Case-toggling rules"
.include [List.Rules:NT]
```

I would think that doing

# For Wordlist mode and very fast hashes
[List.Rules:Jumbo]
.log "Jumbo"
.include [List.Rules:Wordlist]
.include [List.Rules:Single]
.include [List.Rules:Extra]
.include [List.Rules:NT]

[List.Rules:Wordlist]
.log "Wordlist rules"
# whatever rules are in wordlist 

[List.Rules:Single]
[...]

PR #2171, opened 2016-07-11
added .log message directive code into rules scripts.  #2140

jfoug:
For #2140


jfoug:
@magnumripper  we can drop the branch that this one was stored on.  I do not know how to drop a branch

PR #2196, opened 2016-08-02
.pot chopped file valid() processing moved into common code in loader.c

jfoug:
NOT ready for merge.  Just putting out a PR so that this gets tracked, and so that we know when to dump the branch (at merge, or when we scrap this idea)


jfoug:
It also appears that CircleCI does not start a build for a branch until a PR is build ???

jfoug:
This one should be ready to go.

PR #2199, opened 2016-08-08
regex 1.4 integration, and full restore logic

jfoug:
Updated to function with rexgen 1.4 with full restore logic.


jfoug:
This PR is replacing PR  #2084

jfoug:
Bugs  #2084, #570 and #2082

PR #2214, opened 2016-08-15
Re-enable mscash2-opencl format on Travis CI.

claudioandre:
For testing purposes; not a real PR.

Please, merge only IF this is what you guys would like to see in.


magnumripper:
Looks good, I'm merging this.

magnumripper:
Hmmm I wonder if we could try the same with gpg-opencl now that we lowered the kernel duration limit.

Issue #2215, opened 2016-08-16
sha512crypt-opencl fails self tests on NVidia hardware


crioux:
SHA512 kernel is missing a few #pragma unpacks that are present on the SHA256 opencl kernel.
The result is my GTX 980 was failing self tests for sha512crypt-opencl. Adding the #pragma unpacks caused it to 'work for me'.

The patch for this is on my jtrdll fork, here:
https://github.com/L0phtCrack/jtrdll/commit/61644721965b81f836eb5b49f784194fe8cb9d6e

magnumripper:
You mean unrolls. This really shouldn't make any device fail - it should only affect speed. Likely an nvidia driver bug.

Anyway we might want to merge your patch but I'm not sure whether it was left out on purpose or not. @claudioandre this is yours, right?

magnumripper:
Oh and @crioux what nvidia driver version did you use when it failed?

crioux:
Yea, oops, I meant unroll. An it's Nvidia driver version 368.81

```
Platform #0 name: AMD Accelerated Parallel Processing, version: OpenCL 2.0 AMD-APP (2117.9)
    Device #0 (0) name:     AMD FX(tm)-8350 Eight-Core Processor
    Device vendor:          AuthenticAMD
    Device type:            CPU (LE)
    Device version:         OpenCL 1.2 AMD-APP (2117.9)
    Driver version:         2117.9 (sse2,avx,fma4)
    Native vector widths:   char 16, short 8, int 4, long 2
    Preferred vector width: char 16, short 8, int 4, long 2
    Global Memory:          31.0 GB
    Global Memory Cache:    16.0 KB
    Local Memory:           32.0 KB (Global)
    Max memory alloc. size: 7.0 GB
    Max clock (MHz):        4018
    Profiling timer res.:   69 ns
    Max Work Group Size:    1024
    Parallel compute cores: 8
    Speed index:            128576
    PCI device topology:

Platform #1 name: NVIDIA CUDA, version: OpenCL 1.2 CUDA 8.0.0
    Device #0 (1) name:     GeForce GTX 980
    Device vendor:          NVIDIA Corporation
    Device type:            GPU (LE)
    Device version:         OpenCL 1.2 CUDA
    Driver version:         368.81
    Native vector widths:   char 1, short 1, int 1, long 1
    Preferred vector width: char 1, short 1, int 1, long 1
    Global Memory:          4.0 GB
[...]

PR #2221, opened 2016-08-17
Fix sha512crypt-opencl failing self tests on NVIDIA OpenCL driver 361

claudioandre:
Is this a reasonable solution for #2215 ?

It works on super (which uses CUDA 8). I bet no new problems introduced.
****

In this change:
1. if there is a known problem, we do something.
2. otherwise, let the compiler do its best.

### DIFF
--- a/src/opencl/cryptsha512_kernel_GPU.cl
+++ b/src/opencl/cryptsha512_kernel_GPU.cl
@@ -160,6 +160,8 @@ inline void sha512_block(sha512_ctx * ctx) {
     #pragma unroll 2
 #elif defined(NVIDIA_STUPID_BUG_1) && DEV_VER_MAJOR == 361 && DEV_VER_MINOR == 28
     #pragma unroll 16
+#else
+	#pragma unroll
 #endif
     for (uint i = 16U; i < 80U; i++) {
 	w[i & 15] = w[(i - 16) & 15] + w[(i - 7) & 15] + sigma1(w[(i - 2) & 15]) + sigma0(w[(i - 15) & 15]);
@@ -467,6 +469,7 @@ void kernel_preprocess(
 	uint32_t total = 0;
 	uint32_t j = generator_index[i];
 
+	#pragma unroll
 	for (uint32_t k = 0; k < 8; k++)
 	   work_memory[OFFSET(i, k)] = 0;
 
--- a/src/opencl/cryptsha512_kernel_GPU.cl
+++ b/src/opencl/cryptsha512_kernel_GPU.cl
@@ -38,6 +38,32 @@
     #define VECTOR_USAGE    1
 #endif
 
+#if amd_vliw5(DEVICE_INFO) || amd_vliw4(DEVICE_INFO)
+/* 
+   Needed (at least) in 14.9 and 15.7.
+   kernel prepare fails if no unroll hint is offered.
+   - the workaround is not needed on GCN (tested on Tahiti, Bonaire and Capeverde)
+   - amd_vliw4() is a guess, no hardware available to test.
+   - can't remove the [unroll] (at least) on Radeon HD 6770.
+   - R9 290X w/ 1800.11 works fine with or without explicit unroll.
+*/
+#define AMD_UNROLL_BUG_1    1
+#endif
+
+#if gpu_nvidia(DEVICE_INFO) && DEV_VER_MAJOR > 352
+/* 
+   Needed for OpenCL NVIDIA driver version 361 (CUDA 7.5+).
+   kernel prepare fails if no unroll hint is offered.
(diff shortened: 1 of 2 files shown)


magnumripper:
IMHO it's a bit messy as of now (as in hard to follow). For instance, NVIDIA_STUPID_BUG_1 is **always** defined for any nvidia. Then if it's defined **and** we are within a version span, we work around bugs. Doesn't really make sense.

I think something along the lines of this would be easier to understand/maintain:

opencl_misc.h
```c
#if gpu_nvidia(DEVICE_INFO) && DEV_VER_MAJOR >= 361
/* some comment on this particular bug */
#define NVIDIA_STUPID_BUG_1    1
#endif

#if gpu_nvidia(DEVICE_INFO) && DEV_VER_MAJOR >= 310 && DEV_VER_MAJOR <= 313
/* some comment on this particular bug */
#define NVIDIA_STUPID_BUG_2    1
#endif

(...)

#if gpu_amd(DEVICE_INFO) && DEV_VER_MAJOR >= this && DEV_VER_MAJOR <= that
/* some comment on this particular bug */
#define AMD_STUPID_BUG_1    1
#endif

(...)
```

some_kernels.cl
```c
(...)
#if AMD_STUPID_BUG_1
      #pragma unroll 2 /* or whatever workaround */
#elif NVIDIA_STUPID_BUG_1
      #pragma unroll 16 /* or whatever workaround */
#endif
(...)
```

magnumripper:
BTW determining a correct or non-ambiguous "CUDA version" from OpenCL output can be tricky.

The stuff @solardiz installed on super was called "CUDA 7.5". According to --list=cuda-devices, it's CUDA runtime 7.5 and CUDA driver 8.0 and according to --list=opencl-devices, it's "CUDA 8.0.13".

claudioandre:
> IMHO it's a bit messy as of now (as in hard to follow). For instance, NVIDIA_STUPID_BUG_1 is always defined for any nvidia. Then if it's defined and we are within a version span, we work around bugs. Doesn't really make sense.

I can see your point, but there are some things I though when I first did this:
* avoid messing with shared code and opencl_misc.h is used a lot.
* keep driver versions inside the .cl file, easy to find and change - probably only to me.
* allow an easy way to solve a problem if one happens (just change the .cl file in the right place). Consider the situation when an user opened 2215.
* I called them bugs, but in fact I was only documenting (somehow) my findings. Seems, nobody else faced problems

But it is fair, I can start to document my findings in a better way.

claudioandre:
> according to --list=opencl-devices, it's "CUDA 8.0.13"

Strange (since I never did myself anything NVIDIA related I will say nothing). 
* But there is a printf inside NVIDIA software outputing this.

magnumripper:
> avoid messing with shared code and opencl_misc.h is used a lot.

Fair enough, if you believe a bug only affects your code you can keep it all within your kernel. Still, it could be something like:
```c
#if gpu_nvidia(DEVICE_INFO) && DEV_VER_MAJOR >= 361
/* some comment on this particular bug, what device, issue # or whatever */
#define NVIDIA_UNROLL_BUG    1
#endif

#if gpu_amd(DEVICE_INFO) && DEV_VER_MAJOR >= this && DEV_VER_MAJOR <= that
/* some comment on this particular bug, what device, issue # or whatever */
#define AMD_UNROLL_BUG    1
#endif

(...)
#if AMD_UNROLL_BUG
      #pragma unroll 2
#elif NVIDIA_UNROLL_BUG
      #pragma unroll 16
#endif
```

This applies to myself too; my documentation of bugs has varied between poor and good.

magnumripper:
Some more data points:
- Titan X w/ 361.42 also fails without the explicit unroll.
- Titan X w/ 352.63 works fine with or without explicit unroll.
- R9 290X w/ 1800.11 works fine with or without explicit unroll.

claudioandre:
> Titan X w/ 361.42 also fails without the explicit unroll.
Titan X w/ 352.63 works fine with or without explicit unroll.

As it used to work on super in the past.

> R9 290X w/ 1800.11 works fine with or without explicit unroll.

Only saw fails on VLIW5 (remember, I tested on well boards too).

> so it's better to over-use it than not.

Do you mean enable the workaround always?

magnumripper:
>> so it's better to over-use it than not.

> Do you mean enable the workaround always?

Perhaps not, but we should err to the safe side so we rather over-use it than miss some, eg:
```c
#if gpu_nvidia(DEVICE_INFO) && DEV_VER_MAJOR > 352
      // So far any version above 352 we've tested needed this to work
      #pragma unroll 16
#endif
```

magnumripper:
Looks fine to me now. We may need to improve it over time of course, as vendors develop more bugs...

PR #2235, opened 2016-08-25
Publishing from Travis to Ubuntu Snap Store

claudioandre:
Please, I need to know if this patch is useful and could be merged. If so:

WIP
- I have to share a secret with travis:magnumripper/JohnTheRipper (but I can't save it using cli - I'm not magnum); I tried to submit it via web page; _Check if it is working_
- Easy to see that the 'ok' or 'fail' takes the same amount of time (not sure if anyone is online now to confirm); I did the check.
- These binaries can also became a ZIP to be used in Windows with  WSL (at some point in the 'near' future).
- Maybe something else 'secret' related.

It already works when the commit happens in 'my' fork.

---

Publishing bleeding-edge JtR to the snap store from Travis

Allow snap users to install, test and use the latest and hot bleeding-jumbo from git. The snap build process is not going to make Travis slower: build completeness will finish as soon gcc and clang have completed (as it does now).

If the user wants something hot:
$ snap refresh --channel=edge john-the-ripper

Unhappy? No problems, jump back when he/she can no longer take the heat:
$ snap refresh --channel=stable john-the-ripper

### DIFF
--- a/src/cracker.c
+++ b/src/cracker.c
@@ -950,12 +950,13 @@ static int crk_salt_loop(void)
 	}
 	do {
 		crk_methods.set_salt(salt->salt);
-		status.resume_salt_md5 = salt->salt_md5;
+		status.resume_salt_md5 = (crk_db->salt_count > 1) ?
+			salt->salt_md5 : NULL;
 		if ((done = crk_password_loop(salt)))
 			break;
 	} while ((salt = salt->next));
-	if (!salt || salt->count < 2)
-		status.resume_salt_md5 = 0;
+	if (!salt || crk_db->salt_count < 2)
+		status.resume_salt_md5 = NULL;
 
 	if (done >= 0) {
 #if !HAVE_OPENCL


magnumripper:
I'm not too happy with all the files in base dir, especially a Makefile that will confuse users. Can we create a new directory 'Travis' or '.travis' and have (some of) them there?

claudioandre:
Done. What about the other concerns?

magnumripper:
That was my only concern. Did you manage to share a secret? Or do we not quite know until it's actually merged?

claudioandre:
> That was my only concern. 

After merge, any commit can be at users hands faster than before; I mean a 'Joe User'. There are commits that are not meant to be published, or shouldn't be.

> Did you manage to share a secret?

No, it seems that I can't encrypt using your public key (I used something, it is not mine but it is not accepted inside your magnumripper).

claudioandre:
Just to document, I changed things a little:
- https://github.com/claudioandre/JohnTheRipper/commits/daily

magnumripper:
If there's anything I can do to fix the shared secret issue, please let me know. I don't really have any idea what is needed.

claudioandre:
I guess you have to run the 'travis --set-env' yourself - after I share the secret with you. 
****

Well, in order to test, I have been posting x64, armhf, arm64 and ppc to the edge channel (once a day, in fact). x64 with all fancy stuff using Travis. Others using canonical servers. BTW: I don't think I can do it all using only Travis CI.

That said:
- without the secret, my 'daily' branch behaves in Travis as it does now (build gcc, clang and test). Only if it knows the secret it does something else. I mean, a merge is already possible - after an approval.
- I'm not confident we can build edge packages for all archs without human intervention.
- for x64, any commit in my branch 'daily' will upload to the Store. So, a full automation is already in place.

claudioandre:
I' going to use the build bots listed above:
- not 100% automated; A request via web is necessary.
- packages upload from build bots only [no user mistakes, sysadmins around, ...].
- the package announce itself using a commit hash.
- I know some people have installed the software, but I have no idea what real users are expecting:
  - daily builds? build more than once a day (why?)? Weekly?

So, I would like to hear from you guys:
- how often should I publish? What should be in stable channel? What about edge? 

****
Since .travis.yml is upstream, a merge is still desirable. Other files can be safely ignored.

claudioandre:
It is not impossible to build to IBM System z (s390x). But, I guess nobody is interested.

PR #2247, opened 2016-08-29
logger.c log_file_init check permissions so as to not perform unnecessary chmod

walkerk1980:
Only chmod file if permissions aren't already correct as chmoding will unnecessarily fail if another user owns john.pot, even if john.pot is readable and writeable.

### DIFF
--- a/src/logger.c
+++ b/src/logger.c
@@ -95,9 +95,12 @@ static void log_file_init(struct log_file *f, char *name, int size)
 	if (f == &log && (options.flags & FLG_NOLOG)) return;
 	f->name = name;
 
-	if (chmod(path_expand(name), S_IRUSR | S_IWUSR))
-	if (errno != ENOENT)
-		pexit("chmod: %s", path_expand(name));
+        if (!(access(path_expand(name), R_OK) == 0) || !(access(path_expand(name), W_OK) == 0))
+        {
+                if (chmod(path_expand(name), S_IRUSR | S_IWUSR))
+                if (errno != ENOENT)
+                        pexit("chmod: %s", path_expand(name));
+        }
 
 	if ((f->fd = open(path_expand(name),
 	    O_WRONLY | O_CREAT | O_APPEND, S_IRUSR | S_IWUSR)) < 0)


magnumripper:
@solardiz please comment. I'm not sure you'll like this patch. Personally I'm pretty much OK with it.

We are normally chmod'ing to 600 so for this to happen, that other user must have set 660 or "worse". With this patch, we would not try to reset it back.

Also, from access(2) man page
```
CAVEAT
     access() is a potential security hole and should never be used.
```

solardiz:
I think we should NAK this patch, at least unless and until its rationale is explained better (and then we'd need to rework it to match the actual rationale). My impression is that @walkerk1980 assumed the chmod() was to grant permissions, and thus would be unnecessary if the permissions already allow read and write, but that wasn't the actual primary intent. Rather, the primary intent was to restrict permissions, since the default umask is often too relaxed for working with sensitive data like that.

jfoug:
What about in a system wide build??

frank-dittrich:
> What about in a system wide build??

@jfoug I doubt that all users on a system should share the same .pot file.

jfoug:
How would they have separate .pot files on a john core ?   I thought only jumbo has the -pot= command line switch.

walkerk1980:
@solardiz I had assumed that sorry. I'm new to John and to C and had not noticed that Jumbo includes --pot as well.  The situation I have is a team that is insistant on having a centralized John binary on a multi-gpu-rig and multiple non-sudo users sharing the same pot file as to not duplicate work on the same pentest and to build on passwords from previous pentests. Would it be more appropriate if there were a switch like --shared-pot or --group-pot which would change the default desired perms are for john.pot? If this is in no way desired functionality please let me know because I can always work around this with a script to run john instead that takes care of the perms. 

I can understand why by default a user would want 600 when generating passwords and default umask is more permissive. Although if a user runs the John binary without specifying --session or --pot and then another user does the same we fail out on chmod. I do not understand trying to chmod a file that belongs to somebody else that will always fail anyway, warning instead of exiting would be good enough to make my use case possible with users manually chmod 660 the pot file.

Please go ahead and close this pull request, but if someone could point me in the right direction as to creating a more appropriate patch for a team of users in the same group I would appreciate it greatly!

frank-dittrich:
@walkerk1980 

```
(bleeding-jumbo)run $ git grep -A 8 "extra pot file"
john.conf:# This allows you to read extra pot files when loading hashes. Nothing will
john.conf-# ever be written to these files, they are just read. Any directory in this
john.conf-# list will be traversed and files in it with an extension of .pot will be
john.conf-# read. However there will NOT be any recursion down further directory levels.
john.conf-# Any entries that don't exist will be silently ignored.
john.conf-[List.Extra:Potfiles]
john.conf-#somefile.pot
john.conf-#somedirectory
john.conf-#$JOHN/my.pot
```

So, a user can add corresponding lines into his john-local.conf.

May be we need to document this in a file doc/README.teamwork or similar?

walkerk1980:
@frank-dittrich Sharing pot files would still need users to manually chmod 660 their pot files after each John run though since currently working shared pot files get's chmod 600 each run correct?

frank-dittrich:
@walkerk1980 Yes you need some additional scripts.

each user: regularly copy either ether complete pot file or just the diff under a unique to a common directory, making sure not to overwrite existing files, use chmod to make sure the file is readable by others.

some central agent:
process all the individual pot files uploaded by others:
-did someone accidentally upload a pot file containing hashes which were not supposed to be cracked during the pen test? 
-do some consistency checks (different passwords for same hash?)
-different representations of the same hash (due to different john versions or other tools being used etc.) 
-remove duplicates, to reduce load time

solardiz:
I don't mind us adding a PotMode config file setting, 0600 by default.  Process it as a base 8 number by default, such as passing 8 as the base argument to strtoul().

walkerk1980:
@frank-dittrich This is what I was using... which works but could get messy if multiple users call script at once.

Root cron job:
/bin/chmod 0660 /usr/local/bin/john/*.log
/bin/chmod 0660 /usr/local/bin/john/*.pot    
sleep 10 # repeat x 5

Replace John binary with script that calls John Binary:
mv ./john.pot ./john.pot.bak
cat ./john.pot.bak >> ./john.pot
./john_binary $1 $2 $3 $4 $5 $6 $7 $8 $9

@solardiz That sounds great, I'll make that patch instead.  

@everyone Thanks for your help.

PR #2263, opened 2016-09-08
Added support for HDAA as defined in RFC 2069.

alehaa:
The current implementation for HDAA only covered RFC 2617, but was not able to
test digest auth responses in the old scheme.

### DIFF
--- a/doc/HDAA_README
+++ b/doc/HDAA_README
@@ -36,3 +36,15 @@ Loaded 2 password hashes with 2 different salts (HTTP Digest access authenticati
 kikou            (moi)
 nocode           (user)
 guesses: 2  time: 0:00:01:27 (3)  c/s: 670223  trying: nocode
+
+
+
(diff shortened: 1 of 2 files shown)


magnumripper:
Excellent, thank you!

jfoug:
Do we have a hdaa2john type converter ?

magnumripper:
Converter from what? I guess we should add support for it in pcap2john.

jfoug:
How are these hashes 'built' ???  I would expect it to be in pcap2john.py, but do not see anything.  The only place I find 'response' that builds a hash, was in pass_gen.  That is why my question.

magnumripper:
For one-time uses (in general), I often pick stuff manually from wireshark, for example. That would be awful in larger scale though...

Perhaps the http dissector in Ettercap already outputs JtR format, I have no idea.

PR #2269, opened 2016-09-12
Keep default Log file permissions, but add option to set in john.conf.

walkerk1980:
Please see previous pull request for use case https://github.com/magnumripper/JohnTheRipper/pull/2247

Added config file parameters LogFilePermissions, PotFilePermissions, IgnoreLogChmodErrors.

Small issue: open() on line 102 of logger.c ignores perms_t on O_CREAT and perms are still set on create to the less permissive defaults.  Perms don't actually get set to more permissive perms until second run of John when chmod() is called in log_file_init().  I'm not that familiar with C or the John code, so it may be a umask I'm not seeing somewhere.  Besides this issue, feature works properly in my testing so far.


magnumripper:
We do not want your local merges upstream. Please learn git, rebase and re-submit (push force).

magnumripper:
It looks like cherry-picking 2e9093f would suffice.

walkerk1980:
Sorry, will resubmit.

walkerk1980:
Resubmitted as https://github.com/magnumripper/JohnTheRipper/pull/2270

PR #2270, opened 2016-09-12
Added LogFilePermissions, PotFilePermissions, IgnoreLogChmodErrors

walkerk1980:
Please see previous pull request for use case #2247

Added config file parameters LogFilePermissions, PotFilePermissions, IgnoreLogChmodErrors.

Small issue: open() on line 102 of logger.c ignores perms_t on O_CREAT and perms are still set on create to the less permissive defaults. Perms don't actually get set to more permissive perms until second run of John when chmod() is called in log_file_init(). I'm not that familiar with C or the John code, so it may be a umask I'm not seeing somewhere. 

Besides the previously mentioned issue, feature works properly in my testing so far.

### DIFF
--- a/src/logger.c
+++ b/src/logger.c
@@ -68,9 +68,6 @@ static char *LogDateStderrFormat;
 static int LogDateFormatUTC=0;
 static char *log_perms;
 static char *pot_perms;
-static char *p_u;
-static char *p_ug;
-static char *p_ugo;
 mode_t perms_t;
 
 /*
@@ -96,17 +93,8 @@ static struct log_file pot = {NULL, NULL, NULL, 0, -1};
 
 static int in_logger = 0;
 
-static void log_file_init(struct log_file *f, char *name, char *perms, int size)
+static void log_file_chmod(char *name, mode_t perms_t)
 {
-	p_u = "0600";
-	p_ug = "0660";
-	p_ugo = "0666";
-	if (!((strcmp(perms, p_u) == 0) || (strcmp(perms, p_ug) == 0) || (strcmp(perms, p_ugo) == 0)))
-		pexit("PotFilePerms or LogFilePerms %s invalid, valid values: %s, %s, %s\n\n", perms, p_u, p_ug, p_ugo);
-	perms_t = strtoul(perms, NULL, 8);
-	if (f == &log && (options.flags & FLG_NOLOG)) return;
-	f->name = name;
-
 	if (chmod(path_expand(name), perms_t))
 	{
 		if (cfg_get_bool(SECTION_OPTIONS, NULL, "IgnoreLogChmodErrors", 0))
@@ -119,9 +107,33 @@ static void log_file_init(struct log_file *f, char *name, char *perms, int size)
 				pexit("chmod: %s", path_expand(name));
 	}
 	
+}
+
+static void log_file_init(struct log_file *f, char *name, char *perms, int size)
+{
+	perms_t = strtoul(perms, NULL, 8);
+	if ((perms_t & 0x49) || !(perms_t & 0x92))
+	{
+		fprintf(stderr, "PotFilePerms or LogFilePerms %s invalid\n", perms);
+		error();
+	}
+	
+	if (f == &log && (options.flags & FLG_NOLOG)) return;
+	f->name = name;
+
+	log_file_chmod(name, perms_t);
+	
+#ifndef _MSC_VER
+    umask(000);
+#endif
+
 	if ((f->fd = open(path_expand(name),
 	    O_WRONLY | O_CREAT | O_APPEND, perms_t)) < 0)
 		pexit("open: %s", path_expand(name));
+
+#ifndef _MSC_VER
+    umask(077);
+#endif
 
 	/*
 	 * plain will now always be < LINE_BUFFER_SIZE. We add some extra bytes
--- a/src/logger.c
+++ b/src/logger.c
@@ -66,6 +66,12 @@ static int cfg_showcand;
 static char *LogDateFormat;
 static char *LogDateStderrFormat;
 static int LogDateFormatUTC=0;
+static char *log_perms;
+static char *pot_perms;
+static char *p_u;
+static char *p_ug;
+static char *p_ugo;
+mode_t perms_t;
 
 /*
  * Note: the file buffer is allocated as (size + LINE_BUFFER_SIZE) bytes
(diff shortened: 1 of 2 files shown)


magnumripper:
Looks good on first sight. I will test it a bit and see if I can nail the mentioned problem before merging. Thanks!

magnumripper:
BTW I see it currently only allows 600, 660 or 666. I can imagine someone wanting eg. 640 or 664 and other combinations (eg. allowing group `john -show` but not to write to it).

magnumripper:
from man open(2)
```
     The oflag argument may indicate that the file is to be created if it does not exist
     (by specifying the O_CREAT flag).  In this case, open() and openat() require an addi-
     tional argument mode_t mode; the file is created with mode mode as described in
     chmod(2) and modified by the process' umask value (see umask(2)).
```

src/logger.c
```diff
@@ -90,17 +96,31 @@ static struct log_file pot = {NULL, NULL, NULL, 0, -1};
...
 
-static void log_file_init(struct log_file *f, char *name, int size)
+static void log_file_init(struct log_file *f, char *name, char *perms, int size)
 {
+	p_u = "0600";
+	p_ug = "0660";
+	p_ugo = "0666";
+	if (!((strcmp(perms, p_u) == 0) || (strcmp(perms, p_ug) == 0) || (strcmp(perms, p_ugo) == 0)))
+		pexit("PotFilePerms or LogFilePerms %s invalid, valid values: %s, %s, %s\n\n", perms, p_u, p_ug, p_ugo);
+	perms_t = strtoul(perms, NULL, 8);
```
jfoug: Why only rw------- or rw-rw---- or rw-rw-rw-  What about rw-r--r-- or  rw-rw-r--  All are equally valid (for differing reasons)

walkerk1980:
I just didn't want to make the code too long, I have nothing against the other perms. As far as testing after conversion, I was having strange results when converting if a user didn't include the leading 0 in the permissions string, it was a couple weeks ago when I was working on it.  I was trying to make sure the string coming in could always be used, but there may be a better way than the way I went.

src/logger.c
walkerk1980: I just didn't want to make the code too long, I have nothing against the other perms. As far as testing after conversion, I was having strange results when converting if a user didn't include the leading 0 in the permissions string, it was a couple weeks ago when I was working on it.  I was trying to make sure the string coming in could always be used, but there may be a better way than the way I went.

jfoug: This if might be better.

int bad = 0;
if (perms_t & 0x49)
   bad = 1;  //  an execute bit is set
if (!(perms_t & 0x92))
   bad = 1;  // no write bit set.
if (bad) {
   // perform bad exit
}

This would make sure that there is no execute bit  (0x49 is 0b01001001  (i.e exe bits), and 0x92 is 0b010010010 (write bits).

walkerk1980:
@magnumripper "the file is created with mode mode as described in chmod(2) and modified by the process' umask value (see umask(2))."

So, if I change umask 077 to 000 in john.c, then open() on line 102 of logger.c works as expected on file creation.  But this seems undesirable as I have no idea what other files that effects.  Do you think I should I just force a chmod after file creation instead?

walkerk1980:
Making chmod a method and calling before and after file creation works.
```
static void log_file_chmod(char *name, mode_t perms_t)
{     
        if (chmod(path_expand(name), perms_t))
        {      
                if (cfg_get_bool(SECTION_OPTIONS, NULL, "IgnoreLogChmodErrors", 0))
                {       
                        if (errno == EPERM)
                                fprintf(stdout, "chmod: %s failed, but IgnoreChmodError is set, moving on..\n\n", path_expand(name));
                }     
                else 
                        if (errno != ENOENT)
                                pexit("chmod: %s", path_expand(name));
        }
        
}
```
```
       log_file_chmod(name,perms_t);
        
        if ((f->fd = open(path_expand(name),
            O_WRONLY | O_CREAT | O_APPEND, perms_t)) < 0)
                pexit("open: %s", path_expand(name));

        log_file_chmod(name,perms_t);
```

magnumripper:
Let's ponder the umask issue for a while. I'm inclined to actually obey the umask (with a helpful comment in that john.conf section). If user does't want it she should change umask prior to running john. @solardiz what do you think?

magnumripper:
Or maybe not - if user explicitly said eg. 0660 in john.conf, that could override umask. I haven't decided yet but that does make sense too.

walkerk1980:
@magnumripper So, I could either call log_file_chmod() twice like above, which fixes the umask issue, or I can add another config file Parameter for ProcessUmask, and specify in comments on the other parameters that ProcessUmask must be set properly to allow the other two parameters to work on first run.

Current Parameters:
``` #LogFilePermissions = 0600,
 #PotFilePermissions = 0600,
 #IgnoreLogChmodErrors = N```

New Parameter:
```ProcessUmask = 700```

File: src/logger.c (hunk 1 above)
Comment (walkerk1980) [inline src/logger.c]: @jfoug Thanks! I will try testing what happens after conversion.  The issue I was having is that if a user specified for example 666 instead of 0666 it would set sticky bit and be like 400 for the rest of the perms or something weird like that.  I might have been passing the param as an int at the time and not a ulong, which was dropping my preceding zero no matter what. I'll have to test and see.  Besides that, much simpler thank you.

**Comment by magnumripper** [comment]
When I wrote:
> I'm inclined to actually obey the umask

I didn't realize john.c was dropping umask to 077. So the user's original umask is ignored anyway.

I see no need for yet another config parameter. Your chmod function might be ok. Or even simpler, just drop that umask temporarily:
```c
	if (chmod(path_expand(name), perms_t) && errno != ENOENT)
	{
		if (errno == EPERM && cfg_get_bool(SECTION_OPTIONS, NULL, "IgnoreLogChmodErrors", 0))
			fprintf(stdout, "chmod: %s failed, but IgnoreChmodError is set, moving on..\n\n", path_expand(name));
		else
			pexit("chmod: %s", path_expand(name));
	}

#ifndef _MSC_VER
	umask(000);
#endif

	if ((f->fd = open(path_expand(name),
	    O_WRONLY | O_CREAT | O_APPEND, perms_t)) < 0)
		pexit("open: %s", path_expand(name));

#ifndef _MSC_VER
	umask(077);
#endif
```
I'm just dry-running code though, I didn't test this.

src/logger.c
```diff
@@ -90,17 +96,31 @@ static struct log_file pot = {NULL, NULL, NULL, 0, -1};
...
 static int in_logger = 0;
 
-static void log_file_init(struct log_file *f, char *name, int size)
+static void log_file_init(struct log_file *f, char *name, char *perms, int size)
 {
+	p_u = "0600";
+	p_ug = "0660";
+	p_ugo = "0666";
+	if (!((strcmp(perms, p_u) == 0) || (strcmp(perms, p_ug) == 0) || (strcmp(perms, p_ugo) == 0)))
+		pexit("PotFilePerms or LogFilePerms %s invalid, valid values: %s, %s, %s\n\n", perms, p_u, p_ug, p_ugo);
```
magnumripper: Please note that `pexit()` should only be used for things that set errno, eg. `chmod()`, but this is not the case here. You should change it to
```c
	if (<insert Jim's clever tests here>) {
		fprintf(stderr, "PotFilePerms or LogFilePerms %s invalid\n", perms);
		error();
	}
```

walkerk1980: @magnumripper Thank you, will do. I will submit a new commit with all the changes.

walkerk1980: @magnumripper This works perfectly, thanks!

```
#ifndef _MSC_VER
    umask(000);
#endif

    if ((f->fd = open(path_expand(name),
        O_WRONLY | O_CREAT | O_APPEND, perms_t)) < 0)
        pexit("open: %s", path_expand(name));

#ifndef _MSC_VER
    umask(077);
#endif

```

magnumripper:
Thanks!

walkerk1980:
@magnumripper @jfoug Thanks guys! This will be a huge help to me in supporting my team's use case.

magnumripper:
@walkerk1980 please note that 304da8803 refactored `IgnoreLogChmodErrors` to `IgnoreChmodErrors` (because it affects pot too). Other than that, I tweaked the permitted perms logic a bit after finding out a mode like 0066 would render unreadable for owner...

walkerk1980:
@magnumripper My thoughts on IgnoreLogChmodErrors naming was because it's directed at the chmod() in log_file_init(). I wasn't thinking about the end user perspective, that makes sense.

PR #2280, opened 2016-09-19
path_init() must be called prior to CPU fallback mechanism

### LINKED ISSUE #2277: *2john tools fail in CPU fallback due to path_init() is not been called soon enough
Isn't a patch like this mandatory?

```
index f230335..6c82b10 100644
--- a/src/john.c
+++ b/src/john.c
@@ -1408,7 +1408,6 @@ static void john_init(char *name, int argc, char **argv)
        if (make_check)
                argv[1] = "--test=0";
 
-       path_init(argv);
        CPU_detect_or_fallback(argv, make_check);
 
 #if HAVE_MPI
@@ -1891,6 +1890,7 @@ int main(int argc, char **argv)
 
        /* put the crc table init here, so that tables are fully setup for any ancillary program */
        CRC32_Init_tab();
+       path_init(argv);
 
        if (!strcmp(name, "unshadow")) {
                CPU_detect_or_fallback(argv, 0);
```

claudioandre:
Closes #2277

- [x] CI
- [x] Testing

### DIFF
--- a/src/john.c
+++ b/src/john.c
@@ -1408,7 +1408,6 @@ static void john_init(char *name, int argc, char **argv)
 	if (make_check)
 		argv[1] = "--test=0";
 
-	path_init(argv);
 	CPU_detect_or_fallback(argv, make_check);
 
 #if HAVE_MPI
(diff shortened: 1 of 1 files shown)


claudioandre:
I run TS on super (first time in a long time).

I can see some errors. If nobody says this is expected, I will open a ticket:
****
Some tests had Errors. Performed 687 tests.  3 errors  1 errors reprocessing the .POT files

magnumripper:
Not sure what the TS errors are but they're not related to this bugfix anyway.

Issue #1976, opened 2016-01-03
--regex='\0' fails for passwords with non-ascii characters


frank-dittrich:
latest bleeding-jumbo, latest test suite version, latest rexgen master commit with this patch to drop librexgen debug output:
```
diff --git a/src/librexgen/string/simplestring.cpp b/src/librexgen/string/simplestring.cpp
index 9c52188..0e1d5de 100644
--- a/src/librexgen/string/simplestring.cpp
+++ b/src/librexgen/string/simplestring.cpp
@@ -45,7 +45,7 @@ void SimpleString::toupper(unsigned int n) {
 }
 
 void SimpleString::append(const char* ch, size_t ch_len) {
-  fprintf(stderr, "APPEND\n");
+//  fprintf(stderr, "APPEND\n");
        /* ensure that buffer is not full */
        if (length > SIMPLESTRING_MAXLEN) {
                return;
@@ -60,7 +60,7 @@ void SimpleString::append(const char* ch, size_t ch_len) {
        for (size_t idx=0; idx<ch_len; ++idx) {
                characters[length++] = ch[idx];
 
-    fprintf(stderr, "%04x\n", characters[length-1].codepoint);
+//    fprintf(stderr, "%04x\n", characters[length-1].codepoint);
        }
 }
```

```
(master)test $ ./jtrts.pl -q -passthru="--regex='\0'"
-------------------------------------------------------------------------------
- JtR-TestSuite (jtrts). Version 1.13, Dec 21, 2014.  By, Jim Fougeron & others
- Testing:  John the Ripper 1.8.0.6-jumbo-1-1814-gca11872 [linux-gnu 64-bit AVX-ac]
--------------------------------------------------------------------------------
form=dynamic_0                    guesses: 1195 -show=1197 0:00:00:00 DONE : Expected count(s) (1500)(-show1502)  [!!!FAILED1!!!  ret_val=0]
form=dynamic_0                    guesses:  876 -show= 876 0:00:00:00 DONE : Expected count(s) (1500)(920)  [!!!FAILED1!!!  ret_val=0]
form=dynamic_0                    guesses: 1436 -show=1436 0:00:00:00 DONE : Expected count(s) (1500)  [!!!FAILED1!!!  ret_val=0]
form=dynamic_0-raw                guesses: 1195 -show=1195 0:00:00:00 DONE : Expected count(s) (1500)  [!!!FAILED1!!!  ret_val=0]
[...]

PR #2294, opened 2016-09-29
Fix for a warning seen when compiling against regex master

### LINKED ISSUE #2293: Update regex refs from upstream
- I automated regex building inside CI (during SNAP package creation). 
- I already created an issue in regex upstream asking for a release / git tag for some kind of stable version (so I can use an stable version, not head, during build).
- meanwhile, I guess JtR needs to update its regex references.

```
regex.c: In function ‘do_regex_hybrid_crack’:
regex.c:218:33: warning: passing argument 2 of ‘c_regex_cb’ from incompatible pointer type [-Wincompatible-pointer-types]
   regex_ptr = c_regex_cb(regex, callback);
                                 ^
In file included from regex.h:35:0,
                 from regex.c:12:
/home/claudio/fontes/packages/john-the-ripper/stage/include/librexgen/c/librexgen.h:40:13: note: expected ‘callback_fp {aka long unsigned int (*)(int *, const long unsigned int)}’ but argument is of type ‘size_t (*)(char *, const size_t) {aka long unsigned int (*)(char *, const long unsigned int)}’
 c_regex_ptr c_regex_cb(
             ^
regex.c: In function ‘do_regex_crack’:
regex.c:328:32: warning: passing argument 2 of ‘c_regex_cb’ from incompatible pointer type [-Wincompatible-pointer-types]
  regex_ptr = c_regex_cb(regex, callback);
                                ^
In file included from regex.h:35:0,
                 from regex.c:12:
/home/claudio/fontes/packages/john-the-ripper/stage/include/librexgen/c/librexgen.h:40:13: note: expected ‘callback_fp {aka long unsigned int (*)(int *, const long unsigned int)}’ but argument is of type ‘size_t (*)(char *, const size_t) {aka long unsigned int (*)(char *, const long unsigned int)}’
 c_regex_ptr c_regex_cb(
             ^
ar: creating aes.a
```

claudioandre:
- Fixes and closes #2293

### DIFF
--- a/src/regex.c
+++ b/src/regex.c
@@ -138,7 +138,7 @@ static void rexgen_setlocale()
 // Would be nice to have SOME way to be thread safe!!!
 static char BaseWord[1024];
 
-size_t callback(char* dst, const size_t buffer_size)
+size_t callback(wchar_t* dst, const size_t buffer_size)
 {
 	int len;
 


magnumripper:
Dang, is there still no way to detect version? We should ideally have this ifdef'ed depending on version.

claudioandre:
Just a side note: regex created a release that JtR should follow or recommend - 1.4.0.

> is there still no way to detect version?

I would say 'yes'. But I saw no other releases, only 1.4.0. I mean, I know nothing about old version numbers.

claudioandre:
BTW: 1.4 is required in JtR. So, I understand what you meant, but on a second thought:
- for JtR HEAD, there is only one regex version that is valid >= 1.4.0. And this patch is required.

magnumripper:
Aha so this prototype discrepancy was there ever since 1.4? I never saw any compiler complaint but let's just merge this then!

claudioandre:
Since 12 Feb

claudioandre:
BTW: is anyone using and testing the multi-byte chars usage? I tested it only for pure ASCII.

Not sure if it can handle all encoding witchcraft. See also: #1976

magnumripper:
Apparently I didn't upgrade my librexgen for quite a while.

However, I now get this
```
regex.c: In function 'callback':
regex.c:148:19: warning: passing argument 1 of 'strnzcpyn' from incompatible pointer type [-Wincompatible-pointer-types]
  len =  strnzcpyn(dst, BaseWord, 1024);
                   ^~~
In file included from formats.h:16:0,
                 from loader.h:18,
                 from regex.h:47,
                 from regex.c:12:
misc.h:118:12: note: expected 'char *' but argument is of type 'wchar_t * {aka int *}'
 extern int strnzcpyn(char *dst, const char *src, int size);
            ^~~~~~~~~
```
This is just a matter of casting explicitly. I'll fix that. I'll also have a look at current state of #1976 and other things (we might want/need to add conversion stages before/after librexgen calls, when using a non-UTF8 encoding)

magnumripper:
cb9a35e

I'll try to do a broader review of encodings in JtR vs. librexgen soon-ish.

jfoug:
It is NOT a matter of casting.  strnzcpyn can not be used.  It is a utf16 value (@claudioandre change).  Thus we can NOT treat the data in any was as a char * in our code.  

I am VERY unhappy all the way overall with rexgen. That thing has had so many interface changes and NO control over compatibility.  What it does is not all that big a thing. We should simply write the damn thing ourselves, and quit depending upon Jan for any form of stability, since from history, we are not going to get stability.

jfoug:
Or fork the damn thing, and deal with it internally.

magnumripper:
Oh, I thought it'd be UTF-8 but obviously it's not... Damn. So I guess anything with `\0` is borken now?

jfoug:
Well, we could convert out of utf16 into utf8, do our processing, and marshal back into utf16 if we needed to return data.   I have not look at this code for a while, so I am not really sure any more.

But yes, we would only end up with 1 character (at most) if we use any str*() functions during the callback.

magnumripper:
Perhaps we should revert the latest commits and stick to a sane version of rexgen (eg. 159baaf). This really sucks.

magnumripper:
Briefly looking at mbtowc and similar functions, I only see character conversions. We'd need to write string conversions ourselves from that.

What we'd really need is something like `enc_to_wc(dst, src)` converting from our currently selected encoding (or internal codepage) to system's wchar_t strings.

jfoug:
> this really sucks.

lol, yup.
Like I said, what we really should do is simply fork rexgen, and make things that WORK  (in a branch).  Then we own the rexgen (at least the one john uses).

jfoug:
I do not know of any open 'needs list' for rexgen that are not there.  BUT we need to properly integrate.   Since rexgen has switch 100% to unicode under the hood, there is quite a bit of stuff in john not working. Things like the '-case' is not there (due to how he did the unicode stuff), BUT I think it is still possible.  Also here, the interface as john uses it changed.

But if we simply use HIS git repo, he will change things every time he tweaks, bug fixes or adds some non-useful (as far as john cares) feature.

magnumripper:
I think most changes to rexgen were totally needed... BUT I think this one should have been UTF-8 and not the system-specific wchar_t. I regard the latter as something you'd only use WITHIN black boxes, not in any interfaces.

magnumripper:
Anyway I'll probably fix this tomorrow. I have sort-of working code already.

PR #2298, opened 2016-10-06
OpenCL: a platform could have no devices

claudioandre:
- I have 1 zVIDIA GPU + 1 zMD GPU. Two platforms installed. 
- I removed the zVIDIA board, no software changes.
- the system will report 2 installed platforms with only one device available in one of them.
- Or one of my drivers starts to misbehave after an upgrade.

In any case, this patch is safe and tested. But:
- I have no access to a real crazy machine to improve it further.


magnumripper:
Looks good

PR #2301, opened 2016-10-10
CI: allow to disable ASAN using the commit message

claudioandre:
- Passed 'my' CI
- Also, I tested it using a commit message containing [no asan]

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,10 +25,12 @@ before_install:
   - sudo apt-get install fglrx-dev opencl-headers || true
   - export OMP_NUM_THREADS=4
 script:
-  - cd src 
-  - if [[ "$CC" == "gcc" ]]; then ./configure --enable-asan; fi
-  - if [[ "$CC" == "clang" ]]; then ./configure; fi
-  - make -sj4 
+  - cd src
+  - export LOG=$(git log -n 1)
+  - export ASAN=""
+  - if [[ $LOG != *'[no asan]'* && "$CC" == "gcc" ]]; then export ASAN="--enable-asan"; fi
+  - ./configure $ASAN
+  - make -sj4
   - ../.travis/test.sh
 
 addons:


magnumripper:
I'm not sure I see the point of this. So if collaborator1 pushes a "bad" commit with [no asan], it will pass. Then when collaborator2 pushes a good commit, it will fail due to the previous commit.

With current settings, topic branches can't be merged (by anyone but me, IIRC) as long as they fail CI/Travis. In my example, collaborator2's commit would be unmergable even though it was OK.

jfoug:
This is why I wondered (and asked) if a config file to do this was a better way to do this, since it would stay in the project and only be removed once the asan problem was finally nailed down.  BUT as @claudioandre stated, it is just as easy to remove the --enable-asan from the yml file, and then reinsert it later, that using a config file to do this does not make sense. 

It would really only  be for some last resort, where we were having problems getting someone else's asan problem fixed, BUT we wanted that code out there, and did not want to keep the build in a long state of 'failure' over some semi safe asan edge condition (especially something like a buffer read-past error, which is typically non-'fatal', but does need corrected)

magnumripper:
Right. It's a pretty remote scenario so I think just editing the yml file is the way to go, if anything. On a side note you could disable ASan for a function with `ATTRIBUTE_NO_ADDRESS_SAFETY_ANALYSIS` which expands to the function attribute `__attribute__((no_address_safety_analysis))`.

jfoug:
That might not be a bad way to do it either, IF there is some function that we just can't get figured out, that the test-full is killing us over, so we can have normal CI results UNTIL someone can nail the bug.

I love having ASAN added to the CI.  I just don't want to have jumbo appear to be fuk'd a lot of the time, when it really isn't.  This will make @frank-dittrich fuzzing hopefully not have to trigger dozens of bugs every time he does it.  Yes, the fuzzing will find things that -test-full does miss, but test-full finds things which are more thorough than more normal usage of john would ever encounter (pushes the hell out of edge conditions)

claudioandre:
I don't have a strong opinion. Anyway, edit the yml is a bad thing in my opinion. 
- I prefer the principle of the publicity, public use of reason and social control.

So, the usage of an attribute or to disable the format or to use the commit message.
***
If the PR is not the way to go, I guess one of you guys should close this.

PR #2302, opened 2016-10-11
CI: create a clang environment to test ASAN (skip OpenCL formats)

claudioandre:
- Test ASAN on clang (but skip OpenCL formats).
- Remove unused/unnecessary code.


magnumripper:
Can't merge this as it contains 5fc8e78 too

PR #2304, opened 2016-10-11
CI: test JtR (also) on a fresh environment

claudioandre:
- Addresses #2303 
- I still have to move to CentOS

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,10 +25,12 @@ before_install:
   - sudo apt-get install fglrx-dev opencl-headers || true
   - export OMP_NUM_THREADS=4
 script:
-  - cd src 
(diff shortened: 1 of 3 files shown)


jfoug:
I am not sure we need to do a ASAN on clang or not.

claudioandre:
Since JtR tests on gcc and clang, my preference is to do:
- a complete test using both;
- and test against a LTS and a fresh OS (using the 'default' compiler, at least).

claudioandre:
I created a new branch to allow you guys an easier review process. It adds a new OS on CI but:
- without ASAN on clang and;
- without the ability to disable ASAN using the commit message.

https://github.com/magnumripper/JohnTheRipper/compare/bleeding-jumbo...claudioandre:new_approach?expand=1

magnumripper:
Can't merge this as it contains 5fc8e78 too

PR #2309, opened 2016-10-13
update OSX compile instructions

FireFart:
This PR updates the compile instructions for OSX (tested with OSX Sierra 10.12)

- The brew link command is not recommended by homebrew. So the path to the include files is passed to `./configure`
- Use an option to provide the GCC compiler instead of symlinking it systemwide
- Added configure and make commands


magnumripper:
Thanks!

PR #2320, opened 2016-11-04
github: add guidelines to github users when they create an issue or a PR

### LINKED ISSUE #2319: Auto include guidelines when an user create a new issue (or PR)
As you can see below, it is possible to fill the 'new issue' text field with some guidelines.

- The goal is to prevent users from using github as a community forum.
****
image

claudioandre:
- Fixes and closes #2319

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,0 +1,11 @@
+### IMPORTANT
+Use issues to keep track of ideas, enhancements, tasks, and bugs. NEVER as a support forum. If you have questions, please ask them first in the mailing list john-users at lists.openwall.com.
+
+### Steps to reproduce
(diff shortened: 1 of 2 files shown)


jfoug:
We been needing this more and more lately.  I think the wording might be a bit more strong that it is.

Most problems are: 
- usage issues (RTFM)
- compile issues
- cuda crap (I would love to unsupport that cluster fuk)
- crashes due to opencl version, or low grade GPU

claudioandre:
> I think the wording might be a bit more strong that it is.

Probably. And we can state clearly about CUDA, OpenCL, whatever.

jfoug:
We likely should just have one open thread about each.  Then new 'issues' can be posted to that, and the thread owner can update the primary message (with a task list), and when completed, check off the task list.   Having dozens (or hundreds) of the exact same issues is not a great way to proceed.

PR #2327, opened 2016-11-12
Added hex masks to rules. lower, upper and mixed case hex classes and…

jfoug:
This is a pretty trival code change. EASY to test validity.

Here is a simple john-local.conf file to easily test and see that the new classes and anti-classes are working perfectly, and that no additional code is needed.

This branch is merge ready:

$ cat john-local.conf
[List.Rules:test]
/?h \Az"?h"
/?H \Az"?H"
/?i \Az"?i"
/?I \Az"?I"
/?j \Az"?j"
/?J \Az"?J"


magnumripper:
I do not want this merged. I see no need for it.

PR #2341, opened 2016-11-22
CI: improvement

claudioandre:
Instead of https://github.com/magnumripper/JohnTheRipper/pull/2302 and https://github.com/magnumripper/JohnTheRipper/pull/2304, I was tempted to use this approach:

In the PR you can see (just an example from a remote past):
- gcc (ASAN) and clang (without ASAN) evaluating the build
- gcc (ASAN) on a recent compiler (in case of problems, it won't fail the CI build)
- TS on gcc on 16.04 LTS (in case of problems, it won't fail the CI build)

Thinking about  5fc8e78,  if JtR is failing because of ASAN, the new "infra" allow to 'fix' the build, e.g.:
- evaluate the build using gcc and clang on a stable OS **without** ASAN 
- test gcc (or clang) with ASAN only on a recent compiler - without a build fail.
- TS and whatever  - without a build fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,17 +16,18 @@ matrix:
   - compiler: clang
     env: ASAN="" OPENCL="yes"
 
-  - env: ASAN="--enable-asan" FRESH="yes"
+  - env: ASAN="--enable-asan" TEST="fresh test"
+
+  - env: ASAN="" TEST="TS --restore"
 
   allow_failures:
-  - env: ASAN="--enable-asan" FRESH="yes"
+  - env: ASAN="--enable-asan" TEST="fresh test"
+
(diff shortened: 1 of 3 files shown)


claudioandre:
There are `--restore` issues on JtR. They were caught by CI (if anyone disagrees, please say so) in the PR. I guess #2082 is the main log of it.

BTW: I run it with -v -v and it is easy to see when the TS run stops to crack any new candidate.

claudioandre:
BTW: it is possible to reproduce the error on super. 
****
Seems (to me) a 'show stoppers' in #1879.

magnumripper:
Awesome. I audited session restoring thoroughly less than a year ago iirc and unfortunately it seems things have regressed since.

PR #2343, opened 2016-11-25
fix build against libressl

rofl0r:
libressl deprecated the SHA_* functions and kept only the SHA1_* ones.
since they do the same (the existing code even mixed  calls to them up!)
let's just use the SHA1 ones everywhere so we're compatible with openssl
and libressl.

### DIFF
--- a/src/hmac_sha_plug.c
+++ b/src/hmac_sha_plug.c
@@ -42,7 +42,7 @@ void JTR_hmac_sha1(const unsigned char *key, int key_len, const unsigned char *d
 
 	if (key_len > 64) {
 		uint32_t *p = (uint32_t*)buf;
-		SHA_Init(&ctx);
+		SHA1_Init(&ctx);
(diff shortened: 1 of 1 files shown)


solardiz:
For hmac_sha_plug.c, this looks like the correct fix.

I guess rawSHA0_fmt_plug.c fails self-test when patched like this? I think we should simply exclude it from the build when -lcrypto does not provide SHA-0.

jfoug:
This is not the correct fix.  Do not merge this.

magnumripper:
> This is not the correct fix. Do not merge this.

It is the correct fix for hmac_sha_plug.c. It obviously isn't for raw-SHA0, which is why the bots fail.

rofl0r:
updated. i suppose for the other code in the sha0 plugin the best fix would to be to check for existence of SHA_Init() in configure, and disable the plugin if not found, or have some fallback copy of the code in-tree.

Issue #2358, opened 2016-12-13
Hash identification reports invalid hash types


hgbstu:
When starting john without the --format parameter, hash type detection reports invalid hash types.

For example when trying to crack a md5 hash invalid information is printed
- dynamic=md5($p) is not a valid hash type. There is no such hash type in ./john --list=formats
- a md5 hash is identified as Raw-SHA1-* hash

### Steps to reproduce

> $ #md5(john)=527bd5b5d689e2c32ae974c6229ff785
> $ echo "527bd5b5d689e2c32ae974c6229ff785" > MD5.hash
> $ ./john MD5.hash
> Warning: detected hash type "LM", but the string is also recognized as "dynamic=md5($p)"
> Use the "--format=dynamic=md5($p)" option to force loading these as that type instead
> Warning: detected hash type "LM", but the string is also recognized as "HAVAL-128-4"
> Use the "--format=HAVAL-128-4" option to force loading these as that type instead
> Warning: detected hash type "LM", but the string is also recognized as "MD2"
> Use the "--format=MD2" option to force loading these as that type instead
> Warning: detected hash type "LM", but the string is also recognized as "mdc2"
> Use the "--format=mdc2" option to force loading these as that type instead
> Warning: detected hash type "LM", but the string is also recognized as "mscash"
> Use the "--format=mscash" option to force loading these as that type instead
> Warning: detected hash type "LM", but the string is also recognized as "mscash2"
> Use the "--format=mscash2" option to force loading these as that type instead
> Warning: detected hash type "LM", but the string is also recognized as "NT"
> Use the "--format=NT" option to force loading these as that type instead
> Warning: detected hash type "LM", but the string is also recognized as "Raw-MD4"
> Use the "--format=Raw-MD4" option to force loading these as that type instead
> Warning: detected hash type "LM", but the string is also recognized as "Raw-MD5"
> Use the "--format=Raw-MD5" option to force loading these as that type instead
> Warning: detected hash type "LM", but the string is also recognized as "Raw-MD5u"
> Use the "--format=Raw-MD5u" option to force loading these as that type instead
> Warning: detected hash type "LM", but the string is also recognized as "Raw-SHA1-AxCrypt"
> Use the "--format=Raw-SHA1-AxCrypt" option to force loading these as that type instead
> Warning: detected hash type "LM", but the string is also recognized as "Raw-SHA1-Linkedin"
[...]

PR #2359, opened 2016-12-13
Fix axcrypt prepare to keep the input fields unchanged

claudioandre:
For review. See #2358


src/rawSHA1_common_plug.c
```diff
@@ -98,6 +98,17 @@ int rawsha1_common_valid(char *ciphertext, struct fmt_main *self)
...
+	static char out[41];
+	int extra;
+
+	if (hexlen(ciphertext, &extra) != 32 || extra)
+		return rawsha1_common_valid(ciphertext, self);
+	sprintf(out, "%s00000000", ciphertext);
+	return rawsha1_common_valid(out, self);
+}
+
 char *rawsha1_common_split(char *ciphertext, int index, struct fmt_main *self)
```
magnumripper: You should be able to drop the `static` from `out` now.

PR #2367, opened 2016-12-20
loader: fix the --show

### LINKED ISSUE #2366: Option '--show' doesn't show the cracked passwords for a given password file
Note that the 2nd run is lying to me: all passwords were cracked! 
- BTW: the same happens on Linux. I was on (b957228)

```
C:\Users\claud\bleeding\run>john C:\Users\claud\Desktop\hash.txt --show:left

2 password hashes cracked, 0 left

C:\Users\claud\bleeding\run>john C:\Users\claud\Desktop\hash.txt --show
0 password hashes cracked, 2 left
```

### Steps to reproduce
I used these 2 self-test hashes and cracked them
```
$6$ojWH1AiTee9x1peC$QVEnTvRVlPRhcLQCk/HnHaZmlGAAjCfrAN0FtOsOnUk5K5Bn/9eLHHiRzrTzaIKjW9NTLNIBUCtNVOowWS2mN.
$6$saltstring$svn8UoSVapNtMuq1ukKS4tPQd8iKwSMHWjl/O817G3uBnIFNjnQJuesI68u4OTLiBFdcbYEdFCoEOfaS35inz1
```

claudioandre:
It is not "the" fix, but it shows what is wrong. See #2366.

It fixes and closes #2366.

### DIFF
--- a/src/loader.c
+++ b/src/loader.c
@@ -1896,6 +1896,7 @@ static void ldr_show_pot_line(struct db_main *db, char *line)
 	char *ciphertext, *pos;
 	int hash;
 	struct db_cracked *current, *last;
+	struct fmt_main *fmt_tmp;
 
 	ciphertext = ldr_get_field(&line, db->options->field_sep_char);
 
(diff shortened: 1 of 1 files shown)


magnumripper:
Thanks! This bug was introduced by me recently. This is probably very slow for large pot files and it also doesn't look fool-proof to me (for some other situations than what was shown in #2366). I'll give this some thought and see if there is a better solution (such as dropping the Jumbo-specific call to split completely or at least move it into the single-format situation).

claudioandre:
`--show` (w/ and w/o left) are very verbose on memdbg. While at it, you can also ponder over it.
* https://travis-ci.org/claudioandre/JohnTheRipper/jobs/185430725 (do search for 'leaks')

magnumripper:
> This solution is probably very slow for large pot files and it also doesn't look complete to me

Turns out it doesn't affect performance. And it was almost complete, I changed it do ldr_trunc_valid() in order to support truncated pot entries.

PR #2383, opened 2017-01-19
circle: fix the build issue introduced in a63b819

claudioandre:
See:
- https://circleci.com/gh/claudioandre/JohnTheRipper/13


magnumripper:
Perhaps a similar fix is needed for Makefile.legacy?

PR #2420, opened 2017-02-06
Add a configure option to treat build warnings as errors

claudioandre:
I'm working to make CI act as my watchman.


jfoug:
Good addition for the little change.

PR #2424, opened 2017-02-12
New Office OpenCL format joining the separate ones

magnumripper:
Need some bot checks on this one.


magnumripper:
@kholia speed did increase a tad, but not significant. But it's great not to depend on many CPU cores.

magnumripper:
So the bots fail. I see now there seems to be an intermittent problem, only seen with --test-full=0. I have no idea why.

magnumripper:
Doesn't seem to happen if I limit it to 2013 (`-cost=2013:2013`), only with 2007 and 2010.

magnumripper:
I can't reproduce the problem on my Macbook but at least Super's device 3 and 6 has been seen erroring out intermittently. Always at cmp_all(49) with --test-full=0

@claudioandre can you test this on your gear?

I have rules out the atomic stuff, they are not the problem. Sometimes that SHA-1 kernels simply end up with some incorrect calculation! Or do we have a buffer overrun somewhere?

magnumripper:
BTW please note this commit has been force pushed a number of times - use the latest!

solardiz:
On Sun, Feb 12, 2017 at 03:52:11PM -0800, magnum wrote:
> I can't reproduce the problem on my Macbook but at least Super's device 3 and 6 has been seen erroring out intermittently. Always at cmp_all(49) with --test-full=0

"cmp_all(49) with --test-full" reminds me of this gpg-opencl issue, which Dhiru says is likely in CPU code:

https://github.com/magnumripper/JohnTheRipper/issues/2242#issuecomment-274303803

Could be same bug.

magnumripper:
The self-tests for OpenCL always run with LWS=7 and GWS=49 and --test-full puts the correct key in the last index. So the similarity is likely an illusion.

Both are using SHA-1 though. I have narrowed this down to the loop kernel which does 500 iterations of SHA-1 per call. Sometimes it produces the correct result, sometimes not. I have yet to understand why only --test-full triggers it. This HAS to be driver bugs but since it's seen on nvidia also, I need to find a workaround. Not easy :-( I have thrown things at it for hours to see what sticks, but nothing do.

solardiz:
On Sun, Feb 12, 2017 at 05:31:52PM -0800, magnum wrote:
> The self-tests for OpenCL always run with LWS=7 and GWS=49 and --test-full puts the correct key in the last index. So the similarity is likely an illusion.

Oh.  We need --test-full to also run a test with all keys being correct and to require matches for all indices, if it doesn't do that already.

magnumripper:
> Oh.  We need --test-full to also run a test with all keys being correct and to require matches for all indices, if it doesn't do that already.

Maybe it does that too somewhere. I hate the --test-full code, it's hard to follow and shows incorrect errors. For example, if crypt_all returns 0 when there should have been a match, the error printed is cmp_one(0).

magnumripper:
The weirdest thing is the loop kernel is effectively unchanged from the old office2007-opencl and office2010-opencl formats. They have no problem but this one has. And not only with some particular driver but several.

claudioandre:
* I was at: Ubuntu 14.04.5 LTS (without the latest updates);
* I was using HEAD + [this patch](https://github.com/magnumripper/JohnTheRipper/commit/8adbca2082c735ca147b0fe7917c76287d6ea280.patch) [1];
* The error I saw for test and test-full [2];
* Everything was failing, except [3];


[1]
```
7cf8207835830371e040007fb59c14838a8b65a5 New Office OpenCL format joining the separate ones for Office 2007, Office 2010 and Office 2013. 
5082bdf8f9fec30ab142bc1f3f595b7c9292724e Enpass-opencl: Copy/paste bug
f4c209d38f2d766798ee3f26bda213bdaeb3fb27 OpenCL AES formats: Optimization just for good measure.
42011bde4f28c59e47c0c58945b0ccf0ca5d4c0c Add volatile to variables handled with atomic functions.
e66f6d146e5c4cb3728be51f0912f2cf787d7493 OpenCL AES: Add ECB interface.
77c438b70c56fb665b1a39a523fba7823a44dec6 7z-opencl: Drop an unused variable.
8226cb6082edf66423e6c6e945c900a8a9de7552 Merge pull request #2423 from claudioandre/push
87a5640dabaaed0ac0497cd4eb15db512487f065 OpenCL: be more resilient on --list=opencl-devices
```
[2]
```
$ ../run/john  --test-full=0 --format=office-opencl
Device 0: Juniper [AMD Radeon HD 6700 Series]
Testing: office-opencl, MS Office [SHA1/SHA512 AES OpenCL]... FAILED (cmp_one(0))
```
[3]
``` 
$ ../run/john  --test --format=office-opencl -cost=2007:2007
Device 0: Juniper [AMD Radeon HD 6700 Series]
Benchmarking: office-opencl, MS Office [SHA1/SHA512 AES OpenCL]... DONE
Speed for cost 1 (MS Office version) of 2007, cost 2 (iteration count) of 50000
Raw:	8827 c/s real, 75294 c/s virtual
```

```
$ ../run/john  --list=build-info
Version: 1.8.0-jumbo-1-5533-g7cf8207
Build: linux-gnu 64-bit SSE2-ac OMP
SIMD: SSE2, interleaving: MD4:3 MD5:3 SHA1:1 SHA256:1 SHA512:1
$JOHN is ../run/
Format interface version: 14
Max. number of reported tunable costs: 3
Rec file version: REC4
Charset file version: CHR3
CHARSET_MIN: 1 (0x01)
CHARSET_MAX: 255 (0xff)
CHARSET_LENGTH: 24
SALT_HASH_SIZE: 1048576
Max. Markov mode level: 400
gcc version: 4.8.4
GNU libc version: 2.19 (loaded: 2.19)
OpenCL headers version: 2.0
Crypto library: OpenSSL
OpenSSL library version: 01000106f
OpenSSL 1.0.1f 6 Jan 2014
File locking: fcntl()
fseek(): fseek
ftell(): ftell
fopen(): fopen
memmem(): System's
```

PS: 
* I tried LWS=128 GWS=1024 to check if it changes the error message (the first fails, the last ...). It doesn't.
* In fact I am a gearless guy. But the testing machine has the same hardware I used to have.
```
$ ../run/john  --list=opencl-devices
Platform #0 name: AMD Accelerated Parallel Processing, version: OpenCL 2.0 AMD-APP (1800.11)
    Device #0 (0) name:     Juniper
    Board name:             AMD Radeon HD 6700 Series  
    Device vendor:          Advanced Micro Devices, Inc.
    Device version:         OpenCL 1.2 AMD-APP (1800.11)
    Driver version:         1800.11 - Catalyst 15.9 [supported]
    Native vector widths:   char 16, short 8, int 4, long 2
    Preferred vector width: char 16, short 8, int 4, long 2
    Global Memory:          512.5 MB
    Local Memory:           32.0 KB (Local)
    Max memory alloc. size: 128.1 MB
    Max clock (MHz):        850
    Profiling timer res.:   1 ns
    Max Work Group Size:    256
    Parallel compute cores: 10
    Stream processors:      800  (10 x 80)
    Speed index:            680000
    SIMD width:             16
    Wavefront width:        64
    ADL:                    Overdrive5, device id 0
    PCI device topology:    01:00.0
```

magnumripper:
I have stared at all the code, host and CL, and diffs against the older working formats, for hours and hours. I'm completely stumped!

On my Macbook (CPU or GPU) I can run --test-full=0 in a loop all day, it never misses a beat.

My debug prints on super show that when it fails, the correct data is fed to the loop kernel's first run but incorrect data comes out of the last one. Run it again and (sometimes) all is fine. Two failures usually have the same incorrect data IIRC.

magnumripper:
Found it! Damn, a buffer overflow in first kernel. Many things had me concentrate too much on the loop kernel but the init kernel probably busted the state info for the loop one. Oh well. Will commit soon.

PR #2439, opened 2017-03-05
o5logon-opencl: fix an intermittent bug

claudioandre:
There are unnecessary changes here (I was playing with it). You can create your own patch and ignore the PR. 

I tested on super CPUs, but not on GPUs. I will rebase to test on CI.

### DIFF
--- a/src/opencl/o5logon_kernel.cl
+++ b/src/opencl/o5logon_kernel.cl
@@ -31,7 +31,7 @@ typedef struct {
 
 __kernel void
 o5logon_kernel(__global const uint *keys, __constant salt_t *salt,
-               __global const uint *index, __global volatile uint *result)
+               __global const uint *index, volatile __global uint *result)
 {
 	uint W[16] = { 0 }, salt_s[3], output[5];
 	uint gid = get_global_id(0);
@@ -53,9 +53,6 @@ o5logon_kernel(__global const uint *keys, __constant salt_t *salt,
 	} pt;
 	uchar iv[16];
 	AES_KEY akey;
-
-	if (gid == 0)
-		result[0] = 0;
 
 	keys += base >> 6;
 
@@ -89,5 +86,5 @@ o5logon_kernel(__global const uint *keys, __constant salt_t *salt,
 	AES_cbc_decrypt(&salt->ct[32], pt.c, 16, &akey, iv);
 
 	if ((result[gid + 1] = (pt.l[1] == 0x0808080808080808UL)))
-		atomic_max(&result[0], gid + 1);
+		atomic_max(result, gid + 1);
 }
--- a/src/opencl_o5logon_fmt_plug.c
+++ b/src/opencl_o5logon_fmt_plug.c
@@ -269,6 +269,7 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 	const int count = *pcount;
 	size_t gws;
 	size_t *lws = local_work_size ? &local_work_size : NULL;
+	result[0] = 0;
 
 	gws = GET_MULTIPLE_OR_BIGGER(count, local_work_size);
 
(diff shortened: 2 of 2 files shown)


magnumripper:
Hmm so what part of it is the necessary one?

src/opencl/o5logon_kernel.cl
```diff
@@ -54,9 +54,6 @@ o5logon_kernel(__global const uint *keys, __constant salt_t *salt,
 	uchar iv[16];
 	AES_KEY akey;
 
-	if (gid == 0)
```
claudioandre: Remove 57, 58, 59. Necessary

src/opencl_o5logon_fmt_plug.c
```diff
@@ -269,6 +269,7 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 	const int count = *pcount;
 	size_t gws;
 	size_t *lws = local_work_size ? &local_work_size : NULL;
+	result[0] = 0;
```
claudioandre: Initialize result[0]. Necessary

src/opencl_o5logon_fmt_plug.c
```diff
@@ -280,6 +281,10 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 			"failed in clEnqueueWriteBuffer buffer_keys");
 
 	BENCH_CLERROR(
+		clEnqueueWriteBuffer(queue[gpu_id], buffer_out, CL_FALSE, 0, sizeof(cl_uint), result, 0, NULL, multi_profilingEvent[1]),
+		"failed in clEnqueueWriteBuffer buffer_idx");
+
+	BENCH_CLERROR(
```
claudioandre: Transfer initialized result[0] to the device. Necessary.

magnumripper: I don't like these news at all. It probably means many other formats has potential problems. BTW I also tried putting a global memory barrier after L59 in the kernel and it didn't help! Given that, how am I supposed to depend on anything if thread n completes before thread 0 even starts?! The atomic functions are pretty much useless this way.

claudioandre:
I'm afraid I agree with the compiler in this case. Think about the code below.

```
__kernel void
  o5logon_kernel(volatile __global uint *result)
 {
	if (get_global_id(0) == 0)
		result[0] = 0; //Line 5.

	atomic_inc(result);
}
```

Suppose N "threads" are running:
* You would expect line 5 to have been run in "thread" 0 before ANY atomic() in line 7 for ALL threads;
* Why?
* The compiler and scheduler can see why looking at the code?

So:
- Is the initialization a parallel task? Not in my opinion.
- The result[0] has to be initialized before any parallel processing. 
****

If "thread" 3 run the atomic before the initialization in "thread" 0, you will have:
* result[0] = real_and_safe_atomic_operation(random + 1);
* and later, result[0] = 0;

magnumripper:
OK, but this code:
```
__kernel void
  o5logon_kernel(volatile __global uint *result)
 {
	if (get_global_id(0) == 0)
		result[0] = 0; //Line 5.

	barrier((CLK_GLOBAL_MEM_FENCE);

	atomic_inc(result);
}
```
This should guarantee proper function - but it doesn't help.

claudioandre:
I won't expect it to work. Read the specification:  
> All work-items in a **WORK-GROUP** executing the kernel on a processor must execute this function before any are allowed to continue execution beyond the barrier.

It guarantees the synchronization inside a work-group (only 1 compute unit), not across multiple work-groups (and/or CUs). 

I am playing devil's advocate. But, in fact, I read you can't trust a barrier to do what you need.

magnumripper:
You are absolutely right. There's simply no way to initialise that variable other than from host side or in an other kernel. I dropped the whole idea in 2860efa, speed wasn't affected anyway.

PR #2486, opened 2017-03-28
general: mute a warning and a possible truncation

claudioandre:
It is C99, I see no problems. Do you?


magnumripper:
Gosh, I thought it was a macro that would work regardless of type. Why wouldn't it?

claudioandre:
To avoid unsafe macro expansion, I guess.

* function is safe: `abs(i++);`
* unsafe macro expansion: `(i++ < 0) ? -(i++) : (i++)`

Issue #2538, opened 2017-05-10
(Some formats) auto-tune OpenCL for actual loaded hashes (as opposed to test vectors), if applicable


magnumripper:
This is very notable in the DMG format. I believe auto-tune is always done with test-vector iteration counts of 1000, while real-world hashes of today are in the hundred-thousands.

Should be easy, most things are in place already. Oh, and bitlocker-opencl should do the same, off the top of my head.

magnumripper:
BTW if we're really lucky, this can be a fix in shared auto-tune code that would affect all OpenCL formats with little effort.

claudioandre:
> this can be a fix in shared auto-tune code that would affect all OpenCL formats 

This is not possible 'as is'. You need decide when auto-tune (first or second call of `reset()`). Look at:
```
static void reset(struct db_main *db)
{
  if (db==test_vector && JtR_is_cracking) //it is NOT a run to benchmark, self-test, ...
     do_nothing(); //1st call of reset()
  else // db contains loaded hashes (if applicable)
     auto_tune(db); // 2nd call of reset; 1st call for a --test run.
}
```
You can't move this decision to shared code because `reset()` itself might need to do its tasks before or after tuning: move data to GPU, kernel (re)build, etc.

magnumripper:
Ah, yes. But anyway it's a fairly trivial change in each format (any many formats don't really need this).

magnumripper:
@claudioandre I found a way to do it in shared code. But it needs testing/review.

magnumripper:
Ooops, terrible bug fixed in ff70033

PR #2544, opened 2017-05-14
Autotune cost

magnumripper:
Auto-tune for actual loaded hashes instead of test-vectors, when applicable. See #2538.

This is not ready for merge yet. More testing needed.

### DIFF
--- a/src/common-opencl.c
+++ b/src/common-opencl.c
@@ -1216,7 +1216,6 @@ static cl_ulong gws_test(size_t gws, unsigned int rounds, int sequential_id)
 	int number_of_events = 0;
 	void *salt;
 	int amd_bug;
-	char *ciphertext;
 
 	for (i = 0; i < MAX_EVENTS; i++)
 		benchEvent[i] = NULL;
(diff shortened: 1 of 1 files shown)


src/common-opencl.c
```diff
@@ -1252,13 +1251,24 @@ static cl_ulong gws_test(size_t gws, unsigned int rounds, int sequential_id)
...
 	dyna_salt_init(self);
-	if (!self->params.tests[0].fields[1])
-		self->params.tests[0].fields[1] = self->params.tests[0].ciphertext;
-	ciphertext = self->methods.prepare(self->params.tests[0].fields, self);
-	ciphertext = self->methods.split(ciphertext, 0, self);
-	salt = self->methods.salt(ciphertext);
-	if (salt)
-		dyna_salt_create(salt);
+	if (self->methods.tunable_cost_value[0] && autotune_db->real) {
+		struct db_main *db = autotune_db->real;
```
claudioandre: The real db availability in the first reset() call (via *real) solves all problems. I was not aware of it.
- I will run all 'my' tests against it. But, I see no problems in autotune code. 

Maybe one concern: are you sure there is no "fast" format with cost and 100 millions of loaded hashes?

magnumripper: > are you sure there is no "fast" format with cost and 100 millions of loaded hashes?

It shouldn't be a problem. We are still only tuning with one salt (the slowest). The tuning will be suboptimal for faster costs but that's better than the old behavior.

magnumripper:
This was rebased but the actual patch is not changed as of now.

PR #2556, opened 2017-05-23
maintenance: fix the usage of an uninitialized variable

claudioandre:


magnumripper:
Ouch. I wonder how my compilers could miss warn about this. I made a slightly different fix for being closer to original `fgetl()`.

magnumripper:
I'm an idiot, my fix didn't quite help. Hopefully 4396d2517 puts an end to this.

PR #2557, opened 2017-05-23
kerberom: new version adding Windows support and Windows implicit authentication mechanisms

Fist0urs:
### Summary

This is a new version of the run/kerberom tool. It provides Windows support (needing PyInstaller if you want to create a standalone binary) and especially the support of Windows implicit authentication. This is particularly interesting if one managed to get another current logon context, for example in redteam assessments, thus not needing any credential to use kerberom and retrieve TGS to crack with krb5tgs format.

### Other Information

The Linux part was a bit modified so that it output samaccountname:ticket, but this does not impact john implementation :+1:  .

This should not be the final version, when I find time I'll implement the other krb5tgs format (thus not based on ARC4 but SHA1) and adapt the tool to retrieve the corresponding hashes :wink: 

Cheers!

### DIFF
--- a/run/kerberom/README.md
+++ b/run/kerberom/README.md
@@ -11,21 +11,21 @@ Cracking these tickets gives you the associated accounts' password within the Ac
 
 **You do not need any third-party tools that are OS dependents (like mimikatz or PowerShell) and do not need privileged rights to use kerberom**
 
-
 Author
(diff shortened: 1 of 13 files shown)


magnumripper:
Cool, thanks!

PR #2598, opened 2017-06-30
Rename for more consistent directory names; added fpga-descrypt/READM…

Apingis:
…E.md and summary  fpga-descrypt\ztex_inouttraffic_summary.html

### Summary

Functionality isn't touched.


claudioandre:
A side note and suggestion: 
* I missed a readme file, a link or a real one, inside the doc folder (recently, while I was adding ztex to CI [1] and needed some info about it).
* I had to sed to find one (it wasnt obvious to me where to find it).

[1] of course, to test the code itself.

solardiz:
Claudio, you're right.  I think the only README files we have now are on the FPGA designs (Denis wrote one on bcrypt first, and now I asked him to do the same on descrypt, which he did in this pull request).  We also need an end-user focused README-ZTEX file or such in the top-level doc/ directory.  Denis?  (Not to be made part of this same pull request - no reason to delay merging it.)

solardiz:
> recently, while I was adding ztex to CI

So, have you added it?  I think this amounts to having libusb-devel (or the like) installed, and passing the `--enable-ztex` option to `configure`.  Of course, this will only make sure the host side code can be built and doesn't hurt other functionality... which I'm afraid it might, by design: "--test" without a specified "--format" will fail upon encountering a ZTEX format and no ZTEX board.  But if CI doesn't do that (what tests does it run?), we're fine.  Thanks!

claudioandre:
> So, have you added it?

Yes. It already runs when I ask to.
- werror, ASAN and memdbg enabled.
- and I check if it fails with no ztex device.

PR #2602, opened 2017-07-02
Оn bad input packet, board correctly removed from operation.

### LINKED ISSUE #2599: descrypt-ztex: segfault on "no task" / "bad input packet"
I have a board with what appears to be connectivity or other issues. It sometimes resets successfully, but eventually, it segfaults and cannot recover, so the entire job aborts.

Once the board is no longer reachable, it would be an improvement if the job could continue with the remaining boards, because jobs are often unattended and long-running.

Example of successful reset:

```
$ ./john -form=descrypt-ztex -inc=alpha -min-len=8 -max-len=8 -mask='?w?l?l?l?l' pw-fake-unix
[...]
morecats         (u2406-des)
asdfasdf         (u915-des)
SN XXXXXXXXXX: Timeout.
Deassigned: 1
Found 1 device(s) ZTEX 1.15y
SN: XXXXXXXXXX productId: 10.15.0.0 "inouttraffic UFM 1.15y" busnum:1 devnum:38
starwars         (u797-des)
[...]
133g 0:00:07:27  0.2974g/s 0p/s 9590Mc/s 13996MC/s loveaaaa..lovealks
clarkson         (u1872-des)
[...]

```
Example of segfault:

```
tehdumwk..tutealks
meatloaf         (u2363-des)
XXXXXXXXXX CMP_EQUAL id=921: no task
SN XXXXXXXXXX: bad input packet
Deassigned: 4
Segmentation fault

```

Apingis:
Closes #2599

### DIFF
--- a/src/ztex/device_format.c
+++ b/src/ztex/device_format.c
@@ -212,7 +212,13 @@ int device_format_crypt_all(int *pcount, struct db_salt *salt)
 		}
 
 		// Process input packets, store results in task_result
-		jtr_device_list_process_inpkt(task_list);
+		for (;;) {
+			struct jtr_device *dev
+					= jtr_device_list_process_inpkt(task_list);
+			if (!dev)
+				break;
+			device_stop(dev->device, task_list, "bad input packet.");
+		}
 
 		// Computation done.
 		if (task_list_all_completed(task_list))
--- a/src/ztex/device_format.c
+++ b/src/ztex/device_format.c
@@ -25,6 +25,12 @@
 // then it counts the device as not functioning one.
 #define DEVICE_TASK_TIMEOUT	5
 
+// Control the behavior in case where no devices are ready
+// because of errors. Startup behavior is not affected.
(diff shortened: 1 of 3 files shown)


src/ztex/device_format.c
```diff
@@ -25,6 +25,12 @@
 // then it counts the device as not functioning one.
 #define DEVICE_TASK_TIMEOUT	5
 
+// Control the behavior in case where no devices are ready
+// because of errors. Startup behavior is not affected.
+// Set to 1 if you want to wait until any device is up.
+// Consider ztex_scan.h:ZTEX_SCAN_INTERVAL_DEFAULT
+const int WAIT_UNTIL_DEVICE_UP = 1;
```
solardiz: This is inconsistent use of #define vs. "const int", but regardless if we use "const int" we should also use the "static" specifier as otherwise the symbol is unnecessarily exported.

PR #2625, opened 2017-07-10
Update INSTALL-UBUNTU

wmacevoy:
Fix path in REGEN support line (install script is in project root directory, not src subdirectory)


magnumripper:
Thanks

Issue #2652, opened 2017-08-04
descrypt-opencl slow and confusing startup at many salts (appears stalled)


aguvillalba:
When trying to execute:
`./john --test=0 --format=descrypt-opencl --device=0`

I got this output:
```
Device 0: GeForce GTX 970
Testing: descrypt-opencl, traditional crypt(3) [DES OpenCL]... Build log: 
ptxas info    : 0 bytes gmem
ptxas info    : Compiling entry function 'DES_bs_25' for 'sm_52'
ptxas info    : Function properties for DES_bs_25
ptxas         .     0 bytes stack frame, 0 bytes spill stores, 0 bytes spill loads
ptxas info    : Used 168 registers, 1796 bytes smem, 336 bytes cmem[0]
Build log: 
ptxas info    : 0 bytes gmem
ptxas info    : Compiling entry function 'DES_bs_25' for 'sm_52'
ptxas info    : Function properties for DES_bs_25
ptxas         .     0 bytes stack frame, 0 bytes spill stores, 0 bytes spill loads
ptxas info    : Used 168 registers, 1796 bytes smem, 336 bytes cmem[0]
Build log: 
ptxas info    : 0 bytes gmem
ptxas info    : Compiling entry function 'DES_bs_25' for 'sm_52'
ptxas info    : Function properties for DES_bs_25
ptxas         .     0 bytes stack frame, 0 bytes spill stores, 0 bytes spill loads
ptxas info    : Used 168 registers, 1796 bytes smem, 336 bytes cmem[0]
Build log: 
ptxas info    : 0 bytes gmem
ptxas info    : Compiling entry function 'DES_bs_25' for 'sm_52'
ptxas info    : Function properties for DES_bs_25
ptxas         .     0 bytes stack frame, 0 bytes spill stores, 0 bytes spill loads
ptxas info    : Used 168 registers, 1796 bytes smem, 336 bytes cmem[0]
PASS
```

When trying to crack using --format=descrypt-opencl I get the same output.

Here my info for ./john --list=build-info:
```
Version: 1.8.0-jumbo-1-5774-g8d7ccaf
Build: linux-gnu 64-bit SSE4.1-ac OMP
SIMD: SSE4.1, interleaving: MD4:3 MD5:3 SHA1:1 SHA256:1 SHA512:1
$JOHN is ./
Format interface version: 14
Max. number of reported tunable costs: 3
Rec file version: REC4
Charset file version: CHR3
[...]

PR #2655, opened 2017-08-06
Warn about the slow startup at many salts in descrypt-opencl

claudioandre:
See #2652
****
* Notes:
  * Parallel per-salt kernel build is disabled by default on HEAD.
  * The 4096 limit is not working properly. That said, anyone can crash JtR HEAD using a proper hash file.

```
#0  0x000000000079dc64 in create_tables (htype=<value optimized out>, loaded_hashes_ptr=<value optimized out>, 
    num_ld_hashes=<value optimized out>, offset_table_ptr=0x7fffffffa090, offset_table_sz_ptr=0x7fffffffa098, 
    hash_table_sz_ptr=0x7fffffffa09c, verb=0) at bt.c:474
#1  create_perfect_hash_table (htype=<value optimized out>, loaded_hashes_ptr=<value optimized out>, num_ld_hashes=<value optimized out>, 
    offset_table_ptr=0x7fffffffa090, offset_table_sz_ptr=0x7fffffffa098, hash_table_sz_ptr=0x7fffffffa09c, verb=0) at bt.c:696
#2  0x0000000000611192 in fill_buffer (salt=<value optimized out>, max_uncracked_hashes=<value optimized out>, max_hash_table_size=0xd1eb04)
    at opencl_DES_bs_plug.c:224
#3  0x0000000000611752 in build_tables (db=<value optimized out>) at opencl_DES_bs_plug.c:423
#4  0x000000000060ca3d in reset (db=0xd64600) at opencl_DES_bs_h_plug.c:667
#5  0x000000000074aee5 in john_run () at john.c:1710
#6  0x000000000074b785 in main (argc=8, argv=0x7fffffffe398) at john.c:1996
```

### DIFF
--- a/src/opencl_DES_bs_f_plug.c
+++ b/src/opencl_DES_bs_f_plug.c
@@ -683,12 +683,25 @@ static void reset(struct db_main *db)
 			salt_list[num_salts++] = (*(WORD *)salt -> salt);
 		} while ((salt = salt -> next));
 
+		if (num_salts > 1)
+			fprintf(stderr, "Note: building per-salt kernels. "
+				"This takes e.g. 2 hours for 4096 salts.\n");
 #if _OPENMP && PARALLEL_BUILD
 #pragma omp parallel for
 #endif
-		for (i = 0; i < num_salts; i++)
+		for (i = 0; i < num_salts; i++) {
 			init_kernel(salt_list[i], gpu_id, 1, 0, forced_global_keys ? 0 :local_work_size);
 
+#if _OPENMP && PARALLEL_BUILD
+			if (omp_get_thread_num() == 0)
+#endif
+			{
+				opencl_process_event();
+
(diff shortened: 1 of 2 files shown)


solardiz:
The changes in this pull request look good to me. Thank you, Claudio!

> Parallel per-salt kernel build is disabled by default on HEAD.

So we lost the ability to have the kernel builds run in parallel with initial (slower) cracking in order to gain parallel builds, but in the end have neither. Maybe we should review (much) older revisions and revert? I'd also like to have the ability to just run with a generic and slower kernel (working for all salts), like we used to have, for those cases where lower startup time is more important than c/s. Maybe we need to open separate issue(s) for this/these.

> The 4096 limit is not working properly. That said, anyone can crash JtR HEAD using a proper hash file.

Please open a separate issue for this. Thanks!

PR #2665, opened 2017-08-14
Mute ptxas info (from build log) in default verbosity

claudioandre:
If nobody complains, I will merge.


solardiz:
Looks good to me. Thanks.

Issue #2740, opened 2017-09-08
Report Argon2 flavor as a "tunable cost"


solardiz:
Jim merged Argon2d and Argon2i into one format, but it is useful to have this difference for the loaded hashes reported as a "tunable cost". These are similar speed, but a slight speed difference does exist.

This will be especially important as/if we add support for more flavors (2id, 2ds).

A related issue is that ideally 2d (and 2id?) should be benchmarked with t=1, but 2i with t=3. Right now, we only get benchmarks for t=3, which is not a recommended and thus not the most relevant setting for benchmarking and optimizing 2d. Can we fix that without un-merging the formats? If we have to, maybe un-merge only in terms of format struct's, but keep them in the same source file?

While at it, I'd prefer the benchmarks to be for m=2048 rather than the current m=4096 as 2048 would be comparable to what some others have and also show code speed differences better due to greater use of CPU caches, so maybe revise the test vectors accordingly? Was there any specific reason to choose 4096? I think it's neither on the fast end (2048 is), nor on the safer end (16384 is).

And I think we should s/argon2/Argon2/ and s/Blake2/BLAKE2b/ in this format's reporting.

And I guess OMP_SCALE 16 is arbitrary, and should probably set to 1 given how slow Argon2 typically is - or otherwise it'd be unnecessarily uncomfortable to run attacks on large numbers of salts.

frank-dittrich:
@solardiz 
Currently, we report: t, m, l.
In https://www.cryptolux.org/images/0/0d/Argon2.pdf, I see m and t mentioned, but  not l.
Should we change `l` to `p` or `lanes` ?

If we report the type (0: Argon2d, 1: Argon2i), we have a total of 4 tunable costs.
The current max. number of tunable costs is 3:

```
$ ./john --list=build-info | head -n 6
Version: 1.8.0.6-jumbo-1-3250-gaa7634f
Build: linux-gnu 64-bit AVX-ac OMP
SIMD: AVX, interleaving: MD4:3 MD5:3 SHA1:1 SHA256:1 SHA512:1
$JOHN is ./
Format interface version: 14
Max. number of reported tunable costs: 3
```

So, `FMT_TUNABLE_COSTS` needs to be increased in formats.h.
It should be possible to do this without changing  `FMT_MAIN_VERSION`.

I can prepare a patch, but would like to know whether `l` should be renamed or not.

frank-dittrich:
I decided to rename l to p.

[...]

PR #2742, opened 2017-09-09
Enhance Argon2 tunable cost reporting

frank-dittrich:
This should fix https://github.com/magnumripper/JohnTheRipper/issues/2740


frank-dittrich:
If #2740 requires splitting argon2d and argon2i into different formats, reporting a 4th tunable cost is not needed anymore.

PR #2746, opened 2017-09-10
Mute a compiler warning [-Wmacro-redefined]

claudioandre:
I found no reason for a printf re-definition.

### DIFF
--- a/src/asn1_plug.c
+++ b/src/asn1_plug.c
@@ -18,7 +18,7 @@
 #include "asn1.h"
 #include "jumbo.h"
 
-#define printf(...)
+#define wpa_printf(...)
 
 int asn1_get_next(const uint8_t *buf, size_t len, struct asn1_hdr *hdr)
 {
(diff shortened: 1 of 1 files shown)


claudioandre:
I probably miss the point. If anyone want to mute the debug message, why there was a rename?
* Was there a rename, right? wpa_printf() -> printf()
* Mute - Isn't easier just to do `#define wpa_printf(...)`?
* Debug - use the real printf `#define wpa_printf    printf`?

magnumripper:
I have no idea why it's written as it is, but we probably want the current behavior where these messages are never shown.

magnumripper:
> we probably want the current behavior where these messages are never shown.

Note that I didn't really look into this, the above is a guess based on the assumption that we decrypt some data with a key derived from a guess, then try to decode it with this ASN.1 code. We're expecting errors (wrong candidate -> random data fed into ASN.1 decoder) so it should silently reject ASAP.

claudioandre:
Ok, but what is wrong with the original/upstream function name `wpa_printf`?
* the new patch, closer to upstream, might make things clear?

magnumripper:
Nothing is wrong with that, we should go that way. Apparently it's been like this since this file was introduced in 063b838e0 - I guess that better notion came later in upstream. White at it, is there any other improvements in upstream?

claudioandre:
> White at it, is there any other improvements in upstream?

I would say there isn't.

PR #2756, opened 2017-09-15
bugfix in bcrypt-ztex bitstream; $JOHN; --devices

### LINKED ISSUE #2714: "ztex/*" runtime pathnames should use $JOHN and be path_expand()'ed
descrypt-ztex failed with the following messages:
```
$ LANG=C john --test=5 --format=descrypt-ztex
Benchmarking: descrypt-ztex, traditional crypt(3) [DES ZTEX]... fopen(ztex/inouttraffic.ihx): No such file or directory
fopen(ztex/inouttraffic.ihx): No such file or directory
no valid ZTEX devices found
```

There is ztex/inouttraffic.ihx in run/ and in src/ . I can run the
test ok standing in run/ . After that, I can run descrypt-ztex from
other folders too. But bcrypt-ztex fails in other folders:
```
$ LANG=C john --test=5 --format=bcrypt-ztex
Benchmarking: bcrypt-ztex [Blowfish ZTEX]... fopen(ztex/ztex115y_bcrypt.bit): No such file or directory
```

I think, path starting from $JOHN should be used always.

john was rebuilt today after `git pull`:
Version: 1.8.0-jumbo-1-5825-g2a18678
Build: linux-gnu 64-bit SSE4.1-ac
SIMD: SSE4.1, interleaving: MD4:3 MD5:3 SHA1:1 SHA256:1 SHA512:1

Apingis:
### Summary

- The bug in bcrypt-ztex bitstream caused error when all 4 ranges were sent to boards. That could happen with low setting of hashes and short ranges (such as ?d?d?d?d)
- Using $JOHN, path_expand() so it accesses ztex files from any CWD. Closes #2714
- Added --devices command-line options. Unlike GPUs, devices must be listed by their Serial Numbers e.g. --dev=04A36E0000,04A36D0000

### Other Information

- New bitstream has slightly different timing. It still guarantees 141 MHz but overclocking abilities might differ.
- Tested several john invocations with different set of devices.


src/ztex/ztex.c
```diff
@@ -468,6 +496,12 @@ int ztex_scan_new_devices(struct ztex_dev_list *new_dev_list, struct ztex_dev_li
 	}
 
 	libusb_free_device_list(usb_devs, 1);
+
+	if (warn_open && num_fail_open) {
+		fprintf(stderr, "Warning: unable to access %d board(s), that could be"
+			" insuffisient permissions\nor other john invocation is running.\n",
```
solardiz: Should be spelled "insufficient". Also let's close and re-open the quotes after "\n" for readability. Something like this:

```c
		fprintf(stderr, "Warning: unable to access %d board(s), could be "
			"insufficient permissions or\n"
			"another instance of john running\n",
```

src/ztex/ztex.c
```diff
@@ -69,6 +71,23 @@ int vendor_request(struct libusb_device_handle *handle, int cmd, int value, int
...
+		if ( !( (sn[i] >= '0' && sn[i] <= '9') || (sn[i] >= 'A' && sn[i] <= 'F')
+				|| (sn[i] >= 'a' && sn[i] <= 'f')) )
+			return 0;
+	}
+	if (sn[i])
+		return 0;
+
+	return 1;
+}
+
```
solardiz: Is there a risk to run into issues similar to what we had with non-default firmware - someone having a serial number patched to be of non-standard format? Or are they converted from raw binary data to hex by the firmware itself, so have to be of the correct format as long as the firmware is correct? BTW, I noticed a serial number may change from fake but correct format (could it as well not have been correct format?) to real upon our firmware upload - but I guess we won't even proceed with firmware upload if the initially seen SN is not of correct format.

We should be able to use and ideally refer to with the `--devices` option to all devices, even those with incorrect SNs, although I understand this can't be always possible as e.g. two devices might have not only incorrect, but also matching SNs.

src/ztex/inouttraffic.c
```diff
@@ -512,7 +509,8 @@ int fpga_progclk_raw(struct fpga *fpga, int clk_num,
 int fpga_progclk(struct fpga *fpga, int clk_num, int freq)
 {
 	if (DEBUG)
-		printf("fpga_progclk(%d,%d,%d)\n", fpga->num, clk_num, freq);
+		fprintf(stderr, "fpga_progclk(%d,%d,%d)\n",	fpga->num,
+				clk_num, freq);
```
solardiz: Looks like a tab character slipped onto the `fprintf` line.

solardiz:
> New bitstream has slightly different timing. It still guarantees 141 MHz but overclocking abilities might differ.

FWIW, my testing on 4 boards so far shows no adverse change in overclocking ability compared to testing a couple of weeks ago with the previous bitstream. 150 MHz appears to work OK (getting the expected 239 guesses in each of 12 runs, 2 minutes each), 155 MHz not.

solardiz:
Why is the compile-time default 140 MHz if the design tools guarantee 141 MHz? Shouldn't we change it to be 141?

solardiz:
Testing the `--devices` option, I split my 4 boards into 2 sets of 2 boards, with 2 concurrent invocations of john with bcrypt-ztex. Somehow this results in errors every few minutes, even though the same 4 boards work without errors when all are used from one instance of john. The errors vary between at least `pkt_comm_status=0x01`, `pkt_comm_status=0x02`, `app_status=0x20` and are reported for varying FPGAs, but are always reported at once for both FPGA boards shared by an instance of john. That is, every few minutes one instance of john prints two errors at once, for its two boards. The other instance does not print errors at the same time, but does at different times (also in pairs, for two of its boards). USB bandwidth usage is low, per `usbtop`:

```
  Device ID 40 :			64.30 kb/s	79.03 kb/s
  Device ID 41 :			63.75 kb/s	71.72 kb/s
  Device ID 42 :			76.55 kb/s	73.73 kb/s
  Device ID 43 :			65.52 kb/s	73.71 kb/s
```

I guess it's low due to on-device mask. The c/s rates are good except when errors occur. This is at 141 MHz.

src/ztex/ztex.c
Apingis: According to USB spec, SN can be arbitrary string up to 253 bytes. That means a corrupted or intentionally developed firmware can return anything. (In practice Ztex SDK assumes that a descriptor fits into one 64-byte hardware buffer EP0, hence 62 bytes limit).

A user can set any valid SN using FWLoader or API. In such case original SN is preserved and remains hidden, user-defined SN is serviced on USB Get String Descriptor requests. That's if firmware contains standard routines from Ztex SDK. John firmware also behaves this way.

Firmware on one of your boards may be the one that does not honor user-defined SN, uses only original SN.

I've checked include files from Ztex SDK, it operates as follows:
- There're two places where SN is stored: user-modified settings and constant settings. Both data structures reside in external I2C-connected EEPROM chip. On MCU startup, it reads user settings first. 10-byte SN is stored in ASCII and there's no check for valid hex on MCU side in either read or write function. If user-defined SN equals to "0000000000" then it proceeds to reading constant settings (*). There're 5 bytes which are converted to hex.

(*) These constant settings might be called constant only because standard firmware never writes there.

If there's a board with some incorrectly behaving firmware, the user can be advised to:
1. power-cycle the board with JP1 (the only jumper) switched. It boots some "internal firmware";
2. then upload "default firmware" or john's firmware into EEPROM using FWLoader. (this can be skipped).

Apingis:
I'm testing `--devices` option having 2 `john` processes: the 1st operates 2 boards and the 2nd operates 1 board. No problems so far (cygwin x64).

While in development, I've noticed following libusb feature: it requires to open() the device to get its iSerialNumber. I'm not aware of the reasons for such API requirement, as it allows to get idVendor and idProduct without open()'ing.
When the device is opened by other process, it can't be opened one more time, returns error. At startup, `john` reports number of boards it's unable to access. On subsequent USB scans that happen every 15 seconds, it doesn't display any message.
I wonder if such attempts to open() can iterfere with already opened boards.

Apingis:
> Why is the compile-time default 140 MHz if the design tools guarantee 141 MHz? Shouldn't we change it to be 141?

It happens that parameters to hardware units that generate clocks must be defined at the start of synthesis. After tools perform all the stages (that takes several hours), you can see the timing for the worst path.

On stable designs such as this one, I see difference 0.2-0.3 MHz between runs, I think startup clock frequency can be adjusted more precisely as resulting timing is known in advance with high probability.
Actually there're 3 frequency values:
- hardcoded into bitstream;
- hardcoded into a format that uses given bitstream;
- in config file.

solardiz:
My suggestion is simply to adjust the value "hardcoded into a format that uses given bitstream" after the bitstream is built, and commit both the bitstream update and, if changed, the format's source file at once. You do know it's 141 MHz now (after you've generated the bitstream), so perhaps the format should use that. There's no chicken-egg problem here, since the format isn't input to synthesis.

solardiz:
I ran another test of `--devices`, with the same 2+2 split as before. This time, I replaced `--mask='?l?l?l?l'` with `-inc=lower -min-len=4 -max-len=4`. I ran this 30 times on each set of 2 boards, with a one-line script, two instances of the script running at once targeting the two groups of boards. This time, the paired error (two boards of same instance) appeared only once, and also a single error (one board) appeared once. The remaining 58 runs completed without error. Also, all 60 cracked the expected 239 passwords all times. Thus, it appears that moving from on-device mask to passing of candidate passwords from host reduces the occurrence of errors - or rather, as I suspect, the detection of errors. Maybe with on-device mask it is more likely that communication errors would be detected rather than remain undetected? Are candidate passwords sent from host part of what's protected with a checksum or not? The fact that all passwords got cracked all the time anyway isn't convincing evidence for lack of communication errors because most candidate passwords are not expected to result in a successful guess. (We need support for >3 hashes/salt, with transfer of computed hashes back to host, to run more convincing tests.)

`usbtop` output when running incremental mode as above, two samples (the exact bandwidth usage figures kept changing rapidly - maybe `usbtop` measures over too short intervals compared to our crypt_all() duration):

```
  Device ID 40 :			407.36 kb/s	93.83 kb/s
  Device ID 41 :			842.27 kb/s	93.89 kb/s
  Device ID 42 :			605.69 kb/s	92.99 kb/s
  Device ID 43 :			603.08 kb/s	96.25 kb/s
```

```
  Device ID 40 :			817.24 kb/s	93.97 kb/s
  Device ID 41 :			575.18 kb/s	91.79 kb/s
  Device ID 42 :			588.53 kb/s	96.25 kb/s
  Device ID 43 :			466.76 kb/s	96.25 kb/s
```

Still plenty of idle bandwidth. :-)

Apingis:
> My suggestion is simply to adjust the value "hardcoded into a format that uses given bitstream" after
> the bitstream is built, and commit both the bitstream update and, if changed, the format's source file at once. 

I agree on the point.
Should there be some debugging issues, both files - bitstream and format's source - can be reverted at once.

> Are candidate passwords sent from host part of what's protected with a checksum or not?

Sure, everything is protected: both header and data of packets; packets of every type.

I continue testing using mask, no errors so far. I'm going to run different test, also going to test on other hardware.

solardiz:
> I agree on the point.

So maybe you commit a change from 140 to 141 in `ztex_bcrypt.c` now (that is, for the current bitstream)?

solardiz:
The 140 to 141 change went in with a separate PR.

The typo fixes I suggested in here are still not committed.

The errors seen with concurrent use of two groups of boards also occur for me when I reduce the size of each group to just one board - so 1+1, leaving the other 2 boards unused (but powered on and connected to the same USB hub; I haven't tried physically leaving just 1+1). They become moderately less frequent (e.g., once per 10 minutes when using on-device mask) than they are in the 2+2 case, though.

Apingis:
I'm preparing commit for typos.

As for testing. So far I'm not able to reproduce errors. I'll be testing on other PC with different USB subsystem which seems to be more buggy (such as I'm able to reproduce the following: when I unplug or power-off 1 board, others display USB I/O errors).

If you continue testing, could you please narrow down the issue doing the following:
in src/ztex/ztex_scan.h greatly increase
// Scan interval in seconds.
#define ZTEX_SCAN_INTERVAL_DEFAULT 15
That will de-facto disable timely USB rescan while john is running and it would become clear if errors are connected to timely scanning.

solardiz:
OK, I increased `ZTEX_SCAN_INTERVAL_DEFAULT` from 15 to 150000 and re-ran the 2+2 mask mode test of bcrypt-ztex. After a few minutes, one of two instances reported a paired error (for both of its boards at once), and after that point it got stuck with no further progress and no response to keypresses yet still slowly eating up CPU time. gdb shows it doing `usleep()` from `
device_format_crypt_all()`. Is that an expected side-effect of the lack of rescans, or should it have recovered anyway?

solardiz:
With the rescans reintroduced so that a process doesn't get stuck on first occurrence of the problem, I am able to get very frequent errors (many times per minute) by running two dissimilar scripts/instances: one with a decent mask as before (for a few minutes per session) and the other with a tiny mask (for a few seconds per session, obviously also achieving poor efficiency but that's unimportant). It appears that startup of the frequently-restarting short sessions with tiny mask often (not always, but often) triggers errors in the longer-running sessions. Thus, it appears there's something in startup code that is causing this. Maybe not only in startup code, but at least in startup code - initial opening of the USB devices, checking firmware and bitstream, etc.

solardiz:
Yet another new thing: having run the two dissimilar scripts as above for a while, I started getting persistent `Self test failed (cmp_one(0))` at startup of the longer-running sessions. This persisted even after I killed off the other script. Yet somehow `--test` on all four boards worked fine. Even more puzzling is that even after successful `--test`, I still got the self-test failure when I tried to run a cracking session on those two boards. Well, maybe `--test` of four boards does not test the specific indices that are used by self-test of only two boards. I managed to recover by temporarily switching to descrypt and back to bcrypt, so that the bitstreams were re-uploaded. After that, cracking on the two boards that previously failed self-test works again.

Apingis:
> [...] it got stuck with no further progress and no response to keypresses yet still slowly eating up CPU time. gdb shows it doing usleep() from device_format_crypt_all(). Is that an expected side-effect of the lack of rescans, or should it have recovered anyway?

Current code requires rescan after any error.

> [...] I started getting persistent Self test failed (cmp_one(0)) at startup [...] managed to recover by temporarily switching to descrypt and back to bcrypt, so that the bitstreams were re-uploaded.

It looks like the following.
Data in some FPGA units (parts of units), most notably DRAM/BRAM content, DCM's programmed multiplier/divisor - persists after GSR is applied. GSR is performed at startup and after successful rescan. To check if that actually happens I run john against a hash with setting 15 (that takes ~5s) and hear increased noise from the PSU (that adjust fan speed dependent on power consumption). If I immediately press Ctrl+C twice the noise level remains the same for ~5s until it finishes computation. But if I immediately run john again, it starts self-test and the noise reduces, increases again after self-test passes.
I paid attention for graceful handling of GSR in the design, it either initializes memory before read or reads read-only memory.

solardiz:
I ran another test: increased mask to 5 characters in both instances (2+2), increased `ZTEX_SCAN_INTERVAL_DEFAULT` from 15 to 150000 (effectively disabled). That way, I have two instances that run for a long time and do not rescan. Both of the concurrent invocations have completed in 2 hours without any errors! This confirms that the problems are only seen during startup or rescan (probably by the other instance than where the problems are seen).

Apingis:
I've reproduced the issue.

Host 1. While running one "long" session using 2 boards, at the same time repeating short-duration session (1 hash, --skip-self-test) using other board. Usually the short-duration session is unable to access 2 boards and accesses 3rd board, and every 50-100 runs it's also unable to access all 3 boards. Long-running session is not affected (tried over 5,000 short sessions for more than an hour). The issue is not reproducible when `ZTEX_SCAN_INTERVAL_TIMEOUT` is set very high.

Host 2. Shows effects as you described. The only case where 2 processes run well for a long time is when scanning disabled.

The main difference between above hosts is hardware. Host 1 has USB 3.0 and Host 2 has USB 2.0. While Host 1 provides USB 2.0 interface to devices internally it's different.

solardiz:
Are versions of libusb the same between the hosts? Could this be a libusb issue?

Regardless, maybe we can have a workaround? For example, a setting to disable the rescans without having to patch the source code. Maybe also ability to automatically resume use of temporarily-failed boards without a rescan as long as those boards haven't disappeared/reappeared (thus, are available to the running john process without a rescan), so that rescans would become less important? Alternatively, a mode where the rescans would only be performed after a board failed, but not periodically to discover potential new boards.

Apingis:
Updated libusb to the lastest version 1.0.21, everything remains the same.

That could be some issue (libusb, driver, OS or how they iteract). I think our usage (several processes access USB devices of same type) is not typical and might not be considered/tested well by system software developers.

Some solutions can be considered such as: 1) create custom OS driver; 2) have some userland dispatcher process; 3) use OS-specific methods such as read from /proc. Those are imperfect for usability but look like able to guarantee no collisions between several processes accessing devices of same type.

Thinking of our current code, rescan is the thing that greatly simplified program execution flow and overally reduced USB development effort (anything wrong - board goes into error state, becomes subject for the next rescan).

It might be the time to enhance existing codebase with more graceful handling of errors of different types and with more precise reflection of current board state in runtime data structures. That would allow to perform "full" scan on startup only.

PR #2766, opened 2017-09-18
Removed limit of (2^18-4) keys/board per crypt_all()

Apingis:
### Summary

Such a limit appeared because of 16-bit word(key) ID in word_list/template_list packet and internally in FPGA where IDs are stored/processed.
Now it just sends several packets where necessary.

### DIFF
--- a/src/ztex_bcrypt.c
+++ b/src/ztex_bcrypt.c
@@ -56,7 +56,7 @@ static struct device_bitstream bitstream = {
 	// computing performance estimation (in candidates per interval)
 	// (keys * mask_num_cand)/crypt_all_interval per jtr_device.
 	1,	// set by init()
-	262140,	// Absolute max. keys/crypt_all_interval for all devices.
+	1 * 1024*1024,	// Absolute max. keys/crypt_all_interval for all devices.
 	3,		// Max. number of entries in onboard comparator.
 	124,	// Min. number of keys for effective device utilization
 	1, { 141 },	// Programmable clocks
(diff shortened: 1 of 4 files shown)


src/ztex_bcrypt.c
```diff
@@ -56,7 +56,7 @@ static struct device_bitstream bitstream = {
 	// computing performance estimation (in candidates per interval)
 	// (keys * mask_num_cand)/crypt_all_interval per jtr_device.
 	1,	// set by init()
-	262140,	// Absolute max. keys/crypt_all_interval for all devices.
+	1 * 1024*1024,	// Absolute max. keys/crypt_all_interval for all devices.
```
solardiz: If max per device is 65535 and not 65536, should the maximum for all devices possibly be a multiple of 65535 rather than of 65536? Wouldn't a value as above, when we do bump into it, result in tiny last batches?

src/ztex_descrypt.c
```diff
@@ -54,7 +54,7 @@ struct device_bitstream bitstream = {
 	// (keys * mask_num_cand)/crypt_all_interval per jtr_device.
 	35 * 1024*1024,
 	// Absolute max. keys/crypt_all_interval for all devices.
-	262140,
+	2 * 1024*1024,
```
solardiz: Same concern as above applies here.

solardiz:
Thank you, Denis! I intend to test this after merging. Meanwhile, I've added a couple of comments on the hard-coded maximums - please let me know if you'd like to reply and possibly make adjustments before we merge this, or if we should merge first. I have no preference on this.

src/ztex_bcrypt.c
Apingis: This update removes maximum per device (was 65535, now can be any value).
The above value (1M keys for bcrypt, 2M keys for descrypt - that's template keys, can contain '#' for on-device mask) is because of limited USB bandwith.
That allows to set a cap on USB traffic per crypt_all so it don't run into timeout.

PR #2775, opened 2017-09-23
./john --list=ext-modes should only list standalone external modes

frank-dittrich:
I want to be able to use
```
for mode in `./john --list=ext-modes`; do ./john --external=$mode ...; done
```
without this change, I get 
```
No generate() for external mode: hybrid_example
```
etc.

Not sure whether these hybrid modes should be listed under ext-filters.
I think so, because they can filter input words just like regular filters.
And even if they don't, they can be used in addition to another mode which generates the candidates.

Please comment, and I'll adjust my pull request.

BTW: What's the intended difference between ext-filters and ext-filters-only?

Does a generator with a filter make sense?
I think an external mode which has both is somehow broken.

### DIFF
--- a/src/listconf.c
+++ b/src/listconf.c
@@ -455,7 +455,6 @@ void listconf_parse_late(void)
 	if (!strcasecmp(options.listconf, "ext-modes"))
 	{
 		cfg_print_subsections("List.External", "generate", NULL, 0);
-		cfg_print_subsections("List.External", "new", NULL, 0);
 		exit(EXIT_SUCCESS);
 	}
 	if (!strcasecmp(options.listconf, "ext-hybrids"))


magnumripper:
> BTW: What's the intended difference between ext-filters and ext-filters-only?
> 
> Does a generator with a filter make sense?
> I think an external mode which has both is somehow broken.

A generator with a filter does make sense. A filter isn't just a "reject or allow" function, it can be a case toggle or a ROT-13 or whatever.

frank-dittrich:
So, you approved and agreed we should list hybrids as filters as well.
Should I adjust my current commit?

> A generator with a filter does make sense. A filter isn't just a "reject or allow" function, it can be a case toggle or a ROT-13 or whatever.

But couldn't the generate() just do the filtering?
Currently, there is no external mode which has both generate() and filter().

$ ./john --list=ext-filters-only |wc -l
23
$ ./john --list=ext-filters |wc -l
23

magnumripper:
> So, you approved and agreed we should list hybrids as filters as well.
> Should I adjust my current commit?

Perhaps it's better not to list them as filters, since we can list hybrids separately. So just merge this and we're fine!

magnumripper:
> But couldn't the generate() just do the filtering?

Not necessarily. Until external mode got support for UTF-32 internally, the dumb16/dumb32 and repeats16/repeats32 external modes all used a filter that converted their UTF-32 representation to UTF-8.

PR #2803, opened 2017-10-01
doc/README-ZTEX; Simulation testbench cleaned up and extended

Apingis:
### Summary

- doc/README-ZTEX
- Simulation testbench cleaned up and extended: moved everything into 1 file; added more tests; verified in ISIM

### Other Information

Changes don't affect build process


doc/README-ZTEX
```diff
@@ -0,0 +1,80 @@
...
+It requires WinUSB driver to access the board.
+You can install the driver using Zadig 2.2 software.
+
+----------------------
+Format Specific Notes
+----------------------
+descrypt-ztex. That's a fast "format", USB 2.0 is unable to transfer
+password candidates from the host at the rate they are computed.
+Use mask mode to allow on-board candidate generation.
+Other feature is the limit of no more than 2,047 hashes per salt.
```
solardiz: Should add a note that mask mode may also be used along with another mode, with `?w`.

The limit isn't really a "feature". Better wording would be "Also note that there's a limit ..."

PR #2845, opened 2017-11-03
Raw-MD{4,5{,flat}}: handle uppercase hex digits

dnet:
Although it shouldn't matter in case of hexadecimal encoding, the Raw MD4, MD5 and MD5 flat format not only couldn't handle uppercase characters, but the error message was not really helpful. This issue was discussed in #2842.

### DIFF
--- a/src/rawMD5_fmt_plug.c
+++ b/src/rawMD5_fmt_plug.c
@@ -83,6 +83,7 @@ static struct fmt_tests tests[] = {
 	{"5a105e8b9d40e1329780d62ea2265d8a", "test1"},
 	{FORMAT_TAG "5a105e8b9d40e1329780d62ea2265d8a", "test1"},
 	{"098f6bcd4621d373cade4e832627b4f6", "test"},
+	{"098F6BCD4621D373CADE4E832627B4F6", "test"},
 	{FORMAT_TAG "378e2c4a07968da2eca692320136433d", "thatsworking"},
 	{FORMAT_TAG "8ad8757baa8564dc136c1e07507f4a98", "test3"},
 	{"d41d8cd98f00b204e9800998ecf8427e", ""},
@@ -177,7 +178,7 @@ static int valid(char *ciphertext, struct fmt_main *self)
 		p += TAG_LENGTH;
 
 	q = p;
-	while (atoi16l[ARCH_INDEX(*q)] != 0x7F)
+	while (atoi16[ARCH_INDEX(*q)] != 0x7F)
 		q++;
 	return !*q && q - p == CIPHERTEXT_LENGTH;
 }
@@ -191,6 +192,7 @@ static char *split(char *ciphertext, int index, struct fmt_main *self)
 		return ciphertext;
 
 	memcpy(out + TAG_LENGTH, ciphertext, CIPHERTEXT_LENGTH);
+	strlwr(&out[TAG_LENGTH]);
 
 	return out;
 }
--- a/src/rawMD5_fmt_plug.c
+++ b/src/rawMD5_fmt_plug.c
@@ -472,7 +472,7 @@ struct fmt_main fmt_rawMD5 = {
 #ifdef _OPENMP
 		FMT_OMP | FMT_OMP_BAD |
 #endif
-		FMT_CASE | FMT_8_BIT,
+		FMT_CASE | FMT_8_BIT | FMT_SPLIT_UNIFIES_CASE,
 		{ NULL },
 		{ FORMAT_TAG, FORMAT_TAG2 },
 		tests
--- a/src/rawMD5_fmt_plug.c
+++ b/src/rawMD5_fmt_plug.c
@@ -187,7 +187,8 @@ static char *split(char *ciphertext, int index, struct fmt_main *self)
 {
 	static char out[TAG_LENGTH + CIPHERTEXT_LENGTH + 1] = FORMAT_TAG;
 
-	if (!strncmp(ciphertext, FORMAT_TAG, TAG_LENGTH))
+	if (ciphertext[0] == '$' &&
+			!strncmp(ciphertext, FORMAT_TAG, TAG_LENGTH))
 		ciphertext += TAG_LENGTH;
 
 	memcpy(out + TAG_LENGTH, ciphertext, CIPHERTEXT_LENGTH);
--- a/src/rawMD5_fmt_plug.c
+++ b/src/rawMD5_fmt_plug.c
@@ -187,9 +187,8 @@ static char *split(char *ciphertext, int index, struct fmt_main *self)
 {
 	static char out[TAG_LENGTH + CIPHERTEXT_LENGTH + 1] = FORMAT_TAG;
 
-	if (ciphertext[0] == '$' &&
-	    !strncmp(ciphertext, FORMAT_TAG, TAG_LENGTH))
-		return ciphertext;
+	if (!strncmp(ciphertext, FORMAT_TAG, TAG_LENGTH))
+		ciphertext += TAG_LENGTH;
 
 	memcpy(out + TAG_LENGTH, ciphertext, CIPHERTEXT_LENGTH);
 	strlwr(&out[TAG_LENGTH]);
--- a/src/rawMD4_fmt_plug.c
+++ b/src/rawMD4_fmt_plug.c
@@ -167,7 +167,8 @@ static char *split(char *ciphertext, int index, struct fmt_main *self)
 {
 	static char out[TAG_LENGTH + CIPHERTEXT_LENGTH + 1];
 
-	if (!strncmp(ciphertext, FORMAT_TAG, TAG_LENGTH))
+	if (ciphertext[0] == '$' &&
+			!strncmp(ciphertext, FORMAT_TAG, TAG_LENGTH))
 		ciphertext += TAG_LENGTH;
 
 	memcpy(out, FORMAT_TAG, TAG_LENGTH);
--- a/src/rawMD4_fmt_plug.c
+++ b/src/rawMD4_fmt_plug.c
@@ -168,7 +168,7 @@ static char *split(char *ciphertext, int index, struct fmt_main *self)
 	static char out[TAG_LENGTH + CIPHERTEXT_LENGTH + 1];
 
 	if (!strncmp(ciphertext, FORMAT_TAG, TAG_LENGTH))
-		return ciphertext;
+		ciphertext += TAG_LENGTH;
 
 	memcpy(out, FORMAT_TAG, TAG_LENGTH);
 	memcpy(out + TAG_LENGTH, ciphertext, CIPHERTEXT_LENGTH + 1);
--- a/src/rawMD4_fmt_plug.c
+++ b/src/rawMD4_fmt_plug.c
@@ -81,6 +81,7 @@ static struct fmt_tests tests[] = {
 	{"8a9d093f14f8701df17732b2bb182c74", "password"},
 	{FORMAT_TAG "6d78785c44ea8dfa178748b245d8c3ae", "magnum" },
 	{"6d78785c44ea8dfa178748b245d8c3ae", "magnum" },
+	{"6D78785C44EA8DFA178748B245D8C3AE", "magnum" },
 	{FORMAT_TAG "31d6cfe0d16ae931b73c59d7e0c089c0", "" },
 	{FORMAT_TAG "934eb897904769085af8101ad9dabca2", "John the ripper" },
 	{FORMAT_TAG "cafbb81fb64d9dd286bc851c4c6e0d21", "lolcode" },
@@ -157,7 +158,7 @@ static int valid(char *ciphertext, struct fmt_main *self)
 		p += TAG_LENGTH;
 
 	q = p;
-	while (atoi16l[ARCH_INDEX(*q)] != 0x7F)
+	while (atoi16[ARCH_INDEX(*q)] != 0x7F)
 		q++;
 	return !*q && q - p == CIPHERTEXT_LENGTH;
 }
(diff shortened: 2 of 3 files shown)


magnumripper:
I have little time right now but on just a quick review I think the changes look sane - yet the build bots complain about things not ending up correctly.

@dnet you should probably verify your changes not only passes `-test` but also `-test-full=0`.

dnet:
It fails with...

```
Testing: Raw-MD4 [MD4 128/128 AVX 4x3]... FAILED (should not set FMT_SPLIT_UNIFIES_CASE (#4))
Testing: Raw-MD5 [MD5 128/128 AVX 4x3]... FAILED (should not set FMT_SPLIT_UNIFIES_CASE (#4))
```

...which I already saw at aa7e0e5, hence me following up with f804fda. However, I don't see what the problem is now. Why should it not set `FMT_SPLIT_UNIFIES_CASE`? `split` now clearly unifies case in the same manner as `Raw-MD5u`, with the same method (calling `strlwr` on `out` with the offset of `TAG_LENGTH`).

Looking at `formats.c` which throws this error, it seems to come up if `is_split_unifies_case_4` has the value of `2`, which can only be set in `test_fmt_split_unifies_case_4`, since its third parameter named `is_split_unifies_case` is the address of `is_split_unifies_case_4`. The value `2` is only written if `change_count` is zero.

However in `test_fmt_split_unifies_case_4`, **there's no write to `change_count`** apart from initializing it to `0` upon definition. So if `test_fmt_split_unifies_case_4` is called, it's practically **guaranteed to set `is_split_unifies_case_4` to `2`, resulting in this failure. But why?** Am I missing something here?

solardiz:
I don't know the detail of these tests, but I think they expect that if the flag is set, then `split()` always unifies case - whereas with your code that would be skipped when `ciphertext` already starts with the format tag. Try fixing that?

solardiz:
For example, `rawSHA256_common_plug.c: sha256_common_split()` does:

        if (!strncmp(ciphertext, HEX_TAG, HEX_TAG_LEN))
                ciphertext += HEX_TAG_LEN;

        memcpy(out, HEX_TAG, HEX_TAG_LEN);
        memcpy(out + HEX_TAG_LEN, ciphertext, HEX_CIPHERTEXT_LENGTH + 1);
        strlwr(out + HEX_TAG_LEN);

whereas you have an extra `return` statement before reaching `strlwr()`:

 	if (!strncmp(ciphertext, FORMAT_TAG, TAG_LENGTH))
 		return ciphertext;
 
 	memcpy(out, FORMAT_TAG, TAG_LENGTH);
 	memcpy(out + TAG_LENGTH, ciphertext, CIPHERTEXT_LENGTH + 1);
	strlwr(&out[TAG_LENGTH]);

dnet:
@solardiz thanks! It seems that this was the last thing I needed to make the tests pass.

src/rawMD5_fmt_plug.c
```diff
@@ -187,9 +187,8 @@ static char *split(char *ciphertext, int index, struct fmt_main *self)
 {
 	static char out[TAG_LENGTH + CIPHERTEXT_LENGTH + 1] = FORMAT_TAG;
 
-	if (ciphertext[0] == '$' &&
-	    !strncmp(ciphertext, FORMAT_TAG, TAG_LENGTH))
```
solardiz: IIRC, the performance improvement, while minimal, was non-negligible on multi-million hash files. So I'd prefer that we keep the separate first character check, and maybe introduce it to other `raw-md5*` formats as well. OK to add it to `raw-md4` too.

It is really unfortunate that the lower-casing itself will have a much bigger performance impact, though. But it looks like we'll have to accept it for now.

dnet: So... should I add it to all three files I touched within this PR?

solardiz:
As I wrote, "these tests [...] expect that if the flag is set, then `split()` always unifies case". I'll add to that: as a separate issue, we might want to look into this and possibly relax the tests (and then optimize the formats' `split()` methods accordingly). There's probably no actual need, but it has a performance impact, to always lower-case hex strings - even those already format-tagged, which are being read from `john.pot` - those are presumably already all-lowercase, and we can stipulate that our format tag implies that. Ideally, we'd have a way to communicate this knowledge to the tests (yet another format flag? or do we just exempt strings starting with '$' from this test?)

src/rawMD5_fmt_plug.c
solardiz: Please feel free to do that, or at least don't remove it from `raw-md5`, which is presumably the most frequently used one of the three.

dnet:
How about this? (566389e09765766cbbb6e90431f7bb0aa892ed27)

magnumripper:
Looks sweet now, I'm merging it.

frank-dittrich:
Wouldn't it be better to reject hashes with a prefix and uppercase hex, and only convert raw hashes without a prefix to lower case?

solardiz:
@frank-dittrich As I pointed out elsewhere (in comments on the PR perhaps?), yes that would be better (especially for performance reasons - faster processing of `john.pot`), but our `--test-full` is currently unprepared for that. Please feel free to open a separate issue for that.

frank-dittrich:
@solardiz --test-full does not cause problems if valid() rejects hashes with a format tag and uppercase hex.

The following patch works (passes --test-full=0)

```
diff --git a/src/rawMD4_fmt_plug.c b/src/rawMD4_fmt_plug.c
index fce8d4b..ead14b2 100644
--- a/src/rawMD4_fmt_plug.c
+++ b/src/rawMD4_fmt_plug.c
@@ -154,12 +154,17 @@ static int valid(char *ciphertext, struct fmt_main *self)
 	char *p, *q;
 
 	p = ciphertext;
-	if (!strncmp(p, FORMAT_TAG, TAG_LENGTH))
+	if (p[0] == '$' && !strncmp(p, FORMAT_TAG, TAG_LENGTH)) {
 		p += TAG_LENGTH;
-
-	q = p;
-	while (atoi16[ARCH_INDEX(*q)] != 0x7F)
-		q++;
+		q = p;
+		while (atoi16l[ARCH_INDEX(*q)] != 0x7F)
+			q++;
+	}
+	else {
+		q = p;
+		while (atoi16[ARCH_INDEX(*q)] != 0x7F)
+			q++;
+	}
 	return !*q && q - p == CIPHERTEXT_LENGTH;
 }
 
@@ -169,7 +174,7 @@ static char *split(char *ciphertext, int index, struct fmt_main *self)
 
 	if (ciphertext[0] == '$' &&
 			!strncmp(ciphertext, FORMAT_TAG, TAG_LENGTH))
-		ciphertext += TAG_LENGTH;
+		return ciphertext;
 
 	memcpy(out, FORMAT_TAG, TAG_LENGTH);
 	memcpy(out + TAG_LENGTH, ciphertext, CIPHERTEXT_LENGTH + 1);
```

solardiz:
@frank-dittrich Oh, that's good. Please feel free to implement such `valid()` logic for all relevant formats. Or would @dnet like to contribute this as well? :-) Thanks!

Also, we could speed things up by changing code like:

```
static char out [...];
[...]
memcpy(out, FORMAT_TAG, TAG_LENGTH);
memcpy(out + TAG_LENGTH, ciphertext, CIPHERTEXT_LENGTH + 1);
strlwr(&out[TAG_LENGTH]);
```

to something like:

```
static char out [...] = FORMAT_TAG;
[...]
memcpylwr(out + TAG_LENGTH, ciphertext, CIPHERTEXT_LENGTH + 1);
```

where we'd need to define `memcpylwr()` with the hybrid functionality in a single loop and without creating an intermediary copy. And we'd rely on the format tag just sitting there in the static return buffer (I hope no other code in jumbo overwrites those?) Performance would need to be actually tested e.g. on loading 30M hashes.

dnet:
@solardiz I implemented `memcpylwr()` in 839e008 on [branch `memcpylwr` of my fork](https://github.com/dnet/JohnTheRipper/tree/memcpylwr). Is this anything like you imagined? I only added `rawMD5`, but obviously I can apply it later to all the other places that has code like those above. Right now, I'll wait for your opinion.

Also, I'm interested in how you'd measure loading 30M hashes. Obviously, I can generate such input files, the thing I have less ideas for right now is how to accurately measure the time it takes to _load_ the hashes.

jfoug:
For 30M hashes, wall clock timings should suffice. Run things with code prior to any changes, wall clock time the amount of time it takes between when you issue the john command, and when you see output that cracking hash started.  Then wall clock time  after changes.

If times initially take 85s, and after takes 95s, then the changes appear to be not overly intrusive. If however, times initially took 85s, and now take 200s, then we have more to discuss. There has been a pretty large outlay of development time over the last few years optimizing load times. It would be a shame to lose the work done there, coding for something which really should be up to the user to make sure the input file is in right format.

dnet:
@jfoug thanks, I generated 30M random hashes with mixed upper and lowercase hexadecimal representations, and f5da7d1d2 (current `bleeding-jumbo`) took around 15s, while 839e0084e (my version) took around 12s.

jfoug:
please generate 30M candidates which are all lowercase also.  I would think that would cause older bleeding to load slower.   So if timings there jump to 25s, but your code still runs fast but does both, then we certainly have a winner as far as loading times go ;)

dnet:
That's what I did, the older bleeding was fed with 30M lowercase-only values, since it wouldn't accept mixed or uppercase MD5 hashes.

jfoug:
It sounds like Good test.  Gala Apples vs Honey-Crisp apples (but still apples to apples, lol)

dnet:
I split this into a separate PR for easier tracking: #2917

frank-dittrich:
Can you lease also measure loading 30m hashes which do already have a format tag and lower case hex?

Once without the "unify case" logic in split(), once with the "unify case" logic in split, once with rejecting hashes with a format tag and upper case hex in valid() plus skipping the "unify case" logic in split for hashes with a format tag, see the sample code in https://github.com/magnumripper/JohnTheRipper/pull/2845#issuecomment-343122052

dnet:
@frank-dittrich I understand that you want me to prepend `$dynamic_0$` to 30M hashes. But I don't really understand your list of test scenarios. Could you be a bit more specific/verbose about those?

frank-dittrich:
OK, scenario 1: Test loading the 30m hashes with commit 93197893786cb85391c53321010e848719b5d68e, i.e., just before your commit 09f644aaa760953c470303b1d3843d71c828ce68.

The second scenario is the one with your most recent patch (also with 30m hashes with format prefix and lower case hex.

The third version is with a slit() which just returns ciphertext if ciphertext already has the format pefix, and only converts raw hex into lower case.
To pass self test in this 3rd scenario, you also have to adjust valid(), so that it rejects hashes as invalid, if the hashes start with a format prefix, but contain mixed case hex instead of lower case hex, see the sample patch  I added 2 weeks ago: https://github.com/magnumripper/JohnTheRipper/pull/2845#issuecomment-343122052
So, the third scenario should have a valid() which verifies that hashes with a format prefix do have lower case hex, so that in this case split() can skip the copying and converting to lowercase.

This is useful, because all cracked hashes in the .pot file will have the format prefix and lower case hex. Plus, `./john --show=left --format=...` will also convert all hashes into the canonical representation (format prefix plus lower case hex).

dnet:
I ran all of these 5 times each by adding `exit(0)` into `john_loaded_counts` in `src/john.c` and running it using `time(1)`. The results are each the average of those 5 runs.

#### Scenario 1:

 - 931978937 with 30M lowercase hashes **without** format prefix: 10.44 s
 - 931978937 with 30M lowercase hashes **with** format prefix: 13.34 s

#### Scenario 2:

 - e4852ce77 with 30M lowercase hashes **without** format prefix: 11.23 s
 - e4852ce77 with 30M lowercase hashes **with** format prefix: 13.82s

#### Scenario 3:

```diff
--- a/src/rawMD5_fmt_plug.c
+++ b/src/rawMD5_fmt_plug.c
@@ -199,7 +199,8 @@ static char *split(char *ciphertext, int index, struct fmt_main *self)

        if (ciphertext[0] == '$' &&
                        !strncmp(ciphertext, FORMAT_TAG, TAG_LENGTH))
-               ciphertext += TAG_LENGTH;
+               return ciphertext;
```

 - 931978937 + early return with 30M lowercase hashes **without** format prefix: 11.21 s
 - 931978937 + early return with 30M lowercase hashes **with** format prefix: 13.55 s

solardiz:
> @solardiz I implemented memcpylwr() in 839e008 on branch memcpylwr of my fork. Is this anything like you imagined? [...] Right now, I'll wait for your opinion.

Yes, and I noticed you already addressed the various issues that you and others found. Thanks.

I also suggested that we optimize the `split()` functions not to copy the format tag into the output buffer each time, but this should probably be a separate pull request.

> Also, I'm interested in how you'd measure loading 30M hashes. Obviously, I can generate such input files, the thing I have less ideas for right now is how to accurately measure the time it takes to load the hashes.

Your approach with patching `exit(0);` into the code is fine. You could also have used an empty wordlist, or `--stdin` and input redirect from `/dev/null`.

For generating fake hashes, I use scripts like:

    perl -e 'use Digest::MD5 qw(md5_hex); for ($i = 0; $i < 1000000; $i++) { print md5_hex($i), "\n"; }'

    perl -e 'use Digest::SHA1 qw(sha1_hex); for ($i = 0; $i < 1000000; $i++) { print sha1_hex($i), "\n"; }'

dnet:
Since it didn't really matter in this case whether the hashes were real or just random `/[0-9a-f]{32}/` strings, I used this, which generated 30M "hashes" to `/tmp` (which is mounted as `tmpfs` on my box) in 5 seconds:

```c
#include <stdio.h>
#include <stdint.h>

#define HASHLEN 32

int main(int argc, char **argv) {
    FILE *output = fopen("/tmp/hashes.txt", "w"), *urandom = fopen("/dev/urandom", "r");
    char *hexes = "0123456789abcdef0123456789ABCDEF";
    uint8_t buf[HASHLEN];

    for (uint64_t i = 0; i < 30000000; i++) {
        fread(buf, HASHLEN, 1, urandom);
        fputs("$dynamic_0$", output); /* optional */
        for (uint8_t j = 0; j < HASHLEN; j++) {
            fputc(hexes[buf[j] % 32], output); /* s/32/16/ for lowercase-only */
        }
        fputc('\n', output);
    }

    fclose(output);
    fclose(urandom);
}
```

PR #2849, opened 2017-11-06
An OpenCL #define was defined in the wrong place

claudioandre:
It is not OpenCL related. That said, a 'core' file like cracker.c
cannot depend on any OpenCL (opencl_*.h) file.

### DIFF
--- a/src/opencl_mask.h
+++ b/src/opencl_mask.h
@@ -1,6 +1,16 @@
+/*
+ * OpenCL mask defines
+ *
+ * This file needs only one #define that is also needed by
+ * non-OpenCL mask routines. The 'define' was copied here
+ * to be used by Kernels ('*.cl') source code.
+ *
+ */
+
 #ifndef _JOHN_OPENCL_MASK_H
 #define _JOHN_OPENCL_MASK_H
 
+// See also mask.h.
 #define MASK_FMT_INT_PLHDR 		4
 
 #endif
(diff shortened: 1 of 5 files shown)


src/opencl_mask_extras.h
```diff
@@ -14,8 +14,6 @@
 #ifndef _JOHN_OPENCL_MASK_H
 #define _JOHN_OPENCL_MASK_H
```
magnumripper: I see a bug here: this should be `_JOHN_OPENCL_MASK_EXTRAS_H` or something, and definitely not same guard as in opencl_mask.h

claudioandre:
> Ideally we should define that macro in one single place. Not sure how to accomplish it though.

You'll need to copy a non opencl* header to run/kernels. Are you ready to this?

src/opencl_mask.h
```diff
@@ -1,3 +1,12 @@
...
+ * OpenCL mask defines
+ *
+ * This file needs only one #define that is also needed by
+ * non-OpenCL mask routines. The 'define' was copied here
+ * to be used by Kernels ('*.cl') source code.
+ *
+ */
+
 #ifndef _JOHN_OPENCL_MASK_H
 #define _JOHN_OPENCL_MASK_H
```
magnumripper: Insert after line 12
```diff
+ // See also mask.h.
```

PR #2876, opened 2017-11-11
Some format fixes (split() function ...), and improved format self test

frank-dittrich:
Enhanced format self test, making sure that hash formats with format tags do have a split() function which adds the format tag when processing raw hashes.
(Otherwise, raw hashes end up in the pot file).

Plus fixes for problems uncovered with the enhanced self test

Some formats added the format tag in binary(), but that's the wrong place.
Some formats listed the wrong tag in the format's signatures.

The first commit includes the fixes for failed tests, the second commit includes the enhanced self test.
(I want to avoid `git bisect` trouble.)

### DIFF
--- a/src/HDAA_fmt_plug.c
+++ b/src/HDAA_fmt_plug.c
@@ -82,8 +82,8 @@ static unsigned int omp_t = 1;
 
 #define SEPARATOR			'$'
 
-#define MAGIC				"$response$"
-#define MAGIC_LEN			(sizeof(MAGIC)-1)
+#define FORMAT_TAG				"$response$"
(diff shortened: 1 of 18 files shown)


magnumripper:
> The first commit includes the fixes for failed tests, the second commit includes the enhanced self test. (I want to avoid git bisect trouble.)

You are clever 👍

frank-dittrich:
I'll add another commit to cleanup 9 formats that didn't fail the enhanced self test.
These formats binary() functions checked whether or not the ciphertext has a format tag.
But since binary() is called with after split(), such a test is not necessary.

PR #2883, opened 2017-11-12
Proposal: increase MaxLen if MinLen is specified and it is > MaxLen

claudioandre:
If needed and allowed, increase MaxLen when a valid value of MinLen is specified. Warn otherwise.

### DIFF
--- a/src/inc.c
+++ b/src/inc.c
@@ -555,12 +555,19 @@ void do_incremental_crack(struct db_main *db, char *mode)
 #endif
 	}
 
+	if (options.req_minlength >= 0 && min_length > max_length &&
+	   (options.req_maxlength == 0 || maxlength_computed) &&
+	    options.req_minlength <= CHARSET_LENGTH) {
+		max_length = min_length;
+	}
+
 	if (min_length > max_length) {
 		log_event("! MinLen = %d exceeds MaxLen = %d",
 		    min_length, max_length);
 		if (john_main_process)
-			fprintf(stderr, "MinLen = %d exceeds MaxLen = %d\n",
-			    min_length, max_length);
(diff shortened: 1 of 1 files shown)


src/inc.c
```diff
@@ -555,12 +555,19 @@ void do_incremental_crack(struct db_main *db, char *mode)
 #endif
 	}
 
+	if (options.req_minlength >= 0 && min_length > max_length &&
```
claudioandre: `>` is better.

src/inc.c
```diff
@@ -555,12 +555,19 @@ void do_incremental_crack(struct db_main *db, char *mode)
...
+	   (options.req_maxlength == 0 || maxlength_computed) &&
+	    options.req_minlength <= CHARSET_LENGTH) {
+		max_length = min_length;
```
frank-dittrich: I wouldn't silently adjust max_length.
At least the user must be informed about such an adjustment (at default verbosity)

Just aborting with an error message telling the user what to do might be better.

claudioandre: It is fair, but, in my opinion, if I say `--min-len=X`, at least, I'm saying `--max-len=X` (it could be greater). 
* So, I dislike the idea to quit with error.

frank-dittrich: OTOH, I dislike wasting CPU time on something I didn't really intend to do.

So it all depends. If the min length vs. max length check occurs after john loaded millions of hashes, then automatically adjusting the max length to the smallest possible value might be preferable.

But when the max length is checked first, I would still prefer an error message.

Let's wait for opinions of other people.

magnumripper:
I think I agree with Claudio. If I said `-min-len=10` I obviously intended max length to be at least the same.

What's important is we never bump it beyond the **format's** max length though.

PR #2918, opened 2017-11-23
Move an #include to before its usage

claudioandre:


magnumripper:
If we put this fix in #2847 instead, it may save you a rebase of that PR.

claudioandre:
But you need this in bleeding ASAP. Can't wait that PR.

magnumripper:
#nitpick: This should ideally be a "git revert f8debd25" for clarity. Pretty cosmetical tho'.

PR #2946, opened 2017-12-01
Fix bash completion for ./john --show=types-json

frank-dittrich:
The old logic didn't work for ./john --show=types-[tab] ...


magnumripper:
Thanks, this had me scratch my head. I thought it worked because it would write out all alternatives, but in the end it would never really complete it.

PR #2957, opened 2017-12-02
Register dynamic formats last.

magnumripper:
We certainly don't want it to have precedence over optimized formats. The old comment about it needing to be registered early (bc of thin formats) does not seem to apply anymore.

### DIFF
--- a/src/john.c
+++ b/src/john.c
@@ -368,18 +368,19 @@ static void john_register_all(void)
 	john_register_one(&fmt_AFS);
 	john_register_one(&fmt_trip);
 
+	/* Add all plug-in formats */
+#include "fmt_registers.h"
+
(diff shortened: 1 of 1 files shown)


magnumripper:
@jfoug any last words before I merge this? I can't seem to find any problems with it. Here's the old comment that I believe no longer applies:
```c
       // NOTE, this MUST happen, before ANY format that links a 'thin' format
       // to dynamic.
       // Since gen(27) and gen(28) are MD5 and MD5a formats, we build the
       // generic format first
```
We now register thin format plugs before dynamic but there doesn't seem to be any problems.

jfoug:
I have NOT had a chance to review this (and am actually VERY busy), but if all the build bots are happy, then for whatever reason that comment was made in past, must not be required ATM, so if it builds, merge it.

PR #2960, opened 2017-12-04
Core CVS merge

magnumripper:
Dropped use of our custom int64 in favor of standard uint64_t


magnumripper:
Just letting the bots have a look before merging.

src/bench.c
```diff
@@ -214,24 +214,16 @@ char *benchmark_format(struct fmt_main *format, int salts,
...
-		sprintf(buffer, "%uK", cps_hi / 1000);
-	else
-	if (cps_hi >= 100)
-		sprintf(buffer, "%u", cps_hi);
-	else {
-		mul64by32(&tmp, 10);
-		cps_lo = div64by32lo(&tmp, time) % 10;
-		sprintf(buffer, "%u.%u", cps_hi, cps_lo);
+	unsigned int cps = crypts * clk_tck / time;
+	if (cps >= 1000000) {
```
jfoug: Is this a chance for overflow (the crypts*clk_tck ?  It was done as int64 in a prior life?  Should the cps be set to int64_t and the printf's changed to handle it?

magnumripper: Solar opted to do it this way, and it avoids the printf nightmare with PRIu64 and such crap so I'm happy with it.

jfoug: Well, the math looks to be done uint64_t, with the final truncation to unsigned.  Yep, it does avoid the printf hell you get into when moving outside int / unsigned world.

src/bench.c
```diff
@@ -623,10 +617,8 @@ void gather_results(struct bench_results *results)
 		MPI_SUM, 0, MPI_COMM_WORLD);
 	MPI_Reduce(&results->virtual, &combined.virtual, 1, MPI_LONG,
 		MPI_SUM, 0, MPI_COMM_WORLD);
-	MPI_Reduce(&results->crypts.lo, &combined.crypts.lo, 1, MPI_UNSIGNED,
-		MPI_SUM, 0, MPI_COMM_WORLD);
-	MPI_Reduce(&results->crypts.hi, &combined.crypts.hi, 1, MPI_UNSIGNED,
-		MPI_SUM, 0, MPI_COMM_WORLD);
+	MPI_Reduce(&results->crypts, &combined.crypts, 1,
+	           MPI_UNSIGNED_LONG_LONG, MPI_SUM, 0, MPI_COMM_WORLD);
```
magnumripper: Really old/odd MPI environments may not have MPI_UNSIGNED_LONG_LONG, it's optional. We'll see if that's a problem in real life. If it is, we'll just add it to autoconf and disable MPI.

solardiz: Why don't we simply keep the two `MPI_Reduce` lines using `& 0xffffffffU` and `>> 32`, respectively? Sounds like a cheap enough price for not relying on optional functionality of the MPI library.

src/cracker.c
```diff
@@ -996,24 +989,11 @@ static int crk_salt_loop(void)
...
-		          mask_int_cand.num_int_cand);
-#else
-		/* Safe for > 4G crypts per call */
-		int64 totcand;
-		mul32by32(&totcand, crk_key_index, mask_int_cand.num_int_cand);
-		add64to64(&status.cands, &totcand);
-#endif
-	}
+	if (done >= 0)
+		status.cands += crk_key_index * mask_int_cand.num_int_cand;
```
solardiz: I think this is a bug: we need explicit cast to `uint64_t` before the multiplication.

src/bench.c
magnumripper: Unless I'm confused it will (or may) produce incorrect result if we eg. have only two nodes, and their figures are just above and just below 0x100000000? We've had that problem until now though.

magnumripper: We could do so only when MPI_UNSIGNED_LONG_LONG isn't supported though. Let's see if this is a real problem or not.

src/wordlist.c
```diff
@@ -387,7 +387,9 @@ void wordlist_hybrid_fix_state(void)
 
 static double get_progress(void)
 {
-	int64_t pos, size;
+	struct stat file_stat;
+	long pos;
```
solardiz: This is a bug: `pos` must be `int64_t` or similar in jumbo. It's only `long` in core.

src/wordlist.c
```diff
@@ -405,12 +407,14 @@ static double get_progress(void)
...
 	} else {
+		if (fstat(fileno(word_file), &file_stat))
+			pexit("fstat");
 		pos = jtr_ftell64(word_file);
 		jtr_fseek64(word_file, 0, SEEK_END);
 		size = jtr_ftell64(word_file);
 		jtr_fseek64(word_file, pos, SEEK_SET);
 #if 0
-		fprintf(stderr, "pos="LLd"  size="LLd"  percent=%0.4f\n", (long long)pos, (long long)size, (100.0 * ((rule_number * (double)size) + pos) /(rule_count * (double)size)));
+		fprintf(stderr, "pos=%ld  size=%"PRIu64"  percent=%0.4f\n", (long long)pos, (long long)size, (100.0 * ((rule_number * (double)size) + pos) /(rule_count * (double)size)));
```
solardiz: Continuing the bug here, but also with inconsistent format specifier and typecast.

src/bench.c
solardiz: OK. I didn't realize this was fixing a maybe-real problem.

magnumripper:
Thanks, I'll look into it.

src/bench.c
solardiz: Maybe we have a very real problem in jumbo here for GPU/FPGA formats, where the c/s rate may reasonably exceed 4G. I merely didn't want to make unrelated changes along with the move from `int64` to `uint64_t`.

magnumripper:
Fixed in 8cf38fc0

src/bench.c
magnumripper: We are only looking at the original core diff here, not the merged Jumbo diff. I believe Jumbo still has it right.

magnumripper: $ git diff master bench.c
```diff
(...)
@@ -217,6 +591,13 @@ char *benchmark_format(struct fmt_main *format, int salts,
 void benchmark_cps(uint64_t crypts, clock_t time, char *buffer)
 {
        unsigned int cps = crypts * clk_tck / time;
+       uint64_t cpsl = crypts * clk_tck / time;
+
+       if (cpsl >= 1000000000000ULL) {
+               sprintf(buffer, "%uG", (uint32_t)(cpsl / 1000000000ULL));
+       } else if (cpsl >= 1000000000) {
+               sprintf(buffer, "%uM", (uint32_t)(cpsl / 1000000));
+       } else
        if (cps >= 1000000) {
                sprintf(buffer, "%uK", cps / 1000);
        } else if (cps >= 100) {
(...)
```
Jumbo adds support for M and G, with a non-conflicting diff as above.

src/bench.c
magnumripper: I opted to fix this in 9e02be954 before anyone complained: For systems lacking 64-bit reduce we fall back to old buggy behavior of adding halves without carry. Is there even any way to work around that problem? I mean, we have *n* nodes, potentially thousands.

Anyway it only affects --test under MPI so it's cosmetical.

Issue #2881, opened 2017-11-12
Reuse JtR runtime CPU detection code from configure script


Florianwall:
Can anyone help me? installing works just fine, but i cant execute any of the programs like zip2john etc.. always tells me that AVX2 is required for this build. I don't know how to fix this. I am currently running Kali linux on a virtual machine. Couldn't find any similar bugs.

solardiz:
This isn't a user support forum. Helping one particular person isn't the intent of GitHub issues, and thus not a priority when handling of issues reported in here. We have the john-users mailing list for that.

It is known that VMs sometimes incompletely support or proxy host's support for SIMD instruction sets, resulting in inconsistencies like this. A workaround might be to configure with `./configure --disable-native-tests`, although this is likely excessive (will also disable plain AVX, etc., resulting in a greater than necessary performance hit).

That said, thank you for your problem report. What we might be able to do to work around this in JtR jumbo for everyone (which is finally on-topic for a GitHub issue) is to make its configure script's and its runtime CPU detection checks more consistent (ideally, invoke the same code), so that either both will detect e.g. AVX2 or both won't.

We could also make `--disable-native-march` the default, since gcc's CPU detection is also different from ours and is also sometimes wrong in VMs, but per previous discussions @magnumripper does not want to do this (and it makes sense to me as this is how we currently make use of instructions such as MOVBE where available).

magnumripper:
@Florianwall when you say "installing" do you mean building Jumbo from (GitHub) source, or installing some Kali packet?

claudioandre:
https://github.com/claudioandre/JohnTheRipper/commit/e727e7a82de6e592f10723b260afca6847cb31d3

CI is still checking it. But:
* It is working on well (AVX2);
* It is working on Travis VMs (some AVX and some AVX2).

I have no idea what will happen on VC.

claudioandre:
It is not hard to Ctrl+c and Ctrl+v when a change is needed.

solardiz:
Thanks, Claudio. I commented on register clobbering, but more importantly I dislike us introducing (and then maintaining) an inline asm revision of the code. It's certainly not what I meant... although maybe it can help us for now, as a temporary hack.

I'd prefer that we build the same `.S` file from `configure` and later from `make`, and call the exact same `CPU_detect()` function. This might mean also building the actual crypto code found in the same source files, or alternatively we could split out the CPU detection into separate files.

[...]

Issue #2764, opened 2017-09-17
Inconsistent build on AVX-capable CPU with pre-AVX kernel


solardiz:
Trying to build on an AVX-capable CPU (in this case Ivy Bridge) with a pre-AVX Linux kernel (in this case RHEL5) results in an inconsistent binary not runnable on the system: it prints "Sorry, SSE4.1 is required" and exits. Examining `x86-64.o` reveals it actually checks for AVX. Examining the generated `Makefile` shows:

```
CFLAGS = -DAC_BUILT -march=native -msse4.1 -DJOHN_SSE4_1 $(CFLAGSX)
ASFLAGS = -DAC_BUILT  -DARCH_LITTLE_ENDIAN=1 -march=native -msse4.1 -DJOHN_SSE4_1 -c -fopenmp   -DHAVE_HT
```

Apparently, `-march=native` defines `__AVX__`, but our `-DJOHN_SSE4_1` takes priority in our `arch.h`.

`./configure --disable-native-march` produced a consistent SSE4.1 build usable on the system.

Can/should we somehow improve the situation? If we've detected SSE4.1 and are passing the explicit `-m` option for it, maybe we should not be also passing `-march=native`? Why do we?

magnumripper:
> Apparently, -march=native defines `__AVX__`

So that's a gcc bug. You can't reliably use AVX without kernel support for it, right?

> If we've detected SSE4.1 and are passing the explicit -m option for it, maybe we should not be also passing -march=native? Why do we?

Because (at least in theory), some future CPU feature could make wonders for JtR without us writing any explicit support for it. Without -march-native, it wouldn't be used.

I haven't decided what to do here, if anything. If you don't want -march=native you should tell ./configure so.

solardiz:
> So that's a gcc bug. You can't reliably use AVX without kernel support for it, right?

Yes. This also results in e.g. `php_mt_seed` non-working in a default build on systems like that.

However, we clearly have a JtR jumbo bug here as well: it fails to even correctly refer to what's required - complains SSE4.1 is required and that's why it'll refuse to run, but SSE4.1 is in fact available.

Besides, we should be able to workaround this specific gcc bug by omitting `-march=native` when our own tests detect SSE4.1 and don't detect AVX. Or we could be even more specific and only omit that option when our own tests (to be made more selective) detect that AVX is supported by the CPU but not enabled by the kernel. It's a matter of checking for a combination of bit flags, should be easy.

[...]

Issue #2716, opened 2017-09-02
Change the AVX2 detection in x86.S and x86-64.S to check for features we actually need.


magnumripper:
#2384 
https://github.com/magnumripper/JohnTheRipper/issues/2674#issuecomment-326740705

We should check for actual AVX2 support (and whatever else we need, if anything) and not features that merely suggest we have AVX2.

magnumripper:
@solardiz I recall I was actually hoping for you to add such test to John proper so I could just merge that 😉

magnumripper:
@kholia from https://github.com/magnumripper/JohnTheRipper/issues/2674:
> From x86-64.S,
>
> ```
> #define CF_AVX                  $0x1C000000 /* AVX + XSAVE + OSXSAVE */
> #define CF_AVX2                 $0x1C401000 /* the above + MOVBE + FMA3 */
> ```
>
> From commit f9338ccaeec94949a2e7f4a7cb54da9a1919750d,  "The AVX2 check is very quick and dirty, not using leaf 7 nor extended flags but it should be sufficient for a starter".

You fooled me there. That's not the latest commit to this file. The below is also there (but I forgot I already did this!):
```
#define C7_AVX2                 $0x00000128 /* BMI2 + AVX2 + BMI1 */
#define C7_AVX512F              $0x00010000
#define C7_AVX512BW             $0x40010000 /* AVX512BW + AVX512F */
```
Those are leaf 7 tests. Actually we do test **everything** that we should - and nothing else.

@solardiz  from https://github.com/magnumripper/JohnTheRipper/issues/2674:
> per Dhiru's comments above we actually know our check for AVX2 is wrong - it doesn't check for AVX2, but rather checks for MOVBE and FMA3, neither of which we use.

Dhiru's comment was incorrect, but the last part of what you say puzzles me. Of course we have to check for MOVBE and FMA3. How would we be sure some compiler doesn't emit them?! The tests I do are everything Intel says we have to check for and nothing else.

From [Intel spec](https://software.intel.com/en-us/articles/how-to-detect-new-instruction-support-in-the-4th-generation-intel-core-processor-family):
> Generally, compilers and libraries generating code for 4th generation Intel Core processors are expected and allowed to use all the instructions listed above, with the exception of Intel TSX. Below is the complete list of CPUID flags that generally must be checked:
> 
> CPUID.(EAX=01H, ECX=0H):ECX.FMA[bit 12]==1 &&
[...]

Issue #2679, opened 2017-08-16
CPU fallback might break in some builds.


magnumripper:
f7235fca13 broke CPU detection/fallback in certain cases. Jim put the CRC32_Init_tab() in john.c and it may end up containing eg. AVX2 instructions. See https://github.com/magnumripper/JohnTheRipper/issues/2674#issuecomment-322801462

solardiz:
Strictly speaking, jumbo's calling `sig_preinit()` and `path_init`() before CPU fallback is also a bug, until/unless we build `signals.c` and `path.c` without CPU type specific optimizations (like I think/hope we do for `john.c`, or don't we?) We're just "lucky" the compiler doesn't yet use (advanced) SIMD instructions in those functions.

I've just verified that none of this is in core, where `CPU_detect_or_fallback()` is reached without those external files' function calls. I don't know why we introduced/moved them in jumbo - I am not sure that was sound reasoning, but OTOH I don't rule out that core might have needed to do it for some reason I am unaware of. The core tree's CPU fallback works just fine in the Owl package.

solardiz:
> without CPU type specific optimizations (like I think/hope we do for john.c, or don't we?)

We do (or at least can) in core, due to these lines in `Makefile`:

```
john.o: john.c
        $(CC) $(CFLAGS_MAIN) $(OPT_NORMAL) $*.c
```

Owl's `john.spec` in some cases manages `CFLAGS_MAIN` separately.

claudioandre:
> Strictly speaking, jumbo's calling [...] path_init() before CPU fallback is also a bug

Jumbo uses `path_expand("$JOHN/" CPU_FALLBACK_BINARY)` inside `CPU_detect_or_fallback()`; so, if you do not do path_init() before it, you do not have $JOHN properly expanded and JTR will fail.

BTW: fallback works nicely on Linux (maybe also on Windows, I can't remember). On Linux, I've been packaging/testing it every week or so. Note, only for gcc.

magnumripper:
So we should build signals.o and path.o with same CFLAGS as john.o. This should be really trivial. Reopening for this.

magnumripper:
The move of sig_init (now sig_preinit) was caused by #798 where Frank came up with some pretty ridiculous test cases that I couldn't really rule out as not having any significance.

magnumripper:
> So we should build signals.o and path.o with same CFLAGS as john.o. This should be really trivial. 

[...]

PR #2963, opened 2017-12-04
Change run-time CPU detection for AVX2 to actually only check for AVX2

magnumripper:
...but neither of BMI1, BMI2, MOVBE and FMA3.

Change ./configure default to *not* use `-march=native` or the like.

Note: 32-bit x86 CPU detection still use the Quick'n'Dirty approach of inferring AVX2 from MOVBE+FMA3 (which is against recommendations but IMHO better than not testing at all).

See #2881, #2764, #2716, #2679, #2674, #2384 and a bunch of others.

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -1401,10 +1401,8 @@ Optional Features:
   --disable-rexgen        Disable use of rexgen library
   --disable-pcap          Do not build helpers depending on PCAP
   --disable-native-tests  Do not test build system for target features.
-                          Implies --disable-native-march.
-  --disable-native-march  Do not use things like -march=native even if valid
-                          compiler option(s). Native tests still can be run in
-                          this mode
+  --enable-native-march   Use things like -march=native if valid compiler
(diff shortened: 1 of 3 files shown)


magnumripper:
I'm actually not sure we want this but I think so (while waiting for a better solution).

@solardiz the more I look at these issues, the more I think you (core john) should have a more generic cpu_detect() function that returns a struct of uint32_t's (for different leafs) that the *caller* can compare with whatever it wants to. That would make future changes much easier (I have a hard time modifying the asm code without breaking things).

For example, our autoconf could also check for BMI (for `__builtin_bswapxx`), possibly adding `-mbmi`, and then we'd want to check for eg. AVX2 as well as BMI. But we simply can't expand the asm file for tens of combinations like that, it would end up as the legacy Makefile.

magnumripper:
Prehaps s/BMI/MOVBE/ but it was just an example

PR #2969, opened 2017-12-06
base64_decode replaced with base64_convert

jfoug:
Also a few optimizations to base64_convert, and some bug fixes.


jfoug:
Looks like the 32 bit mingw builds of keepass is not finding the aligned alloc crap.  Damn, I hate mingw :(   I will see what i can do.  I am pretty sure it is some missing #define or something like that in the new keepass build line,

magnumripper:
Travis CI:
```
opencl_salted_sha_fmt_plug.c:32:20: fatal error: base64.h: No such file or directory
 #include "base64.h"
                    ^
compilation terminated.
make[1]: *** [opencl_salted_sha_fmt_plug.o] Error 1
make: *** [default] Error 2
```

jfoug:
I am sticking a fork in this one.  If someone wants to rebase, with structure and enumeration name changes, that can be done at a later time.

This version has passed all tests, both simd and non-simd, gomp and non-gomp, and now the opencl.  I have also tested fully using ASAN and UBSan, and these changes, and the entire base64conv 'test bed' tests run with no problems.

src/CircleCI-MinGW.sh
```diff
@@ -180,7 +180,7 @@ echo "now building/testing a 32 bit SSE2 Linux build"
 echo '******************************************************************************'
 echo ""
 make -f Makefile.legacy -s clean
-JOHN_CFLAGS=-m32 JOHN_ASFLAGS=-m32 JOHN_LDFLAGS=-m32 make -f Makefile.legacy -sj4 linux-x86-sse2
+JOHN_CFLAGS=-m32 JOHN_ASFLAGS=-m32 JOHN_LDFLAGS=-m32 make -f Makefile.legacy -j4 linux-x86-sse2
 # do NOT exit on error from make.  We expect an error in the libpcap stuff
```
magnumripper: Did you really want the changes above to be permanent? If so, I don't mind - just asking.

frank-dittrich:
I think such a change is OK to temporarily find out what is going on.
If enabled permanently, it is easier to ignore (i.e., not detect) warnings.

jfoug:
I did.  When I had to investigate things having NO information provided no help.   I know Travis is also still using -sjX, and i think as soon as someone has to track into a bot build problem that is not superficial, information not shown would be important.

jfoug:
@frank-dittrich yes, warnings DO hide, without the -s  BUT information hides also ;)

It would really be nice if we could dump the CFLAGS/CPPFLAGS/LDFLAGS, etc in some manner, yet keep the '-s' part to help flush out warnings.   Lack of those variables was one big thing.  Also seeing which .o files is sometimes needed.  That is why I went away from the -s.  But if others see it as more important (to help flush out warnings), then so be it.

Hell, we could change the makefile to fail on warning.   Anyone up for that trial by fire ;)

jfoug:
For the dump of CFLAGS, we could make a target.  Then on that target, it ONLY does somehting if an env var is set  (like BOT_ENVIRONMENT)   Then we can set that env var on the bot builds.  Thus for normal builds, someone can use -s or not use it.  Then on the bots we have a target like:

```
# NOTE, not tested

ENV_DUMP:   /dev/null
     if test xBOT_ENVIRONMENT != "x" ; then
        echo "CFLAGS = $CFLAGS"
        echo "LDFLAGS = $LDFLAGS"
        # .....
     fi
```
and add that as a required target to the john build (or to any ancillary file build)    Do ppl like this at all, or even think it is feasible.

I know that using the -s has become the norm (to flush out warnings), but could it be time to force failure!?!? instead.

frank-dittrich:
If we are currently without build warnings, I am OK with -Werror, at least for normal builds which we expect to succeed without warnings.
As long as nobody looks at the build bot output, we wouldn't detect warnings, either.
For non-standard builds, people interested in build warnings could also search the build logs for `warning:` etc.

magnumripper:
We have warnings for things like "format not built because it needs SSE4.1" so that would explode in our faces.

frank-dittrich:
Shouldn't `-Werror -Wno-error=#warnings` work?

magnumripper:
Never heard of it but if the compiler supports it I guess that would be fine

frank-dittrich:
The `#` needs to be escaped.

`-Werror -Wno-error=\#warnings` works as intended. 

```
$ make -s
john.c:194:2: warning: Test Preprocessor Warning [-W#warnings]
#warning Test Preprocessor Warning
 ^
1 warning generated.

Make process completed.
```

Instead of
```
c$ export LC_ALL=C; make -s
john.c:194:2: error: Test Preprocessor Warning [-Werror,-W#warnings]
#warning Test Preprocessor Warning
 ^
1 error generated.
Makefile:1457: recipe for target 'john.o' failed
make[1]: *** [john.o] Error 1
Makefile:181: recipe for target 'default' failed
make: *** [default] Error 2
```
with just `-Werror`

magnumripper:
> The # needs to be escaped

Are you sure? That should only happen if it's preceded by a space.

frank-dittrich:
Without escaping, I get:
```
$ make -s
error: unknown -Werror warning specifier: '-Wno-error=' [-Werror,-Wunknown-warning-option]
Makefile:1455: recipe for target 'john.o' failed
make[1]: *** [john.o] Error 1
Makefile:181: recipe for target 'default' failed
make: *** [default] Error 2
```
for clang. (I got the came or a similar warning for gcc.)

But `-Werror -Wno-error=\#warnings` works for clang and gcc.

magnumripper:
That's odd. OTOH most things in compiler-land are odd so I'm not stunned.

magnumripper:
```
$ echo #hello

$ echo something#hello
something#hello
$ echo -Wno-error=#warnings
-Wno-error=#warnings
```

frank-dittrich:
I think it is a make oddity. vim also seems to think so, because without escaping, the end of line is turned into a comment.
Let's see make without `-s`:
```
c$ make
make find_version
make[1]: Entering directory '/home/fd/git/JtR/src'
echo "#define JTR_GIT_VERSION \"1.8.0.6-jumbo-1-3591-g954a837+\"" > version.h.new
diff >/dev/null 2>/dev/null version.h.new version.h && rm -f version.h.new || mv -f version.h.new version.h
make[1]: Leaving directory '/home/fd/git/JtR/src'
make[1]: Entering directory '/home/fd/git/JtR/src'
clang -DAC_BUILT -DJOHN_AVX -c -Werror -Wno-error= -funroll-loops -O0 john.c
error: unknown -Werror warning specifier: '-Wno-error=' [-Werror,-Wunknown-warning-option]
Makefile:1455: recipe for target 'john.o' failed
make[1]: *** [john.o] Error 1
make[1]: Leaving directory '/home/fd/git/JtR/src'
Makefile:181: recipe for target 'default' failed
make: *** [default] Error 2
```

claudioandre:
> We have warnings for things like "format not built because it needs SSE4.1" so that would explode in our faces.

It is a possibility in the future, of course. Anyway, you can enable -werror inside Travis with no problems (running it for months):
- https://travis-ci.org/claudioandre/JohnTheRipper/jobs/312577448

magnumripper:
I see Jim's latest such warnings were made with `#pragma message` instead. Perhaps they won't bail when using `-Werror`?
```c
#ifdef __GNUC__
#pragma message(": target system requires aligned memory access, rar format disabled:")
#elif _MSC_VER
#warning ": target system requires aligned memory access, rar format disabled:"
#endif
```
We need to establish when `#pragma message` was introduced though, just in case it's fairly new.

jfoug:
I actually got them reversed ;)   Usually I #warn on linux, but VC does not have that.  It has #pragma message().

gcc did not have the #pragma message in 4.1.2

clang 2.8 (but it may take a command line switch to work, not sure)
For gcc, a nice page, but lists no 'when it was added' section  : https://gcc.gnu.org/onlinedocs/gcc/Diagnostic-Pragmas.html#Diagnostic-Pragmas

magnumripper:
Crap, then we probably need to fix that. I was hoping for -Werror!

jfoug:
Ok, so we want to #warn on linux. I can easily reverse that logic  (I got it backwards anyway, and it would not have compiled if 'hit' on VC.  Vc has no #warn.

magnumripper:
Yeah just reverse it so it's like stribog_fmt_plug.c

magnumripper:
BTW you are not likely to need the _MSC_VER clause in RAR3 because AFAIK all Windows versions run on hardware that doesn't require alignment 😆

magnumripper:
OTOH you can keep that anyway just in case someone copy-pastes it for some other situation.

frank-dittrich:
I just tested the `#pragma message` notes with -Werror. No problem.

magnumripper:
So it's a matter of how old gcc's will handle it at all. I'm not normally running any old gear but some people may need to for whatever reason. @solardiz how far would you think we can go with a requirement of gcc version?

solardiz:
I think we should see whether using a given feature would bump up the gcc version requirement or not (maybe we already heavily depend on some other feature requiring at least same version), and whether the feature benefits us enough to go for it. For something as minor as this, I think we should not accept any increase in required gcc version (not worth it).

jfoug:
> BTW you are not likely to need the _MSC_VER clause in RAR3 because AFAIK all Windows versions run on hardware that doesn't require alignment

Yep, I fully understand. BUT putting out code which would not compile under VC, and then having that code copied to other places, is not ideal. It does happen.

As for #warn vs #pragma message, I agree with Solar, that it should not bump a compiler version requirement.  BUT the usaage of it IS much better, if it can be done.

It could be something like:

```c
#if _MSC_VER || __GCC_VER > xxx
#pragma message "Bugger off"
#elif defined __GCC__
#warn "Bugger off"
#endif
```

or some code like that.  This would allow closer movement to -Werror or whatever the command option is to force warnings to fail.

PR #2974, opened 2017-12-06
ppc64 and ppc32 ports.  With/without ALTIVEC code builds.  Some forma…

jfoug:
Pull request added, so Solar can easily test this also.

There are still formats failing.  I will be working through those.

However:

1. both 32 bit and 64 bit ppc now build with altivec (if the compiler allows)

2. neither 32 or 64 bit probe IF altivec builds 'work', just that the compiler allows  (TODO)

3. ppc32.h and ppc64.h are much close to core john versions. The call the *alt.h if using ALTIVEC

4. ppc32alt.h and ppc64alt.h have the SIMD_PARA defines in them (only used on jumbo)

5. At one time, I thought I had all formats working on altivec builds, but now, I am having 14 format failures.  I may have missed the format file porting logic.

6. The non-SIMD builds have not been tested yet (I need to find a good way to do that using configure).  I can always build using Makefile.legacy however until I figure it out  ;)

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -668,6 +668,7 @@ M_LIBS
 OPENSSL_LIBS
 OPENSSL_CFLAGS
 COMMONCRYPTO_LIBS
+SIMD_var
 EGREP
 PKG_CONFIG_LIBDIR
 PKG_CONFIG_PATH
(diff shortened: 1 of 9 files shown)


jfoug:
I figured this file was going to be a mess, lol.

jfoug:
Most of the format failures are fixed.  Problem was in the common-simd-sekey32[64].h code.  I had to handle the 0x80 byte setting for even length passwords (i.e. even mod4==0 length for 32 bit, mod8==0 length for 64 bit). 

I still have a problem with leet, but that is the last one.

jfoug:
Down to this on ppc32 (altivec)
```
ghost@local:~/JtR/bleed32/src$ ../run/john -test-full=0 | grep -v PASS
Will run 4 OpenMP threads
Testing: known_hosts, HashKnownHosts HMAC-SHA1 [SHA1 32/32]... (4xOMP) Warning: salt() returned misaligned pointer
Testing: leet [SHA-512(128/128 AltiVec 2x) + Whirlpool(OpenSSL/32)]... (4xOMP) FAILED (get_hash[0](2047) a8!=4e)

Testing: PKZIP [32/32]... (4xOMP) Warning: salt() returned misaligned pointer
```

PR #2975, opened 2017-12-06
Fixes for new CPU detection in ./configure, for non-Linux.

magnumripper:
This is not quite complete (circular dependencies) but probably works better than
current code.


magnumripper:
Circular dependencies hopefully fixed. Letting the bots do the rest now.

magnumripper:
OK, that failed miserably

PR #2985, opened 2017-12-09
Suggestion for code style

cpsw:
### Summary

I'm suggesting a code style change to avoid a possible dangling else in the future.

### DIFF
--- a/src/DOMINOSEC8_fmt_plug.c
+++ b/src/DOMINOSEC8_fmt_plug.c
@@ -541,22 +541,27 @@ static void decode(unsigned char *ascii_cipher, unsigned char *binary)
 				out <<= 6;
 				ch = *ascii_cipher;
 
-				if (ch < '0' || ch > '9')
-					if (ch < 'A' || ch > 'Z')
-						if (ch < 'a' || ch > 'z')
-							if (ch != '+')
-								if (ch == '/')
+				if (ch < '0' || ch > '9') {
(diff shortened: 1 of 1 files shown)


src/DOMINOSEC8_fmt_plug.c
```diff
@@ -541,22 +541,27 @@ static void decode(unsigned char *ascii_cipher, unsigned char *binary)
...
+							if (ch != '+') {
+								if (ch == '/') {
 									out += '?';
-								else
-								{ ; } /* shit happens */
-							else
+								} else {
+									/* shit happens */
+								}
+							} else {
```
magnumripper: I'm not sure the dropped `;` would go unnoticed with any compiler, doesn't some complain about empty braces? OTOH with brackets everywhere else we can drop lines 550-551 altogether because it's no longer gets ambiguous.

cpsw: I have not seen warnings about empty braces, but I can do the changes without removing the ; and putting everything in the same line: { ; } /* shit happens */. I submitted the patch because I would use the brackets in my code, but let me know if it makes sense according to the project's guidelines.

magnumripper: Like I said it's better to drop that else clause altogether. It was only there because it had to as long as no braces were used.

cpsw:
You are right, @magnumripper! Sorry for my delay to get your point.

PR #2987, opened 2017-12-09
configure:  add flag to disable SIMD logic in a build

jfoug:
configure flag  --disable-simd will fully disable SIMD logic in a build.

All CPU types doing SIMD are handled, with exception of MIC, which I am not sure how to handle without SIMD, or if we should even try.  All others  (any x86, PPC, ARM) have the SIMD logic removed and a base OpenSSL only ctx version of john built.

For x86_64 builds, I do leave __SSE__ enabled, since there are many compiler/environments which will not compile without that flag.  But inside John's code, we do not use that flag, so we are still effectively SIMD blind.


magnumripper:
Perhaps for this branch only, we could add `--disable-simd` to the bots' ./configure options, and once everything passes we can remove that again?

magnumripper:
> and once everything passes we can remove that again?

BTW our existing non-SIMD build(s) for build bots should definitely use this from now on, so they should be converted in this PR and obviously NOT reverted before merging!

magnumripper:
Also, a few of all checks should be kept without `--disable-simd` so we ensure a normal build didn't break.

magnumripper:
I did such changes, let's see how the bots cope with it.

jfoug:
> "AES-NI support ... run-time detection"

Is that during configure, or build, or at runtime?

jfoug:
> BTW our existing non-SIMD build(s) for build bots should definitely use this from now on, so they should be converted in this PR and obviously NOT reverted before merging!

that was the plan, but not until we have a 100% known working build product.

magnumripper:
>> BTW our existing non-SIMD build(s) for build bots should definitely use this from now on, so they should be converted in this PR and obviously NOT reverted before merging!
>
> that was the plan, but not until we have a 100% known working build product.

Not quite sure how you reasoned there. We do it in this PR (I already did it now) and will merge it when all is fine.

magnumripper:
>> "AES-NI support ... run-time detection"
>
> Is that during configure, or build, or at runtime?

That's during configure. I guess we could place some `#ifndef JOHN_SIMD_OFF` in some file(s) under the aes/ directory. If you find it out we can do it now, but we could also do it later.

magnumripper:
Hmmm wait. I think the AES-NI detection is actually what it says: **run-time detection**. So we never need to disable it, it will use AES-NI if present at run-time, but the very same build will not use it if it's not present. My bad.

jfoug:
Ok, so you want me to make short term changes to the CI script files, for testing.  I cna easily do that.

As for the errors listed in your first post, I have never seen anything like that.  You get that with a clean --disable-simd configure / make   ?   I have never seen tht.

jfoug:
I have the AES-NI detection covered.  Pretty simple to just put a if test "x$simd" = "xyes" block around the detection logic.  I see the right (using OSSL) line but have not 100% tested.  I need to also make damn sure that the makefile under eas gets built correctly.      right now, I am a bit short for time, so it may not get completed today (the CI bots, and checking the AES-NI logic also being off).  

Good catch, though on the EAS.  I was only think SIMD logic from x86.S and simd-intrins stuff.

jfoug:
Ok, just read your post on AES. But I do not think that is right, if we are wanting a SIMD free build. The reason being that we will still be building in the AES-NI AVX code (AVX I think).   Thus that is the bugaboo if we are targeting a non SIMD build.    

But it appears simple to turn off (at least the summary message states it is off), in the configure.ac file.  If testing holds, I will also add that to the final product.

jfoug:
Just had an idea.  An easy way to test this, is to set the default to 'no' for simd within configure.ac (just for now).  So we would get the option ```--enable-simd```  but we are only using that for build bots, and once happy with the end result, that default goes back to 'yes'

magnumripper:
> Just had an idea. An easy way to test this, is to set the default to 'no' for simd within configure.ac (just for now). So we would get the option --enable-simd but we are only using that for build bots, and once happy with the end result, that default goes back to 'yes'

Not sure what you did now. I already pushed commits that disabled simd for most builds. Now you disabled it for all builds, so we wont catch any introduced problems with SIMD.

magnumripper:
Anyway, even Linux build are failing now on the build bots so you've got work to do.

magnumripper:
> Ok, just read your post on AES. But I do not think that is right, if we are wanting a SIMD free build. The reason being that we will still be building in the AES-NI AVX code (AVX I think).   Thus that is the bugaboo if we are targeting a non SIMD build.    

Read this again: It is run-time detection. We can build this for anything and it will work on anything else.

magnumripper:
I believe undef'ing __SSE2__ and the like is a good method of getting into problems. You should probably only rely on `JOHN_SIMD_OFF` and not try to f@ck with compiler-added macros. That's just a guess though.

jfoug:
> I believe undef'ing SSE2 and the like is a good method of getting into problems. You should probably only rely on JOHN_SIMD_OFF and not try to f@ck with compiler-added macros. That's just a guess though.

That will require changes to 100's of lines of code, that USE those flags. i can try and remove those checks, especially now that -mnative is no longer getting in the way.  But what we may want to do, is to add -msse to the CFLAGS, and still have the ```undef __MMX__``` in the arch file.

jfoug:
I had not seen that you pushed the build bot scripts already.

magnumripper:
> I had not seen that you pushed the build bot scripts already.

Breath through nose, have a beer, enjoy life.

>> I believe undef'ing SSE2 and the like is a good method of getting into problems. You should probably only rely on JOHN_SIMD_OFF and not try to f@ck with compiler-added macros. That's just a guess though.
>
> But what we may want to do, is to add -msse to the CFLAGS, and still have the ```undef __MMX__``` in the arch file.

I **really** think that's a totally detrimental approach but that's just a gut feeling.

jfoug:
It may well be, but we have gotten ourselves into to this, but directly within our code, using __SSE2__  __SSSE31__   __SSE4__  etc

magnumripper:
> It may well be, but we have gotten ourselves into to this, but directly within our code, using `__SSE2__`  `__SSSE31__`  ` __SSE4__`  etc

That's a totally sane way of probing. FCS it's HOW YOU SHOULD PROBE FOR IT.

The really stupid shit here is what we are trying to do with this PR, IMHO. Things like this are solved with things like `-mno-avx` and so on. You and Solar don't get that so I'll leave it to you to solve the hundreds of problems you've now got ahead. Good luck with that.

magnumripper:
OTOH you could leave this to be a `./configure`-only thing. Do not probe for any SIMD. That I could buy.

magnumripper:
@solardiz I think we need some energy here 😆

solardiz:
I'm afraid I have no energy for this now - got a growing backlog (beyond JtR). My comments on easy stuff like `isdec*()` are quick and not really distracting from whatever else I am doing, I am already into the PPC64 build topic, whereas I haven't yet seriously looked into this issue and it'd take concentration to do so properly and to contribute positively.

magnumripper:
OK I'll try to be less negative. I think we simply *have* to stay away from undefining compiler-defined macros, not because I just have that idea but because it'll likely be easiest. If there are 100's of lines of code affected, tough shit but then we'll actually need to surround every instance of them with `JOHN_SIMD_OFF` or change them from eg. `#ifdef __that__` to `#if not defined JOHN_SIMD_OFF && defined __that__`.

jfoug:
Ok, how about this thought

If the --disable-simd switch is used, we make some changes to the CFLAGS, prior to calling any of the JTR_ARCH_SPECIAL_LOGIC functions.  So for any x86 build, we would add -mno-sse2 (ppc/arm will require other chagnes).  We also add -DJOHN_SIMD_DISABLE as a GCC/AS flag.  Then we fix any fallout in the formats required using ```#if !defined(JOHN_SIMD_DISABLE)```

This would remove all of the undef crap added to files like x86_64.h I have done. There still may be a sprinkling of the ```#if !defined(JOHN_SIMD_DISABLE)``` within them, but hopefully not much.  Then a few formats 'may' need a bit more nudging (blake, stribog come to mind), IF needed.

I think this sort of direction may be more of what you had in mind, and really it should end up being that a call to ```./configure --disable-simd``` really ends up much like a shortcut replacement for ```CFLAGS=-mno-sse2 ./configure``` in the end.

jfoug:
I am going to start a new branch (leaving this one intact), on my repo, and see how things go, and how easy it is to get this working in that manner.

solardiz:
I tried adding this comment via e-mail a couple of hours ago, but somehow it hasn't appeared yet, so:

> 69afa08  cracker.c  removed __SSE__ if disabling SIMD
> 6eaf2da  params.h  keep cracker prefetch from being set if __SSE__ is set

This is not ideal.  Ideally, we should be able to disable JtR's use of SIMD (I mean actual SIMD, not just prefetch instructions that happen to have been introduced along with first SSE on x86) while keeping the compiler's support for SIMD (and prefetch) intrinsics enabled so that uses like cracker.c's (which is not really a SIMD use) and maybe system headers' would continue working in non-SIMD builds.

jfoug:
@solardiz   I am looking into different overall usages.

the problems with cracker.c were we wrap the include of ```#include <xmmintrin.h>``` with ```__SSE___```, but inclusion of that header fails unless ```__SSE2___``` is defined, thus either we are including the wrong header, OR the intrinsic we want should only be used for an SSE2+ build.  I am not sure which, and am not continuing to chase it down.

solardiz:
> either we are including the wrong header, OR the intrinsic we want should only be used for an SSE2+ build

I think it's neither. This is an SSE intrinsic, thus available since Pentium 3 and usable in such 32-bit builds. However, x86-64 implies SSE2 support, so system headers (from application point of view, any of them!) are in their right to depend on SSE2 being enabled in the compiler when building for x86-64. We should accommodate that: be able to disable our use of SIMD while leaving it enabled in the compiler and preprocessor.

Moreover, I think recent OS X implies SSE4 support, so its system headers might require that to be enabled in the compiler & preprocessor as well.

solardiz:
> with exception of MIC, which I am not sure how to handle without SIMD, or if we should even try.

It should be possible to build for MIC without SIMD just like you do for other archs. In fact, our first builds for MIC were without SIMD and they worked fine (just slow). Whether we should even try is another matter; for consistency, that would be good to have, and I don't see MIC as any different from other archs in this respect.

jfoug:
For MIC, what would need changed in mic.h ?  It appears the entire file is SIMD logic. Is there version history of the original non-simd logic.  if so, I probaly can recreate it.  MY biggest concern, is I have no way to test, prior to checking code in, and we have no bots that build for it.  Thus it is double blind on that format.  That is why i simply pushed it off for now.  I do have ARM with neon capability (RPI3). I have not built on it yet with the --disable-simd logic, but will certainly do that prior to calling it complete.

solardiz:
For non-SIMD build on MIC, `DES_BS*` would need to be set the same as for a non-SIMD build on x86-64 (including even choice of the S-box expressions to use). For the jumbo-specific SIMD defines, I guess it's similar, but I'm not familiar with those. (BTW, all of those arch.h alikes say they're only copyrighted by me, but in jumbo they also have lots of defines I did not write.)

And you certainly can test this prior to checking in. You do have an account on super.

jfoug:
@magnumripper have you ever been able to get a non-SIMD build working with openCL ?  I believe that in the current rendition of john, to build openCL, you have to have SIMD enabled.

A non-OpenCL, that is easy.   I can build without simd easily.  But not so much with non-SIMD and opencl.  I get all sorts of failures once it starts including CL/CL.h  It includes intrinsic headers, and needs the defines.

@solardiz, it looks like if we want to enable a non-SIMD build of john, then we also have to force no OpenCL logic.  I am not sure that is what we are looking for in the end, but that is the way it is now.

That is also why TRAVIS is failing.  It is on OpenCL builds.

magnumripper:
> @magnumripper have you ever been able to get a non-SIMD build working with openCL ? I believe that in the current rendition of john, to build openCL, you have to have SIMD enabled.
>
> A non-OpenCL, that is easy. I can build without simd easily. But not so much with non-SIMD and opencl. I get all sorts of failures once it starts including CL/CL.h It includes intrinsic headers, and needs the defines.

I'm sure I've done that several times. Don't we even have such build on some of the build bots? However, I never got the idea to undef any compiler-defined macro like `__SSE2__`. That just doesn't seem right to me, just like redefining or undefining `__LP64__` would not make sense.

magnumripper:
I just successfully built non-SIMD w/ OpenCL, on macOS even. I had to drop this though:
```diff
diff --git a/src/simd-intrinsics.h b/src/simd-intrinsics.h
index 27afaa1d2..f322475e2 100644
--- a/src/simd-intrinsics.h
+++ b/src/simd-intrinsics.h
@@ -19,9 +19,6 @@
 #include "simd-intrinsics-load-flags.h"
 #include "aligned.h"
 
-#ifndef _EMMINTRIN_H_INCLUDED
-#define __m128i void
-#endif
 #define vtype void
 
 #define STRINGIZE2(s) #s
```
Those line are VERY old and I'm almost sure they are obsolete. Needs more testing before committing though.

magnumripper:
> Needs more testing before committing though.

I'm pretty sure you should commit that change to this branch right now and see if it doesn't actually fix a LOT of problems.

jfoug:
> However, I never got the idea to undef any compiler-defined macro like __SSE2__

Please drop this thought, it is NOT being done any more.  this PR is out of date, but I am not jamming my current trial up until I am pretty sure it is working, and will pass the bots.

The reason I am pretty sure opencl can not be built with ```CFLAGS=-mno-sse2 -mno-mmx```  is until recently I did not think you could do a -m64 non SIMD build at all (at least on cygwin).  I had tried before, but the build always fails.   BUT now that I am much more in touch with things, it is not the build.  it is the OpenCL part that is a no go.   A CPU only build is very easy to do with those cflags.   But on an opencl build, the --disable-simd (simply a mimic of the CFLAGS no simd stuff), works slick on the CPU formats, but then dies with undefined items on the openCL.  It is a classic cart before horse problem.

So here is my current recommendations (and what I  am working on in a personal branch)

1. add --disable-simd  flag to configure.ac

2. if in --disable-SIMD   we get a flag -DJOHN_SIMD_OFF set.

3. in x86 builds, it also adds to CFLAGSEX  -mno-sse2 -mno-mmx.  Then the code that does the SIMD testing will fail.

4. on ppc/arm builds, within the macros in m4 (m4/jtr_[ppc][arm].m4) the code that does the SIMD setting (and sets the -DJOHN_ARM/-DJOHN_PPC is not performed.

5. later in configure, when we test for openCL, if opencl is set, and --disable-simd is set, then configure bails with an explanation of simd being required for opencl, and if they really want a non-simd version, they have to also run configure with --disable-opencl

6. there will have to be wrapper code added to blake2, and to stribog that does not compile them (or add them into the fmt_*.h files), if -DJOHN_SIMD_OFF is set.

7. we may or may not need to change the arch.h files.  I am still working on this. I do have a version running and am testing.  I will be working on getting minimal patches, so I will be working to make no changes at all to the arch.h file.  Since arm and ppc both require flags TO get SIMD working, and we are not placing any into the CFLAGS, they should be fine.  As for x86 we are passing -mno-* flags, so again, I think they are fine.

8. the logic added to configure to keep AES simd logic may or may not be needed. When added, we get a 'using openssl' message listed, but i still need to make sure the code builds in pure ssl, and not using the SIMD optional stuff.

9. the prefetch stuff in bench/wordlist will stay intact, if it can be.

I just read your comment.  That is what was lacking. If that fixes things, then this PR (the next PR I do) will be much less complex.  I will test with that change, thank you. If I get this no-simd stuff working, I REALLY would like to see OpenCL still able to function.

jfoug:
> I tested like this:    ```./configure CFLAGS="-mno-mmx -mno-sse2 -g -O2"```

This is close to what the new version does. It will add the -DJOHN_SIMD_OFF  That flag may needed at lease in BLAKE2 and STRIBOG stuff.  

But I will put that patch into this PR, and also use it in the other cleaner branch I am working on.  I will not touch this branch (other than this change) until I am pretty damn sure I have things working right.  Hopefully, when done, there will be zero changes to any of the arch.h files.

jfoug:
Ok, I am now down to these errors:   (all the simd missing definitions are now gone)

```
In file included from formats.c:24:0:
common-opencl.h:118:2: error: unknown type name ‘hw_bus’
  hw_bus pci_info;
  ^
common-opencl.h:128:29: error: ‘MAX_GPU_DEVICES’ undeclared here (not in a function)
 extern cl_device_id devices[MAX_GPU_DEVICES];
                             ^
formats.c: In function ‘fmt_self_test_body’:
formats.c:857:4: warning: implicit declaration of function ‘advance_cursor’ [-Wimplicit-function-declaration]
    advance_cursor();
    ^
make[1]: *** [Makefile:1660: formats.o] Error 1
```

Any ideas?

jfoug:
Ok, on the branch I am working on, I had to add an include to "common-gpu.h" within common-opencl.h  (not sure why, but I did).  

After that change, the --disable-simd i have builds fine, with or without openCL (with by not having libs, or by forcing off opencl using --disable-opencl).  I will now work at testing this solution, AND to mimimize the changes.  

I can check things into this PR, or close this, create a new one, which will be a clean patch, and easier to see.  Whichever seems easiest for you to look at and critique.

NOTE, in the end, this will really just be a known working way of emulating  ```CFLAGS="-nno-sse2 -mno-mmx" ./configure [options]```  BUT it will not be limited to just x86, and the build issues such as opencl problems without sse2 will be worked out.  But there will be a simple flag, which users having problems building can be asked to try, to eliminate problems, especially where there are CPU's which have only limited porting done, and thus may or may not build (such as power, or when the next version of Intel's SIMD is released prior to us doing the porting needed).   I believe this was what what both Solar and i were hoping to achieve, nothing else.  Just a know working helper flag.   Yes @magnumripper we do understand that you are NOT happy, and would much rather see configure and john simply properly handle these machine differences.  I am not trying to stop that in the least.  I would just rather see something usable other than telling joe luser to revert back to Makefile.legacy and building a generic jumbo

magnumripper:
> I just successfully built (main bleeding-jumbo branch) non-SIMD w/ OpenCL, on macOS even

The same simd-intrinsics.h fix (with otherwise untouched bleeding-jumbo) and `-mno-mmx -mno-sse2` works fine for non-SIMD OpenCL on Linux too. I'm puzzled you had to change common-opencl.h.

> I can check things into this PR, or close this, create a new one, which will be a clean patch, and easier to see. Whichever seems easiest for you to look at and critique.

Whatever works for you. You could force-push to this branch and that'll be clean too. Or push a new branch and make a new PR for that one, no problem.

jfoug:
I am working on a clean branch.  I do not think there is going to be much lost from conversation, if this branch goes away.  In the end, if there is a clean branch, and we all say, 'yup that is clean, and works', then there likely is no reason at all for these comments about false starts.

jfoug:
NOTE the failures here, are likely do to the ```#undef __CPUTYPE___``` logic in x86-64.h   but I am not really interested in getting these failed starts 'fixed'. I would rather just push a good minimal representation of the ```CFLAGS=-mno-sse2 -mno-mmx ./configure``` version   I still will have a define for JOHN_SIMD_OFF, in case it is needed (it may be for ppc/arm or some later CPU type).  It does not take much change, so it will be left intact.  But other than that, this hopefuly will be a minimal change.

The key IMHO, was getting simd-intrinsics.h 3 line cutout in the mix.  I was stuggling with the opencl vs non catch-22 issues.  With that change, that situation is no longer a problem, and the simple CFLAGS forcing of simd off for x86 is happily working.

magnumripper:
Replaced by #2995

Issue #2862, opened 2017-11-09
Introduce configure options to enable/disable SIMD


solardiz:
We should introduce end-user friendly configure options to enable/disable SIMD (and have those options substitute the needed platform-specific compiler flags automatically).

Maybe also be able to choose which SIMD, on architectures that support many (like on x86: MMX, SSE*, AVX, XOP, AVX-2, AVX-512)? Of course, this can be done with custom `CFLAGS`, but the point of this issue is to make things simpler for end-users, and for ourselves too.

magnumripper:
Basically, I see two types of users:
1. J. Random Luser, who wants a build fairly optimized for the build host. He won't use this.
2. Mrs. Dist, who knows all about configure and CFLAGS. Neither will she.

solardiz:
What about "unnamed" users who would nevertheless be able to use `./configure --disable-simd` if the default SIMD-enabled build fails (and it is in fact more likely to fail) or if they need to test longer passwords (other than with hash types simple enough that Jim's command-line dynamic formats could be used for them, hopefully providing a high enough limit)? It's useful for us to be able to give such simple instructions in those cases.

It is also useful for us to be able to test such builds in a generic manner, without diving into platform-specific CFLAGS.

jfoug:
I kind of agree with Solar on this. It would be 'nice' to simply do a ```./configure --disable-simd && make -s clean && make -sj8``` without having to remember CFLAGS, or CPPFLAGS or whatever 'magic' is needed to build a non-SIMD build.

magnumripper:
Adding these options will introduce bugs and problems and confusion. If "the default SIMD-enabled build fails" we should put our efforts in fixing that, not working around it.

On the contrary, we should drop some of our present options IMHO. They just give me headache.

solardiz:
Adding these options will also make us fix bugs earlier, help us and users work around problems, and reduce confusion when we come up with and give instructions to make non-SIMD builds. Right now, I don't even know how to make a non-SIMD build for PPC64 in #2861. Sure it's some bugs we'll fix anyway, and SIMD will also work then, but it's useful to be able to quickly and easily test both kinds of builds.

While we should fix SIMD build bugs seen on "supported platforms" whenever we learn of them, "the default SIMD-enabled build fails" will be a recurring issue on uncommon and older platforms anyway.

solardiz:
[...]

PR #2991, opened 2017-12-11
minimal method of --disable-simd.  No change to source or headers (ex…

jfoug:
Ok, I think this is the minimal version.  It has no change to any john source/headers, with the exception of simd-intrinsics.h which was required to get OpenCL to build properly without SIMD.

This version disables searching for SIMD on arm/ppc.  on x86, it behaves almost exacty like CFLAGS="-mno-sse2 -mno-mmx" before calling configure.

There is a new define set when running in non-SIMD mode.  This is JOHN_SIMD_OFF   Right now, nothing seems to need this flag.  That may not always be the case, so its inclusion will tell anyone needing that information and allow them to be able to add #ifdef/#ifndef logic based upon this to make code work where SIMD logic is turned off.

There are 2 places in john code where we have left the SIMD logic intact (possibly 3).  These are on cracker.c / wordlist.c where we have prefetching.  Those are currently triggered off of __SSE__  Now, since we have -mno-mmx, both of these MAY be disabled already, I am not sure.  There is also the possibility that the AES will use SIMD, but again, since the -mno-mmx switch is used, I am pretty sure this is off also.  It does show it is off on the configure summary, but I am not absolutely sure how this works since it has always been runtime checked.

This version should be much cleaner than prior attempts.  also it should (hopefully) pass all build bots.

NOTE, this code is NOT merge ready.  we have made changes to the build bot scripts to force non-SIMD building.  Once everything is acceptable, we would certainy want to remove SOME of that logic. We should be using this logic wherever possible in the circle-CI, for doing non-simd builds. I.e. we would NOT revert some of that script, but keep it like it is checked in now.


jfoug:
#2862

.circle/CircleCI-MinGW.sh
```diff
@@ -176,7 +176,7 @@ echo ""
 echo ""
 echo ""
 echo '******************************************************************************'
-echo "now building/testing a 32 bit SSE2 Linux build"
+echo "now building/testing a 32 bit legacy SSE2 Linux build"
 echo '******************************************************************************'
 echo ""
```
magnumripper: All changes above should be kept when merging.

.travis.yml
```diff
@@ -55,5 +55,5 @@ addons:
       description: "http://openwall.com/john"
     notification_email: john.magnum@hushmail.com
     build_command_prepend:
-    build_command: "cd src && ./configure && make -sj4 && echo -e '[Disabled:Formats]\nRaw-SHA512-free-opencl = Y\nXSHA512-free-opencl = Y' > john-local.conf && ../run/john -test-full=0"
+    build_command: "cd src && ./configure --disable-simd && make -sj4 && echo -e '[Disabled:Formats]\nRaw-SHA512-free-opencl = Y\nXSHA512-free-opencl = Y' > john-local.conf && ../run/john -test-full=0"
```
magnumripper: The above should be reverted right before merging

.travis/check.sh
```diff
@@ -23,6 +27,9 @@ elif [[ -z "$TEST" || "$TEST" == "encoding" ]]; then
     if [[ "$OPENCL" == "yes" ]]; then
         sudo apt-get install fglrx-dev opencl-headers || true
 
+	# --disable-simd and opencl do not work together.  Turn off disable-simd mode.
```
magnumripper: This statement is no longer true, right?

src/configure.ac
```diff
@@ -69,6 +69,7 @@ AC_ARG_ENABLE([memdbg], [AC_HELP_STRING([--enable-memdbg], [Enable memdbg memory
 AC_ARG_ENABLE([asan], [AS_HELP_STRING([--enable-asan], [* Build with AddressSanitizer])], [asan=$enableval], [asan=no])
 AC_ARG_ENABLE([ubsan], [AS_HELP_STRING([--enable-ubsan], [* Build with UndefinedBehaviorSanitizer])], [ubsan=$enableval], [ubsan=no])
 AC_ARG_ENABLE([ubsantrap], [AS_HELP_STRING([--enable-ubsantrap], [* Build with UndefinedBehaviorSanitizer, crashes on failure])], [ubsantrap=$enableval], [ubsantrap=no])
+AC_ARG_ENABLE([simd], [AS_HELP_STRING([--disable-simd], [* Build forcing SIMD logic to be OFF])], [simd=$enableval], [simd=yes])
```
magnumripper: Super minor nitpick: I dislike saying "off" here and I also dislike over-use of capitals. IMHO this should say "Build forcing SIMD logic to be disabled".

src/configure.ac
```diff
@@ -402,6 +403,11 @@ dnl AC_CHECK_SIZEOF([int *function()]
 # See http://ftp.netbsd.org/pub/NetBSD/NetBSD-release-6/src/external/bsd/pcc/dist/pcc/configure.ac for many alternatives
 # *****************************************************************
 #
+if test "x$simd" = xno ; then
+[echo "SIMD logic... DISABLED by request"]
+AC_SUBST([JOHN_SIMD_OFF],["-DJOHN_SIMD_OFF"])
```
magnumripper: Actually I'd also rather have this as `JOHN_NO_SIMD` or `JOHN_DISABLE_SIMD` but I can live with this. You even accidentally used `JOHN_NO_SIMD` yourself in 69afa080 so I take it that name might be more logical for yourself too 😆

magnumripper:
I wonder what makes the Travis checks fail. I'll do some testing with this branch and get back!

magnumripper:
This (as of 07a2a997) works just fine as-is on my Macbook (latest everything, real gcc), including with OpenCL.

magnumripper:
I'll do some Ubuntu testing later. Hopefully I can reproduce the Travis bots' problems and find out how to fix them.

magnumripper:
> Now, since we have -mno-mmx, both of these MAY be disabled already, I am not sure

On systems I tested, -mno-mmx did not disable SSE (nor even SSE2 so that's why we do -mno on both of them)

.travis/check.sh
jfoug: True, but I think all of travis build scripts get reverted.

We 'could' keep it the way it is (scrap the comments and commented out stuff in the OpenCL), but keep the other, if we want.  then we simply comment out the export line at top of file, and a build will be done without this logic. Then if we need to test, simply uncomment that line and we have a version testing right off.

src/configure.ac
jfoug: I will make this fix.

src/configure.ac
jfoug: I agree that JOHN_NO_SIMD is probably the right choice.  it is easy to fix, and now that I think we have a better working version (although not done), I am not averse to fixing the nits in it.

jfoug:
NOTE, we do not disable sse.  BUT I do not think we have any usage of it, other than possibly in cracker/wordlist.

magnumripper:
> NOTE, we do not disable sse. BUT I do not think we have any usage of it, other than possibly in cracker/wordlist.

SSE is only in cracker.c and like Solar said it should not be disabled. So all is golden except for the mysterious fails on Travis CI.

magnumripper:
I see the problem (you can test it on well). Apparently the older xmmintrin.h used on Travis and well requires MMX support (which would normally be there if there is SSE support). So any such system we either need to add `-mno-sse` or remove `-mno-mmx`.

I propose we do the former, but only where actually needed. So we need an autoconf test that tests for SSE (specifically, for `_mm_prefetch()`) with `-mno-mmx -mno-sse2`. If it doesn't work, we need to add `-mno-sse`.

For now, I'm adding `-mno-sse` blindly (1062ca089), let's see if that fixes everything on Travis too.

magnumripper:
That did not work out well. Many system headers assumed SSE support. I reverted it with a force push.

So on systems like this we can't even disable MMX. Now MMX will not result in any SIMD intrinsics, just the old MMX assembler stuff - do we still have those in the tree at all?

magnumripper:
Trying a new commit that disable prefetch for non-SIMD builds.

magnumripper:
This works on well, but OpenCL is disabled as a result. Personally I think that's fine (since it gets disabled during configure - the build doesn't fail).

magnumripper:
Same happens on super: OpenCL gets disabled during configure, build is OK.

magnumripper:
Now the CircleCI build just seemed to have a hickup. @jfoug you need to trigger a rebuild there, I don't have permission.

Some Travis build actually works fine but like on well and super they end up with OpenCL disabled - and then they try to run `--test --format=opencl` which then fail. No real problem.

The clang builds however, fails with this:
```
fatal error: error in backend: SSE2 register return with SSE2 disabled
clang-3.9: error: clang frontend command failed with exit code 70 (use -v to see invocation)
clang version 3.9.0 (tags/RELEASE_390/final)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /usr/local/clang-3.9.0/bin
clang-3.9: note: diagnostic msg: PLEASE submit a bug report to http://llvm.org/bugs/ and include the crash backtrace, preprocessed source, and associated run script.
clang-3.9: note: diagnostic msg: 
********************
PLEASE ATTACH THE FOLLOWING FILES TO THE BUG REPORT:
Preprocessed source(s) and associated run script(s) are located at:
clang-3.9: note: diagnostic msg: /tmp/timer-5d1773.c
clang-3.9: note: diagnostic msg: /tmp/timer-5d1773.sh
clang-3.9: note: diagnostic msg: 
********************
make[1]: *** [timer.o] Error 70
make: *** [default] Error 2
```
So what's with timer.c/timer.h and clang? It looks like an ICE to me, perhaps just a compiler bug? Or maybe it simply can't understand how an x86-64 can lack SSE2?

magnumripper:
More specific
```
timer.c: In function 'sTimer_GetSecs':
timer.c:105:1: error: SSE register return with SSE disabled
 {
 ^
make[1]: *** [timer.o] Error 1
```

magnumripper:
With clang on well, we do this
```
clang -DAC_BUILT  -DJOHN_x86_64 -c -m64 -mno-mmx -mno-sse2 -I/usr/local/include -I/opt/AMDAPPSDK-3.0-0-Beta/include -DARCH_LITTLE_ENDIAN=1   -Wall -Wdeclaration-after-statement -fno-omit-frame-pointer -Wno-deprecated-declarations -Wformat-extra-args -Qunused-arguments -std=gnu89 -D_POSIX_SOURCE -D_GNU_SOURCE -D_XOPEN_SOURCE=600      -pthread -DHAVE_HT  -I/usr/local/include -I/opt/AMDAPPSDK-3.0-0-Beta/include -funroll-loops timer.c -o timer.o
```
...but the backend does this
```
 "/usr/bin/clang" -cc1 -triple x86_64-redhat-linux-gnu -emit-obj -mrelax-all -disable-free -disable-llvm-verifier -main-file-name timer.c -mrelocation-model static -mdisable-fp-elim -fmath-errno -masm-verbose -mconstructor-aliases -munwind-tables -fuse-init-array -target-cpu x86-64 -target-feature -mmx -target-feature -sse2 -target-linker-version 2.23.52.0.1 -D "AC_BUILT" -D "JOHN_x86_64" -D "ARCH_LITTLE_ENDIAN=1" -D "_POSIX_SOURCE" -D "_GNU_SOURCE" -D "_XOPEN_SOURCE=600" -D "HAVE_HT" -Wall -Wdeclaration-after-statement -Wno-deprecated-declarations -Wformat-extra-args -std=gnu89 -ferror-limit 19 -fmessage-length 100 -funroll-loops -pthread -mstackrealign -fobjc-runtime=gcc -fdiagnostics-show-option -fcolor-diagnostics -vectorize-slp -x c timer-e5ca91.c
```
Note the `-target-feature -mmx -target-feature -sse2` 😞

magnumripper:
Executive summary:  Some systems simply can't can't build non-SIMD with compiler flags (we can revert to SSE2 but not less) so we'd still need to make lots of use of JOHN_NO_SIMD.

Crap. I'm taking a nap, TTFN.

jfoug:
Sounds like we might have to find the right /arch:value to set things.  Yes, is sounds like /arch:pentum4 is the default (SSE2)

http://lists.llvm.org/pipermail/cfe-commits/Week-of-Mon-20150629/132028.html

src/m4/jtr_x86_logic.m4
```diff
@@ -34,6 +34,16 @@ dnl TODO: Ultimately we should not depend on any predefined stuff in arch.h
...
 AC_DEFUN([JTR_X86_SPECIAL_LOGIC], [
+if test "x$simd" = xno ; then
+if test "x$CC" != xgcc ; then
+echo --disable-simd only tested with gcc for Intel
+AC_SUBST([JOHN_NO_SIMD],[""])
+else
+AC_SUBST([JOHN_NO_SIMD],["-DJOHN_NO_SIMD -mno-sse2 -mno-mmx"])
+CFLAGS="-mno-sse2 -mno-mmx $CFLAGS"
+fi
+fi
```
magnumripper: First, you can't just test for `$CC != "gcc"` beause it may be eg. `/usr/bin/gcc` or `/usr/local/bin/gcc-5.2`.

Also, printing a warning is fine but we should definitely still obey the option in case of non-intel. Let the user try - in many cases it may work.

magnumripper: BTW on macOS native tools, /usr/bin/gcc is actually a clang so a name test is not enough anyway you do it.

jfoug: It did not work anyway.  Look at .Travis.    The failures for OpenCL gcc builds,fail in configure, saying it can find CL.h but it can not compile it.  Configure knows, but then the script goes ahead and calls for the gpu tests and the john process does not know about them, so the script fails.

jfoug: So that test was a total red herring

solardiz:
> Now MMX will not result in any SIMD intrinsics, just the old MMX assembler stuff - do we still have those in the tree at all?

FWIW, at least in the core tree enabling MMX (but not higher) and OpenMP will result in use of MMX intrinsics in `DES_bs_b.c`. It should be the same in jumbo.

solardiz:
> Executive summary: Some systems simply can't can't build 64-bit non-SIMD with compiler flags only

This is what I expected: SSE2 is part of x86-64, so systems may depend on it being enabled in the compiler. Probably similar issues exist for other archs where some SIMD instruction sets have been available for a long while.

jfoug:
> FWIW, at least in the core tree enabling MMX 

That was what I meant @solardiz   I knew  there was some core DES stuff still out there.  Just nothing using simd-intrinsics.c   I am not sure if the DES coding has ever had compilation problems.

magnumripper:
Replaced by #2995

PR #2997, opened 2017-12-13
fmt_split_case issue dyna_compiler and --test-full=1

jfoug:
This has been a LONG standing issue (in the test_fmt_split_unifies_case_3() function) which gave this error:

```
FAILED (should unify cases in split() and set FMT_SPLIT_UNIFIES_CASE (#3))
```

Due to the false positives, we have moved that logic from --test-full=0 and placed it into --test-full=1 so that the build bots would work.  Well with this fix, all of the false pos are gone. The one REAL problem (dyna_compiler) has been fixed, by making split deal with case.

The first patch does NOT move it back into --test-full=0.  I want to run on build bots first.  Once they are all happy, I will put a 2nd fix out that will move this logic from --test-full=1 and place it into --test-full=0 and make sure the build bots are happy still.  Once that happens we can merge.

### DIFF
--- a/src/dynamic_compiler.c
+++ b/src/dynamic_compiler.c
@@ -2398,17 +2398,28 @@ char *dynamic_compile_prepare(char *fld0, char *fld1) {
 }
 char *dynamic_compile_split(char *ct) {
 	extern int ldr_in_pot;
+	static char Buf[1024], Buf2[1024];
 	if (strncmp(ct, "dynamic_", 8)) {
-		return dynamic_compile_prepare("", ct);
+		char *cp;
+		ct = dynamic_compile_prepare("", ct);
+		cp = strrchr(ct, '@');
+		while (*cp) {
+			if (*cp >= 'A' && *cp <= 'Z') {
+				snprintf(Buf, sizeof(Buf), "%s", ct);
+				cp = strrchr(Buf, '@');
+				strlwr(cp);
+				return Buf;
+			}
+			++cp;
+		}
+		return ct;
 	} else if (strncmp(ct, "@dynamic=", 9) && strncmp(ct, dyna_signature, dyna_sig_len)) {
 		// convert back into dynamic= format
 		// Note we should probably ONLY do this on 'raw' hashes!
-		static char Buf[512];
 		snprintf(Buf, sizeof(Buf), "%s%s", dyna_signature, ct);
 		ct = Buf;
 	} else {
 		if (ldr_in_pot == 1 && !strncmp(ct, "@dynamic=", 9)) {
-			static char Buf[512], Buf2[512];
 			char *cp = strchr(&ct[1], '@');
 			if (cp) {
 				strnzcpy(Buf, &cp[1], sizeof(Buf));
--- a/src/dynamic_compiler.c
+++ b/src/dynamic_compiler.c
@@ -2398,28 +2398,17 @@ char *dynamic_compile_prepare(char *fld0, char *fld1) {
 }
 char *dynamic_compile_split(char *ct) {
 	extern int ldr_in_pot;
-	static char Buf[1024], Buf2[1024];
 	if (strncmp(ct, "dynamic_", 8)) {
-		char *cp;
-		ct = dynamic_compile_prepare("", ct);
-		cp = strrchr(ct, '@');
-		while (*cp) {
-			if (*cp >= 'A' && *cp <= 'Z') {
-				snprintf(Buf, sizeof(Buf), "%s", ct);
-				cp = strrchr(Buf, '@');
-				strlwr(cp);
-				return Buf;
-			}
-			++cp;
-		}
-		return ct;
+		return dynamic_compile_prepare("", ct);
 	} else if (strncmp(ct, "@dynamic=", 9) && strncmp(ct, dyna_signature, dyna_sig_len)) {
 		// convert back into dynamic= format
 		// Note we should probably ONLY do this on 'raw' hashes!
+		static char Buf[512];
 		snprintf(Buf, sizeof(Buf), "%s%s", dyna_signature, ct);
 		ct = Buf;
 	} else {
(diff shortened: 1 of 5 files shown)


jfoug:
Once the CI's are happy, and I see no warnings there, I will check this in.  That test which was under --test-full=1 is now under --test-full=0.

there WERE 2 formats that were split-unifies-case buggy. Both have been fixed.  One handles cased input (dynamic compiler). The other, palshop now returns 0 in valid() for wrong case.

The case-check-#3 function is also now fixed, and works under >= 0 conditions (not > 0 like before)

src/dynamic_compiler.c
```diff
@@ -2398,17 +2398,28 @@ char *dynamic_compile_prepare(char *fld0, char *fld1) {
...
+		while (*cp) {
+			if (*cp >= 'A' && *cp <= 'Z') {
+				snprintf(Buf, sizeof(Buf), "%s", ct);
+				cp = strrchr(Buf, '@');
+				strlwr(cp);
+				return Buf;
+			}
+			++cp;
+		}
+		return ct;
```
magnumripper: What is the above code doing? Will it not lower-case constants in the expression?

src/formats.c
```diff
@@ -1469,6 +1469,8 @@ static void test_fmt_split_unifies_case_3(struct fmt_main *format,
 
 		*has_change_case = 1;
 		split_ret = format->methods.split(cipher_copy, 0, format);
+		if (strcmp(cipher_copy, ciphertext))
+			*is_need_unify_case = 0;
```
frank-dittrich: If the formats.c changes are required, they should be a separate commit.
The dyna changes then should be a second commit, IMHO.

src/formats.c
```diff
@@ -1464,42 +1465,46 @@ static void test_fmt_split_unifies_case_3(struct fmt_main *format,
 	// Lower case
 	strlwr(cipher_copy + index);
 
-	if (strcmp(cipher_copy, ciphertext))
-	if (format->methods.valid(cipher_copy, format)) {
+	if (strcmp(cipher_copy, ciphertext)) {
+		if (format->methods.valid(cipher_copy, format)) {
```
frank-dittrich: May be @magnumripper didn't indent properly, to make merging core changes easier, I don't know.

Also, I wouldn't mix format.c changes and format specific changes into one commit.
Instead, the formats.c changes should come first, format specific changes in a separate commit.

jfoug: formats.c changes can not come first.  they will not pass bot checks without the changes to the format files.

jfoug: This code is not part of core.  Core has no concept of --test-full=x so there should be no merge conflict

src/dynamic_compiler.c
jfoug: You are right, but not constants.   The only thing it will screw up (I think) is salts.  Constants are p[art of the format, not the input hashes.  I could add a check for a '$' before the check for upper case, and if I find the '$', simply return ct.

I also see that Buf2 (earlier code) is not being used, and needs removed.

jfoug: Looking deeper (thanks for the poke), there is about no way to do this, and this code will have to be reverted.  It would kill things like:

```--format='MD5($p)'````

Also fixing valid will be equally as hard (for same reasons), !!!   damn.  

Looks like I will have to keep the code in the case-test-3 as a ```--test-full=1``` level, since I am now not sure I can get dyna expression format to pass that test.

jfoug:
Ok, the dyna compiler is not going to be easy to fix.  It has to handle valid inputs that lc hex, uc hex, base64, etc.  Also the prior patch would have stomped on salts, but if there were the only issue, it would have been trivial to fix.  But the UC/LC/mime issue is not as trivial, and will take more thought, since I do not want to parse the expression each time just to determine what to do on each hash input line.

So for now, I have fixed one of the 2 formats with this case split bug (palshop).  I have fixed the logic in the salt-case-test-3 which has been faulty since it was written.  BUT i have kept this test ```#3``` at --test-full=1 level, so that the build bots will pass.

Now, the only way (at the current time), to put it to level 0 (which is where it SHOULD be), is to put code into this test that ignores the known issue with dyna-compiler format, or to fix that format.

jfoug:
Do not merge yet.     I may be able to get dynamic valid working

jfoug:
Ok, I found/fixed dynamic's valid()   BUT this brings up a huge issue.  WHY was this test not triggering on every dynamic hash that was storing hex output???  There was actually a bad test string in dynamic.conf file (for 1010 RAdmin), that once the valid was fixed, failed valid.

BUT the test #3 was not catching ANY salt-case failures in dyna (AND IT SHOULD HAVE)  I will look a bit closer at that function (and the dynamic signature).

jfoug:
OK, I found out and fixed dynamic.  In dynamic, I have split() (handles any base64 hashes) and split_UC (unified case, not upper case).  I simply had to fix both of those, to handle the @dynamic=...@hash$salt type hashes, and then change dyna compiler format to call the parent split after doing the dyna-compiler special split logic.

So before, dyna was 100% handled, due to split or split_UC doing the proper case handling code (along with other canonicalization logic.).  But the test#3 was 100% correct, dyna was compliant, dyna compiler WAS not.  Now, dyna compiler calls into the proper split (post its own work in split), and piggybacks on dynamics 'proper' split setup (handles all cases).

I also reset --test-full=0 so that split-case#3 also in this commit.

jfoug:
This new build is fully passed, including the test#3 in the self-test --test-full=0 mode.

I think we are now golden.

src/formats.c
magnumripper: No this is Jumbo-only, originating from some GSoC task. Perhaps the original intention was to avoid indenting so much.

I agree that ideally this should be separate commits (although they can very well be in one same PR). Probably not important enough to require a rebase though.

Issue #2922, opened 2017-11-24
SIMD enable on some formats building under SIMD


jfoug:
The prior BE porting project 'skipped' some formats. These simply have the SIMD disabled when building under a BE SIMD machine.  This bug is to re-enable the SIMD logic so it works in these formats. NOTE, some of these are NOT trivial when it comes to SIMD usage, lol.   Dynamic format is handled all in a separate bug #2892, but is 100% related to this one.    But that beast is very ugly, and should be done with its own PR, since there will likely be hundreds of porting points in the code.

- [ ] hdaa
- [ ] sapb
- [ ] sapg
- [ ] saph   (and utf8)
- [ ] SunMD5
- [ ] wpapsk
- [ ] 7z    (and utf8)

jfoug:
Got sapG done.  NOTE, I see a LOT of performance improvement, on SIMD, doing sorted (by limb count) of the hashes within crypt_all  (like done in cryptsha256 and others)

jfoug:
WPAPSK was fixed with simply removing the inline pbkdf2 shit, and using the proper header include.

jfoug:
HDAA was a damn mess.  I am creating a PR to redo the SIMD totally.  That hash uses multiple SIMD limbs, of varying length.  Also it was WAY over optimized (to do a lot of the raw2hex shit), and deref a LOT of stuff using 32 bit values.  I think that format should be looked as as moving to the flat model, I think it would work as quick (OR QUICKER), and should have almost not BE porting code left in it.

An extra is it almost certainly will run faster.  Right now, we spin through all hash items, plucking off the done ones as we go. BUT we could use hash re-arrangement logic (like is done in sha256crypt), get blocks of all the same limb count, then just SIMD burn through them.

BUT that will be for a later (rainy) day.

PR #3005, opened 2017-12-14
Port of wpapsk to use proper pbkdf2_hmac-sha1.h file.  Works on BE

jfoug:
Win baby win.  Faster code, simpler (to read) code, works on LE or BE.

### DIFF
--- a/src/wpapsk_fmt_plug.c
+++ b/src/wpapsk_fmt_plug.c
@@ -21,15 +21,11 @@ john_register_one(&fmt_wpapsk);
 
 #include "arch.h"
 
-#if !ARCH_LITTLE_ENDIAN
-#undef SIMD_COEF_32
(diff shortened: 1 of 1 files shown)


jfoug:
#2922

jfoug:
On my testing, I was seeing about 5% to 10% improvements in performance!!!

Do we have other files (not the opencl stuff) where we have inline pbkd2.  If so, we really should look at removing it.    Now, we do have a few formats which still have BE porting, and it is possible one or more of them will be 'fixed' just like this one was.  I am pretty sure that 7z format uses pbkdf2, and is failing on BE. I have not looked to see, and will not have time for a bit (getting dragged out Christmas shopping, ugh)

jfoug:
Thats the easy way to port this one to BE.  Just do it with the simple header include stuff, and bam! fixed lol.

PR #3010, opened 2017-12-15
Fix for non SIMD builds

claudioandre:
In non SIMD builds, it is printing nothing where the CPU name is expected.

```
Target CPU ................................. , 64-bit LE
```

****
Well, to make things clear, maybe:
```
if test -n "$SIMD_NAME"; then
   CPU_NAME="$host_cpu $SIMD_NAME"
else
   CPU_NAME="$host_cpu no-SIMD"
fi
```

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -730,6 +730,7 @@ infodir
 docdir
 oldincludedir
 includedir
+runstatedir
 localstatedir
 sharedstatedir
 sysconfdir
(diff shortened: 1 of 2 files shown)


claudioandre:
BTW: OpenCL works with no SIMD. I have it inside Travis in case of doubts.

jfoug:
I am not sure of what version of autoconf you are using (or if there is some option enabled we have not enabled before), but the autoconf is spitting out a different set of options for ./configure than we have had before.

magnumripper:
To avoid an extra space IMHO it should be
```c
if test -n "$SIMD_NAME"; then
   CPU_NAME="$host_cpu $SIMD_NAME"
else
   CPU_NAME=$host_cpu
fi
```

> autoconf is spitting out a different set of options for ./configure

Exact same lines was added to me too, I just didn't add them to commit. No harm in doing so but once we add them, we need to avoid other people dropping them in later commits.

PR #3016, opened 2017-12-15
New cpuid

magnumripper:
Enhance the excellent tests @claudioandre fixed, by using proper AC macros instead of files and a "bastard compile" within configure. Also use the same logic for AVX512 (should work AFAIK).


magnumripper:
Somehow the AVX512 tests always end up as true so this fails on any AVX2 build. Need to find out some little detail that is wrong.

magnumripper:
@claudioandre I'm stumped. Using your test_SIMD manually, everything is correct. But same tests within AC always return 1 for AVX512.

src/m4/jtr_x86_logic.m4
```diff
@@ -244,13 +258,16 @@ if test "x$simd" = xyes; then
 
   AS_IF([test "x$CPU_NOTFOUND" = x0],
   [
-  CC="$CC_BACKUP -mavx2"
-  AC_MSG_CHECKING([for AVX2])
+  CFLAGS="$CFLAGS_BACKUP -mavx2 -P $EXTRA_AS_FLAGS $CPPFLAGS $CFLAGS_EXTRA $CPUID_ASM"
```
claudioandre: You missed the (e.g.) `-DCPU_REQ_AVX2 -DCPU_REQ`, no? It is required by the ASM code.

magnumripper: No, I dropped it because it's added by arch.h

magnumripper: Same command line works perfectly fine up to AVX2, then starts returning "true" for AVX512* even though I fixed mic.h

claudioandre: Thinking aloud:
- you need -DCPU_REQ_SSE2 when testing SSE2
- you need -DCPU_REQ_SSE3 when testing SSE3
- etc.

Are you sure it is changing as needed. If not, you are always testing SSE2.

magnumripper: That's not it. I'm running AVX. It returns true for SSSE3, SSE4.1, AVX and then returns false for XOP and AVX2. While debugging I let it continue trying AVX512* anyway - and then it returns true for both types of AVX512.

When doing the f***n exact same thing with test_SIMD.c, it returns false for AVX512.

claudioandre: That is interesting. I'll try to see if I can see a bug.

magnumripper: I pushed a new version but it only dropped the `#include`s becuase they're not needed.

magnumripper: I'm taking a nap, TTFN. Hopefully I'll know the answer when I wake up.

claudioandre: There is a typo in your code.

src/m4/jtr_x86_logic.m4
```diff
@@ -263,16 +270,16 @@ if test "x$simd" = xyes; then
...
-        #include <stdio.h>
-        extern void exit(int);
-        int main(){__m512i t, t1;*((long long*)&t)=1;t1=t;t=_mm512_mul_epi32(t1,t);if((*(long long*)&t)==88)printf(".");exit(0);}]]
-    )]
+  AC_RUN_IFELSE([AC_LANG_SOURCE(
+    [[extern int CPU_detect(void); extern char CPU_req_name[];
+      unsigned int nt_buffer8x[4], output8x[4];
+      int main(int argc, char **argv) { return !CPU_detect(); }
+    ]])],
     ,[CPU_BEST_FLAGS="-mavx512f"]dnl
```
claudioandre: Ops. Extra comma. Compare with AVX2.

```
checking for MMX... yes
checking for SSE2... yes
checking for SSSE3... no
checking for SSE4.1... no
checking for AVX... no
checking for XOP... no
checking for AVX2... no
checking for AVX512F... no
checking for AVX512BW... yes
```

magnumripper:
Wow I think I would never have found that one, thanks! M4 is a bastard of a language.

PR #3019, opened 2017-12-17
Handle a build outside git and archive

claudioandre:
I agree that seems unusual, but it can happen. The build will fail if there is any '%'.

The error is:
```
options.c:359:9: error: invalid conversion specifier ' ' [-Werror,-Wformat-invalid-specifier]

        printf(JOHN_USAGE, name);

               ^

options.c:301:20: note: expanded from macro 'JOHN_USAGE'

"John the Ripper " JTR_GIT_VERSION _MP_VERSION DEBUG_STRING MEMDBG_STRING ASAN_STRING UBSAN_STRING " [" JOHN_BLD "]\n" \

                   ^

./version.h:1:48: note: expanded from macro 'JTR_GIT_VERSION'

#define JTR_GIT_VERSION JOHN_VERSION "ormat:-%h %ci"

                                             ~~^

options.c:359:21: error: format specifies type 'int' but the argument has type 'char *' [-Werror,-Wformat]

        printf(JOHN_USAGE, name);

                           ^~~~

options.c:359:9: error: more '%' conversions than data arguments [-Werror,-Wformat]

        printf(JOHN_USAGE, name);

               ^

options.c:305:10: note: expanded from macro 'JOHN_USAGE'

"Usage: %s [OPTIONS] [PASSWORD-FILES]\n" \

        ~^

3 errors generated.

Makefile:1419: recipe for target 'options.o' failed

make[1]: *** [options.o] Error 1

Makefile:186: recipe for target 'default' failed

make: *** [default] Error 2
```

### DIFF
--- a/src/Makefile.in
+++ b/src/Makefile.in
@@ -42,12 +42,13 @@ subdirs = aes secp256k1 escrypt @ZTEX_SUBDIRS@
 top_srcdir = @top_srcdir@
 srcdir = @srcdir@
 prefix = @prefix@
+F = ""
 
 JTR_GIT_COMMIT = \"-$(shell git describe --dirty=+ --always 2>/dev/null)\"
 ifeq ($(JTR_GIT_COMMIT), \"-\")
 # this format string will be replaced by git archive:
 JTR_ARCHIVE_VERSION_STRING = $Format:-%h %ci$
-JTR_GIT_VERSION = JOHN_VERSION "\"$(JTR_ARCHIVE_VERSION_STRING)\""
+JTR_GIT_VERSION = JOHN_VERSION "\"$(shell echo "$(JTR_ARCHIVE_VERSION_STRING)" | "$(SED)" 's/ormat:-%h %ci/-/g')\""
 else
 JTR_GIT_HEAD_TS = \" $(shell git show -s --format=%ci HEAD 2>/dev/null)\"
 JTR_GIT_VERSION = JOHN_VERSION "$(JTR_GIT_COMMIT) $(JTR_GIT_HEAD_TS)"


frank-dittrich:
@claudioandre When I worked on this, I thought about moving src and run outside of the git repository and test this as well.
Then I thought: Who would ever do that.

But if you want to be that "paranoid", you have to add a line 
```
F = ""
```
to avoid this:
```
export F="test% % % Test" ; make -s distclean ; ./configure && make -s clean && make -s -j 16
```
resulting in
```
options.c: In function ‘print_usage’:
options.c:301:1: warning: conversion lacks type at end of format [-Wformat=]
 "John the Ripper " JTR_GIT_VERSION _MP_VERSION DEBUG_STRING MEMDBG_STRING ASAN_STRING UBSAN_STRING " [" JOHN_BLD "]\n" \
 ^
options.c:359:9: note: in expansion of macro ‘JOHN_USAGE’
  printf(JOHN_USAGE, name);
         ^
options.c:301:1: warning: conversion lacks type at end of format [-Wformat=]
 "John the Ripper " JTR_GIT_VERSION _MP_VERSION DEBUG_STRING MEMDBG_STRING ASAN_STRING UBSAN_STRING " [" JOHN_BLD "]\n" \
 ^
options.c:359:9: note: in expansion of macro ‘JOHN_USAGE’
  printf(JOHN_USAGE, name);
         ^
options.c:301:1: warning: unknown conversion type character ‘T’ in format [-Wformat=]
 "John the Ripper " JTR_GIT_VERSION _MP_VERSION DEBUG_STRING MEMDBG_STRING ASAN_STRING UBSAN_STRING " [" JOHN_BLD "]\n" \
 ^
options.c:359:9: note: in expansion of macro ‘JOHN_USAGE’
  printf(JOHN_USAGE, name);
```
and
```
$ ../run/john --list=build-info|head -n 1
Version: 1.8.0.12-jumbo-1-bleedingtest% % % Test-20171215
```
BTW: Why not 20171217?

magnumripper:
> The build will fail if there is any '%'.

Where? In the CWD? In the archive name?

frank-dittrich:
@magnumripper what about `s/20171217/unsupported/`?

frank-dittrich:
@magnumripper 
> Where? In the CWD? In the archive name?

In Makefile.in, when you are not inside a git repository and don't use an archive that has been created using `git arvchive`. (github download links for .tar.gz or .zip archives will point to files created using `git archive`.

Bat AFAIK, the build will not fail. Instead, you see some build warnings (possibly unwanted behavior of the build due to those warnings) and a strange version string, because of the `JTR_ARCHIVE_VERSION_STRING = $Format:-%h %ci$` line.

magnumripper:
> @magnumripper what about `s/20171217/unsupported/`?

Yes that would be better. Or even //.

>> Where? In the CWD? In the archive name?

> In Makefile.in, when you are not inside a git repository and don't use an archive that has been created using `git arvchive`. (github download links for .tar.gz or .zip archives will point to files created using `git archive`.

Oh. Since we'll have tar files at Openwall sooner or later, we really should fix this. Or will an official release look different? I can't remember, we don't release very often 🤣

frank-dittrich:
> Oh. Since we'll have tar files at Openwall sooner or later, we really should fix this. Or will an official release look different? I can't remember, we don't release very often

Why?

You can produce a working tar file by using `git archive` or just download the .tar.gz file from github.
This would substitute the format string in Makefile.in with abbreviated hash and commit time stamp, same as with the openwall's existing links to the latest development source code on github, because (obviously) github is using `git archive` to produce these files:
https://github.com/magnumripper/JohnTheRipper/archive/bleeding-jumbo.tar.gz
https://github.com/magnumripper/JohnTheRipper/archive/bleeding-jumbo.zip
(Bonus: the CI specific files and github specific files will be excluded from the archive)

And for releases, we do have a separate logic:
```

listconf.h:55:#if JTR_RELEASE_BUILD
listconf.h-56-#undef JTR_GIT_VERSION
listconf.h-57-#define JTR_GIT_VERSION JOHN_VERSION
listconf.h-58-#endif
```

frank-dittrich:
> Yes that would be better. Or even //.

Makes sense. When we see a version string which isn't an official release and doesn't have a date, we know the user did something strange. In that case, we can just ask the user to reproduce with latest bleeding jumbo.

PR #3045, opened 2017-12-22
NT-old: Honor the --disable-simd configure option. Closes #3044.

### LINKED ISSUE #3044: A few formats still use SIMD in --disable-simd builds
I believe this is the only hold out hash algo still using SIMD code when a non simd build is requested.

magnumripper:
Not sure if something is needed for 32-bit builds

### DIFF
--- a/src/x86-64.S
+++ b/src/x86-64.S
@@ -1741,6 +1741,8 @@ CPU_detect_fail:
 	ret
 #endif
 
+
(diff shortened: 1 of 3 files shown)


magnumripper:
Works fine on well `./configure CC="gcc -m32" --host=i686-linux-gnu --disable-simd`

PR #3048, opened 2017-12-23
idle: add Haiku support

alaviss:
### Summary

This commit add support for [Haiku][0], which is an BeOS-inspired operating system

[0]: https://www.haiku-os.org/

### DIFF
--- a/src/idle.c
+++ b/src/idle.c
@@ -37,7 +37,7 @@ static int use_yield = 0;
 extern int nice(int);
 #endif
 
-#ifdef __BEOS__
+#if defined(__BEOS__) || defined(__HAIKU__)
 #include <OS.h>
 #endif
 
@@ -84,7 +84,7 @@ void idle_init(struct fmt_main *format)
 #if defined(__MINGW32__) || defined (_MSC_VER)
 	SetPriorityClass(GetCurrentProcess(), IDLE_PRIORITY_CLASS);
 	SetThreadPriority(GetCurrentThread(), THREAD_PRIORITY_IDLE);
-#elif defined(__BEOS__)
+#elif defined(__BEOS__) || defined(__HAIKU__)
 	set_thread_priority(getpid(), 1);
 #else
 /*


kholia:
@alaviss What happens if this patch is not applied? Does the compile process fails?

magnumripper:
More files use the `__BEOS__` macro, I suspect more/all of them should get the same treatment?
```
$ git grep -l __BEOS__
bench.c
brg_endian.h
configure.ac
idle.c
os-autoconf.h
os.h
```

alaviss:
@magnumripper Haiku supports more POSIX features, which were disabled by those `#ifdef`s, so not all `__BEOS__` macro is required

@kholia The compile process will fail without the patch as Haiku doesn't provide POSIX [`nice(3)`](https://www.unix.com/man-page/posix/3p/nice/) function.

Also, I've already created a patch which let John the Ripper expand `~config/` and `~data/` dynamically to OS-specific folders. This is not strictly necessary for Haiku, it's only needed as part of their packaging guidelines (programs must save their user data to the correct folder, found dynamically with Haiku APIs), so I'm wondering if this should be upstreamed. The said changes could be found [here](https://github.com/alaviss/JohnTheRipper/tree/haikuDirs).

kholia:
Haiku's interface is really cool.

haiku-is-cool

kholia:
This patch improves things :+1: 

Compilation progresses further after applying this patch. However, I am getting stuck at the following compile error message,

```
$ ./configure
...

$ make
...
~/boot/system/develop/tools/bin/../lib/gcc/x86_64-unknown-haiku/5.4.0/../../../../x86_64-unknown-haiku/bin/ld: x86-64.o: relocation R_X86_64_PC32 against symbol `nt_buffer8x' can not be used when making a shared object; recompile with -fPIC
/boot/system/develop/tools/bin/../lib/gcc/x86_64-unknown-haiku/5.4.0/../../../../x86_64-unknown-haiku/bin/ld: final link failed: Bad value
collect2: error: ld returned 1 exit status
Makefile:1334: recipe for target '../run/john' failed
```

@alaviss Are you not running into this problem?

alaviss:
@kholia Looks like the error above only occurs on x86_64. I'm looking into it

PR #3054, opened 2017-12-25
Make some compilation warnings just a message

claudioandre:
The message is informational. It is neither a compilation warning nor an error.

### Summary

- Will allow to use -werror everywhere.
- Works on Linux and CygWin. Lets see how other OS will behave.
- gcc and clang.

### DIFF
--- a/src/compiler_message.h
+++ b/src/compiler_message.h
@@ -1,0 +1,25 @@
+/*
+ * This file is part of John the Ripper password cracker.
+ *
+ * Developed by Claudio André <claudioandre.br at gmail.com> in 2017
+ *
+ * Copyright (c) 2017 Claudio André <claudioandre.br at gmail.com>
+ * This program comes with ABSOLUTELY NO WARRANTY; express or implied.
+ *
+ * This is free software, and you are welcome to redistribute it
+ * under certain conditions; as expressed here
+ * http://www.gnu.org/licenses/gpl-2.0.html
+ *
+ */
+#ifndef _JTR_CC_MESSAGE_H
+#define _JTR_CC_MESSAGE_H
+
+#if __GNUC__ && GCC_VERSION >= 40201	// 4.2.1
+#pragma message JTR_CC_MESSAGE
+#elif _MSC_VER
+#pragma message(JTR_CC_MESSAGE)
+#else
+#warning JTR_CC_MESSAGE
+#endif
+
+#endif /* _JTR_CC_MESSAGE_H */
--- a/src/pkzip_fmt_plug.c
+++ b/src/pkzip_fmt_plug.c
@@ -1779,11 +1779,8 @@ struct fmt_main fmt_pkzip = {
 #else
 
 #if !defined(FMT_EXTERNS_H) && !defined(FMT_REGISTERS_H)
-#ifdef __GNUC__
-#warning pkzip format requires zlib to function. The format has been disabled
-#elif _MSC_VER
-#pragma message(": warning pkzip format requires zlib to function. The format has been disabled :")
-#endif
+#define JTR_CC_MESSAGE "Notice: pkzip format requires zlib to function; format disabled."
+#include "compiler_message.h"
 #endif
 
 #endif /* HAVE_LIBZ */
--- a/src/rar_fmt_plug.c
+++ b/src/rar_fmt_plug.c
@@ -438,11 +438,8 @@ struct fmt_main fmt_rar = {
 
 #else
 #if !defined(FMT_EXTERNS_H) && !defined(FMT_REGISTERS_H)
-#ifdef __GNUC__
-#warning ": target system requires aligned memory access, rar format disabled:"
-#elif _MSC_VER
-#pragma message(": target system requires aligned memory access, rar format disabled:")
-#endif
+#define JTR_CC_MESSAGE "Notice: target system requires aligned memory access; rar format disabled."
+#include "compiler_message.h"
 #endif
 
 #endif	/* ARCH_ALLOWS_UNALIGNED */
--- a/src/stribog_fmt_plug.c
+++ b/src/stribog_fmt_plug.c
@@ -472,10 +472,7 @@ struct fmt_main fmt_stribog_512 = {
 
 #else
 #if !defined(FMT_EXTERNS_H) && !defined(FMT_REGISTERS_H)
-#ifdef __GNUC__
-#warning Stribog-256 and Stribog-512 formats require SSE 4.1, formats disabled
-#elif _MSC_VER
-#pragma message(": warning Stribog-256 and Stribog-512 formats require SSE 4.1, formats disabled:")
-#endif
+#define JTR_CC_MESSAGE "Notice: Stribog-256 and Stribog-512 formats require SSE 4.1, formats disabled."
+#include "compiler_message.h"
 #endif
 #endif /* __SSE4_1__ */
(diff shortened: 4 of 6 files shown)


frank-dittrich:
Instead of introducing a new include and a new macro definition, we could also just use
`-Werror -Wno-error=\#warnings` in the Makefile, see https://github.com/magnumripper/JohnTheRipper/pull/2969#issuecomment-349790481

If we introduce a new macro, should we juse the JTR namespace, to avoid naming conflicts in the future?

magnumripper:
I can't find `-Wno-error=#warnings` at all but there is `-Wno-cpp` in my gcc (version 7.2.0) that does exactly what we want and nothing else, according to man page. I'm currently testing adding this to `./configure --enable-werror`.

magnumripper:
OK so `-Wno-cpp` would even mute the `#warning` messages but `Wno-error=cpp` works fine.

magnumripper:
`-Wno-error=cpp` Added in 6cb76bf3a

frank-dittrich:
BTW: Shouldn't `--enable-werror` be a "Development option", to prevent regular users from using it?
`* Treat warnings as errors`

magnumripper:
No

frank-dittrich:
Why not?
With your latest change, we suppress preprocessor warnings, so we might get support questions why the stribog formats are not working anymore, etc.

magnumripper:
`-Wno-error=cpp` works as we intended: Warnings are printed but not regarded as errors even with `-Werror`.

magnumripper:
@frank-dittrich I realize now that gcc wants `-Wno-error=cpp` while clang needs `-Wno-error=#warnings`. I have yet to find a way to add this to configure.ac:
```diff
diff --git a/src/configure.ac b/src/configure.ac
index 2eb29b975..b94e6a97e 100644
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -237,6 +237,7 @@ fi
    AS_IF([test "x$werror" = xyes],
       JTR_FLAG_CHECK([-Werror], 1)
       JTR_FLAG_CHECK([-Wno-error=cpp], 1)
+      JTR_FLAG_CHECK([-Wno-error=#warning], 1)
    )
    if test "x$asan" = xyes || test "x$ubsan" = xyes || test "x$ubsantrap" ; then
       JTR_FLAG_CHECK([-fno-omit-frame-pointer], 1)
```
The closest I get is enclosing the crap in double `[[` `]]`. That way, the flag check succeeds. The problem is it will not put `\#` in the Makefile so it will bugger off later.

I am currently ignoring this stupid problem.

jfoug:
For # characters, I am pretty sure you have to use ```[[```   ```]]``` to make it work. There are a few other things which I have found need escaping, but off top of my head, I do not remember (might be @ * $)

magnumripper:
Sure, it does work. The problem is we really need `\#` to end up in the Makefile... I have brute-forced tens of alternatives of syntaxes but nothing does both.

magnumripper:
I believe I tried all these and more
```
JTR_FLAG_CHECK([-Wno-error=#warning], 1)
JTR_FLAG_CHECK([[-Wno-error=#warning]], 1)
JTR_FLAG_CHECK([-Wno-error=\#warning], 1)
JTR_FLAG_CHECK([[-Wno-error=\#warning]], 1)
JTR_FLAG_CHECK([-Wno-error=\\#warning], 1)
JTR_FLAG_CHECK([[-Wno-error=\\#warning]], 1)
JTR_FLAG_CHECK("-Wno-error=\\#warning", 1)
```

jfoug:
Check these
```
JTR_FLAG_CHECK([[-Wno-error=\\\\#warning]], 1)
JTR_FLAG_CHECK([[-Wno-error=\\\#warning]], 1)
```

jfoug:
Can you get the flag to check?   If so, then hard code this into Makefile.in

```
WARNING_HACK=-Werror=\#warning
.if SOME_SETTING
CFLAGS=$CFLAGS $WARNING_HACK
.endif
```

note syntax may not be right, but I think you get the point.  Then check the flag. If the flag is OK, then define SOME_SETTING to trigger the make to include that warning.  ugly, yes, but if you are able to check the flag (just not get it into proper output to dump into makefile), then that work around would do the trick.

magnumripper:
Yeah I see what you mean. I had this thought too but right now I don't have the energy. How did they get the idea to use a compiler flag that is not compatible with gmake?! 😠

magnumripper:
IIRC, this gets the flag check right: `JTR_FLAG_CHECK([[-Wno-error=#warning]], 1)`. After that, we just need to add a `\` before the `#` in the CFLAGS_EX -  but that one isn't empty to start with unless we do this a little differently. This isn't impossible at all, it's just a matter of persistance 🙄

jfoug:
Non standard shit in M4 is just that, ugly and trial and error at times.  For most of the things, M4 is not that bad.  But when it gets in the way, it really gets in the way.

claudioandre:
> I realize now that gcc wants -Wno-error=cpp while clang needs -Wno-error=#warnings. 

No exactly. It works on clang 5 and fails on clang 4.

magnumripper:
Does anything work with clang 4? I mean, does `-Wno-error=cpp` work for example? I presume it doesn't.

frank-dittrich:
@magnumripper 
`-Wno-error=#warning` is wrong. You forgot the trailing `s`.
See the preprocessor warning output:
```
john.c:197:2: warning: test [-W#warnings]
#warning test
```
Each warning will be printed together with the switch which enables or disables this warning.
For preprocessor warnings, it is ` [-W#warnings]`.

That's why, the general `-Wno-error=switch` needs `-Wno-error=#warnings` here.
The additional `\` is only needed because make treats an unescaped `#` in a Makefile as the start of a comment.
(May be that's a bug in `make`, not sure.)

magnumripper:
I did try it with an s in the end too. I got to a stage where I detected the option correctly,  but that didn't end up with a backslash in the Makefile so it would blow up anyway.

magnumripper:
This is the closest I get
```diff
diff --git a/src/configure.ac b/src/configure.ac
index 2eb29b975..d1ef840de 100644
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -237,6 +237,8 @@ fi
    AS_IF([test "x$werror" = xyes],
       JTR_FLAG_CHECK([-Werror], 1)
       JTR_FLAG_CHECK([-Wno-error=cpp], 1)
+      JTR_FLAG_CHECK([[-Wno-error=#warnings]], 1)
+      [CFLAGS_EX=$(echo "$CFLAGS_EX" | sed s/#warn/\\\\#warn/)]
    )
    if test "x$asan" = xyes || test "x$ubsan" = xyes || test "x$ubsantrap" ; then
       JTR_FLAG_CHECK([-fno-omit-frame-pointer], 1)
```
I have tried it with, like zero to a billion back slashes. I give up here. M4 is the one thing I have ever met that is worse than Microsoft shite, and it's 100x worse.

frank-dittrich:
@magnumripper 
OK, a crazy idea:
What if we define a separate makefile to add these options to CFLAGS, and optionally include that other makefile into the main Makefile?

magnumripper:
Solved it in a37f10abe. I had to escape it **only** for when writing to Makefile as the escaping would not work in rest of ./configure.

Now, clang 4 and clang 5 works fine, as well as the clang shipped with macOS, "Apple LLVM version 9.0.0 (clang-900.0.39.2)".

PR #3065, opened 2017-12-27
3limb salt handling for all 32 bit opencl pbkdf2-hmac-* #3035

jfoug:
All changes in 1 commit. Rebased with new -Werror building logic.


magnumripper:
Good Stuff[tm]

jfoug:
Thanks. Took a while to get all ducks to line up.   I still think we might want to try to share headers between CPU/GPU for lengths, it may help, now that they all can handle unlimited length.

NOTE, mscash2 opencl probably should be looked at, since it is using oddball pbkdf2 code.  (I have not looked for others).

Issue #2858, opened 2017-11-09
Auto-tune instead of OMP_SCALE


magnumripper:
I sometimes see my laptop core i7 wanting a different OMP_SCALE than a desktop one. Not sure but it might be due to cache sizes. There's even more difference between different CPU types - some AMD CPU may want a completely different OMP_SCALE than an Intel.

Perhaps we could ditch the idea of OMP_SCALE and instead introduce trivial shared code that does something similar to OpenCL auto-tune, but much simpler and faster - just trying different numbers of keys-per-crypt.

magnumripper:
The format "interface" should be something like
```c
#include "omp_autotune.h"

(...)

static void init(struct fmt_main *self)
{
        omp_autotune(self);
        (...)
}
```

solardiz:
Optimal `OMP_SCALE` may also vary by OS, kernel version, OpenMP implementation/version, OpenMP env vars, and other system load (ideally, none). But I think auto-tuning it won't work well enough to bother, will slow down startup, and will result in even more inconsistent speeds than we're seeing now.

Instead, my tentative plan is to try and find time to rework the way we use OpenMP. This is related to making the formats interface multi-block and async, which is also needed for more efficient and user-friendlier (lower max_keys_per_crypt) support for offloading to external devices such as FPGA boards. Once those changes are in place, our use of OpenMP could be moved from formats to `cracker.c`, making OpenMP efficiently usable even for the fastest hashes. Now, whether I do find time any time soon (certainly not this year, and certainly not before next core & jumbo releases, as this is an invasive change) is an open question... but auto-tuning `OMP_SCALE` sounds bad to me anyway. Thank you for suggesting it, though!

solardiz:
What we may really need for now is a tool (or JtR builtin feature) to auto-tune `OMP_SCALE` just once, by our team when we're making changes to the formats. With a tool like this, we could also easily run it with varying thread counts and on multiple systems at once and choose what appears to be near-optimal for all.

solardiz:
[...]

PR #3068, opened 2017-12-30
Introduce shared auto-tune code.

magnumripper:
Only implemented in oldoffice format here. Just a PoC. See #2858

### DIFF
--- a/src/omp_autotune.c
+++ b/src/omp_autotune.c
@@ -1,0 +1,256 @@
+/*
+ * This software is Copyright (c) 2018 magnum
+ * and is hereby released to the general public under the following terms:
+ * Redistribution and use in source and binary forms, with or without
+ * modifications, are permitted.
+ */
+
+#ifdef _OPENMP
+#include <omp.h>
+#endif
+
+#include "john.h"
+#include "options.h"
+#include "logger.h"
+#include "formats.h"
+#include "timer.h"
+#include "config.h"
+#include "memdbg.h"
+
+#define CONF_SECTION SECTION_OPTIONS, ":CPUtune"
+
+extern volatile int bench_running;
+
+static int use_preset, max_no_progress;
+static double sample_time, req_gain, max_tune_time;
+
+static struct fmt_main *fmt;
+static int omp_autotune_running;
+static int mkpc;
+static int omp_scale;
+static int report;
+static int fmt_preset;
+static int tune_preset;
+
+void omp_autotune_init(void)
+{
+	int ci;
+
+	use_preset = cfg_get_bool(CONF_SECTION, "UsePreset", 1);
+	if ((ci = cfg_get_int(CONF_SECTION, "AutoTuneSampleTime")) < 0)
+		ci = 10;
+	sample_time = (double)ci / 1000.0;
+	if ((ci = cfg_get_int(CONF_SECTION, "AutoTuneReqGain")) < 0)
+		ci = 5;
+	req_gain = (double)ci / 100.0 + 1.0;
+	if ((ci = cfg_get_int(CONF_SECTION, "AutoTuneMaxDuration")) < 0)
+		ci = 100;
+	max_tune_time = (double)ci / 1000.0;
+	if ((max_no_progress =
+	     cfg_get_int(CONF_SECTION, "AutoTuneMaxNoProgress")) < 0)
+		max_no_progress = 3;
+
+	if (options.tune) {
+		if (!strcmp(options.tune, "auto"))
+			use_preset = 0;
+		else if (!strcmp(options.tune, "report")) {
+			use_preset = 0;
+			report = 1;
+		} else {
+			use_preset = 1;
+			tune_preset = atoi(options.tune);
+		}
+	}
+}
+
+int omp_autotune(struct fmt_main *format, int preset)
+{
+#ifdef _OPENMP
+	int threads = omp_get_max_threads();
+#else
+	int threads = 1;
+#endif
+	int ret_scale;
+
+	fmt_preset = preset;
+
+	omp_scale = tune_preset ? tune_preset : (use_preset ? preset : 0);
+	ret_scale = omp_scale ? omp_scale : 1;
+
+	if (omp_autotune_running)
+		return fmt->params.max_keys_per_crypt / mkpc;
+
+	if (!use_preset || !preset) {
+		fmt = format;
+		mkpc = format->params.max_keys_per_crypt;
+	}
+	format->params.min_keys_per_crypt *= threads;
+	format->params.max_keys_per_crypt *= threads * ret_scale;
+
+	return threads * ret_scale;
+}
+
+void omp_autotune_run(struct db_main *db)
+{
+#ifdef _OPENMP
+	int threads = omp_get_max_threads();
+#else
+	int threads = 1;
+#endif
+	int best_scale = 1, scale = 1;
+	int best_cps = 0;
+	int no_progress = 0;
+	int min_crypts = 0;
+	int tune_cost;
+	void *salt;
+	char key[PLAINTEXT_BUFFER_SIZE] = "tUne0000";
+	sTimer timer;
+	double duration;
+
+	if (!fmt || omp_scale == 1)
+		goto cleanup;
+
+	if (john_main_process && bench_running &&
+	    (!strcmp(options.tune, "report") || options.verbosity > VERB_DEFAULT))
+		fprintf(stderr, "\n");
+
+	omp_autotune_running = 1;
+
+	// Find most expensive salt, for auto-tune
+	{
+		struct db_main *tune_db = db->real ? db->real : db;
+		struct db_salt *s = tune_db->salts;
+
+		tune_cost = MIN(tune_db->max_cost[0], options.loader.max_cost[0]);
+
+		while (s->next && s->cost[0] < tune_cost)
+			s = s->next;
+		salt = s->salt;
+	}
+
+	if (john_main_process && options.verbosity == VERB_MAX) {
+		fprintf(stderr, "%s OMP autotune using %s db",
+		        fmt->params.label, db->real ? "real" : "test");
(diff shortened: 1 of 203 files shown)


magnumripper:
While adding 7z format I saw some caveats. Working on it. Meanwhile please do not add anything to this branch (because interface needs a change).

jfoug:
Question.  Why only do this for OMP?

src/omp_autotune.c
```diff
@@ -0,0 +1,120 @@
...
+			best_scale = scale;
+		}
+		else {
+			if (john_main_process && options.verbosity == VERB_MAX)
+				fprintf(stderr, " -\n");
+			no_progress++;
+		}
+
+		// Double each time
+		scale *= 2;
```
jfoug: how about:
```c
scale += (scale)/2+1;
```
That would chose for values under 200k : 
1 2 4 7 11 17 26 40 61 92 139 209 314 472 709 1064 1597 2396 3595 5393 8090 12136 18205 27308 40963 61445 92168 138253
vs:
1 2 4 8 16 32 64 128 256 512 1024 2048 4096 8192 16384 32768 65536 131072

Just gives a bit more granularity.

Or possibly this expression:

```c
scale += scale < 5 ? 1 ? (scale/3)+1;
```
which gives this result (even better granularity):
1 2 3 4 5 8 12 18 26 36 50 68 92 124 167 224 300 402 538 719 960 1282 1711 2283 3046 4063 5419 7227 9638 12852 17138 22852 30471 40630 54175 72235 96315 128422 171231

src/omp_autotune.c
```diff
@@ -0,0 +1,120 @@
...
+		do {
+			int count = this_kpc;
+
+			self->methods.crypt_all(&count, NULL);
+			crypts += count;
+		} while (sTimer_GetSecs(&timer) < 0.010);
+		sTimer_Stop(&timer);
+
+		// Release buffers
+		self->methods.done();
```
jfoug: I was also calling clear_keys()   is there no reason to do this?

src/omp_autotune.c
jfoug: With a very limited testing (of the formats you already have done), the scale *=2 seems to work as good or better than either of the other 2 'finer' granular methods I proposed, even when I messed around with the ```if (cps >= (best_cps * 1.05))``` code a bit.

KISS method looks damn good ;)

jfoug:
```
Testing: netlm, LM C/R [DES 32/64]... (2xOMP) =================================================================
==23708==ERROR: AddressSanitizer: global-buffer-overflow on address 0x55c653389707 at pc 0x55c652fa6f01 bp 0x7fff33299850 sp 0x7fff33299840
READ of size 1 at 0x55c653389707 thread T0
    #0 0x55c652fa6f00 in get_salt /cwd/src/NETLM_fmt_plug.c:282
    #1 0x55c652e0ab72 in omp_autotune /cwd/src/omp_autotune.c:39
    #2 0x55c652fa7526 in init /cwd/src/NETLM_fmt_plug.c:102
    #3 0x55c6531c9d43 in fmt_init /cwd/src/formats.c:80
    #4 0x55c6531eaaa5 in ldr_init_test_db /cwd/src/loader.c:1396
    #5 0x55c6531ae8e0 in benchmark_all /cwd/src/bench.c:807
    #6 0x55c6531dd23b in john_run /cwd/src/john.c:1617
    #7 0x55c6531deaea in main /cwd/src/john.c:2000
    #8 0x7f96b5d691c0 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x211c0)
    #9 0x55c652d73bf9 in _start (/cwd/run/john+0x1c7bf9)
0x55c653389707 is located 57 bytes to the left of global variable '*.LC12' defined in 'NETLM_fmt_plug.c' (0x55c653389740) of size 10
  '*.LC12' is ascii string 'G3RG3P00!'
0x55c653389707 is located 6 bytes to the right of global variable '*.LC11' defined in 'NETLM_fmt_plug.c' (0x55c653389700) of size 1
  '*.LC11' is ascii string ''
SUMMARY: AddressSanitizer: global-buffer-overflow /cwd/src/NETLM_fmt_plug.c:282 in get_salt
  0x0ab94a669290: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0ab94a6692a0: 00 00 00 00 00 f9 f9 f9 f9 f9 f9 f9 00 00 00 00
  0x0ab94a6692b0: 01 f9 f9 f9 f9 f9 f9 f9 00 00 01 f9 f9 f9 f9 f9
  0x0ab94a6692c0: 00 f9 f9 f9 f9 f9 f9 f9 06 f9 f9 f9 f9 f9 f9 f9
  0x0ab94a6692d0: 07 f9 f9 f9 f9 f9 f9 f9 00 02 f9 f9 f9 f9 f9 f9
=>0x0ab94a6692e0:[01]f9 f9 f9 f9 f9 f9 f9 00 02 f9 f9 f9 f9 f9 f9
  0x0ab94a6692f0: 05 f9 f9 f9 f9 f9 f9 f9 00 00 00 00 00 00 01 f9
  0x0ab94a669300: f9 f9 f9 f9 00 02 f9 f9 f9 f9 f9 f9 00 00 01 f9
  0x0ab94a669310: f9 f9 f9 f9 00 00 00 00 00 00 00 00 00 01 f9 f9
  0x0ab94a669320: f9 f9 f9 f9 00 02 f9 f9 f9 f9 f9 f9 00 00 00 00
  0x0ab94a669330: 00 00 00 00 00 01 f9 f9 f9 f9 f9 f9 00 07 f9 f9
  Addressable:           00
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==23708==ABORTING
```

jfoug:
NetLM needs prepare called, I think.  Or else, we need to validate that all formats have test hash 0 in properly prepared format.

src/omp_autotune.c
magnumripper: Right, that needs to be added

magnumripper:
> Question. Why only do this for OMP?

Good question! We can look into that too.

magnumripper:
NETLM and others are fixed in my latest code but it's not quite finished yet. It also adds tuning for most expensive salt of all test vectors or - in case we have real hashes loaded - the most expensive salt of those (similar to OpenCL autotune). I will probably not finish it or push anything new here in 48h or so (guess why 🎆 )

magnumripper:
In order to tune for actual loaded hashes and/or heaviest cost, I had to change interface for `omp_autotune()`. We need to get a pointer to `self` (format struct) as well as the `db`. This is not sweet but works:
```diff
 static void init(struct fmt_main *self)
 {
 #ifdef _OPENMP
-       int threads = omp_get_max_threads();
-
-       if (threads > 1) {
-               self->params.min_keys_per_crypt *= threads;
-               threads *= OMP_SCALE;
-               self->params.max_keys_per_crypt *= threads;
-       }
+       omp_autotune(self, NULL);
 #endif
        saved_key = mem_calloc(self->params.max_keys_per_crypt,
                               sizeof(*saved_key));
@@ -122,6 +108,13 @@ static void init(struct fmt_main *self)
        output = mem_calloc(self->params.max_keys_per_crypt, sizeof(*output));
 }
 
+static void reset(struct db_main *db)
+{
+#if defined (_OPENMP)
+       omp_autotune(NULL, db);
+#endif
+}
+
 static void done(void)
 {
        MEM_FREE(output);
```
The call from `init()` merely saves off the `self` pointer and returns number of actual threads. The call from `reset()` will actually tune a scale. If real hashes are loaded, they will be used for tuning instead of test vectors. Otherwise the test vectors are used.

I'm actually not quite sure whether we should tune for heaviest or lightest cost? An experienced user will usually not run with very different costs at once though. Lightest cost would sometimes mean a much faster autotune.

@solardiz any thoughts (this is still just an experiment BTW)?

magnumripper:
I think this is about as good as this gets. Mostly KISS, mostly working as a champ. Some quirks. This really reveals formats like 7z where the speed difference between fast and slow salts are gigantic. We have no option other than to benchmark for slowest salt, or self-test does take ages. OTOH we have the exat same problem with any scheme of manual tuning too.

Not sure whether we should merge this or just consider it an interesting experiment.

magnumripper:
Brand new version, less intrusive. Keeps the OMP_SCALE values and uses them by default.

More to come.

magnumripper:
Rebased a cleaned-up version. Now formats.h in turn sources omp_autotune.h so most formats merely have this change:
```diff
diff --git a/src/AzureAD_fmt_plug.c b/src/AzureAD_fmt_plug.c
index 35f5b8011..b6a64bf72 100644
--- a/src/AzureAD_fmt_plug.c
+++ b/src/AzureAD_fmt_plug.c
@@ -80,13 +80,7 @@ static uint32_t (*crypt_out)[BINARY_SIZE / sizeof(uint32_t)];
 static void init(struct fmt_main *self)
 {
 #ifdef _OPENMP
-       int threads = omp_get_max_threads();
-
-       if (threads > 1) {
-               self->params.min_keys_per_crypt *= threads;
-               threads *= OMP_SCALE;
-               self->params.max_keys_per_crypt *= threads;
-       }
+       omp_autotune(self, OMP_SCALE);
 #endif
        saved_key = mem_calloc_align(self->params.max_keys_per_crypt,
                                     sizeof(*saved_key), MEM_ALIGN_WORD);
```

magnumripper:
I simply had to merge this before @kholia forced me to rebase yet another again 😄 

All tests passed and behavior isn't changed unless you ask for it.

PR #3117, opened 2018-01-17
Enable fallback for CygWin

### LINKED ISSUE #3076: Fallback on CygWin
It is probably a good idea to enable fallback on Windows (CygWin) on Core.

### IMPORTANT
- http://www.openwall.com/lists/john-dev/2018/01/01/1

I used:
```
index ff30b1668..38b1cfb4d 100644
--- a/src/john.c
+++ b/src/john.c
@@ -463,7 +463,7 @@ static void john_omp_init(void)
 }
 
 #if OMP_FALLBACK
-#if defined(__DJGPP__) || defined(__CYGWIN__)
+#if defined(__DJGPP__)
 #error OMP_FALLBACK is incompatible with the current DOS and Windows code
 #endif
 #define HAVE_JOHN_OMP_FALLBACK
@@ -1393,7 +1393,7 @@ static void CPU_detect_or_fallback(char **argv, int make_check)
        if (!CPU_detect()) {
 #if CPU_REQ
 #if CPU_FALLBACK
-#if defined(__DJGPP__) || defined(__CYGWIN__)
+#if defined(__DJGPP__)
 #error CPU_FALLBACK is incompatible with the current DOS and Windows code
 #endif
                if (!make_check) {
```
Removing the fallback binary
```
C:\bleeding\run>john --list=build-info
execv: $JOHN/john-sse2: No such file or directory
Sorry, AVX is required for this build
```

*****
PS: https://github.com/claudioandre/packages/blob/master/john-the-ripper/readme.md#windows

claudioandre:
### Summary
* Closes #3076;
* More info [here](http://www.openwall.com/lists/john-dev/2018/01/01/1).

### DIFF
--- a/src/john.c
+++ b/src/john.c
@@ -464,8 +464,8 @@ static void john_omp_init(void)
 }
 
 #if OMP_FALLBACK
-#if defined(__DJGPP__) || defined(__CYGWIN__)
-#error OMP_FALLBACK is incompatible with the current DOS and Windows code
+#if defined(__DJGPP__)
+#error OMP_FALLBACK is incompatible with the current DOS code
 #endif
 #define HAVE_JOHN_OMP_FALLBACK
 static void john_omp_fallback(char **argv) {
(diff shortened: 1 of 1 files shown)


magnumripper:
Thanks!

PR #3193, opened 2018-03-07
OpenCL: warn if --dev lists more OpenCL devices than fork

### LINKED ISSUE #3176: Warn when -dev lists more OpenCL devices than the number in -fork
Our current poor man's support for use of multiple OpenCL devices by a single instance of John is through "--fork". However, when a user specifies e.g. "--dev=0,1" without a "--fork", John silently uses the first device only. This is a major usability issue, especially for new users. At the very least, John should print a Warning when invoked without "--fork" or with fewer forks than the number of OpenCL devices.

Important: this new logic should apply to OpenCL devices only (or maybe to "-opencl" formats only, if that's simpler to check for). It must not apply to ZTEX devices, where multiple devices are usable without "--fork", and where "--fork" is likely to result in misbehavior (maybe we should even disallow use of "--fork" with those, but that's a separate issue).

claudioandre:
Print a warning when JtR is invoked without "--fork" or with fewer forks than the
number of OpenCL devices.

Closes #3176


claudioandre:
Pushed again.

PR #3194, opened 2018-03-08
OpenCL: only complain about --dev and fork on non MPI

claudioandre:
As seen in https://github.com/magnumripper/JohnTheRipper/issues/3176#issuecomment-371292880 commit 302a6ec3888354d03f51b60a0e1fa6ad67305f1c broke MPI.

* I'm really not sure what kind of warning (as requested in #3176) should be printed on a MPI run. So, at least as a workaround, I'm disabling the new warnings in case of MPI.

### DIFF
--- a/src/john.c
+++ b/src/john.c
@@ -1217,6 +1217,9 @@ static void john_load(void)
 			log_event("Loaded a total of %s", john_loaded_counts());
 			/* make sure the format is properly initialized */
 
+#if HAVE_MPI
+	if (mpi_p == 1) {
+#endif
 #if HAVE_OPENCL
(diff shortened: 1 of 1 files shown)


magnumripper:
There's another problem. `options.fork` is actually 0 when not used so the original fix doesn't work properly for non-MPI builds when using just one device:
```
$ ../run/john scraped.sam -form:nt-opencl -w 
Device 2: GeForce GT 650M
To fully use the 1 devices requested, you must specify --fork=1
```
I'm reverting the orignal fix now, I will modify this so it works properly with or without MPI.

PR #3196, opened 2018-03-09
mask: add FMT_MASK flag to raw-SHA512-opencl

claudioandre:
```
$ ../run/john  -format:mask -list=formats
../run/john: /usr/local/cuda-8.0/targets/x86_64-linux/lib/libOpenCL.so.1: no version information available (required by ../run/john)
oldoffice-opencl, descrypt-opencl, krb5pa-md5-opencl, LM-opencl, 
mscash-opencl, mysql-sha1-opencl, nt-opencl, ntlmv2-opencl, Raw-MD4-opencl, 
Raw-MD5-opencl, Raw-SHA1-opencl, Raw-SHA256-opencl, Raw-SHA512-opencl, 
salted-sha1-opencl, SL3-opencl, XSHA512-opencl
```


magnumripper:
I just saw this too, missed that there were two format structs in same file. I think this was the last missing one.

PR #3200, opened 2018-03-14
Major mask mode changes

magnumripper:
Executive summary:
- Mask stretching can now stretch a mask starting with a range, not only ones ending with one. For example, you always could `-mask=admin?d -min-len=6` but now you can also do `-mask=?dadmin -min-len=6`. There's a pending change that will also stretch a mask like `FOO?dBAR` into `FOO?d?dBAR`, `FOO?d?d?dBAR` and so on but that's not included here.
- Any incremental (iterated length) mask with GPU-internal-mask formats is now supported, even from length 0. This means we'll have to call `reset()` between lengths and that takes a while (new auto-tune) but it sure is faster than not handling it at all.
- Many other bug fixes, tweaks and enhancements.

Two formats are yet to handle all of the above, namely `LM-opencl` and `DEScrypt-opencl`.  I worked like a galley slave with this for more than two (three?) weeks but they didn't make it yet. They will however cope a whole lot better now, adapting and warn when things are not 100% what you may have asked for. I intend to fix them as well, sooner or later.

### DIFF
--- a/doc/MASK
+++ b/doc/MASK
@@ -68,14 +68,16 @@ This should be used with -max-len (and possibly -min-len) to do any good.
 The -max-len=N option will truncate the mask so no words longer than N are
 produced.
 
-The -min-len=N option will skip generation of words shorter than N, and more
-importantly in this case, it will iterate length from -min-len to -max-len (or
-format's max length, if -max-len was not given).  So to produce all possible
-words from 3 to 5 letters, use -mask=?l -min-len=3 -max-len=5.  In case the
-specified mask is shorter, the last part of it will be expanded, for example
-"-mask=?u?l -max-len=5" will use an effective mask of ?u?l?l?l?l. Whenever
-iterating over lengths, the ETA shows estimated time to complete the *current*
-length, as opposed to the whole run.
+The -min-len=N option will skip generation of words shorter than N.
+
+If not in "hybrid mask" mode, and either -min-len or -max-len option was used,
+we will iterate lengths (as in "incremental mask") from -min-len to -max-len
+(or format's min or max length, if one was not given).  So to produce all
+possible words from 3 to 5 letters, use -mask=?l -min-len=3 -max-len=5.  In
+case the specified mask is shorter, the last part of it will be expanded, for
+example "-mask=?u?l -max-len=5" will use an effective mask of ?u?l?l?l?l.
+Whenever using incremental mask, the ETA at any given time shows estimated
+time to complete the *current* length, as opposed to the whole run.
 
 You can escape special characters with \.  So to produce a literal "?l" you
 could say \?l or ?\l and it will not be parsed as a placeholder.  Similarly,
--- a/src/bt.c
+++ b/src/bt.c
@@ -35,7 +35,6 @@ typedef struct {
 	unsigned short collisions;
 	unsigned short iter;
 	unsigned int offset_table_idx;
-
 } auxilliary_offset_data;
 
 /* Interface pointers */
@@ -465,7 +464,7 @@ static unsigned int create_tables()
 	alarm(0);
 
 	hash_table_idx = 0;
-	while (offset_data[i].collisions > 0) {
+	while (i < offset_table_size && offset_data[i].collisions > 0) {
 		done++;
 
 		while (hash_table_idx < hash_table_size) {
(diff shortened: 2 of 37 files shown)


magnumripper:
@claudioandre I intend to merge this if the build bots passes but in case you have time, please hit this with your largest gun so we don't introduce worse code (bloody unlikely though) than we already had. I actually do know a few bugs but all of them was present even before this!

claudioandre:
Running (it takes about 2 hours)

1. https://travis-ci.org/claudioandre/JohnTheRipper/builds/353539067
2. https://circleci.com/gh/claudioandre/JohnTheRipper/1120
3. https://circleci.com/gh/claudioandre/JohnTheRipper/1119
4. https://ci.appveyor.com/project/claudioandre/johntheripper/build/1.8J1-39cc35194

magnumripper:
Is any of those test actually testing GPU-side mask mode? If not, they won't likely find any problem. I intent do enhance TS soon with real, systematic, tests.

OTOH as long as we don't introduce a build-bot failure, things are not worse than before 🙄

claudioandre:
Not exactly. But I have some tests with mask + the CPU OpenCL driver.
---

And, I have a set of tests that does test GPU OpenCL (including mask). But I have to run them by hand (on super). Working on it.

magnumripper:
I've tested stuff so much my fingers are bleeding. Once this gets in, we can close several issues.

claudioandre:
The regular mask code is doing good on other archs (s390x and ppc64el finished):
```
 Ok: --format=sha512crypt --mask=jo?l[n-q]
```

magnumripper:
Here's one test I ran for GPU-side-mask formats
```
$ cat pre.lst 

1
7a
8ad
9adm
1admi
45admin
932admin
4022admin
40079admin
```
Then created test input files from that for each format, then
```
rm -f ../run/john.pot && for i in Raw-MD4 Raw-MD5 Raw-SHA1 Raw-SHA256 Raw-SHA512 salted-sha1 XSHA512 krb5pa-md5 mysql-sha1 netntlmv2 ; do ../run/john pre.$i.in -form:$i-opencl -mask:?dadmin -max-len=10 -v:1 -dev=cpu && ../run/john -form:$i-opencl -show=left pre.$i.in | tail || break ; done
```

claudioandre:
One error happened. I'll dig into it to see what changed and why.

```
Running raw-SHA256 cracking tests...
ERROR:  ../../run/john -ses=tst-cla -pot=tst-cla.pot alltests.in -form=Raw-SHA256-opencl -mask:?l -min-len=4 -max-len=7 &> /dev/null
Expected value: 2, value found: 3. ../../run/john -show=left -pot=tst-cla.pot alltests.in -form=Raw-SHA256-opencl &> _tmp_output.DvOOgchB
```

claudioandre:
It is the empty password. Not generated previously by `-mask:?l -min-len=4 -max-len=7`.

```
0g 0:00:00:03  (5) 0g/s 0p/s 0c/s 0C/s GPU:45°C
                 (?)
test             (?)
```

magnumripper:
> It is the empty password. Not generated previously by `-mask:?l -min-len=4 -max-len=7`.

That's a bit worrying. Shouldn't happen. Did you check that the hash for it is actually a null hash?

magnumripper:
In the pot file, I mean

claudioandre:
Another issue, but also related to the blank password (inverted now: the old generates the blank, the new one is not generating). So, nothing really bad so far.
`-mask:xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx[x-z] -min=55 -max-l=55`

claudioandre:
It is the blank from https://github.com/magnumripper/JohnTheRipper/blob/bleeding-jumbo/src/rawSHA256_common.h#L46:
```
../../run/john -form:cpu --list=format-tests 2> /dev/null | cut -f3 1> alltests.in
```

> That's a bit worrying.

Maybe not. 
* If KPC=1024 but the mask only generates 1023 candidates, one (on a clean buffer) is the empty. So, I expect it to happen at some stages.

Do you disagree?

claudioandre:
I will rerun the adjusted tests. Anyway, it is safe to merge.

claudioandre:
```
All tests passed without error! Performed 89 tests in 1240 seconds.
```

PR #3205, opened 2018-03-17
docs: alert about a distro package of JtR

claudioandre:
If nobody complains, I'll merge this.

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,6 +1,7 @@
 ### IMPORTANT
 - If you have questions, please ask them first in the mailing list john-users at lists.openwall.com;
 - Use issues to keep track of ideas, enhancements, tasks, and bugs. NEVER as a support forum;
+- Usually, it is a bad idea to use the version of John the Ripper packaged by your distro;
 - Please, DO NOT create a new issue if your bug is already known:
   - OpenCL known issues #2331.
   - CUDA support is dropped from tree now. Do not post issues about that.


frank-dittrich:
If we add such a comment, we should at least partly blame ourselves for not releasing a new stable jumbo version for a very long time.
After all, what should a distro do here?
Distribute just a john core package, without any jumbo additions?
Distribute a jumbo package based on our last jumbo release?
Pick some random git commit for packaging?

And when we finally manage a new jumbo release, we need to rephrase this comment again.

frank-dittrich:
@magnumripper 
> However, it's a bad idea to run a distro's (old) version and then report bugs to us. If running a distro, user should report bugs to them, not us.

And what are the distros supposed to do with these bug reports? Report them upstream? Cherry-pick individual git commits for new distro releases?

claudioandre:
> I wouldn't say it's generally a bad idea 

In fact, it is up to you to define what should be stated (and keep it, as is, maybe, is the best choice). But, personally, I would like to make it clearer somewhere.
* distro packages are from a remote past (years);
* why report a bug based on "literally several thousands of commits" in the past)
* report bugs to the distro is a waste of time, IMO;
* report bugs in the distro and here is 2x waste of time; 

Ok (in my opinion), it is always a bad idea to use the distro package.

magnumripper:
> And what are the distros supposed to do with these bug reports? Report them upstream? Cherry-pick individual git commits for new distro releases?

I have no idea but I do know that such bug reports are just useless noise to me. This is why I think we should either release at least 1-2 times a year (I wish we could but we simply can't because we don't have the resources) or drop the notion of releases.

solardiz:
> I think we should either release at least 1-2 times a year

Yes, I think we should.

> (I wish we could but we simply can't because we don't have the resources)

I think we do have the resources, we're just not trying to make releases - but we should.

magnumripper:
I'll rewrite this a little

PR #3207, opened 2018-03-17
OpenCL: add workaround for '-I' on CygWin

claudioandre:
Magnum, this is the workaround I used to handle 3191. 
* I was able to make CygWin pretend it is at `/`, as reported (e.g., see it doing `-I /run/kernels`);
* I'm not sure if it should be merged, but since it was useful in one of my tests, throwing it out does not seem appropriate either;
* Please, check if you find it useful. Otherwise, I'll discard it.

E.g:
```
C:\bleeding
|_ run
|_ doc
|_ ...
```

### DIFF
--- a/src/opencl_common.c
+++ b/src/opencl_common.c
@@ -1009,6 +1009,56 @@ static char *quote_str(char *orig)
 	return new;
 }
 
+#if defined(__CYGWIN__)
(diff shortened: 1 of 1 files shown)


claudioandre:
> Is there any risk of regression on some Cygwin environments?

Well, line 1021 should make it safe and free of regressions.

PR #3212, opened 2018-03-20
OpenCL: fix for an uninitialized pointer in bitwarden

claudioandre:


magnumripper:
Thanks. That's what happens when using -F3 with `patch`... effective but dangerous.

PR #3230, opened 2018-04-11
Add a cracking rule that converts words in t9

fpesce:
Following the comment from some company claiming they can get passwords
typed from a phone call via T9.

There was a discovery that Fidelity investment was doing that kind of password conversion to be able to get passwords typed by phone.

I've successfully run this rule against Troy Hunt pwned password lists and found a few hundreds.

I think List.Rules:Extra is a good place don't hesitate to tell me if there's a better one.

### DIFF
--- a/run/john.conf
+++ b/run/john.conf
@@ -745,6 +745,9 @@ l Az"[1-90][0-9][0-9]" <+
 -c (?a c Az"[1-90][0-9][0-9]" <+
 <[\-9] l A\p[z0]"[a-z][a-z]"
 <- l ^[a-z] $[a-z]
+# T9 text
+l sa2 sb2 sc2 sd3 se3 sf3 sg4 sh4 si4 sj5 sk5 sl5 sm6 sn6 so6 sp7 sq7 sr7 ss7 st8 su8 sv8 sw9 sx9 sy9 sz9 s?D*
+l sa2 sb2 sc2 sd3 se3 sf3 sg4 sh4 si4 sj5 sk5 sl5 sm6 sn6 so6 sp7 sq7 sr7 ss7 st8 su8 sv8 sw9 sx9 sy9 sz9 s?D#
 
 # Wordlist mode rules
 [List.Rules:Wordlist]
--- a/run/john.conf
+++ b/run/john.conf
@@ -866,8 +866,8 @@ o[0-9A-E][ -~] Q M o[0-9A-E][ -~] Q
 i[0-9A-E][ -~] i[0-9A-E][ -~]
 
 [List.Rules:T9]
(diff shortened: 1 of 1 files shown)


magnumripper:
For fast hashes, this is more or less futile because it's more effective to just brute-force digits. But for slower hashes it might have some merit.

run/john.conf
```diff
@@ -745,6 +745,9 @@ l Az"[1-90][0-9][0-9]" <+
 -c (?a c Az"[1-90][0-9][0-9]" <+
 <[\-9] l A\p[z0]"[a-z][a-z]"
 <- l ^[a-z] $[a-z]
```
magnumripper: Thanks! You are currently adding the new rules to `[List.Rules:Extra]`. For me to merge it, I'd want you to instead create a new section eg. `[List.Rules:T9]`. Please do that and I'll merge it.

fpesce: Ok done. Re: Fast hash vs slow hash. I've found a fair amount (few dozen) of 14-16+ digits long passwords (from Troy Hunt's pwned pass file) with that rule and big dictionaries and I hadn't pushed my Digits brute force above 10 chars long 😅

fpesce:
Changes submitted ( e42ce8c )

solardiz:
> It might be better to skip this word if it only contains digits.

We could also have the /?D on the first rule before all of the substitutions. Like this:

```
/?D l sa2 sb2 sc2 sd3 se3 sf3 sg4 sh4 si4 sj5 sk5 sl5 sm6 sn6 so6 sp7 sq7 sr7 ss7 st8 su8 sv8 sw9 sx9 sy9 sz9 s?D*
/?D l sa2 sb2 sc2 sd3 se3 sf3 sg4 sh4 si4 sj5 sk5 sl5 sm6 sn6 so6 sp7 sq7 sr7 ss7 st8 su8 sv8 sw9 sx9 sy9 sz9 /?D s?D#
```

Yes, it's two `/?D` checks in the second rule. And in this case no use of the preprocessor.

magnumripper:
> Shouldn't `[List.Rules:T9]` be appended at the end of `[List.Rules:All]`

Agreed. Just thinking loud, we could drop `all` from the config and make that a hard-coded thing in `rules.c` instead, so it includes all rule-sets in any sourced config file without need for updates. Or maybe that's just code bloat.

> (Looks looks like All and Jumbo are currently the same, except that All includes some sections which now are already part of Jumbo.)

I need to look into that. IIRC, `all` would include all korelogic stuff while `jumbo` should not, or something like that, given that korelogic rules are fairly low quality.

fpesce:
@solardiz  I think your second suggestion is more elegant, saving all digits only to be generated twice. I've updated my pull request accordingly.
@magnumripper I've appended T9 to List.Rules:All

magnumripper:
Merged, thanks!

PR #3370, opened 2018-08-15
maint: refuse --device if the format is not device-aware

### LINKED ISSUE #3369: Refuse to run with the "--device" option when the format is neither OpenCL nor ZTEX
In #3368 
> we should probably refuse to run with the "--device" option when the format is neither OpenCL nor ZTEX (nor whatever future format types might be device-aware).  Including for auto-detected formats, like in this example.

### Initial Proposal ###
```
index 9b7a254..5e1e96a 100644
--- a/src/john.c
+++ b/src/john.c
@@ -1231,6 +1231,17 @@ static void john_load(void)
                        else
                                log_event("Starting a new session");
                        log_event("Loaded a total of %s", john_loaded_counts());
+                       /* only allow --device for OpenCL or ZTEX formats */
+#if HAVE_OPENCL || HAVE_ZTEX
+                       if (options.acc_devices->count &&
+                         !(strstr(database.format->params.label, "-opencl") ||
+                           strstr(database.format->params.label, "-ztex"))) {
+                               fprintf(stderr,
+                                   "The \"--devices\" option is valid only for OpenCL or ZTEX formats\n");
+                               error();
+                       }
+#endif
+
                        /* make sure the format is properly initialized */
 #if HAVE_OPENCL
                        if (!(options.acc_devices->count && options.fork &&
```

claudioandre-br:
Closes #3369 

CI is happy.

### DIFF
--- a/src/john.c
+++ b/src/john.c
@@ -1231,6 +1231,17 @@ static void john_load(void)
 			else
 				log_event("Starting a new session");
 			log_event("Loaded a total of %s", john_loaded_counts());
+			/* only allow --device for OpenCL or ZTEX formats */
+#if HAVE_OPENCL || HAVE_ZTEX
+			if (options.acc_devices->count &&
+			  !(strstr(database.format->params.label, "-opencl") ||
+			    strstr(database.format->params.label, "-ztex"))) {
+				fprintf(stderr,
+				    "The \"--devices\" option is valid only for OpenCL or ZTEX formats\n");
+				error();
+			}
+#endif
+
 			/* make sure the format is properly initialized */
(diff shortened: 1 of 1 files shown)


claudioandre-br:
> it wont kick in until after a potentially long load time. I guess we can't avoid it.

Indeed. Well, as an end user (as I was in the "CMIYC pre-contest challenge" cracking), I was happy that JtR can help me when I was doing something silly. Even if I have to pay for it.

> I suppose the FMT_GPU flag can be a separate issue and shouldn't stop this PR.

We have to change all OpenCL formats. Keeping it in your own crate is best.

PR #3374, opened 2018-08-18
sha256crypt-ztex

Apingis:
sha256crypt-ztex
- incl. Verilog sources
- Design tools reported possible frequency is 166.3. At that frequency tested boards fail
often (timeout) and lose successful guesses. Default frequency is set to 135.
- Performance: 71.5 Kc/s for length 8; 60.6 Kc/s for length 12.
- Measured power consumption (12V): ~2.6 А. Idle power consumption is ~1.0 А.


src/ztex/fpga-sha256crypt/README.md
```diff
@@ -0,0 +1,143 @@
...
+           +----------+-----------                     |
+           |                                           |
+ . . . . . | . . . . . . . . . . . . . . . . . . . . . |. .
+           |                                          /
+           |                                         /
+  +---------------+         -------------           /
+  |               |        /             \         /
+  | process_bytes |       /    "main"     \ <------
+  |               |<--+--|      memory     |
+  +---------------+   |   \    (16x256b)  / <------
```
solardiz: Should be 6x128B. BTW, let's use "B" for bytes if we use "b" for bits (ditto for the sha512crypt-ztex README). These documentation issues can be fixed in a separate commit and PR for both sha256crypt-ztex and sha512crypt-ztex.

src/ztex/fpga-sha256crypt/README.md
```diff
@@ -0,0 +1,143 @@
...
+ | procb |  |       /                           +------------+
+ | saved |  |      /    +------------------+    | unit_input |
+ | state |  |     /     | thread_state(x6) |    +------------+
+ +-------+  |    |      +------------------+              ^
+            |    |                                        |
+  +-----------+   \   +-----------------------------+     |
+  | procb_buf |    -->|          C.P.U.             |     |
+  +-----------+       |  +-----------------------+  |     |
+          ^           |  | integer ops. incl.    |  |     |
+           \          |  | 6 x 16 registers(x16b)|  |     |
```
solardiz: Here's where we use "b" for bits, and that's OK.

src/ztex/fpga-sha256crypt/README.md
```diff
@@ -0,0 +1,143 @@
...
+  +---------------------+----------------------+
+  |  Prog. clocks  | USB I/O   |  FPGA reset   |
+  +--------------------------------------------+
+
+```
+fig.2. Overview, FPGA application
+
+- Each FPGA has 25 computing units, that's 75 cores, 150 keys are
+computed in parallel.
+
```
solardiz: It would be helpful to add (with the separate commit/PR) some info on relative resource usage by SHA-256 cores vs. the "overhead" (CPUs, glue logic). We do have this info in sha512crypt-ztex's README.

run/john.conf
```diff
@@ -376,6 +376,14 @@ Frequency = 141
 # This turns off all 10 units (resulting in a timeout).
 #Config1_04A36E0FD6_0 = \xff\x03
 
+[ZTEX:sha256crypt]
+# Design tools reported possible frequency is 166.3 but tested boards
+# miss guesses, often fail unless frequency is decreased.
```
solardiz: It'd be helpful to add a similar comment for sha512crypt and Drupal7, stating those formats' design tools' reported frequency.

run/john.conf
```diff
@@ -376,6 +376,14 @@ Frequency = 141
...
+#Frequency = 165
+Frequency = 135
+#TargetRounds = 1000000
+
+
```
solardiz: Extra linefeed here.

Apingis: Thank you!
The documentation definitely requires an update / correction. I'm preparing a PR to include topics you've commented on. Maybe some other items require more detailed description, such as resource usage, performance of 16-bit CPU in sha256crypt-ztex vs 32-bit CPU in sha512crypt-ztex.
Regarding extra linefeed in run/john.conf:L386, I was unsure on style in john.conf. Sometimes sections that refer to different subsystems are separated with 2 empty lines, i.e. L419, L474.

solardiz: You're right about the extra linefeed, I missed the fact this was the very last ZTEX entry, and is followed by OpenCL stuff. Makes sense.

PR #3376, opened 2018-08-19
Improved ztex documentation

Apingis:


run/john.conf
```diff
@@ -361,11 +361,17 @@ Frequency = 141
 
 [ZTEX:sha512crypt]
 #TargetRounds = 5000
+# Design tools reported possible frequency to be 225.1 MHz.
+# On tested boards it often fail (with error or timeout), or doesn't
+# find passwords at such high frequency. Default frequency is set
+# to 135 MHz.
```
solardiz: Grammar fix: s/fail/fails/

More importantly, I think we never encountered a board where this worked at anywhere close to 225 MHz for a bitstream with all 10 units in it. The best I got out of it is 180 MHz or so with 5 units disabled in john.conf (and 5 left enabled). With all 10 units, the best realistic speed is 160 MHz. Perhaps we should give 160 MHz as an example in the comment, e.g. saying "Some lucky boards run this OK at 160 MHz, but usually not more."

src/ztex/fpga-sha256crypt/README.md
```diff
@@ -141,3 +141,27 @@ fig.2. Overview, FPGA application
...
+
+- "procb_buf" (intermediate storage of data submitted by PROCESS_BYTES
+instructions): 48 LUT (2.0%)
+- "realign" (fetches data from 32-bit memory and performs proper
+alignment): 92 LUT (3.8%)
+- "process_bytes" (operates the above, creates SHA256 blocks, adds
+padding, counts total bytes etc.): 235 LUT (9.6%)
+- "thread_state" with 3 independent read and 4 write channels (reflects
+state of each thread for CPU and other subsystems): 35 LUT (1.4%)
+
```
solardiz: Thank you for this additional detail.

The sha512crypt-ztex README mentions manual layout. Since this format is similar/related, perhaps this should be mentioned for it as well (if sha256crypt-ztex doesn't use a manual layout, this can be mentioned as a difference from sha512crypt-ztex).

PR #3394, opened 2018-09-04
Botcheck

magnumripper:


magnumripper:
For the life of me, I can't understand why one PR creates two identical jobs in Travis CI. They must feel an urge to warm up the planet or something.

PR #3398, opened 2018-09-06
Botcheck

magnumripper:


src/pp.c
```diff
@@ -1505,7 +1508,12 @@ void do_prince_crack(struct db_main *db, char *wordlist, int rules)
   }
 
 #ifdef HAVE_MMAP
-  if (options.flags & FLG_PRINCE_MMAP)
+  if (mmap_max == -1)
+  {
+    mmap_max = 1 << 20;
+  }
+  if (options.flags & FLG_PRINCE_MMAP &&
+      mmap_max && mmap_max >= (file_len >> 20))
```
solardiz: The check `mmap_max >= (file_len >> 20)` will allow mmap'ing files that are almost 1 MiB above the limit. I think we should use `file_len <= (uint64_t)mmap_max * 1000000ULL`. Ditto in wordlist.c.

magnumripper: True. I merged this before I saw your comments but I will consider them both, thanks!

PR #3421, opened 2018-10-06
improvements to md5crypt-ztex, phpass-ztex

Apingis:
- Increased PLAINTEXT_LENGTH for phpass-ztex to 64, tested on a hash provided by Solar;
- In md5crypt-ztex, removed dummy "iteration count" tunable cost from a view of a user, let it pass "silently" to fpgas, without a modification of bitstream's application-level input packet processing unified for both "formats";
- Reviewed where it sets programmable clocks' frequency on fpga's "soft" reset (Global Set Reset, GSR).  Internal programmable clock sources (DCMs, PLLs) don't reset with GSR. It sets: if it finds "Frequency" in ```john.conf``` then it sets given frequency, else uses frequency value from ```struct device_bitstream```. Everything seems correct.
- Minor corrections.

### DIFF
--- a/src/ztex_md5crypt.c
+++ b/src/ztex_md5crypt.c
@@ -71,7 +71,7 @@ static struct device_bitstream bitstream = {
 	1536 * 1024,	// Would be 48 MB of USB traffic on 32-byte keys
 	512,		// Max. number of entries in onboard comparator.
 	384,		// Min. number of keys for effective device utilization
-	1, { 135 },	// Programmable clocks
+	1, { 180 },	// Programmable clocks
 	"md5crypt",	// label for configuration file
 	"\x00", 1	// Initialization data
 };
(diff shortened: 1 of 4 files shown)


src/ztex_md5crypt.c
```diff
@@ -131,16 +131,22 @@ static void init(struct fmt_main *fmt_main)
...
 	// 1 byte unused, 1 byte salt_len, 16 bytes salt in network byte order
 	salt_buf[0] = 0;
 	salt_buf[1] = salt_len;
 	memcpy(salt_buf + 2, salt->salt, salt_len);
+	// The device expects salt (18 bytes) then 4 bytes for each
+	// "tunable cost", for those found in struct fmt_main.
+	// We add-up "tunable cost" of 1,000 like if it was in "struct fmt_main".
+	salt_buf[18] = 1000 % 256;
+	salt_buf[19] = 1000 / 256;
+	*((uint16_t *)(salt_buf + 20)) = 0;
```
solardiz: This is a bug: `salt_buf` is not guaranteed to be aligned to allow for the 16-bit write. Please use two single-byte writes instead.

PR #3422, opened 2018-10-08
handle limit on input (to the board) packet length; minor corrections

Apingis:


solardiz:
@Apingis With these changes, my tests with `-ext=repeats -max-len=64` no longer trigger a problem, however this still fails:

```
$ ./john --test-full=1 --format=md5crypt-ztex
[...]
pkt_new(type 1): data_len(1048576) exceeds 1048558 bytes
```

solardiz:
Perhaps you should add a new macro for `PKT_MAX_LEN - PKT_HEADER_LEN - 2 * PKT_CHECKSUM_LEN` and use it in both places.

PR #3425, opened 2018-10-10
sha512crypt-ztex improvement (incl.new bitstream)

Apingis:
- Improved "glue logic", the following is gone: "On some hashes, dependent on key length and salt length, performance degradation up to 10% is observed. The cause is not-so-effective iteraction between CPU and process_bytes unit."
- Different approach to clocking the system. Only 2 clocks now: interface one (IFCLK) and "main" one (CORE_CLK). Previously used PKT_COMM_CLK was removed. That removes potential synchronization issues.
- Minor improvements.
- On 2 boards tested it works fine at 160 MHz. Previous bitstream worked reliably at 135 MHz.
- Power consumption is 3.2-3.3A at 12V.


run/john.conf
```diff
@@ -363,11 +363,11 @@ Frequency = 141
...
 #TargetRounds = 5000
-# Design tools reported possible frequency to be 225.1 MHz.
+# Design tools reported possible frequency to be 215 MHz.
 # We never encountered a board where this worked anywhere close
-# to such high frequency. Default frequency is set to 135 MHz.
-# Some lucky boards run this OK at 160 MHz, but usually not more.
-#Frequency = 135
+# to such high frequency. Default frequency is set to 160 MHz.
+# Some lucky boards might run at some higher frequency.
+Frequency = 160
```
solardiz: We need to similarly change the default frequency in the section below for `drupal7-ztex`, and also change both default frequencies in the `.c` files.

solardiz:
I confirm that 160 MHz is the new reliable frequency, in testing on my 4 boards. On two lucky boards (the same ones that also run sha256crypt-ztex better), this new design is stable at 170 MHz (and also runs at 175 MHz briefly, but starts to fail as the boards heat up).

PR #3427, opened 2018-10-13
ztex --test works (also with --mask), displays correct numbers

Apingis:
- New global variable ```int self_test_running``` available in ```john.h``` to distinguish between self-test and benchmark
- Log message "Candidate passwords %s be buffered and tried in chunks of %d" delayed to let  ```mask_init()```, ```fmt_reset()``` be performed first
- Reviewed test vectors in ztex formats

### DIFF
--- a/src/idle.c
+++ b/src/idle.c
@@ -66,6 +66,8 @@ int idle_requested(struct fmt_main *format)
 	if (strstr(format->params.label, "-opencl"))
 		return 0;
 #endif
+	if (strstr(format->params.label, "-ztex"))
+		return 0;
 
 	return 1;
 }
--- a/src/john.c
+++ b/src/john.c
@@ -189,6 +189,8 @@ char *john_terminal_locale ="C";
 
 unsigned long long john_max_cands;
 
+int self_test_running = 0;
+
 static int children_ok = 1;
 
 static struct db_main database;
@@ -412,7 +414,6 @@ static void john_register_all(void)
 
 static void john_log_format(void)
 {
-	int min_chunk, chunk;
 	int enc_len, utf8_len;
 	char max_len_s[128];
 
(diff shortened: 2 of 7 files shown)


src/idle.c
```diff
@@ -66,6 +66,8 @@ int idle_requested(struct fmt_main *format)
 	if (strstr(format->params.label, "-opencl"))
 		return 0;
 #endif
+	if (strstr(format->params.label, "-ztex"))
+		return 0;
```
solardiz: Could check `#ifdef HAVE_ZTEX` here for consistency with what we do for OpenCL, but that's optional. We're not doing it consistently across the codebase anyway.

src/john.c
```diff
@@ -1721,7 +1729,9 @@ static void john_run(void)
 				                           &database);
 			else
 				test_db = &database;
+			self_test_running = 1;
 			where = fmt_self_test(database.format, test_db);
+			self_test_running = 0;
```
solardiz: We could possibly have this variable in `formats.[ch]` and set/reset it from within `fmt_self_test()`. That would cover other possible calls to `fmt_self_test()` from elsewhere. But I'm OK with the current way of doing it as well.

src/john.c
```diff
@@ -1813,6 +1823,11 @@ static void john_run(void)
 		if (trigger_reset)
 			database.format->methods.reset(&database);
 
+		if (!(options.flags & FLG_STDOUT) && john_main_process) {
+			john_log_format2();
+			log_flush();
+		}
```
solardiz: FWIW, I've confirmed that the `if` condition here is effectively the same as the one we use for calling `john_log_format()` above - it is.

I'm not happy about the `john_log_format` vs. `john_log_format2` function names, but I don't readily have a better suggestion.

solardiz:
In my testing, `-test -mask` does in fact work correctly now, however this PR caused other issues, such as timeouts with some non-mask tests. I've e-mailed @Apingis about those.

magnumripper:
> I guess magnum could want to review this as well before we merge

Too late now, but it looks good to me. Sooner or later I'll likely move the `self_test_running` in the way you mentioned but that's mostly cosmetic. More importantly I think I recall I've been mis-using the `bench_running` in weird ways in Jumbo for meaning benchmark and/or self-test even though the distinction is often valuable. I should review all uses of `bench_running` and possibly change some of them to `self_test_running` (and perhaps some to `(bench_running || self_test_running)`).

solardiz:
Great. FWIW, in this case the motivation to split `self_test_running` and `bench_running` was to have the self-test relatively quick (fewer candidate passwords) yet have the benchmark more closely represent speeds seen in actual cracking.

magnumripper:
I made an overhaul of `bench_running` and `self_test_running` in 9cc4e0193. Now `formats.h` can be sourced to get access to `self_test_running`, `benchmark_running` and `bench_or_test_running` (the last is a macro).

Issue #3122, opened 2018-01-24
ARMv7 problems


claudioandre-br:
### Synopsis
- Since the changes introduced by BE/alignment patches, armhf is malfunctioning.

```
Testing: as400-ssha1, AS400-SaltedSHA1 [sha1(utf16be(space_pad_10(uc($s)).$p)) (IBM AS/400 SHA1) 32/32]... *** stack smashing detected ***: ../run/john terminated
./build.sh: line 124: 21675 Aborted                 (core dumped) "$JTR_BIN" -test=0 --format=dynamic
```

```
Using default input encoding: UTF-8
Loaded 1 password hash (sha512crypt, crypt(3) $6$ [SHA512 32/32 OpenSSL])
Will run 4 OpenMP threads
Cost 1 (iteration count) is 5000 for all loaded hashes
./build.sh: line 124: 21686 Bus error               (core dumped) "$JTR_BIN" tests.in --format=sha512crypt --mask=jo?l[n-q]
```

claudioandre:
BTW: it is "Applied Micro Circuits Corporation (APM)" (my best guess)

```
processor	: 3
BogoMIPS	: 100.00
Features	: fp asimd evtstrm
CPU implementer	: 0x50
CPU architecture: 8
CPU variant	: 0x0
CPU part	: 0x000
CPU revision	: 1
```

claudioandre-br:
I do not remember this (I cant see a complete build history):

```
In file included from /usr/include/string.h:635:0,
                 from dynamic_fmt.c:54:
In function ‘memcpy’,
    inlined from ‘salt_hash’ at dynamic_fmt.c:2467:2:
/usr/include/arm-linux-gnueabihf/bits/string3.h:53:10: warning: call to __builtin___memcpy_chk will always overflow destination buffer
   return __builtin___memcpy_chk (__dest, __src, __len, __bos0 (__dest));
```

claudioandre-br:
Seems fixed.

claudioandre-br:
Follow-up 
* (At least) one of the build machines is a HP Moonshot

PR #3430, opened 2018-10-15
Dynamic: Change an erratic unsigned long to uint64_t. See #3122.

magnumripper:


solardiz:
magnum, what makes you think this is supposed to always be 64-bit rather than whatever is most efficient on a given platform (but not less than 32 bits)?

src/dynamic_fmt.c
```diff
@@ -2455,14 +2455,14 @@ static void *get_salt(char *ciphertext)
...
 	if (!salt) return 0;
 	if ( (curdat.pSetup->flags&MGF_SALTED) == 0)
 		return 0;
 
 	// salt is now a pointer, but WORD aligned.  We remove that word alingment, and simply use the next bits
 #if ARCH_ALLOWS_UNALIGNED
-	H = *((unsigned long*)salt);
+	H = *((uint64_t*)salt);
 #else
 	memcpy(&H, salt, 8);
```
solardiz: Should use `sizeof(H)` in place of 8 here.

solardiz:
Oh, noticed the `memcpy()`. Yes, the code was buggy as it was. OTOH, a comment near the hash function formula suggests that this code was using pointers as salts? If so, casting to `uint64_t` would be wrong in that it can be larger than a pointer. We need to figure out if pointers are in fact being cast to integers here, and be careful with that if so (use `uintptr_t`, which I am already using in yescrypt code, which we'll have in the tree anyway). Also need to ensure the format's reported `salt_size` is correct.

magnumripper:
The memcpy and the comment about word alignment made me think it should be uint64_t. On the other hand I believe `long` is 32-bit on Windows - and Jim wrote this code. Also, looking again at the few lines after this, I think we can safely use `uint32_t` and we should use `sizeof` regardless.

magnumripper:
We only use 29 bits of it anyway:
```c
	return ( (H^(H>>9)) & (SALT_HASH_SIZE-1) );
```

solardiz:
I think `uint32_t` is wrong. On 64-bit big-endian, it will end up taking upper 32 bits of 64-bit pointers, likely resulting in the same salt hash value for different salts.

solardiz:
There are also places in that file where `pFmt->params.salt_size` is set to 0. While maybe (just maybe) the salt hash function doesn't matter in that case, strictly speaking it is supposed to always return 0 in that case. Should we check for it explicitly? How, given that `pFmt` isn't available from that function? Perhaps that's an issue we should discuss separately from merging a fix for the original issue.

magnumripper:
So I now learned about `uintptr_t` and it looked like it was straight-forward - is the current version OK?

Regarding a salt_size of 0, that would be a separate bug but it appears it's taken care of somewhere - I tried the following and tested all formats:
```diff
diff --git a/src/formats.c b/src/formats.c
index 73efe6da3..60192ce65 100644
--- a/src/formats.c
+++ b/src/formats.c
@@ -739,6 +739,10 @@ static char *fmt_self_test_body(struct fmt_main *format,
 		    format->methods.source(ciphertext, binary)))
 			return "source";
 
+		if ((unsigned int)format->methods.salt_hash(salt) &&
+		    !format->params.salt_size)
+			return "salt_hash mismatch with salt_size";
+
 		if ((unsigned int)format->methods.salt_hash(salt) >=
 		    SALT_HASH_SIZE)
 			return "salt_hash";
```

solardiz:
> Regarding a salt_size of 0, that would be a separate bug but it appears it's taken care of somewhere - I tried the following and tested all formats:

Great. I suggest that you actually add that test, perhaps with a separate PR. Thank you!

PR #3432, opened 2018-10-16
various ztex improvements

Apingis:
- Use of verbosity. With -v=4 it prints mask size, keys_per_crypt on ```reset()```. With -v=5 it prints every ```crypt_all``` call with result count.
- Indexing the list of results (```struct task_result_list```) collected in ```crypt_all()```. With ```--test```, ```--test_full=1``` it sometimes (on a pseudo-random fashion) returns large number of results, sometimes >60K in 1 crypt_all, unless indexed it looks like it hangs (takes tens of seconds of CPU time to process).
- Reviewed parameters in some ztex format to better fit tests. Still there're questions - if tests properly represent various real uses and corner cases? And it also raises questions on how tests operate, e.g. for sha512crypt-ztex ```test-full=1``` runs crypt_all with max.number of keys 10 times, for Drupal7-ztex 4 times, for phpass-ztex 32 times.

### DIFF
--- a/src/ztex/device_format.c
+++ b/src/ztex/device_format.c
@@ -1,5 +1,5 @@
 /*
- * This software is Copyright (c) 2016 Denis Burykin
+ * This software is Copyright (c) 2016-2018 Denis Burykin
  * [denis_burykin yahoo com], [denis-burykin2014 yandex ru]
  * and it is hereby released to the general public under the following terms:
  * Redistribution and use in source and binary forms, with or without
@@ -16,11 +16,13 @@
 #include "../memory.h"
 #include "../misc.h"
 #include "../options.h"
+#include "../params.h"
 
 #include "jtr_device.h"
 #include "task.h"
 #include "jtr_mask.h"
 #include "device_bitstream.h"
+#include "device_format.h"
 
 // Problem: Inclusion of ztex.h results in inclusion of libusb-1.0.h
 // which includes more stuff, on Cygwin that results in redefinition
@@ -67,13 +69,16 @@ struct task_list *task_list;
 struct fmt_params *jtr_fmt_params;
 struct device_bitstream *jtr_bitstream;
 struct list_main *jtr_devices_allow;
+int jtr_verbosity;
 
 void device_format_init(struct fmt_main *fmt_main,
-		struct device_bitstream *bitstream, struct list_main *devices_allow)
+		struct device_bitstream *bitstream, struct list_main *devices_allow,
+		int verbosity)
 {
 	jtr_fmt_params = &fmt_main->params;
 	jtr_bitstream = bitstream;
 	jtr_devices_allow = devices_allow;
+	jtr_verbosity = verbosity;
 
 	struct list_entry *entry;
 	int found_bad_sn = 0;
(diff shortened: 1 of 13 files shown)


src/ztex/device_format.c
```diff
@@ -67,13 +69,16 @@ struct task_list *task_list;
 struct fmt_params *jtr_fmt_params;
 struct device_bitstream *jtr_bitstream;
 struct list_main *jtr_devices_allow;
+int jtr_verbosity;
```
solardiz: Global variables not used from outside of a translation unit should be declared `static`. I doubt all of these are used from elsewhere.

src/ztex/task.c
```diff
@@ -373,23 +418,49 @@ void task_result_execute(struct task_list *task_list,
...
 	}
 }
 
 
+void task_list_create_index(struct task_list *task_list)
+{
+	struct task *task;
+	for (task = task_list->task; task; task = task->next)
+		task_result_list_create_index(&task->result_list);
+}
```
solardiz: Is this function/loop a no-op (never calls `task_result_list_create_index`) when `crypt_all` would return 0 (no matches)? I hope so, I just want this discussed/confirmed explicitly.

src/ztex/task.c
```diff
@@ -23,13 +23,50 @@
...
+		list->index = NULL;
+	}
+}
+
+
+static void task_result_list_create_index(struct task_result_list *list)
+{
+	if (list->index) {
+		MEM_FREE(list->index);
+		list->index = NULL;
```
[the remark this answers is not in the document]
solardiz: Hmm, I guess the explicit setting to `NULL` actually makes sense here - as annotation that we actually depend on that. `MEM_FREE` setting its argument to `NULL` is more of a just-in-case precaution. So OK to leave things as they are.

src/ztex/task.c
```diff
@@ -23,13 +23,50 @@
...
+	}
+	if (!list->count)
+		return;
+
+	list->index = mem_alloc(list->count * sizeof(struct task_result *));
```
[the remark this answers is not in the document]
Apingis: We already have dynamic memory allocation on every match (to be precise, on every application-level packet received from a board, be it a packet with a match or e.g. PROCESSING_DONE).
I've decided to allocate memory dynamically because it has to allow every candidate be a match (and it sometimes happen in tests) and the number of candidates is high (e.g. in descrypt 140 M/board, in md5crypt >0.5 M/board).

What could be an alternative approach? Some applications allocate a big chunk of memory and then perform management of allocated memory on its own. Because of application specifics such approach might be more efficient. In this application, memory management seem doesn't consume significant fraction of resources, I decided to use a ready-to-go allocator.

In this particular case, I think we can skip allocation and creation of index if the number of matches is low (e.g. 1-5).

src/ztex/task.c
Apingis: I think it would look better to skip indexing operations when ```crypt_all()``` would return 0, because on the subsequent ```crypt_all()``` it drops existing ```task_list``` (with ```task_result```'s and their indices) and creates new ```task_list``` (with new portion of data).

solardiz:
This PR passed my tests so far, thanks!

PR #3440, opened 2018-10-24
Tweak error/warning handling in gcc-8

magnumripper:
See https://github.com/magnumripper/JohnTheRipper/pull/3439#issuecomment-432281202

Autoconf:
- Change  `-Wno-stringop-truncation` to `-Wno-error=stringop-truncation`
- Add `-Wno-error=format-overflow`.

NOTE that this make the warnings reappear - it just stops build from failing due to `--enable-werror`. Is this the way to go? I'd rather not suppress the warnings. Comments appreciated.

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -6135,58 +6135,6 @@ ac_compiler_gnu=$ac_cv_c_compiler_gnu
 
 
 
-   warn="-Wstringop-truncation"
- nowarn="-Wno-stringop-truncation"
- if test 1 -gt 0; then :
-  { $as_echo "$as_me:${as_lineno-$LINENO}: checking if $CC supports $nowarn" >&5
-$as_echo_n "checking if $CC supports $nowarn... " >&6; }
-fi
-  ac_ext=c
-ac_cpp='$CPP $CPPFLAGS'
-ac_compile='$CC -c $CFLAGS $CPPFLAGS conftest.$ac_ext >&5'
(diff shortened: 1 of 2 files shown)


src/configure
```diff
@@ -6135,8 +6135,58 @@ ac_compiler_gnu=$ac_cv_c_compiler_gnu
...
+$as_echo_n "checking if $CC supports -Werror... " >&6; }
+fi
+  ac_ext=c
+ac_cpp='$CPP $CPPFLAGS'
+ac_compile='$CC -c $CFLAGS $CPPFLAGS conftest.$ac_ext >&5'
+ac_link='$CC -o conftest$ac_exeext $CFLAGS $CPPFLAGS $LDFLAGS conftest.$ac_ext $LIBS >&5'
+ac_compiler_gnu=$ac_cv_c_compiler_gnu
+
+  ac_saved_cflags="$CFLAGS"
+  CFLAGS="-Werror -Werror"
```
solardiz: The repeated `-Werror -Werror` looks weird.

magnumripper: Right, lulz, apparently we're using `-Werror` to test if `-Werror` is supported 🤣 

It's a peculiar result of the "compiled" macros, but I'm not going to touch it as long as everything seems to work (also it's unrelated to this PR).

magnumripper:
Bots accepted it. I'm merging, we'll see how it holds up in the long run.

PR #3445, opened 2018-10-26
More hashcat rules

magnumripper:


.editorconfig
```diff
@@ -247,7 +247,7 @@ indent_size = 8
 [vax.[ch]]
 indent_size = 8
 
-[wordlist.[ch]]
```
frank-dittrich: This seems unrelated. Was it intended?

magnumripper: Yes, it was "needed" for the little change in `wordlist.c`. Our version of that file is so different from the original I decided to simply use the "jumbo indentation" of just 4, rather than trying to break up lines just making them less readable.

Issue #3444, opened 2018-10-26
Add shared code for entropy tests


magnumripper:
See https://github.com/hashcat/hashcat/issues/1729#issuecomment-433291076

As mentioned in that comment, a cold (really proper test) version could be used for detecting bogus input data. We might also want a shared hot (as fast as possible) one for post-crypt rejection.

magnumripper:
https://github.com/jj1bdx/ent seems to be source code for the classical `ent(1)` tool.

solardiz:
FWIW, rather than or in addition to trying to compute Shannon entropy (and strictly speaking, failing to do that), we may also compute the "expected different" like e.g. passwdqc does. In my experience, the "expected different" (for a properly tuned symbol size) is a more useful metric, allowing for direct comparison of what's observed against what's expected.

magnumripper:
I need to read up on that, and experiment a little. Also, I guess the requirements are very different depending on use case:

The original hashcat issue is one thing: Fairly good size of supposedly encrypted blob, in this case obviously not encrypted. Also, at that stage we can opt to just warn, no harm done.

Trying to guess correctness of a short decrypted message (like one or two AES blocks) is a whole different beast: Less data, and the outcome of a failed test could be a false negative. We hates that!
I recall our past struggle to come up with fast yet reliable tests for DMG.

magnumripper:
> FWIW, rather than or in addition to trying to compute Shannon entropy (and strictly speaking, failing to do that), we may also compute the "expected different" like e.g. passwdqc does. In my experience, the "expected different" (for a properly tuned symbol size) is a more useful metric, allowing for direct comparison of what's observed against what's expected.

I've finally had some time to look into this. Yes, the "expected different" seems to be a good and KISS measure. When you say "symbol size" you mean what's called `charset` in the code below, right? Typically 256 for our use.

```c
#define FIXED_BITS                      15

typedef unsigned long fixed;

static int expected_different(int charset, int length)
{
        fixed x, y, z;

        x = ((fixed)(charset - 1) << FIXED_BITS) / charset;
        y = x;
        while (--length > 0)
                y = (y * x) >> FIXED_BITS;
        z = (fixed)charset * (((fixed)1 << FIXED_BITS) - y);

        return (int)(z >> FIXED_BITS);
}
```

Then we have the `analyze.c` you sent me. I'm not sure what to make of that.
```
$ dd if=/dev/urandom bs=1 count=16 | ./analyze 
n = 16
[...]

PR #3452, opened 2018-11-07
format openssl-enc. Implement ascii percentage check.

julianjm:
Sometimes you don't know any part of the plaintext, but the plaintext is mostly ascii characters. 
With this patch, you can define the ascii threshold.

### DIFF
--- a/src/openssl_enc_fmt_plug.c
+++ b/src/openssl_enc_fmt_plug.c
@@ -88,6 +88,7 @@ static struct custom_salt {
 	int kpa;
 	int datalen;
 	unsigned char kpt[256];
+	int asciipct;
 	unsigned char data[1024];
 	unsigned char last_chunks[32];
 } *cur_salt;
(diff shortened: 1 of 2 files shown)


src/openssl_enc_fmt_plug.c
```diff
@@ -265,27 +272,57 @@ static void *get_salt(char *ciphertext)
...
 
+static int countascii(unsigned char *data, int len) {
+	int nascii = 0;
+	int c;
+	for(c=0; c<len; c++) {
+		if (data[c]==0x0a || (data[c]>=32 && data[c]<127))
+			nascii++;
+	}
+	return nascii;
+}
```
magnumripper: You should probably count `0x0d` as well, in case the data has Windows' CRLF line endings.

While at it, please add some air around operators, as in `for (c = 0; c < len; c++)`. That's our code style although I can't remember where it's documented, if at all.

magnumripper:
I like the idea. This is somewhat related to #3444

src/openssl_enc_fmt_plug.c
julianjm: Found it... I think it's ok now

magnumripper: Almost. I added further minor tweaks in a commit using GitHub's online editor. That was cool, although somewhat scary.

magnumripper:
Squashed and merged. Thanks!

PR #3465, opened 2018-11-22
sunMD5: define an OpenMP variable data scope

claudioandre-br:
See lines 529 and 531. Why there is a difference? Commit 2a63d1b54f8e5e7a22b310d6baf27dba82466a79 says nothing.
* Intel do not handle the "const-qualified type"? So, "count" should be added too (L548)? I expect a failure otherwise.

*****
Even better, why not remove the `#ifdef __INTEL_COMPILER` and define scope for all variables?

Reasoning:
gcc 9 is, for a while, unhappy with the source code.

### DIFF
--- a/src/sunmd5_fmt_plug.c
+++ b/src/sunmd5_fmt_plug.c
@@ -532,9 +532,9 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 #endif // SIMD_COEF_32
 #else
 #ifdef SIMD_COEF_32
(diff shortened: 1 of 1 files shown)


claudioandre-br:
CI fails because gcc compiler do not accept to be informed about something it already knows. clang and Mac work fine.

Still not clear to me if it works on Intel.

magnumripper:
Not sure why icc requires it while gcc can't stand it. I usually just do `#pragma omp parallel for` and nothing more., leaving it to the compiler. I suppose that might hide bugs though.

PR #3470, opened 2018-11-23
Colorize output for cracked administrator passwords.

magnumripper:


roycewilliams:
Interesting! I know this is quick and dirty, but it matches strings like "nonroot"; could the matching for some of the strings be bounded by word boundaries? Also, 'toor' might be a good add.

magnumripper:
I thought of making it configurable. Have you seen "nonroot" in real life though? Or "toor" (I've only seen that as a password).

I use word boundaries for "adm" because I've seen false positives otherwise. I'm not sure we want to do so for root... what if someone has eg. "secretroot" or "hiddenroot" that is actually a root account? As long as we have uid 0 we go by that, but people often just use a shadow file without caring for the passwd file or "unshadow" tool. And then we don't have any uid!

roycewilliams:
'nonroot' was a contrived example to demonstrate unexpected partial matches (like the ones you're heading off for 'adm') 

'toor' is a stock BSD user - also UID 0, but often with an alternate (usually full-featured) shell.

I see your point about legit partial matches; it's a tricky one. Configurability may be a good way to go.

magnumripper:
Thanks, I'll add toor then. And I'll add a config later.

magnumripper:
BTW I should probably add SA as well, as an exact full match though. Perhaps DBA too? I guess I'd better just make it configurabe 😆

magnumripper:
1ac9c7f added toor and SA, both as exact matches but case insensitive.

Issue #3386, opened 2018-08-21
Warn about too few candidate passwords


solardiz:
It is very common for users to run -opencl and -ztex formats on small wordlists, or even to try "single crack" mode, not realizing they're leaving most hardware resources idle. Maybe we should have a check in shared code (in cracker.c?) where we'd print a warning when the count passed to `crypt_all()` is below `min_keys_per_crypt` (include both numbers in the warning message, e.g. "Warning: only %u candidate passwords left, which is below the minimum of %u needed to utilize the available hardware resources"). Maybe print it the first time only, or maybe print it each time the count is different from what was seen previously (such that there would be many of those in "single crack" mode, consistent with how extremely bad the resulting performance is, but only one for the last batch in wordlist mode as the count will stay the same across all salts), maybe up to a certain number of these messages printed (then print that "further warnings of this type are suppressed" or similar). Maybe the verbosity setting should affect whether/when this warning is printed.

Alternatively, we could implement this inside the -ztex formats for now, printing a warning when some of the hardware cores are left idle. But I think the problem is more generic.

magnumripper:
While on that page we could also want to make "batch mode" different when format is ZTEX or OpenCL. It could be confusing though.

Just thinking loud, another alternative would be to have batch mode enforce CPU format for single mode, so that `./john hashes -form:nt-opencl` would start with single mode on CPU despite the format `-opencl` option, then continue with wordlist and incremental using OpenCL as requested. This too could be confusing but pretty handy (and it could be explained with some informative message).

FWIW some early versions of OpenCL support for RAR3 was built into the existing CPU format (iirc there was only `-format:rar` and it would pick OpenCL if applicable). For single mode and small batches it would simply call the CPU version of `crypt_all()` instead (and I think the decision was made per call so a wordlist run would often use OpenCL for all of the session except the very last batch). It worked just fine but I abandoned the idea anyway because it was too confusing - and made it hard to run a CPU session alongside an OpenCL one in a controlled manner. But I digress. Nowadays we have many OpenCL formats that are so different from their CPU counterparts it wouldn't be possible at all to just call "the other" `crypt_all()` transparently.

magnumripper:
PoC code. This seems to induce a 2% speed drop for NT. Not sure what would be a good default for the `max_warn` variable? It could be a params.h or even john.conf setting if we wanted to. This code currently lacks `john_main_process` as a condition for printing (and maybe that's correct).
```diff
[...]

PR #3480, opened 2018-12-04
cracker.c: Warn about too few candidate passwords. See #3386

magnumripper:
botcheck

### DIFF
--- a/src/cracker.c
+++ b/src/cracker.c
@@ -50,6 +50,7 @@
 #include "recovery.h"
 #include "external.h"
 #include "options.h"
+#include "config.h"
 #include "mask_ext.h"
 #include "mask.h"
 #include "unicode.h"
@@ -60,6 +61,7 @@
 #include "john_mpi.h"
 #include "path.h"
 #include "jumbo.h"
+#include "opencl_common.h"
 #if HAVE_LIBDL && defined(HAVE_OPENCL)
 #include "gpu_common.h"
 #endif
@@ -95,6 +97,7 @@ static struct db_keys *crk_guesses;
 static uint64_t *crk_timestamps;
 static char crk_stdout_key[PLAINTEXT_BUFFER_SIZE];
 int64_t crk_pot_pos;
+static int kpc_warn, single_running;
 
 /* expose max_keys_per_crypt to the world (needed in recovery.c) */
 int crk_max_keys_per_crypt(void)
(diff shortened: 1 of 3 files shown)


src/cracker.c
```diff
@@ -818,6 +821,19 @@ static int crk_password_loop(struct db_salt *salt)
...
 		fp_fix_state();
 
+	if (kpc_warn && crk_key_index < crk_params->min_keys_per_crypt) {
+		static int last_warn_kpc;
+
+		if (!mask_increments_len && last_warn_kpc != crk_key_index) {
+			last_warn_kpc = crk_key_index;
+			fprintf(stderr,"Warning: %u: Only %d candidates left, minimum %d "
+			        "needed for performance.\n%s",
+			        NODE, crk_key_index, crk_params->min_keys_per_crypt,
```
solardiz: Don't we normally print the node number before any messages (so "%u: Warning: ...") and only when there's more than one node?

src/cracker.c
```diff
@@ -818,6 +821,19 @@ static int crk_password_loop(struct db_salt *salt)
 	if (fp_fix_state)
 		fp_fix_state();
 
+	if (kpc_warn && crk_key_index < crk_params->min_keys_per_crypt) {
+		static int last_warn_kpc;
+
+		if (!mask_increments_len && last_warn_kpc != crk_key_index) {
```
solardiz: Are we sure we want to always completely mute these when `mask_increments_len`? That's OK for first commit, but maybe we'd want to add some kind of specialized reporting for that case.

src/cracker.c
```diff
@@ -818,6 +821,19 @@ static int crk_password_loop(struct db_salt *salt)
...
+			last_warn_kpc = crk_key_index;
+			fprintf(stderr,"Warning: %u: Only %d candidates left, minimum %d "
```
solardiz: Missing space after "stderr,".

src/cracker.c
magnumripper: Yes, I actually have a pending commit that adds a `log_print(int verbosity, bool main_only, char *format, ...)` function with all that functionality built-in and that's why it's not here. That change is much harder to get right than I could ever anticipate. This PR won't be merged right now anyway because I've seen some caveats that I need to dig into.

src/cracker.c
magnumripper: Right, we'll see how that ends up in practice. The rationale right now is that no matter how ineffective might be, it's still better than doing a separate run for length 1, one for length 2 and so on until we get some real keyspace.

src/cracker.c
magnumripper: Fixed, thanks

PR #3486, opened 2018-12-05
Add support for RACF KDFAES hashes (v0.15)

bigendiansmalls:
Support for KDFAES RACF hashes.

This fixes https://github.com/magnumripper/JohnTheRipper/issues/2511 (Support RACF KDFAES hashes).

Note: This is a reference implementation. A usable, GPU implementation will be coming up in the future.

### DIFF
--- a/src/racf_kdfaes_fmt_plug.c
+++ b/src/racf_kdfaes_fmt_plug.c
@@ -1,0 +1,466 @@
+/*
+ * JtR format for cracking RACF KDFAES hashes (reference implementation).
+ *
+ * This software is Copyright (c) 2018, Bigendian Smalls <mainframe [at]
+ * bigendiansmalls.com> and it is hereby released to the general public under
+ * the following terms:
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted.
+ */
+
+#if FMT_EXTERNS_H
+extern struct fmt_main fmt_racf_kdfaes;
+#elif FMT_REGISTERS_H
+john_register_one(&fmt_racf_kdfaes);
+#else
+
+#include <stdint.h>
+#include <openssl/des.h>
+
+#ifdef _OPENMP
+#include <omp.h>
+#endif
+
+#define OMP_SCALE               1
+
+#include "arch.h"
+#include "misc.h"
+#include "common.h"
+#include "formats.h"
+#include "params.h"
+#include "options.h"
+#include "aes.h"
+#include "sha2.h"
+#include "hmac_sha.h"
+#include "memdbg.h"
+
+#define FORMAT_LABEL            "RACF-KDFAES"
+#define FORMAT_NAME             ""
+#define FORMAT_TAG              "$racf$*"
+#define FORMAT_TAG_LEN          (sizeof(FORMAT_TAG)-1)
+#define ALGORITHM_NAME          "KDFAES (DES + HMAC-SHA256/" ARCH_BITS_STR " + AES-256)"
+#define BENCHMARK_COMMENT       ""
+#define BENCHMARK_LENGTH        0
+#define PLAINTEXT_LENGTH        8
+#define CIPHERTEXT_LENGTH       96
+#define BINARY_SIZE             16
+#define SALT_SIZE               sizeof(struct custom_salt)
+#define BINARY_ALIGN            sizeof(uint32_t)
+#define SALT_ALIGN              1
+#define MIN_KEYS_PER_CRYPT      1
+#define MAX_KEYS_PER_CRYPT      1
+
+#define MAX_SALT_SIZE           16
+#define PAD_SIZE                64
+#define HEADER                  "E7D7E66D00018000"
+#define HEADER_LEN              (sizeof(HEADER)-1)
+#define HASH_OUTPUT_SIZE        32
+
+static const unsigned char a2e[256] = {
+	0,  1,  2,  3, 55, 45, 46, 47, 22,  5, 37, 11, 12, 13, 14, 15,
+	16, 17, 18, 19, 60, 61, 50, 38, 24, 25, 63, 39, 28, 29, 30, 31,
+	64, 79,127,123, 91,108, 80,125, 77, 93, 92, 78,107, 96, 75, 97,
+	240,241,242,243,244,245,246,247,248,249,122, 94, 76,126,110,111,
+	124,193,194,195,196,197,198,199,200,201,209,210,211,212,213,214,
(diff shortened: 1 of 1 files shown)


solardiz:
Why exactly is `mfact` limited to 16 bits here? The way it's encoded - as base-2 logarithm - allows to specify much larger values. So is the limit to 16 bits just an arbitrary detail/bug here, or is it part of the spec?

In fact, is there a spec? And if it is, what does it call these `mfact` and `rfact` things? So that we report them under proper names as our tunable costs.

magnumripper:
https://www.detack.com/_docs/media/19_1_1641358905.pdf calls them **memory factor** and **repetition factor**, which might be better for costs reporting. It also mentions security by obscurity as in "_Obscurity vs. security, would it make sense that intended changes in password crypto is based on standards and fully disclosed, instead of keeping it secret?_" indicating there's no public spec.

magnumripper:
I think https://community.ibm.com/community/user/ibmz-and-linuxone/blogs/ross-cooper/2021/09/02/thwarted-by-ibm-z-episode-3?communityKey=01ddfc0e-8a3c-4873-ad0b-76a90c1adcca calls rfact **rounds**.

magnumripper:
> https://www.detack.com/_docs/media/19_1_1641358905.pdf calls them **memory factor** and **repetition factor**

https://slideplayer.com/slide/13142977/ also call them that

magnumripper:
OTOH few people will not understand **_m_** and **_r_**

bigendiansmalls:
To answer your question, @solardiz - there is no published spec. All inferred / tested etc. by myself and a couple others.  There are max limits for mfact and rfact (memory factor / repitition factor). When I tested, it's been a few years, but no reason to believe it has changed, mfact min/max  were 8/16 and rfact min/max were 50/1000. 

Sorry for the delay reply, holiday weekend :) 

Let me know if you need anything further.  Cheers.

solardiz:
> https://www.detack.com/_docs/media/19_1_1641358905.pdf calls them **memory factor** and **repetition factor**

This slide deck is very interesting. The algorithm they partially describe on slide titled "The KDFAES Algorithm" looks _different_ from what we have! It also has special-case support for lengths above 8, whereas our format here is limited to length 8 and below. They also have a screenshot with their custom JtR format on slide "KDFAES Performance: Brute Force", where they presumably implemented the algorithm they had described with SIMD (unlike what we have here).

> https://slideplayer.com/slide/13142977/ also call them that

Specifically:

```
Memory factor (PMEM)
Repetition factor (PREP)
```

and:

```
[22 ](https://slideplayer.com/slide/13142977/79/images/22/DES+v.+KDFAES+%E2%80%93+execution+time.jpg)DES v. KDFAES – execution time
Algorithm – (MF,RF)
Test system run time
DES – Original
KDFAES – (8,50) *Default
KDFAES – (9,64)
KDFAES – (11,50)
KDFAES – (10,1000)
KDFAES – (16,50)
KDFAES – (16,1000)
```

Suggests we may want to use 8, 50 for our default benchmark?

> There are max limits for mfact and rfact (memory factor / repitition factor). When I tested, it's been a few years, but no reason to believe it has changed, mfact min/max were 8/16 and rfact min/max were 50/1000.

Thank you! I also wonder if we should report mfact as 8 or as whatever gets into the `mfact` field in our code now? That field's value is then encoded into the extended salt used in further computation. But it sounds like we should report just the original value as it was in the hash encoding?

> Sorry for the delay reply, holiday weekend :)

That was no delay at all.

solardiz:
> This slide deck is very interesting. The algorithm they partially describe on slide titled "The KDFAES Algorithm" looks different from what we have!

Here's copy-paste from that slide:

```
The KDFAES Algorithm
• Step #1: Calculate input for key derivation, Ki
Passwords: Ki[8] = Old DES password = DES(K=password, In=username); permits one-way migration of old DES
Password phrases: Ki[40] = SHA256(phrase) + length(phrase); no migration possible for DES phrases
• Step #2: Key derivation based on custom IBM version of PBKDF2-SHA256; 45008 rounds*
Kd = IBM-PBKDF2-SHA256(RANDOM_SALT[16],Ki,45008)
• Step #3: Encrypt, by using AES256, the user name with the derived key
Hash[32] = AES256(K=Kd, In=username) + RANDOM_SALT[16]
• The icing on the cake: It is quantum-safe  Both SHA256 and AES256 are believed to be safe.
*The algorithm includes 2 parameters (memory and repetition factors) which can be customized and get stored
within the hash. This allows the key derivation step to be future-proof, i.e., get slower. Not in use for now, but if
maxed out, the computation time is increased ~4500 times
Link: https://github.com/openwall/john/blob/bleeding-jumbo/src/racf_kdfaes_fmt_plug.c
```

I am puzzled - they describe something subtly different from what we have, yet refer to this very format file from this PR?

bigendiansmalls:
I'm familiar with that deck, as I co-wrote it :) 

The algorithm is custom but built on those primitives, as we stated in the deck. 

The algorithm supports passphrases as well, but I cannot recall if that was implemented in the jtr code or not. There are a few tweaks to implement passphrases vs the passwords that are there now, but not too many. I haven't dug back into it in a while.

solardiz:
> I'm familiar with that deck, as I co-wrote it :)

Oh, cool. Do you also have / could contribute that other JtR format referenced in the screenshot in there? Or was it contributed and we somehow lost it or decided not to merge? I don't recall.

> The algorithm supports passphrases as well, but I cannot recall if that was implemented in the jtr code or not.

Apparently not in code that actually got into JtR. Also we have no test vectors for lengths above 8 - perhaps you do?

bigendiansmalls:
Yea I believe I have this in a test repo somewhere, just never got around to finishing it. I’ll look it up next week when I’m back. It is subtly different in the hash metadata but otherwise same algorithm.

On May 25, 2025, at 12:32, Solar Designer ***@***.***> wrote:

﻿
[https://avatars.githubusercontent.com/u/18089797?s=20&v=4]solardiz left a comment (openwall/john#3486)<https://github.com/openwall/john/pull/3486#issuecomment-2907962134>

I'm familiar with that deck, as I co-wrote it :)

Oh, cool. Do you also have / could contribute that other JtR format referenced in the screenshot in there? Or was it contributed and we somehow lost it or decided not to merge? I don't recall.

The algorithm supports passphrases as well, but I cannot recall if that was implemented in the jtr code or not.

Apparently not in code that actually got into JtR. Also we have no test vectors for lengths above 8 - perhaps you do?

solardiz:
> I'm familiar with that deck, as I co-wrote it :)

This slide deck also gives GPU speeds for KDFAES, but neither us nor current public hashcat have that. More unreleased code?

Issue #3497, opened 2018-12-08
Introduce proper support for non-hashes


magnumripper:
WPAPSK might be the best example, but there are other non-hashes where we actually have a same-salt opportunity that we miss due to the mis-use of the format interface:

- [x] WPA-PSK: salt is ESSID.
- [x] WPA-PSK-PMK: same, although salt isn't used until post-processing so little gain. It does however share much code with WPA-PSK so it was done anyway.
- [x] oldOffice: salt will (or may) be reused when saving an altered document, so new verifiers but same salt. CPU format done, GPU format has internal mask so will be tricky.
- [x] Office: same as above.
- [ ] Apple Notes (see https://github.com/magnumripper/JohnTheRipper/issues/2999#issuecomment-536925527).
- [x] RAR (salts seem to be derived from some notion of time, same-salt not often seen in the wild).
- [ ] ...very likely more

The **real** salt in WPAPSK is the ESSID: Thousands of different handshakes (possibly from completely different networks) can be attacked at the cost of one, as long as they have the same ESSID.

The problem is we treat WPAPSK as a "salt only" non-hash, abusing the legacy format interfaces. I'm pretty sure there are no workarounds for that unless we change the interfaces. A bad side effect now is that JtR core do not see any "same salt" because from its POV, the salt in WPAPSK is much more data than the ESSID. For WPAPSK (and a few more IIRC) we have a workaround hack in that we sort salts by ESSID, and a kludge for not re-doing the PBKDF2 if `current_essid` == `last_essid`. This solves some of the problem but we still get incorrect/inefficient behaviour with Single mode.

@solardiz I'm not quite sure how to fix this. I guess we need to add a notion of `salt2/get_salt2()/set_salt2()` but when at it we could probably just name them `data/get_data()/set_data()` and over time migrate all non-hash formats to using those instead (with a few, eg. WPAPSK, also using `salt/get_salt/set_salt` for the **real** salt only).

magnumripper:
I think we can get away with moderate changes to the format struct and merely a comment change to the loader struct: All non-hashes has a binary_size of 0 [except some that put a data verifier in the binary, but we'll handle that better] so we just re-purpose the binary pointer!

That pointer will in case of FMT_BLOB point to a fmt_data struct describing a variable size data blob.
```c
/*
 * Flags for fmt_data.
 */
/* Blob portion is tiny-alloc. */
#define FMT_DATA_TINY                   0x00000001
[...]

PR #3498, opened 2018-12-09
Format struct change for non-hashes (do not merge)

magnumripper:
This is WIP and not likely to be merged soon, see #3497

* Format flag added: FMT_BLOB (requires a binary_size of 0).
* Format methods added: `data()` [as in `get_data()`], `cmp_all_data()`, `cmp_one_data()`.

Loader struct unchanged. For FMT_BLOB we just re-purpose the binary pointer.

### DIFF
--- a/src/formats.c
+++ b/src/formats.c
@@ -1646,6 +1646,11 @@ void *fmt_default_salt(char *ciphertext)
 	return ciphertext;
 }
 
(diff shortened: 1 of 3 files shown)


magnumripper:
This is obsoleted by an even more KISS solution, PR coming soon.

PR #3501, opened 2018-12-09
Assorted fixes

### LINKED ISSUE #3495: Update AMD OpenCL docs
This has not been applicable since the release of amdgpu-pro. It needs to be updated.

```
==== If you have an AMD GPU (or want to use CPU as OpenCL device)

- Up to Ubuntu 15.10 (<= 15.10). This will work on Intel hardware but warnings will be seen.
sudo apt-get install ocl-icd-opencl-dev opencl-headers fglrx-dev

- Ubuntu Xenial 16.04 LTS and above (>= 16.04)
The fglrx driver would not be supported in 16.04 (XServer 1.18 is not supported).
Canonical and AMD decided to use the new hybrid driver stack for 16.04. JtR OpenCL
users should stay on a supported release until the hybrid stack is available.
```

BTW: it seems CPU support no longer exits.

claudioandre-br:
Basically, set GWS to next LWS multiple larger than count for "my" formats.

Closes #3495

### DIFF
--- a/src/opencl_rawsha256_fmt_plug.c
+++ b/src/opencl_rawsha256_fmt_plug.c
@@ -607,9 +607,10 @@ static int crypt_all(int *pcount, struct db_salt *_salt)
 {
 	const int count = *pcount;
 	const struct db_salt *salt = _salt;
-	size_t gws = count;
-	size_t *lws = (local_work_size && !(gws % local_work_size)) ?
-		&local_work_size : NULL;
+	size_t gws, initial = 128;
+	size_t *lws = local_work_size ? &local_work_size : &initial;
+
+	gws = GET_MULTIPLE_OR_BIGGER(count, local_work_size);
 
 	//Check if any password was cracked and reload (if necessary)
 	if (num_loaded_hashes != salt->count)
--- a/doc/INSTALL-UBUNTU
+++ b/doc/INSTALL-UBUNTU
@@ -28,15 +28,22 @@ Please, read INSTALL for general information about installing John on your syste
 
     sudo apt-get install nvidia-opencl-dev
 
-==== If you have an AMD GPU (or want to use CPU as OpenCL device)
+==== If you have an AMD GPU (OpenCL support)
 
-    - Up to Ubuntu 15.10 (<= 15.10). This will work on Intel hardware but warnings will be seen.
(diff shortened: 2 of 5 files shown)


claudioandre-br:
This code (that enforces 128) is 4 years old. See  https://github.com/magnumripper/JohnTheRipper/commit/f5631161ae6362a6033bd6f4e340a1d2a0751745

1. the commits above are, basically, a `git revert`;
2. so, I prefer to keep the commits as is (and create a new one to use LWS=NULL);

I don't know for sure, but off the top of my head, there is a reason (at least, there was); related to GPU mask mode
I need to stress  it to remember.

claudioandre-br:
I'll migrate to circleCI.com. My Circle run is no longer connected to GitHub PRs.

No wait, I'm already on whateverCI.com (https://circleci.com/gh/claudioandre-br/JohnTheRipper/1352). I'll take a look.

claudioandre-br:
> For the raw formats you break that now and enforce 128 instead - any special reason?

Wait a second. What happens if one is using `__local` memory? What happens in a call of `get_local_id(0)`? I will look at it closely later, but, I'm confident I had a reason.

src/opencl_rawsha256_fmt_plug.c
```diff
@@ -607,9 +607,10 @@ static int crypt_all(int *pcount, struct db_salt *_salt)
 {
 	const int count = *pcount;
 	const struct db_salt *salt = _salt;
-	size_t gws = count;
-	size_t *lws = (local_work_size && !(gws % local_work_size)) ?
-		&local_work_size : NULL;
+	size_t gws, initial = 128;
+	size_t *lws = local_work_size ? &local_work_size : &initial;
+
+	gws = GET_MULTIPLE_OR_BIGGER(count, local_work_size);
```
claudioandre-br: It seems I picked a bad name. I will propose a patch to rename it do `GET_NEXT_MULTIPLE`.

magnumripper:
> > For the raw formats you break that now and enforce 128 instead - any special reason?
> 
> Wait a second. What happens if one is using `__local` memory? What happens in a call of `get_local_id(0)`? I will look at it closely later, but, I'm confident I had a reason.

Ah yes, that's a reason, I think my RC4 formats also need to enforce some LWS limits. Anyway, I'll approve this and we can change things later if at all needed.

PR #3505, opened 2018-12-12
OpenCL versus Fork validation

claudioandre-br:
I tested it like this:
```
rm -f ../run/pot; ../run/john -form:SHA256crypt-opencl alltests.in -incremental -max-run=30 -fork=2            #OK
rm -f ../run/pot; ../run/john -form:SHA256crypt-opencl alltests.in -incremental -max-run=30 -fork=2 -dev=0,1,2 #error 
rm -f ../run/pot; ../run/john -form:SHA256crypt-opencl alltests.in -incremental -max-run=30 -fork=2 -dev=6     #OK
rm -f ../run/pot; ../run/john -form:SHA256crypt-opencl alltests.in -incremental -max-run=30 -fork=2 -dev=6,7   #OK

rm -f ../run/pot; ../run/john -form:SHA256crypt-opencl alltests.in -incremental -max-run=30 --fork=3 -dev=1,2   #OK
rm -f ../run/pot; ../run/john -form:SHA256crypt-opencl alltests.in -incremental -max-run=30          -dev=1,2,3 #error 

rm -f ../run/pot; ../run/john -form:SHA256crypt-opencl alltests.in -incremental -max-run=30 --dev=gpu           #error
```

Error is shown only if I have more listed devices than forked instances (and I prefer this way, john still allows the user run advanced use cases.).

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -586,6 +586,8 @@ void opt_init(char *name, int argc, char **argv, int show_usage)
 	if (options.format && strcasestr(options.format, "opencl") &&
 	    (options.flags & FLG_FORK) && options.acc_devices->count == 0) {
 		list_add(options.acc_devices, "all");
+		/* Set a flag that JtR has changed the value of --devices. */
+		default_device_selected = 1;
 	}
 #endif
 
--- a/src/options.c
+++ b/src/options.c
@@ -585,7 +585,7 @@ void opt_init(char *name, int argc, char **argv, int show_usage)
 #if HAVE_OPENCL
 	if (options.format && strcasestr(options.format, "opencl") &&
 	    (options.flags & FLG_FORK) && options.acc_devices->count == 0) {
-		list_add(options.acc_devices, "all");
+		list_add(options.acc_devices, "best");
 		/* Set a flag that JtR has changed the value of --devices. */
 		default_device_selected = 1;
 	}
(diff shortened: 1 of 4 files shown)


src/options.c
```diff
@@ -585,7 +585,9 @@ void opt_init(char *name, int argc, char **argv, int show_usage)
 #if HAVE_OPENCL
 	if (options.format && strcasestr(options.format, "opencl") &&
 	    (options.flags & FLG_FORK) && options.acc_devices->count == 0) {
-		list_add(options.acc_devices, "all");
+		list_add(options.acc_devices, "gpu");
+		/* Set a flag that JtR has changed the value of --devices. */
+		default_device_selected = 1;
 	}
```
magnumripper: What if we only have accelerators, or only CPUs? Perhaps after adding gpu's you could check count again, if still zero try next preferable (accel), and then again (CPU).

claudioandre-br: OpenCL CPUs are in the past (the new AMD driver do not support CPUs). So, only Intel does(?)

> Perhaps after adding gpu's you could check count again, if still zero try next preferable (accel), and then again (CPU).

This is feasible. And it will happens only in the **default** situation.

magnumripper: The fact AMD stopped supporting them in their GPU drivers doesn't mean it's the past: Intel has good drivers now, pocl has it, beignet, macOS... Perhaps AMD plans to release a separate CPU driver/runtime.

claudioandre-br: Every time I tried to find any CPU implementation (to do CI), I was disappointed. Anyway, it is not important.

claudioandre-br:
I'm waiting CI. 

What about the name? Anyone? Something like `--device=preferable`, or something like that.

magnumripper:
`--device=best` perhaps? That can mean singular or plural, and we can modify this later to be sorted by speed

PR #3507, opened 2018-12-13
Sort OpenCL devices

claudioandre-br:
More testing is needed.

### DIFF
--- a/src/opencl_common.c
+++ b/src/opencl_common.c
@@ -87,7 +87,8 @@ size_t ocl_max_lws;
 static char opencl_log[LOG_SIZE];
 static int opencl_initialized;
 
-static char* opencl_get_dev_info(int sequential_id);
+static void load_device_info(int sequential_id);
+static char* get_device_capability(int sequential_id);
 static int find_valid_opencl_device();
 
 // Used by auto-tuning to decide how GWS should changed between trials.
@@ -174,6 +175,7 @@ int get_number_of_available_platforms()
 	return i;
 }
 
+/* Get the number of available devices (all the OpenCL devices) */
 int get_number_of_available_devices()
 {
 	int total = 0, i = 0;
@@ -184,22 +186,32 @@ int get_number_of_available_devices()
 	return total;
 }
 
+/*
+ * Get the total number of devices that were requested (do not count duplicates)
+ * --device=2,2 result that "one" device is really in use;
+ */
 int get_number_of_devices_in_use()
 {
 	int i = 0;
 
-	while (gpu_device_list[i++] != -1);
+	while (gpu_device_list[i] != -1)
+		i++;
 
-	return --i;
+	return i;
 }
 
+/*
+ * Get the total number of requested devices (count duplicates)
+ * --device=2,2 result that "two" devices will be used. E.g., to split tasks;
+ */
 int get_number_of_requested_devices()
 {
 	int i = 0;
 
-	while (requested_devices[i++] != -1);
+	while (requested_devices[i] != -1)
+		i++;
 
-	return --i;
+	return i;
 }
 
 int get_platform_id(int sequential_id)
@@ -437,7 +449,10 @@ static int get_if_device_is_in_use(int sequential_id)
 	return found;
 }
 
-static void get_opencl_environment()
+/* Load information about all platforms and devices available in the
+ * running system
+ */
+static void load_opencl_environment()
 {
 	cl_platform_id platform_list[MAX_PLATFORMS];
 	cl_uint num_platforms, device_pos = 0;
@@ -554,7 +569,7 @@ static int start_opencl_device(int sequential_id, int *err_type)
 
 	// Get the detailed information about the device
 	// (populate device_info[d] bitfield).
-	opencl_get_dev_info(sequential_id);
+	load_device_info(sequential_id);
 
 	// Get hardware bus/PCIE information.
 	get_pci_info(sequential_id, &ocl_device_list[sequential_id].pci_info);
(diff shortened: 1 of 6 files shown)


src/opencl_common.c
```diff
@@ -663,10 +687,30 @@ static void add_device_to_list(int sequential_id)
...
+/* Used below (inside add_device_type routine) to sort devices */
+static int comparator(const void *p1, const void *p2)
+{
+	const speed_sort_t *c1 = (const speed_sort_t *)p1;
+	const speed_sort_t *c2 = (const speed_sort_t *)p2;
+	int diff = (int)c2->value - (int)c1->value;
+	if (diff)
+		return diff;
+	return c1->index - c2->index;
+}
```
magnumripper: This will ensure that devices with same speed will end up sorted by device number, lowest first, right? I like it.

claudioandre-br: Yep. Well, this is how JtR proper works, I just keep it.

magnumripper:
Tried it out on my Macbook, worked fine except we still have this:
```
$ ../run/john all_tests.in -form:ntlmv2-opencl -dev=best -w
Loaded 6 password hashes with 6 different salts (ntlmv2-opencl, NTLMv2 C/R [MD4 HMAC-MD5 OpenCL])
Remaining 3 password hashes with 3 different salts
Error: To fully use the 2 devices requested, you must specify --fork=2
(see doc/README-OPENCL)
```
Do we now have some variable to detect that we used "best" and not a specific list? Then we can change this to "*Warning*: To fully use the 2 devices *available*" and not bail out in that case, while keeping the message as-is and bail if I had actually said `-dev=2,1`.

Also,
```
$ ../run/john -form:ntlmv2-opencl -test=0 -dev=xx
Error: --device must be numerical, or one of "all", "cpu", "gpu" and
"acc[elerator]".
```
We need to add "best" to the list as well. I'm not sure about "top", with the above fix we may be good without even having that option coded.

claudioandre-br:
> We need to add "best" to the list as well. I'm not sure about "top", with the above fix we may be good without even having that option coded.

I haven't added it because it seems like an hidden/internal option. To use "best", you just have to specify nothing. Anyway, if you want it, I will add.

claudioandre-br:
> Tried it out on my Macbook, worked fine except we still have this:

How many GPUs do you have? Do you have 2 OpenCL enabled GPUs?

magnumripper:
> > We need to add "best" to the list as well. I'm not sure about "top", with the above fix we may be good without even having that option coded.
> 
> I haven't added it because it seems like an hidden/internal option.
> 
> * To use "best", you just have to specify nothing (you have to be using using `--fork`).
> * To use "top", you just have to specify nothing (only WITHOUT `--fork`).
> 
> Anyway, if you want it, I will add.

Perhaps not, then - no need to list things not needed to type!

magnumripper:
> > Tried it out on my Macbook, worked fine except we still have this:
> 
> How many GPUs do you have? Do you have 2 OpenCL enabled GPUs?

Yes, 1 CPU and 2 GPUs although one of the GPUs is just the Intel HD Graphics. OTOH it's not a lot slower than the AMD (this is a laptop so no monster stuff in it).

magnumripper:
> > > > Tried it out on my Macbook, worked fine except we still have this:
> > > 
> > > 
> > > How many GPUs do you have? Do you have 2 OpenCL enabled GPUs?
> 
> > Yes, 1 CPU and 2 GPUs although one of the GPUs is just the Intel HD Graphics. OTOH it's not a lot slower than the AMD (this is a laptop so no monster stuff in it).
> 
> So, the software did the right thing. What should I do? A solution could be:
> 
> * remove "top" and
> * "best" with fork should do what it does now
> * "best" without fork should do what "top" does

Yes, almost - but on super, "best" is 5 devices. If I use it without fork, it should pick the Titan X and only give a brief warning there are more devices. But if I use it with `--fork=2` it should pick the first two from the list, `--fork=3` should pick the three first and so on.

claudioandre-br:
Top 3 are new behavior
```
rm -f ../run/*.pot; ../run/john ~/alltests.in -form:ntlmv2-opencl -dev=best -w                  #Dev=6
rm -f ../run/*.pot; ../run/john ~/alltests.in -form:ntlmv2-opencl -fork=2 -w                       #Dev=6,7
rm -f ../run/*.pot; ../run/john ~/alltests.in -form:ntlmv2-opencl -fork=2 -w  --dev=gpu     #Dev=6,7 WARNING

rm -f ../run/*.pot; ../run/john -form:SHA256crypt-opencl alltests.in -incremental -max-run=30 -fork=2            #OK Dev=6,7
rm -f ../run/*.pot; ../run/john -form:SHA256crypt-opencl alltests.in -incremental -max-run=30 -fork=2 -dev=6     #OK Dev=6
rm -f ../run/*.pot; ../run/john -form:SHA256crypt-opencl alltests.in -incremental -max-run=30 -fork=2 -dev=6,7   #OK Dev=6,7

rm -f ../run/*.pot; ../run/john -form:SHA256crypt-opencl alltests.in -incremental -max-run=30 --fork=3 -dev=1,2  #OK Dev=1,2,1

rm -f ../run/*.pot; ../run/john -form:SHA256crypt-opencl alltests.in -incremental -max-run=30          -dev=1,2,3 #error 
rm -f ../run/*.pot; ../run/john -form:SHA256crypt-opencl alltests.in -incremental -max-run=30 -fork=2 -dev=0,1,2  #error 
rm -f ../run/*.pot; ../run/john -form:SHA256crypt-opencl alltests.in -incremental -max-run=30 --dev=gpu           #error 
```

PR #3515, opened 2018-12-16
OpenCL: xSHA512 lacks proper binary_hash()

claudioandre-br:
It is a bug.


magnumripper:
But then you must change from `fmt_default_get_hash` to `get_hash_0`, `get_hash_1` and so on. Doesn't it just fail with this patch?

claudioandre-br:
> Doesn't it just fail with this patch?

My fault. 
***
I have no idea why I missed `binary_hash()` (the raw format has it).

Issue #3516, opened 2018-12-16
Autotune runs more than needed


claudioandre-br:
In the example below it runs 3 times.

```
$ rm ../run/*.pot; LWS=128 ../run/john -form=nt-opencl -mask:?l -min-len=7 -max-len=9 alltests.in -dev=2 --verb=5 --skip
```

How `reset()` is working now?

claudioandre-br:
So, this is a bug that seems to be in many GPU mask enabled formats. It is very low priority.
****

BTW: is this warning correct? If I have the necessary minimum, why was I warned?
```
Warning: Only 172288 candidates buffered, minimum 172288 needed for performance.
```

magnumripper:
I started to work on this problem in general a week ago or two but much more interesting work got my attention.

magnumripper:
FWIW that erroneous warning was fixed in b30c280eb

magnumripper:
This will automagically be revisited when we fix #3249 and #3697 later

magnumripper:
Closing in favor of #4012

PR #3518, opened 2018-12-16
OpenCL: self test using sane GWS and LWS

claudioandre-br:
See #3516

### DIFF
--- a/src/opencl_rawsha256_fmt_plug.c
+++ b/src/opencl_rawsha256_fmt_plug.c
@@ -331,7 +331,7 @@ static void tune(struct db_main *db)
 
 static void reset(struct db_main *db)
 {
-	static size_t saved_lws, saved_gws;
+	static size_t saved_lws, saved_gws, should_tune;
 
 	offset = 0;
 	offset_idx = 0;
@@ -346,24 +346,38 @@ static void reset(struct db_main *db)
 	//Adjust kernel parameters and rebuild (if necessary).
 	build_kernel();
 
-	if (!autotuned) {
+	if (!should_tune) {
 		/* Read LWS/GWS prefs from config or environment */
 		opencl_get_user_preferences(FORMAT_LABEL);
 
 		//Save the local and global work sizes.
 		saved_lws = local_work_size;
 		saved_gws = global_work_size;
+	}
 
-		//GPU mask mode in use, do not auto tune for self test.
-		//Instead, use sane defauts. Real tune is going to be made below.
-		if ((options.flags & FLG_MASK_CHK) &&
-		   !((options.flags & FLG_TEST_CHK) || (options.flags & FLG_NOTESTS)))
-			opencl_get_sane_lws_gws_values();
-
+	/*
+	 * First reset() call. Don't run autotune.
+	 *    -> If self test is running.
+	 *    -> If --skip-self-test is running.
+	 *    -> And if benchmark is NOT running.
+	 *   Instead, use sane defauts. Tune is going to run later/below.
+	 */
+	if (!should_tune && (self_test_running || options.flags & FLG_NOTESTS) &&
+	    !benchmark_running) {
+		opencl_get_sane_lws_gws_values();
 		tune(db);
 
+		//Tune later.
+		autotuned = 0;
+	} else if (!autotuned) {
+		//Retrieve LWS/GWS prefs saved on first round
+		local_work_size = saved_lws;
+		global_work_size = saved_gws;
+
+		tune(db);
 	} else if ((options.flags & FLG_MASK_CHK)) {
-		//Tune for mask mode.
+		//Tune for mask mode (for each mask change).
+		// auto-tune for eg. ?a and then a reset for re-tuning for ?a?a
 		local_work_size = saved_lws;
 		global_work_size = saved_gws;
 
@@ -375,6 +389,7 @@ static void reset(struct db_main *db)
 
 		create_clobj(global_work_size, self);
 	}
+	should_tune++;
 	hash_ids[0] = 0;
 	load_hash();
 }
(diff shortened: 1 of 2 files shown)


claudioandre-br:
This only affects a self test of a cracking session.

What will happen is:
* LWS=0 GWS=0 `?a` `reset` for tuning 
* LWS=x GWS=y `?a?a` `reset` for re-tuning (it is up to autotune code to retune or use x and y)
* LWS=x GWS=y `?a?a?a` `reset` for re-tuning (it is up to autotune code to retune or use x and y)

Are telling me I have to set LWS=0 GWS=0 to force a re-tune?

magnumripper:
I was just afraid you kept the auto-tuned values for first length and re-used it.

I'm thinking if `-lws` and `-gws` are supplied (by variables or options) they should always be used (except possibly for self-test, which could still use get_sane_defaults or whatever it's called). If LWS/GWS are *not* supplied, we need to re-autotune any time `reset()` is called.

Now, `reset()` could obviously recognize that nothing's changed and just give the same values again without re-tuning and that would be perfectly fine **but** it's much better if we can instead ensure it's not called in the first place! In this case it's mask.c that should never call `reset()` unless actually needed, but I think it sometimes does. It's fairly tricky to know whether internal things changed or not IIRC.

claudioandre-br:
Now I get it. The PR is wrong.

claudioandre-br:
I'm still looking for a solution. But, the idea is [what in a remove past was `if (db == NULL)`]:

```
if (_self_testing)
    get_small_sane_lws_gws_values() // Never self test using GWS=4M
    create_buffers() 
else
    tune(db)
```

magnumripper:
6ea7946f09 added a global `autotune_real_db` variable. If it's true, the db sent with `opencl_init_auto_setup()` was a real db and not a test db. However, this may still happen during a self-test call, so it mostly just tell us this is not a `--test` session.

We also have recently reviewed and now correct variables `self_test_running`, `benchmark_running` and `bench_or_test_running`.

My problem when I last looked at this (for nt-opencl) was that even using all the above it wasn't quite always possible to know whether to use `small_sane` or perform an autotune. I can't remember why and I'll look at it later.

claudioandre-br:
I'm looking at `self_test_running` and `benchmark_running` now. They look very promising.

magnumripper:
Here's a good debug line to place at top of `reset()` for ensuring things are what we think they are:
```c
	fprintf(stderr, "%s(), %s, %s\n", __FUNCTION__,
	        (db && db->real && db->real == db) ? "real db" : "test db",
	        self_test_running ? "self-test" :
	        benchmark_running ? "benchmark" : "real crack session");
```

**This is fine:**
$ ../run/john -test -form:nt-opencl
Benchmarking: NT-opencl [MD4 OpenCL]... **reset(), test db, self-test**
DONE
Note: This format may also be benchmarked using --mask (see doc/MASK).
Raw:	28504K c/s real, 48131K c/s virtual

**This too is fine:**
$ ../run/john scraped.sam -form:nt-opencl -w
Loaded 5294 password hashes with no different salts (NT-opencl [MD4 OpenCL])
Remaining 3470 password hashes with no different salts
**reset(), test db, self-test**
Note: This format may be a lot faster with --mask acceleration (see doc/MASK).
**reset(), real db, real crack session**
Local worksize (LWS) 256, global worksize (GWS) 32768 (128 blocks)
0g 0:00:00:01 DONE (2018-12-17 00:31) 0g/s 2175p/s 2175c/s 7548KC/s 123456..sss
Session completed

**This has one superflous reset() but I believe we do that according to format interface spec. However, we should not re-tune the second time because nothing changed. But it looks we are.**
$ ../run/john scraped.sam -form:nt-opencl -w -skip
Loaded 5294 password hashes with no different salts (NT-opencl [MD4 OpenCL])
Remaining 3470 password hashes with no different salts
**reset(), real db, real crack session**
Local worksize (LWS) 256, global worksize (GWS) 32768 (128 blocks)
Note: This format may be a lot faster with --mask acceleration (see doc/MASK).
**reset(), real db, real crack session**
Local worksize (LWS) 256, global worksize (GWS) 65536 (256 blocks)
0g 0:00:00:00 DONE (2018-12-17 00:31) 0g/s 8865p/s 8865c/s 30761KC/s 123456..sss
Session completed

**I need to debug the last session's double resets for length 0:**
$ ../run/john scraped.sam -form:nt-opencl -mask=?a -max-len=4 -v:4
Loaded 5294 password hashes with no different salts (NT-opencl [MD4 OpenCL])
Remaining 3470 password hashes with no different salts
**reset(), test db, self-test**
Local worksize (LWS) 64, global worksize (GWS) 768 (12 blocks)
**reset(), real db, real crack session** <-- length 0
**Internal mask not utilized (target: 1024)** <-- length 0
**Internal mask not utilized (target: 1024)** <-- length 0 again?
Local worksize (LWS) 256, global worksize (GWS) 32768 (128 blocks)
**reset(), real db, real crack session** <-- length 1
**Internal mask, multiplier: 95 (target: 1024)** <-- length 1
Local worksize (LWS) 256, global worksize (GWS) 131072 (512 blocks)
0g 0:00:00:01  (1) 0g/s 0p/s 0c/s 0C/s
0g 0:00:00:01 1.04% (2) (ETA: 00:35:43) 0g/s 75.39p/s 75.39c/s 261626C/s aa..aa
0g 0:00:00:01 1.05% (3) (ETA: 00:35:42) 0g/s 7181p/s 7181c/s 24918KC/s aaa..aaa
0g 0:00:00:01 1.06% (4) (ETA: 00:35:41) 0g/s 682279p/s 682279c/s 2367MC/s aaaa..aaaa
0g 0:00:00:01 N/A 0g/s 61430Kp/s 61430Kc/s 285640MC/s #u3:..#u3:
Session completed

magnumripper:
Anyway the problem is in mask.c and I'll look at it.

claudioandre-br:
The PR now:
* use sane LWS and GWS for self testing;
* skip tune during the 1st `reset()` if `--skip-self-test`;

Both are `reset()` responsability.

PR #3519, opened 2018-12-17
Comment about rawSHA256/512 rounds reversing

### LINKED ISSUE #2227: Reverse 4 rounds more in SHA-256 and SHA-512 OpenCL kernels
Just off Twitter:

```
<hashcat> New SHA2 meet-in-the-middle optimization quick writeup: https://hashcat.net/forum/thread-5778.html
<@solardiz> @hashcat So it's 57 steps for SHA-256, 73 for SHA-512, 61 for SHA-224, 77 for SHA-384. That's what we have in simd-intrinsics.c.
<@hashcat> @solardiz I made sure it's not in JtR, but i was looking in kernels/sha256_kernel.cl -- not there
<@solardiz> @hashcat Yeah. We also lack on-GPU mask mode for raw SHA-512, but we do have it for raw SHA-256, so makes sense to add this opt.
<@hashcat> @solardiz Good, it then explains why I couldn't see it
<@solardiz> @hashcat Thanks for reminding us to add it for OpenCL. Looks like Aleksey Cherepanov was first to find it http://www.openwall.com/lists/john-dev/2015/09/23/2
```

We also have these in CUDA - perhaps we don't care to optimize the CUDA code anymore (and it also lacks mask), but maybe add comments about possible yet unimplemented optimizations in there, along with a suggestion to look at and use the OpenCL kernels instead.

claudioandre-br:
Closes #2227

### DIFF
--- a/src/opencl/sha256_kernel.cl
+++ b/src/opencl/sha256_kernel.cl
@@ -75,6 +75,38 @@ inline void any_hash_cracked(
     }
 }
 
+/*
+ * Reverse 4 rounds more in SHA-256 and SHA-512 OpenCL kernels
+ *
+ * 1. To avoid data transfers from GPU to CPU, I'm using a Bloom filter;
+ * 2. It can produce false positives;
+ * 3. The best approach is to discard as many candidates as possible on GPU
+ *    and avoid costly transfers;
+ * 4. If I reverse rounds I will have less data to compare;
+ * 5. So I'll face more false positives;
+ * 6. That will impact performance.
+ *
+ * That said, for a future version, based on how many hashes were loaded to
+ * crack, we can:
+ * -> use a reversed sha_block() version when running a session with only a few
+ *    keys.
+ *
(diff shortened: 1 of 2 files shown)


claudioandre-br:
> Please remove the words " on super" from the two instances of the comment, because this is too specific to our current machines whereas the comment should be generic enough to stay relevant for years and not only to ourselves. 

In fact, I should change from super to GeForce GTX TITAN X. Otherwise, neither I can understand.

solardiz:
Perhaps to "Titan X Maxwell", since there's also a "Titan X Pascal".

claudioandre-br:
I don't have enough power to merge this one.

solardiz:
I also don't. Let's wait for CI or magnum.

magnumripper:
Have I messed something up with the build bots, or did they spontaneously break at the same time?

claudioandre-br:
> Have I messed something up with the build bots

In this case, GitHub did.

Issue #3517, opened 2018-12-16
Audit rules engine optimizations


magnumripper:
See #3468, https://github.com/magnumripper/JohnTheRipper/commit/15163f45fa95d3f68c2ae58971f233d706d1279f#commitcomment-31692841 and https://github.com/magnumripper/JohnTheRipper/commit/79559008ace3797ab6d5b6c4da1a4711261b820a#commitcomment-31692820

- [ ] Verify that commands like `aN` behaves correctly depending on the `FMT_TRUNC` flag.
- [ ] Verify that 15163f4 didn't add `-c` somewhere where it should be `-[:c]` (followed by `\p1[lu]`).

magnumripper:
(there's some more to it than the diff above, but that's the gist of it)

solardiz:
@magnumripper I think you picked a wrong example. I intentionally chose "Summer" and not the numbers, etc. For the numbers, etc., we assume we also have rules that test shorter ones, so testing truncated numbers would be redundant. This is not the case for words like "Summer".

I don't understand the patch you propose, but FWIW it needs braces around `db->format->params.flags & FMT_TRUNC`.

magnumripper:
> I think you picked a wrong example. 

OK, I thought you placed your comment near the rules you discussed. Anyway, that makes more sense. You mean we should only reject if we can't append anything even one letter of it.

magnumripper:
For the Summer case, the only sane way to get it the way you want is to revert that change. Personally I'd rather have it with `a6` but I can make a private rule set for that.

I should probably just ditch #3522. What it does is setting rules_max_length as well as `*` and related variables to `RULES_MAX_LENGTH` whenever `FMT_TRUNC` but while that often is logical and good, it sometimes isn't (eg. `'*` would not truncate DEScrypt at 8).

I think we need to discuss exactly how and when we should take FMT_TRUNC in consideration.

magnumripper:
Here's current behavior. @solardiz is there anything we want to change here? If not, I will add some of these details to the docs, and then I can go on with the OP possibly only reverting some rule changes.

```
->N	reject this rule unless length N or longer is supported
-<N	reject this rule unless length N or shorter is supported (--min-length)
```
The above refer to format's reported length, unless overridden with `--min/max-len` options. `FMT_TRUNC` is not considered. 
[...]

PR #3522, opened 2018-12-17
Rules engine: Fixes for truncating formats (like DEScrypt)

magnumripper:
For truncating formats (like DEScrypt), set max_length variable(s) to 125 (+/-1) instead of format's reported length. See #3517.

@solardiz note that this affects all of `*`, `-` and `+`. I can't think of any ill side effects, can you?

At `out_OK:`
- If `FMT_TRUNC` and not `--max-len`: The word is truncated to whatever is correct, then accepted.
- If `!FMT_TRUNC` or `--max-len`: The word is rejected if it was longer than max length.

Regardless, we also never reject or truncate a word that will later pass rules again (`--rules-stack`).

### DIFF
--- a/src/rules.c
+++ b/src/rules.c
@@ -81,8 +81,11 @@ int rules_stacked_number;
  */
 static int rules_line;
 
-static int rules_max_length = 0, min_length, skip_length;
-int hc_logic; /* can not be static. rpp.c needs to see it */
+static int rules_max_length, min_length, reject_length;
+static int rules_trunc_len;
+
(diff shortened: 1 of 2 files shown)


magnumripper:
Oh BTW the wordlist.c thing is purely refactoring "skip_length" -> "reject_length".

magnumripper:
> @solardiz note that this affects all of *, - and +. I can't think of any ill side effects, can you?

Oh I can now. This will break things like `'*`. Apparently we have nothing like that in any current rule.

PR #3527, opened 2018-12-18
Changes required to again build on VC

jfoug:
2 files required changing (after a lot of time getting things working again on VC).
There was also a execute bit set .c file checked in a month ago, fixed here.


jfoug:
The exec bit was not me this time, so  I claim at least 3 free screw ups ;)

PR #3529, opened 2018-12-18
logger.c has exe bit set per 679ae3c

jfoug:
Since this was not my bad, I claim right for at least 3 screw ups in the future ;)


jfoug:
This is a very safe PR.  Simply removing exec bits from logger.c

magnumripper:
Wow, it was actually your's truly. I think a USB key was involved...

jfoug:
On 12/18/2018 6:31 PM, magnum wrote:
>
>     I thought there was a pre-push hook that chmod cleared things? Did
>     that go away?
>
> Hooks needs to be installed locally by each one of us, and I can't 
> remember where we discussed/documented them so I'm running without 
> any. Was it on john-dev? We should have it on the wiki here.
>
>
This has been so long ago, i do not remember.  But since I was usually 
the target of shame, I did think there was some talk of hooks (and I 
thought they were there).  Things like exe bits, dos eols, etc, and a 
commit would fail if things were wrong.  I know we had some editor tool 
(I will have to re-install it).  I have none of the computers I had a 
year ago, everything is new, so I'm spending a large amount of time 
getting crap setup and usable again.

magnumripper:
LOL I found the hook now, by pure coincidence. It's in the base directory and called `.pre-commit.sh`. You activate it with `ln -s .pre-commit.sh .git/hooks/pre-commit`

PR #3539, opened 2018-12-19
updated the unused pre-built version of this file to its proper state

### LINKED ISSUE #3538: dynamic_big_crypt.c in unused is not up to date
This file should be copied into unused from the scr directory after a successful build, AFTER the *.cin file is updated.  They were updated (by me) a while back, removing the undef of SIMD defines when building under a BE system.   I failed to update the correctly built 'unused' file, and need to do this.

jfoug:
Fixes #3538


jfoug:
Simple fix. if there is a way to cancel Travis for this one, we probably should.  Travis will not test these changes at all, simply waste CPU cycles.

PR #3540, opened 2018-12-19
changed the structure packing code to be more generic

jfoug:
This was failing to be biddable in VC.  This method used in several other header files, so I simply used the same technique here.

### DIFF
--- a/src/diskcryptor_common.h
+++ b/src/diskcryptor_common.h
@@ -27,6 +27,9 @@ struct custom_salt {
 #define PKCS_DERIVE_MAX         (MAX_KEY_SIZE*2)
 #define CF_CIPHERS_NUM          7  // 7 types of ciphers possible
 
+#pragma pack(push)  /* push current alignment to stack */
+#pragma pack(1)     /* set alignment to 1 byte boundary */
+
 struct dc_header {
 	uint8_t  salt[PKCS5_SALT_SIZE]; /* pkcs5.2 salt */
 	uint32_t sign;                  /* signature 'DCRP' */
@@ -45,7 +48,9 @@ struct dc_header {
 	uint8_t  tmp_wp_mode;           /* data wipe mode */
 
 	uint8_t  reserved[1422 - 1];
-} __attribute__((packed));
+};
+
+#pragma pack(pop)
 
 extern struct fmt_tests diskcryptor_tests[];
 


solardiz:
@magnumripper Per my review, no, neither the original nor the revised version won't work right with `!ARCH_ALLOWS_UNALIGNED`. At least the `flags` field is misaligned, and accessing it should fail on those archs. We need to at least document that in a comment, and perhaps exclude this format (or is it two formats - CPU and OpenCL?) from `!ARCH_ALLOWS_UNALIGNED` builds.

solardiz:
Oh, and I think I didn't explicitly say how glad I am that @jfoug is active again. Very glad. Thank you, Jim!

magnumripper:
> @magnumripper Per my review, no, neither the original nor the revised version won't work right with `!ARCH_ALLOWS_UNALIGNED`. At least the `flags` field is misaligned, and accessing it should fail on those archs. We need to at least document that in a comment, and perhaps exclude this format (or is it two formats - CPU and OpenCL?) from `!ARCH_ALLOWS_UNALIGNED` builds.
> 
> But this can be done separately. It doesn't have to be in this same PR since the issue existed before.

Wouldn't that depend on just how such system would access the struct members though? IIRC we have something like `#pragma pack (1)` in wpapsk formats and they work just fine on Sparc and such. I guess we never access the larger-than-1 variables directly. I have no idea how this format reads things, I'm just trying to understand.

jfoug:
On 12/19/2018 12:59 PM, Solar Designer wrote:
>
> *@solardiz* requested changes on this pull request.
>
> Neither version is generic. I think we need conditional compiles, 
> continuing to use |__attribute__((packed))| with GCC. We should not 
> use pragmas anywhere in our source files outside of compiler version 
> checks.
>
>
I certainly see your point here.  The pragma pack syntax 'has' become a 
defacto working standard, but that does not mean it will work everywhere.

At one point, I thought we had some header which helped with packing 
syntax.  I think it should be found and looked at.  It may be that all 
of our aligned structure declarations can carry some JOHN_ALIGN(1) or 
some syntax that works properly, without spewing ifdef stuff all over 
the place (it will still be there, but hidden in a single header I think).

For now, I can easily revert this to a ifdef _MSC_VER set of logic.  But 
as the original file was submitted, it craps out in VC, so something 
does need to be done.  After that, we can look into removing the pragma 
pack coding, along with validating that the usage of these packed 
structures is not going to core some non-aligned system.

solardiz:
> Wouldn't that depend on just how such system would access the struct members though?

Definitely. If the code is careful not to read the (unaligned) multi-byte variables as such, we're fine.

> For now, I can easily revert this to a ifdef _MSC_VER set of logic.

I would also be fine with checking for `__GNUC__` (which will also apply to clang and icc, and that's right) and using `attribute`, and using the pragmas otherwise.

magnumripper:
@ivan-freed this is likely of interest for you

PR #3545, opened 2018-12-21
Fix structure packing in DiskCryptor stuff

ivan-freed:
This might fix https://github.com/magnumripper/JohnTheRipper/issues/3542 and address https://github.com/magnumripper/JohnTheRipper/pull/3540 as well.

### DIFF
--- a/src/diskcryptor_common.h
+++ b/src/diskcryptor_common.h
@@ -27,6 +27,17 @@ struct custom_salt {
 #define PKCS_DERIVE_MAX         (MAX_KEY_SIZE*2)
 #define CF_CIPHERS_NUM          7  // 7 types of ciphers possible
 
+#ifdef __GNUC__
+#define PACKED __attribute__ ((__packed__))
+#else
+#define PACKED
+#endif
+
+#ifndef __GNUC__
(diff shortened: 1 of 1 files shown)


ivan-freed:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=52991 could be relevant for us.

src/diskcryptor_common.h
```diff
@@ -27,6 +27,17 @@ struct custom_salt {
...
+#ifdef __GNUC__
+#define PACKED __attribute__ ((__packed__))
+#else
+#define PACKED
+#endif
+
+#ifndef __GNUC__
+#define PACKED
+#pragma pack(push,1)
+#endif
```
magnumripper: Perhaps move the lower one of these to the else clause of the upper one?

ivan-freed: I am not about this but perhaps the `#pragma pack(pop)` part needs to come after the structure definition (and thus needs to belong to a new `ifdef` block)?

ivan-freed:
@jfoug Does this PR help with getting things built in Visual C?

Issue #3417, opened 2018-10-05
stack smashing detected using dynamic format


jeroen-80:
Running the latest GIT code on up-to-date Ubuntu 16.04 LTS server. If a dynamic_mode is started in single mode, john crashes:
```
..
Node numbers 1-4 of 4 (fork)
Proceeding with single, rules:Single
Press 'q' or Ctrl-C to abort, almost any other key for status
2 0g 0:06:06:48 0.44% (ETA: 2018-12-01 16:54) 0g/s 8264p/s 8264c/s 189401KC/s <removed>
*** stack smashing detected ***: ./john terminated
*** Error in `./john': free(): invalid size: 0x00007fb78046f040 ***
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(+0x777e5)[0x7fb77ed907e5]
/lib/x86_64-linux-gnu/libc.so.6(+0x8037a)[0x7fb77ed9937a]
/lib/x86_64-linux-gnu/libc.so.6(cfree+0x4c)[0x7fb77ed9d53c]
./john[0x47367a]
./john[0x6bd84a]
./john[0x6c8c51]
./john[0x6c8f72]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf0)[0x7fb77ed39830]
./john[0x4057e9]
======= Memory map: ========
00400000-0091b000 r-xp 00000000 fc:03 55184516                           /opt/JohnTheRipper/run/john
00b1a000-00b1b000 r--p 0051a000 fc:03 55184516                           /opt/JohnTheRipper/run/john
00b1b000-00ba9000 rw-p 0051b000 fc:03 55184516                           /opt/JohnTheRipper/run/john
00ba9000-01f53000 rw-p 00000000 00:00 0
022d3000-557e4000 rw-p 00000000 00:00 0                                  [heap]
557e4000-7ca8ec000 rw-p 00000000 00:00 0                                 [heap]
7fb767efe000-7fb77616c000 rw-p 00000000 00:00 0
7fb778000000-7fb778021000 rw-p 00000000 00:00 0
7fb778021000-7fb77c000000 ---p 00000000 00:00 0
7fb77e02a000-7fb77e040000 r-xp 00000000 fc:03 21627405                   /lib/x86_64-linux-gnu/libgcc_s.so.1
[...]

Issue #2425, opened 2017-02-12
very bad (and decreasing) performance on mode "Single" in certain circumstances


p4tpr0:
I have a hash file where every line looks like this:

candidate:$dynamic_25$hash$salt

"candidate" is a very serious password candidate to test, with 50-60% success rate without even using mangling rules.
When using mode "single" on this big hash file with "SingleRetestGuessed = N" and no rules, I see a bad performance that decrease along the cracking session. Sample session:

```
$ time ./john --single=None --nolog --verbosity=1 pw-1M-real  --pot=pw-1M-real.pot
Using default input encoding: UTF-8
Loaded 1000000 password hashes with 1000000 different salts (dynamic_25 [sha1($s.$p) 128/128 AVX 4x1])
Press 'q' or Ctrl-C to abort, almost any other key for status
0g 0:00:00:02 50.00% (ETA: 16:24:20) 0g/s 0p/s 0c/s 0C/s
30769g 0:00:00:08 50.00% (ETA: 16:24:33) 3445g/s 5563p/s 5563c/s 5563C/s aoygm
48159g 0:00:00:18 50.00% (ETA: 16:24:53) 2544g/s 4108p/s 4108c/s 4108C/s agmys
61712g 0:00:00:29 50.00% (ETA: 16:25:15) 2062g/s 3341p/s 3341c/s 3341C/s aprzjx
82423g 0:00:00:52 50.00% (ETA: 16:26:01) 1557g/s 2523p/s 2523c/s 2523C/s cxox
93556g 0:00:01:08 50.00% (ETA: 16:26:33) 1357g/s 2201p/s 2201c/s 2201C/s yvoq
105281g 0:00:01:29 50.00% (ETA: 16:27:15) 1170g/s 1897p/s 1897c/s 1897C/s pfkwx
117217g 0:00:01:54 50.00% (ETA: 16:28:04) 1028g/s 1665p/s 1665c/s 1665C/s vpqu
132177g 0:00:02:29 50.00% (ETA: 16:29:15) 881.5g/s 1428p/s 1428c/s 1428C/s acwzhx
145131g 0:00:03:06 50.00% (ETA: 16:30:29) 776.4g/s 1257p/s 1257c/s 1257C/s airzx
193370g 0:00:06:35 50.00% (ETA: 16:37:27) 488.4g/s 791.9p/s 791.9c/s 791.9C/s amoxr
...
```

[...]

PR #3547, opened 2018-12-21
fixed dynamic overflows in single mode

jfoug:
ties to #3417  Possibly #2425 

NOTE, may not be ready to merge.  @magnumripper has worked out some debugging code, and I think there may still be an issue when single mode completes, and all the partial buckets are run.  I think the last array element may still exhibit the problem, and may take some additional work.  

HOWEVER, what we found was dynamic 'min_keys_per_crypt' was hard coded to 64.  This did not work all the time. Prime example was AVX2 md5.  Here SIMD32 was 8, and SIMD_PARA was 3. That means each SIMD block was 24.  24 does not evenly divide into 64.  Thus, the code would clear 64 buffers, BUT later code uses full buffer counts (which in this case were 72).  SO the last 8 buffers would get data appended to them, but NEVER properly reset between runs.  So those 8 buffers grew, until they started to overflow.

Since #2425 was listing things like dynamic, it is very possible that this could be the problem.  What @magnum experienced was the heap got corrupted, putting min_keys_per_crypt to 0, which was causing very slow performance (i.e. 1 candidate at a time, no buffering, but performing 72 crypt cracks.  So I hope this is at least 1 of the issues causing the bug which was reported.

### DIFF
--- a/src/dynamic_fmt.c
+++ b/src/dynamic_fmt.c
@@ -848,9 +848,7 @@ static void init(struct fmt_main *pFmt)
 	force_md5_ctx = curdat.force_md5_ctx;
 
 	fmt_Dynamic.params.max_keys_per_crypt = pFmt->params.max_keys_per_crypt;
-	fmt_Dynamic.params.min_keys_per_crypt = pFmt->params.max_keys_per_crypt;
-	if (pFmt->params.min_keys_per_crypt > 64)
-		pFmt->params.min_keys_per_crypt = 64;
+	fmt_Dynamic.params.min_keys_per_crypt = pFmt->params.min_keys_per_crypt;
 	fmt_Dynamic.params.flags              = pFmt->params.flags;
 	fmt_Dynamic.params.format_name        = pFmt->params.format_name;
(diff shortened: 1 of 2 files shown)


magnumripper:
This was a HARD one to debug!

magnumripper:
Bumping min. kpc doesn't help at all though. Single mode does run many many times during a session with less than min. kpc, and even more so after you bump it. And all modes may run the final batch incomplete.

I think you need to ensure that all functions process the same amounts (i.e. `count`), be it min. or max. or anything inbetween. The problem I saw was that some functions processed a hard-coded size of some multiple of SIMD_COEF while other functions only did `count`, leaving the rest uninitialized.

jfoug:
@magnumripper please merge this. I will start a new issue for cleaning up the straggler single buffers, and to better compute the 'min' buffer count size allowable per format.  The code in this PR 'is' the correct fix.  Yes, it can and will get additional fixes, but this should handle the one issue for sure, and I think this may well be why we were also experiencing severe single slowdowns (as seen within your testing).

claudioandre-br:
After the "approval", you can merge it yourself.

PR #3554, opened 2018-12-23
Fix a leak in cracker.c

claudioandre-br:
The crk_init rounine is called more than once.

539b8fd72a560a25c3f2ae5af7c9fcdce4204156 is not enough.

### DIFF
--- a/src/cracker.c
+++ b/src/cracker.c
@@ -215,6 +215,7 @@ void crk_init(struct db_main *db, void (*fix_state)(void),
 
 	if (db->loaded) {
 		size = crk_params->max_keys_per_crypt * sizeof(uint64_t);
+		MEM_FREE(crk_timestamps);
 		memset(crk_timestamps = mem_alloc(size), -1, size);
 	} else
 		crk_stdout_key[0] = 0;


claudioandre-br:
My debug session was not so complete. Anyway, it was single in a single and wordlist run.

magnumripper:
@solar I think this is a core bug, of sorts.

magnumripper:
Good stuff @claudioandre-br, even if your proposed patch was not quite it.

solardiz:
While core's single.c not calling crk_done() might indeed be a bug, I'm puzzled by @jfoug's introduction of c_cleanup() and calling it from crk_done(). (BTW, after magnun's fix here the call Jim added to single_done() becomes redundant.) I worry that this addition of c_cleanup() breaks external mode filters when used along with batch mode... oh, I just realized this is currently unsupported in terms of command line parsing: when `--external` is specified without other cracking modes, it replaces the batch mode with external mode. So we dodged that problem for now. But if we ever introduce a way to enable batch mode along with external filter, like `--batch --external=...`, I guess Jim's addition/use of c_cleanup() will hurt because we're not re-initializing the external mode filter as we proceed through the cracking modes that are part of batch mode.

Issue #3549, opened 2018-12-22
dynamic single mode issues


jfoug:
a recent PR fixed some serious issues in single when it came to dynamic (i.e. min count not always being a require integral of the SIMD buffer size).  However, there were at least 2 things still needing done.

1. when complete, single runs with fewer than min were being run, and it appeared the same buffer grow type bug was seen. Should be somewhat easy to fix, but has to be done without adding overhead within dyna.
2. add a min count on a per format basis. Right now, min-count is globally set for all dynamic formats, thus it is larger than it has to be for individual formats. It is done like the max size, where min is set large enough to handle all hash types (thus it is 2 to 12x larger than it really needs to be, for an individual hash).

jfoug:
#3556 should fix this bug (totally).  The other fix (#3547) was  also 100% needed and without it, performance would certainly suffer.

This fix handles the transition from single into wordlist mode, where there are many 'non-full' buffers.  It also handles ANY situation where crypt_all is called and the count is under sized.

I will close this bug once both of those PR's are into jumbo.

PR #3556, opened 2018-12-23
Dyna overflow2

jfoug:
Perform a full clean if m_count is < max hash count


jfoug:
This should fix  #3549 and ANY run of dynamic where crypt_all is called where key count is < max keys.

jfoug:
@magnumripper This should be 'THE' fix. It will force a complete full clean of all buffers any time crypt_all is called where key count is 'short'.

Note, the prior partial fix would have kept dyna from overflowing outside of the input buffers. There could have been some internal overflowing happening, within the transition from single into wordlist mode, but the main state would NOT have been damaged.  BUT that overflow could have spilled over, once the wordlist mode was started, and there may have been damaged buffers, thus there could have been false negatives, due to damaged input buffers.

Now, there should be NO overflowing of the input buffers at all.  During the 'run', when there are short count, the append type code WILL be performed for all buffers up to a SIMD*PARA boundary.  BUT the inputs will all be at zero, thus any appends will be normal 'valid' appends, bounds checked by the validity of the format itself.  The problem before, was the prior run's dirty buffers were being used on these > m_count input buffers (buffers < than m_count+(m_count+SIMD*PARA)%(SIMD*PARA).  Those buffers were not being cleaned, thus contained prior run input data, and appends could overflow.  Now, everything starts at 0 buffer length, thus are protected from overflow.

PR #3559, opened 2018-12-23
dynamic_types: fix non-SIMD compile error caused by c31b66d

magnumripper:
This replaces #3558 and should get Circle CI botcheck

### DIFF
--- a/src/dynamic_types.h
+++ b/src/dynamic_types.h
@@ -228,6 +228,7 @@ typedef struct private_subformat_data
  #define ALGORITHM_NAME			"32/" ARCH_BITS_STR
  #define ALGORITHM_NAME_S		"32/" ARCH_BITS_STR
  #define ALGORITHM_NAME_4		"32/" ARCH_BITS_STR
+ #define MIN_KEYS_PER_CRYPT		(MD5_X2+1)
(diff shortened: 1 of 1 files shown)


magnumripper:
@jfoug the irony now is since I pushed this I can't give it an approving review but you can 🤣  But on the other hand I can override that. So let's just wait for Circle CI now.

jfoug:
I was working trying to the the CI to work. I 'thought' I got it, since right after a change, IT WORKED, but I am pretty sure it was you fuxoring around.  But the other PR fired up a circleCI just before you closed it.  Right now, I am trying to simply remove my own fork as being a project. I would assume that the main tree I fork from has build-PR's.  So if I am not setup as a project, possibly it will work.

NOTE, it used to work.  But all that are there are v 1.0 circleCI builds, and a big warning about upgrading to 2.0, but I am pretty sure we have been 2.0 for a couple years (on the master tree).  Oh well, I still have a few other things to work on, so each time, I will mess around with circle stuff, until somehow it magically starts working, at which time I will not breath on it again (until it fails, lol).

Yes, I figured I could easily review this one, since it is the exact same line from the PR I made.

Issue #3562, opened 2018-12-24
Dynamic 'compiler' work (main work task)


jfoug:
All subtasks will be linked to this main Issue.   ANY new user issues started should also be linked into this topic.  Current known subtasks and bug reports:   #2800 - #3389  - (#3125 this one may not be part of this task conglomeration) - #1746 
As discussed with @magnumripper off-list, here is the plan

1. Build the scripts as best as possible (like is done today, just make some improvements where they can be made).

2. at time of 'linkage' to dyna, the compiler will run a couple samples.

3. if the samples 'pass' and crack, then the linkage will happen like it does today (to the native dynamic_fmt format).

4. if the samples fail, then I will instead link to a 'new' format.

This 'new' format will perform oSSL using a recursive decent parser against the already built recursive decent parser tree. This will be very similar to the code "in" the compiler which generates the test strings.   This code will not have the limitations of the 2 input and 2 output buffers which dynamic formats are limited to.  This is the problem when we try to actually 'emit' the code from the compiler.  Going from a recursive decent tree, into an iterative language, which has only 2+2 variables, and NO stack, will only work 'so far'.  Thus once there is a format built that properly uses the RDP tree of data, any valid expression (no matter how complex) will work fine.  It will be slow however, single iteration ssl, little to NO chance for optimizations.   I do think I will emit a warning message at the start of the run, listing to the user that this format is non optimal (and direct them to john ML or documentation), and listing that writing a proper dynamic format would likely run much faster, if there is much work at all to do with hashes of this expression

5. I will also work to have more expressions work within the compiler.  Many 'could' work fine, but no matter how much better I make it, it is VERY easy to make a valid expression which CAN NOT work within the limitations of dynamic. Any format which would require 3 temp 'registers' can not be done with the current dynamic.  So something like this

md5($s.md5($p.md5($p.$s)).md5($s.$p))

simply can not be done with the existing flat dynamic format.  It 'can' be done within dynamic, using all of the input buffers AND the output buffers.  But that usage does not work in flat mode (without changes), which would cause dynamic code base to grow substantially.   Also, even using all 4 variables (both inputs and both outputs), it is still trivial to generate expressions which require 5 temp registers to complete. An expression that complex (or more complex), CAN NOT be handled by the dynamic format AT ALL.  So no matter what, there will be expressions which dynamic of today CAN NOT handle.  So the new RDP dynamic format, will have to be done, else the 'compiler' will always have bug reports like we see that show up on a regular basis.

PR #3568, opened 2018-12-27
added dynamic-compiler RDP format when script builds incorrectly

jfoug:
Here is working version.  This simply detects a failed script at linkage time, emits a warning, sets a flag. Then within dynamic, if that flag lists script failure, then a RDP function within the dynamic compiler code is called, on each password to actually perform the work, since the dynamic code would not work.

Note, the dynamic compiler can still be optimized to handle more expressions.  But at some point of expression complexity, the dynamic format CAN NOT be used successfully. That is the point where the RDP parsing of the expression will take over and complete the work for the expression.

This work is for #3562


magnumripper:
Cool! I will try this out and give feedback, just don't hold your breath cause I've got all sorts of things going on around here!

jfoug:
Looks like circle-CI is now working again , yea!

kholia:
I will test this with some known dynamic expressions which were failing earlier.

This is awesome 👍

src/dynamic_compiler.c
```diff
@@ -2432,6 +2479,9 @@ char *dynamic_compile_split(char *ct) {
 int dynamic_assign_script_to_format(DC_HANDLE H, struct fmt_main *pFmt) {
 	int i;
 
+	/* assume comipler success. */
```
kholia: Small typo here.

jfoug: numerous small errors (mostly signed/unsigned char*), but this one also.  All fixed, and are in 2nd push.
The new format should handle any failing expressions.  It has no optimizations, and is slow due to marshal code at each step, but it is fast enough to get the job done, AND if the expression IS a common expression (i.e. used a lot), then we can see about working them into a real format, whether it be with a dynamic script, or a fat format.

jfoug:
NOTE, even though it is passing the checks, do not merge it yet. There are still issues. I am getting get_key(0) failures on my AVX2 builds.  There is some settings that I have to make, so that the SIMD buffers are not used. I have not figured all of that out yet.
Also, I would like to get the 'name' of the algorithm used done properly.  Currently it is this: 
```
Testing: dynamic=md5($s.sha1($s).md5($p)) [256/256 AVX2 8x1]... FAILED (get_key(0))
```

The 256/256 AVX2 8x1 is totally wrong.

But I think both of those issues are trivial, and should easily be fixable from within the dynamic_compiler linkage function (where we determine that the RDP code is needed).

jfoug:
14fd1fb should be much closer to working.  The code was failing when MD5_X2 was set (that is usual). This is why it was working like a champ on my VC test system. There, MD5_X2 is not set, so it was cracking all. But with X2 set, the only every other crypt key was being used, and the 2nd usage overwrote the first (not what we want at all).   
I also added the algorithm name of [Dynamic RDP] so that we know it is not using SIMD or any other enhancements, and is using the Recursive Decent Parser code to perform the hash logic.

jfoug:
Here are examples showing the speed difference. The reason the 2nd one works, is because the compiler optimizes out the sha1($s) into the 'salt' (i.e. it is computed only at load time).  But this shows the difference in speed.

NOTE, I still see a bug, where the 2nd one (which actually does a dynamic script), still lists that it will use the RDP.   This should be a trivial bug to find.


```
$ ../run/john -test=3 -form='dynamic=md5(sha1($s.$p).md5($p))'
cmp_one() failed. This format will FAIL and needs the Slower dyna-compiler format
This expression will use the RDP dynamic compiler format.
Benchmarking: dynamic=md5(sha1($s.$p).md5($p)) [Dynamic RDP]... DONE
Many salts:     439907 c/s real, 465553 c/s virtual
Only one salt:  463917 c/s real, 473644 c/s virtual

$ ../run/john -test=3 -form='dynamic=md5(sha1($s).md5($p))'
This expression will use the RDP dynamic compiler format.
Benchmarking: dynamic=md5(sha1($s).md5($p)) [256/256 AVX2 8x3]... DONE
Many salts:     7019K c/s real, 7726K c/s virtual
Only one salt:  2573K c/s real, 3089K c/s virtual
```

kholia:
> BTW I have some faint memory of .pot file issues with this format. We should look at them too if possible.

https://github.com/magnumripper/JohnTheRipper/issues/2437 perhaps.

magnumripper:
Here's some strangeness:
```
$ echo '900150983cd24fb0d6963f7d28e17f72' > test.in
$ ../run/john test.in -form:dynamic='md5($p)'
Loaded 1 password hash (dynamic=md5($p) [256/256 AVX2 8x3])
Proceeding with single, rules:Wordlist
Proceeding with wordlist:../run/password.lst, rules:Wordlist
abc              (?)
1g 0:00:00:00 DONE 2/3 (2018-12-28 16:11) 100.0g/s 336000p/s 336000c/s 336000C/s 123456..Geronimo
Use the "--show --format=dynamic=md5($p)" options to display all of the cracked passwords reliably
Session completed

$ cat ../run/john.pot 
$dynamic_6000$900150983cd24fb0d6963f7d28e17f72:abc
```
Dynamic 6000? Where did that come from?
```
$ ../run/john test.in -show
0 password hashes cracked, 2 left
```

jfoug:
NOTE, I am adding a new flag, ``` ,rcp ```  If that switch is there, then the format will use the RDP format, even IF the dynamic compiler 'could' build a proper script.  This flag will most likely be used mostly by me, as I spend some time getting the RDP code working faster, but you never know.  The ,debug flag was added also as a helping debug tool for me, but it sounds like it gets used by others.

I have thought of a few optimizations I could add to the RDP format for some formats. One big thing, is to parse and find common expressions, eliminating wasted hash calls. Those can not be added to the dynamic script, but within the RDP, since we are not limited to just having the 2 variables of dynamic.  So something like md5(md5($p).md5($s.$p).md5($p)) we could store off the md5($p) into a 'local' variable, pre-hash the salt, and then simply do:
append_p
md5
store results into t1
append $s
append $p
md5
store results in t2
append t1
append t2
append t3
md5 (and return results)

Now that I have RDP working well, I will work to clean things up, and we can get this new format checked in and then I will start to work on optimizations.  I am not 100% sure we need to wait on an optimized version prior to jumbo2.  If things get speed improved prior to jumbo2, good, if not, I vote it is not a show stopper.

jfoug:
|$ echo '900150983cd24fb0d6963f7d28e17f72' > test.in |
> |$ ../run/john test.in -form:dynamic='md5($p)' Loaded 1 password hash 
> (dynamic=md5($p) [256/256 AVX2 8x3]) Proceeding with single, 
> rules:Wordlist Press 'q' or Ctrl-C to abort, almost any other key for 
> status Proceeding with wordlist:../run/password.lst, rules:Wordlist 
> abc (?) 1g 0:00:00:00 DONE 2/3 (2018-12-28 16:11) 100.0g/s 336000p/s 
> 336000c/s 336000C/s 123456..Geronimo Use the "--show 
> --format=dynamic=md5($p)" options to display all of the cracked 
> passwords reliably Session completed $ cat ../run/john.pot 
> $dynamic_6000$900150983cd24fb0d6963f7d28e17f72:abc |
>
> Dynamic 6000? Where did that come from?
>
> |$ ../run/john test.in -show 0 password hashes cracked, 2 left |


I see the same results as you (the dynamic_6000 signature in john.pot).

however, with a removed .pot file and when run this way:

  ../run/john test.in -form='dynamic=md5($p),nolib'

this is the result:

  cat ../run/john.pot
@dynamic=md5($p),nolib@900150983cd24fb0d6963f7d28e17f72:abc


So this bug has something to do with a format 'coming' from the 
pre-built library, and is certainly a bug I will fix now.  also, a tiny 
issue (possible issue), is that the ,nolib probably should not be there 
either.  I will look at that also.  Things like ,nolib  ,debug  ,rdp 
should not be part of the signature. They are flags and not variables.  
the variables (such as ,saltlen=8 or ,c1=junk_text ) should be there, 
since they are 'variables'

jfoug:
> 
> 
> Using this branch
> 
> ```
> $ cat test.in
> 900150983cd24fb0d6963f7d28e17f72
> 
> $ ../run/john test.in -form:'dynamic=md5($p)'
> cmp_one() failed. This format will FAIL and needs the Slower dyna-compiler format
> This expression will use the RDP dynamic compiler format.
> Using default input encoding: UTF-8
> Loaded 1 password hash (dynamic=md5($p) [Dynamic RDP])
> Warning: no OpenMP support for this hash type, consider --fork=8
> Self test failed (cmp_all(1))
> ```

This is fixed in latest push.  Something changed in core john startup, and common_init() was not being called, but 'only' for the hashes in the dynamic-compiler-lib.  The md5($p) is one of them in the compiler lib.  Yes, I also found this when adding (and testing) the ,rdp switch.  That switch 'forces' usage of the RDP code.  So I tried it on some compiler-lib formats, only to find they were failing badly.  The problem was the binary() function was returning a null'd out buffer, since the atoi16 stuff was not initialized.  Thus the results were never matching a null binary buffer.

jfoug:
Fixed the john.pot output of things from the dynamic-compiler-lib  (such as md5($p) ) when run on 'bare' hashes. The prepare within dynamic-compiler was not handling raw hashes properly here, because it did not know the raw hash length.  I had to add that to the compiler lib structure. They all are 16 bytes now (all are MD5 hashes), but I did not want to force ONLY md5 hashes to be available in the lib.  So the length of the hash had to be added to the lib structure.  This was not an easy bug to find.  The worst thing, is that 'most' of the functionality did work even with the dynamic_6000 signature being used. However, all formats within the lib would have 'appeared' to have cracked the hash when using the -show function, even if that hash was not right.  Now, the proper @dynamic=md5($s)@01234...abc:password signature will be output into the john.pot file, so there is no mis-interpretation in the -show command.

src/dynamic_compiler.c
jfoug: Will be fixed next time I check things in.

src/dynamic_compiler.c
```diff
@@ -1425,6 +1426,52 @@ static char *rand_str(int len) {
...
 }
+
+// This is the 'crypt-all' for the RecursiveDecentParser code
+// This will be called from within dynamic_fmt crypt_all, IF
+// the compiler does not generate a valid script. john will
+// still process the data properly, but it does so much slower.
+void run_one_RDP_test(DC_ProcData *p) {
+	int n;
+	unsigned char *in;
+	strnzcpy(gen_pw, p->iPw, p->nPw+1);
```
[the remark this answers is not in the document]
jfoug: Fixed this issue in many functions (will check in shortly)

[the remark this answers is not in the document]
jfoug: Fixed in 39ad0fe

src/dynamic_compiler.c
jfoug: Fixed in 39ad0fe

magnumripper:
I think we can merge this now, right? Remaining issues are either of other scope or can be done later.

magnumripper:
This is a terrific improvement! :+1:

jfoug:
Yes, this was a very good place to merge.  It is working well.  The only change I have done past this, was to get 1 hash to work in the script format (a utf16), but it is still lacking, AND not really a generic fix. There are simply too many 1-offs failing the script, the code emit part gets confused VERY easily.

That is why the RDP format is so important. It is slower, yes, BUT it overcomes the deficiencies of the code emission part of dyna-compiler, and allows it to be fixed/upgraded/replaced over time, without having users failing to be able to complete their work.

Issue #3574, opened 2018-12-30
bug in fgetl/fgetll  (misc.c)


jfoug:
In the case we read a null byte in the stream, the code does not check after the null byte, looking for the \n (followed by a null), up to size, before starting to read byte/byte from the stream, looking for the \n.

Thus, if we have this file:

here is \0line 1\n
here is line 2\n
here is line 3\n
...

on call #1 to fgetl/fgetll, we get "here is "   On call 2, we get "here is line 3" missing the line "here is line 2" completely.

One other bug found.  fgetll would return the original pointer, in a long line, of that long line had a NULL byte in it, even though we have allocated a buffer (which the caller must free).  Memory leak, and we are returning a shorter section of the line.

solardiz:
I noticed this issue and intend to look into it in core's context later.

jfoug:
The change is actually pretty trivial, and safe.  Once you put logic in which was triggered by a short line that did not end in a \n, it will 'known' that the line had at least 1 embedded NULL.  Then simply walk the buffer of 'actual' read data.  If the \n is seen, then the call is done.  No need to read any more from the file.  Now, IF the line was full length, and did not contain a \n, then what we have is a line that is TOO long for our buffer. That's ok, and we simply read more data (throwing it way), until we he that \n EOL byte (or run out of file).

PR #3575, opened 2018-12-30
fgetl and fgetll lines with NULLs.  fgetll memory leak

jfoug:
#3574

Both functions would (usually) skip the next line, if there was a NULL byte in the current line.

fgetll returned the wrong pointer if there was a NULL in the line (on long lines). Looks like a cut/paste bug from the code handling nulls prior to the section to handle long lines.

### DIFF
--- a/src/misc.c
+++ b/src/misc.c
@@ -126,11 +126,20 @@ char *fgetl(char *s, int size, FILE *stream)
 			*pos = 0;
 			if (pos > res)
 			if (*--pos == '\r') *pos = 0;
-		} else
+			return res;
+		}
+		if ((pos-res) + 1 < size) {
+			/* There was a NULL byte in this line.
+			   Look for \n past the 'read' null byte */
+			while ((++pos - res) + 1 < size)
+				if (*pos == '\n')
+					return res; /* found it read no more */
+		}
 		if ((c = getc(stream)) == '\n') {
 			if (*pos == '\r') *pos = 0;
 		} else
 		while (c != EOF && c != '\n')
+			/* Line was longer than our buffer. discard extra */
 			c = getc(stream);
 	}
 
@@ -174,6 +183,11 @@ char *fgetll(char *s, size_t size, FILE *stream)
 		return s;
 	}
 	else if ((len + 1) < size) { /* We read a null byte */
+		/* first check past the null byte, looking for the \n */
+		while (++len < size)
+			if (s[len] == '\n')
+				return s; /* found it. Read no more. */
+		/* did not find the \n, so read and discard rest of line */
 		do {
 			c = getc(stream);
 		} while (c != EOF && c != '\n');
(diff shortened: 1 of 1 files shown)


jfoug:
The whole issue came from 3 things. 

1 looking at fgetll and 'seeing' that the handling of lines with NULLs in them was wrong, and dropping a line.
2. I could see the same problem in fgetl (hidden a bit).
3. I saw near the bottom of fgetll an obvious bug (returned s, even though cp was the newly allocated buffer, which the caller must clean up).  This was almost certainly a cut/paste bug when the null logic was added to the 'normal' sized lines within fgetll

Now, we can handle this in several ways.
1. we have to fix the memory leak bug.
2. the code as supplied in this commit fixes the drop of the next line of a binary file.  We can go with this.
3. we can ignore the problem, users should not be using these functions for binary data.

It appears that mgetl worked fine for the binary data.  I tested this with wordlist.c  (simply using ./john inp --stdout)   I then changed the read 1 line statement in wordlist, to get the function I wanted to test.

the mgetl worked out of the box.
both fgetl and fgetll would skip the line following a line with a NULL in it.
my VC started out with fgetl and showed the bug.  my cygwin build started out with mgetl, and hid the bug (and made me scratch my head a bit about why it was working when it should not have been).

But since this is a core function, we do have to get buy in from Solar on this one.

jfoug:
in wordlist mode, there was no warning about embedded NULL bytes, at least in my testing.

magnumripper:
I think we should accept all of this, I just wanted Solar to have a look see first. The fgetll part is definitely going in.

> in wordlist mode, there was no warning about embedded NULL bytes, at least in my testing.

I guess it only happens in some cases, ie. when using memory buffer and rules, or something like that. I'm sure it does warn in *some* situations.

magnumripper:
Solar took a walkover. I assume you've tested this fairly well, so I merged it

jfoug:
> Solar took a walkover. I assume you've tested this fairly well, so I merged it

I tested this on 4 environments (linux 32/64 bit, cygwin64 bit and VC. I tested with normal unix txt files (\n), and dos files \r\n  and f'd up files with some lines \n and some \r\n  (3 files only).  I wrote the lines to be easily tested, to make it very visible if lines were not read.  Prior to changes, both functions would drop a line following a line with an embedded null.  After the change, both functions would read the line with the null, and printf 'up to' the null (expected behavior), and then the next line would be properly read.
I tested from within john, but modifying wordlist.c, so that the function I wanted tested would always be used.   Then I simply used ``` ./john input-test.txt --stdout ```  Observing the output lines.  Note this was not comprehensive, but I was assuming that most files were properly handled by the functions, so other than running a single file that contained no nulls, through both functions, all testing was on null embedded files.  However, stating all of that, I am very happy with the code changes turned in.

solardiz:
I looked into including a fix for this issue into core's `fgetl`, but unfortunately the fix committed here looks subtly wrong - I think it'll look for `'\n'` in uninitialized or stale memory when `fgets` encounters EOF or an error without first seeing a linefeed - e.g., if the last line in a file doesn't end with `'\n'`. I'm not aware of a way to fix the original issue properly while continuing to use `fgets`. So I'll probably leave it unfixed in core for now.

PR #3577, opened 2019-01-03
configure: detect openssl installed by MacPorts on macOS

LRGH:
Without this patch, configure needs to be called with additional arguments.
This patch is made to be similar to the surrounding lines.


LRGH:
Yes, you are right.

PR #3578, opened 2019-01-03
HDAA: remove check that cnonce is lowercase hex

LRGH:
RFC 7616 section 3.4 does not put such constraints on cnonce: it specifies that cnonce is ASCII-only (we may want to check this property, but I don't see why john should reject invalid cnonce values).
The examples of RFC 7616 section 3.9.1 contain non-hex values for nonce and cnonce.
Note that RFC 7616 section 3.3 recommends that the nonce is Base64 or hexadecimal data, but it is not mandatory.

### DIFF
--- a/src/HDAA_fmt_plug.c
+++ b/src/HDAA_fmt_plug.c
@@ -221,8 +221,6 @@ static int valid(char *ciphertext, struct fmt_main *self)
 		goto end_hdaa_legacy;
 	if ((p = strtokm(NULL, "$")) == NULL) /* clientnonce */
 		goto err;
-	if (!ishexlc(p) )
-		goto err;
 	if ((p = strtokm(NULL, "$")) == NULL) /* qop */
 		goto err;
 	if ((p = strtokm(NULL, "$")) != NULL)


solardiz:
We shouldn't simply be removing such restrictions. When we do, we should also be adding case unification to `split` and setting `FMT_SPLIT_UNIFIES_CASE`, for subtle reasons that were discussed before many times (perhaps we need to add that to development documentation).

LRGH:
I don't understand your comment.

Without my patch, john is not able to reverse legitimate hashes such as the ones from the RFC.

With my patch you can find that `Circle of Life` is the password for
```MufasaRFC7616:$response$8ca523f5e9506fed4657c9700eebdbec$Mufasa$http-auth@example.org$GET$/dir/index.html$7ypf/xlj9XXwfDPEoM4URrv/xwf94BcCAzFZH4GiTo0v$00000001$f2/wE4q74E6zIJEtWaHKaf5wv/H5QzzpXusqGemxURZJ$auth```

Please give me an example of a legitimate input that gets broken by my patch.

solardiz:
I'm not saying the change is undesirable (I think it is desirable to support all valid inputs) nor that it breaks any legitimate input (I think it does not). I am saying that the patch is incomplete (allows JtR to enter an incorrect state wrt (non-)detection of effectively equivalent hashes), and should be completed in the way I described.

See #2842 and/or search issues (including closed ones) for `FMT_SPLIT_UNIFIES_CASE`.

Thank you!

LRGH:
The issue that you indicate is about the way the hash is encoded. I did not remove this check (lines 208 and 209).
I removed an invalid check of the cnonce value.

magnumripper:
@LRGH thank you for this patch, we'll likely merge it sooner or later but Solar's comment is relevant so we need to dig into that first.

@solardiz I'm not sure we would want to unify case though. While input might be hex it could also be Base64 or something else. Perhaps we could unify case only for hex-only ciphertexts... but would that make sense? My first reflection is to just leave case alone.

magnumripper:
> I'd expect this PR as it currently is to fail --test-full. We can use this as an opportunity to test that.

BTW it didn't.

solardiz:
@magnumripper Of course, I meant unifying the case only for the portion in hex where we remove the case check. Not of the entire string, nor with other encodings.

solardiz:
OK, maybe I misunderstood. Is the cnonce string essentially arbitrary and case-sensitive? And we just happened to wrongly require it to be lowercase hex? If so, the patch is complete as-is.

LRGH:
The cnonce is case-sensitive.
According to the RFC, it is not arbitrary, because it should be ASCII-only (meaning octets between 0 and 127). But in reality, I think that if we want to make a verification, we should probably forbid non-printable ASCII characters: I am pretty confident that no implementation of HTTP Digest Access Authentication generates non-printable ASCII characters in their cnonce.

magnumripper:
Looks like @jfoug added the too restrictive check in b3669c20ec. While we could add some other restrictions, my opinion is we should simply merge this as-is.

magnumripper:
Oh, and the actual hash still needs to be lower-case, so no need for `SPLIT_UNIFIES_CASE`.

magnumripper:
Thanks!

PR #3583, opened 2019-01-08
Common.[ch] and misc.[ch] changes

jfoug:
Here are changes to common.[ch] and misc.[ch] which have been triggered by findings from within the unit-tests project.

Also, one format was changed (it's valid).  It was the only format not handling the change to ishex("") now returning false (prior logic returned true).

There was also a change made in jumbo.[ch] but only for a VC build, and only for VC 2015 or newer.  2015 fixed the snprintf function (leaving the older 'busted' logic in the _snprintf function).  Now, (v 2015), the snprintf works properly, and having a wrapper varargs function to handle it makes things more complex.

There is a comment in the first commit, that lists all things changed.

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -125,7 +125,10 @@ int ishexn(const char *q, int n);
 int ishexucn(const char *q, int n);
 int ishexlcn(const char *q, int n);
 /* length of hex. if extra_chars not null, it will be 1 if there are more
- * non-hex characters after the length of valid hex chars returned. */
+ * non-hex characters after the length of valid hex chars returned.
+ * NOTE, the return will always be an even number (rounded down). so if we
+ * want the length of "ABCDE", it will be 4 not 5.
+ */
 size_t hexlen(const char *q, int *extra_chars);
 size_t hexlenl(const char *q, int *extra_chars); /* lower cased only */
 size_t hexlenu(const char *q, int *extra_chars); /* upper cased only */
--- a/src/misc.c
+++ b/src/misc.c
@@ -257,7 +257,7 @@ void *strncpy_pad(void *dst, const void *src, size_t size, uint8_t pad)
 	uint8_t *d = dst;
 	const uint8_t *s = src;
 
-	if ( ((long long)size) < 1)
+	if ((int64_t)size < 1)
 		return dst;
 
 	while (*s && size) {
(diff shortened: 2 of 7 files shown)


src/common.h
```diff
@@ -125,7 +125,10 @@ int ishexn(const char *q, int n);
 int ishexucn(const char *q, int n);
 int ishexlcn(const char *q, int n);
 /* length of hex. if extra_chars not null, it will be 1 if there are more
- * non-hex characters after the length of valid hex chars returned. */
+ * non-hex characters after the length of valid hex chars returned.
+ * NOTE, the return will always be an even number (rounded down). so if we
+ * want the length of "ABCDE", it will be 4 not 5.
+ */
```
magnumripper: Does "ABCDE" not trigger "extra_chars"? Given it returns 4, it should.

src/misc.c
```diff
@@ -250,15 +251,19 @@ char *fgetll(char *s, size_t size, FILE *stream)
...
 }
-#endif
 
 void *strncpy_pad(void *dst, const void *src, size_t size, uint8_t pad)
 {
 	uint8_t *d = dst;
 	const uint8_t *s = src;
 
-	while (*s && size--)
+	if ( ((long long)size) < 1)
```
magnumripper: We should refrain from using `long` and `long long` in favor of `int64_t`

magnumripper: Fixed now

src/common.h
jfoug: I am not sure.  Let me check.....

yes, hexlen("ABCDE", &ex)    returns 4, and ex is set to 1.

magnumripper: Good, then we're fine!

magnumripper:
As long as these functions include tests for caller bugs at all, why not bail with error instead of silently chugging along? Anyway, feel free to commit it.

jfoug:
On 1/9/2019 11:46 AM, magnum wrote:
>
> As long as these functions include tests for caller bugs at all, why 
> not bail with error instead of silently chugging along? Anyway, feel 
> free to commit it.
>
>
I asked about doing that, but crickets were louder, LOL.  Yes, since the 
infrastructure is there, we can easily call error() and bail (which we 
SHOULD do), and will have CI failures, along with a semi unusable 
file/line (the real file/line is in the calling function, lol).  But 
dying (or crashing), really IS the proper thing to do.  Dying gracefully 
with error text (and semi-usable file/line), sure beats bullet proofing, 
running and having something much deeping in the code die, causing the 
dog-chasing-tail situation.

PR #3587, opened 2019-01-11
fixed bullet-proof code to not core on 32 bit builds

jfoug:
This is required for 32 bit builds (where size_t is not 64 bits).


jfoug:
Once this change is merged, then the branch 'unit-tests-fix' should pushed this change. Once that is done, the 'unit-tests-fix' branch will build properly within the CI's for 32 bit builds.

PR #3588, opened 2019-01-11
sha256crypt-ztex new version

Apingis:
New version, some details in <a href='https://github.com/magnumripper/JohnTheRipper/compare/bleeding-jumbo...Apingis:ztex?expand=1#diff-0da3fdf2b3a43018e9102ff82d7407b3'>README</a>


src/ztex_sha256crypt.c
```diff
@@ -76,9 +76,9 @@ static struct device_bitstream bitstream = {
 	// (keys * mask_num_cand)/crypt_all_interval per jtr_device.
 	1,			// set by init() base on john.conf setting
 	4096,		// keys/fpga for self-test
-	600 * 1024,	// Would be ~20 MB of USB traffic on 32-byte keys
+	565248,		// Would be ~20 MB of USB traffic on 32-byte keys
 	512,		// Max. number of entries in onboard comparator.
-	150,		// Min. number of keys for effective device utilization
+	23 * 12,	// Min. keys for effective device utilization
 	1, { 135 },	// Programmable clocks
```
solardiz: For consistency, we should set the default frequency to 175 in here as well - not only in default `john.conf`.

PR #3600, opened 2019-01-16
add -fstack-clash-protection gcc flag

frank-dittrich:
See https://github.com/magnumripper/JohnTheRipper/issues/3599
I tested this change on well (CPU-only build).

I ran `time ./jtrts.pl` after compiling john with and without this flag.
with:
```
real	29m14.597s
user	58m9,515s
sys	1m43,198s
```
without:
```
real	29m10.945s
user	58m1,502s
sys	1m42,724s
```
(Both runs completed without errors.)

That looks like a 0.2 percent performance penalty (0.4 percent for sys).
Not sure yet whether this is repeatable.

Then I tested memory usage on my laptop.
I compared memory usage restoring a --wordlist --rules session with almost 14 million NT hashes.

Using top, I saw no differences:
```
  PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND    
14471 fd        39  19   10,2g  10,0g   4,8g R 100,0  32,1   1:06.50 john       
14527 fd        39  19   10,2g  10,0g   4,8g R 100,0  32,1   0:56.93 john       
14529 fd        39  19   10,2g  10,0g   4,8g R 100,0  32,1   0:56.86 john       
14528 fd        39  19   10,2g  10,0g   4,8g R  99,7  32,1   0:56.92 john       
```

### DIFF
--- a/run/benchmark-unify
+++ b/run/benchmark-unify
@@ -101,7 +101,9 @@ while (<>) {
 # (ignoring case, so use sort -f or sort --ignore-case, please)
 __DATA__
 bf-opencl, OpenBSD Blowfish (x32)	bcrypt-opencl ("$2a$05", 32 iterations)
+blackberry-es10	Blackberry-ES10 (101x)
 BSDI DES (x725)	bsdicrypt, BSDI crypt(3) ("_J9..", 725 iterations)
+chap, iSCSI CHAP authentication	chap, iSCSI CHAP authentication / EAP-MD5
 Cisco PIX MD5	pix-md5, Cisco PIX
 CRC-32	CRC32
 crypt-MD5	md5crypt, crypt(3) $1$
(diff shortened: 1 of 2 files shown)


frank-dittrich:
>  I feel old.

Why that?

jfoug:
> 
> 
> Cool. I feel old.

I feel cold (but its just winter).  I know I'm old, lol.

jfoug:
Note, we do need to test this on non-linux, just in case.  Again, since this IS dependant upon a problem in a linux systemd journal, it really should not impact any system outside of linux anyway.

frank-dittrich:
@jfoug 

> Note, we do need to test this on non-linux, just in case. Again, since this IS dependant upon a problem in a linux systemd journal, it really should not impact any system outside of linux anyway.

I think you misunderstood 

https://www.qualys.com/2019/01/09/system-down/system-down.txt

This was a systemd bug which couldn't be exploited on some distros which compile all userland (like systemd) with `-fstack-clash-protection`.

This change is totally unrelated to that systems bug. just
I only referenced it so that you know what this gcc flag is about.
From `man gcc`:
```
      -fstack-clash-protection
           Generate code to prevent stack clash style attacks.  When this
           option is enabled, the compiler will only allocate one page of
           stack space at a time and each page is accessed immediately after
           allocation.  Thus, it prevents allocations from jumping over any
           stack guard page provided by the operating system.

           Most targets do not fully support stack clash protection.  However,
           on those targets -fstack-clash-protection will protect dynamic
           stack allocations.  -fstack-clash-protection may also provide
           limited protection for static stack allocations if the target
           supports -fstack-check=specific.
```
 
And I thought bleeding-jumbo might also have some bugs which allow stack smashing, e.g., when feeding arbitrary input files to certein formats.
And compiling john with `-fstack-clash-protection` just makes such bug not exploitable (on systems which provide the necessary infrastructure for the `-fstack-clash-protection` flag.

 
But I agree we should test this on non-linux systems, including the impact on run time and memory usage.

Issue #3598, opened 2019-01-16
Unhide commented out options variables in john.conf


jfoug:
We should try to NOT have a bunch of commented option variables in john.conf.  We have the john option of  ``` --list-parameters:Options ```  which dumps a list of what the current settings are.  If lines are commented out, then they do NOT show up in this dump, EVEN THOUGH there is default value for this item(s).

So:

```
# Un-commenting this stops Single mode from re-testing guessed plaintexts
# with all other salts.
#SingleRetestGuessed = N
```
But should be
```
# Change to 'N' to stop Single mode from re-testing guessed plaintexts
# with all other salts.
SingleRetestGuessed = Y
```
 
These commented out lines in the Options section should all be looked at, and (if possible), uncommented with proper default setting set, and the comment changed. This will allow the option to show up if the --list=parameters:Options

Yes, a bit off-topic here, but I wanted this to be talked about. If it is deemed 'a good thing', then we can easily create a issue to hang it on.

NOTE, this needs to be done carefully.  We need to look at how each variable is used, and properly set the variable BASED upon how it's defaults appear to be.  It can be opposite of how the commented out variable is (as in the SingleRetestGuessed instance).  But there are others where simply removing the comment and leaving it set as is, is correct.

These are the current commented variables:

```
#WordlistRules = Wordlist
#DefaultIncrementalUTF8 = UTF8
#LogDateFormat = %Y-%m-%dT%H:%M:%S%z
#LogDateFormatUTC = Y
#LogDateStderrFormat = %b %d %H:%M:%S
#LogFilePermissions = 0600
#PotFilePermissions = 0600
#IgnoreChmodErrors = N
#SingleRetestGuessed = N
#SessionFileProtect = Named
#LogFileProtect = Named
#DefaultInternalCodepage = ISO-8859-1
```

frank-dittrich:
The current `--list-parameters:section` has been implemented with scripts in mind.
For end users, an enhanced version might be needed.

That enhanced version should ideally
- print any preceding comments fro mthe config file
- print the parameter names in CamelCase (like in the config file)
- indicate whether the current setting is the default setting or a setting which has been adjusted by the user

jfoug:
> indicate whether the current setting is the default setting or a setting which has been adjusted by the user
Very important, AND useful, for sure!   But this issue is simply to adjust the john.conf file itself, removing the comments (effectively) for items in the [Options] section.

[...]

PR #3606, opened 2019-01-18
Uncommented params from john.conf Options section.

jfoug:
Tied to #3598

### DIFF
--- a/run/john.conf
+++ b/run/john.conf
@@ -42,9 +42,10 @@ SingleRules = Single
 # Default batch mode Wordlist rules
 BatchModeWordlistRules = Wordlist
 
-# Default wordlist mode rules when not in batch mode (if any)
-# If this is set and you want to run once without rules, use --rules:none
-#WordlistRules = Wordlist
+# Default wordlist mode rules when not in batch mode (if any).  If this is
+# changed from 'none' to have default rules applied, and you DO want to run
+# once without rules, use --rules:none on the command line.
+WordlistRules = none
 
 # Default loopback mode rules (if any)
 # If this is set and you want to run once without rules, use --rules:none
@@ -60,8 +61,9 @@ RelaxKPCWarningCheck = N
 
 # Default/batch mode Incremental mode
 # Warning: changing these might currently break resume on existing sessions
+# one option frequently changed (with above caveat) is setting DefaultIncrementalUTF8 = UTF8
 DefaultIncremental = ASCII
-#DefaultIncrementalUTF8 = UTF8
+DefaultIncrementalUTF8 = ASCII
 DefaultIncrementalLM = LM_ASCII
 
 # Time formatting string used in status ETA.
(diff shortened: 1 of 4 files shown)


run/john.conf
```diff
@@ -42,9 +42,10 @@ SingleRules = Single
 # Default batch mode Wordlist rules
 BatchModeWordlistRules = Wordlist
 
-# Default wordlist mode rules when not in batch mode (if any)
-# If this is set and you want to run once without rules, use --rules:none
-#WordlistRules = Wordlist
+# Default wordlist mode rules when not in batch mode (if any).  If this is
+# changed from 'none' to have default rules applied, and you DO want to run
+# once without rules, use --rules:none on the command line.
+WordlistRules = none
```
magnumripper: I think this might be a regression. We should compare using `-w` vs. `-w -ru:none` for some of the fastest formats.

jfoug: Agreed.   I know the 'results' should be the same.  BUT there could be certain additional issues, going from 'no' rules, to a 'dummy' rule. 

I am glad question this showed up in the PR.  This change (and  the change to ``` DefaultIncrementalUTF8 = ASCII ``` ) are the 2 which troubled me the most.  Keeping the logic 'dictates' them, BUT they certainly give me the chills, thinking there is something not quite right.  The date formats, setting same logic for NULL as for "" is no big issue.  It is only at the 'exceptional event' of a log, and an extra if is no big deal.

magnumripper: It should be easy to change the code to not enable rules at all for a null string.

jfoug: That is probably a better solution, than using rules:none (much like the handling of date logic).  I will work on making it so, then it should be 'identical' both in logic and implementation, where with the patch as it stands it is only equal in logic.

Is the change to ``` DefaultIncrementalUTF8 = ASCII ``` OK?  That was about the only other thing that made the hair on the back of my neck stand up a bit.

magnumripper: > Is the change to `DefaultIncrementalUTF8 = ASCII` OK? That was about the only other thing that made the hair on the back of my neck stand up a bit.

Yes, given that's the default it's actually better - because it's harder to miss!

jfoug:
@magnumripper I believe that once the CI's complete, this is ready. 

It was tested using real wordlist, rules, no rules, etc.  The internal ST would not catch this.  The jtrts did catch an initial version which had problems.  But I rolled back changes, and simply detect when reading the config file that we have a "" string, and then set the variable to be NULL (i.e. original behavior).  Now, the jtrts.pl is happy, and all tests I could think of using rules were happy.

This should remove any possibility to a regression issue for this config option.

src/unicode.c
```diff
@@ -971,7 +971,7 @@ int cp_name2id(char *encoding)
 	char enc[16] = "";
 	char *d = enc;
 
-	if (!encoding)
+	if (!encoding || !encoding[0])
 		return CP_UNDEF;
 	if (strlen(encoding) > sizeof(enc))
 		goto err;
```
magnumripper: What is the rationale for this change?

magnumripper: Never mind - I got the idea the change was the other way round. This makes sense!

PR #3608, opened 2019-01-18
using void* in arithmetic is not valid, even though gcc extension allows

jfoug:
Doing arithmetic on the contents of a void pointer is not defined (or at least NOT portable).  I believe this is an implementation defined operation (vs undefined), but in the same case, this is not portable.

This all falls down to this (and what the sizeof actually is

```c
#include <stdio.h>
int get_size(void *p) {
     return sizeof(p[0]);
}
int main() {
     printf ("size of void object is:  %d\n", get_size(0));
}
```

This compiles on gcc and VC.  Gcc returns sizeof void object as 1.  VC returns its size as 0 (which 0 REALLY is correct, as per the C standard).

I did find this:   https://www.geeksforgeeks.org/void-pointer-c-cpp/   ``` 2) The C standard doesn’t allow pointer arithmetic with void pointers. However, in GNU C it is allowed by considering the size of void is 1 ```   and this:    https://stackoverflow.com/questions/20967868/should-the-compiler-warn-on-pointer-arithmetic-with-a-void-pointer  ``` Pointer arithmetic with void * is a GCC extension and not standard C. ```

NOTE, adding ``` -Wpointer-arith``` to the build forces the compiler to SHOW this problem __(we might want to look at adding that switch, if it is available)__.  Here is a full build, where I added this warning flag.  This warning flag only showed up these 2 items.  It can also flag code like the sizeof code in my first sample code post, as being ```warning: invalid application of ‘sizeof’ to a void type [-Wpointer-arith]```  The john project has only the 2 warnings I have fixed in this PR.

```
$ make -sj3
sapB_fmt_plug.c: In function ‘cmp_one’:
sapB_fmt_plug.c:365:27: warning: pointer of type ‘void *’ used in arithmetic [-Wpointer-arith]
            !memcmp(binary + BINARY_SIZE / 2, zeros, BINARY_SIZE / 2)));
                           ^
sapG_fmt_plug.c: In function ‘cmp_one’:
sapG_fmt_plug.c:344:27: warning: pointer of type ‘void *’ used in arithmetic [-Wpointer-arith]
            !memcmp(binary + BINARY_SIZE / 2, zeros, BINARY_SIZE / 2)));
                           ^
ar: creating aes.a
ar: creating secp256k1.a
ar: creating ed25519-donna.a
```

### DIFF
--- a/src/sapB_fmt_plug.c
+++ b/src/sapB_fmt_plug.c
@@ -362,7 +362,7 @@ static int cmp_one(void *binary, int index)
 	if (half_hashes)
 		return (!memcmp(binary, crypt_key[index], BINARY_SIZE) ||
 		        (!memcmp(binary, crypt_key[index], BINARY_SIZE / 2) &&
-		         !memcmp(binary + BINARY_SIZE / 2, zeros, BINARY_SIZE / 2)));
+		         !memcmp(((unsigned char*)binary) + BINARY_SIZE / 2, zeros, BINARY_SIZE / 2)));
 	else
 		return (!memcmp(binary, crypt_key[index], BINARY_SIZE));
 #endif
--- a/src/sapG_fmt_plug.c
+++ b/src/sapG_fmt_plug.c
@@ -341,7 +341,7 @@ static int cmp_one(void *binary, int index)
 	if (half_hashes)
 		return (!memcmp(binary, crypt_key[index], BINARY_SIZE) ||
 		        (!memcmp(binary, crypt_key[index], BINARY_SIZE / 2) &&
-		         !memcmp(binary + BINARY_SIZE / 2, zeros, BINARY_SIZE / 2)));
+		         !memcmp(((unsigned char*)binary) + BINARY_SIZE / 2, zeros, BINARY_SIZE / 2)));
 	else
 		return (!memcmp(binary, crypt_key[index], BINARY_SIZE));
 #endif


jfoug:
Should I add ``` -Wpointer-arith ``` to the configure, to have that param show up if the warning is valid, to keep this from happening again.

jfoug:
> I wonder why it's not enabled by -Wall or the like.

Not sure, but -Wall is anything but 'all'  ;)   From this page (not dox, but it does put things into 1 place)

https://stackoverflow.com/questions/2408038/what-does-wall-in-g-wall-test-cpp-o-test-do

```


See man gcc.

-Wall turns on these warnings:

-Waddress -Warray-bounds (only with -O2) -Wc++0x-compat -Wchar-subscripts
-Wenum-compare (in C/Objc; this is on by default in C++) -Wimplicit-int (C and
 Objective-C only) -Wimplicit-function-declaration (C and Objective-C only) 
-Wcomment -Wformat -Wmain (only for C/ObjC and unless -ffreestanding) 
-Wmissing-braces -Wnonnull -Wparentheses -Wpointer-sign -Wreorder -Wreturn-type 
-Wsequence-point -Wsign-compare (only in C++) -Wstrict-aliasing 
-Wstrict-overflow=1 -Wswitch -Wtrigraphs -Wuninitialized -Wunknown-pragmas 
-Wunused-function -Wunused-label -Wunused-value -Wunused-variable 
-Wvolatile-register-var

```
-Wextra contains:

-Wclobbered -Wempty-body -Wignored-qualifiers -Wmissing-field-initializers
-Wmissing-parameter-type (C only) -Wold-style-declaration (C only) -Woverride-init
-Wsign-compare -Wtype-limits -Wuninitialized -Wunused-parameter (only with -Wunused
 or -Wall) -Wunused-but-set-parameter (only with -Wunused or -Wall)
```


There are many more warnings which you have to turn on explicitly.
```
-Wall -Wextra -Waggregate-return -Wcast-align -Wcast-qual -Wdisabled-optimization -Wdiv-by-zero -Wendif-labels -Wformat-extra-args -Wformat-nonliteral -Wformat-security -Wformat-y2k -Wimplicit -Wimport -Winit-self -Winline -Winvalid-pch -Wjump-misses-init -Wlogical-op -Werror=missing-braces -Wmissing-declarations -Wno-missing-format-attribute -Wmissing-include-dirs -Wmultichar -Wpacked -Wpointer-arith -Wreturn-type -Wsequence-point -Wsign-compare -Wstrict-aliasing -Wstrict-aliasing=2 -Wswitch -Wswitch-default -Werror=undef -Wno-unused -Wvariadic-macros -Wwrite-strings -Wc++-compat -Werror=declaration-after-statement -Werror=implicit-function-declaration -Wmissing-prototypes -Werror=nested-externs -Werror=old-style-definition -Werror=strict-prototypes 
```

So even if we do turn on ```-Wpointer-arith``` (which we should)  there are still dozens of 'potential' warnings we are not accounting for.

NOW, many of them are likely not part of -Wall, because they false alarm, or there is too much code written that triggers them, or some other reason.   I know the addition of -Wpointer-arith to configure.ac is beyond the scope of this PR.  Getting the void* += x out of our code IS the point.  IF we want 'additional' warnings, then it merits it's own PR (for sure).  There seem to be lots of them.  Now, a lot of the warnings  pretty much p1ss me off, because they point out shortcut code I have been doing for years, which 'is' rock solid.  However, just as often, they do point out areas which ARE bug ridden, and should be reworked, to clarify things properly to the compiler.

PR #3612, opened 2019-01-20
Tune raw-SHA256/512 on super

claudioandre-br:


claudioandre-br:
PS:
* Note that the AMD Vega error was caused exactly by the use of a previous "tune" value.

Issue #3611, opened 2019-01-20
audit gcc warnings (adding new ones which 'help' our project)


jfoug:
From external discussion in #3608   If this is deemed not a big deal, then we can simply close this issue.
To start with, I will make a build turning on ALL the listed warnings, and just give it a shot, lol).

> I wonder why it's not enabled by -Wall or the like.

Not sure, but -Wall is anything but 'all'  ;)   From this page (not dox, but it does put things into 1 place)

https://stackoverflow.com/questions/2408038/what-does-wall-in-g-wall-test-cpp-o-test-do

```
See man gcc.

-Wall turns on these warnings:

-Waddress -Warray-bounds (only with -O2) -Wc++0x-compat -Wchar-subscripts
-Wenum-compare (in C/Objc; this is on by default in C++) -Wimplicit-int (C and
 Objective-C only) -Wimplicit-function-declaration (C and Objective-C only) 
-Wcomment -Wformat -Wmain (only for C/ObjC and unless -ffreestanding) 
-Wmissing-braces -Wnonnull -Wparentheses -Wpointer-sign -Wreorder -Wreturn-type 
-Wsequence-point -Wsign-compare (only in C++) -Wstrict-aliasing 
-Wstrict-overflow=1 -Wswitch -Wtrigraphs -Wuninitialized -Wunknown-pragmas 
-Wunused-function -Wunused-label -Wunused-value -Wunused-variable 
-Wvolatile-register-var
```

-Wextra contains:
```
-Wclobbered -Wempty-body -Wignored-qualifiers -Wmissing-field-initializers
-Wmissing-parameter-type (C only) -Wold-style-declaration (C only) -Woverride-init
-Wsign-compare -Wtype-limits -Wuninitialized -Wunused-parameter (only with -Wunused
 or -Wall) -Wunused-but-set-parameter (only with -Wunused or -Wall)
```

There are many more warnings which you have to turn on explicitly.
```
-Wall -Wextra -Waggregate-return -Wcast-align -Wcast-qual -Wdisabled-optimization -Wdiv-by-zero 
-Wendif-labels -Wformat-extra-args -Wformat-nonliteral -Wformat-security -Wformat-y2k -Wimplicit 
-Wimport -Winit-self -Winline -Winvalid-pch -Wjump-misses-init -Wlogical-op -Werror=missing-braces 
-Wmissing-declarations -Wno-missing-format-attribute -Wmissing-include-dirs -Wmultichar -Wpacked 
[...]

PR #3616, opened 2019-01-22
Gcc extra warnings (do not merge)

jfoug:
#3611

There are still a few errors left.  But those have much larger footprint on source code changes.  I am working to put those commits together.

THIS PR is not merge ready yet. There will be more commits.


jfoug:
Ok, if this fails again, I will put it off for the night, get some sleep, then work to log into the CI's and modify the configure.ac until things work again (or figure out which source files require improvements).

jfoug:
FINALLY.  Now I have big log files to look over, looking through all of these warnings for data which 'may' need fixed (i.e. warnings that REALLY do matter).  A lot of them are likely fluff, or invalid compiler warnings.

I see a LOT of them being things like strict alias rules.  But this seems to be a VERY old compiler, and this warning is not apparent in more recent compilers.  This could be combinations of compiler/OS, and possibly there are issues being picked up by the CI's.    But since I am building 32/64 bit cygwin and 64 bit Fedora29 and not seeing those, them im not over concerned by them.  I DID find several 'real' alias rules being broken, and fixed those (in 646f7b1 ).  I will start digging through the CI data.

Now, as it stands, I have added a --enable-exwarn  This turns on all of the extra warnings.  It also changes a 'few' of the --error=warning  into -Wwarning so that the compile will continue.   I had to do this to get it to pass Travis.  There were simply way too many errors (which were not showing up in other areas), on travis. So I (temp) changed the travis build to do the --enable-exwarn   and removed the --enable-werror.

jfoug:
Now that i have this where it will pass the CI's, I will get the additional things I have accounted for  ``` -Wmissing-declarations ```  And then continue looking into the warnings to see which should be addressed, and which should be ignored as false warnings providing no better quality to the code base

claudioandre-br:
Deprecated. Not mergeable.

PR #3619, opened 2019-01-24
Make sha512_reverse() available for non-SIMD

claudioandre-br:
This is what makes sense to me: move it to sha2.c

I can also wrap it using an:
```
#if HAVE_OPENCL || HAVE_SIMD
```

See: https://github.com/magnumripper/JohnTheRipper/pull/3617#issuecomment-457009975
See: https://github.com/magnumripper/JohnTheRipper/pull/3617#issuecomment-457181387


src/sha2.c
```diff
@@ -639,3 +648,63 @@ void jtr_sha512_final(void *_output, jtr_sha512_ctx *ctx)
...
+	maj = (b & c) ^ (b & d) ^ (c & d);
+
+	hash[0] = e - (a - (s0 + maj));
+}
+
+void sha512_unreverse(uint64_t *hash)
+{
+	fprintf(stderr, "sha512_unreverse() not implemented\n");
+	error();
+}
```
magnumripper: I suppose we can drop the unreverse function completely since we aren't exporting (nor using) it anyway.

src/sha2.h
```diff
@@ -68,4 +69,12 @@
 
 #endif
 
+#if defined(SIMD_COEF_64) || defined(HAVE_OPENCL)
+/*
+ * Reverse SHA512 rounds. The function can be used on OpenCL.
+ * - It must be available for a SIMD or non-SIMD build.
+ */
+void sha512_reverse(uint64_t *hash);
+#endif
```
solardiz: I think this `#if` and comment should be in `sha2.c` as well. There's no reason to compile the code in when we're not providing the prototype, and this comment is actually informative (unlike "Here is code for reverse SHA512 rounds", which doesn't say anything beyond what's implied by the function name).

src/sha2.c
solardiz: I agree, no reason to have a function that merely says it's unimplemented.

PR #3622, opened 2019-01-30
md5crypt-opencl optimizations favoring same-length passwords

solardiz:
This is much better than we had before, but is still much worse than Hashcat and presumably Hash Suite. A drawback is that speeds at mixed-length passwords change slightly to the worse, so it becomes more important to group candidate passwords by length.

```
[solar@super run]$ rm -r ~/.nv/ComputeCache
[solar@super run]$ LWS=256 GWS=10240 ./john -test=5 -form=md5crypt-opencl -dev=3 -v=5 -mask='?a?a?a?a?a?a?a' 
initUnicode(UNICODE, ASCII/ASCII)
ASCII -> ASCII -> ASCII
Device 3: GeForce GTX 1080
Benchmarking: md5crypt-opencl, crypt(3) $1$ [MD5 OpenCL]... Loaded 61 hashes with 39 different salts to test db from test vectors
Options used: -I /home/solar/j/bleeding-jumbo-20190120/run/kernels -cl-mad-enable -DSM_MAJOR=6 -DSM_MINOR=1 -cl-nv-verbose -D__GPU__ -DDEVICE_INFO=524306 -DSIZEOF_SIZE_T=8 -DDEV_VER_MAJOR=410 -DDEV_VER_MINOR=79 -D_OPENCL_COMPILER -DPLAINTEXT_LENGTH=15 $JOHN/kernels/cryptmd5_kernel.cl
Build log: 
ptxas info    : 0 bytes gmem, 54 bytes cmem[3]
ptxas info    : Compiling entry function 'cryptmd5' for 'sm_61'
ptxas info    : Function properties for cryptmd5
ptxas         .     480 bytes stack frame, 0 bytes spill stores, 0 bytes spill loads
ptxas info    : Used 32 registers, 344 bytes cmem[0], 108 bytes cmem[2]

Local worksize (LWS) 256, global worksize (GWS) 10240 (40 blocks)
DONE
Raw:	7198K c/s real, 7198K c/s virtual, GPU util: 98%
```

### DIFF
--- a/src/opencl/cryptmd5_kernel.cl
+++ b/src/opencl/cryptmd5_kernel.cl
@@ -2,6 +2,7 @@
  * This software is
  * Copyright (c) 2011 - 2013 Lukas Odzioba
  * Copyright (c) 2012 - 2013 magnum
+ * Copyright (c) 2015, 2018 Solar Designer
  * and it is hereby released to the general public under the following terms:
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted.
@@ -9,15 +10,11 @@
 
 #include "opencl_misc.h"
 
-#if nvidia_sm_3x(DEVICE_INFO)
-#define USE_BITSELECT 1
-#endif
-
-#if gpu_nvidia(DEVICE_INFO)
-#define BITALIGN_AGGRESSIVE
-#else
+//#define USE_BITSELECT 1
+//#define BITALIGN_AGGRESSIVE
 #define BUF_UPDATE_SWITCH
-#endif
+//#define UNROLL_AGGRESSIVE
+//#define UNROLL_LESS
 
 #undef MD5_LUT3 /* No good for this format, just here for reference */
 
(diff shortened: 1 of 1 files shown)


solardiz:
BTW, apparently (per my testing) specifying `-mask` along with `-test` disables self-test (so we'd see speeds even for an unusable kernel)?! I find this weird.

I did also test this without mask, which should have run self-test.

claudioandre-br:
See: https://github.com/magnumripper/JohnTheRipper/issues/3249

solardiz:
I realized there's an issue with this PR - I think it might produce occasional (very rare) false negatives. We have these patterns (per a comment in the code):

```c
        //pattern[0]=alt pass
        //pattern[1]=alt pass pass
        //pattern[2]=alt salt pass
        //pattern[3]=alt salt pass pass
        //pattern[4]=pass alt
        //pattern[5]=pass pass alt
        //pattern[6]=pass salt alt
        //pattern[7]=pass salt pass alt
```

In patterns 5 and 7 and depending on pass and salt lengths, `x[9]` might be part of alt, where that value being 0 doesn't guarantee that all further ones are 0 as well. (I think `x[9]` might also be the last component of alt in pattern 6, but that's OK.)

I'll think of a fix.

PR #3632, opened 2019-02-03
Run more SHA512crypt loops per kernel call

claudioandre-br:
Using latest (powerful) GPUs, the performance gain can be > 3%.

Still need to stress it on super (not now).


magnumripper:
I'll let you merge it yourself when you want to

PR #3637, opened 2019-02-05
[List.ZTEX:Devices] - ability to hide Serial Numbers

Apingis:
Each board has a factory programmed Serial Number (SN). JtR displays
SNs on startup and in information/error messages. To hide SNs, you can
list SNs of your boards in [List.ZTEX:Devices] section in john.conf.
Board numbers (starting from 1) will appear instead of SNs. You can
specify these numbers in --dev command-line option.

### DIFF
--- a/run/john.conf
+++ b/run/john.conf
@@ -395,6 +395,13 @@ AbortTemperature = 95
 
 
 # ZTEX specific settings
+[List.ZTEX:Devices]
+# If you list Serial Numbers (SN) of ZTEX boards here, it will display
+# numbers (starting from 1) instead of factory programmed SN's.
+# These numbers can be used in --dev command-line option.
+#04A31E0123
+#00AD034E01
+
 [ZTEX:descrypt]
 # The design has two programmable clocks. The 1st one is supplied to
 # pipelines of DES rounds, the 2nd clock is supplied to comparators.
--- a/src/ztex/device_format.c
+++ b/src/ztex/device_format.c
@@ -1,5 +1,5 @@
 /*
- * This software is Copyright (c) 2016-2018 Denis Burykin
+ * This software is Copyright (c) 2016-2019 Denis Burykin
  * [denis_burykin yahoo com], [denis-burykin2014 yandex ru]
  * and it is hereby released to the general public under the following terms:
  * Redistribution and use in source and binary forms, with or without
@@ -9,6 +9,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <sys/time.h>
+#include <ctype.h>
 
 #include "../loader.h"
 #include "../formats.h"
@@ -22,12 +23,8 @@
 #include "jtr_mask.h"
 #include "device_bitstream.h"
 #include "device_format.h"
+#include "ztex_sn.h"
 
-// Problem: Inclusion of ztex.h results in inclusion of libusb-1.0.h
-// which includes more stuff, on Cygwin that results in redefinition
-// of MEM_FREE macro. Using forward declaration instead.
-//#include "ztex.h"
-extern int ztex_sn_is_valid(char *sn);
 
 // If some task is not completed in this many seconds,
 // then it counts the device as not functioning one.
@@ -79,15 +76,35 @@ void device_format_init(struct fmt_main *fmt_main,
 	jtr_devices_allow = devices_allow;
 	jtr_verbosity = verbosity;
 
-	struct list_entry *entry;
-	int found_bad_sn = 0;
-	for (entry = devices_allow->head; entry; entry = entry->next)
-		if (!ztex_sn_is_valid(entry->data)) {
-			fprintf(stderr, "Error: bad Serial Number '%s'\n", entry->data);
-			found_bad_sn = 1;
+
+	static int init_conf_devices = 0;
+	if (!init_conf_devices) {
+
+		// Initialize [List.ZTEX:Devices] configuration section.
+		ztex_sn_init_conf_devices();
+
+		// Check -dev command-line option.
+		struct list_entry *entry;
+		int found_error = 0;
+		for (entry = devices_allow->head; entry; entry = entry->next) {
+
+			if (isdigit(entry->data[0]) && (entry->data[1] == 0
+					|| (isdigit(entry->data[1]) && entry->data[2] == 0)) ) {
+				if (!ztex_sn_check_alias(entry->data))
+					found_error = 1;
+			}
+
+			else if (!ztex_sn_is_valid(entry->data)) {
+				fprintf(stderr, "Error: bad Serial Number '%s'\n", entry->data);
+				found_error = 1;
+			}
 		}
-	if (found_bad_sn)
-		error();
+		if (found_error)
+			error();
+
+	init_conf_devices = 1;
+	}
+
 
 	// Initialize hardware.
 	// Uses globals: jtr_device_list, device_list, jtr_bitstream.
--- a/src/ztex/ztex.c
+++ b/src/ztex/ztex.c
@@ -15,6 +15,7 @@
 #include <libusb-1.0/libusb.h>
 
 #include "ztex.h"
+#include "ztex_sn.h"
 
 #include "../path.h"
 
@@ -71,23 +72,6 @@ int vendor_request(struct libusb_device_handle *handle, int cmd, int value, int 
 //
 ////////////////////////////////////////////////////////////////////////////////
 
-// checks if given string is a valid Serial Number
-int ztex_sn_is_valid(char *sn)
-{
-	int i;
-	for (i = 0; i < ZTEX_SNSTRING_LEN; i++) {
-		if (!sn[i])
-			return i < ZTEX_SNSTRING_MIN_LEN ? 0 : 1;
-		if ( !( (sn[i] >= '0' && sn[i] <= '9') || (sn[i] >= 'A' && sn[i] <= 'F')
-				|| (sn[i] >= 'a' && sn[i] <= 'f')) )
-			return 0;
-	}
-	if (sn[i])
-		return 0;
-
-	return 1;
-}
-
 // Creates 'struct ztex_device' out of 'libusb_device *'
 // gets data from the device
 int ztex_device_new(libusb_device *usb_dev, struct ztex_device **ztex_dev)
@@ -134,7 +118,7 @@ int ztex_device_new(libusb_device *usb_dev, struct ztex_device **ztex_dev)
 	}
 
 	result = libusb_get_string_descriptor_ascii(dev->handle, desc.iSerialNumber,
-			(unsigned char *)dev->snString, ZTEX_SNSTRING_LEN);
+			(unsigned char *)dev->snString_orig, ZTEX_SNSTRING_LEN);
 	if (result < 0) {
 		ztex_error("ztex_device_new: libusb_get_string_descriptor_ascii(iSerialNumber): %s\n",
 				result, libusb_error_name(result));
@@ -142,13 +126,9 @@ int ztex_device_new(libusb_device *usb_dev, struct ztex_device **ztex_dev)
 		return result;
 	}
 
-	// Before firmware upload, board may have SN of different format
-	//
-	//if (!ztex_sn_is_valid(dev->snString)) {
-	//	ztex_error("ztex_device_new: bad Serial Number (%s)\n", dev->snString);
-	//	ztex_device_delete(dev);
-	//	return -1;
-	//}
+	// Original SN remains in dev->snString_orig.
+	char *sn = ztex_sn_get_by_sn_orig(dev->snString_orig);
+	strncpy(dev->snString, sn, ZTEX_SNSTRING_LEN);
 
 	result = libusb_get_string_descriptor_ascii(dev->handle, desc.iProduct,
 			(unsigned char *)dev->product_string, ZTEX_PRODUCT_STRING_LEN);
@@ -319,7 +299,7 @@ void ztex_dev_list_print(struct ztex_dev_list *dev_list)
 	}
 }
 
-
+/*
 // Finds valid device with given Serial Number in ztex_dev_list
 struct ztex_device *ztex_find_by_sn(struct ztex_dev_list *dev_list, char *sn)
 {
@@ -335,6 +315,7 @@ struct ztex_device *ztex_find_by_sn(struct ztex_dev_list *dev_list, char *sn)
 	}
 	return NULL;
 }
+*/
 
 // Finds valid device by libusb_device *
 struct ztex_device *ztex_find_by_usb_dev(struct ztex_dev_list *dev_list, libusb_device *usb_dev)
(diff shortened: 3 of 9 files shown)


solardiz:
Thanks @Apingis!

This currently numbers the devices starting from 1. @magnumripper @claudioandre-br Do you think we should keep that or maybe switch to 0-based numbering to be consistent with what we do for OpenCL devices? And if we keep it 1-based here, should we possibly switch the numbering of OpenCL devices to be 1-based as well, for consistency with Hashcat? The 0-based numbering is consistent with OpenCL APIs, but that consistency breaks once there's more than one OpenCL platform anyway as OpenCL's numbering is per-platform and ours (and Hashcat's) is global across all platforms.

run/john.conf
```diff
@@ -395,6 +395,13 @@ AbortTemperature = 95
 
 
 # ZTEX specific settings
+[List.ZTEX:Devices]
+# If you list Serial Numbers (SN) of ZTEX boards here, it will display
+# numbers (starting from 1) instead of factory programmed SN's.
+# These numbers can be used in --dev command-line option.
+#04A31E0123
+#00AD034E01
```
solardiz: Should this possibly use the same example SNs that README-ZTEX does?

src/ztex/device_format.c
```diff
@@ -79,15 +76,35 @@ void device_format_init(struct fmt_main *fmt_main,
...
-	struct list_entry *entry;
-	int found_bad_sn = 0;
-	for (entry = devices_allow->head; entry; entry = entry->next)
-		if (!ztex_sn_is_valid(entry->data)) {
-			fprintf(stderr, "Error: bad Serial Number '%s'\n", entry->data);
-			found_bad_sn = 1;
+
+	static int init_conf_devices = 0;
+	if (!init_conf_devices) {
+
```
solardiz: Inconsistent empty line.

src/ztex/device_format.c
```diff
@@ -79,15 +76,35 @@ void device_format_init(struct fmt_main *fmt_main,
...
+		// Initialize [List.ZTEX:Devices] configuration section.
+		ztex_sn_init_conf_devices();
+
+		// Check -dev command-line option.
+		struct list_entry *entry;
+		int found_error = 0;
+		for (entry = devices_allow->head; entry; entry = entry->next) {
+
```
solardiz: Another inconsistent empty line.

src/ztex/device_format.c
```diff
@@ -79,15 +76,35 @@ void device_format_init(struct fmt_main *fmt_main,
...
+			if (isdigit(entry->data[0]) && (entry->data[1] == 0
+					|| (isdigit(entry->data[1]) && entry->data[2] == 0)) ) {
```
solardiz: ctype macros are defined to operate on `int` and are allowed to malfunction when used on `char`, especially when the 8th bit is set, although typical libc's work around these potential issues. They're also dependent on current locale, which isn't the behavior we want here. I think it's preferable to use comparisons against '0' and '9' here, or maybe use our custom `isdec` function and a length check.

src/ztex/device_format.c
```diff
@@ -79,15 +76,35 @@ void device_format_init(struct fmt_main *fmt_main,
...
+				if (!ztex_sn_check_alias(entry->data))
+					found_error = 1;
+			}
+
+			else if (!ztex_sn_is_valid(entry->data)) {
```
solardiz: Coding style: I'd write this as `} else if (` on one line.

src/ztex/device_format.c
```diff
@@ -79,15 +76,35 @@ void device_format_init(struct fmt_main *fmt_main,
...
+				fprintf(stderr, "Error: bad Serial Number '%s'\n", entry->data);
+				found_error = 1;
+			}
 		}
-	if (found_bad_sn)
-		error();
+		if (found_error)
+			error();
+
+	init_conf_devices = 1;
```
solardiz: Wrong indentation.

src/ztex/ztex.c
```diff
@@ -335,6 +315,7 @@ struct ztex_device *ztex_find_by_sn(struct ztex_dev_list *dev_list, char *sn)
 	}
 	return NULL;
 }
+*/
```
solardiz: Let's use `#if 0` ... `#endif` rather than "comments" for excluding pieces of code.

src/ztex/ztex.h
```diff
@@ -42,6 +43,7 @@
 #define CAPABILITY_MAC_EEPROM 0,6
 // Capability index for multi FPGA support.
 #define CAPABILITY_MULTI_FPGA 0,7
+// Copy-pasted from ZTEX SDK! (java)
```
solardiz: It's unclear how much is copy-pasted. If it's just one line, then it's confusing why the comment was needed. If it's much more than that, there can be copyright+licensing concerns.

src/ztex/ztex_scan.c
```diff
@@ -57,14 +57,28 @@ static int ztex_scan(struct ztex_dev_list *new_dev_list, struct ztex_dev_list *d
...
 				continue;
 			}
 		}
+		/*
+		if (jtr_devices_allow->count && firmware_is_ok(dev)) {
+			if (!list_check(jtr_devices_allow, dev->snString)) {
+				ztex_dev_list_remove(new_dev_list, dev);
+				continue;
+			}
+		}*/
```
solardiz: Let's use `#if 0` ... `#endif` rather than "comments" for excluding pieces of code.

src/ztex/ztex_sn.c
```diff
@@ -0,0 +1,119 @@
...
+
+	if (!ztex_sn_conf_devices) {
+		if (!ztex_sn_get_by_alias_error) {
+			fprintf(stderr, "Error: [List.ZTEX:Devices] not found\n");
+			ztex_sn_get_by_alias_error = 1;
+		}
+		return 0;
+	}
+
+	int id = atoi(alias); // alias specified in -dev option
```
solardiz: This assumes `alias` has been validated by the caller to actually be a decimal number, right? If that's not the intent, we could check with our `isdec` here.

src/ztex/ztex_sn.c
```diff
@@ -0,0 +1,119 @@
...
+		fprintf(stderr, "Error: invalid -dev=0. Devices are numbered "
+			"starting from 1.\n");
+		return 0;
+	}
+
+	struct cfg_line *line;
+	int last_id = 0;
+	for (line = ztex_sn_conf_devices->head; line; line = line->next) {
+		if (id == line->id + 1)
+			return 1;//line->data;
```
solardiz: Why the commented-out `line->data`?

magnumripper:
> This currently numbers the devices starting from 1. @magnumripper @claudioandre-br Do you think we should keep that or maybe switch to 0-based numbering to be consistent with what we do for OpenCL devices? And if we keep it 1-based here, should we possibly switch the numbering of OpenCL devices to be 1-based as well, for consistency with Hashcat? The 0-based numbering is consistent with OpenCL APIs, but that consistency breaks once there's more than one OpenCL platform anyway as OpenCL's numbering is per-platform and ours (and Hashcat's) is global across all platforms.

Personally I prefer 0-based. IMHO Hashcat is very confusing with its 1-based numbers not matching logs, tools and diagnostics.

src/ztex/ztex_sn.c
Apingis: ```ztex_sn_check_alias()``` checks if supplied argument is correct,
without a need to reveal actual SN.

src/ztex/ztex.c
Apingis: OK, I see a rationale of ```#if 0 ... #endif``` - the construct can be nested.

src/ztex/ztex_sn.c
Apingis: Indeed ```alias``` has been validated by the caller. The caller has
to distinguish between alias and actual SN, perform proper check,
take proper branch.
A suggestion to validate input to the function is definitely a good
idea, results in cleaner and more stable codebase.
To avoid code duplication I'll create ```ztex_sn_alias_is_valid()```.

src/ztex/device_format.c
Apingis: I use to insert a blank line where I feel code is too condensed, hard to read/understand, however I'm not ready to provide a formalized criteria for "density". Could you please explain what is "inconsistency" here from your point of view?

run/john.conf
Apingis: I'm not sure if that matters. In both places there're example SNs rather than ones from actual boards.
Probably I can choose a few SNs that looks like they are "example" SNs, use them across the documentation.

src/ztex/device_format.c
Apingis: OK, agreed.

src/ztex/device_format.c
Apingis:
```suggestion
		init_conf_devices = 1;
```

src/ztex/device_format.c
Apingis: I'll remove blank line (like ```indent -ce``` would do).
I'm not aware of coding style guidelines on the issue, and my preferece (as I explained in the above comment) is to do it base on a feeling if that's "too condensed" or not.
Sometimes I write ```} else if (```, ```} else {``` on one line.

src/ztex/ztex.h
Apingis: I've copied definitions of bits in configuration bytes from ZTEX specific USB descriptor. There was some issue, I've commented with exclamation to return later. I'll take a look and write more descriptive comment.

solardiz:
@magnumripper OK, thanks for your opinion.

We use 1-based numbers for forks and nodes, though - is this something we should change, too? I think not. I think e.g. `--node=13-24/24` is clearer than `--node=12-23/24` would be, so probably that should stay 1-based.

I think there are no standard ZTEX tools that use device numbers.

So should we ask @Apingis to implement 0-based ZTEX device numbers in JtR?

src/ztex/ztex_sn.c
solardiz: Sure, I meant what meaning does the comment carry to someone reading the code, and if it isn't a relevant meaning anymore then should the comment perhaps be dropped.

src/ztex/device_format.c
solardiz: I think/hope we never have empty lines between the `if` or such and the following code block, like you did here and maybe in your other code contributions. To me, that empty line breaks the logical flow of code and makes the code harder to read, not easier.

We sometimes have blank lines _before_ the `if` in cases where that block needs to be separated for easier reading. In your case, the blank line before `static int init_conf_devices = 0;` looks sufficient to separate this portion of code from what preceded it.

run/john.conf
solardiz: Right, that's what I suggest - choose some example SNs and use them consistently. Since you already had some in README-ZTEX, I thought you'd just continue to use them in other examples.

src/ztex/device_format.c
magnumripper: On another note we currently require declarations to be in top of block in the rest of Jumbo, so this `static int init_conf_devices` will emit a `-Wdeclaration-after-statement` warning (or even error with `-Werror`). I have no plans to enforce that for the ZTEX code though, perhaps we should even drop that requirement some day.

magnumripper:
I had a hard time getting used to 1-based nodes (because MPI numbers are 0-based) but eventually I did, and I agree about it being more clear. So I'm not sure. I could get used to 1-based OpenCL numbers as well.

solardiz:
@magnumripper So do we keep the 1-based numbers in this PR, and only ask @Apingis to fix the coding style issues? And then you change the OpenCL numbers to be 1-based as well? Personally, I found the JtR vs. Hashcat discrepancy to be worse than JtR vs. OpenCL APIs/tools. I think we don't need to match OpenCL APIs/tools because our device numbers are across all OpenCL platforms, whereas OpenCL's are per-platform. Switching to 1-based would also allow for forked node numbers to match GPU device numbers in common special cases where exactly the first few devices are used and no Intel OpenCL for CPU is installed.

magnumripper:
> @magnumripper So do we keep the 1-based numbers in this PR, and only ask @Apingis to fix the coding style issues? And then you change the OpenCL numbers to be 1-based as well?

OK, I'll open an issue for 1-based OpenCL.

solardiz:
@Apingis Are you going to push more commits to this PR to fix the coding style issues, or open a separate PR for that? I'm not merging this one yet in case you'd prefer to add more commits to here.

Apingis:
I'd prefer this to be merged as-is, then open a new PR with corrections.

solardiz:
I'm now testing this and #3643 at once. I have four boards where one board changes its SN upon firmware upload. I listed that one's final/real SN on line 4, and initial/weird SN on line 5. I am getting:

```
$ ./john -2='?d?u' --mask='?2?2?2?2?2' --format=sha512crypt-ztex --verbosity=1 pw-sha512crypt 
SN 1: firmware uploaded
SN 5: firmware uploaded
SN 2: firmware uploaded
SN 3: firmware uploaded
SN 3: uploading bitstreams.. ok
SN 2: uploading bitstreams.. ok
SN 4: uploading bitstreams.. ok
SN 1: uploading bitstreams.. ok
ZTEX 3 bus:2 dev:30 Frequency:160 160 160 160 
ZTEX 2 bus:2 dev:31 Frequency:160 160 160 160 
ZTEX 4 bus:2 dev:32 Frequency:160 160 160 160 
ZTEX 1 bus:2 dev:29 Frequency:160 160 160 160 
Loaded 512 password hashes with no different salts (sha512crypt-ztex, crypt(3) $6$ [sha512crypt ZTEX])
104g 0:00:00:09 4.12% (ETA: 23:13:26) 11.31g/s 270763p/s 270763c/s 124314KC/s 11IS0..5YIS0
206g 0:00:00:21 9.77% (ETA: 23:13:22) 9.501g/s 272590p/s 272590c/s 110772KC/s 11HT3..5YHT3
222g 0:00:00:25 11.32% (ETA: 23:13:28) 8.851g/s 272842p/s 272842c/s 107140KC/s 40429..5YH29
512g 0:00:00:56 DONE (2019-02-12 23:10) 8.991g/s 273129p/s 273129c/s 71046KC/s 40477..##E77
Session completed
```

This looks fine to me.

I am also able to request use of individual boards by listing their numbers in `-dev`. Great.

And when I try to use device number 0 or device number 6 or higher, I get an appropriate error message. Great.

What doesn't work:

When I list a device number that's present in `john.conf`, but isn't currently available among online devices, I get no error and JtR silently uses fewer devices. So e.g. if I use `-dev=1,5` after firmware had already been uploaded to the fourth board (so it matches the SN on line 4 rather than line 5), JtR starts up and silently uses only one board. I would have liked it to complain (but perhaps as a Warning only, then proceed working) about the requested yet missing board.

Also, referring to the boards by their full SNs no longer works when those SNs are listed in `john.conf`. I expected that either would work.

But these are very minor issues that I'm OK to ignore.

PR #3641, opened 2019-02-08
Add a safe-guard against logins/plaintexts that upset the terminal

magnumripper:


solardiz:
I think the safe-guard should be to avoid printing control characters in the first place, not try to recover after the fact because that won't always fully undo the damage. Some logins/plaintexts might not merely confuse the terminal - they might be actively malicious, playing tricks like overwriting some other part of output to make it misleading or changing the window title or even exploiting a vulnerability in the terminal emulator code (or e.g. X library or server code, where it was in fact possible to overflow a buffer through over-long window title many years ago).

roycewilliams:
Naive question: couldn't any discovered plaintext that isn't detected as ASCII or UTF-8 simply be HEX-ified prior to output?

magnumripper:
> I think the safe-guard should be to avoid printing control characters in the first place

How, exactly, would we do that? Our current code has no knowledge about what the terminal would see as a control character or not, nor whether UTF-8 is supported or not. We're effectively running "raw". This whole thing is actually only a problem when user options/input are incorrect, for instance running a wordlist that don't match the chosen (or default) input encoding.

magnumripper:
> Naive question: couldn't any discovered plaintext that isn't detected as ASCII or UTF-8 simply be HEX-ified prior to output?

We could be slightly more intelligent than that (eg. allowing ISO-8859 or DOS CPxxx or Windows "ANSI" CPxxxx as opposed to UTF-8 depending on our output encoding) but that's another issue. I'm pretty sure we had an issue but I can't find it right now.

solardiz:
@magnumripper Just like some other tools do, like Linux "procps" commands, we'd assume that anything in the ranges '\0' to '\x1f' and '\x80' to '\x9f' are potentially control characters. The latter range overlaps with some CPxxxx, but unfortunately this _does_ mean those CP's are not fully usable on a terminal or emulator (they were on DOS). Alternatively, or as an option, in the 8-bit range we could exclude only '\x9b', even though many terminals actually do process a few other (commonly less dangerous) control characters in that range.

magnumripper:
Well we could assume that `[|x00-\x08\x0b-\x1f]` are dangerous. For virtually anything else, we'd need to change default behaviors and/or add options. I'm pretty sure most users are just run Jumbo without caring much about encodings, and most often that will work just fine and show eg. UTF-8 passwords (as well as "OEM" codepage ones in case of LM, converted to UTF-8 on screen) properly.

If we are to do this, one option is to revert to being much stricter than John proper, and only allow [printable] ASCII unless user explicitly specifies an output encoding. That would however mean we'd need to add such option. We actually don't have any as of now: We have input encoding, intermediate encoding and target encoding but none of them has anything to do with the terminal - it's all about the hash. In `john.conf` we do have a couple of options saying that the *terminal* is UTF-8 though, and they are the default.

This all would be a fairly well defined and "easy" project but it'd be pretty extensive and I don't see much benefit (I see virtually none). I'm pretty sure literally more than 99% of all users are running an UTF-8 terminal.

solardiz:
I think `[\x00-\x08\x0b-\x1f]` is not enough. Why exclude TAB and LF? They're not supposed to be in usernames and tested/cracked passwords, or if they are then we want to protect against them. More importantly, we must exclude at least `\x9b` and preferably all of `[\x7f-\x9f]`.

I think there's no need to limit ourselves to 7-bit ASCII by default. The above should be fine.

magnumripper:
While tabs and linefeeds are extremely rare in passwords, I've seen them in the wild several times and they're not "dangerous", just a little messy. Anyway we should probably pick the `$HEX[deadbabe]` route like Hashcat. This PR will stay until then although I could live with changing the default.

PR #3644, opened 2019-02-11
[SIP] Wrong comment statement which may confuse

antnks:
Wrong comment statement which may confuse: NONCE_COUNT should go prior to CNONCE

RFC: https://tools.ietf.org/html/rfc2617#section-3.2.2.1
request-digest  = <"> < KD ( H(A1),     unq(nonce-value)
                                          ":" nc-value
                                          ":" unq(cnonce-value)
                                          ":" unq(qop-value)
                                          ":" H(A2)
The code which actually concatenates the values however is correct:
snprintf(salt.static_hash_data, STATIC_HASH_SIZE, ":%s:%s:%s:%s:%s",
				login.nonce, login.nonce_count, login.cnonce,
				login.qop, static_hash);


magnumripper:
Thanks

Issue #3642, opened 2019-02-09
Add zip support info to --list=build-info


claudioandre-br:
See https://www.openwall.com/lists/john-users/2019/02/09/1

Something simple. E,g,:
```
#if HAVE_LIBZ
    printf("pkzip support: yes")
#else
    printf("pkzip support: no")
#endif
```

claudioandre-br:
```
$ john test.hash 
Using default input encoding: UTF-8
Loaded 1 password hash (PKZIP [32/64])
Will run 2 OpenMP threads
Proceeding with single, rules:Single
Press 'q' or Ctrl-C to abort, almost any other key for status
Warning: Only 3 candidates buffered for the current salt, minimum 8
needed for performance.
Warning: Only 6 candidates buffered for the current salt, minimum 8
needed for performance.
Warning: Only 5 candidates buffered for the current salt, minimum 8
needed for performance.
Warning: Only 4 candidates buffered for the current salt, minimum 8
needed for performance.
Warning: Only 2 candidates buffered for the current salt, minimum 8
needed for performance.
Almost done: Processing the remaining buffered candidate passwords, if any
Warning: Only 6 candidates buffered for the current salt, minimum 8
needed for performance.
Proceeding with wordlist:/snap/john-the-ripper/current/run/password.lst, rules:Wordlist
12345            (test.zip)
1g 0:00:00:00 DONE 2/3 (2019-02-09 20:23) 1.754g/s 97270p/s 97270c/s 97270C/s 123456..Peter
Use the "--show" option to display all of the cracked passwords reliably
Session completed

$ cat test.hash 
[...]

PR #3648, opened 2019-02-16
List disabled formats during build in --list=build-info

claudioandre-br:
I prefer a simple solution. See #3642 

And it is a fix that can be applied now.

### DIFF
--- a/src/listconf.c
+++ b/src/listconf.c
@@ -131,6 +131,7 @@ static void listconf_list_method_names()
 static void listconf_list_build_info(void)
 {
 	char DebuggingOptions[512], *cpdbg=DebuggingOptions;
+	char *fmt_disabled = DebuggingOptions;
 #ifdef __GNU_MP_VERSION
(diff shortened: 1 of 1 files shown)


solardiz:
@magnumripper Certainly "not a very good solution" (and isn't even much of a solution for most users, as I explained), but better than nothing while we work on a better solution.

@claudioandre-br Your new code with `sprintf` is on one hand clever and compact, but on the other too hackish and unsafe. `sprintf` "shall return a negative value" (per POSIX) on "output error". Do we want to assume that `sprintf` used like it is used here will never incur an "output error", even at the risk of moving the pointer out of bounds if it does incur such error (whatever that means - e.g., out of memory for some stdio internal buffer used to achieve thread-safety, or whatever else)? I think we should aim for correct code rather than theorize that our subtly incorrect code will probably not fail in practice. So I'd prefer one of the approaches I suggested previously. Thanks!

Issue #2789, opened 2017-09-26
Source tree clean-up / reorganize


magnumripper:
It was brought up recently in a mail discussion that our source tree is "disorganized" with like "800 source files in a single directory". This is absolutely true. I'm so used to it I haven't cared much about it. Making it better should be pretty easy. Here are two easy steps for making it a little better:
1. For a starter, we could move all plugins (and nothing else) to `src/plugins`. That will half the number of files directly under `src`... 
2. We have lots of pure OpenCL headers (for inclusion by kernels) directly under `src`, they should either go to `src/opencl` like the kernels - or perhaps `src/opencl/include` if we want to.

After that, it gets trickier to get lots of improvement with little work but we can probably find out a couple more things.

frank-dittrich:
we could also move all non-plugin formats into a separate directory (formats?), and all source files which are not used to build john, but separate binaires, into another directory (tools?).

magnumripper:
Agreed, except we can't move any core (a.k.a. "john proper") files unless that change comes from "upstream".

Theoretically we could have something like this tho'
```
src   (this would almost be as empty as in John Proper)
src/jumbo
src/jumbo/include
src/jumbo/plugins
src/jumbo/plugins/include
src/jumbo/tools
src/jumbo/tools/include  (probably overkill)
src/jumbo/aes(/...)
src/jumbo/opencl
src/jumbo/opencl/include
src/jumbo/...
```

magnumripper:
```
$ l *_plug.c | wc -l
     358
$ l opencl*_plug.c | wc -l
      79
$ bc <<< 358-79
279
$ l opencl_*h | wc -l
      45
$ l *_plug.h | wc -l
       2
$ l opencl*_plug.h | wc -l
       1
```
Most plug headers aren't named `plug` (as they don't need to), eg. `dmg_common_plug.c` vs. `dmg_common.h`, so the header figures above are way low.

claudioandre:
Anyone with perl skills can change  plugin_deps.pl? The `-f $object` at line 30 has to \#include (at least):
```
src/jumbo/include              // general *.h
src/jumbo/opencl/include  // OpenCL *.h
```

I tried, but the result is ridiculous:
```
diff --git a/src/plugin_deps.pl b/src/plugin_deps.pl
index f98f486de..991f5fc25 100755
--- a/src/plugin_deps.pl
+++ b/src/plugin_deps.pl
[...]

PR #3652, opened 2019-02-20
Break source in small code units

claudioandre-br:
Is there any acceptable way to deal with "source packages" in C? 
* one way I can create small units of code and include them once in other source files.

To simplify code management since.
- `-list=opencl-devices` code is self-contained;
- kernel build code is self-contained;
- etc.

I can move them all to another `.c` files (but I dont want to change the `#include common-opencl.h`).

The `common-opencl.h` is the **package**, included once.
****

The PR is an example


magnumripper:
There's nothing wrong with sourcing (with `#include`) a C source file. I would eg. break out the list-devices stuff in some new file (eg. `opencl_list_dev.c`) and simply include it from `opencl_common.c`.

Some people would call that new file `opencl_list_dev.h` even though it contains code - just because it's meant to be sourced but IMHO that's very confusing. Headers are one thing, code is another. Both can be sourced with `#include`.

magnumripper:
I approved but didn't merge. In general, you can merge at will once any of us has approved.

claudioandre-br:
People would say that a `.c` file is meant to produce a `.o` file. 
* Anyway, I would wait to see if any C-ist guy complains.

solardiz:
Thanks, Claudio. I like having more smaller files with good logical separation between them. However:

I strongly dislike the introduction of a subdirectory called `code`. I think this is confusing. Why introduce a subdirectory at all?

I mildly dislike `#include` of a `.c` file. I sometimes do that myself, but in unusual cases where I think I have more significant reasons to do it. In this case, I think it can be a usual `.c` file compiled into an `.o` file on its own if we simply export 3 functions from the "base" file - right? If so, maybe namespace those functions by adding some name prefix to them, and export them, so that the files can be compiled separately and by usual means.

claudioandre-br:
> I strongly dislike the introduction of a subdirectory called `code`.

We are procrastinating #2789. We will need to, at least:
```
_ src
|_ opencl
     |_ kernels
     |_ include
     |_ code (another src. Should all formats and helper routines be at OpenCL root?)
```

PR #3655, opened 2019-02-25
Fixes command line prompt ambiguity

noahluddy:
Removes the `$ ` prompt for two commands. I think this is appropriate because none of the earlier commands include the `$ ` prompt.

### DIFF
--- a/doc/INSTALL-UBUNTU
+++ b/doc/INSTALL-UBUNTU
@@ -83,11 +83,11 @@ system.
 
 === Test your build
 
-    $ ../run/john --test=0
+    ../run/john --test=0
 
 === Benchmark your build
 
-    $ ../run/john --test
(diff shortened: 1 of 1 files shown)


frank-dittrich:
I think we should rather add the missing `$ ` (in INSTALL-FEDORA as well), since a lot of other docs use `$ ` and `# ` to distinguish command that should be started as a regular user from commands to be started as root.

frank-dittrich:
OTOH, even core doc/EXAMPLES file doesn't use `$ `.

Whatever we do, we should try to be consistent across documents.

magnumripper:
This was inconsistency within a single document, so I just merged it for now.

frank-dittrich:
INSTALL-FEDORA has the same inconsistency.

PR #3657, opened 2019-02-25
Use base1 --device

claudioandre-br:
Waiting CI

### DIFF
--- a/src/opencl_common.c
+++ b/src/opencl_common.c
@@ -669,7 +669,9 @@ static void add_device_to_list(int sequential_id)
 	found = get_if_device_is_in_use(sequential_id);
 
 	if (found < 0) {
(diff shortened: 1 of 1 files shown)


src/opencl_common.c
```diff
@@ -2943,7 +2950,7 @@ void opencl_list_devices(void)
 			p = dname;
 			while (isspace(ARCH_INDEX(*p))) /* Intel quirk */
 				p++;
```
solardiz: Unrelated to the changes, but: the "Intel quirk" comment is wrong (it's not allowed to use ctype macros on "char" - they're to be used on "int" - and Intel isn't to blame), and we should have this code depend on current locale. So this code is ideally changed to:

```c
while (*p == ' ' || *p == '\t')
        p++;
```

Issue #3566, opened 2018-12-25
Autoconf <-> Makefile overhaul


magnumripper:
There's some terrifically b0rken behaviors we should fix (probably after next release) in our build process, in that several things happen in `./configure` that definitely should be in `make`, and perhaps some case of vice versa (although I'm not sure I can think of anything).

Here's how it is supposed to be in my firm opinion:
* `./configure` should set things up for the system you build on. That is, check arch, CPU type, available libs, handle *options* like `--with-openmpi` or "systemwide" stuff and so on. Most is about the **system** and very little is about **JtR Jumbo**. It should definitely not handle anything about checking Jumbo plugins or dependencies (that's what a makefile is all about!).
* `make` should merely build any files missing or not up to date according to whatever spec came from last `./configure`.

If we get this straight and you do a `git pull`, ideally you shouldn't **ever** have to re-config or `make clean`. Only `make`. That's even if you got a few new plugin formats with that git update.
However, if you just installed a GPU and OpenCL (and didn't have that before) or decide to try out the rexgen lib, you'd need a re-conf and full rebuild.

To do:
- [ ] The plugin handling done in `./configure` should be in Makefile just like it was before we added autoconf to Jumbo. We got too excited and tried to move just about everything from Makefile to autoconf, and ended up with rather silly behavior ever since.
- [ ] The dependency handling (currently `plugin_deps.pl`) happens in `./configure` while it should happen during `make`. Moreover, I believe that script should possibly be ditched and we should learn how to do it with make and/or gcc themselves (I can't recall the details but a `deps:` target is often involved).
- [ ] The creation of `dynamic_big_crypt.c` from `dynamic_big_crypt_header.cin` and `dynamic_big_crypt_hash.cin` happens in `./configure` while it too should totally be in Makefile.
- [ ] Probably some more things, add here...

magnumripper:
> Moreover, I believe that script should possibly be ditched and we should learn how to do it with make and/or gcc themselves (I can't recall the details but a deps: target is often involved).

I think #3842 is the answer for ditching that script hack

magnumripper:
> * [ ]  The creation of `dynamic_big_crypt.c` from `dynamic_big_crypt_header.cin` and `dynamic_big_crypt_hash.cin` happens in `./configure` while it too should totally be in Makefile.

This doesn't seem to be the case (anymore?), but it's still kind of a mess and has its own issue in #4869

ghost:
[...]

PR #3663, opened 2019-03-04
Move the unit tests into make check

claudioandre-br:
You guys are complaining about moving unit-tests to 'make check'. But that is an easy task. 
* Rethink how JtR should run unit-tests is another issue and can be done later. E.g., in #3566.

### DIFF
--- a/src/Makefile.in
+++ b/src/Makefile.in
@@ -184,7 +184,7 @@ endif
 #  also need to figure out -mpreferred-stack-boundary=4 and add (or whatever the stack size is)
 #
 #########################################################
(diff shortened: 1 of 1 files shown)


solardiz:
Thank you, Claudio!

I think this PR broke being able to run `make` followed by `make check`, because `unit-tests` isn't built by either target. I think we should have the `check` target depend on `unit-tests`, so it'd get built at that time.

claudioandre-br:
Strange, I tested exactly this.
```
make -sj4 && make check # after ./configure
```

claudioandre-br:
Ok, `make clean` is not removing unit-test.

PR #3664, opened 2019-03-05
keystore2john.py: Python2/3 compatibility

AlbertVeli:
Make it work with both by adding .decode('utf-8')
to hexlified data. Python2 thinks it has type 'str'
and python3 treats it as 'bytes'. After decode()
both think it has type 'str'.

This should fix issue #3658 .


magnumripper:
With @kholia absent I just assume you know what you’re doing. Thanks!

PR #3674, opened 2019-03-19
cracker.c: Detect when a reset() call lowered min. kpc, adjust warnings accordingly

magnumripper:
I think this is more or less exactly as suggested, no reason to make it more complicated.


magnumripper:
Thanks, apparently I'm quickly getting rusty  o.O

PR #3676, opened 2019-03-22
Merge from upstream John proper

magnumripper:


src/config.c
```diff
@@ -44,22 +44,22 @@ const struct cfg_section *get_cfg_db() {
...
+		return s;
 
 	e = s + strlen(s) - 1;
-	while (e >= s && (*e == ' ' || *e == '\t')) e--;
-	/*
-	 * NOTE, if there are trailing spaces, then leave 1 of them. There are
-	 * VALID rules, that need a trailing space like $   i.e. appends space
-	 */
-	if (!force && (*(e+1) == ' ' || *(e+1) =='\t'))
-		++e;
```
solardiz: Was the leaving of a trailing whitespace possibly required to accept some existing Hashcat rulesets? Not an issue in core, which doesn't support those anyway, but could be relevant in jumbo. @jfoug's changes above look like they could invoke undefined behavior - often reading one char beyond end of string - so a correct reimplementation could be needed in jumbo.

src/idle.c
```diff
@@ -73,8 +74,9 @@ int idle_requested(struct fmt_main *format)
 
 void idle_init(struct fmt_main *format)
 {
-#if !(defined(__MINGW32__) || defined (_MSC_VER) || defined(__BEOS__) || defined(__HAIKU__))
-	static int once;
+#if !defined(__BEOS__) && !defined(__HAIKU__) &&	  \
+	!defined(__MINGW32__) && !defined (_MSC_VER)
+	int old_nice;
 #endif
```
solardiz: BTW, we'd simplify this logic a lot by allowing mixed declarations and code. I no longer recall, why is it we don't allow those? I guess I was the one to support such restrictions some years back, but perhaps we no longer support a platform where they'd be relevant?

src/john.c
```diff
@@ -324,7 +324,7 @@ static void john_register_one(struct fmt_main *format)
 	/* Format disabled in john.conf */
 	if (!override_disable &&
 	    cfg_get_bool(SECTION_DISABLED, SUBSECTION_FORMATS,
-	                 format->params.label, 0)) {
+	                 (char*)format->params.label, 0)) {
```
solardiz: Instead of doing these casts, we ought to add `const` to the prototype of `cfg_get_bool`.

src/listconf.c
```diff
@@ -654,7 +654,7 @@ void listconf_parse_late(void)
 			printf(" Disabled in configuration file      %s\n",
 			       cfg_get_bool(SECTION_DISABLED,
 			                    SUBSECTION_FORMATS,
-			                    format->params.label, 0)
+			                    (char*)format->params.label, 0)
```
solardiz: Ditto.

src/config.c
magnumripper: Oh, right, I was going to revisit that thing!

src/idle.c
magnumripper: I can't recall who wanted that or why. I kinda like it but we don't need to enforce it, especially in cases like this

src/memory.h
```diff
@@ -165,7 +165,7 @@ extern void *mem_alloc_copy(void *src, size_t size, size_t align);
 /*
  * Similar to the above function, but for ASCIIZ strings.
  */
-extern char *str_alloc_copy(char *src);
+extern char *str_alloc_copy(const char *src);
```
solardiz: Just got this into core:

```diff
Added const qualifiers to arguments of mem_alloc_copy() and str_alloc_copy()

+++ Owl/packages/john/john/src/memory.c Fri Mar 22 14:58:48 2019
@@ -91,12 +91,12 @@
        return p;
 }

-void *mem_alloc_copy(void *src, size_t size, size_t align)
+void *mem_alloc_copy(const void *src, size_t size, size_t align)
 {
  	return memcpy(mem_alloc_tiny(size, align), src, size);
 }

-char *str_alloc_copy(char *src)
+char *str_alloc_copy(const char *src)
 {
  	size_t size;

+++ Owl/packages/john/john/src/memory.h Fri Mar 22 14:58:48 2019
@@ -88,11 +88,11 @@
 /*
  * Uses mem_alloc_tiny() to allocate the memory, and copies src in there.
  */
-extern void *mem_alloc_copy(void *src, size_t size, size_t align);
+extern void *mem_alloc_copy(const void *src, size_t size, size_t align);

 /*
  * Similar to the above function, but for ASCIIZ strings.
  */
-extern char *str_alloc_copy(char *src);
+extern char *str_alloc_copy(const char *src);
```

src/john.c
magnumripper: Yes, and all other cfg_get_* functions. Will you do it in core first or should I?

solardiz: OK, I'll do it in core first. Let's just not forget to remove these casts.

magnumripper:
> I didn't review if jumbo's john.conf external modes are now in a reasonable order or not. I guess my attempt at putting them in a reasonable order in core didn't map too well onto jumbo's additional external modes - for example, are jumbo's added password generator crackers near core's `Strip`?

Yep, should be fine.

src/config.c
magnumripper: Shouldn't core support a rule like `$ ` or `^ ` though? Not to mention `@ `. Sure, there are ways around it, but why would we trim it?

magnumripper: I reverted your changes for now. On a quick look I see no problem with Jim's code, I believe at worst it's reading the trailing null.

solardiz: I'd rather consistently trim trailing whitespace, and never rely on it in any of our rules. "Ways around it" is the way to go, I think. As an alternative, maybe we should stop trimming trailing whitespace at all, rather than leave just one whitespace character in there in special cases? Jim's hack feels like the worst of possible approaches.

Jim's code might be OK in terms of UB - I haven't reviewed it thoroughly enough in full context to know if it is OK or not. My original code from prior to yesterday's commit certainly had UB in it - it'd sometimes move the "e" pointer to one char before the start of string. So reverting my changes isn't the best idea long-term, but I understand we might want to do it for now.

solardiz: Looks like the right-trimming is needed primarily for the use in:

```c
        if (cfg_database && (p = strchr(line, '='))) {
                *p++ = 0;
                cfg_add_param(strlwr(trim(line)), trim(p));
```

where there might be extra whitespace before the `'='` and it shouldn't become part of the parameter name. Maybe we should handle this case specially.

solardiz: I've just committed changes to trim() and its uses in core that should also be suitable for jumbo as-is. Please merge.

src/john.c
solardiz: I've just added `const` to cfg_* and path_* functions in core.

PR #3682, opened 2019-03-24
Stop duplicating the path in cfg_init()

claudioandre-br:
As it is now, `cfg_init()` misbehaves when a file name contains directory components. To fix that and handle both situations (file names with and without directory information), the use of `basename()` is needed.

See https://github.com/magnumripper/JohnTheRipper/pull/3681#issuecomment-475896251

### DIFF
--- a/src/path.c
+++ b/src/path.c
@@ -145,7 +145,9 @@ char *path_expand_ex(char *name)
 {
 	if (john_home_pathex &&
 	    john_home_lengthex + strlen(name) < PATH_BUFFER_SIZE) {
-		strnzcpy(&john_home_pathex[john_home_lengthex], name,
+		char *p = basename(name);
+
+		strnzcpy(&john_home_pathex[john_home_lengthex], p,
 			PATH_BUFFER_SIZE - john_home_lengthex);
 		return john_home_pathex;
 	}


src/config.c
```diff
@@ -237,7 +237,7 @@ void cfg_init(char *name, int allow_missing)
 	cfg_name = str_alloc_copy(path_expand(name));
 	file = fopen(cfg_name, "r");
 	if (!file) {
-		cfg_name = str_alloc_copy(path_expand_ex(name));
+		cfg_name = str_alloc_copy(path_expand_ex(basename(name)));
```
magnumripper: `basename()` isn't quite portable so we have `jtr_basename()` in in jumbo.h for use here

claudioandre-br: As in 74a386b3775c27f288abf8bcc3a4eab107df0062 [1]? Or should I use `jtr_basename`?

[1] From jumbo.h
```.h
#undef basename
#define basename(a) jtr_basename(a)
```

claudioandre-br: `jumbo.h` is included from `misc.h`

magnumripper: Ah, sorry. That is confusing but anyway you're all set. Perhaps we should stop things like that but that's a whole different issue.

solardiz:
I'm not convinced the use of `basename()` is appropriate here. Why is it that there are already path components? Do they come from the user? Maybe we should simply keep them and not expand the path? I worry that we might be making unexpected changes to user-provided pathnames. Our goal shouldn't merely be to avoid duplicate path expansion - it should be to meet users' expectations in all cases.

claudioandre-br:
> I'm not convinced the use of basename() is appropriate here. Why is it that there are already path components? 

I would say: I'm not convinced the use of `path_expand_ex()` is needed if there are already path components in `name`.

Then, to answer this, I imagined use cases `--config` is trying to solve. So, I did this:
* I have 2 sets of different possible configurations.
  * john.conf was changed, some rules was changed. I put them in conf1 folder.
  * john.conf was changed (again), idem. I put them in conf2 folder.

```
$ mkdir ../conf1
$ mkdir ../conf2
$ cp ../run/*.conf ../conf1
$ cp ../run/*.conf ../conf2
$ cp ../run/rules/* ../conf1/
$ cp ../run/rules/* ../conf2/
$ # Change stuff
$ ../run/john --test=0 --format=raw-sha512 --config=../conf1/john.conf
```

Well, if I'll need `path_expand_ex()` to open other conf files inside conf1 folder, I need (at least) to remove `$JOHN` from them.

Anyway:
* how `--config` should do? It will only replace the main `john.conf` file? Is `path_expand_ex()` needed at all?

magnumripper:
```
$ ../run/john --test=0 --format=raw-sha512 --config=../conf1/john.conf
```
I think this command line will/should set `$JOHN` to `../run` from the binary path, then set config path to `../conf1/john.conf`, both relative from whatever CWD we had. I believe any references to `$JOHN` from that out-of-basepath config file (as well as sourcing written as `.include <file>`) will be based on `../run`. While not quite crystal clear it's fairly logical. Some users would probably expect such sourcing to also load from `../conf1` though.

Like you say, this boils down to us needing to decide the "rules" for what will happen in a case like this. I suspect whoever wrote the --config option mostly had pathless files in mind and perhaps even a `$JOHN` of `./`.

magnumripper:
Perhaps `--config=../conf1/john.conf` should imply us changing `$JOHN` to `../conf1` for everything except the john binary?

claudioandre-br:
Well, someone else have to say if `--config` is useful / relevant. And how. 

That said, my vote is:
* keep it as is (very powerfull);
* OR, remove `path_expand_ex().

solardiz:
Another way to look at it: are the include directives in our conf files relative to `$JOHN` or relative to the including file's directory? If they are (or we make them to be) relative to the including file's directory, then they'll just work right without us needing to adjust `$JOHN` in your example.

As to your vote, I don't know what the current behavior is - I've never used `--config` much, and certainly never with different directories. Maybe you can:

1. Describe the current behavior for us all, and explain how it's very powerful.

2. Explain the rationale for having a `path_expand_ex()` and what it does.

As things are, I don't know why we have `path_expand_ex()`, so removing it feels right. ;-)

claudioandre-br:
> then they'll just work right without us needing to adjust $JOHN in your example.

Not my example, right? I'm not adjusting $JOHN, I'm removing it inside `path_expand_ex`. Otherwise, a filename will became ../conf1/$JOHN/korelogic.conf

> Describe the current behavior for us all, and explain how it's very powerful.

This is how I would use it (as I put above).
* create N sets of different possible configurations (.conf and .rules files);
* put each set in its own directory (conf1, conf2, conf3);
* run the same JtR binary with the desired set (.conf and .rules files);

Disclaimer: is this necessary? is this useful? I don't know. But it seems to be.
***

Question to ALL: Is anyone using `--config`?

magnumripper:
> Question to ALL: Is anyone using `--config`?

I never use it. I see it as an artifact for the time (years and years ago) when eg. `--rules` and `--single` wouldn't take the `=SECTION` argument - the ruleset would be given in the config. Nowadays we say `--rules=foo` instead..

solardiz:
Claudio, actually I did mean _your_ example with `conf1` and `conf2` directories. I think it'll work fine if the include paths are relative to the including file's path - similar to how #include with double-quotes works in C preprocessor. The wordlist and .chr file references won't be relative, though. So maybe we do need to adjust `$JOHN` _if_ we want those to also follow the config file's location - but I don't know whether we do. Maybe someone using `--config` with a directory path nevertheless expects the specified config file's references to `$JOHN` to take those larger files from a directory shared between their multiple sets of config file directories. That's tricky.

I make almost no use of `--config`. I think I used it a few times a few years ago, and without a pathname (for files in my `run` directory, while also being in there).

frank-dittrich:
I think I used it in the past during contests, of for fuzzing using afl (http://lcamtuf.coredump.cx/afl/), but not often.

PR #3683, opened 2019-03-24
Modify cfg_name inside cfg_init() cautiously 

claudioandre-br:
Depends on #3682.
****

Tested:
```
$ ../run/john --test=0 --config=../run/john.ini           # renamed john.conf
Testing: [...]   # => Works

$ ../run/john --test=0 --config=\$JOHN/john.ini           # renamed john.conf
Testing: [...]   # => Works

$ ../run/john --test=0 --config=../abc/config.ini         # wrong conf file name
fopen: ../abc/config.ini: No such file or directory

$ ../run/john --test=0                                    # missing john.conf
fopen: ../run/john.conf: No such file or directory
```

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -230,28 +230,31 @@ void cfg_init(char *name, int allow_missing)
 {
 	FILE *file;
 	char line[LINE_BUFFER_SIZE];
+	char *tmp_name;
 	int number;
 
 	if (cfg_database && !cfg_recursion) return;
 
-	cfg_name = str_alloc_copy(path_expand(name));
-	file = fopen(cfg_name, "r");
+	tmp_name = path_expand(name);
+	file = fopen(tmp_name, "r");
 	if (!file) {
-		cfg_name = str_alloc_copy(path_expand_ex(name));
-		file = fopen(cfg_name, "r");
+		tmp_name = path_expand_ex(name);
+		file = fopen(tmp_name, "r");
 		if (!file) {
 			if (allow_missing && errno == ENOENT) return;
-			pexit("fopen: %s", path_expand(cfg_name));
+			pexit("fopen: %s", path_expand(name));
 		}
 	}
 
 	number = 0;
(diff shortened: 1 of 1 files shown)


solardiz:
Why are we still doing `str_alloc_copy` in two places (BTW, leaking memory)? Wouldn't it be sufficient to do it on the final assignment to `cfg_name`?

claudioandre-br:
Well, what I see is:
* `str_alloc_copy()` uses `mem_alloc_tiny()` (can't be freed);
* The second buffer can be bigger than the first one; So, seems natural to have it in two places;

> Wouldn't it be sufficient to do it on the final assignment to cfg_name?

Yes. I agree that only the final assignment to cfg_name requires it.

solardiz:
1. I think we don't need to initialize `tmp_name` to `NULL` since we either don't use it or reinitialize it before first use.

2. Perhaps this line is now wrong:

```c
	if (cfg_process_line(line, ++number)) cfg_error(cfg_name, number);
```

Maybe it should be reverted to what it is in core:

```c
        if (cfg_process_line(line, ++number)) cfg_error(name, number);
```

However, I think this wouldn't properly report the filename in case `tmp_name` came from `path_expand_ex` and was successfully opened. Looks like the introduction/use of `path_expand_ex` broke previous assumptions in many subtle ways.

claudioandre-br:
> I think this wouldn't properly report the filename in case tmp_name came from path_expand_ex and was successfully opened. 

In this case, name has all information (name plus directory).

claudioandre-br:
I edited previous comment.

> Looks like the introduction/use of path_expand_ex broke previous assumptions in many subtle ways.

Yes. A "solution" could be printing `path_expand` and `path_expand_ex` file names.

claudioandre-br:
BTW, on a second thought, the correct should be:
```
if (cfg_process_line(line, ++number)) cfg_error(tmp_name, number);
```

tmp_name contains the correct and "expanded" name of the opened file.

PR #3688, opened 2019-03-26
Core merge

magnumripper:
I'm pretty certain this needs more fixes


claudioandre-br:
I tried to run git-merge myself a few days ago. It was a nightmare. 

I'm not sure how, but, the workflow have to change.
* maybe one commit at a time.


If I get it right, magnum took 2 days of work to produce a "I'm pretty certain this needs more fixes" PR.

solardiz:
Maybe I shouldn't have gone for the const'ification so close to our upcoming release, as this ended up distracting us all and especially magnum. I only added const where it was easy in core, but that didn't mean the same const wouldn't require lots of other changes in jumbo.

src/compiler.c
```diff
@@ -280,7 +280,7 @@ static int c_expect(char expected)
 }
 
 static struct c_ident *c_find_ident(struct c_ident *list,
-	struct c_ident *globals, char *name)
+	struct c_ident *globals, const char *name)
```
solardiz: I'll probably add the compiler.[ch] const's to core.

src/config.c
```diff
@@ -72,7 +72,7 @@ static int cfg_merge_local_section() {
 	if (!cfg_database) return 0;
 	if (strncmp(cfg_database->name, "local:", 6)) return 0;
 	if (!strncmp(cfg_database->name, "local:list.", 11)) return 0;
-	parent = cfg_get_section(&cfg_database->name[6], NULL);
+	parent = (struct cfg_section*)cfg_get_section(&cfg_database->name[6], NULL);
```
solardiz: Having such casts is kind of defeats the purpose of const'ification. Maybe we shouldn't be const'infying the `struct cfg_section *` pointers for now, and focus on `char *` ones only?

src/fake_salts.c
```diff
@@ -164,7 +164,7 @@ static char *LoadUserClass(char which, int i) {
 		loaded = 1;
 	}
 
-	if ((user_class = cfg_get_param("Regen_Salts_UserClasses", NULL, user_class_num))) {
+	if ((user_class = (char*)cfg_get_param("Regen_Salts_UserClasses", NULL, user_class_num))) {
 		extern char *userclass_expand(const char *src); /* found in rules.c */
 		return userclass_expand(user_class);
 	}
```
solardiz: Do we need the cast here? The context suggests that we don't.

src/inc.c
```diff
@@ -452,9 +453,9 @@ static int inc_key_loop(struct db_main *db, int length, int fixed, int count,
 
 void do_incremental_crack(struct db_main *db, char *mode)
```
solardiz: Can't we make it `const char *mode` and drop the casts?

src/x86-64.h
```diff
@@ -412,11 +345,9 @@
  * some newer CPUs capable of SSE4.2 but not AVX happen to lack SMT, so will
  * likely benefit from the 3x interleaving with no adverse effects for the
  * multi-threaded case.
- *
- * In Jumbo, we may get BF_X2 from autoconf (after testing ht cpuid flag).
```
solardiz: So we decided to no longer check HT flag in configure? I'm OK with that.

src/x86-sse.S
```diff
@@ -1,7 +1,7 @@
 /*
  * This file contains the core of a bitslice DES implementation for x86/SSE2.
  * It is part of John the Ripper password cracker,
- * Copyright (c) 2000-2001,2005,2006,2008,2011,2012,2015 by Solar Designer
+ * Copyright (c) 2000-2001,2005,2006,2008,2011,2012,2015,2019 by Solar Designer
```
solardiz: Oh, I wrongly edited this in `x86-sse.S`. Should have been in `x86.S`. I'll correct this in core.

src/x86-64.S
```diff
@@ -1,7 +1,7 @@
 /*
  * This file contains the core of a bitslice DES implementation for x86-64/SSE2.
  * It is part of John the Ripper password cracker,
- * Copyright (c) 2000-2001,2005,2006,2008,2011,2012,2015 by Solar Designer
+ * Copyright (c) 2000-2001,2005,2006,2008,2011,2012,2015,2019 by Solar Designer
```
solardiz: Maybe there should be a copyright statement for magnum's CPU detection code from 2015,2017. Ditto in x86.S.

solardiz:
Oh, CI failed:

```
now building/testing a 32 bit legacy NON-SIMD Linux build
******************************************************************************

Makefile.legacy:248: *** missing separator.  Stop.
/root/magnumripper/JohnTheRipper/.circle/CircleCI-MinGW.sh: line 164: ../run/john: No such file or directory
Exited with code 127
```

src/x86-64.h
magnumripper: I recall some discussion of that flag being unreliable at best

magnumripper:
I'll go over some of the casts again

src/x86-64.S
solardiz: BTW, the AVX2 and AVX512F detection I added to core was indeed inspired by magnum's and I tried to avoid unnecessary differences, but it is slightly different code because we're detecting fewer SIMD instruction sets in core (no SSSE3, no SSE4*). So I felt it can be copyrighted by either or both of us there. But jumbo's is pure magnum's.

src/x86-64.h
solardiz: Yes, the HT flag is "unreliable at best". I am also considering just making AVX+ builds use BF_X3 as well, based on my latest benchmarks on newer systems and with newer gcc - there are regressions in newer gcc for the BX_X2=1 case, which BF_X3 avoids. %-)

magnumripper:
> I tried to run git-merge myself a few days ago. It was a nightmare.
> 
> I'm not sure how, but, the workflow have to change.
> 
> * maybe one commit at a time.
> 
> If I get it right, magnum took 2 days of work to produce a "I'm pretty certain this needs more fixes" PR.

I should probably merge less at a time. Most of the time now was from tools' acting up though. "Meld" is a total must-have for jobs like this but under OSX right now it somehow ends up with text so tiny I can't read it. Eventually I gave up and booted Linux on my Macbook instead but then I had to start from scratch with some configuring and installing of tools, and some wifi problems, and some keyboard problems, well you get the idea. 😵

src/arm32le.h
```diff
@@ -42,27 +42,12 @@
...
 #define DES_BS_ALGORITHM_NAME		"DES 128/128 NEON"
-#elif 0
-#define DES_BS_VECTOR			5
-#define DES_BS_VECTOR_SIZE		8
-#define DES_BS_ALGORITHM_NAME		"DES 128/128 NEON + 32/32"
 #else
-#define DES_BS_VECTOR			8
-#define DES_BS_ALGORITHM_NAME		"DES 128/128 X2 NEON"
+#define DES_BS_VECTOR			2
+#define DES_BS_ALGORITHM_NAME		"DES 64/64 NEON"
```
solardiz: We should make similar edits to `arm64le.h`. That file doesn't exist in core (I should probably add it separately), so was missed by the merge.

src/config.c
magnumripper: Possibly. And we might want to consider reverting `path_expand_ex()` as well, if we keep it at all.

src/fake_salts.c
magnumripper: I misread the code. I'm reverting it, changing to `const char *user_class`.

BTW that external function declaration is local - is that OK? I recall someone mentioned it's "bad".

solardiz:
Still not quite there:

```
Makefile.legacy:573: *** missing separator.  Stop.
```

src/fake_salts.c
solardiz: I'm OK with us importing scope-local `extern` symbols for now. This can be even cleaner than having them at the top of the file, and thus unnecessary at global scope.

If we want to make things cleaner, we'd need to have all external function prototypes in header files, with no exceptions.

magnumripper:
squashed all to a single merge

claudioandre-br:
I'm canceling all unnecessary builds on CI (those related to intermediate commits/pushes)

magnumripper:
> squashed all to a single merge

I managed to do that without including the changes 🙄  Fixed now, fd28a3202 should be close to final

src/john.c
```diff
@@ -858,30 +858,30 @@ static void john_load_conf(void)
 	}
 
 	if (options.activewordlistrules == NULL)
-		if (!(options.activewordlistrules =
+	  if (!(options.activewordlistrules = (char*)
```
magnumripper: Minor whitespace problem here. I used some default emacs settings in live stick Ubuntu during merge. I'll defer a fix for now.

src/john.c
```diff
@@ -858,30 +858,30 @@ static void john_load_conf(void)
...
 		      cfg_get_param(SECTION_OPTIONS, NULL,
 		                    "BatchModeWordlistRules")))
 			options.activewordlistrules =
 				str_alloc_copy(SUBSECTION_WORDLIST);
 
 	if (options.activesinglerules == NULL)
-		if (!(options.activesinglerules =
+		if (!(options.activesinglerules = (char*)
```
magnumripper: I'm making `activesinglerules` and `activewordlistrules` `const char` in options.h and dropping these casts. This also involves adding const to `rpp_init()`

src/logger.c
```diff
@@ -364,16 +364,16 @@ void log_init(char *log_name, char *pot_name, char *session)
...
 		                                    "MarkAdminEnd"));
-		terminal_reset = parse_esc(cfg_get_param(SECTION_OPTIONS, NULL,
+		terminal_reset = parse_esc((char*)cfg_get_param(SECTION_OPTIONS, NULL,
 		                                         "TerminalReset"));
 	} else
 		admin_start = admin_end = terminal_reset = "";
 
-	admin_string = parse_esc(cfg_get_param(SECTION_OPTIONS, NULL,
+	admin_string = parse_esc((char*)cfg_get_param(SECTION_OPTIONS, NULL,
 	                                       "MarkAdminString"));
```
magnumripper: I'm fixing all of the above in a separate commit, changing parse_esc to not alter string in-place.

src/opencl_sha256crypt_fmt_plug.c
```diff
@@ -313,11 +313,11 @@ static void build_kernel(char *task, char *custom_opts)
 				opt[i] = '_';
 
 		if (!(custom_opts = getenv(opt)))
-			custom_opts = cfg_get_param(SECTION_OPTIONS,
+		  custom_opts = (char*)cfg_get_param(SECTION_OPTIONS,
```
magnumripper: Whitespace, defering fix until bots are ready

src/logger.c
magnumripper: Scrap "separate commit" - I'm squashing it, just later.

src/pp.c
```diff
@@ -1517,7 +1517,7 @@ void do_prince_crack(struct db_main *db, char *wordlist, int rules)
   if (!john_main_process)
     warn = 0;
 
-  wordlist = path_expand(wordlist);
+  wordlist = (char*)path_expand(wordlist);
```
magnumripper: The above will be fixed soon, changing to `do_prince_crack(..., const char *wordlist, ...)`

magnumripper:
Merged manually and pushed as 843638b.

PR #3690, opened 2019-03-26
OpenCL: add const qualifier

claudioandre-br:
I see no warnings on well.


magnumripper:
I suppose a rebase is needed after #3692 is merged.

Issue #3620, opened 2019-01-28
OpenCL autotune: add a backward run to lower GWS in case the device wasn't warmed-up


solardiz:
### Revised issue description ###

I originally wrote this in one of the comments below, but I think it fits to replace the issue description:

Rather than have a separate longer warm-up run, how about we revise the final GWS-finding loop to also go backwards after it's chosen a tentative GWS, to see if a lower GWS has started to perform better (or less than 1% worse), which may happen if the GPU was still waking up during the forward run? With this kernel, it'd enable us to end up with a GWS of 10240 instead of 327680, even if with suboptimal LWS.

This backward run should stop as soon as it finds a lower GWS that performs more than 1% worse than the previously tested higher one, so when it's unneeded it'd stop after the first try of one lower GWS, meaning it won't significantly increase the auto-tuning duration.

### Original issue description ###

I think this is primarily an example of how our LWS/GWS autotuning fails, which might be applicable to other formats as well. There's probably some general issue causing this, which we'd want to identify and fix.

Default run:

```
[solar@super run]$ ./john -test -form=md5crypt-opencl -dev=3 -v=5
initUnicode(UNICODE, ASCII/ASCII)
ASCII -> ASCII -> ASCII
Device 3: GeForce GTX 1080
Benchmarking: md5crypt-opencl, crypt(3) $1$ [MD5 OpenCL]... Loaded 61 hashes with 39 different salts to test db from test vectors
Options used: -I /home/solar/j/bleeding-jumbo-20190120/run/kernels -cl-mad-enable -DSM_MAJOR=6 -DSM_MINOR=1 -cl-nv-verbose -D__GPU__ -DDEVICE_INFO=524306 -DSIZEOF_SIZE_T=8 -DDEV_VER_MAJOR=410 -DDEV_VER_MINOR=79 -D_OPENCL_COMPILER -DPLAINTEXT_LENGTH=15 $JOHN/kernels/cryptmd5_kernel.cl
Calculating best GWS for LWS=32; max. 100ms single kernel invocation.
Raw speed figures including buffer transfers:
pass xfer: 4.064us, crypt: 13.228ms, result xfer: 5.120us
gws:      1024	  77357c/s    77357000 rounds/s  13.237ms per crypt_all()!
pass xfer: 12.448us, crypt: 13.106ms, result xfer: 9.984us
gws:      2048	 155995c/s   155995000 rounds/s  13.128ms per crypt_all()!
[...]

PR #3695, opened 2019-03-27
OpenCL auto-tune tweaks. Add backward GWS runs.

magnumripper:
#3620

### DIFF
--- a/src/opencl_common.c
+++ b/src/opencl_common.c
@@ -93,6 +93,7 @@ static char* get_device_capability(int sequential_id);
 // Used by auto-tuning to decide how GWS should changed between trials.
 extern int autotune_get_next_gws_size(size_t num, int step, int startup,
                                       int default_value);
+extern int autotune_get_prev_gws_size(size_t num, int step);
 
 // Settings to use for auto-tuning.
 static int buffer_size;
@@ -1499,7 +1500,8 @@ static void clear_profiling_events()
 	}
 }
 
-// Do the proper test using different global work sizes.
+// Do a test run with a specific global work size, return total duration
+// (or return zero for error or liomits exceeded)
 static cl_ulong gws_test(size_t gws, unsigned int rounds, int sequential_id)
 {
 	cl_ulong startTime, endTime, runtime = 0, looptime = 0;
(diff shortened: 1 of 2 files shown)


src/opencl_common.c
```diff
@@ -1819,14 +1821,13 @@ void opencl_find_best_lws(size_t group_size_limit, int sequential_id,
...
-	numloops = (int)(size_t)(200000000ULL / (endTime - startTime));
+	cl_ulong roundup = endTime - startTime - 1;
+	numloops = (int)(size_t)((200000000ULL + roundup) / (endTime - startTime));
 
 	clear_profiling_events();
 
 	if (numloops < 1)
 		numloops = 1;
-	else if (numloops > 5)
-		numloops = 5;
```
magnumripper: I dropped this, I'm not sure what that limitation buys us (we're talking 200ms here).

src/opencl_common.c
```diff
@@ -2056,9 +2057,6 @@ void opencl_find_best_gws(int step, int max_duration,
 		raw_speed = (kpc / (run_time / 1E9)) * mask_int_cand.num_int_cand;
 		speed = rounds * raw_speed;
 
-		if (run_time < min_time)
-			min_time = run_time;
-
```
magnumripper: These were artifacts, never used

src/opencl_common.c
claudioandre-br: It is acceptable. But, for very fast formats, from now on, you can run unnecessary tests (when 5 is enough).

src/opencl_common.c
```diff
@@ -1499,7 +1500,8 @@ static void clear_profiling_events()
 	}
 }
 
-// Do the proper test using different global work sizes.
+// Do a test run with a specific global work size, return total duration
+// (or return zero for error or liomits exceeded)
```
claudioandre-br: ?

magnumripper: That old comment would perhaps be proper for `opencl_find_best_gws()`. However, this comment is for `gws_test()` which tests a single GWS as (now) described. `opencl_find_best_gws()` calls `gws_test()` for each GWS.

magnumripper: ...and liomits is my stiff upper lip pronunciation of limits 🤣

src/opencl_common.c
magnumripper: For very fast formats, we just get more samples (and incidentally we increase the chance of winding up the GPU to full speed).

Issue #3696, opened 2019-03-27
Bring x86-64.h and x86-sse.h closer to core's


solardiz:
Especially after the recent core update and merge, jumbo's x86-64.h and x86-sse.h are significantly different from core's in unnecessary ways.

There are some real bugs, too - e.g., `DES_BS` is never set to 4 (so `sboxes-t.c` is never used - but it should be on AVX-512), x86-sse.h tries to use 256-bit XOP (which it calls "XOP2") whereas DES_bs_b.c no longer includes support for that, there are redefinitions of `CPU_FALLBACK_BINARY` on some builds (are likely to result in build-time warnings). These are just things I noticed when skimming the files. There are probably many more issues.

Of course, trying to bring these files closer to core's is a change relative to what we had tested in jumbo, and might introduce new bugs too, but at this point I think we should go for it.

solardiz:
Specifically, I suggest to start with versions of these files from core and then re-introduce new implementations (consistent with core's) of whatever changes are actually needed in jumbo (such as for the extra SSE* levels). This might also require minor edits to the .S files - such as for `#if` vs. `#ifdef` for `CPU_REQ_*` - bringing them closer to core's, too.

As an option, this effort can be split into two steps: first fix two major bugs I mentioned above, then work on bringing the files closer to core's.

magnumripper:
Making it closer to core's isn't very straightforward because that file is so DES-centric. We need CPU_DETECT, CPU_REQ, CPU_REQ_XX, CPU_NAME and CPU_FALLBACK_BINARY_DEFAULT for other things as well. Having said that, I'm reviewing them.

solardiz:
I don't see a problem with defining the `CPU_*` things inside the DES-centric if/elif/else/endif's, yet using those SIMD instruction sets in non-DES formats as well. This might be unclean, but it will help bring the files in core and jumbo closer and avoid duplicate if/elif/else/endif chains (or equivalent) for DES_* and CPU_*.

Well, the SSE* flavors are the same thing for DES, so perhaps there might be an inner if/elif/else/endif chain for those inside the SSE2 block for DES.

magnumripper:
> I don't see a problem with defining the CPU_* things inside the DES-centric if/elif/else/endif's

They will all be turned off for non-gcc AVX+ builds, resulting in SSE2...

solardiz:
> They will all be turned off for non-gcc AVX+ builds, resulting in SSE2...

[...]

PR #3700, opened 2019-03-27
Revision of x86-sse.h and x86-64.h after core merge. #3696

magnumripper:
These are hard to review here. A better way is eg. `git diff origin/master..origin/botcheck x86-64.h` to compare the core files against the Jumbo ones.


magnumripper:
I haven't tested 32-bit at all, I'm not sure I can find a machine for that. Build bots to the rescue though!

solardiz:
> I haven't tested 32-bit at all, I'm not sure I can find a machine for that.

On a 64-bit system, you can probably do `setarch i686 bash` and then build from that shell. You'll need to have 32-bit libraries installed. I think we have such libraries installed at least on "well".

solardiz:
1. These blocks
```
#if CPU_FALLBACK && !defined(CPU_FALLBACK_BINARY)
#define CPU_FALLBACK_BINARY            "john-non-avx512bw"
#define CPU_FALLBACK_BINARY_DEFAULT
#endif
```
Can probably be changed to one-liners like:
```
#define CPU_FALLBACK_BINARY_DEFAULT    "john-non-avx512bw"
```
like I do in core, followed by this shared block I added to core recently (is it somehow missing from master branch here?!):
```
#ifndef CPU_FALLBACK
#define CPU_FALLBACK                    0
#endif
#if !defined(CPU_FALLBACK_BINARY) && defined(CPU_FALLBACK_BINARY_DEFAULT)
#define CPU_FALLBACK_BINARY             CPU_FALLBACK_BINARY_DEFAULT
#endif
```

2. The added `#ifndef BF_X2` ... `#endif` can be dropped - I think they're remnants of jumbo's checking of the HT flag in configure, now dropped. Also, x86-sse.h has this logic in only one of two places, so it's broken anyway.

3. Minor bug: x86-64.h forgets to set `CPU_DETECT` and `CPU_REQ` to 0 when these things are not needed. Perhaps it should.

4. x86-sse.h's hack of `#define __SSE2__ 1` for MSVC without OpenMP (why?) looks very dirty and might be problematic (if MSVC either defines this on its own or would refuse to recognize the intrinsics when SSE2 isn't properly enabled in the compiler - at least we'd have these issues if we tried that hack with gcc, so I'd expect similar from MSVC). It's not a new thing with this PR, though.

magnumripper:
I found some really old Thinkpad (32-bit. only SSE2 and single core) in the garage and everything seemed fine there.

magnumripper:
> The added #ifndef BF_X2 ... #endif can be dropped - I think they're remnants of jumbo's checking of the HT flag in configure, now dropped. Also, x86-sse.h has this logic in only one of two places, so it's broken anyway.

I'll look into this tomorrow. I'll merge this for now.

The `#ifndef BF_X2` is (I think) for being able to do things like `./configure BF_X2=3 (...)`

PR #3704, opened 2019-03-28
OpenCL: mute a warning on Windows

claudioandre-br:
As seeing in https://github.com/magnumripper/JohnTheRipper/blob/bleeding-jumbo/src/opencl_common.c#L1219

### DIFF
--- a/src/opencl_common.c
+++ b/src/opencl_common.c
@@ -1223,7 +1223,7 @@ static char *include_source(const char *pathname, int sequential_id, const char 
 
 	MEM_FREE(pex);
 #else
-	full_path = path_expand_safe(pathname);
+	full_path = (char*)path_expand_safe(pathname);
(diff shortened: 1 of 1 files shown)


solardiz:
Shouldn't we const'ify this variable and argument of `quote_str` instead?

claudioandre-br:
> Shouldn't we const'ify this variable and argument of quote_str instead?

Then, it will complain 4-5 times. I can't remember all, at least in `MEM_FREE(full_path);` 

This patch keeps the stuff as it is now.

solardiz:
Hmm, I understand `MEM_FREE` requires that the pointer itself not be `const`, but it should accept a pointer to a `const` string.

magnumripper:
I saw the wierd MEM_FREE problem too. Haven't had time to investigate what the heck happens.

PR #3732, opened 2019-03-30
mask: free resources only after use

claudioandre-br:


solardiz:
Per @Apingis findings, we have a regression here where the final status line printed no longer corresponds to actually final status. Perhaps we should come up with a better fix for the original issue fixed here, a fix that wouldn't have that side-effect.

claudioandre-br:
The problem is that there is no better fix. We need to create a new method, e.g.,
- `mask_unload()`
- `mask_final()`

And run the `free()` only inside this method. And, this new method will be the default for all final cleanups from now on.

solardiz:
> The problem is that there is no better fix.

That's bad news. Maybe we can instead accept a memory leak? Maybe we won't be leaking too much (just mask status?) and nor too many times (at most as many times as we increase mask length)?

Issue #3721, opened 2019-03-29
some -ztex formats cannot handle empty salt


AlekseyCherepanov:
I tried md5crypt, sha256crypt and sha512crypt with empty salts and -ztex formats fail with error.

After error, same format is loaded well, but the error happens again. Switching format fails. But rerunning fixes it.

Aborting with Control-C causes john to print `Wait...`, but then error message is printed again. Next Control-C exits immediately.

@Apingis, can drupal7 have empty salt?

@solardiz, can phpass or bcrypt have empty salt?

```
$ cat pw-empty-salt
$1$$RmyPVMlhpXjJj8iv4w.Ul.
$5$$ak6faDK9fe20oKsZ7bjHwjN3gWFkUl7Jh1kXkAJFoo3
$6$$.Q5TOqNNyRzKTE5ZY02dsLIOQwkwN0bWooBHB4loBoSN7E68SH7dHJ6SZRyzHvdF0/ybpEO82nrjBd5oI1f3g0

$ john pw-empty-salt --format=md5crypt-ztex
[...]
SN [...] FPGA #1 error: pkt_comm_status=0x80, debug=0x0000
SN [...] error -1 doing r/w of FPGAs (LIBUSB_ERROR_IO)
Found 1 device(s) ZTEX 1.15y
SN: [...] productId: 10.15.0.0 "inouttraffic JtR 1.8.x" busnum:4 devnum:4 
SN [...] FPGA #1 error: pkt_comm_status=0x80, debug=0x0000
SN [...] error -1 doing r/w of FPGAs (LIBUSB_ERROR_IO)
^C
Session aborted

$ john pw-empty-salt --format=sha256crypt-ztex
SN [...]: uploading bitstreams.. ok
SN [...]: device_list_check_bitstreams(): no bitstream or wrong type
no valid ZTEX devices found

$ john pw-empty-salt --format=sha256crypt-ztex
[...]
SN [...] FPGA #1 error: pkt_comm_status=0x80, debug=0x0000
SN [...] error -1 doing r/w of FPGAs (LIBUSB_ERROR_IO)

[...]
$ john pw-empty-salt --format=sha512crypt-ztex
SN [...] FPGA #1 error: pkt_comm_status=0x80, debug=0x0000
SN [...] error -1 doing r/w of FPGAs (LIBUSB_ERROR_IO)

$ john pw-empty-salt --format=md5crypt
[...]
123456           (?)
[...]
[...]

PR #3741, opened 2019-03-31
{sha256|sha512|md5}crypt-ztex: salt_length=0: print warning, don't process

Apingis:
- Discussed in #3721 
- If password file contains only hashes with zero salt length,  John reverts to CPU version.
- After the fix included in the next hardware design versions, the warning will be removed.


Apingis:
Will create more "clean" PR

PR #3742, opened 2019-03-31
{sha256|sha512|md5}crypt-ztex: salt_len=0: warning, don't process

Apingis:
- Discussed in #3721
- If password file contains only hashes with zero salt length, John reverts to CPU version.
- After the fix included in the next hardware design versions, the warning will be removed.

### DIFF
--- a/src/ztex_md5crypt.c
+++ b/src/ztex_md5crypt.c
@@ -105,6 +105,37 @@ static struct fmt_tests tests[] = {
 };
 
 
+static int cryptmd5_common_valid_salt0(char *ciphertext, struct fmt_main *self)
+{
+	char *pos, *start;
+	char *salt_pos, *salt_end_pos;
+
+	if (!strncmp(ciphertext, md5_salt_prefix, md5_salt_prefix_len))
+		ciphertext += md5_salt_prefix_len;
+	else if (!strncmp(ciphertext, apr1_salt_prefix, apr1_salt_prefix_len))
+		ciphertext += apr1_salt_prefix_len;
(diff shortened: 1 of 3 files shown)


src/ztex_md5crypt.c
```diff
@@ -105,6 +105,37 @@ static struct fmt_tests tests[] = {
...
+{
+	char *pos, *start;
+	char *salt_pos, *salt_end_pos;
+
+	if (!strncmp(ciphertext, md5_salt_prefix, md5_salt_prefix_len))
+		ciphertext += md5_salt_prefix_len;
+	else if (!strncmp(ciphertext, apr1_salt_prefix, apr1_salt_prefix_len))
+		ciphertext += apr1_salt_prefix_len;
+	else if (!strncmp(ciphertext, smd5_salt_prefix, smd5_salt_prefix_len))
+		ciphertext += smd5_salt_prefix_len;
```
solardiz: BTW, we don't have any test vectors for apr1 and smd5 in this format. If we support those hash types in this format, then we should add such test vectors. We should add them _before_ possibly revising the hardware design to support zero-length salts, because adding the test vectors might uncover more issues and shortcomings to fix at the same time.

solardiz:
BTW, maybe a simpler implementation of this temporary workaround would have been to have the new `valid` functions call the previous/generic ones and if they return non-zero then post-check for e.g. `!strncmp(..., "$1$$", 4)`, etc. and print a warning and force zero return if so. This would avoid needing to duplicate their code, and would be easier to review now and to revert later.

PR #3745, opened 2019-04-01
Make the windows target the default when running on Windows

claudioandre-br:
Both commits are needed.

### DIFF
--- a/src/Makefile.in
+++ b/src/Makefile.in
@@ -34,8 +34,8 @@ FIND = @FIND@
 NULL = /dev/null
 SORT = @SORT@
 STRIP = @STRIP@
-CPPCHECK=cppcheck
-PEFLAGS=peflags --dynamicbase=true --nxcompat=true
+CPPCHECK = cppcheck
+PEFLAGS = peflags --dynamicbase=true --nxcompat=true
 
 SHELL = /bin/sh
 VPATH = @srcdir@
@@ -192,9 +192,8 @@ default:
 		JOHN_OBJS="$(JOHN_OBJS) @CC_ASM_OBJS@" \
 		AESNI_ARCH=@AESNI_ARCH@
 
-	@if [ "$(OS)" = "Windows_NT" ]; then\
-		$(MAKE) peflags;\
-		$(MAKE) windows;\
+	@if [ "$(OS)" = "Windows_NT" ]; then \
+		$(MAKE) peflags; \
 	fi
 	@echo
 	@echo Make process completed.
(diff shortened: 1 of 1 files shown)


claudioandre-br:
> Claudio, I've just added a commit to here - I hope you don't mind? I think we shouldn't do the sed's and mv's by default because they alter files/names that came from our pristine source tree. I think such behavior of a mere "make" would be unexpected. So I only kept them in the "windows-package" target.

Windows users have to raise their voices here. I don't do this myself while packaging, but in IMO, it makes sense to do.

> Claudio, I've just added a commit to here - I hope you don't mind?

* See #3747
  * remove cppcheck
  * create build blocks

solardiz:
I actually think it does make sense to do the sed's and mv's _while_packaging_. But not necessarily when doing e.g. test (re)builds and intending to send a PR off a Windows system, which I think e.g. @Apingis and @jfoug do. We shouldn't be changing our pristine files on a mere build.

I don't see value in the "create build blocks" part of #3747.

So I'd rather have this PR merged as it currently is, and then have cppcheck removed with another PR - or we can add that removal to here.

solardiz:
Actually, added dropping of cppcheck to here. Not really relevant, but I don't mind having it here as opposed to in another PR.

claudioandre-br:
> But not necessarily when doing e.g. test (re)builds and intending to send a PR off a Windows system.

Without the new target `make -s do-build`, all (re)builds on Windows will change the EXE (it is mandatory). How do I build a regular EXE on Windows if I need one?

solardiz:
> Without the new target `make -s do-build`, all (re)builds on Windows will change the EXE (it is mandatory). How do I build a regular EXE on Windows if I need one?

Why would one possibly need "a regular EXE", and how would that person know that? I think this is an advanced enough case that we don't need to provide an (undocumented, internal) make target for it. Whoever might need this would just as easily e.g. `make PEFLAGS='echo --'` or comment out the `peflags` invocation, or would not know about our different make target anyway.

claudioandre-br:
> Why would one possibly need "a regular EXE", and how would that person know that?

I need because I can't read dozens of peflags comments inside the build's CI log. Anyway, I already created a patch to revert this on(my) CI.

solardiz:
Oh, I didn't know this produced "dozens of peflags warnings". Can you show those? Why are they produced?

claudioandre-br:
Total = warnings * 2_OMP * 5_SIMD (from now on, in red, thanks to PowerShell)
 
```
Warning: setting dynbase on file with no relocation info (base64conv.exe).
Warning: setting dynbase on file with no relocation info (bitlocker2john.exe).
Warning: setting dynbase on file with no relocation info (calc_stat.exe).
Warning: setting dynbase on file with no relocation info (cprepair.exe).
Warning: setting dynbase on file with no relocation info (dmg2john.exe).
Warning: setting dynbase on file with no relocation info (genmkvpwd.exe).
Warning: setting dynbase on file with no relocation info (gpg2john.exe).
Warning: setting dynbase on file with no relocation info (hccap2john.exe).
Warning: setting dynbase on file with no relocation info (john.exe).
Warning: setting dynbase on file with no relocation info (keepass2john.exe).
Warning: setting dynbase on file with no relocation info (mkvcalcproba.exe).
Warning: setting dynbase on file with no relocation info (putty2john.exe).
Warning: setting dynbase on file with no relocation info (racf2john.exe).
Warning: setting dynbase on file with no relocation info (rar2john.exe).
Warning: setting dynbase on file with no relocation info (raw2dyna.exe).
Warning: setting dynbase on file with no relocation info (tgtsnarf.exe).
Warning: setting dynbase on file with no relocation info (uaf2john.exe).
Warning: setting dynbase on file with no relocation info (unafs.exe).
Warning: setting dynbase on file with no relocation info (undrop.exe).
Warning: setting dynbase on file with no relocation info (unique.exe).
Warning: setting dynbase on file with no relocation info (unshadow.exe).
Warning: setting dynbase on file with no relocation info (wpapcap2john.exe).
Warning: setting dynbase on file with no relocation info (zip2john.exe).
``

solardiz:
Oh. IIRC, those warnings were harmless in this case. I don't know if we want to suppress them or not.

PR #3746, opened 2019-04-01
Minor fixes on how to build on Windows

claudioandre-br:


doc/INSTALL-WINDOWS
```diff
@@ -39,8 +39,7 @@ system.
       cd c:\Users\YourName\Desktop\john\src
 
     - Build
-      ./configure && make -s clean && make -sj4
-      make windows
+      configure && make -s clean && make -sj4
```
magnumripper: `.` is in PATH in default cygwin?

solardiz: Since we require running this in Cygwin shell, I think it's better to keep the "./" - it won't hurt, but it and the habit might help on other occasions.

PR #3754, opened 2019-04-05
Various updates to install instructions.

magnumripper:


doc/INSTALL-WINDOWS
```diff
@@ -29,14 +29,14 @@ system.
...
-    Unzip it to your preferred directory   
+    Unzip it to your preferred directory
 
 ==== Build
 
     - Inside a CygWin terminal
 
     - Go to the folder you unzipped John the Ripper source code. E.g.:
-      cd c:\Users\YourName\Desktop\john\src
+      cd c:\Users\YourName\Desktop\JohnTheRipper\src
```
magnumripper: This is correct, right?

doc/README-DISTROS
```diff
@@ -1,50 +1,57 @@
...
-	mv ../run/john ../run/john-non-omp &&
-	./configure --disable-native-tests CPPFLAGS='-DJOHN_SYSTEMWIDE -DJOHN_SYSTEMWIDE_EXEC="\"/usr/local/bin\"" -DJOHN_SYSTEMWIDE_HOME="\"/usr/local/share/john\"" -mavx2 -DOMP_FALLBACK -DOMP_FALLBACK_BINARY="\"john-non-omp\"" -DCPU_FALLBACK -DCPU_FALLBACK_BINARY="\"john-xop\""' &&
+	mv ../run/john ../run/john-avx2-non-omp &&
+	./configure --disable-native-tests --enable-simd=avx2 CPPFLAGS='-DJOHN_SYSTEMWIDE -DJOHN_SYSTEMWIDE_EXEC="\"/usr/local/bin\"" -DJOHN_SYSTEMWIDE_HOME="\"/usr/local/share/john\"" -DOMP_FALLBACK -DOMP_FALLBACK_BINARY="\"john-avx2-non-omp\"" -DCPU_FALLBACK -DCPU_FALLBACK_BINARY="\"john-xop\""' &&
+	make -s clean && make -sj4 strip &&
+	mv ../run/john ../run/john-avx2 &&
+	./configure --disable-native-tests --enable-simd=avx512bw --disable-openmp &&
+	make -s clean && make -sj4 strip &&
+	mv ../run/john ../run/john-avx512-non-omp &&
+	./configure --disable-native-tests --enable-simd=avx512bw CPPFLAGS='-DJOHN_SYSTEMWIDE -DJOHN_SYSTEMWIDE_EXEC="\"/usr/local/bin\"" -DJOHN_SYSTEMWIDE_HOME="\"/usr/local/share/john\"" -DOMP_FALLBACK -DOMP_FALLBACK_BINARY="\"john-non-omp\"" -DCPU_FALLBACK -DCPU_FALLBACK_BINARY="\"john-avx2\""' &&
```
magnumripper: Should we add a binary for AVX-512F w/o AVX-512BW? Are there things like that (except the old MIC)?

solardiz: > Should we add a binary for AVX-512F w/o AVX-512BW?

Maybe, because:

> Are there things like that

Yes: second generation Xeon Phi aka Knights Landing. Our `./configure` correctly detected AVX512F without AVX512BW there when I ran that a week ago, and the resulting binary passed all self-tests.

Second generation Xeon Phi aka Knights Landing is usually the host CPU, so building JtR for it is easy (no need to cross-compile, and already have the usual Linux distro's libraries). Coprocessor/accelerator versions also exist, but I think plans to release them to the open market were cancelled.

> (except the old MIC)?

No, "old MIC" aka first generation Xeon Phi aka Knights Corner only has its own MIC instructions/intrinsics, which are similar to AVX512F at source code level, but are different at binary level and thus this "CPU" is correctly not detected as having AVX512F. For JtR core, I cross-compile from host using `make linux-mic`. For jumbo, I don't know whether/what way we support now for building for MIC - it's tricky since first one needs to cross-compile OpenSSL. @lei-april did that in 2015, but I'm afraid we haven't repeated it since then. Ideally, we'd have a documentation file describing the procedure, and re-test it before a release.

magnumripper: I'll add AVX-512F to the chain.

magnumripper: Fixed. This needs testing!

PR #3759, opened 2019-04-05
Do not disable self-tests when running "-test -mask"

magnumripper:


magnumripper:
Branch restored, this will be reverted for now.

PR #3765, opened 2019-04-07
relbench fix and benchmark-unify adjustments

frank-dittrich:
There are many changes between 1.8.0-jumbo-1 and bleeding-jumbo which cannot be mapped, e.g. changes from Raw to many salts / Only one salt and vice versa.

That also means, some of the performance regressions will not be seen.

### DIFF
--- a/run/benchmark-unify
+++ b/run/benchmark-unify
@@ -101,7 +101,9 @@ while (<>) {
 # (ignoring case, so use sort -f or sort --ignore-case, please)
 __DATA__
 bf-opencl, OpenBSD Blowfish (x32)	bcrypt-opencl ("$2a$05", 32 iterations)
+blackberry-es10	Blackberry-ES10 (101x)
 BSDI DES (x725)	bsdicrypt, BSDI crypt(3) ("_J9..", 725 iterations)
+chap, iSCSI CHAP authentication	chap, iSCSI CHAP authentication / EAP-MD5
 Cisco PIX MD5	pix-md5, Cisco PIX
 CRC-32	CRC32
 crypt-MD5	md5crypt, crypt(3) $1$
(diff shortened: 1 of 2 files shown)


magnumripper:
Perhaps I should change "PASS " and "\b\b\b\b\bDONE" to just "PASS" and ", DONE", please allow that as well

Output would be
```
Benchmarking: sha1crypt-opencl, (NetBSD) [PBKDF1-SHA1 OpenCL]... PASS, DONE 
Speed for cost 1 (iteration count) of 64000 and 40000
Raw:	10304 c/s real, 819200 c/s virtual
```
or, when using --skip
```
Benchmarking: sha1crypt-opencl, (NetBSD) [PBKDF1-SHA1 OpenCL]... SKIP, DONE 
Speed for cost 1 (iteration count) of 64000 and 40000
Raw:	10304 c/s real, 819200 c/s virtual
```

claudioandre-br:
> Perhaps I should change "PASS " and "\b\b\b\b\bDONE" to just "PASS" and ", DONE"

+1

magnumripper:
I'll revert to the old way of printing either PASS or DONE for now. But we'll likely change to "PASS" and then ", DONE" once we revisit the non-mask selftest, mask-mode benchmark stuff later.

frank-dittrich:
@magnumripper 
That would be either
```
($name) = /^\t?Benchmarking: ([^\[]+) \[.*\].* (PASS, |.....\b\b\b\b\b)?(DONE|Warning:.*)$/;
```
or
```
($name) = /^\t?Benchmarking: ([^\[]+) \[.*\].* (PASS, )?(DONE|Warning:.*)$/;
```
(both untested)

Unfortunately, right now I don't even have time to commit such a tiny change.

magnumripper:
I committed a variant right in here. Will test it soon-ish

PR #3768, opened 2019-04-07
OpenCL: SHA256 and SHA512 force tune after fmt_done()

claudioandre-br:


magnumripper:
@solardiz @claudioandre-br we have this problem (and more) in far to many formats. We probably have to revert 7187208d and wait for J2 😢

solardiz:
@magnumripper OK, if we have to. Please have `-test -mask` print a warning that it doesn't actually run self-tests, then. Thanks!

claudioandre-br:
It is up to you. I will support you.

Issue #3436, opened 2018-10-17
Is SIMD code really better on non-x86?


claudioandre-br:
This is something I've seen for a long time. Since I don't have the hardware I can't create a proper issue. Anyway, I've been watching this for a long time.

Applies to ARM and PPC.

```
Testing: sha512crypt, crypt(3) $6$ (rounds=5000) [SHA512 128/128 NEON 2x]... (4xOMP) Loaded 6 hashes with 4 different salts to test db from test vectors
OMP scale 4: 32 crypts (1x32) in 0.102284 seconds, 312 c/s +
Autotuned OMP scale 4, preset is 2
```

```
Testing: sha512crypt, crypt(3) $6$ (rounds=5000) [SHA512 32/32 OpenSSL]... (4xOMP) Loaded 6 hashes with 4 different salts to test db from test vectors
OMP scale 4: 16 crypts (1x16) in 0.030709 seconds, 521 c/s +
OMP scale 16: 64 crypts (1x64) in 0.120682 seconds, 530 c/s
Autotuned OMP scale 8, preset is 2
```

claudioandre-br:
One that is yelling for attention is sha256crypt. It seems to hang all the time.

magnumripper:
Our SIMD code is architecture agnostic. A certain SIMD implementation (eg. ARM NEON) may well be slower than non-SIMD for certain operations (eg. 64-bit stuff). If we can safely identify such cases we could possibly disable SIMD for certain formats on certain hardware.

Also, we probably have no guarantee there will be as many SIMD units as there are CPU cores - in that case non-OMP may show a gain with SIMD while eg. 4xOMP will not.

I'm sure @solardiz can comment on this with much more actual knowledge.

And yes, I've been meaning to look into what slows sha256crypt down so much when auto-tuning OMP. It's likely a bug.

magnumripper:
BTW I've had a hard time benchmarking ARM on Raspberry Pi reliably, because it may throttle its pacemaker eg. due to temperature or (even with a decent power supply) due to voltage dips. I plan to fix that with fans and stuff but that's for a rainy day in distant future.

claudioandre-br:
I can't say much about the hardware, but, I guess we are running inside https://certification.ubuntu.com/hardware/201608-25047/

magnumripper:
[...]

Issue #2313, opened 2016-10-29
Add 64 bit NEON support


jfoug:
I have a raspberry pi3, which is quad-core 64 NEON.   I will work at getting full support added for this chip.

magnumripper:
Cool. I recall @lei-zz did some work with NEON in GSoC. Was that 32-bit only?

jfoug:
All we have right now is 32 bit neon.  I am not sure the neon code is any better or not (note, there appears to be many flavors, I am just getting started).  But the pi3 IS a 64 bit CPU (first one for the pi family).  So even if there are not 'better' neon instructions, the 64 bit should be used, if possible.  NOTE, I am still just getting started with the OS.  Lots to learn.

claudioandre:
ICYMI; the neon test (seen below) used in configure.ac succeeds on aarch64 (__ARM_64BIT_STATE - 64-bit targets only).

```
#include <stdio.h>
	extern void exit(int);
	int main(){uint32x4_t t;*((long*)&t)=1;t=veorq_u32(t,t);if((*(unsigned*)&t)==88)printf(".");exit(0);}
```

jfoug:
The Rpi3 is 32 bit, and NEON worked fine.  It is a 64 bit CPU, but there is no 64 bit OS built for it. There are sketchy instructions on how to build and bootstrap 64 bit yourself, but that was over my head.  So NEON is working fine on 32 bit.

claudioandre:
I was not talking about rpi. Anyway, I can find people that disagree with you.
* https://fedoraproject.org/wiki/Architectures/ARM/Raspberry_Pi#Raspberry_Pi_3_aarch64_support

claudioandre:
Well, I posted https://github.com/magnumripper/JohnTheRipper/issues/2313#issuecomment-351833398 because I was able to "(kind of)" do it. I don't have any hardware, I only can access a build server [1]. So, it is no more than a POC.
* https://github.com/magnumripper/JohnTheRipper/compare/bleeding-jumbo...claudioandre:neon-64

```
Created directory: /root/.john
Version: 1.8.1-jumbo-1-430-g8604ae9+
Build: linux-gnu 64-bit NEON-ac OMP
SIMD: NEON, interleaving: MD4:2 MD5:2 SHA1:1 SHA256:1 SHA512:1
```

```
Will run 4 OpenMP threads
[...]

PR #3772, opened 2019-04-08
Core merge

magnumripper:


src/arm64le.h
```diff
@@ -36,32 +36,40 @@
...
 #define DES_X2				0
-#define DES_MASK			0
+#define DES_MASK			1
 #define DES_SCALE			1
 #define DES_EXTB			0
-#define DES_COPY			1
+#define DES_COPY			0
 #define DES_BS_ASM			0
-#define DES_BS				1
 #if __ARM_NEON && !JOHN_NO_SIMD
```
solardiz: What would define `__ARM_NEON` (checked here and in one place below)? It is _not_ defined by the compiler on AArch64, at least not in my tests. Instead, we could check for `__aarch64__`, but there's no need to do that in this file since our usage of this file implies we're on that arch.

src/arm64le.h
```diff
@@ -36,32 +36,40 @@
...
-#define MD5_IMM				0
+#define MD5_IMM				1
 
 #define BF_ASM				0
-#define BF_SCALE			1
-#define BF_X2				0
+#define BF_SCALE			0
+#define BF_X2				1
 
 #if __ARM_NEON && !JOHN_NO_SIMD
```
solardiz: The other wrong(?) check for `__ARM_NEON`.

magnumripper:
OK so anywhere we print NEON we should instead say ASIMD?

I have no idea where the `__ARM_NEON` check originated. I'll just drop it.

magnumripper:
> OK so anywhere we print NEON we should instead say ASIMD?

Or is this just for aarch64?

magnumripper:
Yes, looking at core that appears to be the case

solardiz:
Right, we should only assume ASIMD is always available and call it that on AArch64. On 32-bit ARM, it's still NEON and it isn't always available - we need to check for it like we do.

magnumripper:
I'm pretty sure `__ARM_NEON` used to be defined on aarch64 builds. Otherwise `pseudo_intrinsics.h` would fail defining virtual intrinsics, for a starter. I'll blindly fix this but I'm not sure I'll have the time to test it. @claudioandre-br do you have some means of testing that?

magnumripper:
...but we still include `#include <arm_neon.h>`, right? You don't seem to do so in core. I'm confused by all this. Jumbo did run fine using SIMD on aarch64 a year ago or so.

solardiz:
Hmm, actually that's what core has in DES_bs_b.c:

```c
#if defined(__ARM_NEON) && DES_BS_DEPTH == 128
#include <arm_neon.h>
```

I don't know where `__ARM_NEON` comes from. I'll do more testing.

magnumripper:
#2313 has but a few bits of info

solardiz:
Uh-oh, looks like we didn't actually enable NEON on Aarch64 for bitslice DES, precisely because this macro wasn't defined. But enabling it hurts performance now that I try. %-(

magnumripper:
> Jumbo did run fine using SIMD on aarch64 a year ago or so.

That was correct +/- 12 hours 😎

solardiz:
I now doubt we _ever_ tested ASIMD for bitslice DES because of this issue. But I am really puzzled what happened with our testing of formats that use pseudo-intrinsics.

magnumripper:
Are you sure older compilers didn't define `__ARM_NEON` for aarch64?

solardiz:
DES_bs_b.c is (too) smart in that it's able to handle large DES_BS_VECTOR even without SIMD. This was originally intended for PA-RISC, where some CPUs have a tiny L0 instruction cache, so benefit from loops around individual S-boxes. I think we happened to trigger this code on Aarch64.

solardiz:
No, I'm not sure about what might have happened with older compilers, etc.

magnumripper:
I added a commit with untested changes, basically adding `|| __aarch64__` to any `__ARM_NEON` check

solardiz:
I'm playing with the below in arm64le.h now:

```c
/*
 * Here we assume that we're on AArch64, which implies we have Advanced SIMD.
 */
#ifndef __ARM_NEON
/* Tell our originally 32-bit ARM code that we sort of have NEON */
#define __ARM_NEON
#endif

/* We do have native vandn() and vsel() for this architecture, but the timings
 * are often such that it might be better to minimize use of vandn() and avoid
 * vsel() altogether.  Otherwise, the best setting would have been DES_BS=3.
 */
```

magnumripper:
I need to sleep soon but I could throw a 64-bit dist into a Raspberry Pi first thing tomorrow.

solardiz:
Yes, let's continue tomorrow. Thank you for all your work!

solardiz:
I'm getting even better speeds by increasing DES_BS_VECTOR from 2 to 3 or 4, for 3x64 or 4x64-bit without SIMD. And maybe it's worth re-testing one of those mixed scalar+SIMD modes we had before...

solardiz:
On another machine `__ARM_NEON` is defined by default, and SIMD actually improves performance from ~1000K to ~1250K without vsel() or to ~1100K with vsel() - I guess the latency hurts again.

magnumripper:
I added benchmark length stuff now

magnumripper:
Note: In all Jumbo-specific code I'm now using things like `#if __ARM_NEON || __aarch64__` but I did not change `DES_bs_b.c` yet, I presume some change to it will come from core.

claudioandre-br:
> I'm pretty sure __ARM_NEON used to be defined on aarch64 builds. Otherwise pseudo_intrinsics.h would fail defining virtual intrinsics, for a starter. I'll blindly fix this but I'm not sure I'll have the time to test it. @claudioandre-br do you have some means of testing that?

A few interesting defines:

```
#define __ARM_32BIT_STATE 1
#define __ARM_ARCH 7
#define __ARM_ARCH_7A__ 1
#define __arm__ 1 // 32 bits only
#define __ARM_FEATURE_SIMD32 1
```

```
#define __ARM_64BIT_STATE 1
#define __ARM_ARCH 8
#define __ARM_ARCH_8A 1
#define __ARM_NEON 1
```
Defined here:
```
#define __aarch64__ 1
```

magnumripper:
I'm adding this
```diff
@@ -461,6 +462,11 @@ static char *fmt_self_test_body(struct fmt_main *format,
            format->params.plaintext_length > PLAINTEXT_BUFFER_SIZE - 3)
                return "plaintext_length";
 
+       if (format->params.benchmark_length < 1 ||
+           (format->params.benchmark_length & 0xff) >
+           format->params.plaintext_length)
+               return "benchmark_length";
+
        if (!is_poweroftwo(format->params.binary_align))
                return "binary_align";
 
```

claudioandre-br:
> enable NEON on Aarch64 [...] hurts performance now

This is not new #3436.

solardiz:
The universal replacement for the old `-1` is `0x107`, not `7`. A simple `7` works just as well only in formats with zero `salt_size` (which with the new code implies not doing separate many vs. one salt benchmarks), which is why I use it e.g. in core's LM_fmt.c now, but it's not universally valid.

solardiz:
@claudioandre-br I've seen no speedup or a few percent slowdown from NEON on some devices before. I've never seen a 25% slowdown before.

magnumripper:
> The universal replacement for the old `-1` is `0x107`, not `7`.

Working on it :-)

src/c3_fmt.c
```diff
@@ -211,7 +211,7 @@ static void init(struct fmt_main *self)
 			salt = options.subformat;
 			/* turn off many salts test, since we are not updating the */
 			/* params.tests structure data.                            */
-			self->params.benchmark_length = -1;
+			self->params.benchmark_length = 0x107;
```
solardiz: Maybe `self->params.benchmark_length |= 0x100;` would be friendlier to further maintenance.

src/dynamic_fmt.c
```diff
@@ -7463,7 +7463,7 @@ int dynamic_SETUP(DYNAMIC_Setup *Setup, struct fmt_main *pFmt)
 	pFmt->methods.salt_hash = salt_hash;
 	//pFmt->params.format_name = str_alloc_copy(Setup->szFORMAT_NAME);
 	pFmt->params.format_name = "";
-	pFmt->params.benchmark_length = 0;		// NOTE 0 'assumes' salted. If unsalted, we set back to -1
+	pFmt->params.benchmark_length = 7;		// If unsalted, we set to 0x107
```
solardiz: Maybe `BENCHMARK_LENGTH`.

src/dynamic_fmt.c
```diff
@@ -7613,7 +7613,7 @@ int dynamic_SETUP(DYNAMIC_Setup *Setup, struct fmt_main *pFmt)
 	if ( (Setup->flags & MGF_SALTED) == 0)
 	{
 		curdat.dynamic_FIXED_SALT_SIZE = 0;
-		pFmt->params.benchmark_length = -1;
+		pFmt->params.benchmark_length = 0x107;
```
solardiz: Maybe `BENCHMARK_LENGTH | 0x100`.

src/dynamic_compiler_fmt_plug.c
```diff
@@ -224,9 +224,9 @@ static void link_funcs() {
 	fmt_CompiledDynamic.params.flags &= ~FMT_SPLIT_UNIFIES_CASE;
 
 	if ((pPriv->pSetup->flags&MGF_SALTED)!=MGF_SALTED)
-		fmt_CompiledDynamic.params.benchmark_length = -1;
+		fmt_CompiledDynamic.params.benchmark_length = 0x107;
```
solardiz: Maybe `BENCHMARK_LENGTH | 0x100`.

src/dynamic_compiler_fmt_plug.c
```diff
@@ -224,9 +224,9 @@ static void link_funcs() {
...
 	else
-		fmt_CompiledDynamic.params.benchmark_length = 0;
+		fmt_CompiledDynamic.params.benchmark_length = 7;
```
solardiz: Maybe `BENCHMARK_LENGTH`.

src/bench.c
```diff
@@ -472,7 +506,7 @@ char *benchmark_format(struct fmt_main *format, int salts,
 		wait_salts = 1;
 		benchmark_time *= -1;
 	}
-	wait = format->params.benchmark_length ? 0 : wait_salts;
+	wait = format->params.benchmark_length & 0xe00 ? wait_salts : 0;
```
solardiz: What's this `wait` thing? Where do we set those bits?

solardiz:
Also, I suspect many of our uses of `-1` were actually misuses. We should have only used `-1`, and now `0x107`, where we're confident the speeds for many vs. one salt are almost the same. However, some of the uses look like they're on fast salted hashes, where I doubt those speeds can possibly be almost the same. Perhaps open a separate issue for that since it's not a core merge thing.

solardiz:
Don't we need any changes to `--test-full` for it to process the new meaning of `benchmark_length` correctly? IIRC, it's doing some sanity checks on that parameter, and those are probably out of date now.

src/c3_fmt.c
magnumripper: Agreed (although I'm just trying to merge now :)

src/bench.c
magnumripper: That's for `--test=-1` which will benchmark as long as it takes to get a complete many-salts figure.

src/dynamic_fmt.c
magnumripper: > Don't we need any changes to `--test-full` for it to process the new meaning of `benchmark_length` correctly?

I have no idea, I always hated that code for making a hopeless spaghetti mess of `formats.c`

magnumripper:
OK, as far as I can see, all is merged and fixed now. However, we've got a new strange segfault (or ASan crash) in agilekeychain.
```
magnum@mbpr:src [core-merge]$ make -sj8 && ../run/john -test -form:agilekeychain

Make process completed.
Will run 4 OpenMP threads
Benchmarking: agilekeychain, 1Password Agile Keychain [PBKDF2-SHA1 AES 256/256 AVX2 8x]... (4xOMP) 
==42108==ERROR: AddressSanitizer: heap-use-after-free on address 0x61d000016888 at pc 0x000101bbeb79 bp 0x70000590dc20 sp 0x70000590dc18
READ of size 4 at 0x61d000016888 thread T1
    #0 0x101bbeb78 in crypt_all._omp_fn.0 agilekeychain_fmt_plug.c:132
    #1 0x1041d4418 in gomp_thread_start (libgomp.1.dylib:x86_64+0xe418)

0x61d000016888 is located 8 bytes inside of 2131-byte region [0x61d000016880,0x61d0000170d3)
freed by thread T0 here:
    #0 0x104272c30 in wrap_free.part.0 (libasan.5.dylib:x86_64+0x6cc30)
    #1 0x101e25ddc in fmt_self_test formats.c:1610
    #2 0x101dff672 in benchmark_format bench.c:409
    #3 0x101e016a9 in benchmark_all bench.c:823
    #4 0x10226108b in main john.c:1700
    #5 0x7fff650d63d4 in start (libdyld.dylib:x86_64+0x163d4)

previously allocated by thread T0 here:
    #0 0x104271cdd in wrap_malloc (libasan.5.dylib:x86_64+0x6bcdd)
    #1 0x101e55b96 in mem_alloc memory.c:76
    #2 0x101e1cd55 in alloc_binary formats.c:1582
    #3 0x101e25d3a in fmt_self_test formats.c:1601
    #4 0x101dff672 in benchmark_format bench.c:409
    #5 0x101e016a9 in benchmark_all bench.c:823
    #6 0x10226108b in main john.c:1700
    #7 0x7fff650d63d4 in start (libdyld.dylib:x86_64+0x163d4)

Thread T1 created by T0 here:
    #0 0x104228894 in wrap_pthread_create (libasan.5.dylib:x86_64+0x22894)
    #1 0x1041d49f8 in gomp_team_start (libgomp.1.dylib:x86_64+0xe9f8)
    #2 0x1041cbe39 in GOMP_parallel (libgomp.1.dylib:x86_64+0x5e39)
    #3 0x7ffeee2c619f  (<unknown module>)

SUMMARY: AddressSanitizer: heap-use-after-free agilekeychain_fmt_plug.c:132 in crypt_all._omp_fn.0
  0x1c3a00002cc0: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x1c3a00002cd0: fd fd fd fd fd fd fd fd fd fd fa fa fa fa fa fa
  0x1c3a00002ce0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x1c3a00002cf0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x1c3a00002d00: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
=>0x1c3a00002d10: fd[fd]fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x1c3a00002d20: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x1c3a00002d30: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x1c3a00002d40: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x1c3a00002d50: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x1c3a00002d60: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  Addressable:           00
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==42108==ABORTING
Abort trap: 6
```
I can't see what's wrong. Perhaps we are now triggering some old bug in the format. However, the crash seem to be in formats.c.

The format wasn't altered, except this correct change:
```diff
$ git diff origin/bleeding-jumbo agilekeychain_fmt_plug.c
diff --git a/src/agilekeychain_fmt_plug.c b/src/agilekeychain_fmt_plug.c
index d99941182..18777f8e8 100644
--- a/src/agilekeychain_fmt_plug.c
+++ b/src/agilekeychain_fmt_plug.c
@@ -47,7 +47,7 @@ john_register_one(&fmt_agile_keychain);
 #define ALGORITHM_NAME          "PBKDF2-SHA1 AES 32/" ARCH_BITS_STR
 #endif
 #define BENCHMARK_COMMENT       ""
-#define BENCHMARK_LENGTH        -1
+#define BENCHMARK_LENGTH        0x107
 #define BINARY_SIZE             0
 #define BINARY_ALIGN            1
 #define SALT_ALIGN              sizeof(int)
```
I'm grabbing some food while pondering this.

magnumripper:
`--test-full=0` seems to work fine as-is (until it reaches agilekeychain)

solardiz:
Great to know there are no false positives from `--test-full`, but we wouldn't know from merely running it whether it still checks for all those properties it used to, or maybe it'd miss improper settings now. Also, maybe it has unneeded checks now, which we should drop. We need to review its code for uses of `benchmark_length`.

solardiz:
Regarding possible misuses or missed uses of `-1` before or `0x107` now, maybe we should add a builtin benchmarking mode for our own use only that would forcibly run many vs. one salt benchmarks for all formats and tell us where the setting is inappropriate, assuming it's run on an otherwise idle system and perhaps without OpenMP. We'll rerun it a couple of times to make sure, then make the edits it suggests.

magnumripper:
> We need to review its code for uses of benchmark_length.

I did now, there is none.

magnumripper:
Nailed it. Though I don't quite get it. Here's what made it work again:
```diff
@@ -422,6 +451,11 @@ char *benchmark_format(struct fmt_main *format, int salts,
                                format->methods.tunable_cost_value[i](salt);
 #endif
        }
+       /*
+        * Core john doesn't have this set_salt at all, only later under
+        * "if (salts > 1)". I first added it only "if (salts == 1)" but
+        * for some odd reason that lead to severe problems.
+        */
        format->methods.set_salt(two_salts[0]);
 
 #ifndef BENCH_BUILD
```

src/bench.c
```diff
@@ -309,15 +339,14 @@ char *benchmark_format(struct fmt_main *format, int salts,
...
+	int pass;
 #if OS_TIMER
 	struct itimerval it;
 #endif
-	clock_t start_real, end_real;
-	clock_t start_virtual, end_virtual;
 #if !defined (__MINGW32__) && !defined (_MSC_VER)
 	struct tms buf;
 #endif
+	clock_t start_real, start_virtual, end_real, end_virtual;
```
magnumripper: No-op changes like the above may look strange or irrelevant but I simply took some chances to get Jumbo's code closer to core, for things like `git diff origin/master bench.c` to be as clean as possible.

src/bench.c
magnumripper: I'll add the braces, thanks

magnumripper:
Merged manually

solardiz:
FWIW, the system that didn't define `__ARM_NEON` for me doesn't do it even when I pass `-march=armv8-a+simd` to its "gcc version 4.8.4 (Ubuntu/Linaro 4.8.4-2ubuntu1~14.04.4)", even though it accepts that option fine.

solardiz:
Further, quite possibly the slowdown is from this version of gcc failing to do register allocation well when compiling with ASIMD: I am getting larger `DES_bs_b.o` and with more instructions in it when I enable SIMD, vs. scalar. I think it's the other way around on systems where I saw a speedup from NEON/ASIMD. So perhaps just poor gcc.

solardiz:
gcc 8.2.0 on the same system does define `__ARM_NEON` and it produces smaller code and better speeds with SIMD (870K best) than with scalar (660K with the same settings that were best for gcc 4.8.4). However, in absolute terms both of those speeds are lower than gcc 4.8.4's scalar code (was 1040K).

solardiz:
@magnumripper I've just pushed some changes related to bitslice DES on Aarch64 to core. Please merge.

magnumripper:
Done

Issue #3737, opened 2019-03-30
finishing --show=types: moving out, renaming(?), documentation


AlekseyCherepanov:
@magnumripper asked me to move my code out from `ldr_split_line()` back in [2015](https://www.openwall.com/lists/john-dev/2015/11/23/1). I failed it miserably. I am sorry.

Since then, @magnumripper implemented json output (#2938).

I tried to make shortest possible format to reduce overhead and to avoid quoting. But the format turned out to be complicated and hard to use. Also it may fail (gracefully) when field separator char happens in field (for instance, in canon hash).

JSON is much more human friendly and does not require checking documentation every time. I tried to estimate size overhead: json is 2-4 times longer. It is not much.

I guess most JSON parsers require the whole input to be read before parsing. But it is easy to bypass this limitation and parse sequentially line by line: read line, strip left `[` and right `,` or `]`, parse as json. It will work as long as dictionary for original line/hash in json does not contain newlines and dictionaries are separated by newlines. JSON does not need newline in outputs, so we may keep it guaranteed.

So I'd drop original custom format and keep only json.

Also I'd rename it into `--show=formats`, while #3513 is not finished. I am not sure: it is about formats, john uses "format" everywhere, so "types" does not seem to fit well into that. OTOH `--show=formats` may be confused with `--list=formats`, but `--show=formats` require a hash file.

AlekseyCherepanov:
@math07 and @shinnok, what do you think about dropping custom output format of `--show=types` in favour of json format? I guess johnny is the only consumer for `--show=types`.

AlekseyCherepanov:
In current json, I'd change numeric "consistencyMark" to something readable.

shinnok:
Hi @AlekseyCherepanov,

From Johnny's perspective, there is a json parser [in Qt](https://doc.qt.io/qt-5/json.html) so it could be quick to implement and probably more robust parser yes.

Regarding dropping `--show=types` custom format altogether, I'll leave that for @solardiz and @magnumripper. Rename to `formats` is consistent too.

solardiz:
This isn't a core feature, so I'll leave it to @magnumripper @shinnok @AlekseyCherepanov to decide on.

On a related note, I still would like us to integrate Johnny into the jumbo tree after our upcoming 1.9.0-jumbo-1 release, but before the next release after it (which might be 2.0.0).

magnumripper:
[...]

PR #3773, opened 2019-04-08
document --show=formats option

AlekseyCherepanov:


AlekseyCherepanov:
Documentation for #3737

magnumripper:
BTW given the content, fitting to 80 would probably end up pretty much unreadable...

doc/SHOW_FORMATS_OPTION.md
```diff
@@ -0,0 +1,175 @@
...
+$ cat 123456.pot
+$dynamic_0$e10adc3949ba59abbe56e057f20f883e:123456
+```
+
+## Interaction with other options
+
+Formats to be tried may be limited with `--format=` option.
+- 1 exact format may be specified (e.g. `--format=raw-md5`),
+  - a disabled dynamic format may be enabled this way for temporary use,
+- other formats will not be checked and reported, it may be useful because `--show=formats` may be slow.
```
magnumripper: Just FWIW one could mention format wildcards around here, although they are (I think) mentioned somewhere else. Examples: `--format=*crypt`, `--format=mssql*`, `--format=raw*256`. They may come handy using this.

AlekseyCherepanov:
I might rewrite it without so many nested lists.

AlekseyCherepanov:
I have blocks of code and I saw DYNAMIC_COMPILER_FORMATS.md, so I considered to use markdown. Most of text in DYNAMIC_COMPILER_FORMATS.md does not use single line for paragraph splitting lines at 79. Few lines break this rules. Also text in blocks is not split. I missed this and made single lines for text too.

doc/SHOW_FORMATS_OPTION.md
AlekseyCherepanov: I intended to write that, but it does not work really.

AlekseyCherepanov: hm, raw*256 works; I tried --format=*-ztex and john tells `Unknown ciphertext format name requested`

doc/SHOW_FORMATS_OPTION.md
```diff
@@ -0,0 +1,175 @@
+# SHOW FORMATS OPTION
+
+This file tries to document `--show=formats` option (and a bit of deprecated `--show=types`).
+
+john loads hashes of one format from given hash files. john gives hints about some other formats seen in the file. `--show=formats` will try every line against every format and show formats that can load it independently. It does not mean that load all lines at the same time.
```
AlekseyCherepanov: `It does not mean that john can load all lines at the same time.`

doc/SHOW_FORMATS_OPTION.md
```diff
@@ -0,0 +1,175 @@
...
+Each dictionary in "rowFormats" list may have the following field/keys:
+- "label" is the name of format that may be used for `--format=` option,
+- "dynamic" is boolean value,
+  - it is true if format uses engine for dynamic formats,
+  - it is absent if it is false,
+- "prepareEqCiphertext" is boolean value,
+  - it is true if `prepare()` method of formats returned same ciphertext after processing (it may be interesting to developers of formats),
+  - it is absent if it is false,
+- "canonHash" is a list containing ciphertext in canonical form,
+  - it may contain multiple values for some formats of hash files (e.g. pwdump),
```
AlekseyCherepanov: It does not depend onto format of file. It is from format purely. 32 hex LM has 2 halves there.

doc/SHOW_FORMATS_OPTION.md
magnumripper: For `--format=*-ztex` to work you need to have a ZTEX-enabled build. Other than that, it really should work.

solardiz: We're no longer registering the ZTEX formats with `--show` in order to avoid initializing the boards unnecessarily and allow for `--show` to work concurrently with actual use of those boards.

solardiz:
We have a list of documentation files in the main README. Do we intend to maintain that or maybe drop it?

doc/SHOW_FORMATS_OPTION.md
magnumripper: Ah, yes. I have considered doing something similar for OpenCL.

magnumripper:
Unless someone other than me are willing to maintain it, let's drop it.

AlekseyCherepanov:
I made most changes to show to you before I rewrite them into 80-char style. I think filename should be just SHOW_FORMATS.

Should I make a single commit with the file in separate pull request?

claudioandre-br:
> Should I make a single commit with the file in separate pull request?

My two cents:
* Github is very powerful, that is not necessary. Besides, anyone who already reviewed this PR, will need to review all document again.
* If you add new commits (or `push -f`) to this one, only the changes will have to be analysed.

doc/SHOW_FORMATS_OPTION.md
```diff
@@ -0,0 +1,229 @@
...
+BBODHXVAdtcmc
+```
+
+(`--format=descrypt` is specified to reduce output.)
+```
+$ ./JohnTheRipper/run/john --show=formats ab.pw --format=descrypt
+[{"lineNo":1,"ciphertext":"AAa6CzJlsalyo","rowFormats":[{"label":"descrypt","prepareEqCiphertext":true,"canonHash":["AAa6CzJlsalyo"]}]},
+{"lineNo":2,"ciphertext":"BBODHXVAdtcmc","rowFormats":[{"label":"descrypt","prepareEqCiphertext":true,"canonHash":["BBODHXVAdtcmc"]}]}]```
+
+Example: 1 descrypt hash and pretty printing:
```
magnumripper: At least the GitHub ML viewer misses the ``` on line 31 because there's no NL before it. Then the rest of file show as "reversed", all pre-formatted is plain and vice versa

magnumripper:
Actually I think I prefer wider format as long as we're using ML anyway. So to me it's fine as-is. @solardiz what do you think about it? I like ML because it's very readable in any dumb viewer as well.

doc/SHOW_FORMATS_OPTION.md
AlekseyCherepanov: There is no NL actually. I did not try to render. My bad.

solardiz:
@magnumripper What do you mean by ML? A generic markup language? Markdown? I'm OK with using Markdown for this file. I'm also OK with occasionally going beyond 80 chars when there's a good reason for that. I think we should try to stay below 80 chars when we reasonably can do so without hurting readability for those people with wider terminals.

For source code in core, I'm currently trying to stay below 80 chars. In jumbo, below 132. That's assuming 8-char tabs, even though I think some files in jumbo now assume 4-char tabs (which annoys me, especially given that those files' portions inherited from core have their 4-space indentation of continuation lines visibly indistinguishable from jumbo's 4-char tab full indentation of sub-blocks, so even setting tab width to 4 doesn't make those files display reasonably).

But documentation is different, with a slightly stronger preference on staying below 80.

AlekseyCherepanov:
Nice! I can split paragraph into multiple lines and Github still thinks it is a paragraph. That's not regular Markdown, and it does not work in comments on github.

AlekseyCherepanov:
Also it seems that github-markdown-preview (version 4.0.2; quite old) puts last point in a sublist too close to next point in the upper/current list.

AlekseyCherepanov:
I did `git mv SHOW_FORMATS_OPTION.md SHOW_FORMATS.md`. I did not fix the title `SHOW FORMATS OPTION`, ok?

AlekseyCherepanov:
I implemented all changes and fixes I wanted to implement for this version.

doc/OPTIONS
```diff
@@ -180,6 +180,7 @@ using --show=left you normally need to supply the --format option too.
 --show=types				or in old custom format
 
 Shows the formats/types of input hashes contained in the input file.
+See doc/SHOW_FORMATS_OPTION.md
```
magnumripper: File renamed!

magnumripper:
> @magnumripper What do you mean by ML? A generic markup language? Markdown? I'm OK with using Markdown for this file.

I probably mean Markdown, me old git not very good at these modern things 🙃

magnumripper:
Thanks! So should I merge now?

doc/OPTIONS
AlekseyCherepanov: I fixed this and rewrote the commit to include the change.

AlekseyCherepanov:
Yes, you may merge. Thanks!

magnumripper:
I think this ended up terrific! It's very readable in `less` with 80 columns, yet fully featured nice looking document in an MD viever. We should do more documentation this way.

PR #3776, opened 2019-04-09
OpenCL: tune for the length a run will be using

claudioandre-br:
For tuning, use the largest of these:
- mask_add_len (if -mask)
- options.req_maxlength (if set, and min not set)
- options.req_minlength (if set, and max not set)
- self->params.benchmark_length & 0x7f

For J2, maybe.

### DIFF
--- a/src/opencl_common.c
+++ b/src/opencl_common.c
@@ -1536,8 +1536,21 @@ static cl_ulong gws_test(size_t gws, unsigned int rounds, int sequential_id)
 	self->methods.clear_keys();
 	{
 		char key[PLAINTEXT_BUFFER_SIZE];
-		int len = MAX(MIN(self->params.plaintext_length, 7),
-		              self->params.plaintext_min_length);
+		int len;
+
+		if (mask_add_len > 0 &&
+		    (options.req_minlength == -1 && options.req_maxlength == 0)) {
+			//-mask and no min/max are set
+			len = mask_add_len;
+		} else {
+			len = (self->params.benchmark_length & 0x7f);
+			len = (len < options.req_minlength) ? options.req_minlength: len;
+			len = (options.req_maxlength && len > options.req_maxlength) ?
+			          options.req_maxlength: len;
+		}
+		// Obey format's min and max length
+		len = MAX(len, self->params.plaintext_min_length);
+		len = MIN(len, self->params.plaintext_length);
 
 		memset(key, 0x61, sizeof(key));
 


solardiz:
The fallback to `self->params.benchmark_length & 0x7f` may result in tuning for a length that isn't in the min to max range when both min and max are set. I suggest:

- if `-mask` and no min/max are set, just use mask length, otherwise:
- tentatively choose `self->params.benchmark_length & 0x7f`, but:
- if min is set and the previously chosen length is below min, then use min, and
- if max is set and the previously chosen length is above max, then use max.

In other words, if our default benchmark & auto-tune length is in range, just use that. Otherwise clamp it to range.

solardiz:
I set this to 1.9.0-jumbo-1 milestone so that we keep this in consideration for this release. I think the algorithm I just outlined is simple enough for that.

src/opencl_common.c
```diff
@@ -1536,8 +1536,21 @@ static cl_ulong gws_test(size_t gws, unsigned int rounds, int sequential_id)
...
+		if (mask_add_len > 0 &&
+		    (options.req_minlength == -1 && options.req_maxlength == 0)) {
+			//-mask and no min/max are set
+			len = mask_add_len;
+		} else {
+			len = (self->params.benchmark_length & 0x7f);
+			len = (len < options.req_minlength) ? options.req_minlength: len;
+			len = (options.req_maxlength && len > options.req_maxlength) ?
+			          options.req_maxlength: len;
+		}
```
magnumripper: Perhaps use the MIN/MAX macros here as well? Not that it matters for the resulting binary.

magnumripper: Doing it one step at a time is good though. Even if `MAX(MIN(self->params.plaintext_length, 7), self->params.plaintext_min_length)` is "effective" it's hard to read, and more so if there are 2-3 more levels of it.

magnumripper:
`../run/john -form:rar-opencl all_tests.in -mask:alpha?abravo` will tune for length 5 instead of 11 because we auto-tune too early - I believe even before self test. This is a temporary problem, after release we'll revise stuff so self-tests never run auto-tuned but with low (eg LWS 7, GWS 49) group sizes. Also, for most formats this will mean an auto-tune length of 7 or 8, which is fine for now.

src/opencl_common.c
```diff
@@ -1536,8 +1536,21 @@ static cl_ulong gws_test(size_t gws, unsigned int rounds, int sequential_id)
...
+
+		if (mask_add_len > 0 &&
+		    (options.req_minlength == -1 && options.req_maxlength == 0)) {
+			//-mask and no min/max are set
+			len = mask_add_len;
+		} else {
+			len = (self->params.benchmark_length & 0x7f);
+			len = (len < options.req_minlength) ? options.req_minlength: len;
+			len = (options.req_maxlength && len > options.req_maxlength) ?
+			          options.req_maxlength: len;
```
solardiz: Nitpick: I think `len = cond ? something : len` (note `len` used on both sides) is misuse of the ternary operator, and would be cleaner written using `if`. So instead of:

```c
len = (len < options.req_minlength) ? options.req_minlength: len;
len = (options.req_maxlength && len > options.req_maxlength) ? options.req_maxlength: len;
```

We can have:

```c
if (len < options.req_minlength)
        len = options.req_minlength;
if (options.req_maxlength && len > options.req_maxlength)
        len = options.req_maxlength;
```

solardiz:
> after release we'll revise stuff so self-tests never run auto-tuned

Actually, I think we should run self-test after auto-tune, because there's risk that a change in GWS/LWS breaks correct behavior of a kernel (buggy, or with a buggy OpenCL backend), which we'd want to detect.

solardiz:
> we auto-tune too early (mask mode not initialized)

Can we possibly run poor man's mask length calculation early and use that for auto-tune? Count the number of non-? characters, the number of ?c specifiers, and count anything in square brackets as one character. It's 5 lines of code or so.

claudioandre-br:
>> after release we'll revise stuff so self-tests never run auto-tuned

> Actually, I think we should run self-test after auto-tune, because there's risk that a change in GWS/LWS breaks correct behavior of a kernel (buggy, or with a buggy OpenCL backend), which we'd want to detect.

You guys should set a rule. Mine is:
* I ALWAYS tune for the real db.
* I don't tune for self-test. And I dont tune twice.

solardiz:
How about: tune for the real attack & db, but self-test after tune. No need to tune twice.

claudioandre-br:
> How about: tune for the real attack & db, but self-test after tune. No need to tune twice.

Probably the best approach.

solardiz:
This PR changed the auto-tune length in `gws_test`, but forgot to make similar change in `opencl_find_best_lws`. I think this is a bug. Also, ideally code duplication between these two functions should be avoided (duplicate code moved to a shared function called by both).

magnumripper:
Some (as in several) formats are so slow, self-test after auto-tune will simply be WAY to slow. I've worked hard to get rid of that. Years ago, a simple -test=0 would take tens of minutes with some formats.

solardiz:
I think self-test after auto-tune is only slow because we invoke `crypt_all` multiple times, testing just one slot at a time. This ancient approach at self-testing (coming from core) will have to go, or at least be avoided/replaced when `max_keys_per_crypt` is very high.

claudioandre-br:
> Some (as in several) formats are so slow, self-test after auto-tune will simply be WAY to slow. I've worked hard to get rid of that. Years ago, a simple -test=0 would take tens of minutes with some formats (we're talking per format).

For example, it will be **impossible** to use bitlocker. It will be annoying to use sha512crypt-opencl.

ME: I would love to split self-test and benchmark in two command line options.

magnumripper:
> I think self-test after auto-tune is only slow because we invoke `crypt_all` multiple times, testing just one slot at a time.

That's not the worst problem. Some formats do take tens of seconds to test any simple password when tuned for performance. *And worse formats will definitely come - Bitlocker is just the beginning*.

Here's how auto-tune is so fast: Let's say we do PBKDF2 with a million iterations. We have a loop kernel that does 250 at a time, and call it 4000 times. But during auto-tune, we call it just ONCE and multiply the performance data with 4000. Et voila, we have a usable data in a tiny fraction of the time. The resulting hash, however, is obviously useless 🤣 

The only way to do things like that in self-tests would be to use fake hashes with altered iteration counts. If that's even possible, Going that route we'd be on much thinner ice than using low worksizes and real self-tests.

magnumripper:
Here's my personal opinion of future scheme:
1. We set LWS 7 and GWS 49
2. reset with test_db (made from test vectors)
3. self-test

Then, if benchmark:
4. reset with test_db, now auto-tune for performance and for benchmark_length or, if given, custom mask
5. benchmark

or, if real run:
4. reset with real_db, now auto-tune for performance, for actual costs and number of salts and considering actual lengths the mode will use
5. fire up some cracking mode

solardiz:
> The only way to do things like that in self-tests would be to use fake hashes with altered iteration counts, if that's even possible. Going that route we'd be on much thinner ice than using low worksizes and real self-tests.

I wouldn't call this "thinner ice" if it's an additional self-test on top of the one you suggest at low LWS/GWS, but with full iteration count test vectors. The only difficulty is that we'd need to produce those lighter test vectors and add a way to specify them (a formats interface extension). I actually like this idea.

I do something similar in crypt_blowfish, where it always self-tests itself with a "bcrypt cost 0" test vector.

PR #3777, opened 2019-04-09
Mask-mode benchmark: Add/replace benchmark_comment with "(length n)" …

magnumripper:
…if a custom mask is being specified.  Also separate used mask from requested mask (so we can see whether mask is default or not).


magnumripper:
Merged manually (not sure why it's not seen here, it used to be seen)

PR #3781, opened 2019-04-10
Add cmp_all() to auto-tune.

magnumripper:
Triggered by https://github.com/magnumripper/JohnTheRipper/issues/3779#issuecomment-481474190

### DIFF
--- a/src/opencl_common.c
+++ b/src/opencl_common.c
@@ -1510,8 +1510,9 @@ static void clear_profiling_events()
 	}
 }
 
-// Fill [set_salt(), set_key()] the OpenCL device with data
-static void* fill_opencl_device(size_t gws)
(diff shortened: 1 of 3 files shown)


claudioandre-br:
BTW: it is still inaccurate since you are not measuring `copy_hash_back()`.

magnumripper:
We should only include `copy_hash_back` in case it's normally called for every `crypt_all()` (all tuning should be for the "no crack" fast paths). So it depends... given a small enough on-device binary we may need to include that one too but normally I'd say we shouldn't.

magnumripper:
This is now complete and I intend to merge it after bots say it's fine. First because not including the compare kernel durations was a bug, and second because this was trivial so I expect no regressions.

PR #3782, opened 2019-04-10
md5crypt-opencl: only detect divergence in 64 lid's near current

solardiz:
This speeds up the check when LWS is greater than 64 _and_ it avoids unnecessary disabling of length specialization when there are different length passwords somewhere within the local group, but not in the immediate group of 64.

### DIFF
--- a/src/opencl/cryptmd5_kernel.cl
+++ b/src/opencl/cryptmd5_kernel.cl
@@ -619,7 +619,7 @@ __kernel void cryptmd5(__global const crypt_md5_password *inbuffer,
  * Disable specialization to password length if it varies within the group.
  * We could use tree-like log2(N) reduction here, but we don't bother.
  */
-	__local uint lengths[0x100];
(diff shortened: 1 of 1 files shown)


solardiz:
Mysteriously, the increase of `length` array size from 0x100 to 0x400 elements that I included here has resulted in a 9% speedup on well's Tahiti at length 8 (from 3909K to 4267K on a test I just ran), even though it tunes to LWS=64 anyway (so the rest of the changes are no-op for this case). I was able to isolate the speedup to just this one change by making edits to just this one digit in the kernel - yes, this is what causes the change.

magnumripper:
My favorite 4x boost was b755050c

solardiz:
That one with RC4 actually makes sense to me - it looks like bank conflict avoidance.

magnumripper:
Absolutely. But how would you know, other than trial and error? Things like that should ideally be taken care of by the toolchain.

Issue #3780, opened 2019-04-09
Complete "OpenCL: tune for the length a run will be using"


solardiz:
The changes merged as part of #3776 are great, but incomplete:

1. That PR changed the auto-tune length in `gws_test`, but forgot to make similar change in `opencl_find_best_lws`. I think this is a bug. Also, ideally code duplication between these two functions should be avoided (duplicate code moved to a shared function called by both).

2. Can we possibly run poor man's mask length calculation early and use that for auto-tune? Count the number of non-? characters, the number of ?c specifiers, and count anything in square brackets as one character. It's 5 lines of code or so.

3. Nitpick: I think `len = cond ? something : len` (note `len` used on both sides) is misuse of the ternary operator, and would be cleaner written using `if`. So instead of:

```c
len = (len < options.req_minlength) ? options.req_minlength: len;
len = (options.req_maxlength && len > options.req_maxlength) ? options.req_maxlength: len;
```

We can have:

```c
if (len < options.req_minlength)
        len = options.req_minlength;
if (options.req_maxlength && len > options.req_maxlength)
        len = options.req_maxlength;
```

magnumripper:
Some (as in several) formats are so slow, self-test after auto-tune will simply be WAY to slow. I've worked hard to get rid of that. Years ago, a simple -test=0 would take tens of minutes with some formats.

magnumripper:
wrong thread, sorry :)

magnumripper:
@claudioandre-br could you take this please?

magnumripper:
Regarding 2. I think it's going backwards. Normally we should self-test with low worksizes. Normally, bugs will show up anyway (but sure, not always). Try doing a self-test with bitlocker-opencl tuned for a real run. Good luck with that.

solardiz:
Wrong thread again? "2" here says nothing about self-test.

magnumripper:
2 is needed only if you want to auto-tune before self-test

solardiz:
You wrote: "`../run/john -form:rar-opencl all_tests.in -mask:alpha?abravo` will tune for length 5 instead of 11 because we auto-tune too early (mask mode not initialized) - I believe even before self test."

I read this as implying the condition "if -mask and no min/max are set, just use mask length" doesn't work yet. I see no reason why we can't make it work now (for the release), as I suggested. I see little relevance to self-tests here.

magnumripper:
[...]

PR #3783, opened 2019-04-10
OpenCL: tune for the length a run will be using

claudioandre-br:
Print the length used for auto-tune.

Avoid code duplication.

For auto-tune keep length in range:
- if -mask and no min/max are set:
  - just use mask length

- otherwise:
  - tentatively choose self->params.benchmark_length & 0x7f, but:
  - if min is set and the previously chosen length is below min, then use min, and
  - if max is set and the previously chosen length is above max, then use max, and
  - similarly obey the format's min/max constraints, if any.


magnumripper:
This was for #3780

PR #3785, opened 2019-04-10
Fix some GPU device number reporting problems. Closes #3639

### LINKED ISSUE #3639: Change to 1-based OpenCL device numbers
This was decided in #3637.

I believe this can be done with **very** little code changes if we do this right, as it's only cosmetic. I guess it's best to stay 0-based throughout all code except when mapping from `--device` option and/or when outputting to screen/log. Oh, and when doing automagic device selection for fork/MPI.

magnumripper:


magnumripper:
I changed "GPU n" to "Device n" for log output. For screen output, lines can potentially be extremely long (think mscash2-opencl using 8 devices) so I changed "GPUn" to "Dev#n".

PR #3790, opened 2019-04-11
Upgrade from escrypt to yescrypt

solardiz:
This replaces the old escrypt tree (containing many JtR hacks) with the current yescrypt tree (almost pristine) in a relatively straightforward way.

I also drop the two `escrypt/Makefile.*` files without providing replacements to them because those were unused anyway - the top-level `Makefile` only used three `escrypt/*.o` files, which it built on its own. The `escrypt/Makefile.*` files thus provided almost nothing but confusion.

After we merge this, we should proceed with many related cleanups and enhancements - addition of error checking, avoidance of freeing and re-allocating memory (issue present in all scrypt-using formats other than the main scrypt format), and addition of support for yescrypt hashes (with this PR, we're merely replacing the implementation used for scrypt, without exposing its ability to also compute yescrypt hashes).


solardiz:
Please don't squash the 3 commits when merging this. I deliberately made them separate.

solardiz:
The "Added ULL to the 64-bit YESCRYPT_ROM_TAG* constants" commit should ideally have been to upstream before getting it in here, but I have no time to redo it all. I've made the same commit upstream separately.

solardiz:
CircleCI was already happy with this before my last commit, which only changed comments, so I'll merge this.

PR #3794, opened 2019-04-12
Fix some unaligned accesses, enhance some salt_hash()

solardiz:
With this, all formats (except for "monero", now disabled) passed `-test` for me on sparc64. (Of course, it's always possible some of the remaining formats passed test by lucky rather than guaranteed alignment.)

### DIFF
--- a/src/phpassMD5_fmt_plug.c
+++ b/src/phpassMD5_fmt_plug.c
@@ -322,7 +322,7 @@ static void *get_salt(char *ciphertext)
 
 static int salt_hash(void *salt)
 {
-	return *((ARCH_WORD *)salt) & 0x3FF;
+	return *(uint32_t *)salt & (SALT_HASH_SIZE - 1);
 }
 
 struct fmt_main fmt_phpassmd5 = {
--- a/src/ansible_fmt_plug.c
+++ b/src/ansible_fmt_plug.c
@@ -47,7 +47,7 @@ john_register_one(&fmt_ansible);
 #define BENCHMARK_LENGTH        7
 #define PLAINTEXT_LENGTH        125
 #define SALT_SIZE               sizeof(struct custom_salt)
-#define BINARY_ALIGN            1
+#define BINARY_ALIGN            sizeof(uint32_t) // see cmp_all() and ansible_common_get_binary()
 #define SALT_ALIGN              sizeof(uint64_t)
 #ifdef SIMD_COEF_32
 #define MIN_KEYS_PER_CRYPT      SSE_GROUP_SZ_SHA256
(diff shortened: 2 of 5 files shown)


src/phpassMD5_fmt_plug.c
```diff
@@ -322,7 +322,7 @@ static void *get_salt(char *ciphertext)
 
 static int salt_hash(void *salt)
 {
-	return *((ARCH_WORD *)salt) & 0x3FF;
+	return *(uint32_t *)salt & (SALT_HASH_SIZE - 1);
```
magnumripper: Ouch. Exact same problem in drupal7_common.h

solardiz:
There are also, but these are not alignment problems - just too small hash value ranges:

```
nsec3_fmt_plug.c:			hash &= 0x3FF;
nsec3_fmt_plug.c:	hash &= 0x3FF;
opencl_bf_fmt_plug.c:	return ((BF_salt *)salt)->salt[0] & 0x3FF ;
```

`opencl_bf_fmt_plug.c` should be trivial to fix to use `SALT_HASH_SIZE - 1`.

magnumripper:
I believe nsec3 is some other thing. Not sure why I missed opencl_bf - I did grep for it. I'll fix it.

magnumripper:
Oh, I missed it because of that f@cking white space violation.

solardiz:
nsec3 looks like the same thing, just a more complex hash function, so not as easy to fix to support arbitrary size.

magnumripper:
That strange strange man put a space before `;` in 68 lines out of 229.

magnumripper:
Shouldn't this work?
```diff
static int salt_hash(void *salt)
{
	unsigned int hash = 0;
	int i;
	for (i = 0; i < SALT_SIZE; ++i) {
		hash <<= 1;
		hash += (unsigned char)((unsigned char *)salt)[i];
		if (hash >> 10) {
			hash ^= hash >> 10;
-			hash &= 0x3FF;
+			hash &= (SALT_HASH_SIZE - 1);
		}
	}
	hash ^= hash >> 10;
-	hash &= 0x3FF;
+	hash &= (SALT_HASH_SIZE - 1);

	return hash;
}
```

magnumripper:
Contest version had the same problem btw

solardiz:
The right shifts by 10 also assume hash size. Perhaps changing them to SALT_HASH_LOG would work.

magnumripper:
I'll just do that. It shouldn't break anything even if not optimal, right?

solardiz:
I guess so. BTW, there's a redundant second cast to `(unsigned char)`.

PR #3821, opened 2019-04-14
OpenCL: SHA256 and SHA512 force tune after fmt_done()

claudioandre-br:
This is harmless and will be needed after J1.


solardiz:
Why will this be needed?

claudioandre-br:
For autotune, in short. After a `fmt_done()`, all "status" info of the format will be cleaned.

The mandatory `-mask` on benchmark will need it.

solardiz:
> The mandatory `-mask` on benchmark will need it.

Does this imply that we already need this for the optional `-mask` on benchmark?

claudioandre-br:
No. Only after magnun's patch (to force `--mask` or to self-test when `--mask`) it will be required.

PR #3827, opened 2019-04-16
Change or remove the link for the Windows Build

claudioandre-br:
This is a reminder to (do something)


solardiz:
Is the change proposed here correct as-is?

Is its rationale so that we can re-point the redirect target later, so that people using this link from 1.9.0-jumbo-1's docs get the currently working link?

Maybe I should setup an equivalent redirect on openwall.com, so that we don't depend on a third-party link shortening/redirect service?

claudioandre-br:
> Is the change proposed here correct as-is?

Yes

> Is its rationale so that we can re-point the redirect target later

No. But that will happen and people using this link from 1.9.0-jumbo-1's docs **WILL** get the (future) version of JtR currently under development.

BTW: the removal could be the first/better choice.

solardiz:
> BTW: the removal could be the first/better choice.

Do you mean removing this line from the file to be included in the release tarball only, but keeping it in the repo?

claudioandre-br:
>>BTW: the removal could be the first/better choice.

>Do you mean removing this line from the file to be included in the release tarball only, but keeping it in the repo?

No, but it is one possible choice.

solardiz:
Claudio, would you like this merged? If so, please feel free to merge it now. Thanks!

PR #3828, opened 2019-04-16
bcrypt-ztex: fixed memory leak, correction in docs

Apingis:


solardiz:
I added some edits to comments before merging this. I'll test the actual fix next.

@Apingis for your next update please also include the default frequency change to 150 in both the format source and john.conf. Thanks!

solardiz:
I re-ran the same attack that previously ate my 8 GB at ~32 minutes to up to 42 minutes now, and it's still going fine without growth in memory usage, which is about 63 MB per process for the four processes. So I think this issue is fixed for good. :-)

PR #3833, opened 2019-04-16
ztex improvements

### LINKED ISSUE #3709: Support concurrent --show while cracking with a -ztex format
The "device busy" conflict identified in #3379 also prevents "--show" from working while cracking. I have to explicitly specify e.g. "--show --format=descrypt" to override autodetection as "descrypt-ztex". It's unlikely our users would guess to do that. But this gave me an idea: maybe we can have john.c skip registration of the -ztex formats when the invocation is for "--show" (check that flag)?

Alternatively, maybe we can have -ztex formats themselves not access the actual devices when used in "--show".

Apingis:
- Moved board initialization from ```init()``` to ```reset()```. That fixes all ```show=``` options, removes the necessity to skip format registration dependent on command-line options.
- Print the list of boards only once per run. Don't print every board if ```verbosity < VERB_DEFAULT```.
- Print count of boards.
- bcrypt-ztex default frequency is set to 150 MHz.
- small fixes, documentation updates.
This fixes #3709

### DIFF
--- a/run/john.conf
+++ b/run/john.conf
@@ -421,9 +421,9 @@ Frequency = 190
 # Define typical setting of hashes it's going to process. It allows
 # to adjust for best performance.
 TargetSetting = 5
-# Startup frequency for bcrypt-ztex is 141. Design tools guaranteed
-# 141.5 in worst-case temperature and voltage.
-Frequency = 141
+# Startup frequency for bcrypt-ztex is 150. At this frequency tested boards
+# perform reliably.
+Frequency = 150
 # For any algorithm it's possible to set frequency on per-board and
 # per-FPGA basis, but the lowest frequency will determine performance.
 #Frequency_04A36E0FD6 = 142
--- a/run/john.conf
+++ b/run/john.conf
@@ -421,8 +421,8 @@ Frequency = 190
 # Define typical setting of hashes it's going to process. It allows
 # to adjust for best performance.
 TargetSetting = 5
-# Startup frequency for bcrypt-ztex is 150. At this frequency tested boards
-# perform reliably.
+# Design tools reported possible frequency to be 141.5 MHz.
+# Tested boards work reliably at 150, so that's what we use by default.
 Frequency = 150
 # For any algorithm it's possible to set frequency on per-board and
 # per-FPGA basis, but the lowest frequency will determine performance.
--- a/src/john.c
+++ b/src/john.c
@@ -361,15 +361,13 @@ static void john_register_all(void)
 
 	/* Let ZTEX formats appear before CPU formats */
 #ifdef HAVE_ZTEX
-	if (!(options.flags & FLG_SHOW_CHK)) {
-		john_register_one(&fmt_ztex_descrypt);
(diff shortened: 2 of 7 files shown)


solardiz:
Looks like we have a regression here. This works OK with yesterday's bleeding-jumbo, but produces errors with today's:

```
$ ./john -mask='?a' -min-len=1 -max-len=5 -verb=1 -no-log pw-bcrypt-50k 
Warning: detected hash type "bcrypt-ztex", but the string is also recognized as "bcrypt"
Use the "--format=bcrypt" option to force loading these as that type instead
Loaded 50000 password hashes with no different salts (bcrypt-ztex [Blowfish ZTEX])
4 device(s) ZTEX 1.15y ready
0g 0:00:00:00  (1) 0g/s 0p/s 0c/s 0C/s
Warning: salt with 50000 hashes, onboard comparators support up to 512 hashes/salt, turned off
10g 0:00:00:00 1.04% (2) (ETA: 16:30:58) 66.66g/s 633.3p/s 633.3c/s 31666KC/s aa..a|
SN 2 FPGA #4 error: app_status=0x10, debug=0xd3d2
SN 2 error -1 doing r/w of FPGAs (LIBUSB_ERROR_IO)
SN 4 FPGA #4 error: app_status=0x10, debug=0xd3d2
SN 4 error -1 doing r/w of FPGAs (LIBUSB_ERROR_IO)
SN 1 FPGA #4 error: app_status=0x10, debug=0xd3d2
SN 1 error -1 doing r/w of FPGAs (LIBUSB_ERROR_IO)
SN 3 FPGA #4 error: app_status=0x10, debug=0xd3d2
SN 3 error -1 doing r/w of FPGAs (LIBUSB_ERROR_IO)
```

I tried switching between the two revisions back and forth, and this is fully reproducible.

solardiz:
When I set `-min-len=2` or higher, the issue goes away. Maybe it's related to the first mask length completing too quickly in terms of candidate count or/and real time.

solardiz:
With `-mask='?d'`, the issue appears with `-min-len=3` and lower (goes away starting with 4).

Apingis:
Confirmed.
Currently the approach to turn off comparators is 'quick & dirty'. It does not allow to turn on them back, doesn't handle errors correctly.
I've got an idea how to do it properly, change hardware design to check number of hashes supplied in ```PKT_CMP_CONFIG``` and do the job regarding comparison and output base on that setting.

solardiz:
@Apingis I don't see how what you describe is relevant because the issue at hand showed up only with your latest set of changes, not with the introduction of transfer of computed hashes to host. Like I say, with yesterday's code revision - excluding _only_ this latest PR - the same test passes without errors.

Apingis:
The transfer-to-host mode setting doesn't persist after board (re-)initialization which was moved to ```reset()``` in the lastest PR.
```reset()``` occurs when mask iterates over lengths, on every new length.

Additionally board re-initialization occurs after any other error, in such case it would not initialize with proper transfer-to-host mode setting and I described that "doesn't handle errors correctly".

PR #3837, opened 2019-04-17
OpenCL: rawSHA512 force auto-tuning

claudioandre-br:
See recent cracking contest discussion.

### DIFF
--- a/src/opencl_rawsha512_gpl_fmt_plug.c
+++ b/src/opencl_rawsha512_gpl_fmt_plug.c
@@ -376,7 +376,7 @@ static void tune(struct db_main *db)
 {
 	char *tmp_value;
 	size_t gws_limit;
-	int autotune_limit = 100;
+	int autotune_limit = 500;
 
 	if ((tmp_value = getenv("_GPU_AUTOTUNE_LIMIT")))
 		autotune_limit = atoi(tmp_value);


solardiz:
Thanks, @claudioandre-br.

This reverts part of:

```
commit f487b6be3b158177a5529416e71d9d4871428f24
Author: magnum <john.magnum@hushmail.com>
Date:   Thu Jun 21 11:29:26 2018 +0200

    Stop piggy-backing arguments and drop the notion of "max crypt_all time"
    from OpenCL autotune, it was an artifact. Closes #3278.
```

We should check if the rest of that commit possibly caused performance regressions for other formats as well.

solardiz:
Meanwhile, CircleCI failed for an unrelated reason:

```
Testing: crypt, generic crypt(3) [?/32]... Warning: hash encoding string length 13, type id #1
appears to be unsupported on this system; will not load such hashes.
FAILED (valid (before init))

1 out of 711 tests have FAILED
Exited with code 1
```

I guess the system somehow has descrypt excluded or disallowed in its libc or libcrypt code or settings. IIRC, I heard this happens with FIPS enforcing mode on Red Hat'ish distros. Perhaps we should separately make changes to our generic crypt(3) format to support such systems. We could choose to leave things as-is (the user is effectively, even if confusingly to a non-expert, informed that their crypt(3) lacks support for these hashes, and the "failing" test is conveniently the very last one), but then we need to somehow avoid or suppress the "error" on our CI systems.

claudioandre-br:
It happens randomly on CI as it is now. A rebuild on Circle will fix it.

I haven't seen this on my Fedora tests. Any workaround seems premature.

solardiz:
> We should check if the rest of that commit possibly caused performance regressions for other formats as well.

We can also simply bump the autotune time limit for all unsalted hashes e.g. to 500ms. This is always acceptable for unsalted, right?

claudioandre-br:
200ms is our safe number.

solardiz:
Why is 200ms safe and 500ms unsafe for unsalted? Driver timeout issues?

magnumripper:
Some ancient AMD cards or runtime would crash at >200 ms but that might not be worth considering anymore

magnumripper:
Generally I’d like to have low limits because most good formats really don’t need high figures, so it simply speeds up autotune a whole lot especially when testing all formats at once

PR #3839, opened 2019-04-18
bcrypt >512 hashes/salt fix

Apingis:
- Corrected error handling
- Works fine with mask iterating over lengths
- Tested on a salt with 3M hashes (performance almost equals to a run with 513 hashes , memory usage ~700M)

### DIFF
--- a/src/ztex/device.c
+++ b/src/ztex/device.c
@@ -27,6 +27,8 @@
 
 #define CONFIG_MAX_LEN	256
 
+// boards are in "no comparison" mode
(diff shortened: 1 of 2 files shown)


solardiz:
CircleCI failed with the usual thing, which I'll ignore:
```
Testing: crypt, generic crypt(3) [?/32]... Warning: hash encoding string length 13, type id #1
appears to be unsupported on this system; will not load such hashes.
FAILED (valid (before init))

1 out of 711 tests have FAILED
```

Issue #3830, opened 2019-04-16
SIP fails cracking qop=auth


nanosonde:
Has a SIP format with qop=auth ever been tested? 
I do not see any test vectors in the code that's why I am asking.

I have the following SIP trace from a REGISTER sip:sip.provider.com SIP/2.0 message:
```
Authorization: Digest username="155350609596347633",realm="sip.provider.com",nonce="7dfb229f37d9834084f1aa5cdd1e2460",uri="sip:sip.provider.com",qop=auth,nc=00000147,cnonce="547401db",response="e1facc132ebe3973521a83950ff54994",algorithm=MD5
```

Username is: 155350609596347633
Password is: abc
Method: REGISTER
EntityBody: <empty>

I am **able** to crack the password using the tool "HTTPBrute" with the above parameters, but **not** using JtR.

![grafik](https://user-images.githubusercontent.com/2073569/56219660-6db53180-6067-11e9-89af-c6cbc13325b8.png)

solardiz:
I don't know the answer, but I think it'd help if you post the exact input to JtR you were testing with. Thanks!

nanosonde:
This is the file that sipdump created from manually entering the above details: 
```
1.2.3.4"5.6.7.8"155350609596347633"sip.provider.com"REGISTER"sip:sip.provider.com"7dfb229f37d9834084f1aa5cdd1e2460"547401db"00000147"auth"MD5"e1facc132ebe3973521a83950ff54994
```
Note: That the two IP addresses (client/server) 1.2.3.4 and 5.6.7.8 aare not used for calculations.
I guess they are just for documentation purposes.

Running the sipdump2john.py on this file generated this output for use with JtR:
```
1.2.3.4-5.6.7.8:$sip$*1.2.3.4*5.6.7.8*155350609596347633*sip.provider.com*REGISTER*sip*sip.provider.com**7dfb229f37d9834084f1aa5cdd1e2460*547401db*00000147*auth*MD5*e1facc132ebe3973521a83950ff54994
```
Running JtR with this file ("./john sip.login.john") does even find the password from the wordlist file where the password "abc" is also included.

solardiz:
I think we have enough info here to proceed to look into this issue when one of us - perhaps @kholia - is available for that. Thanks!

nanosonde:
Here is a patch for pass_gen.pl which add "sip_qop" to generate test vectors for this auth type.

```
[...]

PR #3840, opened 2019-04-18
SIP: fix for fails cracking qop=auth

claudioandre-br:
And adds a test vector that is qop=auth. By nanosonde.


solardiz:
As I understand, this fixes #3830

PR #3843, opened 2019-04-18
mask: add event to free (destroy) mask objects

claudioandre-br:
To be called after *_done(). It seems, that in mask the done() method
is not only a cleanup; it has some pending tasks to do. The cleanup
has to be done later.

### DIFF
--- a/src/bench.c
+++ b/src/bench.c
@@ -962,8 +962,10 @@ next:
 		ldr_free_test_db(test_db);
 		fmt_done(format);
 #ifndef BENCH_BUILD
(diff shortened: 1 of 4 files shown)


claudioandre-br:
An event to handle the cleanup after an event that finishes pending tasks is not a bad idea for modes.
* seems to much for a format;
* but, for modes it should be a good useful.

Anyway, this is only a workaround (and it needs ztex testing).

Apingis:
I tested with ZTEX and it seem displays as it should do.

PR #3846, opened 2019-04-20
Upstream packaging stuff

claudioandre-br:
Magnum, to be part of a repository, a reviewer suggest to upstream these files. So, that is up to you. 

My opinion is:
* `build.sh`: no. It is not related to upstream at al.
* `appdata.xml`: maybe, why not.
* `.destop`: uhmmmm, useless to upstream. Anyway, harmless.

**[edited]**
https://flathub.org/builds/#/builders/32/builds/2621

### DIFF
--- a/src/packaging/build.sh
+++ b/src/packaging/build.sh
@@ -1,0 +1,55 @@
+######################################################################
+# Copyright (c) 2019 Claudio André <claudioandre.br at gmail.com>
+#
+# This program comes with ABSOLUTELY NO WARRANTY; express or implied.
+#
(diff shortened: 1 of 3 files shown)


claudioandre-br:
BTW: 
* `appdata.xml` [1] is a source of information about an OSS. So, keep it alive and updated is a good idea.
* since it has links to upstream, the idea to migrate to `github.com/openwall` just after J1 (to fix the links during the release), if ever done, is not a bad idea.

[1] I guess `appdata.xml` will succeed because it is you talking about your tool (not a bored distro maintainer).

Issue #3849, opened 2019-04-20
ocl_hc_128_prepare_table uninitialized allocation


crioux:
### Issue ###
ocl_hc_128_prepare_table has an uninitialized table, allocated on line 42: 

```hash_ids = (cl_uint*) mem_alloc((3 * num_loaded_hashes + 1) * sizeof(cl_uint));```

The fix is this:

```hash_ids = (cl_uint*) mem_calloc((3 * num_loaded_hashes + 1), sizeof(cl_uint));```

(notable this code is fixed in `ocl_hc_128_prepare_table_test`, but not in `ocl_hc_128_prepare_table`)

This uninitialized hash id table results in intermittent cracking failures for nt-opencl hashes.

### Steps to reproduce ###
Run any self test with nt-opencl format. Intermittent failures depending on compiler defaults and system allocation strategy result.

### System configuration ###
Windows 64 bit, JtR from L0phtCrack integration

**Comment by solardiz** [comment]
I think we might have more instances of this bug. Let's review.

```
$ grep 'hash_ids.*_alloc.*num_loaded_hashes' *.c
opencl_krb5pa-md5_fmt_plug.c:   hash_ids = (cl_uint*) mem_alloc((3 * max_num_loaded_hashes + 1) * sizeof(cl_uint));
opencl_mscash_fmt_plug.c:       hash_ids = (cl_uint*) mem_alloc((3 * max_num_loaded_hashes + 1) * sizeof(cl_uint));
opencl_mysqlsha1_fmt_plug.c:    hash_ids = (cl_uint*) mem_alloc((3 * num_loaded_hashes + 1) * sizeof(cl_uint));
opencl_mysqlsha1_fmt_plug.c:            hash_ids = (cl_uint*)mem_alloc((3 * num_loaded_hashes + 1) * sizeof(cl_uint));
opencl_ntlmv2_fmt_plug.c:       hash_ids = (cl_uint*) mem_alloc((3 * max_num_loaded_hashes + 1) * sizeof(cl_uint));
opencl_rawsha1_fmt_plug.c:      hash_ids = (cl_uint*) mem_alloc((3 * num_loaded_hashes + 1) * sizeof(cl_uint));
opencl_rawsha1_fmt_plug.c:              hash_ids = (cl_uint*)mem_alloc((3 * num_loaded_hashes + 1) * sizeof(cl_uint));
[...]

PR #3850, opened 2019-04-20
hash_table: ocl_hc_128_prepare_table uninitialized allocation

claudioandre-br:
Patch by crioux. See #3849.


solardiz:
BTW, I think we should figure out and add copyright+license for opencl_hash_check_128_plug.c.

PR #3856, opened 2019-04-21
fix buffer overflow in HC rule p (duplicate word N times) cutting length

### LINKED ISSUE #3852: buffer overflow in hc_logic's 'p' rule
`p9` and other `p` + digit rules do not check length and copy candidate after end of buffer too, overflowing into 'memory' buffer and maybe further.

Example where the problem is not visible:
```
$ python -c 'print "a" * 125' | john --pipe --stdout --rules=': r u p9 X010'
aAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
```

The problem may be observable:
```
$ python -c 'print "a" * 125' | john --pipe --stdout --rules=': M r u p9 X010'
AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
```

`M` forces use of the 'memory' buffer. In both cases, there should be letter 'a', but it is overwritten by uppercase 'A' due to overflow. That's observable in the second example when `X010` inserts from the 'memory' buffer into 'out' buffer.

Rule `u` is used to differentiate original data and generated data. Rule `r` is needed to choose internal buffer closer to the 'memory'buffer.

```
$ john --list=build-info
Version: 1.9.0-jumbo-1-bleeding-4bf9652 2019-04-19 19:08:24 +0200
```

AlekseyCherepanov:
closes #3852

### DIFF
--- a/src/rules.c
+++ b/src/rules.c
@@ -890,8 +890,8 @@ char *rules_apply(char *word_in, char *rule, int split, char *last)
 				/* HC rule: duplicate word N times */
 				unsigned char x, y;
 				POSITION(x)
-				if (x * length > RULE_WORD_SIZE && length)
-					x = RULE_WORD_SIZE / length;
+				if (x * length > RULE_WORD_SIZE - 1)
+					x = (RULE_WORD_SIZE - 1) / length;
 				y = x;
 				in[length*(x + 1)] = 0;
 				while (x) {
--- a/src/rules.c
+++ b/src/rules.c
@@ -890,6 +890,8 @@ char *rules_apply(char *word_in, char *rule, int split, char *last)
 				/* HC rule: duplicate word N times */
 				unsigned char x, y;
 				POSITION(x)
+				if (x * length > RULE_WORD_SIZE && length)
+					x = RULE_WORD_SIZE / length;
 				y = x;
 				in[length*(x + 1)] = 0;
 				while (x) {


src/rules.c
```diff
@@ -890,6 +890,8 @@ char *rules_apply(char *word_in, char *rule, int split, char *last)
 				/* HC rule: duplicate word N times */
 				unsigned char x, y;
 				POSITION(x)
+				if (x * length > RULE_WORD_SIZE && length)
+					x = RULE_WORD_SIZE / length;
 				y = x;
 				in[length*(x + 1)] = 0;
```
solardiz: I think the added two lines should use `RULE_WORD_SIZE - 1` in both places, drop the redundant check of `length` (the first condition implies `length` is non-zero), and perhaps use `x + 1` and subtract one in the assignment to `x`.

solardiz:
Aleksey, have you tested this in an `--enable-asan` build, along with your testcases? Please do.

solardiz:
It would also be good to see what hashcat does in case this would overflow and try to match its behavior.

AlekseyCherepanov:
`--enable-asan` does not catch overflow from 'buffer' to 'memory' in 'rules_data'. Separate declarations are needed to trigger asan.

I did not test with separate declarations of buffers.

Full length is `(x + 1) * length`. So the check searches full blocks after end, so we would not copy them at all. It does not try to fit into `RULE_WORD_SIZE` cleanly (the buffer has length `RULE_WORD_SIZE * 2`).

I don't have time to check hashcat.

AlekseyCherepanov:
I am sorry. I prepared the change yesterday, did `git commit --amend ...` and thought everything was ok, but I did not do `git add`, so my change was not in the commit.

Thanks!

PR #3860, opened 2019-04-23
Move regex to experimental

claudioandre-br:


magnumripper:
I already changed it to require —enable-rexgen, is that not enough? I’ll let you decide @solardiz @claudioandre-br

magnumripper:
Ok I see now, this is cosmetic

PR #3862, opened 2019-04-23
radius: just output some recovered passwords when we can

### LINKED ISSUE #3809: Deal with our radius format weirdness
Our "radius" format reports much higher speeds for "Only one salt" than for "Many salts". This prompted me to take a look at its code, and here's what I found:

The format implements two different algorithms, and there are only two test vectors - one per algorithm.

The first test vector corresponds to an algorithm that doesn't even require cracking - instead of XOR'ing and comparing, we could have XOR'ed the other way around and printed the plaintext password. Indeed, the XOR'ing is also extremely fast, even if ridiculous to do like we do it now.

A question is whether that algorithm even fits JtR, or should be implemented elsewhere. If in JtR, then maybe we can have `valid` figure out the plaintext passwords and print them, but that wouldn't get them into `john.pot`. Maybe we need a formats interface extension where the format itself could suggest candidate passwords or explicitly provide cracked passwords.

As it is, our implementation of this algorithm is similar to what we do in the "dummy" and "plaintext" formats. Perhaps we should at least document that in a comment in the "radius" format.

The second test vector corresponds to an algorithm that requires computation of MD5 per candidate password. It is thus not as fast. Ideally, our first two test vectors would correspond to this algorithm, so that we'd have proper "Many salts" benchmark for it.

As a first step at fixing this issue, I'd like a second test vector (with a different salt) added for the algorithm that does require MD5 per-candidate, and those two test vectors specified at the beginning of the `tests` array. Dhiru, can you take care of this now please?

solardiz:
radius: just output some recovered passwords when we can, and don't benchmark "cracking" of directly recoverable passwords. Fixes #3809.

### DIFF
--- a/src/radius_fmt_plug.c
+++ b/src/radius_fmt_plug.c
@@ -58,10 +58,12 @@ john_register_one(&fmt_radius);
 #endif
 
 static struct fmt_tests tests[] = {
-	// tshark -q -Xlua_script:network2john.lua -X lua_script1:password2 -X lua_script1:1 -r freeradius.pcap
-	{"$radius$1*1*password2*acf627289be7e214c8e328d0d01fcea2*fe18a76eec4abeeea09eb90e99b359f7", "Passw0rd2"},
 	// tshark -q -Xlua_script:network2john.lua -X lua_script1:Passw0rd2 -X lua_script1:0 -r freeradius.pcap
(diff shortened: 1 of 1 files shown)


magnumripper:
I found a cool way to do this: For every mode 1 password we decode, copy it to a special `recovered_key[index]` that is otherwise empty (we can't modify the real `saved_key` in case some other salt are mode 0 and needs the real candidate.).

In get_key, return `recovered_key` if not empty, otherwise the real `saved_key`. 

No matter what the candidate word is (such as `--mask=a`), every single password will be "cracked" and put in `john.pot` immediately.

PR #3864, opened 2019-04-24
Update geli_common_plug.c

trounce1:
Modified code to allow md_aalgo of 256.

I tested this on my hash and it worked fine.


solardiz:
Thanks, @trounce1! Can you also contribute a corresponding test vector, to the `geli_tests` array?

While at it, we could also look into using the two commented-out yet claimed more realistic test vectors for our benchmarks (make them the first two) if that's affordable at all. And we could add `static` to most or all functions, as I think they're local to this source file.

solardiz:
@trounce1 Oh. Can you simply add the new test vector to the end of the array, and keep its previous content as-is? We'll make further adjustments ourselves, since this requires knowledge about how our benchmarks work. Thanks!

solardiz:
Meanwhile, the ASan-enabled build errored out with:

```
Testing: geli, FreeBSD GELI [PBKDF2-SHA512 256/256 AVX2 8x]... (2xOMP) 
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

So I guess those two test vectors were in fact too heavy.

trounce1:
i think this may have timed out due to my itteration count. i will set up another with lower itteration count later.

solardiz:
Yes, your newly added test vector is also a possible cause.

trounce1:
> Yes, your newly added test vector is also a possible cause.

i will update it later with smaller count and remove the other test vectors again. 

i dont think using a larger iteration count should affect the integrity of the opencl kernel

solardiz:
BTW, it looks like we don't actually use the value of the md_aalgo field anywhere except in `valid`.

trounce1:
I think i have made a mistake. i was reading through the source and bsd geom source and the 128 or 256 value actually refers to the md_keylen variable.

claudioandre-br:
Seems ok;
- test-full, asan, ...

PR #3865, opened 2019-04-25
Drop non-updated jumbo-specific documentation files

claudioandre-br:
I'm not the best person to deal with docs.
- I removed non-updated stuff;
- and move advanced stuff to an 'extra' folder


solardiz:
Perhaps we should fully remove `README-CUDA` rather than move it to a subdirectory, because we've completely dropped CUDA support.

claudioandre-br:
It is a possibility. But the file says it was dropped on aa3602b (the information might be useful for someone looking for experimenting using CUDA stuff).

claudioandre-br:
This sounds outdated to me. If you was born A.D., you must be running any LTS software.
```
When running on Linux distributions with glibc 2.7+, John 1.7.6+ additionally supports (and autodetects) SHA-crypt hashes (which are actually used by recent versions of Fedora and Ubuntu), with optional OpenMP parallelization (requires GCC 4.2+, needs to be explicitly enabled at compile-time by uncommenting the proper OMPFLAGS line near the beginning of the Makefile).

Similarly, when running on recent versions of Solaris, John 1.7.6+ supports and autodetects SHA-crypt and SunMD5 hashes, also with optional OpenMP parallelization (requires GCC 4.2+ or recent Sun Studio, needs to be explicitly enabled at compile-time by uncommenting the proper OMPFLAGS line near the beginning of the Makefile and at runtime by setting the OMP_NUM_THREADS environment variable to the desired number of threads).

John the Ripper Pro adds support for Windows NTLM (MD4-based) and Mac OS X 10.4+ salted SHA-1 hashes.
```

Again. Only SSE2? If so, are they useful?
```
Additionally, there are assembly language routines for several processor architectures, most importantly for x86-64 and x86 with SSE2.
```

solardiz:
The "When running ..." and "Similarly ..." paragraphs are still in core's README, but the ones on "John the Ripper Pro adds ..." and "Additionally, there are assembly language routines ..." (in fact, the whole sub-section on performance) I've dropped from core's README shortly before the 1.9.0 release - so at least those should also be dropped from jumbo's README.md. I suggest diff'ing against core's current README and keeping only the intentional and currently relevant changes.

Issue #3810, opened 2019-04-13
-test -mask segfaults with Benchmarks_1_8


solardiz:
Similarly to #3808, there's a write to test vectors in `bench.c`:

```c
                if (options.flags & FLG_MASK_CHK) {
                        plaintext[mask_key_len] = 0;
```

As a quick fix, maybe we should just check for this combinations of settings and refuse to run with a reasonable message, rather than segfault.

magnumripper:
Switch to a copy in a ‘static char[]’ right before this is just as quick a fix

solardiz:
While fixing this issue is within consideration for 1.9.0-jumbo-1, it is _not_ a blocker for the release because it's only triggered in an obscure situation that our users shouldn't face (it's relevant to us, not to them) and the only impact is the process crashing instead of e.g. it printing a message and terminating gracefully or ignoring a setting. Not a big deal, just unclean.

PR #3870, opened 2019-04-28
Workaround for -test -mask segfaults with Benchmarks_1_8

claudioandre-br:
Fixes: #3810.

### DIFF
--- a/src/bench.c
+++ b/src/bench.c
@@ -286,11 +286,11 @@ static void bench_set_keys(struct fmt_main *format,
 
 	/* Legacy benchmark mode for performance regression testing */
 	for (index = 0; index < max; index++) {
-		char *plaintext;
+		static char plaintext[PLAINTEXT_BUFFER_SIZE];
 		do {
 			if (!current->ciphertext)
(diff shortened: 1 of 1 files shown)


solardiz:
Thanks, Claudio!

Two issues here:

1. `strncpy` isn't the right function to use here. Its NUL padding is unnecessary, but we require NUL termination, which it doesn't guarantee. Should use e.g. our `strnzcpy`.

2. This change affects benchmark results, especially for "Only one salt" benchmarks of fast hashes. Perhaps we can come up with an alternative that doesn't touch the normally taken code path (at least not to an extent greater than a single extra `if` check), and only deals with the `-test -mask` issue.

solardiz:
I would actually be fine with _not_ supporting legacy benchmarks along with `-mask`. There's simply no need for that combination. So perhaps we should simply detect that condition in some outer layer (not in a performance-critical loop) and print a warning and disable the legacy benchmarks mode if so. While at it, it would also be nice to print a message when legacy benchmarks _are_ in use.

PR #3889, opened 2019-05-05
Fix AVX512BW detection

### LINKED ISSUE #3888: AVX512BW detection is buggy
This works as expected:
``` 
./configure # OR ./configure CPPFLAGS='-mavx512bw'
[...]
Target CPU ................................. x86_64 AVX512BW, 64-bit LE
```

But this fails:
```
./configure --disable-native-tests CPPFLAGS='-mavx512bw'
[...]
checking for MMX... yes
checking for SSE2... yes
checking for SSSE3... yes
checking for SSE4.1... yes
checking for SSE4.2... yes
checking for AVX... yes
checking for XOP... no
checking for AVX2... yes
checking for AVX512F... yes
checking for AVX512BW... no
[...]
Target CPU ................................. x86_64 AVX512F, 64-bit LE
```

```
configure:11595: checking for AVX512BW
configure:11607: gcc -o conftest -m64 -g -O2 -I/usr/local/include   -O0 -mavx512bw -I/usr/local/include -m64  -L/usr/local/lib -L/usr/local/lib64 -L/usr/lib64 -L/lib64 conftest.c  >&5
conftest.c: In function 'main':
conftest.c:30:9: error: can't convert a vector of type '__m512i' {aka '__vector(8) long long int'} to type 'int' which has different size
         int main(){__m512i t, t1;*((long long*)&t)=1;t1=t;t=_mm512_slli_epi16(t1,t);if((*(long long*)&t)==88)printf(".");exit(0);}
         ^~~
configure:11607: $? = 1
configure: failed program was:
```

claudioandre-br:
Closes #3888

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -11600,7 +11600,7 @@ $as_echo_n "checking for AVX512BW... " >&6; }
     #include <immintrin.h>
         #include <stdio.h>
         extern void exit(int);
-        int main(){__m512i t, t1;*((long long*)&t)=1;t1=t;t=_mm512_slli_epi16(t1,t);if((*(long long*)&t)==88)printf(".");exit(0);}
+        int main(){__m512i t, t1;*((long long*)&t)=1;t1=t;t=_mm512_slli_epi32(t1,t);if((*(long long*)&t)==88)printf(".");exit(0);}
 
 
 _ACEOF
--- a/src/m4/jtr_x86_logic.m4
+++ b/src/m4/jtr_x86_logic.m4
@@ -469,7 +469,7 @@ dnl ======================================================================
       [[#include <immintrin.h>
         #include <stdio.h>
         extern void exit(int);
-        int main(){__m512i t, t1;*((long long*)&t)=1;t1=t;t=_mm512_slli_epi16(t1,t);if((*(long long*)&t)==88)printf(".");exit(0);}]]
+        int main(){__m512i t, t1;*((long long*)&t)=1;t1=t;t=_mm512_slli_epi32(t1,t);if((*(long long*)&t)==88)printf(".");exit(0);}]]
     )]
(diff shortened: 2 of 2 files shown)


solardiz:
Hmm, is `_mm512_slli_epi32` unavailable with plain AVX512F? This feels weird to me if so. In fact, I am using it in DES_bs_b.c when building for AVX512F and this builds and works on KNL fine. So I expected that after this PR has been merged ./configure would misdetect AVX512BW on KNL, but this isn't the case, I'm still only getting:

```
Target CPU ................................. x86_64 AVX512F, 64-bit LE
```

```
$ fgrep -r _mm512_slli_epi32 .
./DES_bs_b.c:   (dst) = _mm512_slli_epi32((src), (shift))
./configure:        int main(){__m512i t, t1;*((long long*)&t)=1;t1=t;t=_mm512_slli_epi32(t1,t);if((*(long long*)&t)==88)printf(".");exit(0);}
./m4/jtr_x86_logic.m4:        int main(){__m512i t, t1;*((long long*)&t)=1;t1=t;t=_mm512_slli_epi32(t1,t);if((*(long long*)&t)==88)printf(".");exit(0);}]]
./pseudo_intrinsics.h:#define vslli_epi32             _mm512_slli_epi32
```

I'm puzzled.

solardiz:
For me, trying to build our current AVX512BW test after Claudio's fix here results in the same error that Claudio reported in #3888.

I think this test is just wrong in multiple ways, and was less wrong prior to this PR (it actually checked for an AVX512BW feature, or at least tried to).

solardiz:
BTW, can we please avoid typecasts in those tests? They're a recipe for disaster, and should be fairly easy to avoid here I think - we can use the proper `*set*` intrinsics, or `union` types.

PR #3891, opened 2019-05-06
Fix AVX512BW detection

### LINKED ISSUE #3888: AVX512BW detection is buggy

claudioandre-br:
Closes #3888

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -11654,7 +11654,7 @@ $as_echo_n "checking for AVX512BW... " >&6; }
     #include <immintrin.h>
         #include <stdio.h>
         extern void exit(int);
-        int main(){__m512i t, t1;*((long long*)&t)=1;t1=t;t=_mm512_slli_epi32(t1,t);if((*(long long*)&t)==88)printf(".");exit(0);}
+        int main(){__m512i t=_mm512_slli_epi16(_mm512_set1_epi16(1),1);exit(!(_mm_cvtsi128_si64x(_mm512_extracti32x4_epi32(t,0))==0x2000200020002ULL));}
 
 
 _ACEOF
--- a/src/m4/jtr_x86_logic.m4
+++ b/src/m4/jtr_x86_logic.m4
@@ -469,7 +469,7 @@ dnl ======================================================================
       [[#include <immintrin.h>
         #include <stdio.h>
         extern void exit(int);
-        int main(){__m512i t, t1;*((long long*)&t)=1;t1=t;t=_mm512_slli_epi32(t1,t);if((*(long long*)&t)==88)printf(".");exit(0);}]]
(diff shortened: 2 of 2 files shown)


claudioandre-br:
It works as it is now. Anyway: 
* the if is not necessary, we are testing only the compilation process;
* we can't run the binary (it is for a cross build);
* but, if we do, the result will be 0x2000...

solardiz:
>     * we can't run the binary (it is for a cross build)

Is this really only for a cross build? What do we use to detect AVX512BW support by the hardware when we build on a machine actually supporting this instruction set? I'm confused.

claudioandre-br:
For non cross is JtR `CPU_detect()`

solardiz:
> For non cross is JtR `CPU_detect()`

Ah, yes, IIRC we discussed that on some issue a year or so ago and magnum made the change to reuse the `CPU_detect` code during build. That was a very welcome change.

magnumripper:
Right. Before that, we had two version of almost same tests (like the one now changed). One version tried to build and run it, the other (for cross-compile) just tries to build and link it and if that works, the test passes.

What I don't understand though, is what was wrong with the original version, prior to e3e1f1d0? I did use the correct instruction, then you changed it in e3e1f1d0 and now you changed it back.

claudioandre-br:
It is `_mm512_slli_epi16(vector_type, **scalar_type**)`. What was failing is related to `scalar_type`.

magnumripper:
I see. So much for dry-running code :-)

PR #3892, opened 2019-05-06
Run some tests only after merge

claudioandre-br:
- It is my way of getting more tests without overwhelming Travis;


claudioandre-br:
Fell free to re-enable CI, review, etc, before being able to merge.

PR #3894, opened 2019-05-06
Bail out with message if running legacy benchmarks with --mask

magnumripper:


solardiz:
This broke  "make -f Makefile.legacy generic" again, ironically on the same day that you committed a fix for another change that had also broken it. I'll send a PR.

PR #3898, opened 2019-05-07
Pass correct salt into crypt_all() in benchmarks

### LINKED ISSUE #3897: Benchmarks should pass correct salt into crypt_all()
@magnumripper I just found out that your commit 01b384e7afebff065838f0cbfebc56ab75b24e15 broke `-test` for bcrypt-ztex, in two ways: with `-mask` I am now getting timeouts, and without `-mask` I am getting poor speeds (~60k c/s for 4 boards, whereas previously we were getting ~400k c/s in that test). I think @Apingis isn't currently available to help troubleshoot this. I'd appreciate your help.

So far, I reviewed the test vectors. The first two are cost 5, which is what we intend to benchmark, but because of the issue being fixed with that commit I'm not it's what we were actually benchmarking (even though the speeds suggested we were). The lowest cost test vectors are cost 4, which isn't that much lower - but I've tried moving those to the start of the `tests` array, which doesn't appear to have made any difference (with the above commit still present, I am still getting those same timeouts and poor speeds).

I think I need better understanding of what exactly changed with that commit - what test vectors were were using before and what we're using now.

solardiz:
Fixes #3897

### DIFF
--- a/src/bench.c
+++ b/src/bench.c
@@ -361,6 +361,7 @@ char *benchmark_format(struct fmt_main *format, int salts,
 	void *salt, *two_salts[2];
 	int index, max;
 #ifndef BENCH_BUILD
+	struct db_salt *two_salts_db[2];
 	unsigned int t_cost[2][FMT_TUNABLE_COSTS];
 	int ntests, pruned, i;
 #endif
@@ -458,8 +459,20 @@ char *benchmark_format(struct fmt_main *format, int salts,
 		     format->methods.tunable_cost_value[i] != NULL; i++)
 			t_cost[index][i] =
 				format->methods.tunable_cost_value[i](salt);
+
+		struct db_salt *dbsalt = test_db->salts;
+		if (format->params.salt_size)
+		while (dbsalt && dyna_salt_cmp(dbsalt->salt, two_salts[index], format->params.salt_size))
+			dbsalt = dbsalt->next;
+		if (dbsalt) {
+			two_salts_db[index] = dbsalt;
+		} else {
+			puts("Warning: Could not find salt in db");
+			two_salts_db[index] = test_db->salts;
+		}
 #endif
 	}
+
 	/*
 	 * Core john doesn't have this set_salt at all, only later under
 	 * "if (salts > 1)". I first added it only "if (salts == 1)" but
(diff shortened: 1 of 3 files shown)


solardiz:
I don't understand why formats.c uses `dyna_salt_cmp` instead of simply `memcmp` in similar context. So I simply use `memcmp`.

```c
                if ((dbsalt = db->salts))
                do {
                        if (!dyna_salt_cmp(salt, dbsalt->salt,
                                           format->params.salt_size))
                                break;
                } while ((dbsalt = dbsalt->next));
```

solardiz:
Some non-hash formats triggered my "Warning: Could not find salt in db" when testing on CircleCI. We need to look into that and address this.

Dynamic formats passed tests without that warning.

solardiz:
I'm now suppressing the warning for formats with `FMT_DYNA_SALT`.

magnumripper:
> I don't understand why formats.c uses `dyna_salt_cmp` instead of simply `memcmp` in similar context. So I simply use `memcmp`.
> 
> ```c
>                 if ((dbsalt = db->salts))
>                 do {
>                         if (!dyna_salt_cmp(salt, dbsalt->salt,
>                                            format->params.salt_size))
>                                 break;
>                 } while ((dbsalt = dbsalt->next));
> ```

I don't remember all details off the top of my head but I'm pretty sure we really **should** use `dyna_salt_cmp()` For "normal" salts, a plain `memcmp()` is what it will in up as anyway but for actual dynamic salts, your memcmp will most likely compare irrelevant stuff like pointers.

solardiz:
@magnumripper So you suggest that instead of suppressing the warning for `FMT_DYNA_SALT`, we use `dyna_salt_cmp`?

solardiz:
Meanwhile, a CircleCI run completed without warnings.

magnumripper:
> @magnumripper So you suggest that instead of suppressing the warning for `FMT_DYNA_SALT`, we use `dyna_salt_cmp`?

I'm not sure those warnings are related. But simply doing memcmp with dynamic salts is kinda like comparing the pointers to two strings instead of the strings themselves. The result is useless. However, I haven't got all context around this code, *perhaps* it's fine in this very case. Using dyna_salt_cmp is never wrong though.

solardiz:
OK, switched to using `dyna_salt_cmp` and never suppressing the warning. Let's see.

solardiz:
No warnings in CircleCI's wine-64bits test. I didn't check the rest.

magnumripper:
> Dynamic formats passed tests without that warning.

Dynamic salts is totally unrelated to dynamic format(s). The former is dynamic-size-salts.

magnumripper:
I also looked for warnings but saw nothing. I saw a dmg2john warning though, issue created.

solardiz:
Have we retested OpenCL benchmarks, with and without mask, after having merged this PR? I only tested CPU and ZTEX benchmarks, not OpenCL.

PR #3900, opened 2019-05-07
[skip ci] Add usage examples to doc/MASK

solardiz:
I reuse my john-users posting for this. Ideally, we should also revise the preceding text, but I have no time.


solardiz:
Hmm, how is that "[skip ci]" thing Claudio was using supposed to be used?

claudioandre-br:
> how is that "[skip ci]" thing Claudio was using supposed to be used?

The way you used it. The point here is that magnum requires an ok from CI, so, Travis is waiting one (this is a Travis - itself - issue).

That said, you are sparing resources and that is the main goal.

magnumripper:
I'll throw it in for now, it's not too bad vs. the preceding text. I'll revise it again if I get time

PR #3901, opened 2019-05-07
Botcheck

magnumripper:


magnumripper:
The external.c stuff is tricky, there are some catch-22's I need to sort out post release. This should do for now, tested with both --stdout and with real format.

PR #3904, opened 2019-05-07
Avoid a problem casting size_t to negative and then to off_t

### LINKED ISSUE #3902: dmg2john Warning from CircleCI Wine64
I've never seen this before.
```
dmg2john.c: In function 'hash_plugin_parse_hash':
dmg2john.c:252:17: warning: overflow in conversion from 'long long unsigned int' to 'long int' changes value from '18446744073709550340' to '-1276' [-Woverflow]
   if (lseek(fd, -sizeof(cencrypted_v1_header), SEEK_END) < 0) {
                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

magnumripper:
...when types are different sizes (seen on Win64). Closes #3902

### DIFF
--- a/src/dmg2john.c
+++ b/src/dmg2john.c
@@ -249,7 +249,7 @@ static void hash_plugin_parse_hash(char *in_filepath)
 
 	// fprintf(stderr, "Header version %d detected\n", headerver);
 	if (headerver == 1) {
-		if (lseek(fd, 0 - (off_t)sizeof(cencrypted_v1_header), SEEK_END) < 0) {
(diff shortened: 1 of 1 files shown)


solardiz:
FWIW, I think `-(off_t)sizeof(...)` or `-(int)sizeof(...)` would be cleaner than the `0 - ...` approach as the latter involves subtraction between these different signed types.

magnumripper:
> FWIW, I think `-(off_t)sizeof(...)` or `-(int)sizeof(...)` would be cleaner than the `0 - ...` approach as the latter involves subtraction between these different signed types.

Thanks, I'm not sure I understand why (especially under `-O2`) but I changed it and will squash later.

Issue #3872, opened 2019-04-29
Repurpose aix-smd5 format


solardiz:
Dhiru created the aix-smd5 format in 2013 to support AIX's variation of md5crypt. We've since added support for those hashes to our main md5crypt and md5crypt-opencl and even md5crypt-ztex formats.

Presently, the only value the aix-smd5 format provides is it supporting passwords of arbitrary length (up to JtR's global limit), whereas our other md5crypt formats are length-limited (to 15, except md5crypt-ztex which allows up to 64, but requires the rare hardware).

We might want to rename and repurpose the aix-smd5 format as e.g. md5crypt-long.

magnumripper:
Agreed. On the other hand I'd like to see support for >15 (obviously only when needed) in the optimized md5crypt format as well, but I haven't groked all optimizations.

solardiz:
Right. For the upcoming release renaming this format is within scope and would reduce further confusion and simplify further `relbench`, while making invasive changes to our optimized format is not.

solardiz:
Along with the rename, it's probably also easy enough to enhance this format to:

1. Support apr1 hashes.
2. Avoid modulo division - have counter(s) increase up to the divisor(s), then reset to 0.

solardiz:
Can we rename this format's source file such that it's benchmarked right after our main md5crypt format? Or would that perhaps require a john.c edit, since the main md5crypt format comes from core and isn't a plugin format? Either way, I think ideally we'd have it benchmarked right at that time, so that users learn of its existence and of its relative speed (~7x slower on AVX build).

solardiz:
We could also have this format count how many of the tested passwords, say, in the first 1 million (then disable the counting) are up to 15 chars vs. 16+, and report those statistics at the 1 million mark if the percentage of the up to 15 chars passwords is significant (say, 1% or above), suggesting use of the main md5crypt format on those shorter candidate passwords.

magnumripper:
Instead of spending time on such bodges, how long would it take you to add support for longer passwords to the SIMD format? And what length would "next" optimization be, 30?

solardiz:
> how long would it take you to add support for longer passwords to the SIMD format?

[...]

PR #3909, opened 2019-05-09
Refactor and repurpose aix-smd5 format.

magnumripper:
See #3872.

This commit unplugs the format, renames it to md5crypt-long and adds support for $apr1$.

### DIFF
--- a/src/MD5_fmt.c
+++ b/src/MD5_fmt.c
@@ -28,7 +28,7 @@
 #endif
 
 #define FORMAT_LABEL			"md5crypt"
-#define FORMAT_NAME			"crypt(3) $1$"
+#define FORMAT_NAME			"crypt(3) $1$ (and variants)"
 
 #define BENCHMARK_COMMENT		""
 #define BENCHMARK_LENGTH		7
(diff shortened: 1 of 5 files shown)


src/MD5_fmt.c
```diff
@@ -28,7 +28,7 @@
 #endif
 
 #define FORMAT_LABEL			"md5crypt"
-#define FORMAT_NAME			"crypt(3) $1$"
+#define FORMAT_NAME			"crypt(3) $1$ (and variants)"
```
solardiz: Does this need a `benchmark-unify` update, or does `relbench` rely purely on `FORMAT_LABEL`? I don't recall.

src/md5crypt_long_fmt.c
```diff
@@ -363,14 +394,14 @@ struct fmt_main fmt_smd5 = {
 		MAX_KEYS_PER_CRYPT,
 		FMT_CASE | FMT_8_BIT | FMT_OMP,
 		{ NULL },
-		{ FORMAT_TAG, FORMAT_TAG1 },
+		{ FORMAT_TAG, FORMAT_TAG1, FORMAT_TAG2 },
 		smd5_tests
```
solardiz: I'd also rename `fmt_smd5`, `smd5_tests`, and `smd5_set_key`. But I understand not wanting to put more effort into this now.

PR #3914, opened 2019-05-10
Fix "make -f Makefile.legacy generic" again, was broken by #3894

solardiz:


solardiz:
Should we possibly test "make -f Makefile.legacy generic" on CI?

magnumripper:
> Should we possibly test "make -f Makefile.legacy generic" on CI?

We did, that's why I usually don't care. Maybe Claudio's recent changes put then under "more tests"?

solardiz:
What's meant by "more tests"? Are they not required for merge?

magnumripper:
Not sure. They are not required right now and it seems they won't run unless we do something. @claudioandre-br?

magnumripper:
It seems I could click "details" and start those builds.

BTW I just turned off the requirement of review before merge again, for now. Not because we should skip it but because you should be able to merge without me present.

magnumripper:
IIRC before Claudio's (very good) changes to CI we had some CI produce a MinGW build and I believe some or all of them were legacy builds. Not sure where they are now, if anywhere.

magnumripper:
On a side note we should get this repo owned by Openwall as an "organization" instead of me as a single user. That way we can have several administrators.

magnumripper:
On another side note I think I'm definitely dropping the support for legacy Makefile and all silly `BENCH_BUILD` bodges after this release.

magnumripper:
It doesn't look like the extra tests involves any legacy build either.

claudioandre-br:
>> Should we possibly test "make -f Makefile.legacy generic" on CI?

> We did, that's why I usually don't care. Maybe Claudio's recent changes put them under "more tests"?

We are no longer testing `legacy`. In fact, I'm still wondering that we are running more tests in Travis than we should (for a free plan - we fast answers). So, I'm thinking about removal, not adds.

As a side note: `I dropped the support for legacy Makefile`.

claudioandre-br:
>> What's meant by "more tests"? Are they not required for merge?

> Not sure. They are not required right now and it seems they won't run unless we do something. @claudioandre-br?

Consider them `manual` tests. They are not required for merge, but you can run them whenever you want. We've already completed our parallel slots in the Circle, this is my try to keep things fast and have what we used to have inside CI.

claudioandre-br:
> On a side note we should get this repo owned by Openwall as an "organization" instead of me as a single user. That way we can have several administrators.

+1

solardiz:
> we should get this repo owned by Openwall as an "organization"

I hope we'll do that and take other steps towards making jumbo+Johnny the official JtR after this release and for 2.0. As you've noticed, I'm already changing the language from official vs. unofficial to core vs. jumbo.

Issue #3918, opened 2019-05-10
Supply proper OpenCL headers ourselves, and detect too old OpenCL lib


magnumripper:
The configure script should ensure we have OpenCL >= 1.2 in headers **and** lib, and then some of our headers should enforce the actually sourced OpenCL header also is >= 1.2

solardiz:
As I wrote in https://github.com/magnumripper/JohnTheRipper/issues/3915#issuecomment-491499246:

As I understand, hashcat bundles the Khronos Group headers under deps/OpenCL-Headers in its tree. Maybe we should consider doing the same.

Maybe that would be a better post-release change than adding an autoconf check for the system's headers version.

magnumripper:
Agreed. We should include the official headers and just check for libs.

alainesp:
I can add the Khronos Group headers and fix this issue. Does JtR use git submodules or prefer to copy the code in a folder?

solardiz:
> I can add the Khronos Group headers and fix this issue. Does JtR use git submodules or prefer to copy the code in a folder?

Thank you! No submodules, please add a subdirectory, I guess `src/CL` and I guess we'll need to change `<>` to `""` inside the files (separate commit, please) so that we don't have to specify a `-I` when building.

Once we have that, we could want to allow to build with OpenCL even when the library isn't found at `configure` time. Maybe we should support both `--enable-opencl` and `--disable-opencl`, with the default depending on whether the library is found at `configure` time.

So maybe 4 commits total:
1. Add `src/CL` with Khronos Group files as they are.
2. Change `<>` to `""` inside the files.
3. Change `<>` to `""` when referring to those files from our existing source files, and drop the `__APPLE__` special case.
4. Enhance `configure` as above (edit `configure.ac`, regenerate `configure` with same version of autoconf as before - e.g., on "super").

solardiz:
Items 1, 2, 3 above have been taken care of by @alainesp in #5396, so I'm closing this issue. We haven't yet decided on the `configure` script changes.

ghost:
We are not really checking "old OpenCL library" yet. Something like this is required:

But, there is one testing issue:
- no OpenCL < 1.2 available. Does `bull` have it? 
 
So, in fact, it was not really tested here.

```
diff --git a/src/opencl_dynamic_loader.c b/src/opencl_dynamic_loader.c
index 43b5b8a87d..25880927f7 100644
--- a/src/opencl_dynamic_loader.c
[...]

PR #3920, opened 2019-05-10
Check OpenCL headers version and bail out if 1.2 is not supported.

magnumripper:
See #3918.

I think this is enough pre-release. Changing autoconf is more error-prone so let's wait with that until post release.

### DIFF
--- a/src/opencl_common.h
+++ b/src/opencl_common.h
@@ -31,6 +31,10 @@
 #include <CL/cl_ext.h>
 #endif
 
+#if !CL_VERSION_1_2
+#error We need minimum OpenCL 1.2 to build with OpenCL support. The headers currently used does not comply.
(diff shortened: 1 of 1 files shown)


src/opencl_common.h
```diff
@@ -31,6 +31,10 @@
 #include <CL/cl_ext.h>
 #endif
 
+#if !CL_VERSION_1_2
+#error We need minimum OpenCL 1.2 to build with OpenCL support. The headers currently used does not comply.
+#endif
+
```
magnumripper: Note: According to some spec I took notes from (and put in the wiki here) long ago, this should be defined to `120`, but on my Macbook it's merely defined as `1`. Merely testing it like this should do just fine.

PR #3923, opened 2019-05-10
Fold long lines in doc/NEWS

magnumripper:


solardiz:
Thanks. With this, we have 4 lines that are exactly 80 chars. Ideally, we'd have 79 as the maximum. However, some of these 4 lines are not convenient/consistent to wrap or edit differently. So I just note this in here.

```
$ egrep '^.{80}$' NEWS
Changes made between 1.8.0-jumbo-1 (December 2014) and 1.9.0-jumbo-1 (May 2019):
  --enable-ztex", requires libusb).  Specifically, we support: bcrypt, descrypt,
  - https://www.openwall.com/lists/john-users/2019/02/03/1 sha512crypt & Drupal7
  that very string on the command line as "--format=dynamic='sha1(md5($p).$s)'".
```

PR #3926, opened 2019-05-11
Don't report virtual c/s if virtual time didn't advance

### LINKED ISSUE #3925: Misreported "virtual" speeds on Linux/sparc64
In testing on Linux/sparc64, I notice that `--test` reports for "c/s virtual" are 500 times higher than expected (e.g., 500 times higher than "c/s real" when running one thread). I don't yet know what causes this - could be our bug, could be a Linux kernel bug, or maybe something else.

solardiz:
Fixes #3925

### DIFF
--- a/src/bench.c
+++ b/src/bench.c
@@ -608,14 +608,12 @@ char *benchmark_format(struct fmt_main *format, int salts,
 #else
 	end_real = times(&buf);
 #endif
-	if (end_real == start_real) end_real++;
 
 #if defined (__MINGW32__) || defined (_MSC_VER)
 	end_virtual = end_real;
 #else
 	end_virtual = buf.tms_utime + buf.tms_stime;
 	end_virtual += buf.tms_cutime + buf.tms_cstime;
-	if (end_virtual == start_virtual) end_virtual++;
 #endif
 
 	results->real = end_real - start_real;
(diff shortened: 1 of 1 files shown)


magnumripper:
Aha it was Linux/Sparc. Strange.

solardiz:
> Is this something you could want in core as well?

Yes. I also wonder whether we should drop conditional compilation and simply set end=start to trigger this same check on systems where we have no API to read virtual times. But not now.

solardiz:
> it was Linux/Sparc. Strange.

Definitely strange. I think I saw working virtual times on that platform before. Maybe some temporary kernel or glibc or distro bug.

solardiz:
All tests except for more-tests passed. I think more-tests is pending forever, right? So I'll merge.

claudioandre-br:
I run more-tests after the merge (not for all commits). The reviewer can merge before, if he wants.

solardiz:
> > it was Linux/Sparc. Strange.
> 
> Definitely strange. I think I saw working virtual times on that platform before. Maybe some temporary kernel or glibc or distro bug.

That system is indeed quite broken - also "top" doesn't show processes consuming CPU time, and "ps" shows 0:00 for almost all processes where I'd expect non-zero.

PR #3930, opened 2019-05-11
md5crypt-long optimizations

solardiz:
Sorry I couldn't resist. ;-) In my testing, this is +16% for length 7, +7% for length 24, +13% for length 47 on super.

### DIFF
--- a/src/john.c
+++ b/src/john.c
@@ -143,7 +143,7 @@ extern int CPU_detect(void);
 extern char CPU_req_name[];
 #endif
 
-extern struct fmt_main fmt_DES, fmt_BSDI, fmt_MD5, fmt_smd5, fmt_BF;
+extern struct fmt_main fmt_DES, fmt_BSDI, fmt_MD5, fmt_md5crypt_long, fmt_BF;
 extern struct fmt_main fmt_scrypt;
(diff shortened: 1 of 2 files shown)


solardiz:
Curiously, md5crypt-long is now ~45% faster than the more-optimized md5crypt format on this UltraSparc T5 system I'm testing on. I tried disabling MD5_X2, but this made little difference. I suspect it has to do with other load on the system (it's not otherwise idle at all, with load averages around 15.0 on the 64 logical-CPU system) running its own hardware threads and thus constantly competing for the tiny L1 data caches. This CPU only has 16 KB of L1 data cache per 8 hardware threads, so only 2 KB per thread. Our optimized md5crypt format uses 8 pre-filled MD5 buffers, so ~512 bytes. Our md5crypt-long format, even after optimizations in this PR, has a smaller L1 working set size. Although its buffer is of similar size, it only fills and reads back from the initial portion of the buffer depending on actual password length (and my performance comparison here is for length 7), and then works with just one MD5 context. So for length 7 it probably needs something like 100 bytes.

We should consider reusing this trick in OpenCL. Maybe it'd result in lower GWS becoming more optimal, with the buffers still in cache.

Issue #3932, opened 2019-05-11
Discussion about formats failing on non-x86


claudioandre-br:
Please move the discussion about formats failing on non-x86 for a new issue. Thanks! It's a nasty surprise to me we still have those issues, and will probably want to fix them - or at least try. All of these tests are passing for me e.g. on sparc64.

_Originally posted by @solardiz in https://github.com/magnumripper/JohnTheRipper/issues/3929#issuecomment-491539404_
*****

> * a few formats fail on BE (non-SIMD);

All of those are auto-disabled on BE. All remaining ones pass self-test on BE. So this isn't user-visible unless the user actually wanted to use some of those few formats, which would end up missing.

> * powerpc build has "problems";

I'm not aware of any current problems on PowerPC. I probably need to retest there.
*****

These are the formats I disable before testing:
- https://raw.githubusercontent.com/claudioandre-br/packages/master/john-the-ripper/tests/disable_formats.sh

Some fail on BE, some are so slow that hurts CI, ... 

But, I re-enabled them to check how things are now. Unfortunately:

S390x
```
Testing: enpass, Enpass Password Manager [PBKDF2-SHA1 32/64]... (4xOMP) FAILED (cmp_all(32))
Testing: STRIP, Password Manager [PBKDF2-SHA1 32/64]... (4xOMP) FAILED (cmp_all(256))

2 out of 402 tests have FAILED
 FAILED: -test-full=0 --format=cpu
```

That said, I need them disabled and the "what we know" above is correct.

****
Follow up, as another example:

powerpc

```
Testing: AndroidBackup [PBKDF2-SHA1 128/128 AltiVec 4x AES]... (4xOMP) Warning: salt() returned misaligned pointer
PASS
Testing: radius, RADIUS authentication [MD5 32/32]... (4xOMP) Warning: salt() returned misaligned pointer
PASS
Testing: bitshares, BitShares Wallet [SHA-512 64/32]... (4xOMP) Warning: salt() returned misaligned pointer
PASS
Testing: andOTP [SHA256 32/32]... (4xOMP) Warning: salt() returned misaligned pointer
PASS

dynamic_assign_script_to_format() the dynamic.cmp_all/cmp_one() failed. This expression can not be handled by john!
Benchmarking: dynamic=md5($p) [Dynamic RDP]... FAILED (cmp_all(1))

 FAILED: -test=3 -form="dynamic=md5($p)"
```

solardiz:
[...]

PR #3934, opened 2019-05-12
Fix salt alignment in a bunch of formats, see #3932

magnumripper:


solardiz:
I'm running a `--test-full=1` of this on sparc64 now.

solardiz:
Have you reviewed the fields in each of these formats' `struct custom_salt`? We need to make sure there are no fields of types like `long` the required alignment of which varies between architectures.

solardiz:
> I'm running a `--test-full=1` of this on sparc64 now.

```
Testing: dynamic_2014 [md5($s.md5($p).$s) (PW > 55 or salt > 11 bytes) 32/64 x2]... PASS
Testing: dummy [N/A]... PASS
Testing: crypt, generic crypt(3) [?/64]... (13xOMP) PASS
All 402 formats passed self-tests!
0.00user 0.00system 6:27.03elapsed 0%CPU (0avgtext+0avgdata 515328maxresident)k
```

magnumripper:
> Have you reviewed the fields in each of these formats' `struct custom_salt`? We need to make sure there are no fields of types like `long` the required alignment of which varies between architectures.

Yes. I also use to replace all uses av `long` with `uint64_t` or `int` as needed whenever I see them.

Issue #3924, opened 2019-05-11
times() vs. clock() discrepancies


solardiz:
JtR core uses times(2). JtR jumbo will sometimes use clock(3) instead, for systems where times(2) is presumably unavailable. However, JtR jumbo "forgets" that while times(2) clocks are in units of `sysconf(_SC_CLK_TCK)`, clock(3) ones are in units of CLOCKS_PER_SEC. Hopefully, these just happen to be the same on systems where jumbo uses clock(3) now (MinGW, etc.) - but we need to carefully review the code and make sure we start using the proper units corresponding to the calls we make.

magnumripper:
Apparently `CLOCKS_PER_SEC` is 1000 under Windows (and it's `clock()` doesn't just count CPU time but also sleep time - why am I not surprised?). I yet to find any reference to `sysconf(_SC_CLK_TCK)` under Windows (using Google as a shield).

magnumripper:
Confirmed - they are both 1000 under MinGW, 32-bit or 64-bit. This means we could merge the PR, but it also means we don't need to yet %-)

PR #3938, opened 2019-05-12
Blind fix of #3924. As far as I can see, this should do it.

magnumripper:


solardiz:
Please _don't_ merge this yet. I actually thought of this kind of fix the moment I noticed the problem, and fixing it properly didn't look as simple as this.

solardiz:
Let's leave the original issue unfixed for the release this time, not to introduce worse issues.

magnumripper:
No problem, we'll wait. But I couldn't spot any foot gun here, did you?

magnumripper:
...apart from the missing semi-colon I just added 🙄

solardiz:
> I couldn't spot any foot gun here, did you?

Maybe not. Need to also check the uses from status.c. Anyway, testing this requires manually running a mingw build and sanity-checking the speeds perhaps against what's seen just prior to these changes, for both `--test` and actual cracking. It's not something CI would test for us.

PR #3944, opened 2019-05-13
2x SHA-1 for XOP; OpenBSD-SoftRAID benchmark fix

solardiz:
As discussed starting with https://github.com/magnumripper/JohnTheRipper/issues/2914#issuecomment-491791916


solardiz:
I think we should use "rebase and merge" on this PR, keeping the two commits separate.

PR #3950, opened 2019-05-14
Fix a "bug" in Jim's portable hi-res timer code

### LINKED ISSUE #3949: Jim's "portable high-resolution timer" sometimes get low resolution
This timer is currently used only in OMP/MKPC auto-tuning but since it's portable we could switch to it for OpenCL auto-tuning, benchmarks and other things as well.

Linux gets a `sm_hrPrecision` of 1 us, that's fine. Under macOS though, I get a resolution of about 200 us. Looking at `sm_cPrecision` we seem to detect an actual 1 us resolution from `clock()` so I'm not sure what  is happening.
MinGW claims 99 ns resolution. 64-bit on a CI see about 399 ns while 32-bit see 599 ns. They both might be affected by heavy load of host and are possibly also VMs.

Not sure whether it's just the trimming within `sTimer_Init()` that need to be enhanced, or some bug.

magnumripper:
… that decreased resolution significantly under macOS. Also add some diagnostics to --list=build-info.
Closes #3949.

### DIFF
--- a/src/listconf.c
+++ b/src/listconf.c
@@ -20,6 +20,13 @@
 #define NEED_OS_FLOCK
 #include "os.h"
 
+#include <unistd.h>
+#if HAVE_SYS_TIME_H
+#include <sys/time.h>
+#endif
+#if HAVE_SYS_TIMES_H
+#include <sys/times.h>
+#endif
 #if !AC_BUILT
  #include <string.h>
  #ifndef _MSC_VER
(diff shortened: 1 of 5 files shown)


magnumripper:
This should be safe to merge. I double-checked that supplying NULL for `tz` is allowed in POSIX and actually POSIX says everything but that is undefined 😆

magnumripper:
On the other hand, it's not that important so we might want to wait until after release.

magnumripper:
This is now NOT safe to merge before release. I improved timers under Linux (and others) and macOS to nanosecond ones. That's normally overkill, but why not...

Issue #3956, opened 2019-05-15
fcntl(F_WRLCK) errors with mpirun


ringt:
When using JtR/magnum across a decently large number of nodes, large numbers of write lock errors occur, terminating those specific threads.  

Using default input encoding: UTF-8
Node numbers 1-1056 of 1056 (MPI)
119@nid04460: fcntl(F_WRLCK): Resource temporarily unavailable
82@nid04459: fcntl(F_WRLCK): Resource temporarily unavailable
960@nid04570: fcntl(F_WRLCK): Resource temporarily unavailable
81@nid04459: fcntl(F_WRLCK): Resource temporarily unavailable
969@nid04572: fcntl(F_WRLCK): Resource temporarily unavailable
948@nid04570: fcntl(F_WRLCK): Resource temporarily unavailable
884@nid04569: fcntl(F_WRLCK): Resource temporarily unavailable
963@nid04570: fcntl(F_WRLCK): Resource temporarily unavailable
962@nid04570: fcntl(F_WRLCK): Resource temporarily unavailable
899@nid04569: fcntl(F_WRLCK): Resource temporarily unavailable
967@nid04570: fcntl(F_WRLCK): Resource temporarily unavailable
920@nid04569: fcntl(F_WRLCK): Resource temporarily unavailable
972@nid04572: fcntl(F_WRLCK): Resource temporarily unavailable
970@nid04572: fcntl(F_WRLCK): Resource temporarily unavailable
826@nid04563: fcntl(F_WRLCK): Resource temporarily unavailable
<snip>  

### Steps to reproduce ###
I am not the SA of the system(s) and unfortunately on this particular one they are still using CUDA versus OpenCL, so the configuration from the build was as follows:
./configure --enable-mpi --disable-opencl --enable-pkg-config 
I have tried different versions of MPI/GCC and that hasn't seemed to matter.  I did try using FLOCK instead of FNCTL but it produces similar errors.

Using the built-in queuing system (basically a wrapped mpirun where things like the number of processes are set in a script as variables) as follows:
aprun -F e -B ./john --pot=./john.pot --format=NT --incremental=VarAll12 --session=nt-incva12 ${HOME}/hashes.txt > va12-john.out

[...]

PR #3958, opened 2019-05-15
Handle spurious EAGAIN when using fcntl() file locking

magnumripper:
We should never get EAGAIN with F_SETLKW, but a user saw it, see #3956.

### DIFF
--- a/src/logger.c
+++ b/src/logger.c
@@ -43,6 +43,7 @@
 #include <stdarg.h>
 #include <string.h>
 #include <signal.h>
+#include <stdlib.h>
 
 #include "arch.h"
 #include "misc.h"
@@ -58,6 +59,7 @@
 #include "john_mpi.h"
 #include "cracker.h"
 #include "signals.h"
+#include "logger.h"
 
 static int cfg_beep;
 static int cfg_log_passwords;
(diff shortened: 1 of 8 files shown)


magnumripper:
@solardiz should I merge this as-is or could it have bad side-effects? We could make it a config option eg. `LockBugWorkaround = <bool>`.

magnumripper:
Also, perhaps some sub-second sleeping (only when we see `EAGAIN`, before re-trying) would decrease the hammering on that poor cluster server?

magnumripper:
Also, perhaps we should do the same workaround for `flock()`? User tried using flock instead and saw similar errors.

solardiz:
I think we should:

1. Sleep for a little while using select(2) on `EAGAIN`, yet continue to retry immediately on `EINTR`.

2. Drop `flock` support entirely. Why do we still have it at all?

But I don't mind us merging this PR as-is first, then making the changes above separately.

The issue the user is having is probably a limit on the number of concurrently held locks.

solardiz:
> The issue the user is having is probably a limit on the number of concurrently held locks.

Also, we might not want to fully mute the issue. Printing a warning (also explicitly suggesting that a limit on the number of concurrently held locks be raised?) the first time we hit the `EAGAIN` case could be useful.

magnumripper:
> 2\. Drop `flock` support entirely. Why do we still have it at all?

I wouldn't know. Drop it in core and I'll follow! :-)

Anyway I'll do a more thorough patch instead of merging this:
- I'll move all our locks into a single shared helper function (in `logger.c`, no?) so we can do everything right in **one** place.
- Adding `log_event()` and print to stderr (both once per node). The screen output can suggest checking ulimits etc.
- I'm not sure how I'd use `select(2)` here. Would I put the affected fd in `errorfds`? I need to read up on that.

magnumripper:
> * I'm not sure how I'd use `select(2)` here. Would I put the affected fd in `errorfds`? I need to read up on that.

Are you referring to something like this? https://stackoverflow.com/questions/3125645/why-use-select-instead-of-sleep

magnumripper:
Done. For the sleep, I use `rand()` (which should do fine for this task, seeded with node number) as in
```c
			/* Sleep for a random time of max. 130 ms */
			t.tv_sec = 0; t.tv_usec = (rand() & 1023) << 7;
			select(0, NULL, NULL, NULL, &t);
			continue;
```
This is 0-130 ms, is that fine? I have no idea of what time frames we have here.

magnumripper:
> This is 0-130 ms, is that fine? I have no idea of what time frames we have here. Perhaps once we hit this wall, a much longer wait would do more good and hopefully separate the nodes in time so they clash less at next event.

Hmm come to think of it, the timers wont separate anyway - next `timer_save` will clash just as much, right? Some ideas:
1. The **initial** `timer_save_value` in `sig_init()` could be ` += (NODE & 31)` (or 63) for 32x or 64x better separation without any other drawback than that the very first rec & log save with be up to 31 or 63 seconds later.
2. And/or we could add a new function in `signals.c` that merely does `timer_save_value++;` - and call that function whenever we get an `EAGAIN` here. Perhaps even when we get a `EINTR`?
3. And/or we could reset `timer_save_value` *after* files are saved instead of when setting `event_save = event_pending = 1`.
4. We could automagically increase the `LOG_BUFFER_SIZE` buffers according to total number of nodes.

@solardiz what do think?

solardiz:
This is a lot of questions and a significant rework of the locking. I'll plan on commenting on this in the next few days.

src/logger.c
```diff
@@ -113,6 +115,96 @@ static struct log_file pot = {NULL, NULL, NULL, 0, -1};
...
+ * params.
+ */
+int real_lock(int fd, int type, int mode,
+              const char *function, const char *file, int line)
+{
+#if OS_FLOCK || FCNTL_LOCKS
+
+#ifdef LOCK_DEBUG
+	fprintf(stderr, "%s(%u): Locking %s...\n",
+	        function, options.node_min, f->name);
```
magnumripper: Fixme: We no longer have `f->name` so enabling LOCK_DEBUG wont work.

src/logger.c
```diff
@@ -113,6 +115,96 @@ static struct log_file pot = {NULL, NULL, NULL, 0, -1};
...
+			continue;
+		} else if (errno != EINTR)
+			pexit(LOCK_FUNC "(%s%s)",
+			      type ? "exclusive lock" : "shared lock",
+			      mode ? "wait" : "nowait");
+	}
+
+#ifdef LOCK_DEBUG
+	fprintf(stderr, "%s(%u): Locked %s exclusively\n",
+	        __FUNCTION__, options.node_min, f->name);
```
magnumripper: Same `f->name` problem

src/logger.c
```diff
@@ -113,6 +115,96 @@ static struct log_file pot = {NULL, NULL, NULL, 0, -1};
...
+ * Shared helper for releasing a lock. Caller uses jtr_unlock(fd) which is
+ * actually a macro adding __FUNCTION__, __FILE__ and __LINE__ to params.
+ */
+void real_unlock(int fd, const char *function, const char *file, int line)
+{
+#if OS_FLOCK || FCNTL_LOCKS
+
+#ifdef LOCK_DEBUG
+	fprintf(stderr, "%s(%u): Unlocking %s\n",
+	        __FUNCTION__, options.node_min, f->name);
```
magnumripper: Same `f->name` problem

src/logger.c
magnumripper: https://stackoverflow.com/a/13544447

magnumripper: Not sure if there's a portable way of doing this at all

magnumripper:
I'll rewrite this to have a single function and no `jtr_unlock()`

PR #3963, opened 2019-05-16
Suppress the nagging about mask mode being faster if we're running single mode

magnumripper:
Suppress the nagging about mask mode being faster for FMT_MASK if user runs single mode (but not in batch mode).

This might be arguable.

### DIFF
--- a/src/john.c
+++ b/src/john.c
@@ -1772,7 +1772,8 @@ static void john_run(void)
 
 		/* Format supports internal (eg. GPU-side) mask */
 		if (database.format->params.flags & FMT_MASK &&
-		    !(options.flags & FLG_MASK_CHK) && john_main_process)
+		    !(options.flags & (FLG_MASK_CHK | FLG_SINGLE_CHK)) &&
+		    john_main_process)
 			fprintf(stderr, "Note: This format may be a lot faster with --mask acceleration (see doc/MASK).\n");
 
 		/* Some formats truncate at max. length */


solardiz:
> This might be arguable.

Yes, it is arguable. On one hand, the user can't just add "--mask" on top of "--single". On the other hand, if a format supports on-device mask chances are it'd have awful performance with "--single" and a warning is justified. Even with other modes, on top of which mask can be stacked, actually doing so changes the candidate passwords being tested. So I see no reason to single the "--single" mode out, and I think we shouldn't merge this change.

magnumripper:
In that case I think we should change the message to (in this case) something along the lines of *Note: This format needs --mask acceleration to perform well (see doc/MASK).  For single mode, the CPU format is likely better*. The current message adds confusion without actually giving a good advice for using single mode.

Also, the existing message otherwise printed should also be changed from *may be* to *is* or as above IMHO. Or is there any "may" for any of them?!

solardiz:
> _Note: This format needs --mask acceleration to perform well (see doc/MASK). For single mode, the CPU format is likely better_.

I don't mind, but frankly for single more isn't a CPU format better most of the time - not just for on-device mask capable formats - even after your enhancements that made it _sometimes_ reasonable to use on OpenCL/ZTEX devices? I think your enhancements changed "never run single mode on massively-parallel devices" to "it might _sometimes_ (but rarely) reasonable to run single mode on massively-parallel devices". No? If so, maybe some (different) warning should be printed when running single mode with any OpenCL or ZTEX format, or maybe when min_keys_per_crypt is above a threshold?

magnumripper:
For semi-slow formats like WPAPSK, single mode works like a champ on GPU provided we have lots of rules and seed it with words/morphemes. We might not quite reach normal GPU speed but it's a lot faster than the CPU format.

magnumripper:
Closing this for now; It's way too unimportant to consider until later.

PR #3966, opened 2019-05-17
Ci tweak

magnumripper:
Doing some experiments.

First I moved "win32" to after "win64" (serially) because the two would complete long before "asan" anyway. This made room to promote "non-simd" to default. Then I also put "encoding-opencl" to after "encoding-cpu" (serially) for the same reason.

Then added "delete all plugins" test a.k.a "no-plugs" which also disable the dynamic formats - effectively building a "core-formats Jumbo". This one is super quick so I also added the unit-tests to it. The "non-simd" test will run after "no-plugs" has finished. Disabling dynamics revealed a bug in loader.c that is now fixed in a separate commit.

All in all, we now do three more tests, still within the time of "asan" which is the slowest default one. And we still have a slot free under "more tests".

### DIFF
--- a/.circleci/circle-ci.sh
+++ b/.circleci/circle-ci.sh
@@ -33,6 +33,10 @@ JTR=../run/john
 # Build and testing
 if [[ $1 == "BUILD" ]]; then
 
+    if [[ "$PLUGS" == "none" ]]; then
+        rm -f *plug.[ch]
+    fi
+
     if [[ -n "$WINE" ]]; then
         do_Copy_Dlls
         export WINEDEBUG=-all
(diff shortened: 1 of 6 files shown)


magnumripper:
Screenshot 2019-05-17 at 11 04 54
After 1st commit

magnumripper:
Screenshot 2019-05-17 at 11 04 54
Final (for now).

magnumripper:
I'm happy with this but I'll let @claudioandre-br speak before I merge it. One caveat I can foresee is that if the CircleCI load is very high, perhaps the three second-level tests will be delayed? For example, if it takes 20 minutes before our main tests even start, there might also be another 20 minutes between encoding-cpu ands and encoding-opencl starts? OTOH in that case I could make merging possible without those three being finished (so the one who merges can decide whether they are important at all or not).

.circleci/circle-ci.sh
```diff
@@ -33,24 +33,32 @@ JTR=../run/john
...
+
     if [[ -n "$WINE" ]]; then
         do_Copy_Dlls
         export WINEDEBUG=-all
     fi
 
     # Configure and build
     if [[ "$TARGET_ARCH" == "i686" ]]; then
-        ./configure --host=i686-w64-mingw32 --build=i686-redhat-linux-gnu --target=i686-w64-mingw32
+        ./configure ${CONF_FLAGS} --host=i686-w64-mingw32 --build=i686-redhat-linux-gnu --target=i686-w64-mingw32
```
claudioandre-br: Where does CONF_FLAGS come from?

magnumripper: Good catch, is that still there? I added it, then realized I could use the existing `BUILD_OPTS` instead.
I'll drop `CONF_FLAGS` and add `BUILD_OPTS` instead (not currently used for these lines but may come handy).

magnumripper:
merged manually

PR #3974, opened 2019-05-19
Increase the jumbo version number

claudioandre-br:
If we were calling it `jumbo-1` a year ago, I think we should call it `jumbo-2` from now on.

### DIFF
--- a/src/params.h
+++ b/src/params.h
@@ -40,7 +40,7 @@
  * Jumbo's version number. Note that we must uncomment JTR_RELEASE_BUILD
  * above, in any release tar-balls (and only then, never ever in Git).
  */
-#define JUMBO_POSTFIX			"-jumbo-1"
+#define JUMBO_POSTFIX			"-jumbo-2"
 
 #if JTR_RELEASE_BUILD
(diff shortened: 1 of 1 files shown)


claudioandre-br:
A `pre` or `plus` is needed, because I can't say what I'm running, I need to compare hashes/commits.

magnumripper:
I'd suggest we do this then
```diff
diff --git a/src/params.h b/src/params.h
index 7cf94466a..7994aecf7 100644
--- a/src/params.h
+++ b/src/params.h
@@ -45,7 +45,7 @@
 #if JTR_RELEASE_BUILD
 #define JUMBO_VERSION                  JOHN_VERSION JUMBO_POSTFIX
 #else
-#define JUMBO_VERSION                  JOHN_VERSION JUMBO_POSTFIX "-bleeding"
+#define JUMBO_VERSION                  JOHN_VERSION JUMBO_POSTFIX "+bleeding"
 #endif
 
 /*
```

claudioandre-br:
:+1:

magnumripper:
Fixed in 3129e3a105

PR #3981, opened 2019-05-21
timer.h: Use CLOCK_MONOTONIC_RAW if available.

magnumripper:
See #3949

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -663,6 +663,7 @@ SKEY_LIBS
 GMP_LIBS
 CRYPT_LIBS
 Z_LIBS
+RT_LIBS
 M_LIBS
 OPENSSL_LIBS
 OPENSSL_CFLAGS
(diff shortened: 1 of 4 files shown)


solardiz:
Are we somehow linking with `-lrt` when that code version is enabled? If not, we should expect linkage issues on some systems - not only with this change, but also before, since apparently we were already using `clock_gettime`.

magnumripper:
> Are we somehow linking with `-lrt` when that code version is enabled? If not, we should expect linkage issues on some systems - not only with this change, but also before, since apparently we were already using `clock_gettime`.
> 
> Edit: if we haven't run into such issues, then maybe ` _POSIX_TIMERS && defined(_POSIX_MONOTONIC_CLOCK)` somehow isn't true, and this is dead (source) code?

Oh. You introduced `clock_gettime` in 01085c0a3f, but I missed this detail when merging so we don't have anything in `Makefile.in` (except, for some other reason, for Solaris builds, see `src/m4/jtr_systems_specific_logic.m4`). Apparently no-one has had problem with it yet! But we should fix this.

magnumripper:
> Edit: if we haven't run into such issues, then maybe ` _POSIX_TIMERS && defined(_POSIX_MONOTONIC_CLOCK)` somehow isn't true, and this is dead (source) code?

No, that very code added nanosecond resolution under Linux. Besides, your use of it in yescrypt doesn't seem to fail.

magnumripper:
#3984 added for the lack of `-lrt`.

> (...) see my other comment here for related concerns about both the original and the changed code.

I too am not happy with it: Like I said we should definitely stop using `double` for a starter. But I'll merge this for now, after adding a fix for #3984 as well. And we'll revisit this later.

solardiz:
> Besides, your use of it in yescrypt doesn't seem to fail.

It did fail without `-lrt` on some systems, but the use is only in the `userom.c` program, which we don't build in JtR.

PR #3983, opened 2019-05-21
CPU_detect: add Hygon Dhyana CPU ("HygonGenuine") support

hygonsoc:
Add Hygon CPU Vendor ID ("HygonGenuine") checking in CPU_detect() to enable optimization

### DIFF
--- a/src/x86.S
+++ b/src/x86.S
@@ -1317,6 +1317,8 @@ BF_body:
 #define CV_INTEL			$0x6C65746E
 /* "AuthenticAMD" */
 #define CV_AMD				$0x444D4163
+/* "HygonGenuine" */
+#define CV_Hygon			$0x656E6975
 /* "CentaurHauls" */
 #define CV_CENTAUR			$0x736C7561
 
@@ -1459,6 +1461,8 @@ CPU_detect_end_of_jumbo:
 	orb $1,%al			/* Detected a suitable CPU */
 	cmpl CV_AMD,%ecx
 	je CPU_detect_AMD		/* Is an AMD processor */
+	cmpl CV_Hygon,%ecx
+	je CPU_detect_P1		/* Is an Hygon processor */
 	cmpl CV_CENTAUR,%ecx
 	je CPU_detect_P1		/* Is a Centaur Technology processor */
 	cmpl CV_INTEL,%ecx


solardiz:
Thanks @hygonsoc. Why do you think this CPU should be detected the same as P1, enabling those ancient optimizations? Do you have benchmarks before and after this patch?

solardiz:
Reading up on Hygon Dhyana, this change looks wrong to me. It is said that Hygon Dhyana is based on AMD's Zen. If so, we should detect both it and AMD's other Zen microarchitecture CPUs the same. Based on what I know about Zen, I'd expect P1 optimizations to hurt it, not help. But if they do help, then we need more than just this change. (Or maybe we'll end up dropping this ancient 32-bit build specific assembly code sooner.)

solardiz:
Ah, I see - we do currently choose the P1 code on AMD64 processors, but that's based on testing on _early_ AMD64 processors back in 2006 or so. I doubt it's still correct for newer ones. It's just that this file is for 32-bit builds only, and this code and this choice affects very few things even in those builds nowadays.

So am I correct to assume that your patch is based on grepping our code and trying to match the behavior it has on AMD processors, rather than on actual testing?

hygonsoc:
@solardiz Thanks, We do test in new 64-bit system, it works as expected. 

And you are correct for old 32-bit build, we hope to be compatible with old codes in 32-bit build, so we analyze the AMD64 code path and follow the same behavior.  if there is the plan to drop support for old 32-bit way, that would be great.

solardiz:
Thanks for your answer, @hygonsoc. I understand that you're proposing this sort of changes to lots of projects now, and that's great! However, this also means you can't dedicate a lot of attention to how these changes fit into the individual projects.

Of course, the code still works as intended, especially in 64-bit builds where this source file isn't even compiled. It's also been tested in 32-bit builds on our CI now. However, this says nothing about whether the change improved or hurt performance on your actual hardware.

In 64-bit builds, the change did nothing. Without actual performance testing, my guess is that in 32-bit builds it actually hurt a tiny bit. So your different vendor identification string actually provides you an advantage in this case, letting our legacy 32-bit code automagically _not_ enable obsolete optimizations (that were presumably only good for AMD64 up to some point in the past) on your new CPUs. It's like if we had a check for Zen in there and disabled the obsolete optimizations if so.

If we ever bother benchmarking this obsolete code on Zen (unlikely - everyone does 64-bit builds on new CPUs, and this is a tiny component of our project) and make any changes as a result, we'll try and remember that we need to treat your CPUs the same as Zen.

Meanwhile, I'll close this PR, so as not to complicate the code and not to hurt performance on your CPUs, even if in currently unusual builds and for a tiny and no longer important component.

PR #3987, opened 2019-05-21
Add --rules-skip-nop option

magnumripper:
Add `--rules-skip-nop` option, for when you already ran a slow format without rules and now want to run *with* rules.

This also fixes some bugs with `--rules-stack` vs. internal encodings and other voodoo magic.


solardiz:
I think feature additions like this should have their corresponding NEWS updates in the same PR.

A drawback is that we'll more often have clashes between pending PR's, though.

magnumripper:
That's was my idea but I forgot about it when adding this little feature... need to get used to it. If we maintain this well it's be a lot less release work.

The clashing isn't too much of a problem given the rate of contributions a normal week (bug fixes and things like that shouldn't be mentioned every time, only when they fix a severe problem). If things do clash, I'll just fix the conflicts when merging.

BTW once @kholia gets his format rattle gun working again, we might want to list new formats under a heading/section separate from the other stuff.

PR #3997, opened 2019-05-24
Handle spurious EAGAIN when using fcntl() file locking

magnumripper:
I now dropped `flock` and dropped AC testing of `fcntl()`.

I also had these thoughts in https://github.com/magnumripper/JohnTheRipper/pull/3958#issuecomment-493705445
> Done. For the sleep, I use `rand()` (which should do fine for this task, seeded with node number) as in
> 
> ```c
> 			/* Sleep for a random time of max. 130 ms */
> 			t.tv_sec = 0; t.tv_usec = (rand() & 1023) << 7;
> 			select(0, NULL, NULL, NULL, &t);
> 			continue;
> ```
> 
> This is 0-130 ms, is that fine? I have no idea of what time frames we have here. Perhaps once we hit this wall, a much longer wait would do more good and hopefully separate the nodes in time so they clash less at next event.

I now changed it to 0-260 ms but I'm also considering something like eg. 100-360 ms.

> Hmm come to think of it, the timers (assuming they're the cause of clash) wont separate anyway - next `timer_save` will clash just as much, right? Some ideas:
> 1. The **initial** `timer_save_value` in `sig_init()` could be ` += (NODE & 31)` (or 63) for 32x or 64x better separation without any other drawback than that the very first rec & log save will be up to 31 or 63 seconds later.
> 2. And/or we could add a new function in `signals.c` that merely does `timer_save_value++;` - and call that function whenever we get an `EAGAIN` here. Perhaps even when we get a `EINTR`?
> 3. And/or we could reset `timer_save_value` *after* files are saved instead of when setting `event_save = event_pending = 1`.
> 4. We could also automagically increase the `LOG_BUFFER_SIZE` buffers according to total number of nodes, with "some" formula.

### DIFF
--- a/src/autoconfig.h.in
+++ b/src/autoconfig.h.in
@@ -57,9 +57,6 @@
 /* Define to 1 if you have the `DSA_get0_pqg' function. */
 #undef HAVE_DSA_GET0_PQG
 
-/* Define to 1 if you have the <fcntl.h> header file. */
-#undef HAVE_FCNTL_H
-
 /* Define to 1 if you have the `fopen64' function. */
 #undef HAVE_FOPEN64
 
--- a/src/configure
+++ b/src/configure
@@ -755,7 +755,6 @@ with_openssl
 with_commoncrypto
 with_endian
 with_systemwide
-with_flock
 enable_asan
 enable_ubsan
 enable_ubsantrap
@@ -1425,7 +1424,6 @@ Optional Packages:
                           Set endianness for target if it doesn't detect
                           properly
   --with-systemwide       install for all users
-  --with-flock            use flock() file locking instead of fcntl()
 
 Some influential environment variables:
   CFLAGS_EXTRA
@@ -3439,14 +3437,6 @@ if test "${with_systemwide+set}" = set; then :
   withval=$with_systemwide;
 else
   with_systemwide=no
-fi
-
-
-# Check whether --with-flock was given.
-if test "${with_flock+set}" = set; then :
-  withval=$with_flock;
-else
-  with_flock=no
 fi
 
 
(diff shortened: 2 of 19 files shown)


magnumripper:
OK now we have two commits, probably best to review them separately. The first one drops flock() and adds handling of EAGAIN.

> 3. And/or we could reset `timer_save_value` *after* files are saved instead of when setting `event_save = event_pending = 1`.

The second one adds this.

magnumripper:
Rewritten yet again. I need to test this on a few systems.

2da98b1a drops a handfull of "`fopen(); lock()" in favor of `open(O_EXLOCK)`.
fea0637 is the "initial" thing, dropping `flock()` in favor of `fcntl()` and moving all such locks to a single place.
237b5a27e adds the timing fixes for avoiding clashes when using many (many!) nodes.

magnumripper:
LOL, that blew up in my face. I never noticed the `open(O_**LOCK)` was BSD-only. Reverted to this:

9f6af44d4 Drop use of flock().  Always use fcntl() locks.  Handle spurious EAGAIN
80b713b71 Implement some tricks for automagically decrease risk of file lock clashes

magnumripper:
I'm starting to realize the `timer_save` stuff is completely irrelevant: Those files are not clashing, it's one per node.

So I'll drop the second `timer_save` commit. The pot file and the log file is the two things we need to care about. That means this is what we really should address:
> 4. We could also automagically increase the `LOG_BUFFER_SIZE` buffers according to total number of nodes, with "some" formula.

magnumripper:
> I'm starting to realize the `timer_save` stuff is completely irrelevant: Those files are not clashing, it's one per node.

No, wait... Uh. Yeah I was right - the save timer **does** affect pot & log also, because `log_file_flush()` is called after rec save.

Not only do I have to understand all these things, I also need to remember what I knew a couple of days ago o.O

magnumripper:
merged manually

PR #4000, opened 2019-05-25
Add AppVeyor yaml file (do not merge)

claudioandre-br:
* Nothing will happen after the PR got merged.
* This is a stripped version (since AppVeyor run sequential builds, you have to decide which tests run).
  - no fallback;

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,163 @@
+######################################################################
+# Copyright (c) 2019 Claudio André <claudioandre.br at gmail.com>
+#
+# This program comes with ABSOLUTELY NO WARRANTY; express or implied.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, as expressed in version 2, seen at
+# http://www.gnu.org/licenses/gpl-2.0.html
+######################################################################
+environment:
+  global:
+    PROJ_FOLDER: C:\projects
+    JTR_FOLDER: C:\projects\JohnTheRipper
+    BUILD_FOLDER: C:\projects\JohnTheRipper\src
+
+  matrix:
+    # CygWin 64bits build on Windows Server 2016
+    - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
+      arch: x86_64
+      bits: 64
+      doing: test
+      test: yes
+      location: cygwin64
+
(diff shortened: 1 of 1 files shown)


solardiz:
Thanks, Claudio!

Why "a stripped version"? Especially given that nothing is going to happen when this is merged, why not archive here the full version?

Why do you ask not to merge this?

claudioandre-br:
> Why "a stripped version"? Especially given that nothing is going to happen when this is merged, why not archive here the full version?

The PR is for CI (testing stuff). Every time we push new code to the repository how it will react (?).

Motivation: everyone should be able to have CI run in less than 5 minutes.

Remember, jobs will run sequentially (it is Windows, every cloud minute is -more- expensive). To us it is a scarce resource; at least, to me it is.

So, you and magnum need to think how you will be able to integrate and run CD (Continuous Delivery) in the main repository. I don't have an opinion about it.
****

This is the file I'm using today. Anyone can copy, study and use it.
wget https://raw.githubusercontent.com/claudioandre-br/JtR-CI/master/tests/appveyor.yml

solardiz:
> The PR is for CI (testing stuff).

This wasn't my motivation when I asked you to make this PR. My initial goal was merely to have this part of our source code archived in our main tree and more likely looked at by the rest of our team. (It effectively became part of our source code when we started using it for our Windows build releases.)

CI use may come later, e.g. possibly after migration to Openwall org GitHub account. For now, I'm happy with occasional CI use and Windows builds continuing under your account, Claudio.

> This is the file I'm using today. Anyone can copy, study and use it.

Great. Here's what I think we should do: first merge this PR as-is, then you make another PR with the actual file you use today and we merge that one as well (updating the file in our repo). That way, we'll have a commits history between a basic "stripped version" of the file and the current advanced version. This might not match actual evolution of the file, but that's fine - it will make sense for someone studying this anyway, perhaps for the purpose of making further revisions.

How does this sound to you, Claudio and magnum?

magnumripper:
> Here's what I think we should do: first merge this PR as-is, then you make another PR with the actual file you use today and we merge that one as well (updating the file in our repo). That way, we'll have a commits history between a basic "stripped version" of the file and the current advanced version. This might not match actual evolution of the file, but that's fine - it will make sense for someone studying this anyway, perhaps for the purpose of making further revisions.

Sounds good to me!

PR #4005, opened 2019-05-27
Minor edits to documentation and comments

solardiz:


solardiz:
Makes sense about the errata, @magnumripper. But we need to keep that errata list in the README really short, or/and link to a separate ERRATA file (different from the now-removed BUGS in that ERRATA is about bugs that we've already fixed, but only recently, whereas BUGS was about non-fixed  bugs and shortcomings that were expected to stay for a long while).

solardiz:
GitHub said this (and several other PRs) was based off stale bleeding-jumbo branch revision and suggested I click a button to do a merge commit of newer bleeding-jumbo changes to here, which I did - intending to "Squash and merge", which would hide this extra commit. However, now it's going through CI all over again before I can merge. :-(

I'm not yet clicking the same for other pending PRs. magnum, perhaps you can take care of that for the PRs you've approved (two of mine and one Claudio's), so that you get them merged without CI? They don't touch existing source code anyway, so no need for CI.

magnumripper:
I'll change the setting not to require a rebase upon latest code (seldom needed anyway). Unfortunately that has the effect it can no longer be done easily from the web GUI AFAIK.

PR #4017, opened 2019-06-01
Build a proper tiny exe

ghost:
The CFLAGS_MAIN is used to compile code. We need to link it to make a proper binary.


solardiz:
Rather than compile/link using unnecessarily many options/libraries (which can also slow down program startup), I think it's best to use simply:
```
$(CC) -DMAIN_NAME=\"$(MAIN_NAME)\" symlink.c -O2 -Wall -s -o ../run/john.exe
```

This target is for a very specific platform, so I think it's OK to hard-code a few compiler options like this.

solardiz (approved):
Looks correct to me. Thanks!

PR #4024, opened 2019-06-07
ZTEX: JtR unrelated firmware versioning scheme (current version 1.0.0)

### LINKED ISSUE #3992: Make ZTEX firmware version string independent of JtR version
This is somewhat tricky to edit, so I chose not to bother, nor to bother @Apingis with a request to make this edit prior to our 1.9.0-jumbo-1 release (especially given that he was mostly unavailable at the time). However, the string still saying "JtR 1.8.x" might confuse some people.

I imagine the easiest way to edit this string is in the .ihx file, without actually recompiling the firmware, and in the host code for what string to expect. Should we maybe use JtR-unrelated versioning scheme there, so that this doesn't become confusing each time we make a new major release?

Alternatively, we might hide reporting of this string, and use it for JtR-internal comparisons only (like we already do). But I think coming up with a JtR-unrelated versioning scheme is both easier and better.

Please make suggestions in comments to this issue before actually committing a change.

Apingis:
Closes #3992 .


solardiz (approved):
Looks good to me. Thanks!

PR #4053, opened 2019-07-22
Cross platform compatibility for tezos2john.py

LordDarkHelmet:
The previous version did not account for cross platform execution. macOS was not able to access the bip-0039 directory. To fix this we use python's os.path.join to create proper paths. We also find the current absolute path of the script in the rare case where someone executes it from another directory.


ghost (approved):
Merging as:
* nonessential files;
* the PR is the result of testing/debugging.

PR #4060, opened 2019-07-28
Lower GWS when saving memory

### LINKED ISSUE #4057: Lower GWS when we're saving memory
We might want `--save-memory` to affect auto-tuning, resulting in lower GWS if the speedup of using a higher GWS would have been small yet the value would be large in absolute terms. If by default we require a 1%+ speedup for use of a higher GWS, then with memory saving enabled (at level 2 or higher? since level 1 is historically for not loading login names only) and values being significant in absolute terms (e.g., at least 10240?) we could use e.g. a 5%+ threshold instead.

ghost:
With memory saving enabled (at level 2 or higher since level 1 is historically
for not loading login names only) and GWS values being significant in absolute
terms, we could use, e.g., a higher speed gain threshold instead.

Closes #4057.

### DIFF
--- a/src/opencl_common.c
+++ b/src/opencl_common.c
@@ -1977,6 +1977,24 @@ void opencl_find_best_gws(int step, int max_duration,
 	int save_duration_time = duration_time;
 	cl_uint core_count = get_processors_count(sequential_id);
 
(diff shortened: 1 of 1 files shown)


magnumripper (approved):
Good stuff

PR #4071, opened 2019-08-20
Removed fmt_not_exact

kost123:
Useless whilst there is a second block check on .key files, which effectively removes false positives. Having this option turned on just reduces overall cracking times by a lot, especially for multiple hashes for no good reason.

### DIFF
--- a/src/multibit_fmt_plug.c
+++ b/src/multibit_fmt_plug.c
@@ -503,7 +503,7 @@ struct fmt_main fmt_multibit = {
 		SALT_ALIGN,
 		MIN_KEYS_PER_CRYPT,
 		MAX_KEYS_PER_CRYPT,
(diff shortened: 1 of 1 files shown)


magnumripper:
Why did you revert the PR? Was it 100% bogus, or perhaps we should set this flag per salt (depending on type/version/whatnot) instead?

magnumripper:
I believe salts ending in a base58 check of 16 bytes equals 92.8 bits, should be fine. The worst case for FP should be `is_bitcoinj_protobuf_data()` which should be about 66 bits if I get it right. So we should be able to drop `FMT_EXACT` for all cases. Before 87f72625 it was much worse (only 40 bits).

kost123:
I had to do some further testing to make sure everything was right, as I had some cmp(all) error, but it was unrelated.  As you say, it will be fine to drop it for all cases with no negative effects.

magnumripper (approved):

magnumripper:
Thanks!

PR #4075, opened 2019-08-25
Passgen costs

KoreLogicSecurity:
This diff probably conflicts with (at least) the useusername PR.

Let us know if you would prefer we merge some or all of these into one single PR; they felt like different enough changes that you might prefer to review/comment/reject individually.


ghost (approved):
I'm merging. So, more people can test.

Issue #4077, opened 2019-08-28
Can't build john-1.9.0-jumbo-1 at macOS Sierra and High Sierra


catap:
Het,

I've tried to upgrade brew's john-jumbo formula and it doesn't build on Sierra and High Sierra: https://github.com/Homebrew/homebrew-core/pull/43531

```
==> make -s CC=clang
ar: creating archive aes.a
ar: creating archive secp256k1.a
ar: creating archive ed25519-donna.a
Undefined symbols for architecture x86_64:
  "_fmt_stribog_256", referenced from:
      _john_register_all in john.o
  "_fmt_stribog_512", referenced from:
      _john_register_all in john.o
ld: symbol(s) not found for architecture x86_64
```

@magnumripper can you have any ideas what should I do?

catap:
As discussed at #4078 let move back to this issue to continue discussing.

Cc: @magnumripper, @solardiz

catap:
A bit more info about this machine:

Testing for SSE4.1 that is failed:

```
configure:10866: result: no
configure:10878: checking special compiler flags
configure:10880: result: Intel x86
configure:10893: checking whether OS X 'as' needs -q option
configure:10905: clang -mavx -o conftest -m64 -g -O2 -I/usr/local/include -I/usr/local/opt/openssl/include   -O0  -I/usr/local/include -I/usr/local/opt/openssl/include -m64  -L/usr/local/lib -L/usr/local/opt/openssl/lib conftest.c  >&5
configure:10905: $? = 0
configure:10906: result: no
configure:11359: Testing tool-chain's CPU support with given options
configure:11361: checking for MMX
configure:11373: clang -o conftest -m64 -g -O2 -I/usr/local/include -I/usr/local/opt/openssl/include   -O0  -I/usr/local/include -I/usr/local/opt/openssl/include -m64  -L/usr/local/lib -L/usr/local/opt/openssl/lib conftest.c  >&5
configure:11373: $? = 0
configure:11379: result: yes
configure:11392: checking for SSE2
configure:11404: clang -o conftest -m64 -g -O2 -I/usr/local/include -I/usr/local/opt/openssl/include   -O0  -I/usr/local/include -I/usr/local/opt/openssl/include -m64  -L/usr/local/lib -L/usr/local/opt/openssl/lib conftest.c  >&5
configure:11404: $? = 0
configure:11410: result: yes
configure:11425: checking for SSSE3
[...]

PR #4078, opened 2019-08-28
Fiexd mixing condition

catap:
Mixing `defined` and not defined condition at clang over macOS Sierra and High Sierra, at CPU before SSE4.1 crashed build with logs:

```
==> make -s CC=clang
ar: creating archive aes.a
ar: creating archive secp256k1.a
ar: creating archive ed25519-donna.a
Undefined symbols for architecture x86_64:
  "_fmt_stribog_256", referenced from:
      _john_register_all in john.o
  "_fmt_stribog_512", referenced from:
      _john_register_all in john.o
ld: symbol(s) not found for architecture x86_64
```

This is fixing https://github.com/magnumripper/JohnTheRipper/issues/4077

### DIFF
--- a/src/stribog_fmt_plug.c
+++ b/src/stribog_fmt_plug.c
@@ -8,7 +8,7 @@
 
 #include "arch.h"
 
-#if !defined(JOHN_NO_SIMD) && __SSE4_1__
+#if !defined(JOHN_NO_SIMD) && defined(__SSE4_1__)
 
 #if FMT_EXTERNS_H
 extern struct fmt_main fmt_stribog_256;


magnumripper:
Are you sure this alone fixed the problem? This should effectively be a no-op change: If `__SSE4_1__` is defined, it should be defined to 1 and if it's not defined it will (in this context) be regarded as 0.

Note that eg. `-D__SSE4_1__` will actually define it to 1.

catap:
@magnumripper I'm trying to reproduce it over home brew jenkins where this issue was originally found.

So, let me a bit more time to confirm that it is fixing the issue, and if it doesn't I'll share a bit more logs or make a new attempt.

magnumripper:
I just now tried configuring using `--enable-simd=sse2` and that works out fine.

During config, these lines are seen:
```
checking for MMX... yes
checking for SSE2... yes
checking for SSSE3... no
```
During build, these diagnostic warnings are seen (merely telling us that format was disabled):
```
stribog_fmt_plug.c:472:2: warning: #warning Stribog-256 and Stribog-512 formats require SSE 4.1, formats disabled [-Wcpp]
 #warning Stribog-256 and Stribog-512 formats require SSE 4.1, formats disabled
  ^~~~~~~
ar: creating archive aes.a
ar: creating archive ed25519-donna.a
ar: creating archive secp256k1.a

Make process completed.
```

magnumripper:
Possibly related: When using native clang (`./configure CC=/usr/bin/gcc --enable-simd=sse2`), archs up to SSE-4.1 are enabled anyway because Apple's clang is built that way.
```
checking for MMX... yes
checking for SSE2... yes
checking for SSSE3... yes
checking for SSE4.1... yes
checking for SSE4.2... no
```
The build still works fine though. And anyway we recommend using a real gcc as noted in end of  doc/INSTALL

catap:
@magnumripper as expected it doesn't fixed this.

Yes, it was blind shot and I'm out of ideas to be honest.

The saved logs from home brew jenkins: https://gist.github.com/catap/b41982a74b3acde0fa20a2ca5b16eb29

This build is used https://openwall.com/john/k/john-1.9.0-jumbo-1.tar.xz plus this pull request and one patch that shouldn't be related:
```
--- a/src/params.h	2012-08-30 13:24:18.000000000 -0500
+++ b/src/params.h	2012-08-30 13:25:13.000000000 -0500
@@ -101,15 +101,15 @@
  * notes above.
  */
 #ifndef JOHN_SYSTEMWIDE
-#define JOHN_SYSTEMWIDE			0
+#define JOHN_SYSTEMWIDE			1
 #endif

 #if JOHN_SYSTEMWIDE
 #ifndef JOHN_SYSTEMWIDE_EXEC /* please refer to the notes above */
-#define JOHN_SYSTEMWIDE_EXEC		"/usr/libexec/john"
+#define JOHN_SYSTEMWIDE_EXEC		"HOMEBREW_PREFIX/share/john"
 #endif
 #ifndef JOHN_SYSTEMWIDE_HOME
-#define JOHN_SYSTEMWIDE_HOME		"/usr/share/john"
+#define JOHN_SYSTEMWIDE_HOME		"HOMEBREW_PREFIX/share/john"
 #endif
 #define JOHN_PRIVATE_HOME		"~/.john"
 #endif
```

I'm trying to find access to machine with CPU likes x86_64 SSSE3, 64-bit LE and darwin17.7.0.

But we can try to use home brew jenkins if you have any ideas ;)

catap:
@magnumripper I guess issue is different.

The machine doesn't support SSE4.1:
```
checking for MMX... yes
checking for SSE2... yes
checking for SSSE3... yes
checking for SSE4.1... no
```

and this plugin is required it build.

But I don't understand how it
```
extern struct fmt_main fmt_stribog_256;
extern struct fmt_main fmt_stribog_512;
```

is generated on machine without SSE4.1

magnumripper:
That is very strange, yes. And BTW is that a Hackintosh? As far as I know, Apple never sold an intel machine with less than SSE-4.1 which is exactly why their native tool set always enables it.

That homebrew formula should probably replace `--disable-native-tests --disable-native-march` with `--enable-simd=sse4.1` for the same reason. And anyway they/you would be better off using gcc instead of clang.

catap:
@magnumripper I have no idea that it is the machine to be honest because I'm not related to this project.

I just tried to contribute updated john for personal uses and have some pain.

magnumripper:
I really can't reproduce the problem. If I use `./configure CC="clang -mno-sse4.1" --disable-native-tests --disable-native-march` on my Macbook, I seem to get a similar config (eg. SSSE-3 but no SSE-4.1) but build works just fine, Stribog formats are disabled.

magnumripper (changes requested):
Perhaps we should close this PR and re-open #4077 instead for further discussions.

solardiz:
On Thu, Aug 29, 2019 at 04:25:20AM -0700, magnum wrote:
> As far as I know, Apple never sold an intel machine with less than SSE-4.1

Apple did sell such machines, e.g. their Core 2 Duo based Macbooks from circa 2008 only do up to SSSE3.

solardiz:
On Thu, Aug 29, 2019 at 06:01:46AM -0700, magnum wrote:
> Perhaps we should close this PR and re-open #4077 instead for further discussions.

Yes, we should.

magnumripper:
> On Thu, Aug 29, 2019 at 04:25:20AM -0700, magnum wrote: As far as I know, Apple never sold an intel machine with less than SSE-4.1
> Apple did sell such machines, e.g. their Core 2 Duo based Macbooks from circa 2008 only do up to SSSE3.

OK. So then I assume they never sold a **64-bit** intel machine with less than SSE-4.1... or something like that!

solardiz:
On Thu, Aug 29, 2019 at 07:07:34AM -0700, magnum wrote:
> OK. So then I assume they never sold a **64-bit** intel machine with less than SSE-4.1... or something like that!

Also not true.  Core 2 Duo is 64-bit, and runs 64-bit OS X out of the box.

magnumripper:
That's weird. Their clang should default to SSSE3 then, not SSE-4.1. But never mind. This means the homebrew formula should use `--enable-simd=ssse3` for making a binary that runs on any Mac, instead of going all the way down to SSE-2. Not that it matters that much.

PR #4079, opened 2019-08-29
Botcheck

magnumripper:


solardiz:
On Thu, Aug 29, 2019 at 05:27:28AM -0700, magnum wrote:
>   * Add a shortcut option --length=N which is same as --min-length=N --max-length=N

This should be in NEWS.

Issue #3991, opened 2019-05-21
Improve (and maybe simplify) our "portable hi-res timer"


magnumripper:
At the very least we should drop all uses of `double` in that code (instead of reporting 1.234567891 seconds we'd simply report 1234567891 ns or perhaps just 1234568 µs).

magnumripper:
Also I think we should be able to observe function call latency instead of mistaking it for low resolution.

magnumripper:
9e3122ab reverted `sTimer_GetSecs()` to return `double`. We'll add a nanosec function such as `uint64_t sTimer_GetNano()` if/when needed.

PR #4081, opened 2019-08-31
timer.c: Drop all use of 'double' in favor of uint64_t. See #3991.

magnumripper:


solardiz:
I was wondering if I somehow missed #4489 when reviewing a PR. Found this PR now. There were no reviews by no reviewers. That's a relief.

While I'm not sure I'd have looked carefully enough to spot the issue, I think we'd have spotted it by actually testing the code that uses this timer looking for any differences in behavior before vs. after these changes.

Were were under impression that this timer was unused (dead code?) and thus didn't need testing?

Issue #4056, opened 2019-07-23
oldoffice-opencl, OpenCL CL_OUT_OF_HOST_MEMORY (-6) error in opencl_oldoffice_fmt_plug.c:138 - Error mapping saved_key


dosoos:
### John Version ###
\# john
John the Ripper 1.9.0-jumbo-1+bleeding-9b4335e7c 2019-07-18 11:52:24 +0200 OMP [linux-gnu 64-bit x86_64 AVX2 AC]
Copyright (c) 1996-2019 by Solar Designer and others
Homepage: https://www.openwall.com/john/

### When i use dictioary crack oldoffice hash have this problem ###
Using default input encoding: UTF-8
Loaded 1 password hash (oldoffice-opencl, MS Office <= 2003 [MD5/SHA1 RC4 OpenCL])
Cost 1 (hash type) is 0 for all loaded hashes
Node numbers 1-8 of 8 (fork)
Device 4: P106-100
Device 8: P106-100
Device 1: P106-100
Note: This format may be a lot faster with --mask acceleration (see doc/MASK).
Device 7: P106-100
Device 5: P106-100
Device 6: P106-100
Device 3: P106-100
Device 2: P106-100
8: OpenCL CL_OUT_OF_HOST_MEMORY (-6) error in opencl_oldoffice_fmt_plug.c:138 - Error mapping saved_key
1: OpenCL CL_OUT_OF_HOST_MEMORY (-6) error in opencl_oldoffice_fmt_plug.c:138 - Error mapping saved_key

## my machine have 8GB memory ##
maybe need more memory? or this is issue? bug?

dosoos:
### hash content ###
xxooxxoo.file:$oldoffice$0*66fff42aaf5231cd7cb14cf410d2a443*d143df50c79960663c127997433aaeca*5dc969945f79f6f40ff9044cc41c1ac4:::::xxooxxoo.xls

solardiz:
@dosoos It does sound like you need more memory, but you might be able to reduce JtR's memory needs by first running with "-v=4" (higher verbosity level), noting the auto-tuned values of GWS, and then explicitly specifying a lower value with "GWS=lowervaluehere john ...". You'd want the lower value to be e.g. 1/2 or 1/4 of the auto-tuned value.

Edit: for the verbose auto-tuning to complete without running out of memory, you'd omit `--fork` on that one run. You'd re-add `--fork` when you specify GWS explicitly.

[...]

PR #4082, opened 2019-08-31
Add functions to query host's physical memory

magnumripper:
Perhaps in the end we only want to use the total memory figure, I'm not quite sure.

The last commit uses this for (optionally) automagically increase or decrease the max. buffer size for single mode.

Future code may also use this for knowing beforehand when host-backed GPU buffers will be too big (see #4056).

### DIFF
--- a/doc/NEWS
+++ b/doc/NEWS
@@ -40,6 +40,11 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
 - Added --length=N which is a shortcut for --min-length=N --max-length=N
   [magnum; 2019]
 
+- Add a function for querying host's total or available physical memory, if
+  possible.  Among other things, this will be used for detecting when an OpenCL
+  buffer that is backed by an equal size host buffer gets too big.
+  [magnum; 2019]
+
 
 Major changes from 1.8.0-jumbo-1 (December 2014) to 1.9.0-jumbo-1 (May 2019):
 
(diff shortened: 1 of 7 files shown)


magnumripper:
Do not merge. I'll merge when ready. This PR incorrectly uses the timer-uint branch as base, for separation.

doc/NEWS
```diff
@@ -40,6 +40,11 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
 - Added --length=N which is a shortcut for --min-length=N --max-length=N
   [magnum; 2019]
 
+- Add a function for querying host's total or available physical memory, if
```
magnumripper, line 43: Should be functions, not "a function"

src/misc.c
```diff
@@ -633,3 +635,93 @@ char *rtrim(char *str)
...
+	if (pagesize < 0)
+		return -1;
+
+	long totmem = sysconf(_SC_PHYS_PAGES);
+	if (totmem < 0)
+		return -1;
+
+	avail_mem = (int64_t)totmem * pagesize;
+
+#endif
```
magnumripper, line 722: Oops, lines 710-721 are to be dropped (they were supposed to be moved to `host_total_mem()`, not copied)

src/misc.c
```diff
@@ -633,3 +635,93 @@ char *rtrim(char *str)
...
+	if ((fp = fopen("/proc/meminfo", "r"))) {
+		while (fgets(buf, LINE_BUFFER_SIZE, fp)) {
+			if (strstr(buf, "MemAvailable")) {
+				char *p = strchr(buf, ':');
+				if (p++)
+					avail_mem = strtoull(p, NULL, 10) << 10;
+				break;
+			}
+		}
+	}
```
magnumripper, line 679: We ought to close the file as well...

magnumripper:
Merged manually, with the new option default disabled.

PR #4092, opened 2019-09-04
New scheme for non-hash formats, that can exploit same-salt

magnumripper:
See #3497 

WPAPSK and a few other formats have FMT_BLOB at this time.

Notes:
Tunable costs only look at salts, so we can no longer separate WPA1 from WPA2 or 802.11w (OTOH they are not really costs per se - they take about the same time to process). For costs actually "hidden" in a FMT_BLOB we need to look at the blob data (a.k.a binary) and that'd be really hairy, right now I can't see it happening anytime soon. So for now we can't use costs that need to be separated from actual salt. For formats like office, we're storing actual salt, iteration count **and** office version in the "salt" so it's not a a problem there.

### DIFF
--- a/src/formats.c
+++ b/src/formats.c
@@ -254,6 +254,8 @@ static char* is_key_right(struct fmt_main *format, int index,
 
 	for (size = 0; size < PASSWORD_HASH_SIZES; size++)
 	if (format->methods.binary_hash[size] &&
+	    format->methods.get_hash[size] &&
+	    format->methods.get_hash[size] != fmt_default_get_hash &&
 	    format->methods.get_hash[size](i) !=
 	    format->methods.binary_hash[size](binary)) {
 		if (options.verbosity > VERB_LEGACY) {
@@ -341,6 +343,23 @@ static char* is_key_right(struct fmt_main *format, int index,
 	return NULL;
 }
 #endif
+
+int fmt_bincmp(void *b1, void *b2, struct fmt_main *format)
+{
+	if (!(format->params.flags & FMT_BLOB))
+		return memcmp(b1, b2, format->params.binary_size);
+
+	size_t s1 = ((fmt_data*)b1)->size;
+	size_t s2 = ((fmt_data*)b2)->size;
+
+	if (s1 != s2)
+		return 1;
+
+	b1 = ((fmt_data*)b1)->blob;
+	b2 = ((fmt_data*)b2)->blob;
+
+	return memcmp(b1, b2, s1);
+}
 
 static char *fmt_self_test_body(struct fmt_main *format,
     void *binary_copy, void *salt_copy, struct db_main *db, int full_lvl)
@@ -488,6 +507,13 @@ static char *fmt_self_test_body(struct fmt_main *format,
 
 	if (format->params.flags & FMT_HUGE_INPUT) {
 		for (size = 0; size < PASSWORD_HASH_SIZES; size++) {
+			/*
+			 * We could allow this, but they would need to calculate the
+			 * same MD5 as the $SOURCE_HASH$ of a truncated pot entry and
+			 * return a portion of that.  This test could instead confirm
+			 * that's the case.  Given common properties of these formats,
+			 * it not likely worth bothering with.
+			 */
 			if (format->methods.binary_hash[size] &&
 			    format->methods.binary_hash[size] !=
 			    fmt_default_binary_hash)
@@ -537,6 +563,10 @@ static char *fmt_self_test_body(struct fmt_main *format,
 	if ((format->params.flags & FMT_UNICODE) &&
 	    !(format->params.flags & FMT_8_BIT))
 		return "FMT_UNICODE without FMT_8_BIT";
+
+	if ((format->params.flags & FMT_BLOB) &&
+	    (format->params.binary_size != sizeof(fmt_data)))
+		return "FMT_BLOB with incorrect binary size";
 
 	if ((format->methods.binary == fmt_default_binary) &&
 	    (format->params.binary_size > 0))
@@ -662,10 +692,12 @@ static char *fmt_self_test_body(struct fmt_main *format,
 
 		/* validate that binary() returns cleaned buffer */
 		if (extra_tests && !binary_cleaned_warned && format->params.binary_size) {
+			BLOB_FREE(format, binary);
 			memset(binary, 0xAF, format->params.binary_size);
 			binary = format->methods.binary(ciphertext);
 			if (((unsigned char*)binary)[format->params.binary_size-1] == 0xAF)
 			{
+				BLOB_FREE(format, binary);
 				memset(binary, 0xCC, format->params.binary_size);
 				binary = format->methods.binary(ciphertext);
 				if (((unsigned char*)binary)[format->params.binary_size-1] == 0xCC)
@@ -676,6 +708,7 @@ static char *fmt_self_test_body(struct fmt_main *format,
 				}
 			}
 			/* Clean up the mess we might have caused */
+			BLOB_FREE(format, binary);
 			memset(binary, 0, format->params.binary_size);
 			binary = format->methods.binary(ciphertext);
 		}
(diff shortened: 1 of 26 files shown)


magnumripper:
Format flag added: `FMT_BLOB`.  Apart from that, the format and loader structs are unchanged!  For `FMT_BLOB` we just re-purpose the binary pointer and the `cmp_*` functions.  A new `typedef struct ... fmt_data` is used for blobs, it's always dynamically sized and can be ether tiny_alloc'ed or alloc'ed (cracked alloc'ed blobs will be freed by cracker).  This potentially obsoletes dynamic_salt if/when all such formats are converted to this new style.

src/formats.c
```diff
@@ -1621,7 +1621,7 @@ char *fmt_self_test(struct fmt_main *format, struct db_main *db)
 	self_test_running = 0;
 
 	MEM_FREE(salt_alloc);
-	BLOB_FREE(binary_alloc);
+	//BLOB_FREE(binary_alloc);
```
magnumripper: I can't yet see why this fails with ASan unless commented out. It boils down to this after expansion:
```c
	if (binary_alloc && (format->params.flags & FMT_BLOB) &&
		(((fmt_data*)binary_alloc)->flags & FMT_DATA_ALLOC)) 
			MEM_FREE(((fmt_data*)binary_alloc)->blob);
```

magnumripper: ...Oh and as we all know, `MEM_FREE` boils down to checking that `((fmt_data*)binary_alloc)->blob` is not NULL either. So why the heck does it say BOOM with
```
==49736==The signal is caused by a WRITE memory access.
==49736==Hint: address points to the zero page.
```

magnumripper:
Got another ASan problem during testing. @solardiz what exactly does the following (john proper) code in `crk_process_guess()` do? I don't quite understand it:
```c
        dupe = crk_timestamps[index] == status.crypts;
        crk_timestamps[index] = status.crypts;
```
In order not to miss stuff with `FMT_BLOB` I had to "disable" that, but apparently that was a bit too bold. Since I don't understand the rationale with this I'm not quite sure how I should proceed.

magnumripper (commented):
Whatever that `dupe` stuff does, this b2678258f seems to deal with the problem. We'll see how ASan behaves...

solardiz:
> what exactly does the following (john proper) code in `crk_process_guess()` do?

As I recall, it was a (too) clever hack to avoid writing dupes to `john.pot` and to single crack's buffers when the same password is cracked more than once per `crypt_all` (perhaps because the same hash is loaded more than once, which may happen e.g. in single crack mode). However, I cannot confirm it achieving that now when I look at recent code. Maybe we need to test with duplicate hashes and see what happens.

Anyway, I also don't see why this would be relevant to your `FMT_BLOB` work.

> `MEM_FREE` boils down to checking that `((fmt_data*)binary_alloc)->blob` is not NULL

But it wouldn't check whether `(fmt_data*)binary_alloc)` is NULL, and would crash on NULL when trying to access `->blob`.

src/wpapsk_fmt_plug.c
```diff
@@ -212,33 +210,27 @@ struct fmt_main fmt_wpapsk = {
...
 		crypt_all,
 		{
-			get_hash_0,
-			get_hash_1,
-			get_hash_2,
-			get_hash_3,
-			get_hash_4,
-			get_hash_5,
-			get_hash_6
+			fmt_default_get_hash,
```
solardiz: Looks wrong to me: why use 7 separate `binary_hash_*` functions when we also use `fmt_default_get_hash`, which always returns 0? And, shouldn't there also be hunks removing the now-unused functions `clear_keys` and `get_hash_*`?

solardiz (commented):
I don't have time to seriously review this and wasn't planning on including these changes in the very next release, but if you're now in the mood, I won't stop you.

It is indeed good to be able to take advantage of matching ESSIDs. Perhaps we can also take advantage of matching salts with `openssl enc` blobs, in case someone specified a salt manually (unlikely?) and used the same more than once. Maybe there are more examples.

If this interferes with reporting of tunable costs that significantly affect performance, that's really bad, though. For example, what will happen to the `dmg` formats?

magnumripper:
> > what exactly does the following (john proper) code in `crk_process_guess()` do?
> 
> As I recall, it was a (too) clever hack to avoid writing dupes to `john.pot` and to single crack's buffers when the same password is cracked more than once per `crypt_all` (perhaps because the same hash is loaded more than once, which may happen e.g. in single crack mode). However, I cannot confirm it achieving that now when I look at recent code. Maybe we need to test with duplicate hashes and see what happens.

I see. I do think it works as intended...

> Anyway, I also don't see why this would be relevant to your `FMT_BLOB` work.

Because unlike normal hashes' salt and binary, FMT_BLOB can have one salt with multiple different binaries, where some may correspond to one same password (while others may not). Before I excluded FMT_BLOB from this test, some cracks were reported but not written to pot file. BTW my current code seem to disable it at the very right place, with no ill effects.

> > `MEM_FREE` boils down to checking that `((fmt_data*)binary_alloc)->blob` is not NULL
> 
> But it wouldn't check whether `(fmt_data*)binary_alloc)` is NULL, and would crash on NULL when trying to access `->blob`.

Oh, right. I'll re-visit that, thanks. **EDIT** no that was not the problem - the BLOB_FREE macro did check that before calling MEM_FREE. The problem was I did `BLOB_FREE(binary_alloc)` but I should have used `BLOB_FREE(binary_copy)` - the latter is sometimes aligned (to no more than `BINARY_ALIGN`) within the former.

src/wpapsk_fmt_plug.c
magnumripper: With my current hacks (everything is experimental at this point) these are only for **loader** to process pot file entries quicker (it did get WAY quicker with a couple hundred thousand WPA handshake samples). I added code to ignore these functions in cracker when there's no corresponding `get_hash` functions. Not sure if all sizes are required, I just threw them in there.

Yes, there is unused code that is #ifdef'ed out right now (although the `get_hash_*` functions were from shared code). It will all be cleaned up before we're anywhere near merging. I'm mostly interested in any caveats you might see with this non-hash scheme.

magnumripper:
> It is indeed good to be able to take advantage of matching ESSIDs. Perhaps we can also take advantage of matching salts with `openssl enc` blobs, in case someone specified a salt manually (unlikely?) and used the same more than once. Maybe there are more examples.

Not many, but at least a few. I believe Office will/may re-use the salt upon saving an updated document.

> If this interferes with reporting of tunable costs that significantly affect performance, that's really bad, though. For example, what will happen to the `dmg` formats?

As long as a salt-reuse wouldn't be exploitable anyway, we could include `header_ver` & `iter_cnt` in the "salt" but it wouldn't be any better (or worse) than keeping it as is, as far as I can see. We can leave most formats as-is because this new stuff doesn't really clash with the old way of doing it.

magnumripper:
This (the initial commit) is ripe for final reviews now. The rest are format-specific implementations of the new feature.

magnumripper:
I will "have to" merge this once complete or it'll be in the way for future work. If we really don't want it in Jumbo-2 I see no particular problem with branching off a "jumbo" branch with stable/wanted code (as defined by you @solardiz) while I just hack away a bit more freely in bleeding-jumbo (as its name suggests!). You'd get to cherry-pick stuff from bleeding-jumbo to jumbo if/when you want to. How about that?

solardiz:
> I will "have to" merge this once complete

I'm OK with having this merged into the main tree. Just make sure that, in your testing, this doesn't cause breakage of any format, including tunable costs reporting, etc. Thanks!

magnumripper:
Maybe I'll try converting oldOffice/Office as well before merging. Right now I'm not sure any other formats are worth converting.

src/wpapsk_fmt_plug.c
magnumripper: BTW `clear_keys` and `get_hash_*` are (and were) removed, but they were sitting in wpapsk.h

magnumripper:
The office formats could keep all costs features - they're left in the salt struct and this doesn't harm exploiting same-salt.

PR #4100, opened 2019-09-14
Reset CFLAGS at `-q` option test for darwin

catap:
I've discovered that at some cases clang at macOS Sierra and High Sierra may define `__SSE4_1__` but can't compile test case with error: `needs target feature sse4.1`.

As result we should have an option to switch SSE4.1 at this machines by `-mno-sse4.1` but when I did it I've broke test to determin does `-q` option requires because it requires `-mavx` to itself and `-mavx -mno-sse4.1` fails.

As result of this fails configure things that `-q` is required and adds `-Wa,-q` to each clang call and it brokes things at very differnet and wierd places.

Fixed this test by reseting for testing time `CFLAGS` to plain `-mavx`.

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -10889,7 +10889,8 @@ CFLAGS="$CFLAGS $SIMD_FLAGS -O0"
 
   case $host_os in #(
   darwin*) :
-    CC="$CC_BACKUP -mavx"
+    CC="$CC_BACKUP"
+    CFLAGS="-mavx"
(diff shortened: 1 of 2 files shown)


catap:
After this fix I can easy overstep https://github.com/magnumripper/JohnTheRipper/issues/4077 by adding to `CFLAGS='-mno-sse4.1'`

magnumripper (approved):

magnumripper:
Thanks!

PR #4101, opened 2019-09-14
Fixed homebrew openssl@1.1

catap:
A few change to fixed `OPENSSL_LIBS` and `OPENSSL_FLAGS` in general and at homebrew ;)

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -5222,11 +5222,15 @@ fi
 if test -d /opt/local/include; then
    ADD_CFLAGS="$ADD_CFLAGS -I/opt/local/include"
 fi
-if test -d /usr/local/opt/openssl/lib; then
-   ADD_LDFLAGS="$ADD_LDFLAGS -L/usr/local/opt/openssl/lib"
+if test -z "OPENSSL_LIBS"; then
+  if test -d /usr/local/opt/openssl/lib; then
+     ADD_LDFLAGS="$ADD_LDFLAGS -L/usr/local/opt/openssl/lib"
+  fi
 fi
-if test -d /usr/local/opt/openssl/include; then
-   ADD_CFLAGS="$ADD_CFLAGS -I/usr/local/opt/openssl/include"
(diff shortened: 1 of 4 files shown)


src/configure.ac
```diff
@@ -602,7 +602,7 @@ else
       [AC_MSG_CHECKING([supplied paths for OpenSSL])]
       [JTR_LIST_ADD(CFLAGS, [$OPENSSL_CFLAGS])]
       [if test "x$cross_compiling" = "xyes" -a "x$OPENSSL_LIBS" = "x"; then
-        OPENSSL_LIBS="-lssl -lcrypto"
+        OPENSSL_LIBS="${OPENSSL_LIBS} -lssl -lcrypto"
```
magnumripper, line 605: This will only happen when `$OPENSSL_LIBS` is empty (per line 604) so it's a no-op change

magnumripper (approved):

magnumripper:
Thanks!

src/configure
```diff
@@ -5222,11 +5222,15 @@ fi
...
-   ADD_LDFLAGS="$ADD_LDFLAGS -L/usr/local/opt/openssl/lib"
+if test -z "OPENSSL_LIBS"; then
+  if test -d /usr/local/opt/openssl/lib; then
+     ADD_LDFLAGS="$ADD_LDFLAGS -L/usr/local/opt/openssl/lib"
+  fi
 fi
-if test -d /usr/local/opt/openssl/include; then
-   ADD_CFLAGS="$ADD_CFLAGS -I/usr/local/opt/openssl/include"
+if test -z "OPENSSL_CFLAGS"; then
+  if test -d /usr/local/opt/openssl/include; then
```
magnumripper, line 5231: Both checks above are missing the `$` in `$OPENSSL_(...)`. Fixed in 695d367b

catap: Yay

PR #4102, opened 2019-09-19
Botcheck

magnumripper:


solardiz:
On Thu, Sep 19, 2019 at 03:33:17PM -0700, magnum wrote:
>   * Make --users/groups/shells options case insensitive and allow wildcard

I think the "case insensitive" part is a misfeature.

magnumripper:
> On Thu, Sep 19, 2019 at 03:33:17PM -0700, magnum wrote: * Make --users/groups/shells options case insensitive and allow wildcard
> I think the "case insensitive" part is a misfeature.

Depending on use case, I agree. So in order to keep this simple, how about we make it case insensitive only when a `*` wildcard is present?

solardiz:
> how about we make it case insensitive only when a `*` wildcard is present?

On one hand, this is good enough when you know about that property. On the other hand, it's a subtle extra thing to learn, and most people would be confused by it at first. So I think we should either not implement the case-insensitivity at all or have it controlled separately from any other functionality. Even a john.conf setting would be better.

src/cracker.c
```diff
@@ -477,11 +481,18 @@ static char *crk_loaded_counts(void)
...
+	                crk_db->password_count,
+	                crk_db->salt_count > 1 ?
+	                jtr_itoa(crk_db->salt_count, nbuf, 24, 10) : "no");
+
+	int b = (10 * crk_db->password_count + (crk_db->salt_count / 2)) /
+		crk_db->salt_count;
+
+	if (crk_db->salt_count > 1 && b > 10)
+		sprintf(s_loaded_counts + p, " (%d.%dx same-salt boost)",
+		        b / 10, b % 10);
```
solardiz, line 495: `sprintf` isn't supposed to fail, but formally it's defined to return a negative value on error. If the first `sprintf` call here fails, then the second one will use an out-of-bounds pointer. I think we should avoid this coding pattern, perhaps by checking that `p > 0` before attempting the second `sprintf`

src/john.c
```diff
@@ -890,17 +890,24 @@ static void john_mpi_wait(void)
...
+	                database.password_count,
+	                database.salt_count > 1 ?
+	                jtr_itoa(database.salt_count, nbuf, 24, 10) : "no");
+
+	int b = (10 * database.password_count + (database.salt_count / 2)) /
+		database.salt_count;
+
+	if (database.salt_count > 1 && b > 10)
+		sprintf(s_loaded_counts + p, " (%d.%dx same-salt boost)",
+		        b / 10, b % 10);
```
solardiz, line 910: Ditto re: `sprintf` behavior on error. Also, let's avoid the code duplication.

src/cracker.c
magnumripper: Good point, I'll look into that

magnumripper:
> > how about we make it case insensitive only when a `*` wildcard is present?
> 
> On one hand, this is good enough when you know about that property. On the other hand, it's a subtle extra thing to learn, and most people would be confused by it at first. So I think we should either not implement the case-insensitivity at all or have it controlled separately from any other functionality. Even a john.conf setting would be better.

Until now, nothing in the docs mentioned it **is** case sensitive in the first place. The fact it was might perhaps have been expected for UNIX accounts but certainly not for eg. Windows accounts and many other ones. I, for one, always thought it was case insenstive.

Also, I want to keep this simple.

PR #4131, opened 2019-11-01
Disable OpenCL on ASAN builds

ghost:
We can't run ASAN tests if OpenCL is enabled. In fact, I would say that we are detecting bugs in OpenCL runtimes not in JtR.


ghost:
BTW: the new Docker image has OpenCL available.

magnumripper (approved):

PR #4140, opened 2019-11-06
Use a #define to "inline" in OpenCL

ghost:
Testing now. It contains the revert and (the last commit) the `#define`.

### DIFF
--- a/run/opencl/7z_kernel.cl
+++ b/run/opencl/7z_kernel.cl
@@ -46,7 +46,7 @@
 		} \
 	}
 
-static inline void sha256_zerofinal(uint *W, uint *output, const uint tot_len)
+inline void sha256_zerofinal(uint *W, uint *output, const uint tot_len)
 {
 	uint len = ((tot_len & 63) >> 2) + 1;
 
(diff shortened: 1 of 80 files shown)


ghost:
It seems it will succeed. So, pending is ROCM (in case anyone complained directly to you, the guy should test now).

magnumripper:
I added a few changes. Right now it uses 'static inline' for any AMD that isn't POCL or MESA. Should work fine.

magnumripper (approved):

ghost:
POCL is unhappy when it sees the re-definition of `inline`. I'm thinking about doing:

```
#elif __POCL__
// Do nothing
```

magnumripper:
> POCL is unhappy when it sees the re-definition of `inline`. I'm thinking about doing:
> 
> ```
> #elif __POCL__
> // Do nothing
> ```

That's fine (with the comment!)

magnumripper:
Actually we should do the same (nothing!) as default!

ghost:
> Actually we should do the same (nothing!) as default!

:+1:

magnumripper:
I think I have beignet installed somewhere as well. And we should test all drivers on Solar's test rigs.

ghost:
> I think I have beignet installed somewhere as well. And we should test all drivers on Solar's test rigs.

I have a "script" testing that does that (manual, since I can't add Solar's stuff to CI).

magnumripper:
You might want to squash all commits for opencl_misc.h into a single one before we merge.

ghost:
> You might want to squash all commits for opencl_misc.h into a single one before we merge.

I almost did that, but then I thought more commits tell the story better.

PR #4144, opened 2019-11-10
ssh2john: python3 compatibility

exploide:
I ported `ssh2john` to python3. Actually only three lines are changed :yum:

### DIFF
--- a/run/pem2john.py
+++ b/run/pem2john.py
@@ -125,7 +125,7 @@ if __name__ == "__main__":
         blob = open(filename, "rb").read()
         if b'-----BEGIN ENCRYPTED PRIVATE KEY-----' not in blob:
             if b'PRIVATE KEY-----' in blob:
-                sys.stderr.write("[%s] try using sshng2john.py on this file instead!\n" % sys.argv[0])
+                sys.stderr.write("[%s] try using ssh2john.py on this file instead!\n" % sys.argv[0])
             else:
                 # try as DER payload
                 ret = unwrap_pkcs8_data(blob)
--- a/run/ssh2john.py
+++ b/run/ssh2john.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 # Copyright (C) 2012, Dhiru Kholia <dhiru@openwall.com>
 # Copyright (C) 2015, Dhiru Kholia <dhiru@openwall.com>
@@ -100,7 +100,7 @@ def read_private_key(filename):
     # if we trudged to the end of the file, just try to cope.
     try:
         data = ''.join(lines[start:end]).encode()
-        data = base64.decodestring(data)
+        data = base64.decodebytes(data)
     except base64.binascii.Error:
         e = sys.exc_info()[1]
         raise Exception('base64 decoding error: ' + str(e))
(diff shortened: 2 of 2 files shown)


solardiz:
IMO, we need these scripts to work with _both_ Python 2 and Python 3. Turning a Python 2 only script into Python 3 only isn't obviously an improvement yet, as many systems still have Python 2 and where systems have both there are often differences in available library versions - e.g., I found this especially important with differing Berkeley DB versions for `bitcoin2john.py` (a different script, but I think we should use a common approach to those).

@exploide Can you make changes to have this script work with either Python version? And please test those with both versions on test files at https://openwall.info/wiki/john/sample-non-hashes#SSH - make sure the script's output stays the same. IIRC, this script can also process non-SSH PEM files - please try this as well: https://openwall.info/wiki/john/sample-non-hashes#PEM

Thanks!

exploide:
Thanks for your fast response.

Is cross-python-version compatibility a requirement for this project? Then I can try to accomplish your proposed changes.

But to be honest, I do not see the advantage of this. I don't know of fairly modern systems that do not have python3, so I think it should be available. If both, python2 and python3 are available, it's still fine since the script has the appropriate shebang. And maintaining both, python2 and python3 compatibility is definitely more work and bloats the code (generally, not for this script in particular). Regarding that python2 will be EOL in two month, I don't see why this is worth it. You might have a point with the library versions, but this script has only standard imports.

I noticed there are already other python3 only scripts in John.

solardiz:
For new scripts, we accept whatever we're given - it's better to have a Python 2 only or a Python 3 only script than no implementation at all. For existing scripts, so far we haven't been converting Python 2 only to Python 3 only, but we have enhanced some scripts to work with either Python version. We haven't ever made it "a requirement for this project", but I think we also haven't deviated from this practice so far.

exploide:
All right, I made the script py2 and py3 compatible.

I checked that for all SSH sample files, the old version, the new with py2 and the new with py3 produce the same results.

The PEM sample file could not be processed, even with the old version. However `pem2john.py` works on this one. While testing this, I noticed `pem2john.py` included a reference to the old name of `ssh2john.py`. I fixed that as well.

solardiz (approved):
I trust your testing, so this is good now. Thanks!

My only concern is the added dependency on the "codecs" module - I don't know if it's always installed or might have to be installed by a user. Do you? If "codecs" is not always available, then changes similar to the first commit here but augmented with Python version checks and fallbacks to the existing code would be more portable.

exploide:
`codecs` is a standard import, available in both, python2 and python3. Hence, it is available on every "normal" Python installation.

However, on systems where python (and other packages) are intentionally kept small, for example OpenWrt routers, the `codecs` module, despite being part of the standard library, needs to be installed via a separate package. However, this is also true for most other parts of the library on such systems, so you would have the same trouble with all the other imports in this and other scripts ;)

solardiz:
OK, I'm fine merging this now, but I'd rather let @magnumripper decide on whether to do a merge commit, a rebase of the 3 commits, or a squash and merge. To me, it makes sense to keep your initial commit somewhere in the project's history, such as in case we later choose to support Python 3 only.

magnumripper (approved):
Groovy

magnumripper:
@exploide, thanks! Just so you know, we are SERIOUSLY lacking contributors right now so any future contributions you might come up with just might be what saves us from giving up. Kudos.

solardiz:
@exploide Is the exception we catch here still the right one after the base64 decode line update or does it maybe also need to be updated? See https://github.com/openwall/john/issues/4564#issuecomment-2901567550

exploide:
The function has been changed to `base64.b64decode()` in #4801 anyway (don't know why I didn't used that one directly in this PR). But `base64.binascii.Error` has always been the correct exception subclass.

Issue #4143, opened 2019-11-08
bitcoin2john.py Python 3 compatibility


solardiz:
A revision of `bitcoin2john.py` (per #4142) when run on an altcoin wallet with Python 2 produces a "hash", whereas when run with Python 3 says "this wallet is not encrypted!" (missing a linefeed after that message - something we should fix as well).

This could indicate a Python 3 compatibility issue in the script itself, or it might be a side-effect of differences in Berkeley DB versions in use by these two Python builds on this system (although I'd have expected the `db.open` call to fail, like it sometimes does, if the DB type is unsupported). To get the script to work with Python 3 on this throw-away Fedora VM, I had to explicitly install Berkeley DB support for it:

```
dnf install libdb-devel python3-devel
pip3 install bsddb3
```

I hope this is right. For Python 2, the system had working Berkeley DB support out of the box.

Also, the uses of `print` in `if __name__ == '__main__':` are Python 2 specific.

Related: #3591, #4085.

solardiz:
I'm also seeing the same issue with our sample wallets from the wiki, so I hope anyone can reproduce this.

@exploide I'd appreciate your help on this issue after we merge #4145. Thanks!

solardiz:
I started looking into this issue, but with my current non-knowledge of Python I don't intend to complete fixing it - I'd rather leave that to someone familiar with Python, who would manage to have this work with both Python versions cleanly.

The main issue appears to be Python 3's explicit byte strings, so e.g. what used to be `ckey` in Python 2 becomes `b'ckey'` in Python 3, and the if's fail to match it. When I update the if's to match, a further hurdle is that there's no `encode('hex_codec')` in Python 3's byte strings.

If we somehow don't fix this script soon, then we should have it fail on Python 3 with a more reasonable message - saying the script needs Python 2 rather than wrongly claiming the wallet is not encrypted.

exploide:
@solardiz I can give it a try once I find a bit time, though I have zero knowledge of wallets.

I will base my work on your current `bitcoin2john-cleanup2` branch. Are these the relevant test files? https://openwall.info/wiki/john/sample-non-hashes#Bitcoin-Core-wallet-files

[...]

PR #4145, opened 2019-11-10
bitcoin2john cleanup

### LINKED ISSUE #4142: bitcoin2john.py unneeded parsing
@kholia's `bitcoin2john.py` is based on a third-party Bitcoin wallet parsing script, and it still contains a lot of that parsing, whereas we only use some fields for our script's output. The extra parsing sometimes unnecessarily causes the script to fail on altcoin wallets.

Specifically, `parse_wallet` supports a lot of item types, including e.g. transactions (complicated, but irrelevant to our task). In my very limited testing, it appears to be sufficient to support only `ckey` and `mkey` item types, and only invoke `item_callback` for those rather than for all - or alternatively `item_callback` can be invoked unconditionally like it currently is, but its body revised to silently ignore all other item types. A few other item types might also need to be recognized for other wallet types than what I tested this on - we can keep everything related to keys.

Similarly, `item_callback` can be simplified to only recognize the supported item types.

This should also let us drop some other functions, such as `inversetxid`.

solardiz:
Fixes #4142 and paves the way for #4143.

I've tested this with 3 wallets: 2 samples from our wiki, where this produces the same output as before, and an altcoin wallet where the script used to fail but after these changes produces a "hash" that JtR loads.

I think the two commits here shouldn't be squashed - I kept them separate on purpose.

### DIFF
--- a/run/bitcoin2john.py
+++ b/run/bitcoin2john.py
@@ -28,17 +28,12 @@
 # Alex Martelli (http://www.aleax.it)
 # Ported from C code written by Laurent Haan (http://www.progressive-coding.com)
 
-missing_dep = []
-
 try:
         from bsddb.db import *
 except:
         from bsddb3.db import *
-        # missing_dep.append('bsddb')
 
 import os, sys, time
-pyw_filename = sys.argv[0].split('/')[len(sys.argv[0].split('/')) - 1]
-pyw_path = os.getcwd()
 
 try:
         import json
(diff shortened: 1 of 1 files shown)


ghost (approved):
I'm not a pythonist, anyway, If you need this merged, go ahead.

magnumripper (approved):
I'm not very good at Python but I like the ratio of deleted vs. added lines 🤣

PR #4146, opened 2019-11-11
bitcoin2john further cleanup

solardiz:
This passed the same tests as I used for #4145. Again, the two commits should probably remain separate.
 
I think even further cleanup is possible - perhaps there's no need to parse item types other than "ckey" and "mkey", which would also eliminate the need to have hashing and base58 decoding in here. However, dropping processing of other key-related item types would _potentially_ affect the output for some wallets that I don't have, so I don't dare to make that change yet.


solardiz:
Can someone review this, please? I'd like to proceed with merging this and the other PR, and then make further changes. Thanks!

ghost (approved):

PR #4152, opened 2019-11-12
Add Nexus legacy wallet support

### LINKED ISSUE #4149: Bitcoin format assumes alignment on 64-bit SIMD, uses uninitialized IV
This code snippet from `bitcoin_fmt_plug.c`:

```c
                        AES_KEY aes_key;
                        unsigned char key[32];
                        unsigned char iv[16];

                        // Copy and convert from SIMD_COEF_64 buffers back into flat buffers, in little-endian
#if ARCH_LITTLE_ENDIAN==1
                        for (i = 0; i < sizeof(key)/sizeof(uint64_t); i++)  // the derived key
                                ((uint64_t *)key)[i] = JOHNSWAP64(key_iv[SIMD_COEF_64*i + (index2&(SIMD_COEF_64-1)) + index2/SIMD_COEF_64*SHA_BUF_SIZ*SIMD_COEF_64]);
#else
                        for (i = 0; i < sizeof(key)/sizeof(uint64_t); i++)  // the derived key
                                ((uint64_t *)key)[i] = key_iv[SIMD_COEF_64*i + (index2&(SIMD_COEF_64-1)) + index2/SIMD_COEF_64*SHA_BUF_SIZ*SIMD_COEF_64];
#endif

                        AES_set_decrypt_key(key, 256, &aes_key);
                        AES_cbc_encrypt(cur_salt->cry_master + cur_salt->cry_master_length - 32, output, 32, &aes_key, iv, AES_DECRYPT);
```

has some issues:

1. The casts to `(uint64_t *)` violate C aliasing rules.

2. The writes through those casts assume that either `key` just happens to be aligned or that the architecture allows efficient unaligned writes. This might fail on 64-bit POWER.

3. The `iv` variable is uninitialized, but is then passed to `AES_cbc_encrypt`. The reason this works at all, I think, is that CBC mode compensates for that, and we only check padding on the very last block. So we don't actually need the IV - we only need the penultimate block. Yet we should initialize the `iv` to something, or maybe just reuse `key` in its place, as formally usage of uninitialized memory triggers undefined behavior.

solardiz:
Also fixes #4149 (although there's much more to cleanup), fixes an instance of undefined behavior in a shared function (where a pointer could be set to outside of an object), and drops `FMT_HUGE_INPUT` (unneeded here). I've only tested the SIMD code. Let's wait for CI to test the scalar code as well (does it?)

Similar PR to hashcat: https://github.com/hashcat/hashcat/pull/2222

### DIFF
--- a/src/bitcoin_fmt_plug.c
+++ b/src/bitcoin_fmt_plug.c
@@ -104,6 +104,8 @@ static struct fmt_tests bitcoin_tests[] = {
 	{"$bitcoin$96$d98326490616ef9f59767c5bf148061565fe1b21078445725ef31629e8ee430bf4d04896d5064b6651ab4c19021e2d7c$16$51ee8c9ab318da9e$46008$96$819f6c8e618869c7933b85f6c59d15ca6786876edc435ba3f400e272c2999b43e0e3cda27acd928d1adbccd01b613e66$66$03feefa49b8cbbdbb327b7c477586e4a3275132cf6778f05bc11c517dc2e9107cb", "openwall"},
 	// Truncated PRiVCY hash
 	{"$bitcoin$64$65fe1b21078445725ef31629e8ee430bf4d04896d5064b6651ab4c19021e2d7c$16$51ee8c9ab318da9e$46008$96$819f6c8e618869c7933b85f6c59d15ca6786876edc435ba3f400e272c2999b43e0e3cda27acd928d1adbccd01b613e66$66$03feefa49b8cbbdbb327b7c477586e4a3275132cf6778f05bc11c517dc2e9107cb", "openwall"},
+	/* Nexus legacy wallet */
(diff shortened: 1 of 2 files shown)


solardiz:
Tested scaler non-OMP:
```
[solar@super src]$ make -f Makefile.legacy -sj32 generic
[solar@super src]$ ../run/john -test -form=bitcoin
Benchmarking: Bitcoin, Bitcoin Core [SHA512 AES 64/64]... DONE
Speed for cost 1 (iteration count) of 177864
Raw:    11.4 c/s real, 11.5 c/s virtual
```

solardiz:
Tested scalar OMP:
```
[solar@super src]$ ../run/john -test -form=bitcoin
Will run 32 OpenMP threads
Benchmarking: Bitcoin, Bitcoin Core [SHA512 AES 64/64]... (32xOMP) DONE
Speed for cost 1 (iteration count) of 177864
Raw:    141 c/s real, 5.7 c/s virtual
```

magnumripper (approved):

PR #4153, opened 2019-11-12
Fix legacy build (e.g., "make -f Makefile.legacy generic")

solardiz:
...which was broken (again) by a16ffedb1ad314dc1996141208753f39c0ec3c10 and 838d1ed7688a662708f308b02e54004cc67e8aeb.


magnumripper (approved):

magnumripper:
As long as we don't ditch legacy builds, we should have one in CI

PR #4147, opened 2019-11-11
bitcoin2john: python 3 compatibility

exploide:
Made `bitcoin2john.py` python3 compatible. fixes #4143

I targeted the current `bitcoin2john-cleanup2` branch for this Pull Request.

This version of the script produces the same output than before, but this time, working with python2 as well as python3. I tested with the two test wallets from https://openwall.info/wiki/john/sample-non-hashes#Bitcoin-Core-wallet-files

However, I don't know if I missed some edge case or what happens with other wallet types. It would be great if someone could test a little bit more.

BTW, the script had inconsistent indentation and still has. I would have cleaned that up but wanted to keep the diff manageable. Should I do this now in a further commit?

PR #4155, opened 2019-11-13
bitcoin2john: python 3 compatibility

### LINKED ISSUE #4143: bitcoin2john.py Python 3 compatibility

exploide:
bitcoin2john: python 3 compatibility; fixes #4143

New PR due to misunderstanding in #4147.

@solardiz do you already have your additional changes present (as code) or is this in a planning state? If the latter I would do the code cleanup commit now. Otherwise I will wait until you are done.

### DIFF
--- a/run/bitcoin2john.py
+++ b/run/bitcoin2john.py
@@ -17,19 +17,28 @@
 # https://github.com/gavinandresen/bitcointools
 # Copyright (c) 2010 Gavin Andresen
 
-try:
-        from bsddb.db import *
-except:
-        from bsddb3.db import *
-
-import sys
+import binascii
+import hashlib
 import logging
 import struct
+import sys
 import traceback
-import hashlib
+
+try:
+    from bsddb.db import *
+except:
+    try:
+        from bsddb3.db import *
+    except:
+        sys.stderr.write("Error: This script needs bsddb3 to be installed!\n")
+        sys.exit(1)
+
 
 addrtype = 0
 json_db = {}
+
+def hexstr(bytestr):
+    return binascii.hexlify(bytestr).decode('ascii')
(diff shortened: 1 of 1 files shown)


solardiz:
Thanks, @exploide. I already have my "additional changes present (as code)" - in fact, as several revisions that I intend to commit after merging this PR. So please postpone your code cleanup - in fact, it might not even be needed in the end, as my commits will reduce the number of lines of code much further.

exploide:
Ok, great. Then go ahead :)

solardiz (approved):
I assume these are exactly the same changes as we had in the previous incarnation of this PR.

exploide:
In fact it is the cherry-picked commit from my previous branch ;)

solardiz:
I'm now waiting for CI to complete before I can merge this, due to this repo's settings...

PR #4156, opened 2019-11-13
Further rework of bitcoin2john.py

solardiz:
Having read and understood and revised the Bitcoin format in JtR and mode 11300 in hashcat, I now have better understanding of what's needed from this script. Thus, I finally dare to drop all parsing of items other than `mkey`, since that's the only one we currently use for cracking. This dropping also makes the "hashes" much less security sensitive.

Besides dropping stuff, I now also add stuff (and thus my copyright): checks that the wallet's key derivation method, salt size, and master key size are as expected by us. I also add a lengthy comment explaining the rationale for insisting on specific known master key sizes.

Like before, I'd like my 4 commits to remain separate in our tree. As I understand, magnum has been using "rebase and merge" for that lately.

After these changes, the only cleanup I'm aware still needs to be done is of indentation, which I leave to @exploide, who might also have other cleanups pending.

### DIFF
--- a/run/bitcoin2john.py
+++ b/run/bitcoin2john.py
@@ -18,11 +18,9 @@
 # Copyright (c) 2010 Gavin Andresen
 
 import binascii
-import hashlib
 import logging
 import struct
 import sys
-import traceback
 
 try:
     from bsddb.db import *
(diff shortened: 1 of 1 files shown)


exploide:
Great, looks good to me :+1: (It's good you managed to get rid of the base56 encoding function and more. I was not proud of the way I made that one Python3 compatible :sweat_smile: )

After this is merged I will see if indentation cleanup and such would be reasonable.

magnumripper (approved):

PR #4173, opened 2019-12-17
Do not load john-local.conf from current directory

solardiz:
This must be disabled by default since it's a security risk in case JtR is ever run with untrusted current directory.

### DIFF
--- a/doc/CONFIG
+++ b/doc/CONFIG
@@ -191,32 +191,32 @@ Idle = N
 Editing the local file is good for setting custom local options, for
 setting items specific to this machine (i.e. [Local:Options:OpenCL])
 for setting things like MinLen and MaxLen within the incremental
-sections, and setting things like local Markov settings. Yes, each of
+sections, and setting things like local Markov settings.  Yes, each of
 these can be done by editing john.conf.  However, keep in mind that
-all of these changes to a john.conf file 'can' be lost. If they are
+all of these changes to a john.conf file 'can' be lost.  If they are
 modified within john-local.conf then they are safe from being lost
 during an update.
 
 This file (john-local.conf) can be located in the $JOHN directory
-(where the john executable resides). It can also be in the local
-working directory, AND it can even be in both places.  These files
-will be loaded in this order:
+(where the john executable resides).  With non-default john.conf,
+it can also be in the local working directory, AND it can even be in both
+places.  These files will be loaded in this order:
    1. all of $JOHN/john.conf
    2. $JOHN/john-local.conf
-   3. ./john-local.conf
+   3. ./john-local.conf (only if enabled in john.conf)
 So each john-local.conf can add, or modify the existing config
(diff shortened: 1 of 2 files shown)


solardiz:
`keepass-opencl` fails test on Travis CI - do we perhaps rely on this reading of the config file from the current directory to disable(?) that format on Travis? Can we avoid this?

ghost:
>do we perhaps rely on this reading of the config file from the current directory to disable(?) that format on Travis? Can we avoid this?

Yes and yes.

ghost (approved):

ghost:
You'll have to update your branch with "my changes" or wait magnum to merge for you.

solardiz:
Thanks, Claudio!

I don't know where your changes are, and prefer to wait for magnum anyway.

magnumripper (approved):
I assume we just need to write those CI tests' `john-local.conf` to the `run` directory.

magnumripper:
Don't we have an issue for this as well?

ghost:
> Don't we have an issue for this as well?

I'm testing a fix (after the PR) right now.

ghost:
There is a side effect of this commit. 
* What an user using JtR from a distro package should do to override defaults from john.conf?
* $JOHN is a read-only folder. She/he can't create the file there.
* I see no other place the file will be loaded.

magnumripper:
> There is a side effect of this commit.
> 
> * What an user using JtR from a distro package should do to override defaults from john.conf?
> * $JOHN is a read-only folder. She/he can't create the file there.
> * I see no other place the file will be loaded.

Good catch. For systemwide builds we should try loading it from `JOHN_PRIVATE_HOME` (eg. `~/.john`)

solardiz:
I've just added documentation updates to my one commit here (force-pushed). I didn't know this was documented.

As to system-wide installs, I think `~/.john` should readily work. Doesn't it?

While at it, I also found this in `config.c`:

```c
        if (strstr(Name, "/john-local.conf")) {
                if (!strcmp(Name, "$JOHN/john-local.conf") ||
                    !strcmp(Name, "./john-local.conf")) {
                        if (!strcmp(path_expand("$JOHN/"), "./") &&
                            cfg_loaded_john_local)
                                return 0;
                        else
                                cfg_loaded_john_local = 1;
                }
                cfg_loading_john_local = 1;
        }
```

I don't like the use of `strstr` here - we should check for the pathname ending in that string, not merely having that substring.

ghost:
> As to system-wide installs, I think ~/.john should readily work. Doesn't it?

I looked at the source code and run a system wide (not a real packaging) test and: it does not work. It opens `john.conf` from `~/.john`. But not the `john-local.conf` file. 

* It opens `john-local.conf`only from $JOHN

BTW, I used this branch as it is now.

magnumripper:
Merged manually

PR #4174, opened 2019-12-17
CI: do not create john-local.conf in the current directory

ghost:
Everything seems ok:
- https://github.com/magnumripper/JohnTheRipper/pull/4173#issuecomment-566732570 is another "invitation" to move to Openwall's repo and make Solar one of the Masters of the Universe.

BTW:
- good news: Travis has better hardware available for us now (at least in `.com`).
  - I'll do more tests using my account (still a `.org`).

### DIFF
--- a/.ci/disable_formats.sh
+++ b/.ci/disable_formats.sh
@@ -2,12 +2,12 @@
 
 # There was a bug in echo -e in Travis
 # TODO: we know these formats must be fixed (or removed)
-echo '[Local:Disabled:Formats]' > john-local.conf
-echo 'Raw-SHA512-free-opencl = Y' >> john-local.conf
-echo 'XSHA512-free-opencl = Y' >> john-local.conf
-echo 'gpg-opencl = Y' >> john-local.conf
-echo 'KeePass-opencl = Y' >> john-local.conf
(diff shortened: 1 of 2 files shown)


magnumripper (changes requested):
Just one detail: Don't we have it documented that current directory john-local.conf overrides everything? I think we do so the documentation needs an update

ghost:
> Just one detail: Don't we have it documented that current directory john-local.conf overrides everything? I think we do so the documentation needs an update

If so, that should be in Solar's commit, right?

magnumripper:
I don't really care which commit, or a separate one, but all this should be in one same PR once merged.

solardiz:
Can we please limit this PR to only changes to `.ci/disable_formats.sh` and get it merged already?

solardiz:
> should be in one same PR once merged.

I agree the default `john.conf` change and the documentation update should be in the same PR (and they are in my PR now), but I think the Travis update can be in this one separate PR by Claudio.

solardiz:
Claudio, can you please remove my "Do not load john-local.conf from current directory" commit from this PR, as I revised this commit in its other PR (before noticing you had included it in here) to include the documentation update in it?

ghost:
I'll do it after #4178 due to git conflicts.

PR #4178, opened 2019-12-18
CI: disable OpenCL formats that run slowly

ghost:
* some formats are really bad inside CI
* I bet we will gain more than 5 minutes skipping these 3 formats

### DIFF
--- a/.ci/disable_formats.sh
+++ b/.ci/disable_formats.sh
@@ -8,6 +8,25 @@ echo 'XSHA512-free-opencl = Y' >> john-local.conf
 echo 'gpg-opencl = Y' >> john-local.conf
 echo 'KeePass-opencl = Y' >> john-local.conf
 
-# These formats fails OpenCL CPU runtime
+# These formats fail OpenCL CPU runtime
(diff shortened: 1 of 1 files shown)


ghost:
I'm afraid this PR needs to be merged. Timeouts were frequent yesterday and today.

solardiz:
Maybe we can speed the testing of these formats up by forcing a low GWS?

ghost:
> Maybe we can speed the testing of these formats up by forcing a low GWS?

We do it already. E.g.: Only a few formats do not use shared OpenCL code (and they are ok).

```
$ ../run/john --test=0 -verb=6 --format=ethereum-opencl --dev=4
[...]
TestDB LWS=7 GWS=49 (7 blocks) PASS

```

magnumripper:
Not sure why these formats are so ridiculously slow on CPU device? However, we could try changing the `LWS=7, GWS=49` to eg. `LWS=1, GWS=7` for CPU devices.

magnumripper:
Merged manually. We should look into the "lower worksizes for CPU device" idea.

PR #4185, opened 2019-12-22
Fix PKCS#12 PBE Unicode bugs.  Closes #4179.

### LINKED ISSUE #4179: Our PKCS#12 PBE functions "convert" to UTF-16BE by casting to uint16_t
Apparently our PKCS#12 PBE functions (derived from PolarSSL/mbedTLS) are written by someone believing a UTF-16 conversion is done by inserting nulls to a normal string!? Either that, or actual implementations (eg. Bestcrypt, BKS (BouncyCastle), PFX, SAP-PSE, Signal, Zed Archives) all do exactly that? But, then, what is the expected input? Perhaps non-Latin1 passwords are rejected? Perhaps they all actually use UTF-8 interleaved with nulls (on Windows too? I doubt that).

I'm not sure which - but need to figure it out. I'll dig into OpenSSL and others. **All** formats mentioned above need test vectors added, containing characters outside ISO-8859-1.

We've had this bug (provided it is a bug) ever since June 2016 (directly imported from mbedTLS).

🙄

magnumripper:


magnumripper:
This will be included in #4182

PR #4191, opened 2019-12-29
Add Gitpod config

annonymau5:


ghost (changes requested):
* 100% safe to merge.
* we have 50 hours/month. We can apply for **a free unlimited open source plan**.
* not sure if the image has everything it should.
* it is interesting, but I'm not aware of people that is using it.
  * some introduction text may help.
  * I use VS code myself. So, I'm far from being against it.

ghost:
Are you sure the Dockerfile has anything it should? SSL, zLib, pcap, OpenCL, ...?

I would say you should "sell" your point of view. Why accept the PR? It is interesting, but I don't see any JtR developers using it right now (I could be wrong, not sure about it). [1]
****
BTW, why highlight the IDE/site?

image

****
**[edited]**
[1] Maybe Solar and magnum are fans of _k-pop_ and IDEs nowadays. Who knows!? 

* I have no idea what is _k-pop_, but I love IDEs.

magnumripper (commented):
I see no harm in this PR, but I have absolutely no idea what Gitpod is and have no time investigating it.

magnumripper:
OP lost by WO

PR #4193, opened 2020-01-03
Fix ASCII percentage check for data >= 256 bytes & add related tests.

falaca:
The following call to check_pkcs_pad() is problematic if cur_salt->datalen is 256:

https://github.com/magnumripper/JohnTheRipper/blob/03177817f8753a0205859c366e8513cfa37ea3de/src/openssl_enc_fmt_plug.c#L322

Reasoning:

1) If the original plaintext is >= 256 bytes, it will be PKCS#5 padded to at least 272 bytes before encryption, so cur_salt->data will never contain the last block of ciphertext, since openssl2john.py only extracts the first 256 bytes of the encrypted file. Therefore, the function call will return -1 since no valid padding is found. This causes the ASCII percentage check to fail, even though it shouldn't, since only the last block of ciphertext should be expected to contain the padding.

2) On the other hand, if the original plaintext is between 240 and 255 bytes, it will be PKCS#5 padded to 256 bytes before being encrypted. In this case, cur_salt->datalen will still be 256 bytes, but it will be impossible to distinguish from the first case described above (without knowing the actual size of the encrypted file, which is not present in the format string generated by openssl2john.py).

Considering the above two scenarios, it seems sensible to check for valid padding by default (to make the ASCII percentage calculation more accurate) and then only fall back to using cur_salt->datalen (e.g., without subtracting any padding length) in the event that no padding is found.

I've fixed the bug by adding the following after the line referenced above:

```
if (len == -1 && cur_salt->datalen >= 256)
	len = cur_salt->datalen;
```

The above check would still correctly result in `kpa()` returning -1 if cur_salt->datalen < 256 and no valid padding is found, but this would never actually occur since `decrypt()` already checks for valid padding in cur_salt->last_chunks before calling `kpa()`. So the `&& cur_salt->datalen >= 256` part of the condition can be removed with no consequence, if desired.

A slightly nicer fix is possible if openssl2john.py is updated to output the size of the encrypted file: `decrypt()` should pass the padding length as an argument into kpa(), and if cur_salt->datalen is less than the encrypted file size, `kpa()` should set len = cur_salt->datalen; otherwise (i.e., if cur_salt->datalen equals the encrypted file size), set len = cur_salt->datalen - paddinglen.

I've also added a few relevant test cases that all pass when the fix is applied.


magnumripper:
Interesting, thanks! I'll look into this sooner or later.

ghost (approved):
It passes all (my) tests and makes sense to me.

However, I'll wait magnum (or Dhiru).

magnumripper (approved):

magnumripper:
Good stuff, thanks!

PR #4198, opened 2020-01-08
Disable binary caching for AMD GPUs

ghost:
I am not entirely confident that the change is really necessary. Still, it's safe.


magnumripper (commented):
I'll review (test) the original issue and alternate solutions. The binary caching works and does wonders in start-up time on my macbook. Perhaps we should just mute that specific error message in some cases.

magnumripper:
BTW where is the original issue? john-users only?

ghost:
> BTW where is the original issue? john-users only?

Yes.

PR #4202, opened 2020-01-24
add second block data in office2john output for 40bit RC4 keys

philsmd:
As mentioned here https://github.com/magnumripper/JohnTheRipper/issues/4116 and explained/discussed also on the hashcat github issues https://github.com/hashcat/hashcat/issues/2067 , we have a problem with the current office2john.py output for $oldoffice$3 hashes because the keys for the RC4 decrypt are very weak (40 bit).

What we discovered over at the hashcat issue, is that the problem is that each block of (document) data needs to be decrypted with a different key (because the RC4 key depends on the block number too).

This could lead to passwords that lead to the correct key for the first block, but fail to also work for e.g. the second block.

It seems that block 0 (e.g. the key we need also for verifying encryptedVerifierHash) and block 1 (a data block) is enough to prevent very many false results. Therefore, I suggest here to make these changes to office2john to add 32 bytes of encrypted data from the document and optionally use them in the crackers (jtr/hashcat) to perform a final rejection test (shouldn't affect the speed).

This commit for now only adds some code to office2john.py... code needed for the fmt_plug needs to be added too (but I hope a jtr dev could first give some feedback and implement this eventually, if accepted).

We have already some suggestions on how to perform the check within the crackers see https://github.com/philsmd/hashcat/commit/cb67e0b2e2e3952494877ef4a19ceaf00c727925 , i.e. add a NUL byte count for the raw (decrypted) data, which was always correct in my tests... but please also perform some tests with your example (https://openwall.info/wiki/john/sample-non-hashes#Sample-files-created-using-LibreOffice-and-MS-Office) or even further examples.
Thx

### DIFF
--- a/run/office2john.py
+++ b/run/office2john.py
@@ -2432,10 +2432,23 @@ def find_rc4_passinfo_xls(filename, stream):
                 verifierHashSize = unpack("<I", stm.read(4))[0]
                 assert(verifierHashSize == 20)
                 encryptedVerifierHash = stm.read(verifierHashSize)
-                sys.stdout.write("%s:$oldoffice$%s*%s*%s*%s\n" % (os.path.basename(filename),
+
+                second_block_extra = ""
+                if typ == 3:
+                    offset_cur = stream.tell()
+                    assert(offset_cur < 1024)
+
+                    skip = 1024 - offset_cur
+                    stream.read(skip) # ignore remaining bytes of 1st block
+
+                    second_block_bytes = stream.read(32)
+                    second_block_extra = "*%s" % binascii.hexlify(second_block_bytes).decode("ascii")
+
+                sys.stdout.write("%s:$oldoffice$%s*%s*%s*%s%s\n" % (os.path.basename(filename),
                     typ, binascii.hexlify(salt).decode("ascii"),
                     binascii.hexlify(encryptedVerifier).decode("ascii"),
-                    binascii.hexlify(encryptedVerifierHash).decode("ascii")))
+                    binascii.hexlify(encryptedVerifierHash).decode("ascii"),
+                    second_block_extra))
 
     return None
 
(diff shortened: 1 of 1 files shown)


ghost:
I tested this using the Wiki samples and it is Ok. 

As the PR is now (no needed changes in `fmt_plug` included), it "will miss" some cracks. So, please, let me know if you want this merged now (as soon as possible) or later (after all format changes).

From:
```
22g 0:00:00:00 DONE (2020-01-25 09:42) 70.96g/s 667.7p/s 20032c/s 20032C/s 2003-Office97-2000_1234567890123456_.doc.._myhovercraftisfullofeels
```
To:
```
14g 0:00:00:00 DONE (2020-01-25 09:55) 60.86g/s 904.3p/s 19895c/s 19895C/s 2003-Office97-2000_1234567890123456_.doc.._myhovercraftisfullofeels
```
*********
**[edited]**

In fact, I can crack all 30 hashes using the `2john` tool _as it is now_. 
I was having troubles because `--max-len` skips candidates (I was expecting it to cut at the desired size). So, it is `30` versus `14`:
```
30g 0:00:00:00 DONE (2020-01-25 10:15) 49.18g/s 344.2p/s 10327c/s 10327C/s 2003-Office97-2000_1234567890123456_.doc..myhovercraftisf
```

philsmd:
In theory, the new hash output should be rejected because it has too many fields without patching the cracking code too.
This patch shouldn't cause any other problems for now, because it doesn't change the output except adding another field...

@magnumripper said that he would look into this and see if he is able to change the cracking code to reduce false positives in jtr: https://github.com/hashcat/hashcat/issues/2067#issuecomment-577636590 . I guess he didn't have time yet to look into it... therefore I think we need to wait for magnumripper or someone else to add the code for decrypting the second block data (to filter the results even future to avoid most false positives).

Thank you anyway @claudioandre-br for testing this, much appreciated. In theory we could already try to add the additional decryption code, but I thought it makes sense to wait for @magnumripper to confirm that this strategy works and is a good idea to add also to jtr. :)
Thx

magnumripper:
I am looking into this today. Thank you so far!

magnumripper:
Replaced with #4217 which includes this commit and also fixes the formats

magnumripper:
Thanks!

PR #4204, opened 2020-02-01
update enpass to version 6 and fix #4139

### LINKED ISSUE #4139: Enpass support don't work in current implementation
As described in [old bug](https://github.com/magnumripper/JohnTheRipper/issues/2336), the current Implementation of enpass decryption don't works.
With John the Ripper 1.9.0-jumbo-1 OMP [cygwin 64-bit x86_64 AVX2 AC]
is is not possible to unlock an enpass tresor created with Enpass 6.2.0.

### Steps to reproduce ###
Tested on Environment: Windows 10, Windows 8, Ubuntu and Kali Linux

Steps:

1. Install new Enpass version (not windows store) [Windows bin](https://dl.enpass.io/stable/windows/setup/6.2.0.539/Enpass-setup.exe)  [Linux](https://www.enpass.io/support/kb/general/how-to-install-enpass-on-linux/)
2. Create new tresor with well known password (e.g. "testtest")
3. Find tresor location and copy the walletx.db file to temp directory
4. run `python .\enpass2john.py "C:\Path\to\Tresor\vault.enpassdb" > "C:\Path\to\Tresor\vault_hash_full.txt"`
5. create wordlist.txt with the password from step 2 inside, with some additional random words
6. run `.\john.exe --session=allrulesTest --wordlist="C:\Path\to\Tresor\wordlist.txt" --format=enpass "C:\Path\to\Tresor\vault_hash_full.txt"`

### System configuration ###
`
Version: 1.9.0-jumbo-1
CPU tests: AVX2
CPU fallback binary: john-xop
$JOHN is /run/
Format interface version: 14
Rec file version: REC4
Charset file version: CHR3
CHARSET_MIN: 1 (0x01)
CHARSET_MAX: 255 (0xff)
CHARSET_LENGTH: 24
SALT_HASH_SIZE: 1048576
SINGLE_IDX_MAX: 2147483648
SINGLE_BUF_MAX: 4294967295
Max. Markov mode level: 400
gcc version: 7.4.0
OpenCL headers version: 2.2
Crypto library: OpenSSL
OpenSSL 1.1.1b  26 Feb 2019
GMP library version: 6.1.2
File locking: fcntl()
fseek(): fseek
ftell(): ftell
fopen(): fopen
memmem(): System's
`

damien122:
fix issue #4139


damien122:
@magnumripper I don't know whats going on here, again the error "valid before init". 
But I could build it on a Linux system and cygwin on windows (different pcs). Do you have an idea?

ghost:
Thanks for the PR.

> I don't know whats going on here, again the error "valid before init".

You can reproduce it using your machine running a simple:
```
~/bleeding/src$../run/john --test --format=enpass
Will run 8 OpenMP threads
Benchmarking: enpass, Enpass Password Manager [PBKDF2-SHA512 256/256 AVX2 4x]... (8xOMP) FAILED (valid (before init))
```

It means `valid` is failing when it shouldn't. I will point in a review what is wrong in source code.
*****

BTW: after merging, users will no longer be able to crack a version 0 of the enpass. Is it possible (mandatory I would say) to handle both versions (version 0 and version 1)? Maybe using the same format?

src/enpass_common_plug.c
```diff
@@ -18,11 +18,11 @@ int enpass_common_valid(char *ciphertext, struct fmt_main *self)
 	ctcopy += FORMAT_TAG_LEN;
 	if ((p = strtokm(ctcopy, "$")) == NULL)	/* version */
 		goto err;
-	if (atoi(p) != 0)
+	if (atoi(p) != 1)
```
ghost: You are expecting `1`.

src/enpass_fmt_plug.c
```diff
@@ -66,8 +66,7 @@ john_register_one(&fmt_enpass);
...
...c687060b0b47bd8d004e6b8f6b5a7095dd210c108d9800be471acac750ad33d332138e0fecac173dcc6b1b1aa4fd55d374f22db4f59fde21dfc2de77a2db12a8f1681180428483b2ac134078faf056ad511a515e0824d40dfd63f602d3dabe2be33c3bc5d8408519dbba5478195eb23095b79d7bb705bd0868899e0a12d06cc2d052f5c01c71937082662f6209697a5e6908aeafba6465897fae1b9fbbe42fadc52a823ce2aa191375ad2b93462c84fb74a9eb02b9391a34a3a8ad2c83d646bffa641e78245568fca9f35a07dad31baa7814de590ac63ed4182034bf4ff68be4b428ce1ea679ad146d378cf5de1049b5abe542684cb49183281d68596059691ded3e65913c84c63d49000a927bb6af9d3e2577ee992c9c5a0f952a84e3006a283fd02907421edd90bd5da21386b533a68b933914e0a7b7fa27535008310e0d40d1d6911573cd1d1900d085c509854c415c244aa3a9a937ca29d3f809ec12fc677c1fb70762c4e0e0c463702bdad82e2a6b6bcd2d83c7710a9013497c0a639e5f379e668eea4f4222f9f0f2d00a1ce438c8305d7b04cdb2380f50ee7d774149762d8f40061b743bf9dc7f8411f766e75e9b1c6fba94a1cae6171c27821fcf9b4b9bd3278066aa900f111cdd97cbffe9fad3aa7b5096457677cc544091727d6dfd738e9e2669288182620e3e0d161a0f2f58336f14def91d826be5623970860f0e847d894701e130ccbc822c1c550a4ad6a3be48e905f2fe8d1e837d246f767b0c8454827228c82103a612f405bf7f867ac69a28f880f843e26054012f33273b36870b9b6a82353457cdce1f49301051219", "openwall"},
```
ghost: But you are sending a `0`. 

Fixing the `if` or the test string makes it work.

ghost (changes requested):

src/enpass_fmt_plug.c
damien122: after testing stuf I forgot to change it back, thx for your hint 😄

damien122:
@claudioandre-br do you have an idea what is `pbkdf2_state` in `./run/opencl/enpass_kernel.cl`? I have no idea how to replace it.

ghost:
It is declared in `opencl_pbkdf2_hmac_sha1.h`. Probably, it is not needed or used by SHA2 routines.

I'm afraid that you will need to dig into to source code to find out how to use SHA2 OpenCL routines. OpenCL stuff is more complicated to handle. You can/should focus on the CPU format right now.

Since `pbkdf2_state` is not used/needed by SHA2 API calls, you need to fix all OpenCL shared code calls inside `enpass_kernel.cl`.
*****
BTW: I would say the header you are using is wrong. `opencl_pbkdf2_hmac_sha256.h`. Isn't `512`?

magnumripper:
I am looking into this today. Thank you so far!

magnumripper:
I'm on this right now

magnumripper:
For now, since we don't know how to tell an old vs. new enpass database (I believe there actually isn't any way to tell), I'm simply differentiating `enpass5tojohn.py` (w/ hard-coded 24,000 iterations using SHA-1) and `enpasstojohn.py` (w/ hard-coded 100,000 iterations using SHA-512). The format will handle either/both. 

The changes to the OpenCL version just bogus. I'm rewriting it.

magnumripper:
I'm still on this. I'm slowed down by prototype "incompatibilities" between our PBKDF2-SHA1 and -SHA512 kernels. Luckily only the Tezos format is affected so I'll try to unify it with the SHA1 style, hopefully avoiding future problems. Very trivial but also very boring. We could spend months just cleaning up and unifying code 🙄

magnumripper:
Replaced by #4214. Thanks @damien122 !!

Issue #4171, opened 2019-12-12
Segmentation fault on zip2john


mellowcandle:
## Input File type:

$ 7z l -slt hello-world-module.tar.xz.zip

7-Zip [64] 16.02 : Copyright (c) 1999-2016 Igor Pavlov : 2016-05-21
p7zip Version 16.02 (locale=en_IL,Utf16=on,HugeFiles=on,64 bits,16 CPUs Intel(R) Core(TM) i7-5960X CPU @ 3.00GHz (306F2),ASM,AES-NI)

Scanning the drive for archives:
1 file, 749 bytes (1 KiB)

Listing archive: hello-world-module.tar.xz.zip

--
Path = hello-world-module.tar.xz.zip
Type = zip
Physical Size = 749

----------
Path = hello-world-module.tar.xz
Folder = -
Size = 530
Packed Size = 563
Modified = 2019-11-13 16:54:50
Created = 
Accessed = 
Attributes = 
Encrypted = +
Comment = 
CRC = 
Method = AES-256 Deflate
Host OS = FAT
Version = 20
Volume Index = 0

## CORE dump info:

           PID: 13452 (zip2john)
           UID: 1000 (rfried)
           GID: 1000 (rfried)
        Signal: 11 (SEGV)
     Timestamp: Thu 2019-12-12 11:57:38 IST (6s ago)
  Command Line: ./zip2john ../../hello-world-module.tar.xz.zip
    Executable: /home/rfried/1/Dnv/JohnTheRipper/run/john
 Control Group: /user.slice/user-1000.slice/session-2.scope
          Unit: session-2.scope
         Slice: user-1000.slice
       Session: 2
     Owner UID: 1000 (rfried)
       Boot ID: 0c64b511fae54df59f9f5eb6faedb480
    Machine ID: 96efa27dab3c4c99b9a0909e559dcc7e
      Hostname: rfried-mobl1
       Storage: /var/lib/systemd/coredump/core.zip2john.1000.0c64b511fae54df59f9f5eb6faedb480.13452.1576144658000000.lz4
       Message: Process 13452 (zip2john) of user 1000 dumped core.
                
                Stack trace of thread 13452:
                #0  0x00007f9715f8ebd9 __vsprintf_internal (libc.so.6)
                #1  0x00007f9715f6e3f8 _IO_sprintf (libc.so.6)
                #2  0x0000565406375e61 n/a (/home/rfried/1/Dnv/JohnTheRipper/run/john)
[...]

Issue #4054, opened 2019-07-22
zip2john segfault with AES archive that has 0 as compressed len in local header


srett:
I have an AES encrypted zip file that was created with Threema for Android in 2017. zip2john immediately crashes when processing the file. Upon inspection I found that the local header of every file has the compressed_size set to 0, which then leads to an unsigned underflow when [calculating real_cmpr_len](https://github.com/magnumripper/JohnTheRipper/blob/9b4335e7c004f3fc0986c2e06aa3aa8ba8f5ee74/src/zip2john.c#L292).  According [to wikipedia](https://en.wikipedia.org/wiki/Zip_(file_format)#Data_descriptor) this is legal under some circumstances, and the compessed_length could be extracted from the central directory entry. This would probably require two passes over the archive and building an index, or seeking around quite a bit while processing. I didn't check yet whether the archive I got here actually has that 12/16 byte data descriptor appended to the compressed BLOB at all, and whether it's supposed to come before or after the 10 byte authentication code.

I've been sitting on this for 2 months now planning to take a stab at it during the weekends but never got around to it, so I better open this issue before forgetting about it entirely, so the next person running into this doesn't have to start from scratch.

magnumripper:
Thanks for reporting. IIRC we're supposed to handle such files, but maybe we only do so for legacy pkzip archives?

srett:
Took a quick look, the old code does handle this, but not by parsing the central directory entry, but by scanning for the 16 byte data descriptor that will be appended to the compressed data. The problem here might be that the existing code does so by looking for the *optional* descriptor signature (0x08074b50), so this will not work in case that signature is omitted, or if that sequence appears in the compressed/encrypted data stream. So I think parsing the central directory would be the safest approach, and then skip over the 12/16 bytes data descriptor, or maybe use it for validation, since it should obviously match what's in the central directory.

magnumripper:
Just so you know, we'd love if you'd fix it (ideally for both cases old/new, perhaps with some shared code if applicable) and make a PR. We're seriously low on developers and you seem to know what you're doing here.

srett:
I'm still interested in doing that. I hope I can get back to this in a couple weeks. Currently very busy IRL.

magnumripper:
@srett we seem to have more issues related to (or dupe of) this issue. Do you think you'll get any time to look into it? I might get down to it this year, or not...

srett:
[...]

PR #4207, opened 2020-02-08
zip2john: Refactor and support reading central directory

srett:
By default zip2john now reads the central directory and
processes files from there. This makes handling streaming
archives (where the local header is missing length information)
more robust. The old behavior of scanning the file from the
beginning is still available by passing the -s command line
switch.

### DIFF
--- a/src/zip2john.c
+++ b/src/zip2john.c
@@ -10,6 +10,12 @@
  * pkzip will be $pkzip$ and does not look like the AES version written by Dhiru
  * Also fixed some porting issues, such as variables needing declared at top of blocks.
  *
+ * Updated in 2020 by Simon Rettberg. Handle archives by scanning their central
+ * directory first, which is more robust with archives that have been created using
+ * streams. The old behavior of scanning for local headers from the start of the
+ * file is still available through the -s option. Also fixed some minor issues and
+ * refactored the code for slightly less redundant code beween AES and legacy.
+ *
  * References:
  *
  * 1. http://www.winzip.com/aes_info.htm
@@ -83,7 +89,7 @@
  *  ZFILE This is the literal string ZFILE
  *  Fn    This is the name of the .zip file.  NOTE the user will need to keep the .zip file in proper locations (same as
  *        was seen when running zip2john. If the file is removed, this hash line will no longer be valid.
- *  Oh    Offset to the zip central header record for this blob.
+ *  Oh    Offset to the zip local header record for this blob.
  *  Ob    Offset to the start of the blob data
  *
  *
@@ -127,6 +133,58 @@
 #endif
 #include "johnswap.h"
 
+#define MAX_BLOB_INLINE_SIZE 100000
+
+#define FLAG_ENCRYPTED            1
+#define FLAG_LOCAL_SIZE_UNKNOWN   8
+#define FLAG_STRONG_ENCRYPTION   64
+
+#define AES_EXTRA_DATA_LENGTH 7  // https://www.winzip.com/aes_info.html#extra-data
+#define AES_AUTHENTICATION_CODE_LENGTH 10 // http://www.winzip.com/aes_info.htm#authentication-code
+#define AES_PASSWORD_VERIFICATION_LENGTH 2 // https://www.winzip.com/aes_info.html#pwd-verify
+
+// TODO: Add macro for fprintf(stderr, ...) and make DEBUG a command line switch
+
+// Simple wrappers
+static void xfseek(FILE *stream, long offset, int whence)
+{
+	if (fseek(stream, offset, whence) == -1) {
+		perror("fseek");
+		exit(1);
+	}
+}
+
+static size_t xfread(void *ptr, size_t size, size_t nmemb, FILE *stream)
+{
+	size_t ret = fread(ptr, size, nmemb, stream);
+	if (ret != nmemb) {
+		if (feof(stream)) {
+			fprintf(stderr, "Reached end of stream on fread\n");
+		} else if (ferror(stream)) {
+			fprintf(stderr, "ferror triggered on fread\n");
+		} else {
+			fprintf(stderr, "No reason for short read on fread (%"PRIu64"/%"PRIu64")\n", (uint64_t)ret, (uint64_t)size * nmemb);
+		}
+		exit(1);
+	}
+	return ret;
+}
+
+/**
+ * Read a char from stream, push it back if it doesn't have the expected value.
+ * This is so we can properly look for header magic (32bit).
+ */
+static int fexpect(FILE *stream, int c)
+{
+	int d = fgetc(stream);
+	if (d == EOF)
+		return 0;
+	if (d != c) {
+		ungetc(d, stream);
+	}
+	return d == c;
+}
+
 static int checksum_only = 0, use_magic = 0;
 static int force_2_byte_checksum = 0;
 static char *ascii_fname, *only_fname;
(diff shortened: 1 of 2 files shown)


srett:
As discussed in #4054 

Apart from the central directory stuff I tried to make the naming of things a little more consistent and reduce the amount of duplicated code. ZIP64 support was also improved and tested.

I was not able to test $zip3$ support ("strong encryption" only used by pkzip IIRC), but there's no particular reason to assume it's broken (if it wasn't already broken before).

ghost (approved):
:+1: 

Some notes:
* `exit()` versus `error()` so we get a proper bailout with logging.
* Source code formatting rules.

ghost:
The PR is **NOT** working with (my) AES files. 

Comparing the new hashes, the new local based hashes, and the old hashes (current implementation):
```
-rw-r--r--  1 root    root        2102 fev  8 12:45 local.txt
-rw-r--r--  1 root    root        2102 fev  8 12:37 new.txt
-rw-r--r--  1 claudio claudio  3133760 fev  8 12:42 old.txt
```
**[edited]**

As an example, new hash:
`2007.zip/1.JPG:$zip2$*0*3*0*1400000000000000d20d5064d77f0000*1e0f*0**e3304e71e82993282dbc*$/zip2$:1.JPG:2007.zip:/host/2007.zip
`

The old implementation
`2007.zip/1.JPG:$zip2$*0*3*0*1400000000000000d20d5064d77f0000*1e0f*81fbe*e3304e71e82993282dbc6a354501dd3dc218d4cd3b0572974fbd2b9c99204adaba1df ...[MUCH MORE]`

ghost:
One AES file to test: 
[2019-zipTest.zip](https://github.com/magnumripper/JohnTheRipper/files/4174639/2019-zipTest.zip)

Closes #4171 (the segfault itself is fixed)

srett:
Thanks for finding that one! It's kinda funny; the archive just has a 32bit central directory, but then the local file header contains a zip64 extended field saying the compressed and uncompressed length are both 0, which overwrites the proper 32bit values. I'll try to figure out what the right thing to to is here, otherwise I'll cobble together a simple workaround that seems to make sense. :-)

srett:
I couldn't make too much sense from the APPNOTE.TXT on issues like this. I decided to ignore ZIP64 extended fields if they are zero and we already got something from the normal 32bit fields.

On another note, after a month of not looking at the code I forgot that I actually added "ZFILE" support. The big comment at the top of zip2john.c mentions it, john supports it, but zip2john never made use of that feature, so I added that too: Now, if the compressed data's size exceeds 100kb (arbitrarily chosen by me), it won't be inlined into the hash anymore, but reference the actual .zip file. So that would apply to your test file too. It means you need the original .zip file in addition to the hash to crack the password, so it might make sense to make this a command line switch... Since the inlined data is hex encoded, it's twice as large as the data in the .zip file. If you want to crack the password of a multi-gb file, it might actually make sense to use this.

srett:
I just noticed I somehow killed a return statement after an "if" without braces some time during refactoring. It's just affecting a debug print, but still... Should I just keep committing to this branch or create a new and clean squashed PR?

I'm applying the recommended astyle settings while I'm at it. Wasn't sure about it since a lot of the existing code didn't adhere to the style guide.

ghost:
We can squash while merging. But, the style stuff should be in its own commit (to make review easier). 

In fact, probably, you should forget the astyle stuff for now (it is easier for magnum to run it himself later).

ghost:
AES is cracking the proper password now. Thanks.

I see a side effect while testing the file attached to #4171. But we can't test it (we don't know the password).

ghost:
Everything seems ok to me.

I will wait for comments. I'll merge later today (to allow comprehensive testing).

magnumripper:
Awesome! I'll have a look-see very soon unless Claudio feels confident enough to merge it before that. Thanks!

srett:
Glad I could contribute something, john has been useful in the past. :)

A couple notes, remarks or ideas for the future:

- The zip format has a lot of redundancy and extensions (zip64) that allow for creating contradicting information (e.g. central vs. local data). zip2john isn't designed to verify that a zip file is malicious or suspicious, but it doesn't try too hard to even inform the user that something might be off. I added simple checks for length information mismatch, but that's about it. Well, just in this PR claudio supplied another variant of this happening.. :) But if there's ever going to be a rewrite of this tool I suggest to track file meta data from the central directory, local header and zip64 extended fields separately and at least output some warning.
- It might make sense to have the debug output not hidden behind a compile time flag but command line argument, for easier bug reporting by users, but this might be global john policy. Also the code for legacy pkzip encryption is rather verbose by default, for AES encryption nearly everything is hidden in ifdefs.
- As mentioned above, the threshold for not inlining the compressed data is fixed at 100kb. Making this a command line switch might come in handy, as you don't need the original zip file for cracking then, trading off size for convenience.
- The old version would either process an archive as AES or legacy. The new version handles this on a per-file basis. In theory, if you had an archive that contained both types of encrypted files, it would output both types of hashes, but I doubt this thing exists until deliberately created this way.
- Error checking/handling of file reads is inconsistent. The ```fgetXXLE()``` helpers pretty much prevent checking for errors, in some parts ```fread()```'s return code isn't checked, in some it is. Again, the question is how rigorous such a tool needs to check against these kind of failures, but if it doesn't even bail out on errors, worst case it might output bogus hashes that are uncrackable. I introduced ```xfread()``` which exits on error, but a bit later I realized that this might not be the best behavior if you pass the tool multiple .zip files at once, since it would not even try to read the remaining ones. I'll leave it up to someone else to decide here.
- It'd be great to have more test files, created by different tools and tool versions.
- There's central directory encryption and local file header masking. Didn't really check what they are, so not supported at all.

Attached is a file generated by zip4j, which contains a file >4GB and thus requires valid zip64 extended fields.
[aes-64bit.zip](https://github.com/magnumripper/JohnTheRipper/files/4178862/aes-64bit.zip)

magnumripper:
> the threshold for not inlining the compressed data is fixed at 100kb

Current implementations of all `*2john` tools should *always* inline, never point to a separate file. We started doing this after implementing support for arbitrary large input files a couple years back. A problem/drawback is that our input files are often twice as big as the actual input (eg. 4 GB hex data for a 2 GB archive file) but the pros win over the cons. Ideally we should always use Base64 instead of hex but for hysterical reasons we normally don't.

srett:
I see. I did this when working on ZIP64 since I ran out of memory with my test file (laptop has only 4GB, output hash was 8GB+). Since the comment mentioned it I assumed people just didn't get around to implementing it. So should it just be removed entirely again?

ghost:
>>  all *2john tools should always inline, never point to a separate file

> So should it just be removed entirely again?

It allows you to crack a password in the cloud, for example, without leaking or sharing the zip file itself. So, IMO, please do.

magnumripper:
> > > all *2john tools should always inline, never point to a separate file
> 
> > So should it just be removed entirely again?
> 
> It allows you to crack a password in the cloud, for example, without leaking or sharing the zip file itself. So, IMO, please do.

Not necessarily without leaking: The smallest file will be picked for inlining and if that file is the only file, we're basically leaking the whole archive.

We could keep a threshold for sure, but it should be much larger than 100 KB - perhaps 4 GB or something like that. Also, it should be a run-time option as opposed to compile-time. I'll look into it myself, @srett you don't need to change anything right now.

srett:
> Not necessarily without leaking: The smallest file will be picked for inlining and if that file is the only file, we're basically leaking the whole archive.

That logic only applies to the legacy encryption, where it tries to find the three smallest files. The AES code simply dumps hashes for every file found in the archive, so if you have a 4GB and a 2kb file, it will still dump everything to stdout, unless you use the ```-o``` option or grep the output. Assuming every file is encrypted with the same password, only outputting the smallest file might be another idea for the future here to make this even more of a corner case.  
But I guess for now, moving the threshold into the GB range and making it a cmdline switch should be sufficient.

magnumripper:
> > Not necessarily without leaking: The smallest file will be picked for inlining and if that file is the only file, we're basically leaking the whole archive.
> 
> That logic only applies to the legacy encryption, where it tries to find the three smallest files. The AES code simply dumps hashes for every file found in the archive, so if you have a 4GB and a 2kb file, it will still dump everything to stdout, unless you use the `-o` option or grep the output. Assuming every file is encrypted with the same password, only outputting the smallest file might be another idea for the future here to make this even more of a corner case.

Oh, right. We should fix this. Virtually all encrypted archives use the same password for every file. We should default to pick the smallest [usable] file and perhaps add some option to output a hash for every file (or simply have that situation handled with existing `-o` and multiple runs).

We might have the same "problem" with rar5tojohn - IIRC it outputs all files in the archive. But in that case size doesn't matter and (I think) we normally have a re-used salt, so it doesn't harm a lot.

magnumripper:
Squashed & merged manually. Many thanks!

PR #4209, opened 2020-02-11
Removed some redundancy

serious-angel:


ghost:
It seems magnum did this on purpose: 10d83e6aa0c2f0510efcc3c3931a7534a441398a

Waiting to hear from him.

magnumripper:
Correct

serious-angel:
Just curious why? Doesn't deleting "dict" remove "dict/*" + "dict"? Meaning a root + internals.

magnumripper:
> Doesn't deleting "dict" remove "dict/*" + "dict"?

True, but having both in there reminds us that a **file** called "dict"  or "test" will also be ignored, among other reasons.

More importantly, this PR doesn't add any significant value, correct or not, so why bother? Had this been your 20th commit, I'd give you a way longer leash.

Issue #3686, opened 2019-03-26
John proper should allow to use '$JOHN' in JOHN_SYSTEMWIDE_EXEC and JOHN_SYSTEMWIDE_HOME


claudioandre-br:
**JtR will work as it does now if I use a** `#define JOHN_SYSTEMWIDE_EXEC "/usr/libexec/john"`
****
**[edited]**

As seen in the patch [1] below:
* I guess only path_init() have to be changed to support this. 
* It will allow packagers to install JtR wherever they want.

[1] Allow JtR system wide build to be executed from any location. Why hard code a path?
```diff
diff --git a/src/params.h b/src/params.h
index 0156044b6..a5e576fe7 100644
--- a/src/params.h
+++ b/src/params.h
@@ -100,10 +100,10 @@
 
 #if JOHN_SYSTEMWIDE
 #ifndef JOHN_SYSTEMWIDE_EXEC /* please refer to the notes above */
-#define JOHN_SYSTEMWIDE_EXEC           "/usr/libexec/john"
+#define JOHN_SYSTEMWIDE_EXEC           "$JOHN"
 #endif
 #ifndef JOHN_SYSTEMWIDE_HOME
-#define JOHN_SYSTEMWIDE_HOME           "/usr/share/john"
+#define JOHN_SYSTEMWIDE_HOME           "$JOHN"
 #endif
 #define JOHN_PRIVATE_HOME              "~/.john"
 #endif
```
This hard-coded path "mode" is NOT the best choice for system wide packages like flatpak and snap.

solardiz:
What does [1] refer to? Perhaps you forgot to include the reference?

I'm undecided on making this change, and currently don't intend to look into it before our upcoming release.

claudioandre-br:
Previous comment edited.

magnumripper:
Do you mean the install would be something like `export JOHN=/usr/local/john (...)` instead of editing params.h or build with `(...) -DJOHN_SYSTEMWIDE_HOME=/usr/local/john (...)`?

I think this is confusing.

claudioandre-br:
? Do you mean the install would be something like export JOHN=/usr/local/john (...) instead of editing params.h or build with (...) -DJOHN_SYSTEMWIDE_HOME=/usr/local/john (...)?

No. I mean a JtR system wide build can "calculate" `$JOHN` during runtime and use it (as a non system wide build does nowadays).

magnumripper:
[...]

PR #4218, opened 2020-03-02
Error out when not SYSTEMWIDE and john_home_path detected empty

ulidtko:
See discussion in #3686

### DIFF
--- a/src/path.c
+++ b/src/path.c
@@ -103,8 +103,12 @@ void path_init(char **argv)
 					*pos = '/';
 					pos = strchr(pos, '\\');
 				}
-			} else
-				fprintf(stderr, "Warning: couldn't find John home. Try using full path in argv[0]\n");
+			} else {
+				fprintf(stderr,
+				    "Error: Cannot find John home. Invoke the program via full or relative pathname.\n"
+				    "For example, /full/path/john or path/john, or set and use a shell alias.\n");
+				error();
+			}
 		}
 	}
 #endif
--- a/src/path.c
+++ b/src/path.c
@@ -103,7 +103,8 @@ void path_init(char **argv)
 					*pos = '/';
 					pos = strchr(pos, '\\');
 				}
-			}
+			} else
+				fprintf(stderr, "Warning: couldn't find John home. Try using full path in argv[0]\n");
 		}
 	}
 #endif
--- a/src/path.c
+++ b/src/path.c
@@ -104,10 +104,12 @@ void path_init(char **argv)
 					pos = strchr(pos, '\\');
 				}
 			} else {
+#ifdef __unix__
 				fprintf(stderr,
 				    "Error: Cannot find John home. Invoke the program via full or relative pathname.\n"
 				    "For example, /full/path/john or path/john, or set and use a shell alias.\n");
 				error();
+#endif
 			}
 		}
 	}


solardiz:
This is a good idea, thanks! I think we need to make these adjustments:

1. Make it an error (and call `error();` from `misc.h`), not just a warning. It's actually unsafe to have `john` read its `john.conf` and such from the current directory, which might be untrusted.

2. Avoid talking in `argv[0]` terms in the error message, which might be confusing to users who know how to use the shell, but haven't programmed in C.

3. Don't require full pathname in the message, as relative will also work.

So:

```c
			} else {
				fprintf(stderr,
				    "Error: Cannot find John home. Invoke the program via full or relative pathname.\n"
				    "For example, /full/path/john or path/john, or set and use a shell alias.\n");
				error();
			}
```

solardiz:
The failure on Travis CI is irrelevant:

```
The job exceeded the maximum time limit for jobs, and has been terminated.
```

solardiz (changes requested):
Please make and test changes as I proposed in a comment. This will also re-launch CI, hopefully letting us merge the revised PR. Thanks!

solardiz:
BTW, I think keeping source code lines within 80 characters is a lost cause for jumbo, and we can go for up to 132. However, let's try to keep messages printed by `john` within 80 characters, with exceptions for status lines and for messages that embed filenames, etc., which can sometimes grow longer.

solardiz:
Actually, I've made the edit myself. Untested. Let's see what CI says. Let's plan on squashing the two commits when merging.

ulidtko:
Yeah, sorry, you clearly beat me to it while I was unpiling the rest of the inbox :)

solardiz:
Hmm, I now realize this won't work right on Windows, where (except maybe in Cygwin shell? or in there as well?) one can simply type "john" and it will invoke from the current directory - effectively the same as having "." in PATH, but with this being standard and accepted behavior.

Unfortunately, this also means we'd have to accept the security impact of one possibly putting a `c:\full\path\john\run` into PATH on Windows (a wrong thing to do given our current approach, just like it is on Unix) and then running `john` when not being in that directory - it will try to read `john.conf`, etc. from the current directory.

So maybe we need to add e.g.:

```c
#ifdef __unix__
...
#endif 
```

around the code added here. Or there's that variable `dos` we could check at runtime, but it'd be weird to have this code compiled in at all on systems where it's supposed to be dead code.

solardiz:
BTW, it's interesting our CI runs on Wine succeeded. Perhaps we don't rely on "." in PATH there.

I'd appreciate to hear from Claudio and magnum before we possibly add this ifdef/endif thing.

ghost:
I agree: the `ifdef` is needed because "one can simply type "john" and it will invoke from the current directory". 

In fact, I do test "my" deployed package this way. E.g., inside the **phase 2 CI**:

```
    - cd JtR/run
    - powershell "cmd /c 'john --list=build-info'"
```

solardiz:
Thanks Claudio, I've made the change.

What's "phase 2 CI"?

ghost:
> What's "phase 2 CI"?

Basically, it is because I'm testing the **deployed package**, the **released package**, using the (files inside the) GitHub release. To assure everything is ok after the upload (so, I can detect a misbehavior before real users).

solardiz (approved):
I approve my changes. ;-)

solardiz:
Claudio or magnum, this failed on Travis CI with "The job exceeded the maximum time limit for jobs, and has been terminated." again. Can you override this, so that we can squash and merge? Thanks!

Edit: I guess only magnum can do this?

ghost:
Ready to merge now. Well, what I can do is:
- rerun the job (what I did now);
- make the job optional (allowed to fail)
- remove (again) the slower format(s)

PR #4222, opened 2020-03-05
4220: Fix gcc-9 -Wall warnings in ztex directory

AlbertVeli:
* **device.c**: adjust buffer sizes and use _snprintf_ instead of _sprintf_
* **ztex.c**: work around _strncpy_ not adding _terminating zero_ when truncating
* **ztex_sn.c**: use _snprintf_ instead of _sprintf_ and unsigned int (%u) format string.

### DIFF
--- a/src/ztex/device.c
+++ b/src/ztex/device.c
@@ -92,7 +92,9 @@ static int device_init_fpgas(struct device *device,
 {
 	// Read config for given bitstream, device
 	char *CFG_SECTION = "ZTEX:";
-	char conf_name_board_freq[256], conf_name_freq[256];
+	// conf_name_board_freq must be shorter than conf_name_freq to avoid a
+	// gcc-9 compiler warning. But at least 21 bytes to make the string fit.
+	char conf_name_board_freq[128], conf_name_freq[256];
 	int default_freq[NUM_PROGCLK_MAX], board_freq[NUM_PROGCLK_MAX],
 		fpga_freq[NUM_PROGCLK_MAX];
 
@@ -103,8 +105,8 @@ static int device_init_fpgas(struct device *device,
 				default_freq, NUM_PROGCLK_MAX);
 
 		// Frequency specific to given board
-		sprintf(conf_name_board_freq, "Frequency_%s",
-				device->ztex_device->snString);
+		snprintf(conf_name_board_freq, sizeof(conf_name_board_freq),
+				"Frequency_%s", device->ztex_device->snString);
 		cfg_get_int_array(CFG_SECTION, bitstream->label,
 				conf_name_board_freq, board_freq, NUM_PROGCLK_MAX);
 
@@ -120,7 +122,7 @@ static int device_init_fpgas(struct device *device,
 	// Length is specific to bitstream.
 	// If configuration packet is of incorrect length then
 	// the FPGA would raise an error (app_status=0x08)
-	char conf_name_board_config1[256], conf_name_config1[256];
+	char conf_name_board_config1[128], conf_name_config1[256];
 	char default_config1[CONFIG_MAX_LEN], board_config1[CONFIG_MAX_LEN];
 	int len_default, len_board, len;
 
@@ -132,8 +134,8 @@ static int device_init_fpgas(struct device *device,
 	} else
 		len_default = 0;
 
-	sprintf(conf_name_board_config1, "Config1_%s",
-			device->ztex_device->snString);
+	snprintf(conf_name_board_config1, sizeof(conf_name_board_config1),
+			"Config1_%s", device->ztex_device->snString);
 	ptr = cfg_get_param(CFG_SECTION, bitstream->label,
 			conf_name_board_config1);
 	if (ptr) {
@@ -159,7 +161,7 @@ static int device_init_fpgas(struct device *device,
 		if (bitstream->num_progclk) {
 
 			// Check for frequency for given fpga in the config
-			sprintf(conf_name_freq, "%s_%d", conf_name_board_freq, fpga_num + 1);
+			snprintf(conf_name_freq, sizeof(conf_name_freq), "%s_%d", conf_name_board_freq, fpga_num + 1);
 			cfg_get_int_array(CFG_SECTION, bitstream->label, conf_name_freq,
 					fpga_freq, NUM_PROGCLK_MAX);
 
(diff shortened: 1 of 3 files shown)


src/ztex/device.c
```diff
@@ -92,7 +92,7 @@ static int device_init_fpgas(struct device *device,
 {
 	// Read config for given bitstream, device
 	char *CFG_SECTION = "ZTEX:";
-	char conf_name_board_freq[256], conf_name_freq[256];
+	char conf_name_board_freq[11 + ZTEX_SNSTRING_LEN], conf_name_freq[256];
```
solardiz: Just make it `256 + ZTEX_SNSTRING_LEN`, so that we don't need to (and might forget to) edit this line each time we change the corresponding fixed portion of the string in the `snprintf` call.

src/ztex/device.c
```diff
@@ -120,7 +120,7 @@ static int device_init_fpgas(struct device *device,
 	// Length is specific to bitstream.
 	// If configuration packet is of incorrect length then
 	// the FPGA would raise an error (app_status=0x08)
-	char conf_name_board_config1[256], conf_name_config1[256];
+	char conf_name_board_config1[9 + ZTEX_SNSTRING_LEN], conf_name_config1[256];
```
solardiz: Just make it `256 + ZTEX_SNSTRING_LEN`, so that we don't need to (and might forget to) edit this line each time we change the corresponding fixed portion of the string in the `snprintf` call.

src/ztex/device.c
```diff
@@ -159,7 +159,7 @@ static int device_init_fpgas(struct device *device,
 		if (bitstream->num_progclk) {
 
 			// Check for frequency for given fpga in the config
-			sprintf(conf_name_freq, "%s_%d", conf_name_board_freq, fpga_num + 1);
+			snprintf(conf_name_freq, 256, "%s_%d", conf_name_board_freq, fpga_num + 1);
```
solardiz: Use `sizeof`.

src/ztex/ztex.c
```diff
@@ -135,7 +135,8 @@ int ztex_device_new(libusb_device *usb_dev, struct ztex_device **ztex_dev)
 
 	// Original SN remains in dev->snString_orig.
 	char *sn = ztex_sn_get_by_sn_orig(dev->snString_orig);
-	strncpy(dev->snString, sn, ZTEX_SNSTRING_LEN);
+	strncpy(dev->snString, sn, ZTEX_SNSTRING_LEN - 1);
+	dev->snString[ZTEX_SNSTRING_LEN - 1] = 0;
```
solardiz, line 139: I've confirmed that this appear to be OK: we still have NUL-padding of shorter SN strings, which might or might not matter, and we have NUL-termination of strings of length 10 like we did before. We'd now truncate strings of length 11 to 10, but as I understand SNs are supposed to be length 10. A comment in `ztex_sn.h` says that 11 includes the NUL terminator.

Other than that, the code appears not to rely on this string being NUL-terminated (uses `strncmp`), so there was probably no issue here.

src/ztex/ztex_sn.c
```diff
@@ -134,7 +134,7 @@ char *ztex_sn_get_by_sn_orig(char *sn_orig)
 		return sn_orig;
 
 	static char result[ZTEX_SNSTRING_LEN];
-	sprintf(result, "%d", line->id + 1);
+	snprintf(result, ZTEX_SNSTRING_LEN, "%u", line->id + 1);
```
solardiz: Use `sizeof`.

solardiz (changes requested):
Overall, your changes look good to me, but please make some minor adjustments as I indicated. Thanks!

solardiz:
Since the `strncpy` "fix" here breaks the hopefully-unneeded and unintentional support for SN length 11, we could want to also drop this support from `ztex_sn_is_valid` for consistency. Right now, it is:

```c
int ztex_sn_is_valid(char *sn)
{
        int i;
        for (i = 0; i < ZTEX_SNSTRING_LEN; i++) {
                if (!sn[i])
                        return i < ZTEX_SNSTRING_MIN_LEN ? 0 : 1;
                if ( !( (sn[i] >= '0' && sn[i] <= '9') || (sn[i] >= 'A' && sn[i] <= 'F')
                                || (sn[i] >= 'a' && sn[i] <= 'f')) )
                        return 0;
        }
        if (sn[i])
                return 0;

        return 1;
}
```

Proposed change:

```diff
index 10c6b0f..f04ce62 100644
--- a/src/ztex/ztex_sn.c
+++ b/src/ztex/ztex_sn.c
@@ -27,10 +27,8 @@ int ztex_sn_is_valid(char *sn)
                                || (sn[i] >= 'a' && sn[i] <= 'f')) )
                        return 0;
        }
-       if (sn[i])
-               return 0;
 
-       return 1;
+       return 0;
 }
 
 
```

This also simplifies it. If we later determine we need support for length 11 (highly unlikely), we can instead increase `ZTEX_SNSTRING_LEN` to 12.

solardiz:
Albert, please try making all of the changes I proposed above, including the one to `ztex_sn_is_valid`, and test this with your boards - then revise this PR (you can force-push a revised commit or add another commit, which we'll plan to squash with your first commit on merging). Thanks!

src/ztex/device.c
AlbertVeli: I can make it 128 + ZTEX_SNSTRING_LEN. That leaves enough bytes to change the string later. To avoid the truncation warning conf_name_boad_freq must be a few bytes shorter than conf_name_freq. Or it might be even better to just set it to 128, if alignment is an issue.

AlbertVeli:
I force pushed the change so it is only one commit. And I just used 128 as buffer length and added a comment that it must be shorter than the 256 byte buffer. That is more than enough for the string to fit and there are no warnings now for me. And it works on the ZTEX board.

There is one strange message, but not related to this issue. It was there before too:
```
SN 04A36E226F: uploading bitstreams.. ok
SN 04A36E226F: Timeout.
```
After that john works anyway with the ZTEX board and cracks the password hash I gave it.

PS Since I  force pushed some diffs above are outdated. Not sure how to handle this but I guess it's ok anyway.

src/ztex/ztex.c
AlbertVeli: I left this one as is with strncpy, but it could also have been done with a memcpy of ZTEX_SNSTRING_LEN to avoid the warning.

solardiz (approved):
Thanks. I'll merge this now.

For future occasions, let's avoid unnecessarily saying things like "at least 21 bytes to make the string fit" in comments, because those might get out of date and inconsistent with code with future changes, and because they're either not double-checked in reviews (so might be wrong right away) or waste reviewers' time. Instead, say things like "large enough to make the string fit". Exception: it's appropriate to mention specific numbers if they're part of a file format specification or such.

PR #4228, opened 2020-03-10
Stop complaining about empy home_path on Windows

### LINKED ISSUE #4227: Windows issue after 0674b69
The `unix` is defined in Windows CygWin. Probably, we should change the `if` to something like this:

* `if unix && ! __CYGWIN__ // MingW, etc.
`
```
#define __CYGWIN__ 1
[...]
#define unix 1
#define __INT16_MAX__ 0x7fff
```

```
john --list=build-info
Error: Cannot find John home. Invoke the program via full or relative pathname.
For example, /full/path/john or path/john, or set and use a shell alias.
Command exited with code 1
```

ghost:
On Windows CMD, the user can run "john" and it will invoke it from the current directory.

Closes #4227

### DIFF
--- a/src/path.c
+++ b/src/path.c
@@ -104,7 +104,7 @@ void path_init(char **argv)
 					pos = strchr(pos, '\\');
 				}
 			} else {
-#ifdef __unix__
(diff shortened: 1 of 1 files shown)


solardiz (approved):
As discussed, and looks good to me.

PR #4229, opened 2020-03-11
Don't refuse to run eg. "unique" from path.

magnumripper:
See https://github.com/magnumripper/JohnTheRipper/commit/0674b6994e2fc9eb251643f03402534629024748#commitcomment-37756988

### DIFF
--- a/src/path.c
+++ b/src/path.c
@@ -103,11 +103,11 @@ void path_init(char **argv)
 					*pos = '/';
 					pos = strchr(pos, '\\');
 				}
-			} else {
 #if !(defined(__DJGPP__) || defined(__CYGWIN__) || defined(_MSC_VER) || defined(__MINGW32__))
+			} else if (!strncmp(argv[0], "john", 4)) {
 				fprintf(stderr,
 				    "Error: Cannot find John home. Invoke the program via full or relative pathname.\n"
-				    "For example, /full/path/john or path/john, or set and use a shell alias.\n");
+				    "For example, /full/path/%s or path/%s, or set and use a shell alias.\n", argv[0], argv[0]);
 				error();
 #endif
 			}


solardiz (approved):
On one hand, I don't mind these changes at least as a temporary workaround to this recently introduced code, effectively (maybe temporarily) limiting its scope.

On the other hand, why is it that `unique` failed? Does it call `path_init`? Maybe it shouldn't? And if it does need `path_init` (perhaps for some `john.conf` settings?), then we should probably "protect" it with this check just like we do for `john`, as otherwise it'd be unexpectedly reading something (what?) from the current directory (which is unexpected and unsafe). Ditto for other programs/symlinks that we have.

ghost:
As a side note, the "execute from the current directory" only happens in CMD. Powershell requires the `.\`. 
* So, JtR can require a path in the command line after the next release (saying CMD behavior is bad (TM)).

magnumripper:
In Jumbo, `CPU_detect_or_fallback` needs `path_init` (and all symlinked tools need the former). I caused that when I made Jumbo capable of OMP-fallback for non-systemwide builds. Not sure how to remedy that.

solardiz:
> the "execute from the current directory" only happens in CMD. Powershell requires the `.\`.

That's good to know, and good news.

> JtR can require a path in the command line after the next release (saying CMD behavior is bad (TM)).

I think we shouldn't. Whatever we say, CMD will be relevant for way longer. While this wouldn't break use of CMD completely, it would add to user confusion.

> In Jumbo, `CPU_detect_or_fallback` needs `path_init` (and all symlinked tools need the former).

Maybe we could avoid having the symlinked tools need `CPU_detect_or_fallback` as they don't significantly benefit from SIMD being allowed in the compiler, but it'd complicate our build process. So yes, I also don't know how to remedy that best. The issue is very real, though: when even the symlinked tools may exec a fallback binary, it's especially nasty if they attempt to do it from an untrusted current directory.

For OMP-fallback specifically, maybe we can exclude it from the symlinked tools because they don't currently use OpenMP anyway (although I imagine `unique` could reasonably start to use OpenMP eventually), but there's also that issue with compiler flags allowing use of SIMD beyond what's implied in the base architecture.

magnum, it looks like you were merely lucky that in your case with `unique` in PATH you either had no fallbacks enabled or they were not invoked on your system. Do we really need to support use cases that get away by mere luck? Even at the risk of having a security issue?

magnumripper:
I'm pretty sure I opted to copy `john-sse2-no-omp` by pure instinct without giving it much thought (and without recognizing the threat since they aren't shared computers).

Some alternatives:
- Obviously we could just revert my commit. I should probably start running system-wide anyway.
- We could `#ifdef` it instead, depending on sth like `--with-unsecure-execv` option to configure or even an undocumented `-DSECURITY_PROBLEM.
- We could drop the symlink stuff and make the tools stand-alone. That could end up messier than we expect though, if some of them use functions from eg. misc.o or memory.o
- There is [this library](https://github.com/gpakosz/whereami) but it might be overkill. It is announced as "a drop-in two files library" though. I haven't tried it.

solardiz:
> Obviously we could just revert my commit. I should probably start running system-wide anyway.

I think we should keep your commit, but drop the ` if (!strncmp(argv[0], "john", 4))` check. So that the proper program name would be displayed in the examples, but the requirement would apply to all programs.

>  We could drop the symlink stuff and make the tools stand-alone.

This won't help. The issue isn't symlinks, but builds potentially requiring instruction set extensions, which may include files like:

> That could end up messier than we expect though, if some of them use functions from eg. misc.o or memory.o

solardiz:
How about this? -

```diff
diff --git a/src/john.c b/src/john.c
index dfad2dc..33c87c7 100644
--- a/src/john.c
+++ b/src/john.c
@@ -2160,8 +2160,10 @@ int main(int argc, char **argv)
        }
 #endif
 
-        /* Needed before CPU fallback */
+#if CPU_FALLBACK || OMP_FALLBACK
+       /* Needed before CPU fallback */
        path_init(argv);
+#endif
 
        if (!strcmp(name, "unshadow")) {
                CPU_detect_or_fallback(argv, 0);
@@ -2197,10 +2199,16 @@ int main(int argc, char **argv)
                CPU_detect_or_fallback(argv, 0);
                return zip2john(argc, argv);
        }
+
        if (!strcmp(name, "base64conv")) {
                CPU_detect_or_fallback(argv, 0);
                return base64conv(argc, argv);
        }
+
+#if !(CPU_FALLBACK || OMP_FALLBACK)
+       path_init(argv);
+#endif
+
        john_init(name, argc, argv);
 
        if (options.max_cands) {
diff --git a/src/path.c b/src/path.c
index 4affedf..d1a2533 100644
--- a/src/path.c
+++ b/src/path.c
@@ -104,7 +104,7 @@ void path_init(char **argv)
                                        pos = strchr(pos, '\\');
                                }
 #if !(defined(__DJGPP__) || defined(__CYGWIN__) || defined(_MSC_VER) || defined(__MINGW32__))
-                       } else if (!strncmp(argv[0], "john", 4)) {
+                       } else {
                                fprintf(stderr,
                                    "Error: Cannot find John home. Invoke the program via full or relative pathname.\n"
                                    "For example, /full/path/%s or path/%s, or set and use a shell alias.\n", argv[0], argv[0]);
```

This should allow for copying of tools/symlinks other than `john*` e.g. to `~/bin` or `/usr/local/bin` in without-fallbacks builds.

PR #4235, opened 2020-03-25
support keccak_224 and keccak_384 in dynamic

kazkansouh:
Support within `john` for `keccak` is limited to `256` and `512` digest lengths. I recently needed to use a non-supported digest size so have updated the code to support them.

This PR extends the existing support for `keccak` for digest lengths `224` and `384` within dynamic mode. I did consider providing main formats for `keccak` for these sizes, but I think given their low use, it is acceptable to only provide support within dynamic mode. 

The code is essentially copy, paste and update of existing definitions so should align with existing coding. There is one point that worth mentioning:

* I chose to extend at the end of the existing range of dynamic formats to preserve existing numbering of the formats. That is, the new formats use 43x and 44x ranges and the existing `keccak` are located at 41x and 42x ranges. Thus they are no longer sorted by digest size. If desired, I can update the PR to sort them by size?


jfoug:
We will want to make sure pass_gen.pl is updated.  Also make sure these are in the jtrts.  Those changes should likely be stand alone and not part of this PR, but will help validate this.   I have static looked at code (within github across the web), and things look like they are done properly. I see we have a TravisCI error, but I am not sure it is caused by this code, or if it is some false positive.

I have put these changes on my local PC, and will validate that all changes appear correct and fully done.

ghost (approved):
We have the Ok from Jim. Merging after CI.

jfoug:
I have looked at all LARGE_HASH_EDIT_POINTS  All code is there. I also have a pass_gen.pl (adding raw-keccak224 and 384) and adding the dynamic compiler stuff (and the 430-449 hash info), and things seem 'happy'.  I have not done the test suite yet. But this is the type of formats which 'can' cause problems if there is ANYTHING wrong.  64 bit, full width hashes, push the internal dynamic buffers to the limits.  And especially since these are 'truncated' after the fact, we really want to know things are really happy

magnumripper (commented):
Looks jolly good to me too

PR #4246, opened 2020-04-06
Add --single-user-seed=FILE option.

magnumripper:
Add --single-user-seed=FILE option, where FILE is a file in user:password format that can seed Single mode. Incidentally, it's ESSID:password for WPAPSK.

### DIFF
--- a/doc/NEWS
+++ b/doc/NEWS
@@ -101,6 +101,9 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
 - Keys [<] and [>] can now be used to decrease/increase verbosity level.
   [magnum; 2020]
 
+- Added --single-user-seed=FILE option for single mode. FILE is a wordlist
+  with seeds per username (user:password[s] format)  [magnum; 2020]
+
 
 Major changes from 1.8.0-jumbo-1 (December 2014) to 1.9.0-jumbo-1 (May 2019):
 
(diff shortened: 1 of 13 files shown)


solardiz:
Can we have a copyright+license statement for `mgetl.h`, please?

Also, maybe have a comment in there on the function violating strict aliasing rules - or better yet we'd have rewritten it in a compliant way, but that's tricky.

magnumripper:
Wow, I can't believe I missed the copyright statement. I thought I was adamant.

magnumripper:
> Also, maybe have a comment in there on the function violating strict aliasing rules - or better yet we'd have rewritten it in a compliant way, but that's tricky.

Which one is that now?

magnumripper:
You must mean `mgetl()`. Yes it's a nasty function but it does check for `ARCH_ALLOWS_UNALIGNED` and otherwise will fall down to char. Besides, we've had it for many years I think - I just moved it.

solardiz:
Yes, the problem with `mgetl()` violating strict aliasing rules isn't new, and yes our check for `ARCH_ALLOWS_UNALIGNED` mitigates the problem in practice so far - but technically it's still UB, and a future compiler/version might exploit it. I think a proper fix would be use of `union` types (or pointers to them) for the global variables and for the `mgetl()` argument. The `union` types need to include fields of _all_ types that we access the data through, eliminating the need for typecasts.

BTW, we might want to increase the alignment of `line` in `wordlist.c` from `ARCH_WORD` to `vtype`, which should allow us not to use `vstoreu`. This might provide some speedup in some scenarios on some CPUs.

magnumripper:
That speedup should be long gone (I think) but I get what you mean, I will look into it!

PR #4263, opened 2020-05-19
Fixed typo in axcrypt2john, missing b before byte literal.

alexanderlunden:


solardiz:
Edit: I wrongly thought at first this was a Python version thing, and not a typo. I stand corrected: looking at the changes in context, I see the same variable `axxdata` is treated as a byte string. OK then.

solardiz:
Hmm, I was too quick to say the above. Looks like this same script does use byte strings elsewhere. So perhaps a typo, after all? Anyway, would like someone more knowledgeable in Python to confirm this.

magnumripper (approved):

magnumripper:
I think it's correct, and it seems to work with python2 as well as python3

PR #4290, opened 2020-08-09
OpenCL: (pre)tune future NVIDIA GPUs

ghost:
Treat all future NVIDIA GPUs the same as the latest revision.

https://github.com/magnumripper/JohnTheRipper/pull/4288#issuecomment-670929077


solardiz (approved):
This looks good to me. Of course, we'll need to test and benchmark on a few systems, but OK to do that after merging.

solardiz:
Just tested that (as expected) nothing visibly changed with this PR's changes on a V100.

PR #4302, opened 2020-08-19
zip2john: workaround false negatives with pkzip archive made by macOS eZip app

magnumripper:
A pkzip archive made by macOS eZip app (closed source but seem to use libarchive) only have a single byte valid checksum. We assume here that EFH 0x6c78 means we can't trust the second byte.  This workaround can't break anything: At worst (if it's too promiscuous) it will lead to slightly worse performance from some (very rare) archives.

### DIFF
--- a/src/zip2john.c
+++ b/src/zip2john.c
@@ -184,8 +184,8 @@ static int fexpect(FILE *stream, int c)
 	return d == c;
 }
 
-static int checksum_only = 0, use_magic = 0;
-static int force_2_byte_checksum = 0;
+static int checksum_only, use_magic;
+static int force_1_byte_checksum, force_2_byte_checksum;
 static char *ascii_fname, *only_fname;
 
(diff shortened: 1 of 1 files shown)


ghost (approved):
Just in case you need approval

magnumripper:
I don't seem to be able to even look at the bot checks. Are they running at all?

ghost:
> Are they running at all?

No.

jfoug (approved):
The code appears ok.  with the amount of RW information we have about how the programs build these checksums, I think this is the best way to proceed at this time.

magnumripper:
@solardiz I'm neither powerful enough now to merge this with uncompleted bot checks nor investigate why those checks aren't working. Code is tested and fine though, so please merge at will.

solardiz:
@magnumripper Thanks. I'll see if we can repair the bots before merging this, or will just merge if that would be taking too long.

solardiz:
I've merged this with my admin privs for now. Let's see what happens with the bots next time, now that Travis does see this repo.

PR #4304, opened 2020-08-22
README-OPENCL: give a more relevant build flags override example

solardiz:
In the example, use CUDA rather than AMDAPP, and override CPPFLAGS rather than
CFLAGS.  Overriding CFLAGS resulted in our -O flags getting overridden as well.


solardiz:
Added a commit making a similar change for `INSTALL`:

```
    INSTALL: for providing an include path, override CPPFLAGS (not CFLAGS)
    
    Overriding CFLAGS resulted in our -O flags getting overridden as well,
    which we had to provide a workaround for and explain it.
```

We already use `CPPFLAGS` in many other examples, including in another one in `INSTALL`.

Issue #2812, opened 2017-10-11
JtR too restrictive when verifying encoded UTF-16 files


claudioandre:
Since 0ce265dcffff41864143e3bc48de539b4c62649f, JtR is refusing to load a 8bit wordlist. Well, at least all tools I used detected the wordlist as: ascii/unknown, unknown-8bit or iso8859-1.

To see the issue in action (tested on *well*), just
```
./jtrts.pl -noprelims -type MSSql
```

magnumripper:
Not sure how @jfoug ended up with the crazy idea to hash a UTF-16 BOM and even depend on it, but this is a TS issue and not a JtR one (unless you show me a real-world case where this is a problem *).

I dropped the stupid lines from the dictionaries but now we end up with some formats "failing" due to only cracking 1499 out of 1500. This is just stupid, I don't have time to fix it right now. But again, it's a TS issue.

*) If you do find such case, we could drop the check when using `--encoding=raw` but that's about as far as I'm willing to go.

magnumripper:
OK, after some research I changed my mind. I will probably need to revert to only do this check when input encoding is UTF-8. Perhaps still checking the very first line for other codepages though.

BTW the rationale for checking for BOM's on *all* lines is that user may have concatenated several files, where one or some of them were UTF-16 by mistake. I've seen it happen, I've heard it happen.

bill-e-ghote:
Thanks for reverting this. 'twas puking on crackstation.txt, even.

magnumripper:
@bill-e-ghote well, it should... I had a look now - crackstation.txt contains UTF-16 garbage which is precisely what we wanted to detect because all such data is 100% unusable for any format and any crack mode.

That wordlist is by far the very lowest quality I've ever seen. The size of it is a good hint.

Nonetheless, I now relaxed things even further - we now just warn instead of bailing out. That's only for UTF-16 in wordlists though - we still refuse to read any input file containing a UTF-16 BOM.

androidacy-user:
Why the decision to completely not support utf-16? I have a zip file in utf-16 format that's encrypted, on Linux

Sorry to bump a super old issue but googling took me here

solardiz:
On Sat, May 30, 2020 at 06:36:28PM -0700, Alexia wrote:
> I have a zip file I'm utf-16 format that's encrypted

What do you mean by this, and what makes you think that's what you have?

androidacy-user:
[...]

PR #4306, opened 2020-08-24
Do not abort on seeing a UTF-16 BOM in input file or wordlist, just warn

### LINKED ISSUE #4238: Review UTF-16 detection and other encoding detections (fatal error vs. warning)
https://www.openwall.com/lists/john-users/2020/03/27/4

I can't remember: Maybe the fatal error only happens when detecting UTF-16 in *input files* (as in *hash files*), I'm not quite sure we check *wordlists* for just that at all. Anyway we should probably almost never die from problems seen in a wordlist (potentially a long way into a session), only warn. While we're at it, I believe we currently die on seeing a NULL in a wordlist (probably only when buffering/indexing to RAM) and that too might be better as just a warning.

On a side note, I've been meaning to add (optional) color to warnings, eg. yellow for important notes/reminders and red for worse problems. Sometimes it's easy to miss a crucial note and even more so if we revert to chug-along behavior per above.

magnumripper:
Closes #4238, see also #2812

### DIFF
--- a/src/loader.c
+++ b/src/loader.c
@@ -103,25 +103,24 @@ static int jumbo_split_string;
  */
 static MAYBE_INLINE char *check_bom(char *string)
 {
-	static int warned;
-
 	if (((unsigned char*)string)[0] < 0xef)
 		return string;
 
 	if (!memcmp(string, "\xEF\xBB\xBF", 3)) {
+		static int warned;
+
 		if (options.input_enc == UTF_8)
 			string += 3;
-		else if (john_main_process && !warned++) {
+		else if (john_main_process && !warned++)
 			fprintf(stderr,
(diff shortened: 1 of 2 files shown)


solardiz (approved):
Thank you for fixing that issue! This mostly looks good to me, but I suggest edits to the messages: 
"password hash file" instead of "input file" and "NUL" instead of "NULL" when speaking about characters.

magnumripper:
Done, although I consistently call it the "input file" in discussions and docs (I don't use that term for eg. wordlists).

solardiz:
OK. Looks good to me now.

solardiz:
All tests except for the pending Circle more-tests have passed, and I am able to merge without needing to override anything. So our CI is working well again, to the same extent it worked before the repo move.

I'll merge this now.

PR #4315, opened 2020-08-25
In ASan and -DDEBUG builds, fill allocated memory with non-zeroes

### LINKED ISSUE #4310: In slow debugging builds, have mem_alloc() fill memory with non-zeroes
For example, I am now testing an ASan build with this added:

```diff
@@ -78,6 +78,8 @@ void *mem_alloc(size_t size)
                error();
        }
 
+       memset(res, 0x55, size);
+
        return res;
```

but we'll need to make it conditional on some build-time or run-time configuration option.

solardiz:
Closes #4310

### DIFF
--- a/src/memory.c
+++ b/src/memory.c
@@ -66,6 +66,16 @@ void cleanup_tiny_memory()
 	}
 }
 
+#if defined(WITH_ASAN) || defined(DEBUG)
+static void mem_debug_fill(void *p, size_t size)
+{
+	if (p)
+		memset(p, 0xde, size);
+}
+#else
+#define mem_debug_fill(p, size) /* nothing */
+#endif
+
 void *mem_alloc(size_t size)
 {
 	void *res;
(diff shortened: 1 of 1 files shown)


solardiz:
Curiously, for me all CPU formats (410 of them) passed self-test and benchmark in an ASan build even when I was also filling memory in `mem_realloc`, even though `realloc` is normally supposed to preserve old memory contents (up to the old size). If we never rely on that in JtR, we could call our function differently (e.g. `mem_realloc_nonpreserving`) and exclude that property from it, and then we'd be able to fill memory in there as well. I didn't do that for this PR, keeping the content-preserving property in case we do have some current or future uses where that matters.

magnumripper:
> If we never rely on that in JtR, we could call our function differently (e.g. `mem_realloc_nonpreserving`)

I'm pretty sure we do use `realloc()` depending on it preserving old content (although perhaps not in any format), but I'm not certain we use the `mem_realloc` macro in those places. Reviewing all use of `realloc` should be a quick job though.

PR #4319, opened 2020-08-25
Fix for #3796 (core in dynamic ad hoc compiler)

jfoug:
Updated pass_gen so I could force set a user name within the dynamic formats.  Needed for work fixing dynamic compiler crash #3796

2nd commit is the fix for #3796

PR #4322, opened 2020-08-26
AdHoc dynamic: md5($s.md5($p).$s) broken on non SIMD builds

jfoug:
This was broken from the start. The formatting changes done #4319 simply carried over the original problem.

I first made proper changes using the ```Flag=MGF_KEYS_BASE16_IN1_MD5``` which was the original intention.  But that script was about 50% slower than using the script built for the SIMD.  So in the end, I simply use a single unified script for both SIMD and non-SIMD.

### DIFF
--- a/src/dynamic_compiler_lib.c
+++ b/src/dynamic_compiler_lib.c
@@ -77,7 +77,7 @@ static LIB_struct lib[] = {
 #endif
 	"@dynamic=md5($s.md5($p))@9fd5767e78e00d1c947b4e055831ef9c$df694488"} } },
 
-	// Dyna-lib for md5($s.md5($p))
+	// Dyna-lib for md5(md5($p))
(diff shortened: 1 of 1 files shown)


magnumripper (approved):

magnumripper (commented):
Just noticed that while you mention `md5($s.md5($p))` in the description, the change was to `md5($s.md5($p).$s)`. I guess the commit is correct and the description isn't?

jfoug:
Fixed title. Yes, the fix was to this subformat:   ```md5($s.md5($p).$s)```

solardiz:
Unfortunately the commit message remained wrong, and it's too late to correct now without causing other difficulties.

ghost:
You should add a PR template with some notes about JtR policies.
* https://docs.github.com/en/github/building-a-strong-community/creating-a-pull-request-template-for-your-repository

Issue #4002, opened 2019-05-26
Autodetection of properly tagged dynamic formats is non-working


solardiz:
As mentioned in:

https://github.com/magnumripper/JohnTheRipper/issues/3970#issuecomment-493723879

autodetection of properly tagged dynamic formats is currently non-working. I think it used to work before. We need to repair it.

The above example is for dynamic_82, but I just ran into the same issue with dynamic_62.

jfoug:
This is most certainly by design (not mine).   The file ./run/dynamic_disabled.conf lists a set of dynamic hashes which have had their auto-* logic removed. The will not test, run or do anything, UNLESS you add the -format= tag.

Both dynamic_82 and dynamic_62 are in that file.

In john.conf : 

```
# These formats are disabled from all-formats --test runs, or auto-selection
# of format from an input file. Even when disabled, you can use them as long
# as you spell them out with the --format option. Or you can delete a line,
# comment it out, or change to 'N'
[Disabled:Formats]
#formatname = Y
.include '$JOHN/dynamic_disabled.conf'
```

then dynamic_disabled.conf is simply a long list of dynamic hash types.  All of them are disabled.

magnumripper:
You added it to `john.conf` in 3aa49af9, Jim - and then moved them to its own file in facddf0b 😉 

I was probably involved in some discussion leading to that though. I think we meant to only put really academic formats in there so perhaps dynamic_82 (and dynamic_62?) should be dropped from that list.

jfoug:
are 82 and 62 the only ones?

magnumripper:
I have no idea. Ideal would be a (very) carefully thought out reduced self-test that only tries each "code path" of dynamic. For example, `md5($p)`, `sha1($s.$p)`, `keccak($p.$s)` and so on. But IIRC what we wanted with this was to not (by default) test things like `ripemd160(ripemd160($p).ripemd160($p))` which was likely never used anywhere IRL and the reason was the sheer number of dynamics.

jfoug:
So in ideal, we would have all dynamic formats work 100% of the time EXCEPT for -test* switches (unless -format was added).  Is that the idea?

I think the way it was handled was that the disabled simply did not load the format, unless it was specified.   This certainly keeps -test* from blowing through all the dynamic academic types.  But we DO want john to run properly on input files for any dynamic type (academic or not).

If that is the idea, then I can start looking into it.

jfoug:
[...]

PR #4324, opened 2020-08-28
dynamic: disable ONLY stops -test from auto running

jfoug:
Fix for #3970
Fix for #4002


jfoug:
Changed the 'logic' for the disabled formats.  Now the format gets loaded.  However, we simply do not auto test.  BUT if the hash is in an input file, and we are running, then the hash will auto detect.

As always, ```--format=dynamic-all``` will test all dynamic hashes (even the disabled ones).   Also in debugging builds, all dynamic is tested (as it was before).

jfoug:
This is actually better logic.  We simply added this, to cut down on time (especially in the CI's) which the --test was taking.  Repetitively going over these junk formats was not doing anything about finding any bugs.   BUT totally disabling the formats, i.e. where they were not working without using the --format switch really was wrong run time behavior.

run/dynamic_disabled.conf
```diff
@@ -135,10 +137,10 @@ dynamic_213 = Y
 dynamic_214 = Y
 dynamic_215 = Y
 dynamic_216 = Y
-dynamic_217 = Y
 dynamic_218 = Y
 dynamic_221 = Y
 dynamic_222 = Y
+dynamic_217 = Y
 dynamic_223 = Y
 dynamic_224 = Y
```
magnumripper: Why the reordering here?

magnumripper (approved):
You should revert the inadvertent move of dynamic_217 line. Other than that, it looks reasonable to me.

run/dynamic_disabled.conf
jfoug: Not sure how the swap originated, but this will get it back to original form.

solardiz (approved):
Looks good enough to me.

jfoug:
I see this is committed.  But at work today, I was wondering if we should have renamed the section.  Currently it is called "[Disabled:Formats]"   I was thinking along the lines of "[Muted:Format-tests]"  

But also does this cause problems for people who have a customized john.conf file?  

It is sort of a moot point for this commit, but since we have enabled the formats (but kept the testing muted), where before the entire format was muted (since we did not load the format at startup.

I am just interested in how others feel.  john works 100% like it should now.  This is more for documenting names to match behavior.

magnumripper:
Good point, but yes I think we'll only introduce head ache if renaming it

src/bench.c
```diff
@@ -747,6 +747,27 @@ int benchmark_all(void)
 /* Silently skip formats for which we have no tests, unless forced */
 		if (!format->params.tests && format != fmt_list)
 			continue;
+		
+/* Format disabled in john.conf */
+		if (cfg_get_bool(SECTION_DISABLED, SUBSECTION_FORMATS,
```
magnumripper, line 752: Very strange: Even though line 750 looks fine here, and this supposedly is the very commit that was used, it now has whitespace noise (two tabs).

PR #4326, opened 2020-08-28
Revision of doc/SUBSETS and some comments in subsets.c

magnumripper:


doc/SUBSETS
```diff
@@ -1,7 +1,7 @@
 SUBSETS mode
 
 Subsets is a brute-force variant that tries to produce candidates in order of
-complexity, but without resorting to advanced stuff like Markov chains.  that
+complexity, but without resorting to advanced stuff like Markov chains.  That
 is, it will try a long poor password such as "gggggggggggggggggg#" much earlier
 than a short one with unique characters like "yok3" and in between them it will
 probably try "bubba" which is slightly longer but has repeated characters.
```
solardiz, line 7: Wouldn't it try "bubba" before "gggggggggggggggggg#"? Perhaps just drop the "#" from the first example so that it's actually tested before "bubba".

solardiz (approved):
Looks good to me, and the minor issue I pointed out just happened to be in patch context. I guess I'd have found something else if I looked at a different place.

doc/SUBSETS
magnumripper: True, and also it will actually *never* try "gggggggggggggggggg#" unless you use `--max-len=19` to bump max. length from the default of 16... but I did guard the example with the word "probably" 🤣 so I'll just let it be for now. Hopefully the reader gets the general idea.

Issue #3232, opened 2018-04-15
Add code for AES_cfb128


magnumripper:
`gpg_common_plug.c` needs AES in CFB mode so still uses OpenSSL. Implementing CFB is trivial so we should add it to our AES-NI stuff.

magnumripper:
Actually we even have CFB code in `encfs_common_plug.c`. That should be moved to `aes/` files.

PR #4328, opened 2020-08-31
ssh: support aes256-ctr for private keys encryption

### LINKED ISSUE #4069: ssh2john and john unable to brute-force password
### Steps to reproduce ###
I created the following private key with the password `123`.
[id_rsa.txt](https://github.com/magnumripper/JohnTheRipper/files/3484449/id_rsa.txt)

Converted it via `ssh2john.py` into a john-compatible format.
[id_rsa_johnformat.txt](https://github.com/magnumripper/JohnTheRipper/files/3484473/id_rsa_johnformat.txt)

Tried to get the used password by using john, but it never found a valid solution. For example:
```
john --mask=?d?d?d id_rsa_johnformat.txt
```

### System configuration ###
Attach details about your OS and about JtR, including:
- `$ ./john --list=build-info`.
```
Version: 1.9.0-jumbo-1
CPU tests: AVX2
$JOHN is ./run/
Format interface version: 14
Rec file version: REC4
Charset file version: CHR3
CHARSET_MIN: 1 (0x01)
CHARSET_MAX: 255 (0xff)
CHARSET_LENGTH: 24
SALT_HASH_SIZE: 1048576
SINGLE_IDX_MAX: 32768
SINGLE_BUF_MAX: 4294967295
Effective limit: Max. KPC 32768
Max. Markov mode level: 400
gcc version: 7.4.0
GNU libc version: 2.27 (loaded: 2.27)
Crypto library: OpenSSL
OpenSSL library version: 0100020ef
OpenSSL 1.0.2n  7 Dec 2017
File locking: fcntl()
fseek(): fseek
ftell(): ftell
fopen(): fopen
memmem(): System's
```

vkhromov:
Add support for ssh new-style private keys encrypted using `aes256-ctr`.

Fixes #4069

Test
```bash
$ ssh-keygen -t rsa -b 4096 -f id_rsa-aes256-cbc -Z aes256-cbc -N TestPassword
$ ssh-keygen -t rsa -b 4096 -f id_rsa-aes256-ctr -Z aes256-ctr -N TestPassword
$ echo TestPassword >passwords.lst
$ ./ssh2john.py id_rsa-aes256-cbc id_rsa-aes256-ctr >id_rsa.hash
$ ./john --wordlist=passwords.lst id_rsa.hash
...
TestPassword     (id_rsa-aes256-ctr)
TestPassword     (id_rsa-aes256-cbc)
```


magnumripper (commented):
This looks excellent, except for the fact we've been trying to get rid of high level OpenSSL such as EVP because the long-term goal is to drop OpenSSL altogether. Don't we have AES-256-CTR in our AES-NI stuff already, ready-to-use? I'm not sure right now.

Also, we should add OpenCL support as well, but I can handle that.

solardiz:
This is a much appreciated contribution. Thank you, @vkhromov! Some observations in addition to what magnum said, in arbitrary order:

1. Please add a test vector. That way, even our CI would test your added code right here in this PR. A difficulty is that there's no OpenCL implementation yet. The test vector may be added to `ssh_variable_code.h`, but wrapped in some `#if` enabling it only when included from the CPU format.

2. We probably need copyright statements from you in files with non-trivial edits (new lines of code added, rather than merely some checks expanded). While at it, I notice we could also need to clarify the copyright statements we already have in those files, but that's a separate task.

3. This needs a NEWS entry.

4. We include a space after `if` and before the opening brace. We also don't use the `constant == variable` style of writing comparisons. I'm aware those can catch inadvertent uses of `=` in place of `==`, but modern compilers warn about that anyway (gcc asks for extra braces around assignments used inside conditions) and this rationale doesn't apply to `!=` (except for consistency, if we were to use that for `==`). You could want to revise the coding style of your contribution to match our code in this respect, although I realize that we have lots of wrong/varying-style source code in jumbo as well.

Would you like to make any of these changes before we merge this PR, or should we merge first and then revise separately?

After merging this, we also need to open a new GitHub issue for adding this support to the OpenCL format.

Edit: and until we implement that, we could want to have the OpenCL format's `valid()` warn about the lack of support and suggest use of the CPU format instead. We have a similar problem, needing something like that, for the CPU vs. OpenCL Ethereum wallet formats, where we only support the kind using scrypt KDF on CPU. We know people will be confused if we don't even warn from the OpenCL format.

vkhromov:
@magnumripper @solardiz , thanks for taking a look.

It doesn't seem that `john` contains its own AES-256-CTR implementation at the moment, please correct me if I'm wrong.  I could try to add it if that's fine with you.

It also doesn't seem that `john` supports OpenCL for the new-format ssh private keys with any ciphers, please see https://github.com/openwall/john/blob/2be323f7c536ff7a7a5cc279372f8c0f2002d754/run/opencl/ssh_kernel.cl#L232 .  If that's correct then supporting AES-256-CTR probably means that code for AES-256-CBC also should be added.  I could do that as well.

As for the rest of your comments, I'll try to push the updated code soon.

vkhromov:
I've pushed the update addressing all non-OpenSSL/OpenCL-related comments.

magnumripper:
You seem to be right, the shared `run/opencl/opencl_aes.h` for OpenCL does have CBC and some others but not CTR at the moment. For CPU, I can see "something" in `src/aes/aesni/iaesni.h` with `git grep -i "aes.*256.*ctr" src/aes` but maybe it's not complete code? BTW see also #3232.

We should definitely fix both and you're definitely welcome to give it a shot!

doc/NEWS
```diff
@@ -122,6 +122,8 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
   the Easter egg option "--subsets=full-unicode" which BTW supersedes the
   external modes with far better performance.  [magnum; 2020]
 
+- Add support for ssh new-style private keys encrypted using `aes256-ctr`
+  cipher.
```
magnumripper: This should end with eg. `  [vkhromov; 2020]`

solardiz (approved):
Thanks. I think this is good enough for merging, and we can work on improving it further separately (further contributions are welcome!)

Should we keep the individual commits on merging or are they expected to be squashed?

vkhromov:
Thanks for your feedback!

I've updated `NEWS` file and squashed all commits into a single one.
So, this PR should be ready for the merge.

I'll try to handle the OpenCL implementation and the code which uses `EVP_` functions in a separate PR.

magnumripper:
Good stuff

solardiz:
> It doesn't seem that `john` contains its own AES-256-CTR implementation at the moment

Things have changed - we now have `mbedtls_aes_crypt_ctr` in `mbedtls/aes.h`.

Also, testing recent code on a deliberately old system (to figure out a bug elsewhere in the codebase), I just noticed that this PR broke builds with OpenSSL 1.0.0. If we continue using OpenSSL's AES-256-CTR, then we may want to wrap them in compile-time OpenSSL version checks (let the rest of the functionality build). I don't know which exact version of OpenSSL introduced support for `EVP_aes_256_ctr`. In 1.0.0t, it is somehow within an `#if 0` block inside `openssl/evp.h`, and then linking fails so it's actually missing.

solardiz:
I should add, this is literally the only line that prevents the build on the old system I test on:

```c
        if (EVP_DecryptInit_ex(ctx, EVP_aes_256_ctr(), NULL, key, iv) != 1)
```

With this line commented out, everything builds and works, except for the SSH format indeed.

So we really may want to repair this. Something you'd want to do, @vkhromov?

solardiz:
> I should add, this is literally the only line that prevents the build on the old system I test on:
> 
> ```c
>         if (EVP_DecryptInit_ex(ctx, EVP_aes_256_ctr(), NULL, key, iv) != 1)
> ```

I've now addressed this in #5745 by using our mbedTLS AES. The tree as of that PR now builds cleanly even with OpenSSL 1.0.0t, and SSH passes self-test.

PR #4336, opened 2020-09-10
'Rain' mode

e2002e:
This is clean,  
features are:  
- rotation of the characters  
- displacement of the state's indexes (spread across the word when length is odd, at the start and the end when the length is even).
Slower than subsets by four times on my pc.
I'll be trying with opencl.
Regards.


magnumripper:
Thanks!

First off, our build bots fails building in some conditions:
```
rain.c: In function 'submit':
rain.c:270:1: error: control reaches end of non-void function [-Werror=return-type]
 }
 ^
```
You need to have a look at that of course. Since it doesn't always happen, it's likely some `#ifdef` things out of whack.

Second, I really don't want to discourage you but... I think you still didn't convince us this mode adds any advantage over dumb brute-force. Does it? In what way?

src/options.c
```diff
@@ -135,8 +135,10 @@ static struct opt_entry opt_list[] = {
 		"%u", &options.subset_min_diff},
 	{"subsets-max-diff", FLG_ZERO, 0, FLG_SUBSETS_CHK, OPT_REQ_PARAM,
 		"%u", &options.subset_max_diff},
+	{"rain", FLG_RAIN_SET, FLG_CRACKING_CHK,
+		0, 0, OPT_FMT_STR_ALLOC, &options.rain_full},
 	{"mask", FLG_MASK_SET, FLG_MASK_CHK,
-		0, 0, OPT_FMT_STR_ALLOC, &options.mask},
+		0, 0, OPT_FMT_STR_ALLOC, &options.mask},
```
magnumripper: We're not very fond of added whitespace noise. Please revert that in line 141

src/options.h
```diff
@@ -183,6 +183,8 @@
 /* Subsets cracking mode */
 #define FLG_SUBSETS_CHK			0x0100000000000000ULL
 #define FLG_SUBSETS_SET			(FLG_SUBSETS_CHK | FLG_CRACKING_SET)
+#define FLG_RAIN_CHK			0x2000000000000000ULL //biggest flag yet
+#define FLG_RAIN_SET			(FLG_RAIN_CHK | FLG_CRACKING_SET)
```
magnumripper: Please move these lines to below the `FLG_NO_MASK_BENCH` line, so they come in numerical order

src/rain.h
```diff
@@ -0,0 +1,14 @@
+/*
+ * Copyright (c) 2018 magnum
```
magnumripper: I did not write this, so please remove my name from it

src/rain.c
```diff
@@ -0,0 +1,469 @@
+/*
+ * This software is Copyright (c) 2018 magnum
```
magnumripper: I did not write this, please remove my name from it and put your own there

src/rain.c
```diff
@@ -0,0 +1,469 @@
...
+ *
+ * IDEAS:
+ * - Store charset in target encoding for quicker conversion, even for
+ *   UTF-8 (we can store it in uint32_t)! Beware of endianness.
+ *
+ * RELATED:
+ * - Unicode ranges charsets. Generator? Standalone? --subsets-file=FILE?
+ * - Add global options --skip-odd-lengths and --skip-even-lengths (also
+ *   affecting mask mode and inc, possibly some external modes)
+ */
```
magnumripper: I think most of the comments above are remnants from `subsets.c` which you used as a template. Anything not applicable here should be dropped.

solardiz:
@e2002e On one hand, I appreciate the contribution and the persistence. On the other, I repeatedly asked you (including in private e-mail recently) to start by contributing this (if you insist, as you seem to) as a cleanly written external mode, with the logic clearly visible and also serving as an example of how such modes can be contributed. You instead insist(ed) on completing this pull request. Now it's here, but its future is unclear. While we accepted some dubious stuff into jumbo in the past, a new cracking mode is pretty significant (such as in terms of new user distraction from other modes that actually make sense) and we'd better have some rationale for it. Also, this is anything but "clean". Even the tiny bit of documentation is confusing, seeming to consider that the reader already knows what this mode is about.

magnumripper (changes requested):
I did a quick review but please note that fixing the few issues I commented on does not necessarily mean we'll ever accept this code - simply because we have yet to understand what good it actually does, if any.

ghost:
From john-dev.
> I am going to do these tests you asked for in a few hours from now and will
> let you know, but my hashes are simples and you should expect incremental
> mode to be the winner, but rain mode to find the long passwords before.


* we have yet to understand what good it actually does, if any.

E.g., I didn't see the results of your tests. 

Anyway, if you need a JtR build for your tests (Windows and/or Linux - Flatpak bundle) using this PR, I can do that. I can create a trial version, with those changes applied and that is usable (as a regular JtR package is).

src/rain.c
solardiz: magnum, since your Subsets mode code was used as a template, I guess this is derived work and needs to list both of you as copyright holders, unless the remnants of your code are so insignificant they're not subject to copyright. Separately, a comment may be added that you don't endorse this mode, or something. ;-)

It doesn't look likely that we'd accept this, but even if it exists in a fork of jumbo only, it should better have proper copyright statements and disclaimer.

e2002e:
Many reaction in no time !
I applied what was prescribed.
To answer the matter that is: what is rain mode ? and: is it effective ?
This is right now the combination of the rotation of characters from the set applied to each word, and of a displacement in the index of the bruteforce.
When the word is even, the characters iterate at the beginning of the word and wrap to continue from the end.
If it's odd, the word is split in 4 and the bruteforce goes from these.
So only the odd words are effective now, and the rotation is not enough to beat the entropy despite the displacement.
I'm cancelling this pull request and will be trying to get the rotation take some special effects.
Good day gentlemen.

src/rain.c
magnumripper: I think the "template" was merely for seeing how fix/save/restore_state and such things are implemented so I really don't want to be mentioned at all.

solardiz:
@e2002e Out of respect for your time, I suggest that you don't bother with anything other than possibly an external mode, or something not integrated with JtR at all. There's like a 99% chance we won't accept this as a built-in mode anyway. Thanks.

e2002e:
I think I'm done. 
The state index is strafing freely, and with the rotation of the character set it can be very effective, I'm ready for a pull.
[A video](https://youtu.be/Vqf8Bjyl1Dk) if you want to get a look at how it's going.
Solar thanks for your support, I needed to realize something was not working.

src/rain.c
```diff
@@ -0,0 +1,443 @@
+#include "os.h"
```
magnumripper, line 1: We're probably still not merging your PR but just for clarity, you shouldn't just remove all "my" copyright stuff but instead you should change them to valid ones pointing at yourself. I personally don't care about them a lot, but entities like Debian and Ubuntu do and we're struggling to please them.

e2002e:
Still working on the rotation effect, the strafe which produces a slide on the index works and could be a module in itself, but the rotation either make it slow down very very much or - when kept simple- produces duplicates.

e2002e:
Hi fellows, I reopening.
I understand the importance of a pull request, my apologies for the chaotic start.
The same technique that was used for playing on the characters is working on the index of the state with success,
I've updated the docs.
I wish you a good continuation whether you merge this or not in the end.

solardiz:
@e2002e You ended up replacing magnum's copyright statements with mine. That's wrong - I didn't write this. Please just drop them.

e2002e:
Not working

PR #4342, opened 2020-09-15
Remove "contributors" badge

ghost:
The number differs from the number shown on the main page.

I don't have a strong feeling about it. I guess members are not contributors.


solardiz (approved):
I also have no strong feelings, but since the badge is both redundant and inconsistent with GitHub's display of the number of contributors, let's remove it.

PR #4344, opened 2020-09-17
SSH formats fixes

### LINKED ISSUE #4028: Review ssh formats
See https://www.openwall.com/lists/john-users/2019/06/10/1 and #4026 which added FMT_NOT_EXACT to the GPU format while the CPU format apparently already had that but doesn't need it?!

- [ ] Both formats should set/clear FMT_NOT_EXACT at every `set_salt()` according to the actual conditions for current type. ~Should we instead remove FMT_NOT_EXACT from the CPU format or are *some* of the latter's algos  producing FP's while others doesn't (IIRC the CPU format supports more algos than the GPU format)?~
- [ ] Wouldn't it be trivial to add the needed extra tests to the GPU format, simply post-processing on CPU using code shared by the CPU format?
- [ ] From #4026: "*Another issue is that ssh_fmt_plug.c has custom `split` that converts the "hashes" to all-lowercase, and (correctly for it) sets `FMT_SPLIT_UNIFIES_CASE`. opencl_ssh_fmt_plug.c does neither. I see no reason for the lowercasing there - those are our own custom format "hashes", so we can enforce a particular case - but maybe it changed during our development? Since I don't know the history of this and have no time anyway, I'm not fixing it with this PR. But this inconsistency between the two formats is probably wrong.*"

magnumripper:
- Add FMT_SPLIT_UNIFIES_CASE to the OpenCL format, as the CPU format does so.
- Fix missing glue in "cost" reporting.
- Ensure the OpenCL format rejects hashes with unsupported ciphers.
- Implement full checking in OpenCL format and drop FMT_NOT_EXACT from both as they now seem to produce few enough of FP's and user can still request it manually with --keep-guessing option.
- A logic bug or two was squashed as well.

Closes #4028

### DIFF
--- a/doc/NEWS
+++ b/doc/NEWS
@@ -125,6 +125,12 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
 - Add support for ssh new-style private keys encrypted using `aes256-ctr`
   cipher.  [vkhromov; 2020]
 
+- Implement full checking in SSH OpenCL format, and drop FMT_NOT_EXACT from
+  both it and the CPU format.  Bugs were squashed as well.  For remaining
+  false-positives (if any), user can selectively use the
+  --keep-guessing option.  [magnum; 2020]
+
+
 Major changes from 1.8.0-jumbo-1 (December 2014) to 1.9.0-jumbo-1 (May 2019):
 
 - Updated to 1.9.0 core, which brought the following relevant major changes:
--- a/run/opencl/ssh_kernel.cl
+++ b/run/opencl/ssh_kernel.cl
@@ -40,7 +40,7 @@ typedef struct {
 	uint cracked;
 } ssh_out;
 
-inline void generate_key_bytes(int nbytes, uchar *password, uint32_t len, uchar *salt, unsigned char *key)
+inline void generate_key_bytes(int nbytes, uchar *password, uint32_t len, uchar *salt, uchar *key)
 {
 	uchar digest[16];
 	int keyidx = 0;
@@ -70,14 +70,14 @@ inline void generate_key_bytes(int nbytes, uchar *password, uint32_t len, uchar 
 	}
 }
 
-inline int check_padding_and_structure_EC(unsigned char *out, int length, int strict_mode)
+inline int check_padding_and_structure_EC(uchar *out, int length)
 {
 	struct asn1_hdr hdr;
 	const uint8_t *pos, *end;
 
 	// First check padding
 	if (check_pkcs_pad(out, length, 16) < 0)
-		return -1;
+		return 0;
 
 	/* check BER decoding, EC private key file contains:
 	 *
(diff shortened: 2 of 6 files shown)


magnumripper:
After this is merged I might have a look at implementing the missing cipher types (bcrypt + AES, CBC or CTR) as well.

doc/NEWS
```diff
@@ -125,6 +125,12 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
 - Add support for ssh new-style private keys encrypted using `aes256-ctr`
   cipher.  [vkhromov; 2020]
 
+- Implement full checking in SSH OpenCL format, and drop FMT_NOT_EXACT from
```
solardiz, line 128: `NEWS` is meant to be end-user friendly - we shouldn't mention things internal to source code in there. So instead of mentioning `FMT_NOT_EXACT`, we should explain that change's effect in plain English.

Anyway, I suppose we'll need to go over `NEWS` and add/revise/re-group its entries before the next release.

solardiz:
> Add FMT_SPLIT_UNIFIES_CASE to the OpenCL format, as the CPU

That's weird. Those "hashes" are produced only by our own tool, aren't they? Were different-case hashes ever produced? Why did we add case unification for this format in the CPU format? If there's no valid reason, then maybe drop it from both?

> Ensure the OpenCL format rejects hashes with unsupported ciphers.

Does this imply we'd have false negatives before?

magnumripper:
> > Add FMT_SPLIT_UNIFIES_CASE to the OpenCL format, as the CPU
> 
> That's weird. Those "hashes" are produced only by our own tool, aren't they? Were different-case hashes ever produced? Why did we add case unification for this format in the CPU format? If there's no valid reason, then maybe drop it from both?

For some reason, our tool produces mixed-case hashes, some fields uc and some lc.

I think since the CPU format had the case unify, both now need it: Dropping it would mean we no longer recognize old pot entries.

> > Ensure the OpenCL format rejects hashes with unsupported ciphers.
> 
> Does this imply we'd have false negatives before?

Yep, there must have been 100% false negatives for bcrypt + AES 😢

solardiz:
> Dropping it would mean we no longer recognize old pot entries.

That's only true if the unification actually changed the strings produced by our ssh2john. Did it, or were those already of the target case?

magnumripper:
I amended my post. They actually changed

solardiz (approved):
Looks reasonable to me, and I agree with the reasoning in the discussion we just had. Indeed, I can't confirm that every code change in here is right, but I saw nothing obviously wrong either.

solardiz:
@magnumripper and all: let's adopt this convention for commit messages:

```
Subsystem: Change title up to 72 chars not ending in a dot

Optionally one or more paragraphs of text here with lines wrapped at 79.

Fixes #whatever, etc. (optional)
```

My recent commits follow this pattern. Claudio's are similar too. magnum's are not.

solardiz:
Hmm, or should we wrap the text paragraphs' lines at 75, since git log indents messages by 4 chars? I guess so.

magnumripper:
I'll try to remember that. I normally use Git Gui and I've been trying to keep it "slightly under 80" including the indent. What is the rationale for the shorter first line?

solardiz:
> What is the rationale for the shorter first line?

See how GitHub wraps the long first line e.g. in 4f5f6fc8dca0102da7e307e44d5600af04c00ca9. Actually, looks like it wrapped that one at 70, so maybe 70 and not 72 is the limit?

magnumripper:
Oh, that. I consider it a GitHub "problem". I guess it's mostly related to using bold/larger text - considering a proportional font it might not even be a consistent limit in character count.

PR #4352, opened 2020-09-21
Fix bug in --stress-test option handling.

### LINKED ISSUE #4351: --stress-test option no longer works
Investigating #4295 I just noticed `--stress-test` option is b0rken. It used to be same as `--test` except it looped the test(s) forever.
```
Invalid options combination or duplicate option: "--stress-test"
```

magnumripper:
Closes #4351

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -187,11 +187,10 @@ static struct opt_entry opt_list[] = {
 		OPT_FMT_STR_ALLOC, &show_uncracked_str},
 	{"test", FLG_TEST_SET, FLG_TEST_CHK,
 		0, ~FLG_TEST_SET & ~FLG_FORMAT & ~FLG_SAVEMEM & ~FLG_DYNFMT &
-		~FLG_MASK_CHK & ~FLG_NOLOG & ~OPT_REQ_PARAM,
-		"%d", &benchmark_time},
+		~FLG_MASK_CHK & ~OPT_REQ_PARAM, "%d", &benchmark_time},
 	{"test-full", FLG_TEST_SET, FLG_TEST_CHK,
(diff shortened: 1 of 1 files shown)


solardiz (approved):
So apparently this broke when we enabled mask for "--test" by default. I guess it was also broken for explicit "--test --mask" before. Don't we have the same problem for "--test-full"? Should fix that one as well?

Why are we allowing(?) "--no-log" with "--test" and "--stress-test", can any logging occur during testing? Why are we not allowing that with "--test-full", then?

ghost:
May I merge this?

solardiz:
> May I merge this?

I don't mind, but as I pointed out there are probably a few more closely related things to fix, which we shouldn't forget to at least consider.

magnumripper:
> Why are we allowing(?) "--no-log" with "--test" and "--stress-test", can any logging occur during testing? Why are we not allowing that with "--test-full", then?

I guess they are minor bugs and the reasons are twofold: First, it took me years to even understand how to put these flags right. Second and more important, we were out of options flags for quite some while before I extended them to 64 bits (and now we're severly close to run out of them again).

@claudioandre-br if you want this now, just merge and open an issue for the remaining issues. If you wait with merging I'll try to amend this PR within 24h or so

magnumripper:
> Don't we have the same problem for "--test-full"? Should fix that one as well?

Apparently not, I believe it doesn't use mask.

> Why are we allowing(?) "--no-log" with "--test" and "--stress-test", can any logging occur during testing? Why are we not allowing that with "--test-full", then?

I'll remove that and amend this PR.

magnumripper:
Done, anyone merge once bots are done

solardiz (approved):
The revised commit looks good to me.

Issue #4295, opened 2020-08-12
CL_OUT_OF_HOST_MEMORY with "--test --format=opencl"


solardiz:
When `--test --format=opencl` on a p3.2xlarge AWS instance doesn't crash on #4294, it crashes later with `CL_OUT_OF_HOST_MEMORY`:

```
$ john -test -form=opencl | tee -a log-v100-opencl
[...]
Benchmarking: descrypt-opencl, traditional crypt(3) [DES OpenCL/mask accel]... LWS=128 GWS=65536 Note: Building per-salt kernels. This takes e.g. 2 hours for 4096 salts.
DONE
Warning: "Many salts" test limited: 16/256
Many salts:     1323M c/s real, 1317M c/s virtual
Only one salt:  1265M c/s real, 1252M c/s virtual

Benchmarking: diskcryptor-opencl, DiskCryptor [PBKDF2-SHA512 OpenCL]... (8xOMP) LWS=256 GWS=163840 (640 blocks) DONE
Speed for cost 1 (iteration count) of 1000
Raw:    319687 c/s real, 55586 c/s virtual

Benchmarking: diskcryptor-aes-opencl, DiskCryptor AES XTS (only) [PBKDF2-SHA512 AES OpenCL]... LWS=256 GWS=327680 (1280 blocks) DONE
Speed for cost 1 (iteration count) of 1000
Raw:    1019K c/s real, 976068 c/s virtual

Benchmarking: dmg-opencl, Apple DMG [PBKDF2-SHA1 3DES/AES OpenCL]... LWS=32 GWS=40960 (1280 blocks) DONE
Speed for cost 1 (iteration count) of 1000, cost 2 (version) of 2 and 1
Raw:    1854K c/s real, 1845K c/s virtual

Benchmarking: electrum-modern-opencl, Electrum Wallet 2.8+ [PBKDF2-SHA512 OpenCL]... (8xOMP) LWS=256 GWS=327680 (1280 blocks) DONE
Raw:    21480 c/s real, 2740 c/s virtual

Benchmarking: EncFS-opencl [PBKDF2-SHA1 AES OpenCL]... LWS=32 GWS=81920 (2560 blocks) DONE
Speed for cost 1 (iteration count) of 181474 and 181317
Raw:    18788 c/s real, 18703 c/s virtual

[...]

PR #4353, opened 2020-09-21
Fix OpenCL leak

ghost:
Partial fix of #4295

### DIFF
--- a/src/opencl_autotune.h
+++ b/src/opencl_autotune.h
@@ -210,6 +210,7 @@ static void autotune_run_extra(struct fmt_main *self, unsigned int rounds,
 		}
 		find_best_gws(self, gpu_id, rounds, max_duration1, have_lws);
 	} else {
+		release_clobj();
 		create_clobj(global_work_size, self);
 	}
 
--- a/src/opencl_bitlocker_fmt_plug.c
+++ b/src/opencl_bitlocker_fmt_plug.c
@@ -286,17 +286,19 @@ static size_t get_task_max_work_group_size()
 
 static void release_clobj(void)
 {
-	HANDLE_CLERROR(clReleaseMemObject(d_vmk), "Release");
-	HANDLE_CLERROR(clReleaseMemObject(d_pswI), "Release");
-	HANDLE_CLERROR(clReleaseMemObject(d_pswSize), "Release");
-	HANDLE_CLERROR(clReleaseMemObject(d_found), "Release");
-	HANDLE_CLERROR(clReleaseMemObject(d_wblocks), "Release");
-	HANDLE_CLERROR(clReleaseMemObject(d_salt), "Release");
-	HANDLE_CLERROR(clReleaseMemObject(d_pad), "Release");
+	if (h_found) {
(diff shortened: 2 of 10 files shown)


ghost:
Something like this will be needed in a couple of places:
```
diff --git a/src/opencl_tc_fmt_plug.c b/src/opencl_tc_fmt_plug.c
index c3e7ce57c..1461a469c 100644
--- a/src/opencl_tc_fmt_plug.c
+++ b/src/opencl_tc_fmt_plug.c
@@ -156,14 +156,16 @@ static void create_clobj(size_t gws, struct fmt_main *self)
 
 static void release_clobj(void)
 {
-       HANDLE_CLERROR(clReleaseMemObject(mem_in), "Release mem in");
-       HANDLE_CLERROR(clReleaseMemObject(mem_setting), "Release mem setting");
-       HANDLE_CLERROR(clReleaseMemObject(mem_out), "Release mem out");
-
-       MEM_FREE(inbuffer);
-       MEM_FREE(outbuffer);
-       MEM_FREE(keyfiles_data);
-       MEM_FREE(keyfiles_length);
+       if (inbuffer) {
+               HANDLE_CLERROR(clReleaseMemObject(mem_in), "Release mem in");
+               HANDLE_CLERROR(clReleaseMemObject(mem_setting), "Release mem setting");
+               HANDLE_CLERROR(clReleaseMemObject(mem_out), "Release mem out");
+
+               MEM_FREE(inbuffer);
+               MEM_FREE(outbuffer);
+               MEM_FREE(keyfiles_data);
+               MEM_FREE(keyfiles_length);
+       }
 }
 
 static void done(void)

```

ghost:
Done. 

To test I have to remove MsCash (does it use shared code or not?). Not sure what I should do.

More tests are needed, but, I would say some formats are leaking, I'm almost sure (and not because of shared code). 
The reference test (mine, at least) is sha256crypt-opencl: it has no leak.
****
I'm still testing, anyway, you guys can review and give a feedback.

magnumripper:
Thanks, I'll have a look at this within 18h. During my debugging previously today I didn't see (using assertions) any missing call to `release_clobj()` from autotune so I'm not sure this is all there is to it.

magnumripper:
I think the `if (inbuffer)` was present in all formats I looked at, and anyway missing that would not result in memory leaks but the opposite - trying to free already freed stuff. I *think* it's allowed to call `clReleaseMemObject()` once too much (although perhaps not clearly documented by Khronos so maybe we shouldn't depend on it) and our `MEM_FREE` macro has a safety net in place for null pointers.

magnumripper:
OK so of your two approaches, one (this) is to add a `release_clobj()` in a single file `opencl_autotune.h` and the other (#4354) is to do it to all formats. I think I like this first one better unless you saw some caveat?

ghost:
> add a release_clobj() in a single file opencl_autotune.h and the other (#4354) is to do it to all formats. I think I like this first one better unless you saw some caveat?

A single `release_clobj()` in `opencl_autotune.h` do not fix all leaks (it is NOT enough) .

magnumripper:
Ah OK. Then let's close this PR and merge #4354 for a starter

PR #4358, opened 2020-09-23
zip2john and pkzip format: Bump MAX_PKZ_FILES from 3 to 8

### LINKED ISSUE #3978: zip2john: allow more than 3 files
As the pkzip stuff is included in hashcat now and able to get much higher speeds, it would be beneficial to have some more flexibility in zip2john. Namely, in case of only 1 byte checksums, and enough available files, a hash with 7 or 8 files could be created with the start of the file only, as then still enough data would be available to avoid easy collisions (8x8bit => 64bit).

Also when having 2 byte checksums, 3 checksums might not be enough, depending on the available hardware, so probably could be extended to 4 files (also giving 64bits then) there with the -c flag.

When I implemented the initial kernels with the 8 files, I did some hack to zip2john, but I not able to fully understand the way how zip2john handles all that, so I thought it might be better to let you decide how you could do that.

Thanks in advance :)

magnumripper:
Some code did not honor that macro, this was fixed.  Closes #3978

### DIFF
--- a/src/pkzip.h
+++ b/src/pkzip.h
@@ -14,7 +14,7 @@ u64 fget64LE(FILE *fp);
 u32 fget32LE(FILE *fp);
 u16 fget16LE(FILE *fp);
 
-#define MAX_PKZ_FILES 3
+#define MAX_PKZ_FILES 8
 
 // These came from the gladman headers
 #define KEY_LENGTH(mode)    (8 * ((mode) & 3) + 8)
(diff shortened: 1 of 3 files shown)


solardiz (approved):
I can't really review that as I'm not familiar with the original code, but what I see looks sane to me. I also read about the 8 times 8-bit reasoning (effectively 64-bit to avoid false positives), which justifies why 8.

Issue #4313, opened 2020-08-25
KeePass-opencl ASan crash


solardiz:
On "super", devtoolset-3, testing all formats (CPU and OpenCL), running against the Xeon Phi device (autodetected as "best" since the GPUs are somehow not visible as OpenCL devices in this ASan-enabled build):

```
Benchmarking: KeePass-opencl [SHA256 AES/Twofish/ChaCha OpenCL]... 2 warnings generated.
Build log: Compilation started
2:243:20: warning: passing 'volatile uchar [32]' to parameter of type 'const void *' discards qualifiers
/home/solar/j/bleeding-jumbo/run/opencl/opencl_aes.h:159:36: note: passing argument to parameter '_in' here
2:249:32: warning: passing 'volatile uchar [32]' to parameter of type 'const uchar *' (aka 'const unsigned char *') discards qualifiers
/home/solar/j/bleeding-jumbo/run/opencl/opencl_chacha.h:239:55: note: passing argument to parameter 'c' here
Compilation done
Linking started
Linking done
Build started
Kernel <keepass_init> was successfully vectorized
Kernel <keepass_loop> was successfully vectorized
Kernel <keepass_final> was successfully vectorized
Done.
=================================================================
==28180==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x6110029471c0 at pc 0xa0f483 bp 0x7ffc1aa0b2e0 sp 0x7ffc1aa0b2d8
READ of size 4 at 0x6110029471c0 thread T0
    #0 0xa0f482 in cmp_all /home/solar/j/bleeding-jumbo/src/opencl_keepass_fmt_plug.c:285
    #1 0x94e53c in opencl_find_best_lws /home/solar/j/bleeding-jumbo/src/opencl_common.c:1808
    #2 0x95a6c8 in autotune_find_best_lws /home/solar/j/bleeding-jumbo/src/opencl_autotune.c:93
    #3 0xa11eda in find_best_lws /home/solar/j/bleeding-jumbo/src/opencl_autotune.h:71
    #4 0xa1293f in autotune_run_extra /home/solar/j/bleeding-jumbo/src/opencl_autotune.h:217
    #5 0xa1323f in autotune_run /home/solar/j/bleeding-jumbo/src/opencl_autotune.h:304
    #6 0xa1379f in reset /home/solar/j/bleeding-jumbo/src/opencl_keepass_fmt_plug.c:189
[...]

Issue #4044, opened 2019-07-05
Inconsistent OpenCL error handling in format-specific vs. common code


kevingy:
### Steps to reproduce ###
Using john-1.9.0-jumbo-1-win64 from Openwall, run:
`john --test --format=RAR5-opencl`
from the run directory

### Results: ###
This error is observed:
```
Device 2: Intel(R) HD Graphics 4000
Benchmarking: RAR5-opencl [PBKDF2-SHA256 OpenCL]... 0: OpenCL CL_INVALID_EVENT (-58) error in opencl_common.c:1828 - clWaitForEvents
```

### Notes: ###
I don't know anything about the code or the interface with OpenCL, but these notes might (or might not) help tracking down the problem.
- _john --test --format=RAR5_ works on the same machine without error
- _john --test --format=rar-opencl_ works on the same machine without error
- The RAR5-opencl benchmark test works on another machine with an Nvidia GPU

### System configuration ###
See attached.
[build-info.txt](https://github.com/magnumripper/JohnTheRipper/files/3362653/build-info.txt)
[opencl-devices.txt](https://github.com/magnumripper/JohnTheRipper/files/3362656/opencl-devices.txt)

kevingy:
BTW, I'm getting the same error for 7z-opencl:

`Will run 4 OpenMP threads
Device 2: Intel(R) HD Graphics 4000
Benchmarking: 7z-opencl, 7-Zip (512K iterations) [SHA256 AES OpenCL]... (4xOMP) 0: OpenCL CL_INVALID_EVENT (-58) error in opencl_common.c:1828 - clWaitForEvents`

magnumripper:
My first, second and third guesses would be this is a driver/runtime problem (ie. with your Intel OpenCL driver). Just for curiosity, would the version of said driver satisfy Hashcat's requirements?

kevingy:
That's a good question. I attempted to run to benchmark RAR5 using this commandline:
> hashcat-5.1.0>hashcat64 --benchmark -m 13000

That resulted in the following
```
hashcat (v5.1.0) starting in benchmark mode...

Benchmarking uses hand-optimized kernel code by default.
You can use it in your cracking session by setting the -O option.
Note: Using optimized kernel code limits the maximum supported password length.
To disable the optimized kernel code in benchmark mode, use the -w option.

* Device #2: Intel's OpenCL runtime (GPU only) is currently broken.
             We are waiting for updated OpenCL drivers from Intel.
[...]

PR #4359, opened 2020-09-24
LWS can't be greater than GWS

ghost:
Initialize LWS with the value of GWS if the initial value of LWS is greater than GWS.

### DIFF
--- a/src/opencl_common.c
+++ b/src/opencl_common.c
@@ -1769,6 +1769,10 @@ void opencl_find_best_lws(size_t group_size_limit, int sequential_id,
 		wg_multiple = get_kernel_preferred_multiple(sequential_id,
 		              crypt_kernel);
 
+	// In case LWS would be greater than GWS, adjust LWS value.
+	if (wg_multiple > global_work_size)
+		wg_multiple = global_work_size;
+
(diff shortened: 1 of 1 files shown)


ghost:
I am able to reproduce the `CL_INVALID_EVENT` error. 

It happens because an OpenCL operation [1] failed and the error was ignored, as we are auto tuning. So, I like `CL_INVALID_EVENT`, it highlights bugs.

[1] in the example, a `clEnqueueWriteBuffer` failed. It was before the kernel execution, so `CL_INVALID_EVENT` line 1818.

```
#1 Benchmarking: raw-SHA512-opencl [SHA512 OpenCL/mask accel]... 
[...]
0: OpenCL CL_INVALID_EVENT (-58) error in opencl_common.c:1818 - clWaitForEvents
```

See also #4044

ghost:
The erratic behavior of NVIDIA 1080 is good for detecting bugs.

magnumripper (approved):

magnumripper:
How is it erratic? I love nvidia for generally being so trouble-free compared to AMD.

ghost:
When it throats, it starts to behave like another board. Then, the code path starts to change.

solardiz:
This PR is now known to have fixed #4313.

PR #4370, opened 2020-09-26
Add support for --format=LIST (ie. --format=NAME,NAME[,...])

### LINKED ISSUE #4366: Add --format=LIST feature
From https://github.com/openwall/john/pull/4354#issuecomment-698595869
> I see things like notes-opencl segfaulting on an all-formats test (right after LM-opencl, who would have thought) but not on its own (we've seen that before). I'm actually considering adding option `--format-list=LIST` so I could test with `--format=lm-opencl,notes-opencl`. I'd use the existing list parsing of course, like we do with several other recent-ish options like `--single-seed=WORD[,WORD...]`

Some thoughts right now:
- Perhaps we can use `--format=[FORMAT|WILDCARD|CLASS|LIST]` (with LIST being the only new) instead of a separate `--format-list=LIST`.
- The existing list parsing would (IIRC) cut at the `-` in `foobar-opencl`, which isn't great.
- ~This is not really useful for anything but testing. Perhaps we don't even want it always active? OTOH I can't see it really be in the way or a lot of code.~

magnumripper:
Support a comma-separated list of formats given with --format option.
The order given will be honored.  Closes #4366 and extremely useful (along with `--test`) for several open issues.

### DIFF
--- a/doc/OPTIONS
+++ b/doc/OPTIONS
@@ -394,29 +394,41 @@ their number could potentially be decreasing as some hashes get cracked.
 To have the cracked hashes (and possibly salts) removed from all
 processes, you may interrupt and restore the session once in a while.
 
---format=NAME			force hash type NAME
+--format=NAME[,NAME...]		force hash type NAME
 
-Allows you to override the hash type detection.  As of John the Ripper
-version 1.8.0, valid "format names" are descrypt, bsdicrypt, md5crypt,
-bcrypt, LM, AFS, tripcode, dummy, and crypt (and many more are added in
-Jumbo).  You can use this option when you're starting a cracking session
-or along with one of: "--test", "--show", "--make-charset".  Note that
-John can't crack hashes of different types at the same time.  If you
-happen to get a password file that uses more than one hash type, then
-you have to invoke John once for each hash type and you need to use this
-option to make John crack hashes of types other than the one it would
-autodetect by default.  Mainly for test and list purposes (see --list=WHAT),
-you can use group aliases "dynamic", "cpu", "gpu", "opencl" and "cuda"
-as a format name, or use one wildcard, as in "--format=mysql-*",
-"--format=raw*ng" or "--format=*office".  There's also a special way
-to match any substring within the "algorithm name" (the string shown
-within brackets, e.g. "[DES 128/128 AVX-16]" using @tag syntax as in
-e.g. "--format=@DES" or "--format=@AVX".  Similarly you can match a
-substring in the "format name" (the longer name, not the label) using
-#tag format.  For example, if you forgot which format handles the IPMI
-hashes, you can say --format=#ipmi or --format=#ipmi-opencl and the
-correct format (in this case RAKP or RAKP-opencl) will be used.
-Regardless of method, the matching ignores case.
+Override the hash type auto-detection.  You can use this option when
+you're starting a cracking session or along with things like: "--test",
+"--show", "-list" or "--make-charset".  Note that John can't crack hashes
+of different types at the same time.  If you happen to get a password file
+that uses more than one hash type, then you have to invoke John once for
+each hash type and you need to use this option to make John crack hashes of
+types other than the one it would autodetect by default.  For normal use,
+you'd use a full exact format label such as "--format=crypt-md5".  Mainly
+for test and list purposes (see --list=WHAT), you can use a single wildcard,
+as in "--format=mysql-*", "--format=raw*ng" or "--format=*office", or group
+aliases such as "dynamic", "cpu", "omp", "opencl", "ztex" as a format name.
+There's also a special way to match any substring within the "algorithm
+name" (the string shown within brackets, e.g. "[DES 128/128 AVX-16]")
+using @tag syntax as in e.g. "--format=@DES" or "--format=@AVX". Similarly
+you can match a substring in the "format name" (the longer name, not the
+label) using #tag format.  For example, if you forgot which format handles
+the  IPMI hashes, you can say --format=#ipmi and the correct format (in
+this case RAKP) will be used.  Regardless of method, the matching ignores
+case.  Several formats can be specified at once, separated by commas, and
+prepending a format, wildcard or class with a "-" will exclude instead of
+include (eg. "--format=#ipmi,-cpu" for again finding the RAKP format but
+this time rejecting the CPU version so will end up finding "RAKP-OpenCL"
+as long as your build supports OpenCL).  The order given of any "include"
+parts of a format list will be honored so this it feature can also be used
+for merely changing format precedence during auto-detection. Prepending a
+format, wildcard or class with a "+" will promote it to a requirement as
+opposed to an inclusion. That is, "--format=omp,+cpu" will first "include"
+any OMP formats but then prune all of them that does not match format class
+"CPU".
+
+The ad-hoc dynamic, or "dynamic compiler" format can not be used in a format
+list or with wildcards, only on it's own.  This is because the expression
+may need to include commas, wildcard etc.
 
 "--format=crypt" may or may not be supported in a given build of John.
 In default builds of John, this support is currently only included on
--- a/src/formats.c
+++ b/src/formats.c
@@ -3,6 +3,12 @@
  * Copyright (c) 1996-2001,2006,2008,2010-2013,2015 by Solar Designer
  */
 
+#if AC_BUILT
+#include "autoconfig.h"
+#else
+#define _GNU_SOURCE 1 /* for strcasestr */
+#endif
+
 #include <stdint.h>
 #include <stdio.h>
 #include <string.h>
@@ -14,6 +20,7 @@
 #include "misc.h"
 #include "unicode.h"
 #include "base64_convert.h"
+#include "dynamic_compiler.h"
 #ifndef BENCH_BUILD
 #include "options.h"
 #include "loader.h"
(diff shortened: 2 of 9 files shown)


doc/OPTIONS
```diff
@@ -394,7 +394,7 @@ their number could potentially be decreasing as some hashes get cracked.
 To have the cracked hashes (and possibly salts) removed from all
 processes, you may interrupt and restore the session once in a while.
 
---format=NAME			force hash type NAME
+--format=NAME[,NAME...]		force hash type NAME
```
magnumripper, line 397: I decided not to make the above change in the usage blob, bc it would just lead users to the confusion already discussed about "simultaneous cracking". In here, it's made clear.

src/formats.c
```diff
@@ -54,6 +54,114 @@ static void test_fmt_case(struct fmt_main *format, void *binary,
...
 #endif
 
+static char *fmt_classes[] = { "all", "cpu", "opencl", "dynamic", "mask", "omp", "cpu+omp", NULL };
+
+static char* is_format_class(char *name)
+{
+	int idx = 0;
+
+	if (strchr(name, '*') || strchr(name, '#'))
+		return "wildcards";
```
magnumripper: I had forgotten about the `#` wildcard. It's a gem! And lo and behold, it's documented.

src/formats.c
```diff
@@ -54,6 +54,114 @@ static void test_fmt_case(struct fmt_main *format, void *binary,
...
+{
+	struct fmt_main *current = *full_fmt_list;
+	struct fmt_main *prev = NULL;
+
+	do {
+		if (!strcasecmp(label, current->params.label)) {
+			if (current == *full_fmt_list)
+				*full_fmt_list = current->next;
+			else if (prev && current->next)
+				prev->next = current->next;
```
magnumripper: Lines 125-128 made me 10 years younger in 23 minutes

magnumripper (commented):
Self-review (alas, I can't approve 😆 )

src/john.c
```diff
@@ -430,6 +436,10 @@ static void john_register_all(void)
 	john_register_one(&fmt_crypt);
 #endif
 
+	/* Do we have --format=LIST? If so, re-build fmt_list from it, in requested order. */
+	if (format_list_string && !fmt_check_custom_list(format_list_string))
+		error_msg("Could not parse format list '%s'\n", format_list_string);
```
magnumripper: This should actually be dead code, 440 can't be true. But nevermind, it's good measure I guess.

magnumripper:
Another one of these random buggers from buildbots
```
Testing: rar, RAR3 (length 5) [SHA1 512/512 AVX512BW 16x AES]... (2xOMP) 

Too long with no output (exceeded 10m0s): context deadline exceeded
```

src/formats.c
```diff
@@ -54,6 +54,114 @@ static void test_fmt_case(struct fmt_main *format, void *binary,
 	int *plaintext_has_alpha, struct db_salt *dbsalt);
 #endif
 
+static char *fmt_classes[] = { "all", "cpu", "opencl", "dynamic", "mask", "omp", "cpu+omp", NULL };
```
solardiz: Should be `static const char * const fmt_classes[]`.

src/formats.c
```diff
@@ -54,6 +54,114 @@ static void test_fmt_case(struct fmt_main *format, void *binary,
...
+
+static char* is_format_class(char *name)
```
solardiz: `static const char *is_format_class(const char *name)`

src/formats.c
```diff
@@ -54,6 +54,114 @@ static void test_fmt_case(struct fmt_main *format, void *binary,
...
+			fmt_register(current);
+			return 1;
+		}
+		prev = current;
+	} while ((current = current->next));
+
+	return 0;
+}
+
+int fmt_check_custom_list(char *list_string)
```
solardiz: `const char *list_string`

src/formats.h
```diff
@@ -444,6 +444,11 @@ extern struct fmt_main *fmt_list;
  */
 extern void fmt_register(struct fmt_main *format);
 
+/*
+ * Check for --format=LIST and if so, re-populate fmt_list from it.
+ */
+extern int fmt_check_custom_list(char *list_string);
```
solardiz: `const char *list_string`

src/john.c
```diff
@@ -383,11 +383,17 @@ static void john_register_one(struct fmt_main *format)
 
 static void john_register_all(void)
 {
+	char *format_list_string = NULL;
```
solardiz: `const`

solardiz (approved):
I've suggested some const'ification. I might well have missed more serious stuff, but overall this looks OK to me. Thanks!

ghost:
> Another one of these random buggers from buildbots

GitHub is not updating the status of `circleci: fast-formats-omp`. You can force-push to see if it works (or, you will need to wait Solar for merging)>

solardiz:
> GitHub is not updating the status of `circleci: fast-formats-omp`.

This is not required for merging. I think it's one of those optional tests that we allow to complete after merging.

ghost:
> > GitHub is not updating the status of `circleci: fast-formats-omp`.
> 
> This is not required for merging. I think it's one of those optional tests that we allow to complete after merging.

Indeed (my fault). I was going to say that magnum's last run `35m 36s / 11m ago`, failed again. I'm running it again.

magnumripper:
I'll add the const's. Then perhaps we should merge this before I try to make a larger hack out of combining classes/wildcard/groups and lists (they are now implemented in totally different places so in the end it should be cleaner).

magnumripper:
> I'll add the const's. Then perhaps we should merge this (...)

No let's wait, I think I can picture the much cleaner implementation of wildcards/groups/classes in this new code, restoring `john_register_(one|all)` to mostly unchanged from john proper.

ghost (approved):
Magnum! Magnum! Magnum! Magnum!

Everybody together! Once again!

solardiz:
magnum, do you want this merged as-is for now, or do you prefer to rework it first? This is useful functionality already for our current debugging.

magnumripper:
New version committed, we'll squash before merging. We now support wildcards, groups etc within format lists as well (all such code is in one place now).

Apart from needing a new code review and actual testing I'm not 100% happy with syntax/quirks yet, see below.

This is unchanged from previous code:
- `*` is a wildcard for `format_label`. Only one can be used per item so `oldoffice*`, `*office` or `old*opencl` are valid while `*office*` is not.
- `@string` matches substring "string" in `algorithm_name`.
- `#string` matches substring "string" in `format_name` (n.b. not the label but the longer name).

Classes changed slightly. We now support "dynamic", "cpu", "opencl", "ztex", "mask" and "omp" but not any compound ones like "cpu-dynamic" (now written as `cpu,-dynamic`). The two latter look at format flags ("mask" means "include formats that features device-side mask").

*Any* exact name, wildcard or class can now be prepended with `-` for excluding instead of including.

It works like this:
- The order of exclusions given does *not* matter. All exclusions are made first, dropping them from the original `fmt_list`.
- On second pass all inclusions are made, bulding a brand new `fmt_list` (and order is honored here). If only exclusions were listed, we're using *all* formats remaining after exclusion, in whatever (old) order they happen to come.

Some specific quirks (not sure how to handle it better):
- Any dupe list format items will be silently ignored (we're simply using `list_add_unique`).
- Exclusions not actually matching (excluding) anything, such as `--format=-foobar` will *not* be warned for. I figured you don't want to see warnings about eg. saying `--format=-opencl` using a non-OpenCL build (what do you think?).
- Using the previous code/hack, we could say things like `--format=#ipmi-opencl` which would find substring "ipmi" in the long format name, and then require OpenCL as well (ending up in `RAKP-opencl` only) - a fairly confusing syntax given the dash/minus sign. The new code don't have any compound things so to get the same result we need to say `--format=#ipmi,-cpu` (exclude all CPU formats, then find "ipmi" among the remaining). I'm not 100% happy with having to say "not cpu" when I want "opencl". *EDIT: We can now write it as `--format=#ipmi,+opencl`*
- Previous code also supported some undocumented specials compounds like `--format=cpu+omp`. They are dropped, but can be accomplished with the new more consistent format, for example using `omp,-opencl`. Again I'm not 100% happy with having to list "not opencl" instead of what I want: "cpu".  *EDIT: We can now write it as `--format=cpu,+omp`*

Perhaps I should add `+` as a prefix, sort of like the `-`? Exclusions are made in the first pass, inclusions in second and this new `+` inclusions would **have** to match in the third pass. Then we could say things like `-format=#ipmi,+cpu` and it would mean "all formats matching "IPMI" **and** of class "cpu". I think that would solve what I find slightly backwards above.

magnumripper:
> Perhaps I should add `+` as a prefix, sort of like the `-`? Exclusions are made in the first pass, inclusions in second and this new `+` inclusions would **have** to match in the third pass. Then we could say things like `-format=#ipmi,+cpu` and it would mean "all formats matching "IPMI" **and** of class "cpu". I think that would solve what I find slightly backwards above.

Committed this now. I call it "requirements". So now we can say `--format=#ipmi,-cpu` or `--format=#ipmi,+opencl` and it will give the same result (until someone writes a `RAKP-ztex`...)

So there's three passes now: First we (may) reject stuff from the full format list, using "-". Then we (may) pick stuff from it to our new format list. Finally we (may) require properties, using "+" and any formats in our new format list that doesn't match them will be pruned. I'm pretty happy with this now. Fairly clean, fairly consistent, nice code added and fugly hack removed.

doc/OPTIONS
```diff
@@ -394,29 +394,41 @@ their number could potentially be decreasing as some hashes get cracked.
...
+commas, and prepending a format, wildcard or class with a "-" will exclude
+instead of include (eg. "--format=#ipmi,-cpu" for again finding the RAKP
+format but this time rejecting the CPU version so will end up finding
+"RAKP-OpenCL" as long as your build supports OpenCL).  The order given of
+any "include" parts of a format list will be honored so this it feature can
+also be used for merely changing format precedence during auto-detection.
+Prepending a format, wildcard or class with a "+" will promote it to a
+requirement as opposed to an inclusion. That is, "--format=omp,+cpu"
+will first "include" any OMP formats but then prune all of them that does
+not match format class "CPU".
```
magnumripper: This wall of text should be broken in paragraphs, but the rest of the file isn't made like that

solardiz (approved):
I'm afraid I can't reasonably review this by merely looking at the patch. I'd need to extract pieces of code that's been moved and diff them separately. I'd also need to effectively repeat magnum's reasoning about this whole thing, then see whether/how mine differs. This is just too time-consuming. Given that nothing is expected to blow up even if there's something wrong in here, I'd rather save my time and trust magnum on this, and we'll test the feature after merging.

magnumripper:
Cool. Meanwhile I found minor bugs and inconsistencies and improved documentation by splitting to paragraphs, new PR coming.

PR #4372, opened 2020-09-27
Workaround kernels not found in PowerShell

ghost:
Workaround for OpenCL kernels not found when run from Windows PowerShell.
****

It works, but a broad testing would be interesting.

```
PS C:\> rm .\Temp\johntheripper\run\*.pot; rm Temp\\johntheripper\run\opencl\*.bin; Temp\johntheripper\run\john --format=sha512crypt-opencl tests.in
Device 1: AMD Ryzen 5 3500U with Radeon Vega Mobile Gfx
Loaded 1 password hash (sha512crypt-opencl, crypt(3) $6$ [SHA512 OpenCL])
Cost 1 (iteration count) is 5000 for all loaded hashes
Build log: Compilation started
Compilation done
Linking started
Linking done
Device build started
Device build done
Kernel <kernel_crypt_full> was successfully vectorized (8)
Kernel <kernel_crypt_fast> was successfully vectorized (8)
Done.
Binary Build log: Device build started
Device build done
Reload Program Binary Object.
LWS=128 GWS=384 (3 blocks)
Almost done: Processing the remaining buffered candidate passwords, if any.
Proceeding with wordlist:/run/password.lst
john             (?)
1g 0:00:00:06 DONE 2/3 (2020-09-26 22:45) 0.1660g/s 63.76p/s 63.76c/s 63.76C/s 123456..larry
Session completed.
```

### DIFF
--- a/src/opencl_common.c
+++ b/src/opencl_common.c
@@ -1268,6 +1268,26 @@ static char *include_source(const char *pathname, int sequential_id, const char 
 		full_path = quote_str(full_path);
 	}
 
+#ifdef __CYGWIN__
+	/*
+	 * Workaround for OpenCL kernels not found when run from Windows PowerShell.
+	 * CygWin (at least in PowerShell) maps the main JtR folder to "/".
+	 */
+	const char *sandbox_path;
+
+	sandbox_path = realpath(".", NULL);
+
+	if (sandbox_path && sandbox_path[0] == '/' && !strcmp(full_path, "/run/opencl")) {
+		int main_folder_fd;
+
+		if ((main_folder_fd = open("/run/opencl", O_DIRECTORY)) > 0) {
+			fchdir(main_folder_fd);
+			close(main_folder_fd);
+		}
+	}
+	MEM_FREE(sandbox_path);
+#endif
+
 	snprintf(include, LINE_BUFFER_SIZE,
 	         "-I %s %s %s%s%s%s%d %s%d %s -D_OPENCL_COMPILER %s",
 	        full_path,


src/opencl_common.c
```diff
@@ -1268,6 +1268,26 @@ static char *include_source(const char *pathname, int sequential_id, const char
...
+
+	sandbox_path = realpath(".", NULL);
+
+	if (sandbox_path && sandbox_path[0] == '/' && !strcmp(full_path, "/run/opencl")) {
+		int main_folder_fd;
+
+		if ((main_folder_fd = open("/run/opencl", O_DIRECTORY)) > 0) {
+			fchdir(main_folder_fd);
+			close(main_folder_fd);
+		}
```
solardiz, line 1286: This is just a weird way to `chdir()`. My suggestion to use an fd was for a different purpose - to save the previous working directory and then restore it.

src/opencl_common.c
```diff
@@ -1268,6 +1268,26 @@ static char *include_source(const char *pathname, int sequential_id, const char
...
+#ifdef __CYGWIN__
+	/*
+	 * Workaround for OpenCL kernels not found when run from Windows PowerShell.
+	 * CygWin (at least in PowerShell) maps the main JtR folder to "/".
+	 */
+	const char *sandbox_path;
+
+	sandbox_path = realpath(".", NULL);
+
+	if (sandbox_path && sandbox_path[0] == '/' && !strcmp(full_path, "/run/opencl")) {
```
solardiz, line 1280: Does the check for `sandbox_path && sandbox_path[0] == '/'` serve any purpose? Isn't `realpath()` supposed to always return a full pathname, thus starting with a `/`?

src/opencl_common.c
```diff
@@ -1268,6 +1268,26 @@ static char *include_source(const char *pathname, int sequential_id, const char
...
+			fchdir(main_folder_fd);
+			close(main_folder_fd);
+		}
+	}
+	MEM_FREE(sandbox_path);
+#endif
+
 	snprintf(include, LINE_BUFFER_SIZE,
 	         "-I %s %s %s%s%s%s%d %s%d %s -D_OPENCL_COMPILER %s",
 	        full_path,
```
solardiz, line 1293: I find it weird that even after `chdir()` to the kernels directory we still pass `/run/opencl` as the include path. My guess was that the OpenCL backend is unaware of Cygwin's mappings, which is why it failed for include paths, but then it'd also be failing for `/run/opencl` even when we're in that directory. I expected we'd need to `chdir("/run");` and pass `-I opencl` here, or when we use `chdir("/run/opencl");` then pass `-I .` here.

solardiz (changes requested):
Please comment on my comments, but at least the `fchdir()` misuse should be gone (and we should add proper use for the intended purpose, as I described).

src/opencl_common.c
ghost: Well, if `/doc` is a full pathname, then, yes.

```
Z:\debug\doc>cd
Z:\debug\doc
```
```
..\run\john --test=0 --format=sha512crypt-opencl
Device 1: AMD Ryzen 5 3500U with Radeon Vega Mobile Gfx
Testing: sha512crypt-opencl, crypt(3) $6$ (rounds=5000) [SHA512 OpenCL]... 
sandbox_path: /doc => ../run/opencl
```
My intention was to skip sane environments, if any.

src/opencl_common.c
ghost: In that respect, the patch does nothing. It is the legacy code (which applies to other platforms and in the CMD). I don't need to add a `-I`, so I don't.

src/opencl_common.c
ghost: Anything that returns `C:\...`. Or `\`.

solardiz: So in some cases `realpath()` returns a non-absolute pathname, and in those cases we don't need the new hack, correct? However, that behavior of `realpath()` is against POSIX, which says "The realpath() function shall derive [...] an absolute pathname that names the same file, whose resolution does not involve '.', '..', or symbolic links."

Anyway, I'm OK with that.

src/opencl_common.c
solardiz: Claudio, have you tried using simply `chdir()` here? While it's supposed to work the same as above, I admit I'm not 100% sure both have them same magic in them.

src/opencl_common.c
solardiz: I never implied we needed to _add_ a `-I`; I thought that we'd need to use a different argument to the existing `-I` after having used the `chdir()` trick.

src/opencl_common.c
ghost: I agree it is useless as I see it now (I never saw something without a `/`. A full path is `/cygdrive/z/...`.

src/opencl_common.c
ghost: The original idea:
```
fd = open(".", O_DIRECTORY);
chdir(/run/opencl);
do-stuff();
fchdir(fd);
```

But, what is `.`? What is `/`? The main problem in my patch it that I do not trust these "concepts" inside CygWin.

I do **open** the directory "/run/opencl", so if I succeed, I do something. Otherwise, do nothing but pray.

In my mind, that is safer. Probably, chdir() will do the same thing, but is it going to open the directory before do something else? But wait, do we care? Is that important?

solardiz: I'm now experimenting with what might be a generic solution on Linux. In my testing, not chdir'ing back to the previous current directory results in other path_expand'ed pathnames being wrong, so e.g. there's issue caching the binary kernel. I wonder why you didn't(?) run into such issues with your patch on Windows.

solardiz: BTW, there's also this comment:

```c
        // include source is thread safe.
        build_opts = include_source("$JOHN/opencl", sequential_id, opts);
```

My idea of saving/restoring the cwd might not be thread-safe. As I recall, we experimented with building kernels from an OpenMP parallel region for descrypt-opencl, but disabled it since?

PR #4373, opened 2020-09-27
OpenCL: Build kernels having chdir'ed to John's home directory

### LINKED ISSUE #4335: OpenCL kernels not found when run from Windows PowerShell
As discussed on john-users:

https://www.openwall.com/lists/john-users/2020/09/03/6 PowerShell fails unless in the kernels directory
https://www.openwall.com/lists/john-users/2020/09/03/14 Cygwin shell works

solardiz:
This works for me on Linux. I haven't actually tested on Windows PowerShell yet, nor on any Windows, nor anywhere else at all for that matter. @claudioandre-br I'd appreciate your testing. Thanks!

The description in the commit message and below just assumes that this works across platforms, but I don't know that yet.

This lets us use simply "-I opencl" instead of having to resolve a pathname, which might contain spaces (which we had to quote) and was problematic on Cygwin when run from Windows PowerShell (apparently, with Cygwin resolving pathnames differently than the OpenCL backend would for the includes).

Fixes #4335

### DIFF
--- a/src/opencl_common.c
+++ b/src/opencl_common.c
@@ -60,12 +60,6 @@
 
 #define LOG_SIZE 1024*16
 
-#if !defined(__CYGWIN__) && !defined(__MINGW32__)
-// If true, use realpath(3) for translating eg. "-I./opencl" into an absolute
-// path before submitting as JIT compile option to OpenCL.
-#define I_REALPATH 1
-#endif
-
 // If we are a release build, only output OpenCL build log if
 // there was a fatal error (or --verbosity was increased).
 #ifdef JTR_RELEASE_BUILD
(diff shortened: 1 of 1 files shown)


solardiz:
Apparently, saving/restoring the current directory via an fd won't work on MinGW. Maybe also not on older Windows even under Cygwin. I think we'll need to add fallback code that uses `getcwd()` and `chdir()`.

Do we currently support builds with MinGW, or is that stale anyway?

Regardless, I'd like to see some testing of this PR's branch in its current state.

ghost:
The only failure I saw was:

```
C:\Temp\generico\run>john --test=0 --format=sha512crypt-opencl
Device 1: AMD Ryzen 5 3500U with Radeon Vega Mobile Gfx
Testing: sha512crypt-opencl, crypt(3) $6$ (rounds=5000) [SHA512 OpenCL]... chdir: : No such file or directory
```

magnumripper:
I wouldn't know really, but I thought MinGW builds could had some merit over Cygwin depending on situation/needs. I have no issue with dropping support for it if it gives us headache.

magnumripper (approved):
The `PARALLEL_BUILD` feature ended up as a big disappointment, we never saw any good of it (I was hoping for differences like when building host code, eg. `make -j32`). It may have changed though (no-one probably tested it for years), or could change in the future.

solardiz:
> chdir: : No such file or directory

In other words, we should skip the `chdir()` when the path is empty (which is the case when running from the current directory on Windows and without a `./`, which is optional there). IIRC, we already disallow invocation via `PATH` of non-systemwide installs on non-WIndows. We'll need to also look into what happens for systemwide installs - have you?

> MinGW builds could had some merit over Cygwin depending on situation/needs. I have no issue with dropping support for it if it gives us headache.

I think it would be good to (continue to) support MinGW and it might not give us headache, but without us testing that it's bound to be broken most of the time. Is any one of us testing this? When did we test it last?

> The `PARALLEL_BUILD` feature ended up as a big disappointment, we never saw any good of it

I don't really recall, did we bump into some kind of locks in the OpenCL backends?

BTW, right now `PARALLEL_BUILD` is never defined, nor undefined. Perhaps we should explicitly `#undef` it.

solardiz:
> In other words, we should skip the `chdir()` when the path is empty (which is the case when running from the current directory on Windows and without a `./`, which is optional there).

I wonder whether it would expose the mystery "PowerShell problem" again, and if so whether `chdir(".");` would work around it.

solardiz:
I've just pushed changes skipping the `chdir()` when it is obviously unneeded, and implementing runtime fallbacks to `getcwd()` and `chdir()` for the saving/restoring. We might also need a compile-time fallback if any platform we support fails to even compile code with `fchdir()` or makes it appear to work (per return values) yet doesn't implement it correctly. For now, I think let's try with the runtime fallbacks only. I also had an fd leak there (missed by magnum's review), now fixed.

Claudio, please retest this on Windows. magnum, I'd appreciate it if you take another look. Thanks!

ghost:
Rebuilding it now at https://ci.appveyor.com/project/claudioandre-br/johntheripper/builds/35453915.

ghost:
Passes
```
cd \;                         del \temp\generico\run\opencl\*.bin ; temp\generico\run\john --test=0 --format=sha512crypt-opencl
cd \;                         del \temp\generico\run\opencl\*.bin ; \temp\generico\run\john --test=0 --format=sha512crypt-opencl

cd \Temp;                     del \temp\generico\run\opencl\*.bin ; generico\run\john --test=0 --format=sha512crypt-opencl
cd \Temp\generico;            del \temp\generico\run\opencl\*.bin ; run\john --test=0 --format=sha512crypt-opencl

cd \Temp\generico\doc;        del \temp\generico\run\opencl\*.bin ; ..\run\john --test=0 --format=sha512crypt-opencl
cd \Temp\generico\run;        del \temp\generico\run\opencl\*.bin ; .\john --test=0 --format=sha512crypt-opencl

cd \Temp\generico\run\opencl; del \temp\generico\run\opencl\*.bin ; ..\john --test=0 --format=sha512crypt-opencl
cd \Temp\generico\run;        del \temp\generico\run\opencl\*.bin ; \Temp\generico\run\john --test=0 --format=sha512crypt-opencl
----

cd \ &&                          del \temp\generico\run\opencl\*.bin  &&  temp\generico\run\john --test=0 --format=sha512crypt-opencl
cd \ &&                          del \temp\generico\run\opencl\*.bin  &&  \temp\generico\run\john --test=0 --format=sha512crypt-opencl

cd \Temp &&                      del \temp\generico\run\opencl\*.bin  &&  generico\run\john --test=0 --format=sha512crypt-opencl
cd \Temp\generico &&             del \temp\generico\run\opencl\*.bin  &&  run\john --test=0 --format=sha512crypt-opencl

cd \Temp\generico\doc &&         del \temp\generico\run\opencl\*.bin  &&  ..\run\john --test=0 --format=sha512crypt-opencl
cd \Temp\generico\run &&         del \temp\generico\run\opencl\*.bin  &&  .\john --test=0 --format=sha512crypt-opencl
cd \Temp\generico\run &&         del \temp\generico\run\opencl\*.bin  &&  john --test=0 --format=sha512crypt-opencl

cd \Temp\generico\run\opencl &&  del \temp\generico\run\opencl\*.bin  &&  ..\john --test=0 --format=sha512crypt-opencl
cd \Temp\generico\run &&         del \temp\generico\run\opencl\*.bin  &&  \Temp\generico\run\john --test=0 --format=sha512crypt-opencl

```

solardiz:
Thanks, Claudio! I hope this also works for you with multiple formats benchmarked per run?

Issue #2013, opened 2016-01-29
n-Viterbi


magnumripper:
http://www.openwall.com/lists/john-dev/2015/10/16/3

_n-Viterbi is an extension of the Viterbi algorithm:_

https://en.wikipedia.org/wiki/Viterbi_algorithm

_As far as I'm aware, this extension was first introduced in "Timing Analysis of Keystrokes and Timing Attacks on SSH" by Dawn Song et al:_

http://users.ece.cmu.edu/~dawnsong/papers/ssh-timing.pdf

_As I tweeted a couple of years ago:_

_@dawnsongtweets How about finally releasing the n-Viterbi code for your "Timing Analysis of Keystrokes and Timing Attacks on SSH"?_
_@dawnsongtweets As discussed in 2001, I am not convinced n-Viterbi gives all answers and exactly in decreasing order of probabilities_
_@dawnsongtweets However, if n-Viterbi actually does that and isn't more complex than what the paper says, it has many other applications_
_I did try (re)implementing n-Viterbi from the English description at the time. It'd skip some answers and the order would be suboptimal._

_(I guess I should have pinged Dawn Song via e-mail again, as she didn't proceed to tweet much.  I think she just got on Twitter at the time.)_

_Good news: @mitar_m has just found what looks like an implementation of n-Viterbi, here:_

https://github.com/edechter/PRISM/blob/master/src/c/up/viterbi.c

_it's the compute_n_max() function, as called from:_

https://github.com/edechter/PRISM/blob/master/src/c/up/mcmc_predict.c

``` c
    /* Compute the candidates Es for reranking by n-Viterbi */
    compute_n_max(n);
```

_Other source files/functions in that directory also look relevant (at first glance; to be confirmed)._

_Is anyone willing to give this a try, for a potential new cracking mode for JtR?  Potentially, this would replace incremental or Markov modes' suboptimal ordering with the most optimal ordering that can be inferred from the available statistics (overfitting is a risk, though). My experiments in 2001 didn't result in an algorithm that would behave as well as the paper claimed it should (Viterbi finds one most probable candidate easily, but then it's tricky for the 2nd most probable and on), but maybe it was just me, and maybe the implementation above is actually it.  I think this is worth a try.  Unfortunately, I don't have time for it right now._

solardiz:
Maybe more and simpler implementations:

https://github.com/carthach/kBestViterbi (Python)
https://github.com/meerkat-cv/top_n_viterbi (C++)

[...]

PR #4378, opened 2020-09-29
Due 'rain' as first intended

e2002e:
This is a reworked, faster and truly delinearized generation.


magnumripper:
I respect your perseverance but this won't be merged until you prove that it *actually* finds candidates faster than dumb brute-force in a test case.

e2002e:
Hello !
Here are two small videos to prove that the mangling is far from being a repetitive/dumb cycle:
https://youtu.be/zzSx8OLIXVU
https://youtu.be/-lw2s44hGzI

And this one for the sake of demonstration, using it on a list of hashes with an hexadecimal set. At the end you can see a run of 2^20 words output to console and fathom the root of the algorithm.
https://youtu.be/m5tbdGCuaZU

I would really like to make it benefit from opencl and run it on a list of uncracked hashes.
But I'm having some problems to understand how you do it when having a look a mask.c and mask_ext.c

magnumripper:
> Here are two small videos to prove that the mangling is far from being a repetitive/dumb cycle:

Youtube is for millenials, we need it described properly. I do understand it *looks* much better because you never see the current process as `aaaaaaaahj`. I really get that point. It's just that statistically *it doesn't actually buy you anything* so you need to elaborate on how/why this would crack **any** given set of hashes any faster than any of our existing modes.

If your mode would be powerful for certain situations only (such as finding truly random hashes quicker than our incremental mode), that's totally fine - but you need to prove it with a test case (and you need to supply all info needed for us to reproduce it - links to testing datasets or how to create them, and so on).

> I would really like to make it benefit from opencl and run it on a list of uncracked hashes.
> But I'm having some problems to understand how you do it when having a look a mask.c and mask_ext.c

From a quick look you seem to have implemented it correctly for already being usable with internal masks on GPU. Try this:
```
./john hashes -form:nt-opencl -rain -mask:?w?a?a
```
That will use your mode for creating base words and then append them with two characters generate by GPU-side mask. Don't worry about those last two being dumb BF - they'll make 25 Gp/s on a 2080ti.

Here's a quick test comparing our incremental mode with your mode. Input is 5.294 NT hashes, scraped from pastebin. They range from stupidly simple to uncrackable. I give each mode 60 seconds to find passwords.

Your proposed mode:
```
$ ../run/john scraped.sam -form:nt-opencl -rain -mask:?w?a?a -max-run=60
Device 1: GeForce RTX 2080 Ti
Loaded 5294 password hashes with no different salts (NT-opencl [MD4 OpenCL])
LWS=256 GWS=17408 (68 blocks) x9025 
0g 0:00:00:01 N/A 0g/s 0p/s 0c/s 0C/s Dev#1:48°C
omai             (STUMPL29:4915)
scan             (scan:3651)
das              (SINNP29:5285)
sap              (SAPUPDATE:1256)
Max              (POBATSCHNIGL01:4369)
(...)
andrea           (Schauer:1659)
3wpl6x           (FunParkPBG:18691)
flexxi           (dzlkols01adm:12155)
akethi           (RIEDELA30:3208)
daewoo           (xdnc:5581)
593g 0:00:01:01 N/A 9.572g/s 20944Mp/s 20944Mc/s 99221GC/s Dev#1:73°C z!d$Yaa..AbH1`2ajtEQaa
Remaining 4701 password hashes with no different salts
Warning: passwords printed above might not be all those cracked
Use the "--show --format=NT-opencl" options to display all of the cracked passwords reliably
Session stopped (max run-time reached)
```

Our incremental mode:
```
$ rm -f ../run/john.pot && ../run/john scraped.sam -form:nt-opencl -inc -mask:?w?a?a -max-run=60
Device 1: GeForce RTX 2080 Ti
Loaded 5294 password hashes with no different salts (NT-opencl [MD4 OpenCL])
LWS=256 GWS=34816 (136 blocks) x9025 
stephan          (hagelauer:6446)
mee.             (LOESCHELL04:2704)
09091211         (S:1001)
steff14          (Poley:3933)
melanie          (Hellmann:8245)
(...)
c666o            (oppel:9203)
video            (Surveillance:2530)
vroni            (BITTNERK06:5071)
vamp7            (Henninger:11899)
verxs            (FLIERLERL22:4197)
2234g 0:00:01:01  36.14g/s 25214Mp/s 25214Mc/s 81019GC/s Dev#1:74°C knmad2saa..knmad2saa
Remaining 3060 password hashes with no different salts
Warning: passwords printed above might not be all those cracked
Use the "--show --format=NT-opencl" options to display all of the cracked passwords reliably
Session stopped (max run-time reached)
```
For some reason, your mode is slower. I believe it shouldn't have to be and I might be able to help you with it if I find time and incentive.

For the sake of it, let's compare with subsets as well, even though it's *not* intended as a go-to cracking mode but to find semi-weak passwords before or after running a proper mode:
```
$ rm -f ../run/john.pot && ../run/john scraped.sam -form:nt-opencl -subsets -mask:?w?a?a -max-run=60
Device 1: GeForce RTX 2080 Ti
Loaded 5294 password hashes with no different salts (NT-opencl [MD4 OpenCL])
LWS=256 GWS=34816 (136 blocks) x9025 
0g 0:00:00:01 2.85% (3) (ETA: 23:59:05) 0g/s 0p/s 0c/s 0C/s Dev#1:35°C
omai             (STUMPL29:4915)
scan             (scan:3651)
das              (SINNP29:5285)
adam             (siler:8701)
blau             (KRAUSSR28:2576)
(...)
17111994         (HADERM17#:2103)
18111993         (LECHNERA18#:2647)
18111998         (KRAFTM18:4775)
18111995         (SCHOLLWECKM18#:3451)
18111994         (HAMBERGERD18#:2123)
783g 0:00:01:01 6.19% (4) (ETA: 00:14:54) 12.63g/s 25976Mp/s 25976Mc/s 120297GC/s Dev#1:69°C b11ffbaa..b11ffbaa
Remaining 4511 password hashes with no different salts
Warning: passwords printed above might not be all those cracked
Use the "--show --format=NT-opencl" options to display all of the cracked passwords reliably
Session stopped (max run-time reached)
```

So your mode cracks 11.2% of them in a minute. Incremental mode cracks 42.2% of them in the same time. Even Subsets mode beats your mode, it cracks 13.9% of them.

e2002e:
Now that you gave me the solution for using gpu, I'm going to try with a list of uncracked hashes.
I believe that this is normal that this mode cracks less keys than the others in a run of sixty seconds, for the basic reason that incremental and subset have in common that -at the level of their algorithm- they try the most plausible passwords first.
I'm not sure about the speed issue, the words are processed fully on each generation, unlike mask that will work only on the placeholders that need changes, but I don't know how you are dealing with that in subsets and incremental.

magnumripper:
> Now that you gave me the solution for using gpu, I'm going to try with a list of uncracked hashes.

Just be sure to match the mask with the character set: If you run with lowercase hexadecimal only, you need to use eg. `-mask=?w?h?h` although that may starve the GPU if it's a good one (?h?h give you a mask multiplier of just 256) so you might need `-mask=?w?h?h?h` for a multiplier of 4096, or maybe even `-mask=?w?h?h?h?h` for a multiplier of 64K if you have a flagship GPU.

> I believe that this is normal that this mode cracks less keys than the others in a run of sixty seconds, for the basic reason that incremental and subset have in common that -at the level of their algorithm- they try the most plausible passwords first.

True, but our existing Markov mode explicitly does *not* produce best candidates first, and here's the same test with it:
```
$ rm -f ../run/john.pot && ../run/john scraped.sam -form:nt-opencl -markov -mask:?w?a?a -max-run=60 
Device 1: GeForce RTX 2080 Ti
Loaded 5294 password hashes with no different salts (NT-opencl [MD4 OpenCL])
LWS=256 GWS=69632 (272 blocks) x9025 
MKV start (stats=$JOHN/stats, lvl=200 len=10 pwd=259355623)
chakuza          (PATSCHS06:3010)
chemie           (DRM:1164)
check            (distler:6273)
chinito          (BUERKJ08:1642)
chianty7+        (Boymanns:1710)
(...)
rh4658?          (PCruickshank:18627)
rss              (newsreader:5392)
rtl              (PETRENZM14:4359)
nelly18.5        (SCHRAMMT01:3454)
nelly            (GUNS01:2076)
1008g 0:00:01:02 70.80% (ETA: 00:32:28) 16.23g/s 26728Mp/s 26728Mc/s 128041GC/s Dev#1:68°C neno72aa..neno72aa
Remaining 4286 password hashes with no different salts
Warning: passwords printed above might not be all those cracked
Use the "--show --format=NT-opencl" options to display all of the cracked passwords reliably
Session stopped (max run-time reached)
```
That's 19% compared to your 11.2%. Also, Subsets does not produce best candidates first, just the very weakest.

e2002e:
Very fine, Markov must be powerfull enough. I still think that in some cases where you are looking hard for one key (a man posted an electrum hash on reddit, account on which he had about 100k $ and for which he lost the pass) it would be worth a try before giving up.
I'm letting it run for the night on a list of uncracked hashes.
PS: we should run all the modes for a given time and not clean the pot inbetween, so the last run with rain would show if it's really interesting.

magnumripper:
> I still think that in some cases where you are looking hard for one key (a man posted an electrum hash on reddit, account on which he had about 100k $ and for which he lost the pass) it would be worth a try before giving up.

Sure, just convince us *why* it would be worth a try.

> PS: we should run all the modes for a given time and not clean the pot inbetween, so the last run with rain would show if it's really interesting.

That's up to you. You'll need to come up with a test case that we can reproduce and consider.

Considering your energy I'd really love if you considered implementing #2013 instead! Now *that* would be business. You'll end up on reddit.

e2002e:
I ran:
../run/john -inc=Alnum -min-len=6 -max-len=15 -format=raw-MD5-opencl ../run/testhashes -mask=?w?1?1 -1=?l?u?d -max-run=100
Then:
../run/john -rain=abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789 -min-len=6 -max-len=15 -format=raw-MD5-opencl ../run/testhashes -mask=?w?1?1 -1=?l?u?d -max-run=100
It found more than a hundred hashes, and note that almost all of them were of length 6 (how are lengths 6 and 7 being treated btw, and so even with inc mode ?).

I read a little about Viterbi on wikipedia, I will first try to contact a friend of mine who is a math doctor, if he can guide me with this one I'll do it.

solardiz:
@e2002e In your latest test with 100 seconds incremental + 100 seconds rain, please add a comparison to continuing incremental for another 100 seconds. So instead of the rain, you'd run "--restore". Would incremental or rain crack more passwords in the second 100 seconds? Specifically, compare these two:

```
cd ../run
rm john.pot
./john -inc=Alnum -min-len=6 -max-len=15 -format=raw-MD5-opencl testhashes -mask=?w?1?1 -1=?l?u?d -max-run=100
./john -rest
```

vs.

```
cd ../run
rm john.pot
./john -inc=Alnum -min-len=6 -max-len=15 -format=raw-MD5-opencl testhashes -mask=?w?1?1 -1=?l?u?d -max-run=100
./john -rain=abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789 -min-len=6 -max-len=15 -format=raw-MD5-opencl ../run/testhashes -mask=?w?1?1 -1=?l?u?d -max-run=100
```

You can also save and reuse `john.pot` and `john.rec` from after the first `john` run (first 100 seconds of incremental), so that you don't need to repeat it and so that you remove the variability associated with it possibly running slightly differently when repeated.

e2002e:
@solardiz, after two runs of 100 seconds with incremental, launching with -rain still finds a bit less than two hundred passwords in it's 100 allocated seconds, most of them being of length 6 and being either a repeating pattern like xyzxyz or wwwwww or a combination of lower/upper case characters and digits like t0mc10, Nika66, or the classic q1w2e3. There are more all upper case than what I expected. Most of the lower case have been cracked by incremental.

solardiz:
@e2002e That's not the test I asked you to run.

e2002e:
From the correct tests the results are the same (with a copy of john.pot)
About 200 keys are found on the second run with -rain. Much more with inc.

solardiz:
> About 200 keys are found on the second run with -rain. Much more with inc.

That's what we all expected, right? So how do we justify `-rain`, then? You say it's a last resort, but per this test it appears that it's more efficient to just continue with incremental as that last resort instead.

e2002e:
Of course it is. I had presented this from the start as a hardcore password cracking method, we ran it 3 minutes and compared it with incremental mode on a file that certainly is riddled with passwords that your mode make a fiest upon.
I'll let it run tonight on uncracked hashes, I think that finding one is the only way I can change your minds.

solardiz:
3 minutes against raw-MD5 on GPU is like 3 months against a modern cryptocurrency wallet (to use your example), so it's not that unreasonable.

Finding one uncracked hash wouldn't be convincing: not a comparison against another mode and not statistically significant. You need to show that your mode finds more uncracked hashes than incremental does in the same time. Say, if you run incremental for 24 hours and it cracks 50 hashes, and you separately run rain (from the same starting conditions - hash list, `john.pot`) for 24 hours and it cracks 100 hashes, you'd have a point.

solardiz:
BTW, I find it quite realistic that you'd actually crack more uncracked hashes with rain than with incremental, if others had run incremental on the original (fuller) hash list, leaving only hashes not quickly crackable with incremental in there. I suspect that even more efficient than rain in that case would be starting incremental from some later point (something we don't support officially, although a `john.rec` hack will do it) or using Markov mode with some unusual/high settings. Using something else that is somewhat weird e.g. `-inc -mask='?a?w?a?a'` might also work well.

Rain will lose its weirdness advantage in such cases if it becomes a standard mode that people run. To compensate for that, you also need some kind of user-specified seed to alter the sequence (minimizing overlap with other seeds' initial parts of sequences).

e2002e:
I take note of what you'r saying about testings.
I'm not familliar enough with markov or incremental to really understand where these are going to in terms of generation tree, I'll report the results tomorrow (i'm at greenwich +2)

e2002e:
I have let it run for about 10 hours with alnum set and length from 12 to 17 but no hash from the uncracked list came up...
I'm trying once more with hexadecimals, I think this character set is the most appropriate to use with option -rain.

solardiz:
> length from 12 to 17

BTW, we limit incremental mode to lengths up to 13 (inclusive) by default, as that's where it's most efficient in our testing. (Except on all-digit strings, where the default length limit is 20.) So if you want to try something unusual (which others probably didn't try yet against these hashes) yet reasonable, you can set incremental mode's lengths range e.g. to 14 to 17, and you might crack a password that others' incremental mode runs didn't try.

In fact, if you set it to 12 to 17, it will give so much higher weight to lengths 12 and 13 than it did in other people's runs that it might exceed the portion of those lengths' keyspaces that they searched soon (e.g., in some hours or days or months?) and might crack extra passwords. Moreover, careful analysis of this information could tell you how far others got specifically with our incremental against those hashes, so then you could run a non-length-limited incremental mode attack (allowing low lengths as well) hacked to start at the point to which others got.

Edit: The above applies to usage of incremental mode on its own, without adding a mask. Incremental+mask is possibly already unusual enough that few people tried precisely it (e.g., which specific combination of incremental with which mask).

This is almost the opposite from what we usually recommend, but I understand you're trying to be different than others and see whether these weird approaches have merit against hashes left uncracked by others where you do not readily know which attacks and how far those other people ran.

e2002e:
@solardiz, in your message from two days ago:
> you also need some kind of user-specified seed to alter the sequence
The sequence depends on the ordering of characters in the set. the other way to change it is having different starting value for the variable that modifies the state when picking in the set.
Is is doing as much different sequences as there are characters in the set, no more, even with a `srand(time(NULL))`.
 
> In fact, if you set it to 12 to 17, it will give so much higher weight to lengths 12 and 13 than it did in other people's runs that it might exceed the portion of those lengths' keyspaces that they searched soon 

Ok, I there's one notable point with rain about this it is that it generates the words by looping on maxlength-minlength+1; it will never skip words of higher length.
But in the absolute it has nothing to do with the strafing rotation used with the selection of characters in the set, that I call rain.

Note that the generation is symetrical, with length 6 and a set of '01' we can see that 111000 and 000111 are in the position 32 - 33, and 000110 is the first word while 111001 is the last.

e2002e:
Hi there, I must admit my utter disappointment: the patterns are still repetitive. Not suitable for bruteforcing. It has some interest in some other domains, or maybe reworked. I'll keep trying to do something about it. See ya !

e2002e:
I'm reopening now that everything seems to be fixed:
The rotation is used along the strafe in:
`charset_utf32[(charset_idx[loop][(i+strafe[loop]) % (mpl-2)+2] + rotate[loop]) % charcount]`
This is the case where words are even, and we exclude the two first characters that are anyway always moving because the state evolves from left to right.
when the words are odd, we just skip one character.
And the increment is not so weird, it's just `rotate[loop]++` and `strafe[loop]++` once we have generated a word.
I am going to proceed to some more tests but I already done so against subsets and, at the same speed and running for 60 seconds, rain found 10 less passwords...

e2002e:
I once more had to modify the code, as showed in the previous message, **even** lengths have to 'dodge' the first two placeholders of the word, and **odd** lengths have to do so for three of them, that's the strange thing. I did also put the increase for the 'rotate' variable (used to modify the selection of the character in the set) inside of the iteration on the length of the word. This decreased entropy, instead of having words using the strafed index and completing with the next character from the set on each words, it does change character at each iteration on a placeholder, when they are not 'dodged'. 

Here are the results of the 60 " runs comparing rain and inc mode with min-len=7 & max-len=15
The format is raw-md5-opencl and both modes use the same mask i.e ?w?1?1?1
The file contains 19092 hashes.

With rain, the tests have not finishing to test length 4 (in lowercase), because they put as much weight on other lengths.
For the sake of honesty running the same comparison with a `min-len` set to 7:
Alnum: 
Rain finds 20 keys - 0.10% - from which 16 are found when running rain on top of inc.
Inc finds 1071 keys - 5.6%
Lowernum;
Rain finds 75 keys - 0.39% - from which 61 are found etc ...
Inc finds 2074 keys which is 10.86%
Lower:
Rain finds 761 keys - 3.98% - from which 598 are found etc ...
Inc finds 1280 keys - 6.7%
Digits:
Rain finds 354 keys - 1.85%
Inc finds 361 keys - 1.89%
In this last case, no other password was found by running rain after inc, but length 7, 8, and 9 are exhausted in 1 minute...

There is an entropy weakness with big keyspaces compared to inc, but it can me managed by a wider range of characters treated by the mask, because rain's state evolves from left to right, it will keep the bruteforce dynamic:
With Alnum and a mask of ?w?1?1?1?1, the first characters would still loop fast enough with heavy formats, and the result then with the same settings as above are:
93 words - 0.48% - from which 83 are found etc ...

When pushing the **mask's range to 5**, the first character's entropy would become hard to overcome for the heaviest format, but it is still acceptable for many formats; going at 2G/k seconds, it loops around an alphanum character set in **5 minutes**, run which results in:
Rain finds 521 keys - 2.72% - from which 145 are found etc ...
Inc finds 2092 keys - 10.95% We haven't changed the mask of ?w?1?1?1 for inc since the start.
the found passwords are different, hence my belief for the use of this mode as being complementary as a last resort.

A last test with Lowernum character set and a mask of range of 5 (which is acceptable for heavier formats, i.e it would not leave a static character more than day with a speed about 1K/k sec, independently of the length of the words. But the longer the words, the more words with increasing size chunks of chained characters like ABCD) :
Rain finds 757 keys - 3.96% - from which 466 are found etc ...
Inc finds 1175 keys - 6.15%

The result of these tests show that 'rain' must be tuned with the mask according to the speed of the crack to be able to perform well, and does give a chance of an "other" non-linear bruteforce then.

e2002e:
In fact, as it is now, the algorithm is good, I will play with this to see up to where it's going.
I have reuploaded my files on the repo 'zhou'.
I now realized that there was no need for all this requests of mine, the -stdin switch works very good with the mask, and my software can be restored (wip).
I understand the dubitatives but also enthousiastic reactions to this in the end.
Leaving it as is is the best way, thanks for doing a software that supports this kind of projects. 
And remember, the method is a bit complementary to -inc, so don't hesitate come and pull zhou.

e2002e:
Don't mind the previous tests, this works more simply, but is what I wanted to do.
Check the code or ask me if needed.

magnumripper:
e8fd673e6 contains an unresolved merge conflict

magnumripper:
No wait, it *fixed* an unresolved merge conflict. Never mind me 😄

e2002e:
haha

PR #4384, opened 2020-09-29
Late minor fixes for --format=LIST

magnumripper:
- Move all handling of [Disabled:Formats] to our new functions.
- Drop obsolete composite group dynamic-all. New syntax would be, for
  example, +dynamic,all.
- Do not complain about an inclusion not matching if an earlier inclusion
  added a superset (eg. --format=raw*,raw-md5).
- Update doc/OPTIONS and split it into paragraphs.

Still not breaking `make generic` ;-)

### DIFF
--- a/doc/OPTIONS
+++ b/doc/OPTIONS
@@ -394,37 +394,52 @@ their number could potentially be decreasing as some hashes get cracked.
 To have the cracked hashes (and possibly salts) removed from all
 processes, you may interrupt and restore the session once in a while.
 
+
 --format=NAME[,NAME...]		force hash type NAME
 
-Override the hash type auto-detection.  You can use this option when
-you're starting a cracking session or along with things like: "--test",
-"--show", "-list" or "--make-charset".  Note that John can't crack hashes
-of different types at the same time.  If you happen to get a password file
-that uses more than one hash type, then you have to invoke John once for
-each hash type and you need to use this option to make John crack hashes of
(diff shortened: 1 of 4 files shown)


doc/OPTIONS
```diff
@@ -394,37 +394,50 @@ their number could potentially be decreasing as some hashes get cracked.
 To have the cracked hashes (and possibly salts) removed from all
 processes, you may interrupt and restore the session once in a while.
 
+
 --format=NAME[,NAME...]		force hash type NAME
```
magnumripper, line 398: As the section for --format is so large, I split it into paragraphs. For separating *that* from the rest, I added empty lines (so they are now double empty lines) above and below the whole section.

solardiz: Perhaps we should have empty lines between descriptions of _every_ option, then - for consistency.

magnumripper: Yes, and actually we should review the whole file in case more sections should be split the same way. Also, the used line width varies a little, between below 70 to at least a full 80, which is not so good. Oh, and we're not consistently using double space after period.

That'd be a separate task though.

ghost (approved):
Do you need an approval for merging?

magnumripper:
> Do you need an approval for merging?

Yes, thanks. I'll merge this as it fixes new bugs and not likely introduces new ones

Issue #4388, opened 2020-09-30
bcrypt problem with $2a vs. $2b handling


magnumripper:
I opened https://github.com/openwall/john-tests/issues/88 and started to investigate. It's actually a format quirk, not a TS shortcoming, and I realized I've scratched my head over it before. I'm pretty sure it's reproducible with John Proper. Perhaps this problem is not likely to surface IRL though. Like I said, it's about `$2a` vs. `$2b` (I had a faint memory of looking into it long ago).

Here's the full input file (TS assembled it from format's test vectors along with a corresponding wordlist)
```
$ cat tst-.in 
vec_0:$2a$05$CCCCCCCCCCCCCCCCCCCCC.E5YPO9kmyuRGyh0XouQYb4YMJKvyOeW
vec_1:$2a$05$CCCCCCCCCCCCCCCCCCCCC.VGOzA784oUp/Z0DY336zx7pLYAy0lwK
vec_2:$2a$05$XXXXXXXXXXXXXXXXXXXXXOAcXxm9kjPGEMsLznoKqmqw7tc8WCx4a
vec_3:$2a$05$CCCCCCCCCCCCCCCCCCCCC.7uG0VCzI2bS7j6ymqJi9CdcdxiRTWNy
vec_4:$2a$05$abcdefghijklmnopqrstuu5s2v8.iXieOjg/.AySBTTZIIVFJeBui
vec_5:$2x$05$/OK.fbVrR/bpIqNJ5ianF.CE5elHaaO4EbggVDjb8P19RukzXSM3e
vec_6:$2y$05$/OK.fbVrR/bpIqNJ5ianF.Sa7shbm4.OzKpvFnX1pQLmQW96oUlCq
vec_7:$2x$05$6bNw2HLQYeqHYyBfLMsv/OiwqTymGIGzFsA4hOTWebfehXHNprcAS
vec_8:$2x$05$6bNw2HLQYeqHYyBfLMsv/O9LIGgn8OMzuDoHfof8AQimSGfcSWxnS
vec_9:$2a$05$/OK.fbVrR/bpIqNJ5ianF.swQOIzjOiJ9GHEPuhEkvqrUyvWhEMx6
vec_10:$2a$05$/OK.fbVrR/bpIqNJ5ianF.R9xrDjiycxMbQE2bp.vgqlYpW5wx2yy
[...]

PR #4389, opened 2020-09-30
john.c: Protect from segfaulting from division by zero

magnumripper:
Seen occasionally during bcrypt testing.  Apparently salt_count was zero while password_count was not.  I was using non-standard local config options `ShowSaltProgress = Y` and `ShowRemainOnStatus = Y`, that probably helped triggering the bug by calling `john_loaded_counts()` at a time database is not yet consistent.

See #4388 though - perhaps the `vec3` double reporting is involved? Anyway, this commit adds safety.

### DIFF
--- a/src/john.c
+++ b/src/john.c
@@ -703,7 +703,7 @@ char *john_loaded_counts(struct db_main *db, char *prelude)
 	static char buf[128];
 	char nbuf[24];
 
-	if (db->password_count == 0)
+	if (db->password_count == 0 || db->salt_count == 0)
 		return "No remaining hashes";
 
 	if (db->password_count == 1) {


ghost (approved):

solardiz:
I think we need to fix whatever causes the `vec3` double-reporting issue instead. Trying to paper over inconsistent hash database will hide _some_ bugs, but not all. Even further code in this same function assumes the database is consistent; what this commit does is merely avoid reaching that code in the special case of 0 salts. I'd rather not merge this, at least not until we understand the problem better.

magnumripper:
> I think we need to fix whatever causes the `vec3` double-reporting issue instead.

I thought that issue was unrelated to this bug, but now I'm not so sure. I'll try to have a look-see with a debugger instead, and see what the db looks like when triggered.

PR #4397, opened 2020-10-04
libreoffice2john.py, ODF formats: LibreOffice 7 support, robustness fixes

### LINKED ISSUE #4392: libreoffice2john does not work with LibreOffice 7.* documents
### Steps to reproduce ###
Hello! I use the command to extract hash of ODF document:

`./libreoffice2john.py ./file.odt > odf.hash`

It returns the error:

> Traceback (most recent call last):
>   File "./bin/john-bleeding-jumbo/run/libreoffice2john.py", line 152, in <module>
>     process_file(sys.argv[k])
>   File "./bin/john-bleeding-jumbo/run/libreoffice2john.py", line 86, in process_file
>     salt = binascii.hexlify(base64.b64decode(salt)).decode("ascii")
> UnboundLocalError: local variable 'salt' referenced before assignment

Information:
**file.odt** – is Writer document, LibreOffice 7.*
**JtR** – is the last https://github.com/openwall/john/archive/bleeding-jumbo.zip version.

### System configuration ###
- `./john --list=build-info`.

> Version: 1.9.0-jumbo-1+bleeding-5b35db310d 2020-09-30 02:14:29 +0200
> Build: linux-gnu 64-bit x86_64 AVX2 AC OMP
> SIMD: AVX2, interleaving: MD4:3 MD5:3 SHA1:1 SHA256:1 SHA512:1
> CPU tests: AVX2
> $JOHN is ./
> Format interface version: 14
> Max. number of reported tunable costs: 4
> Rec file version: REC4
> Charset file version: CHR3
> CHARSET_MIN: 1 (0x01)
> CHARSET_MAX: 255 (0xff)
> CHARSET_LENGTH: 24
> SALT_HASH_SIZE: 1048576
> SINGLE_IDX_MAX: 2147483648
> SINGLE_BUF_MAX: 4294967295
> Effective limit: Number of salts vs. SingleMaxBufferSize
> Max. Markov mode level: 400
> Max. Markov mode password length: 30
> gcc version: 10.2.0
> GNU libc version: 2.32 (loaded: 2.32)
> OpenCL headers version: 2.2
> Crypto library: OpenSSL
> OpenSSL library version: 01010108f
> OpenSSL 1.1.1h  22 Sep 2020
> GMP library version: 6.2.0
> File locking: fcntl()
> fseek(): fseek
> ftell(): ftell
> fopen(): fopen
> memmem(): System's
> times(2) sysconf(_SC_CLK_TCK) is 100
> Using times(2) for timers, resolution 10 ms
> HR timer claimed resolution 1 ns, observed 173 ns
> Total physical host memory: 31966 MB
> Available physical host memory: 13779 MB

- `./john --list=opencl-devices`.

> Platform #0 name: Intel(R) OpenCL HD Graphics, version: OpenCL 2.1 
>     Device #0 (1) name:     Intel(R) Gen9 HD Graphics NEO
>     Device vendor:          Intel(R) Corporation
>     Device type:            GPU (LE)
>     Device version:         OpenCL 2.1 NEO 
>     Driver version:         1.0.0 
>     Native vector widths:   char 16, short 8, int 4, long 1
>     Preferred vector width: char 16, short 8, int 4, long 1
>     Global Memory:          25573 MB
>     Global Memory Cache:    512 KB
>     Local Memory:           64 KB (Local)
>     Constant Buffer size:   4095 MB
>     Max memory alloc. size: 4095 MB
>     Max clock (MHz):        1100
>     Profiling timer res.:   83 ns
>     Max Work Group Size:    256
>     Parallel compute cores: 24
>     Stream processors:      192  (24 x 8)
>     Speed index:            211200
> 
> Platform #1 name: NVIDIA CUDA, version: OpenCL 1.2 CUDA 11.1.70
>     Device #0 (2) name:     GeForce GTX 1050 Ti
>     Device vendor:          NVIDIA Corporation
>     Device type:            GPU (LE)
>     Device version:         OpenCL 1.2 CUDA
>     Driver version:         455.23.04 [recommended]
>     Native vector widths:   char 1, short 1, int 1, long 1
>     Preferred vector width: char 1, short 1, int 1, long 1
>     Global Memory:          4042 MB
>     Global Memory Cache:    288 KB
>     Local Memory:           48 KB (Local)
>     Constant Buffer size:   64 KB
>     Max memory alloc. size: 1010 MB
>     Max clock (MHz):        1620
>     Profiling timer res.:   1000 ns
>     Max Work Group Size:    1024
>     Parallel compute cores: 6
>     CUDA cores:             768  (6 x 128)
>     Speed index:            1244160
>     Warp size:              32
>     Max. GPRs/work-group:   65536
>     Compute capability:     6.1 (sm_61)
>     Kernel exec. timeout:   no
>     NVML id:                0
>     PCI device topology:    01:00.0
>     PCI lanes:              16/16
>     Temperature:            54°C
>     Utilization:            1%
> 
> Platform #2 name: Portable Computing Language, version: OpenCL 1.2 pocl 1.5, Release, LLVM 10.0.0, RELOC, SPIR, SLEEF, POCL_DEBUG
>     Device #0 (3) name:     pthread-Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
>     Device vendor:          GenuineIntel
>     Device type:            CPU (LE)
>     Device version:         OpenCL 1.2 pocl HSTR: pthread-x86_64-pc-linux-gnu-znver1
>     Driver version:         1.5 
>     Native vector widths:   char 16, short 16, int 8, long 4
>     Preferred vector width: char 16, short 16, int 8, long 4
>     Global Memory:          29918 MB
>     Global Memory Cache:    9 MB
>     Local Memory:           8 MB (Global)
>     Constant Buffer size:   8 MB
>     Max memory alloc. size: 8 GB
>     Max clock (MHz):        4100
>     Profiling timer res.:   1 ns
>     Max Work Group Size:    4096
>     Parallel compute cores: 12
>     Speed index:            196800

solardiz:
3 commits here, which I'd like to retain separate (merge this PR with rebase-and-merge). Fixes #4392

I tested this on two files: `john-samples/Office/LibreOffice_3.5.1/Libre-Default_myhovercraftisfullofeels_.odt` (also worked before) and `john-samples/Office/Libre-v7_123456.odt` (didn't, but works now), each with two versions of Python (2.6.6 and 3.6.9).

However, I didn't actually test on an unencrypted file - as I understand, @claudioandre-br has a sample where the script previously exhibited the "IndexError: list index out of range" issue, so I'd appreciate it if Claudio confirms that this issue is indeed gone now.

### DIFF
--- a/run/libreoffice2john.py
+++ b/run/libreoffice2john.py
@@ -47,7 +47,7 @@ def process_file(filename):
 
     is_encrypted = False
     key_size = 16
-    for i in range(0, len(elements)):
+    for i in range(0, len(elements) - 4):
         element = elements[i]
         if element.get("{urn:oasis:names:tc:opendocument:xmlns:manifest:1.0}full-path") == "content.xml":
             for j in range(i + 1, i + 1 + 4):
--- a/run/libreoffice2john.py
+++ b/run/libreoffice2john.py
@@ -50,7 +50,7 @@ def process_file(filename):
     for i in range(0, len(elements)):
         element = elements[i]
         if element.get("{urn:oasis:names:tc:opendocument:xmlns:manifest:1.0}full-path") == "content.xml":
-            for j in range(i + 1, i + 1 + 3):
+            for j in range(i + 1, i + 1 + 4):
                 element = elements[j]
                 # print element.items()
                 data = element.get("{urn:oasis:names:tc:opendocument:xmlns:manifest:1.0}checksum")
--- a/run/libreoffice2john.py
+++ b/run/libreoffice2john.py
@@ -75,6 +75,9 @@ def process_file(filename):
                 data = element.get("{urn:oasis:names:tc:opendocument:xmlns:manifest:1.0}key-size")
                 if data:
                     key_size = data
+                data = element.get("{urn:oasis:names:tc:opendocument:xmlns:manifest:1.0}start-key-generation-name")
+                if data:
+                    start_key_generation_name = data
 
     if not is_encrypted:
         sys.stderr.write("%s is not an encrypted OpenOffice file!\n" % filename)
@@ -97,7 +100,7 @@ def process_file(filename):
     elif algorithm_name.find("aes256-cbc") > -1:
         algorithm_type = 1
     else:
-        sys.stderr.write("%s uses un-supported encryption!\n" % filename)
+        sys.stderr.write("%s uses unsupported encryption!\n" % filename)
         return 6
 
     if checksum_type.upper().find("SHA1") > -1:
@@ -105,9 +108,21 @@ def process_file(filename):
     elif checksum_type.upper().find("SHA256") > -1:
         checksum_type = 1
     else:
-        sys.stderr.write("%s uses un-supported checksum algorithm!\n" % \
+        sys.stderr.write("%s uses unsupported checksum algorithm!\n" % \
                 filename)
         return 7
+
+    if start_key_generation_name.upper().find("SHA1") > -1:
+        required_checksum_type = 0
+    elif start_key_generation_name.upper().find("SHA256") > -1:
+        required_checksum_type = 1
+    else:
+        required_checksum_type = -1
+
+    if checksum_type != required_checksum_type:
+        sys.stderr.write("%s uses unsupported combination of algorithms!\n" % \
+                filename)
+        return 8
 
     meta_data_available = True
     gecos = ""
--- a/src/odf_common_plug.c
+++ b/src/odf_common_plug.c
@@ -41,6 +41,7 @@ int odf_valid(char *ciphertext, struct fmt_main *self)
 	char *ctcopy;
 	char *keeptr;
 	char *p;
+	char cipher_type, checksum_type;
 	int res, extra;
 
 	if (strncmp(ciphertext, FORMAT_TAG, FORMAT_TAG_LEN))
@@ -50,27 +51,21 @@ int odf_valid(char *ciphertext, struct fmt_main *self)
 	ctcopy += FORMAT_TAG_LEN;
 	if ((p = strtokm(ctcopy, "*")) == NULL)	/* cipher type */
 		goto err;
-	if (strlen(p) != 1)
+	if ((p[0] != '0' && p[0] != '1') || p[1])
 		goto err;
-	res = atoi(p);
-	if (res != 0 && res != 1)
-		goto err;
-
+	cipher_type = p[0];
 	if ((p = strtokm(NULL, "*")) == NULL)	/* checksum type */
 		goto err;
-	if (strlen(p) != 1)
+	if ((p[0] != '0' && p[0] != '1') || p[1])
 		goto err;
-	res = atoi(p);
-	if (res != 0 && res != 1)
-		goto err;
+	checksum_type = p[0];
 	if ((p = strtokm(NULL, "*")) == NULL)	/* iterations */
 		goto err;
 	if (!isdec(p))
 		goto err;
 	if ((p = strtokm(NULL, "*")) == NULL)	/* key size */
 		goto err;
-	res = atoi(p);
-	if (res != 16 && res != 32)
+	if (strcmp(p, "16") && strcmp(p, "32"))
 		goto err;
 	if ((p = strtokm(NULL, "*")) == NULL)	/* checksum field (skipped) */
 		goto err;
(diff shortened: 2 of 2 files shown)


ghost:
:+1: 
```
$ python libreoffice2john.py ~/Aula\ da\ Prova.odt 
/home/claudio/Aula da Prova.odt is not an encrypted OpenOffice file!
```

solardiz:
I haven't tested this on files that use SHA-1 or/and Blowfish; both samples I tested this on use SHA-256 and AES. I'd appreciate testing on SHA-1 or/and Blowfish files, and perhaps those should be added to john-samples.

solardiz:
Hmm, looks like I misunderstood, and will need to rework the last commit here (checking for different hash types). In the ODF format in JtR, we actually support these two combinations:

1. SHA-1 pre-hashing, PBKDF2-SHA1, Blowfish encryption, SHA-1 checksum.

2. SHA-256 pre-hashing, PBKDF2-SHA1, AES encryption, SHA-256 checksum.

We use SHA-1 in the KDF in both places, so at least my commit message is wrong. And then I don't know what they mean by mentioning SHA-256 in "start-key-generation-name"... oh, perhaps what I just called pre-hashing above? Then I just need to fix the commit message, and we need to separately enhance `odf_valid()` to reject unsupported combinations of hash vs. cipher.

solardiz:
Force-pushed changes per my last comment. 4 commits now, and we actually need to wait for CI since I'm modifying C code now.

run/libreoffice2john.py
```diff
@@ -75,6 +75,9 @@ def process_file(filename):
                 data = element.get("{urn:oasis:names:tc:opendocument:xmlns:manifest:1.0}key-size")
                 if data:
                     key_size = data
+                data = element.get("{urn:oasis:names:tc:opendocument:xmlns:manifest:1.0}start-key-generation-name")
+                if data:
+                    start_key_generation_name = data
 
     if not is_encrypted:
         sys.stderr.write("%s is not an encrypted OpenOffice file!\n" % filename)
```
ghost, line 83: If you are in the mood and agree, why not update all but `--help` from OpenOffice to LibreOffice.

solardiz: I'm not currently in the mood and would rather see someone do that separately. I was also thinking of opening an issue to replace all `sys.exit(-1)` in Python scripts with `sys.exit(1)` (some already do the latter).

magnumripper: Not arguing with that, but what is the difference?
```
$ sh -c "exit 0" && echo good || echo bad
good
$ sh -c "exit 1" && echo good || echo bad
bad
$ sh -c "exit -1" && echo good || echo bad
bad
```

magnumripper: > If you are in the mood and agree, why not update all but `--help` from OpenOffice to LibreOffice.

Is that better at all? Shouldn't it be something like `odf2john`?

I really do not have an opinion on it, I just sense this is weird/unstable ground.

magnumripper: ...and for end users it would obviously make most sense to just have a single `office2john` that caught all of them buggers

solardiz: `sys.exit(-1)` feels like abuse, forcing the larger integer -1 to be turned into 255. If that's the exit code we want, we'd need to cleanly spell it out as 255. But we already use simply 1 in programs where we don't have the -1 to 255 abuse.

The script was apparently called odf2john before and then renamed.

magnumripper: I couldn't care less about politics and especially not in the Open Source world but in my mind "opendocument" is/was (or was intended to be) a standard. Libreoffice is not a standard: Good or bad, it's just a fork or branch of code. It might be the original one or the intended one but that's not even relevant for us. Please correct me if I'm wrong, I'm very ignorant when it comes to things like this.

magnumripper (approved):

run/libreoffice2john.py
solardiz: I think it's a matter of what's more technically correct (ODF) vs. what users look for (LibreOffice). I'm not aware of a good solution. @ch3root might want to weigh in on this.

magnumripper: Users of OpenOffice would not likely look for LibreOffice. I can't think of a great solution either.

magnumripper: I see now the format is correctly called `ODF` which makes it even stranger to call the tool `libreoffice2john.py`. We could make a symlink... but let's not.

PR #4400, opened 2020-10-06
john.c: Bug fixes for ext_init() setting cipher_limit

### LINKED ISSUE #4367: External mode cipher_limit is not set correctly
Unless the `--max-length` option is given, `cipher_limit` is set to 125 and not to the format's maximum length. Among other effects, this results in a segfault with `phpass` and `phpass-opencl` formats when trying to use `--external=Repeats` on them.

magnumripper:
Ensure the length-update call to ext_init() is made after we've processed any length options, and that it's made at all for --stdout. Also fixes a minor bug for external Unicode modes such as repeats32, they benefit from knowing actual cipher_limit (in bytes).  Closes #4367

### DIFF
--- a/src/external.c
+++ b/src/external.c
@@ -197,9 +197,9 @@ int ext_has_function(const char *mode, const char *function)
 void ext_init(char *mode, struct db_main *db)
 {
 	ext_minlen = options.eff_minlength;
-	ext_cipher_limit = maxlen = options.eff_maxlength;
-	ext_target_utf8 =
-		(options.target_enc <= CP_UNDEF || options.target_enc == UTF_8);
+	maxlen = options.eff_maxlength;
+	ext_cipher_limit = (db && db->format) ? db->format->params.plaintext_length : maxlen;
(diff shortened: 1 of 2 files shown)


magnumripper:
I've verified this with debug prints in `ext_init()` and trying the following:
- `--stdout` with or without `--max-len`
- `--stdout=N` with or without further limit with `--max-len`
- a crack session with or without `--max-len`

plus (for the bonus fix for Unicode external modes)
- `./john -stdout -max-len=1 -ext:dumb16` prints all its 55,387 candidates despite all but 95 of them being more than a single byte of output. Same goes for repeats16/dumb32/repeats32 (the two latter producing 143,532 candidates). Before the fix, repeats16/repeats32 would stop after the 95 ASCII characters.

I can't come up with anything more to verify? As far as I can tell, things are fine now.

ghost (approved):
I think you need Solar to test this. Anyway, :+1:

Issue #4363, opened 2020-09-24
Use unified helper functions for creating and destroying OpenCL buffers/mappings - and/or for re-introducing CUDA


magnumripper:
Many formats have variants of this:
```c
#define CLCREATEBUFFER(_flags, _size)         \
    clCreateBuffer(context[gpu_id], _flags, _size, NULL, &cl_error); \
    HANDLE_CLERROR(cl_error, "Error allocating GPU memory");

#define CLKERNELARG(kernel, id, arg)      \
    HANDLE_CLERROR(clSetKernelArg(kernel, id, sizeof(arg), &arg), \
                   "Error setting kernel argument");
(...)
```

We should have such macros in `opencl_common.h` and use them *everywhere*. They can have memory-leak detection like my debug macros for #4354 (could in fact be in place all the time, not just for debugging).

magnumripper:
Note that we could also have more advanced functions like `CREATEPINNEDBUFFER(devbuf, hostbuf, size, (...))` that creates all buffers and maps them, and a corresponding destroyer that unmaps, frees, releases etc.

magnumripper:
Some dicussions in #4401 about functions that keep track. This is cool: so a format can do things like:
```c
    salt = opencl_create_buf(RO, salt_size);
    keys = opencl_create_buf(RO, gws * plaintext_size);
    hashes = opencl_create_buf(WR, gws * binary_size);
```
and later free all of them with a single
```c
    opencl_release_buffers();
```

Taking this a step further we can do the same with kernels. So we have
```c
    prepare_kernel = opencl_create_kernel("kernel_prepare");
    crypt_kernel = opencl_create_kernel("kernel_crypt_xsha");
```
This is not just for being able to release all of them using `opencl_release_kernels()`, a greater win is this function:
```c
static size_t get_task_max_work_group_size()
{
        size_t s;

        s = autotune_get_task_max_work_group_size(FALSE, 0, crypt_kernel);
        s = MIN(s, autotune_get_task_max_work_group_size(FALSE, 0, prepare_kernel));
        return MIN(s, 512);
}
```
We haven't been able to move this function to shared auto-tune code because it needs to ask every kernel of a format, so it varies. But with all kernels in a list, we obviously can!

magnumripper:
[...]

PR #4401, opened 2020-10-06
OpenCL: helper functions for create/destroy buffers

ghost:
I started working on #4363 . I was thinking about what should be done and how. 

I'm using raw-sha512 for testing. 
- Inside the new `release_clobj` I call a kind of `clear_all()` function.
```
static void release_clobj()
{
	if (ocl_initialized) {
		CLRELEASEALLMAPBUFFERS(cl_map_list, cl_map_size, CL_HARD_ERROR);
		CLRELEASEALLBUFFERS(cl_buffer_list, cl_buffer_size, CL_HARD_ERROR);
		MEM_FREE(hash_ids);

		HANDLE_CLERROR(clFinish(queue[gpu_id]), "Error releasing memory");
	}
	cl_buffer_size = 0;
	cl_map_size = 0;
	ocl_initialized = 0;
}
```
- The mask mode buffers are different. More regular, I create and destroy buffers when needed, at specific points. So, I create and release the buffers in a traditional way.
```
	buffer_bitmap = CLCREATEBUFFERx2(
	    CL_MEM_WRITE_ONLY, (bitmap_size / 32 + 1) * sizeof(uint32_t), CL_HARD_ERROR);
        [...]
	CLRELEASEBUFFERx(buffer_bitmap, CL_HARD_ERROR);
}
```

### DIFF
--- a/src/opencl_common.c
+++ b/src/opencl_common.c
@@ -7,7 +7,7 @@
  * Copyright (c) 2010-2012 Samuele Giovanni Tonon <samu at linuxasylum dot net>
  * Copyright (c) 2010-2013 Lukas Odzioba <ukasz@openwall.net>
  * Copyright (c) 2010-2019 magnum
- * Copyright (c) 2012-2015 Claudio André <claudioandre.br at gmail.com>
+ * Copyright (c) 2012-2020 Claudio André <claudioandre.br at gmail.com>
  *
  * and is hereby released to the general public under the following terms:
  * Redistribution and use in source and binary forms, with or without
@@ -87,6 +87,55 @@ static char* get_device_capability(int sequential_id);
 extern int autotune_get_next_gws_size(size_t num, int step, int startup,
                                       int default_value);
 extern int autotune_get_prev_gws_size(size_t num, int step);
+
+/*
+ * List entry.
+ */
+struct kernel_opencl {
+	struct kernel_opencl *next;
+
+	cl_kernel data;
+	cl_kernel *variable;
+};
+
+/*
+ * Main list structure, head is used to start scanning the list, while tail
+ * is used to add new entries.
+ */
+struct opencl_kernels {
+	struct kernel_opencl *head, *tail;
+
+	int count;
+};
+
+/*
+ * List entry.
+ */
+struct buffer_opencl {
+	struct buffer_opencl *next;
+
+	cl_mem data;
+	cl_mem *variable;
+	void **host_ptr;
+};
+
+/*
+ * Main list structure, head is used to start scanning the list, while tail
+ * is used to add new entries.
+ */
+struct opencl_buffers {
+	struct buffer_opencl *head, *tail;
+
+	int count;
+};
+
+//Lists to handle OpenCL buffers and kernels created by OpenCL formats.
+static struct opencl_buffers *cl_buffer_list;
+static struct opencl_kernels *cl_kernel_list;
+static void *cl_map_list[CL_MAX_BUFFERS][2];
+static int cl_map_size;
+static void buf_list_init(struct opencl_buffers **list);
+static void krn_list_init(struct opencl_kernels **list);
 
 // Settings to use for auto-tuning.
 static int buffer_size;
@@ -282,7 +331,8 @@ static char *opencl_driver_ver(int sequential_id)
 	return ret;
 }
 
-static char *remove_spaces(char *str) {
+static char *remove_spaces(char *str)
+{
 
 	char *out = str, *put = str;
 
@@ -855,6 +905,8 @@ void opencl_load_environment(void)
 		amd_probe();
 
 		// Initialize OpenCL global control variables
+		buf_list_init(&cl_buffer_list);
+		krn_list_init(&cl_kernel_list);
 		cmdline_devices[0] = NULL;
 		engaged_devices[0] = DEV_LIST_END;
 		requested_devices[0] = DEV_LIST_END;
@@ -1040,6 +1092,8 @@ void opencl_done()
 	fmt_base_name[0] = 0;
 	opencl_initialized = 0;
 	crypt_kernel = NULL;
+	MEM_FREE(cl_buffer_list);
+	MEM_FREE(cl_kernel_list);
 
 	engaged_devices[0] = engaged_devices[1] = DEV_LIST_END;
 }
@@ -2827,6 +2881,296 @@ char *get_error_name(cl_int cl_error)
 int opencl_calc_min_kpc(size_t lws, size_t gws, int v_width)
 {
 	return gws * v_width;
+}
+
+static void buf_list_init(struct opencl_buffers **list)
+{
+	*list = mem_alloc(sizeof(struct opencl_buffers));
+	(*list)->tail = (*list)->head = NULL;
+	(*list)->count = 0;
+}
+
+static void buf_list_add_link(struct opencl_buffers *list, struct buffer_opencl *entry)
+{
+	entry->next = NULL;
+
+	if (list->tail)
+		list->tail = list->tail->next = entry;
+	else
+		list->tail = list->head = entry;
+
+	list->count++;
+}
+
+static void buf_list_add(struct opencl_buffers *list, cl_mem *data, void **host_ptr)
+{
+	struct buffer_opencl *entry;
+
+	entry = mem_alloc(sizeof(struct buffer_opencl));
+	entry->data = *(data);
+	entry->variable = data;
+	entry->host_ptr = host_ptr;
+
+	buf_list_add_link(list, entry);
+}
+
+static void buf_list_remove(struct opencl_buffers *list, cl_mem data)
+{
+	struct buffer_opencl *current, *prev = NULL;
+
+	for (current = list->head; current; current = current->next) {
+		if (current->data == data) {
+			if (prev)
+				prev->next = current->next;
+
+			if (list->head == current)
+				list->head = current->next;
+			if (list->tail == current)
+				list->tail = prev;
+
+			*(current->variable) = NULL;
+			MEM_FREE(current);
+
+			list->count--;
+
+			/* Done! */
+			break;
+		}
+		prev = current;
+	}
+}
+
+static void krn_list_init(struct opencl_kernels **list)
+{
+	*list = mem_alloc(sizeof(struct opencl_kernels));
+	(*list)->tail = (*list)->head = NULL;
+	(*list)->count = 0;
+}
+
+static void krn_list_add_link(struct opencl_kernels *list, struct kernel_opencl *entry)
+{
+	entry->next = NULL;
+
+	if (list->tail)
+		list->tail = list->tail->next = entry;
+	else
+		list->tail = list->head = entry;
+
+	list->count++;
+}
+
+static void krn_list_add(struct opencl_kernels *list, cl_kernel *data)
+{
+	struct kernel_opencl *entry;
+
+	entry = mem_alloc(sizeof(struct kernel_opencl));
+	entry->data = *(data);
+	entry->variable = data;
+
+	krn_list_add_link(list, entry);
+}
+
+static void krn_list_remove(struct opencl_kernels *list, cl_kernel data)
+{
+	struct kernel_opencl *current, *prev = NULL;
+
+	for (current = list->head; current; current = current->next) {
+		if (current->data == data) {
+			if (prev)
+				prev->next = current->next;
+
+			if (list->head == current)
+				list->head = current->next;
+			if (list->tail == current)
+				list->tail = prev;
+
+			*(current->variable) = NULL;
+			MEM_FREE(current);
+
+			list->count--;
+
+			/* Done! */
+			break;
+		}
+		prev = current;
+	}
+}
+
+/*
+ * Use this function to alloc host and device OpenCL buffers.
+ * Pass host_buffer=NULL if don't want the host buffer to be malloc'ed.
+ * Later, all buffers can be released at once using opencl_release_buf().
+ *
+ * Setting keep_track to zero (0) causes the function to NOT save a reference
+ * of the newly created buffer(s). Therefore, the buffer(s) will not be released
+ * automatically by opencl_release_buf().
+ *
+ * IMPORTANT: the function keeps a reference to host_buffer and buffer_obj. So,
+ *            they can't be freed before a call to opencl_release_buf();
+ */
+cl_mem real_opencl_create_buf_pair(void **host_buffer, cl_mem *buffer_obj, cl_mem_flags flags,
+            size_t size, int keep_track, const char *file, int line)
+{
+	cl_int err_create_buffer;
+
+	if (host_buffer && *(host_buffer))
+#ifdef DEBUG
+		error_msg("** %s:%d Memory leak\n", file, line);
+#else
+		MEM_FREE(host_buffer);
+#endif
+
+	if (*(buffer_obj))
+#ifdef DEBUG
+		error_msg("** %s:%d Memory leak\n", file, line);
+#else
+		opencl_release_buf(buffer_obj);
+#endif
+
+	if (host_buffer)
+		*(host_buffer) = mem_alloc(size);
+
+	*(buffer_obj) =
+		clCreateBuffer(context[gpu_id], flags, size, NULL, &err_create_buffer);
+
+	if (err_create_buffer != CL_SUCCESS) {
+		*(buffer_obj) = NULL;
+		HANDLE_CLERROR(err_create_buffer, "Error allocating GPU memory");
+	} else
+		if (keep_track)
+			buf_list_add(cl_buffer_list, buffer_obj, host_buffer);
+
+	return *(buffer_obj);
+}
+
+/*
+ * Use this function to alloc OpenCL buffers and mapped buffers.
+ * Later, all buffers can be released at once using opencl_release_map().
+ */
+void opencl_create_map(cl_mem_flags buffer_flags, size_t size,
+        cl_mem *out_buffer, void **out_map)
+{
+	*(out_buffer) = clCreateBuffer(context[gpu_id], buffer_flags,
+		size, NULL, &ret_code);
+	HANDLE_CLERROR(ret_code, "Error allocating GPU memory");
+
+	*(out_map) = clEnqueueMapBuffer(queue[gpu_id], *(out_buffer),
+		CL_TRUE, CL_MAP_WRITE_INVALIDATE_REGION, 0, size, 0, NULL, NULL,
+		&ret_code);
+	HANDLE_CLERROR(ret_code, "Error allocating GPU memory");
+
+	if (cl_map_size < CL_MAX_BUFFERS) {
+		cl_map_list[cl_map_size][0] = *(out_buffer);
+		cl_map_list[cl_map_size++][1] = *(out_map);
+	}
+}
+
+/*
+ * Use this function to release the OpenCL buffer(s).
+ * Passing NULL as the buffer_obj parameter, causes it to release all
+ * allocated buffers. Via [cl_buffer_list].
+ */
+void opencl_release_buf(cl_mem *buffer_obj)
+{
+	cl_int ret_code = CL_SUCCESS, failure = CL_SUCCESS;
+
+	if (buffer_obj && (*(buffer_obj))) {
+		failure = clReleaseMemObject(*(buffer_obj));
+		*(buffer_obj) = NULL;
+	} else
+		for (; cl_buffer_list->count;) {
+			cl_mem current = cl_buffer_list->head->data;
+			void **host_buf = cl_buffer_list->head->host_ptr;
+
+			ret_code = clReleaseMemObject(current);
+			*(cl_buffer_list->head->variable) = NULL;
+
+			if (host_buf) {
+				MEM_FREE(*(host_buf));
+				*(host_buf) = NULL;
+			}
+			buf_list_remove(cl_buffer_list, current);
+
+			if (ret_code != CL_SUCCESS)
+				failure = ret_code;
+		}
+	HANDLE_CLERROR(failure, "Error Releasing buffers");
+}
+
+/*
+ * Use this function to release all OpenCL buffers created using
+ * opencl_create_map(). Via [cl_map_list].
+ */
+void opencl_release_map()
+{
+	cl_int ret_code  = CL_SUCCESS, failure = CL_SUCCESS;
+
+	for (cl_map_size--; cl_map_size >= 0; cl_map_size--) {
+		if (cl_map_list[cl_map_size][0]) {
+			ret_code = clEnqueueUnmapMemObject(queue[gpu_id],
+				cl_map_list[cl_map_size][0], cl_map_list[cl_map_size][1], 0, NULL, NULL);
+			if (ret_code != CL_SUCCESS)
+				failure = ret_code;
+			ret_code = clReleaseMemObject(cl_map_list[cl_map_size][0]);
+			if (ret_code != CL_SUCCESS)
+				failure = ret_code;
+		}
+		cl_map_list[cl_map_size][0] = NULL;
+		cl_map_list[cl_map_size][1] = NULL;
+	}
+	HANDLE_CLERROR(failure, "Error Releasing mapped buffers");
+	HANDLE_CLERROR(clFinish(queue[gpu_id]), "Error releasing mapped memory");
+	cl_map_size = 0;
+}
+
+/*
+ * Use this function to alloc OpenCL kernels.
+ * Later, all kernels can be released at once using opencl_release_kernels().
+ *
+ * IMPORTANT: the function keeps a reference to the "kernel_obj". So, it can't
+ *            be freed before a call to opencl_release_kernels();
+ */
+cl_kernel opencl_create_kernel(cl_kernel *kernel_obj, const char *kernel_name)
+{
+	cl_int ret_code;
+
+	if (*(kernel_obj))
+		opencl_release_kernel(kernel_obj);
+
+	*(kernel_obj) = clCreateKernel(program[gpu_id], kernel_name, &ret_code);
+
+	if (ret_code != CL_SUCCESS) {
+		*(kernel_obj) = NULL;
+		HANDLE_CLERROR(ret_code, "Error creating the kernel.");
+	} else
+		krn_list_add(cl_kernel_list, kernel_obj);
+
+	return *(kernel_obj);
+}
+
+/*
+ * Use this function to release the OpenCL kernel(s).
+ * Passing NULL as the kernel_obj parameter, causes it to release all
+ * allocated kernels. Via [cl_kernel_list].
+ */
+void opencl_release_kernel(cl_kernel *kernel_obj)
+{
+	cl_int ret_code = CL_SUCCESS, failure = CL_SUCCESS;
+
+	if (kernel_obj && (*(kernel_obj))) {
+		failure = clReleaseKernel(*(kernel_obj));
+		*(kernel_obj) = NULL;
+	} else
+		for (; cl_kernel_list->count;) {
+			cl_kernel current = cl_kernel_list->head->data;
+
+			ret_code = clReleaseKernel(current);
+			krn_list_remove(cl_kernel_list, current);
+
+			if (ret_code != CL_SUCCESS)
+				failure = ret_code;
+		}
+	HANDLE_CLERROR(failure, "Error Releasing kernels");
 }
 
 /***
--- a/src/opencl_common.h
+++ b/src/opencl_common.h
@@ -4,7 +4,7 @@
  * Common OpenCL functions go in this file.
  *
  *
- * Copyright (c) 2013-2015 Claudio André <claudioandre.br at gmail.com>,
+ * Copyright (c) 2013-2020 Claudio André <claudioandre.br at gmail.com>,
  * Copyright (c) 2012-2013 magnum,
  * Others and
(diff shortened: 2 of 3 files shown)


ghost:
No leaks so far. This is the kind of change I would like to see in `opencl_common.h`.

src/opencl_common.h
```diff
@@ -152,6 +152,11 @@ extern int device_info[MAX_GPU_DEVICES];
 #define DUR_CONFIG_NAME         "_MaxDuration"
 #define FALSE               0
 
+#define CL_SOFT_ERROR           1
+#define CL_HARD_ERROR           2
+#define CL_IGNORE_ERROR         3
+#define CL_MAX_BUFFERS         32
+
```
ghost, line 158: All of these use cases must use `CL_HARD_ERROR`. Remove `#defines` and `BENCH_CLERRORx`!?!

src/opencl_common.h
```diff
@@ -274,6 +298,158 @@ void opencl_process_event(void);
 		} \
 	} while (0)
 
+/*
+ * Use this macro for OpenCL buffer creation.
+ * Later, it can release all buffers at once
+*/
+#define CLCREATEBUFFERx(_flags, _size, error_level, buffer_list, buffer_size) ({    \
+    cl_mem __buffer;                                                                \
```
ghost: I need a better name (`CLCREATEBUFFER_ADD_LIST`)

src/opencl_common.h
```diff
@@ -274,6 +298,158 @@ void opencl_process_event(void);
...
+		  _size, NULL, &__errCreateBuffer);                                             \
+		_out_map = clEnqueueMapBuffer(queue[gpu_id], _out_buffer,                       \
+		  _map_block, _map_flags, _map_offset, _map_size, 0, NULL, NULL,                \
+		  &__errMapBuffer);                                                             \
+                                                                                        \
+		if (__errCreateBuffer != CL_SUCCESS || __errMapBuffer != CL_SUCCESS) {          \
+			_out_buffer = NULL;                                                         \
+			_out_map = NULL;                                                            \
+			if (error_level == CL_SOFT_ERROR)                                           \
+				SOFT_CLERROR(__errCreateBuffer, "Error allocating GPU memory");         \
```
ghost: Ops, should be `__failure`.

src/opencl_common.h
```diff
@@ -274,6 +298,158 @@ void opencl_process_event(void);
...
+		for (buffer_size--; buffer_size >= 0; buffer_size--) {                          \
+			if (buffer_list[buffer_size])                                               \
+				__errRelease = clReleaseMemObject(buffer_list[buffer_size]);            \
+			buffer_list[buffer_size] = NULL;                                            \
+			if (__errRelease != CL_SUCCESS)                                             \
+				__failure = __errRelease;                                               \
+		}                                                                               \
+		if (__failure != CL_SUCCESS) {                                                  \
+			if (error_level == CL_SOFT_ERROR)                                           \
+				SOFT_CLERROR(__errRelease, "Error Releasing buffers");                  \
```
ghost: Ops, should be `__failure`.

src/opencl_common.h
```diff
@@ -274,6 +298,158 @@ void opencl_process_event(void);
...
+				__errRelease = clReleaseMemObject(map_list[map_size][0]);               \
+				if (__errRelease != CL_SUCCESS)                                         \
+					__failure = __errRelease;                                           \
+			}                                                                           \
+			map_list[map_size][0] = NULL;                                               \
+			map_list[map_size][1] = NULL;                                               \
+		}                                                                               \
+		if (__failure != CL_SUCCESS) {                                                  \
+			if (error_level == CL_SOFT_ERROR)                                           \
+				SOFT_CLERROR(__errRelease, "Error Releasing mapped buffers");           \
```
ghost: Ops, should be `__failure`.

src/opencl_common.h
magnumripper: Keeping it short-ish it could be `CLCREATEBUF_LIST`. It could even be `CLCREATEBUFFERS` vs `CLCREATEBUFFER`...

Perhaps some of these things should be real functions in `opencl_common.c` instead of function-like macros.

ghost: Which ones?

ghost:
Better to see the new macros and an example of use.

src/opencl_common.h
solardiz: I was thinking the same (before I saw your comment) - we must not ignore this kind of errors, or we'd crash on a dependent call next. So no reason to support ignoring them.

src/opencl_common.h
solardiz: Different perspective: why are any of these macros at all? Shouldn't all of them be functions?

src/opencl_rawsha512_gpl_fmt_plug.c
```diff
@@ -58,6 +58,11 @@ static cl_mem pass_buffer;      //Plaintext buffer.
 static cl_mem idx_buffer;       //Sizes and offsets buffer.
 static cl_kernel prepare_kernel;
 
+//A list containing all OpenCL buffers created by the format.
+static cl_mem cl_buffer_list[CL_MAX_BUFFERS];
+static void *cl_map_list[CL_MAX_BUFFERS][2];
+static int cl_buffer_size, cl_map_size;
```
solardiz: Should we possibly provide such helpers that we'd only need to define one variable here (of a custom struct type) and the helpers would handle allocation and freeing of the buffer list dynamically? Perhaps it can be a couple of linked lists managed by the helpers, with list tails in the struct.

BTW, if we end up freeing in the reverse order, that can actually be friendly towards the underlying memory allocators.

src/opencl_common.h
magnumripper: Right!

src/opencl_rawsha512_gpl_fmt_plug.c
magnumripper: Something like that is boiling in my head too: A single define for all buffers that depend on GWS. Then a single call to allocate all of them and a single call to destroy them.

Perhaps even a single define for all buffers, where some depend on GWS and some don't? We need to keep scope in mind though.

What is the scope? I think it's "minimize (as in share) code that all formats need", but also - and this is important: "make things easy to read and follow".

ghost:
I refactored and pushed the sha512 source code (make it easier to maintain).

Now I'm moving the code from macros to functions.

ghost:
Please, comment the "C"ism;
* I can't think about more automation. Or, I don't need it. I'm happy; that is my use case;
* formats create only a few buffers. This first version does not have linked list and dynamic allocation;

IMO, the change is invasive. Macros would be a smoother transition.

src/opencl_common.h
magnumripper: > Keeping it short-ish it could be `CLCREATEBUF_LIST`. It could even be `CLCREATEBUFFERS` vs `CLCREATEBUFFER`...

I misunderstood your functions/macros yesterday. I thought the one now called `opencl_create_buf_list()` would be able to allocate several/all buffers for a format, in one call. Now I see `_list` refers to "keeping track".

Given that, I'm not sure the name need to reflect that at all (just like `malloc()` keeps lists and whatnot under the hood but it's not called `malloc_list()`). Maybe we should *only* export a function that does keep track, yet just call it `opencl_create_buf()`.

ghost:
Now:
* using a linked list;
* one createBufferOpenCL API;
```
salt = opencl_create_buf(RO, salt_size, keep_track=1);
mask = opencl_create_buf(RO, mask_size, keep_track=0);
```
Working on (bold):
> we could have it as opencl_create_buf(&buf, flags, size). This makes it easier to optionally verify it was NULL before the alloc and we can keep track of buf in our list so upon **calling opencl_release_buffers() we'd set it to NULL even though it wasn't given as an argument**.

ghost:
I will not add new resources to the PR, only refactor/rename/organize code. I tested for leaks and in a real cracking session.

* it has buffers, pinned buffers and kernels creation. Uses linked list (not yet for `map()`).

src/opencl_common.c
```diff
@@ -2829,6 +2881,269 @@ int opencl_calc_min_kpc(size_t lws, size_t gws, int v_width)
...
+		if (current->data == data) {
+			if (prev)
+				prev->next = current->next;
+
+			if (list->head == current)
+				list->head = current->next;
+			if (list->tail == current)
+				list->tail = prev;
+
+			*(current->variable) = NULL; //Is it really safe?
```
magnumripper: This is no worse than what `MEM_FREE(buf)` do on host side

src/opencl_common.c
```diff
@@ -2829,6 +2881,269 @@ int opencl_calc_min_kpc(size_t lws, size_t gws, int v_width)
...
+ *
+ * IMPORTANT: the function keeps a referent to _buffer. So, it can't be freed
+ *             before a call to opencl_release_buf();
+ */
+cl_mem opencl_create_buf(cl_mem *_buffer, cl_mem_flags _flags, size_t _size,
+            int keep_track) {
+	cl_int __errCreateBuffer;
+
+	if (*(_buffer))
+		opencl_release_buf(_buffer);
```
magnumripper: We could regard this as an error instead. Or, we could make that depend on `-DDEBUG`

src/opencl_common.c
```diff
@@ -2829,6 +2881,269 @@ int opencl_calc_min_kpc(size_t lws, size_t gws, int v_width)
...
+        cl_bool _map_block, cl_map_flags _map_flags, size_t _map_offset, size_t _map_size,
+		void **_out_map) {
+
+	cl_int __errCreateBuffer, __errMapBuffer, __failure = CL_SUCCESS;
+
+	*(_out_buffer) = clCreateBuffer(context[gpu_id], _buffer_flags,
+		_size, NULL, &__errCreateBuffer);
+	*(_out_map) = clEnqueueMapBuffer(queue[gpu_id], *(_out_buffer),
+		_map_block, _map_flags, _map_offset, _map_size, 0, NULL, NULL,
+		&__errMapBuffer);
```
magnumripper: Maybe I'm stupid now but isn't (for our use) always `_size` same as `_map_size`? If so we could drop the latter argument from the function.

Also, I **think** we (so far) never use `CL_MAP_READ` so perhaps we should hard-code that (perhaps even to `CL_​MAP_​WRITE_​INVALIDATE_​REGION`?) and drop the `_map_flags` argument.

magnumripper: All current code has `_map_block` as `CL_TRUE` so we can optimize that away as well

magnumripper: All current code has `_map_offset` as `0` so we can optimize that away as well

ghost: Done.

src/opencl_common.c
ghost: Done

src/opencl_common.c
```diff
@@ -88,6 +88,54 @@ extern int autotune_get_next_gws_size(size_t num, int step, int startup,
                                       int default_value);
 extern int autotune_get_prev_gws_size(size_t num, int step);
 
+/*
+ * List entry, allocated as (sizeof(struct kernel_opencl) + sizeof(data)).
+ */
```
ghost, line 93: To avoid duplication, I can use pseudo-generics. But, the source code becomes hard to follow.

src/opencl_common.c
```diff
@@ -2829,6 +2881,273 @@ int opencl_calc_min_kpc(size_t lws, size_t gws, int v_width)
 	return gws * v_width;
 }
 
+static void buf_list_init(struct opencl_buffers **list) {
```
ghost: Idem. To avoid duplication, I can use pseudo-generics. But, the source code becomes hard to follow.

magnumripper: Very short and concise functions, probably better to keep them as-is for readability.

src/opencl_common.c
```diff
@@ -2829,6 +2881,273 @@ int opencl_calc_min_kpc(size_t lws, size_t gws, int v_width)
...
+ */
+void opencl_create_map(cl_mem_flags buffer_flags, size_t size,
+        cl_mem *out_buffer, void **out_map) {
+
+	cl_int err_create_buffer, err_map_buffer, failure = CL_SUCCESS;
+
+	*(out_buffer) = clCreateBuffer(context[gpu_id], buffer_flags,
+		size, NULL, &err_create_buffer);
+	*(out_map) = clEnqueueMapBuffer(queue[gpu_id], *(out_buffer),
+		CL_TRUE, CL_MAP_WRITE | CL_MAP_WRITE_INVALIDATE_REGION, 0, size, 0, NULL, NULL,
```
magnumripper: If I got it right, the `CL_MAP_WRITE` is superfluous here but I'm not 100% sure

src/opencl_rawsha512_gpl_fmt_plug.c
```diff
@@ -162,33 +196,49 @@ static uint64_t *crypt_one_x(int index) {
...
-static void release_mask_buffers()
+static void create_bitmap_buffers()
 {
-	MEM_FREE(saved_bitmap);
+	release_bitmap_buffers();
+
+	bitmap = (uint32_t *)
+		mem_alloc((bitmap_size / 32 + 1) * sizeof(uint32_t));
+	opencl_create_buf(&dev_bitmap,
+	    CL_MEM_WRITE_ONLY, (bitmap_size / 32 + 1) * sizeof(uint32_t), CL_DONT_KEEP_TRACK);
```
magnumripper: Shouldn't most our functions (or new alternatives to them) cover both?

Something like:
```c
opencl_create_bufs(&bitmap, &dev_bitmap, RO, (bitmap_size / 32 + 1) * sizeof(uint32_t), CL_NO_TRACK);
```
This would cover lines 225-228: Do a malloc for `bitmap` and an OpenCL create for `dev_bitmap` in a single call. With future means for great debugging assertions.

src/opencl_common.c
ghost: Reading again, they are mutually exclusive. So, yes.

ghost: BTW: it is 1.2, is everyone okay with this?

src/opencl_rawsha512_gpl_fmt_plug.c
ghost: Done.

src/opencl_common.c
magnumripper: We have this in `opencl_common.h`
```c
#define CL_TARGET_OPENCL_VERSION 120
```

src/opencl_rawsha512_gpl_fmt_plug.c
ghost: We have a problem here. At least one: salt. We usually don't `malloc()` salt.

magnumripper: So we need alternative functions: Ones that only allocates a device buffer, and ones that allocate a host buffer and a device buffer with same size.

I guess the `_map` functions can always do both though? Or maybe they already do.

magnumripper: It may seem overkill but in my head it will simplify 99% of the formats so I think we should really consider these things. That being said, I'm just brainstorming.

ghost: If you think that `MEM_FREE()` is going to happen automagically, it looks good.

magnumripper: Absolutely. Perhaps we can call this `opencl_create_buf_pair`.  On call to `opencl_release_buf_pairs()` - or whatever we'll call this - we'd release both.

magnumripper: Now, we could make these `_pair` functions yet another set of similar functions, or we could do half of them with macros. I'm still merely brainstorming. It could go something like this (just from head):
```c
cl_mem opencl_create_buf_pair(void **host_buf cl_mem *buffer_obj, cl_mem_flags flags, size_t size, int keep_track)
{
	cl_int err_create_buffer;

	if (*(buffer_obj))
#ifdef DEBUG
		error_msg("** %s:%d Memory leak (device buffer)\n", __FILE__, __LINE__);
#else
		opencl_release_buf(buffer_obj);
#endif

	if (*(host_buf) && **(host_buf))
#ifdef DEBUG
		error_msg("** %s:%d Memory leak (host buffer)\n", __FILE__, __LINE__);
#else
		MEM_FREE(**host_buf);
#endif

	*(buffer_obj) = clCreateBuffer(context[gpu_id], flags, size, NULL, &err_create_buffer);

	if (err_create_buffer != CL_SUCCESS) {
		*(buffer_obj) = NULL;
		HANDLE_CLERROR(err_create_buffer, "Error allocating GPU memory");
	} else
		if (keep_track)
			buf_list_add(cl_buffer_list, buffer_obj);

	if (*host_buf)
		*(host_buf) = mem_alloc(size);

	return *(buffer_obj); /* or should be drop this and just be void? Or return an error code? The buffer was in the parameters */
 }

/* Version for device-side only */
#define opencl_create_buf(buffer_obj, flags, size, keep_track) \
	opencl_create_buf_pair(NULL, buffer_obj, flags, size, keep_track)
```

magnumripper: BTW since `opencl_create_buf_*()` are no longer macros, we lose the value of `__FILE__, __LINE__`. We could fix that by wrapping them with master macros much like `error_msg()` et al.

ghost: If it prints `opencl_common.c line 3007` it is ok to me. The same failure/message can happen in some places, the message will be useful. Anyone (I mean Solar) would like to say something about macroization?

magnumripper: It's trivial really.

If the real function is
```c
cl_mem opencl_create_buf_pair(void **host_buf, cl_mem *buffer_obj, cl_mem_flags flags, size_t size, int keep_track)
```
we modify it to be
```c
cl_mem real_opencl_create_buf_pair(void **host_buf, cl_mem *buffer_obj, cl_mem_flags flags, size_t size, int keep_track, char *file, int line)
```
and use a macro for calling it
```c
#define opencl_create_buf_pair(host_buf, buffer_obj, flags, size, keep_track) \
    real_opencl_create_buf_pair(host_buf, buffer_obj, flags, size, keep_track, __FILE__, __LINE__)
```

ghost:
Sorry, for the spam. BTW: this is ready for review.

src/opencl_common.c
```diff
@@ -2829,6 +2882,291 @@ int opencl_calc_min_kpc(size_t lws, size_t gws, int v_width)
 	return gws * v_width;
 }
 
+static void buf_list_init(struct opencl_buffers **list) {
```
magnumripper: (Many places)

Our code style (wherever it's stated, if it is at all) says this (for functions only)
```diff
- static void buf_list_init(struct opencl_buffers **list) {
+ static void buf_list_init(struct opencl_buffers **list)
+ {
```

src/opencl_common.c
```diff
@@ -2829,6 +2882,291 @@ int opencl_calc_min_kpc(size_t lws, size_t gws, int v_width)
...
+	*(out_buffer) = clCreateBuffer(context[gpu_id], buffer_flags,
+		size, NULL, &err_create_buffer);
+	*(out_map) = clEnqueueMapBuffer(queue[gpu_id], *(out_buffer),
+		CL_TRUE, CL_MAP_WRITE_INVALIDATE_REGION, 0, size, 0, NULL, NULL,
+		&err_map_buffer);
+
+	if (err_create_buffer != CL_SUCCESS)
+		failure = err_create_buffer;
+	if (err_map_buffer != CL_SUCCESS)
+		failure = err_map_buffer;
```
magnumripper: If `clCreateBuffer()` fails with error, shouldn't we catch it and bail *before* going on with `clEnqueueMapBuffer()`? Maybe I overlook something but I can't see the point of keeping separate error variables and chug along.
Also, `ret_code` is already globally declared in opencl_common.c so why not use that one:
```c
	(...)
	*(out_buffer) = clCreateBuffer(context[gpu_id], buffer_flags,
		size, NULL, &ret_code);
	HANDLE_CLERROR(ret_code, "Error allocating GPU memory");

	*(out_map) = clEnqueueMapBuffer(queue[gpu_id], *(out_buffer),
		CL_TRUE, CL_MAP_WRITE_INVALIDATE_REGION, 0, size, 0, NULL, NULL,
		&ret_code);
	HANDLE_CLERROR(ret_code, "Error allocating GPU memory");

	cl_map_list[cl_map_size][0] = *(out_buffer);
	cl_map_list[cl_map_size++][1] = *(out_map);
```
The above will also let the `HANDLE_CLERROR` give (more) correct line numbers for failure, instead of always stating it was line 3059 (BTW we should make them refer to the caller's file:line instead, perhaps I should add that to this PR? Fairly trivial.)

src/opencl_rawsha512_gpl_fmt_plug.c
```diff
@@ -197,145 +246,56 @@ static void create_clobj(size_t gws, struct fmt_main *self)
 		mask_cand = mask_int_cand.num_int_cand;
 		mask_gws = gws;
 	}
+	// create pinned buffers
+	opencl_create_map(
+		CL_MEM_READ_ONLY | CL_MEM_ALLOC_HOST_PTR, BUFFER_SIZE * gws,
+		&pinned_plaintext,     /* buffer created on device */
+		(void *) &plaintext);  /* mapped buffer created on host */
```
magnumripper, line 255: (Many places)

No need to cast to `void*`, that's sort of the idea with using it as function parameter!

ghost:
```
opencl_rawsha512_gpl_fmt_plug.c:255: warning: passing argument 4 of ‘opencl_create_map’ from incompatible pointer type
opencl_common.h:306: note: expected ‘void **’ but argument is of type ‘uint32_t **’

```

magnumripper: Oh I didn't know that, pardon me. Apparently it's not allowed with a double pointer. I can't fathom why.

solardiz:
Actually switching to these helper functions across all OpenCL formats is a fairly invasive change. Maybe we should first work on fixing the known bugs (most notably, #4044), make a release, and only then merge these changes.

magnumripper:
> Actually switching to these helper functions across all OpenCL formats is a fairly invasive change. Maybe we should first work on fixing the known bugs (most notably, #4044), make a release, and only then merge these changes.

That's not a reason for waiting with merging this. It opens up for switching the rest of the formats whenever someone gets the energy, but it doesn't mandate it.

solardiz:
OK, I don't mind merging this PR (shared code + one example).

I notice the linked list implementation is copy-paste-edit from our generic `list[.ch]` code - even the comments are copied, but are no longer right for the modified structs/code. I think instead of copy-paste-edit, we should literally use our shared generic `list[.ch]` code - it's meant to provide an opaque linked list implementation, where we only need to define a custom struct for the `data` field. We can use typecasts to pointers to our custom struct, which is OK per C aliasing rules as long as that's the only way we access the data. That's how we already use this generic/opaque implementation in many places - grep the source tree for "struct list" to find those.

solardiz:
> That's how we already use this generic/opaque implementation in many places

Upon a closer look, all of our current uses of that implementation are for lists of strings, not of arbitrary binary structs, and the comments say so and the `list_*` functions actually assume C strings. So perhaps the copy-paste-edit is right, but we need to revise the comments accordingly.

Longer-term and separately, we could add a `list_add_binary` function, etc. and start to reuse the shared implementation as truly opaque.

src/opencl_common.c
```diff
@@ -88,6 +88,55 @@ extern int autotune_get_next_gws_size(size_t num, int step, int startup,
...
 
+/*
+ * List entry, allocated as (sizeof(struct kernel_opencl) + sizeof(data)).
+ */
+struct kernel_opencl {
+	struct kernel_opencl *next;
+
+	cl_kernel data;
+	cl_kernel *variable;
+};
```
solardiz, line 99: Why does the comment say ", allocated as (sizeof(struct kernel_opencl) + sizeof(data))" when the struct itself has a fully-sized field `data`? I think that comment should be dropped, and the allocation simplified.

src/opencl_common.c
```diff
@@ -88,6 +88,55 @@ extern int autotune_get_next_gws_size(size_t num, int step, int startup,
...
+/*
+ * List entry, allocated as (sizeof(struct buffer_opencl) + sizeof(data)).
+ */
+struct buffer_opencl {
+	struct buffer_opencl *next;
+
+	cl_mem data;
+	cl_mem *variable;
+	void **host_ptr;
+};
```
solardiz, line 120: Ditto about `data`.

src/opencl_common.c
```diff
@@ -88,6 +88,55 @@ extern int autotune_get_next_gws_size(size_t num, int step, int startup,
...
+
+/*
+ * Main list structure, head is used to start scanning the list, while tail
+ * is used to add new entries.
+ */
+struct opencl_buffers {
+	struct buffer_opencl *head, *tail;
+
+	int count;
+};
```
solardiz, line 130: I think it'd be convenient for us to start all of these struct names with some prefix indicating they're ours and come from this source file. Like `john_opencl_kernel`, `john_opencl_kernels`, `john_opencl_buffer`, `john_opencl_buffers`. This will become especially important, and will save us the need to rename them later, if we ever have to make them visible to other source files.

src/opencl_common.c
```diff
@@ -2829,6 +2883,296 @@ int opencl_calc_min_kpc(size_t lws, size_t gws, int v_width)
...
+	list->count++;
+}
+
+static void buf_list_add(struct opencl_buffers *list, cl_mem *data, void **host_ptr)
+{
+	struct buffer_opencl *entry;
+
+	entry = mem_alloc(sizeof(struct buffer_opencl) + sizeof(data));
+	entry->data = *(data);
+	entry->variable = data;
```
solardiz, line 2911: I think the ` + sizeof(data)` merely wastes memory - we never actually write beyond the size of the struct here, right?

Also, why do we store both `*data` and `data`? In other words, we store the same thing by value and by reference. Why?

Finally, assuming we need it, lets drop the braces in `*(data)`.

src/opencl_common.c
```diff
@@ -2829,6 +2883,296 @@ int opencl_calc_min_kpc(size_t lws, size_t gws, int v_width)
...
+	list->count++;
+}
+
+static void krn_list_add(struct opencl_kernels *list, cl_kernel *data)
+{
+	struct kernel_opencl *entry;
+
+	entry = mem_alloc(sizeof(struct kernel_opencl) + sizeof(data));
+	entry->data = *(data);
+	entry->variable = data;
```
solardiz, line 2968: Same comments as for `buf_list_add`.

src/opencl_common.c
```diff
@@ -2829,6 +2883,296 @@ int opencl_calc_min_kpc(size_t lws, size_t gws, int v_width)
...
+
+	buf_list_add_link(list, entry);
+}
+
+static void buf_list_remove(struct opencl_buffers *list, cl_mem data)
+{
+	struct buffer_opencl *current, *prev = NULL;
+
+	for (current = list->head; current; current = current->next) {
+		if (current->data == data) {
```
solardiz, line 2922: Why do we choose to find the thing to remove by value?

src/opencl_common.c
```diff
@@ -2829,6 +2883,296 @@ int opencl_calc_min_kpc(size_t lws, size_t gws, int v_width)
...
+
+	krn_list_add_link(list, entry);
+}
+
+static void krn_list_remove(struct opencl_kernels *list, cl_kernel data)
+{
+	struct kernel_opencl *current, *prev = NULL;
+
+	for (current = list->head; current; current = current->next) {
+		if (current->data == data) {
```
solardiz, line 2978: Why do we choose to find the thing to remove by value?

solardiz (changes requested):
At least the weird overallocations need to be fixed.

solardiz:
Why do we introduce arbitrary list remove functions for the buffers and the kernels? Is this for some kind of special cases? I think normally we'll just be deallocating things in reverse order, so we might need more like a "free up to that entry" function (and if passed NULL, then free all).

src/opencl_common.c
ghost: To allow a `release_buf(pointer)` in the future.

solardiz: OK. We might or might not need that. But anyway, why search by value?

src/opencl_common.c
magnumripper: I've been pushing the use of `opencl_` to indicate this source file. Khronos OpenCL stuff start with `cl_`.

solardiz: OK, if `opencl_` is our convention so far I don't insist on also having `john_`.

src/opencl_common.c
ghost: > Also, why do we store both *data and data? In other words, we store the same thing by value and by reference. Why?

Think about:
1. the OpenCL buffer, the buffer itself, and;
2. the place where the format that made the call stores the information. 

What is of interest to the shared code is the OpenCL buffer, the reference is only for cleaning (assign NULL). I would not like to trust the reference.

ghost:
> Why do we introduce arbitrary list remove functions for the buffers and the kernels? Is this for some kind of special cases? I think normally we'll just be deallocating things in reverse order, so we might need more like a "free up to that entry" function (and if passed NULL, then free all).

IMO, it's the opposite. Formats create and destroy en bloc (usually). Other opinions are welcome.

PR #4404, opened 2020-10-13
Fix a missing paren in doc/NEWS

magnumripper:


doc/NEWS
```diff
@@ -113,9 +113,10 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
 
 - Added -mlc option to unique - "max. length considered".  [magnum; 2020]
 
-- 7z formats:  Added support for BCJ et. al. (that's pre-processors sometimes
-  used before compression.  Before this we had false negatives for those.
-  Also updated LZMA code from upstream, to version 19.00.  [magnum; 2020]
+- 7z formats:  Added support for BCJ et. al. (those are pre-processors
```
solardiz: Nitpick: the dot after "et" is an error because that's a complete word, not an abbreviation, and more importantly "et al." generally refers to people. So maybe "Added support for BCJ and other preprocessors sometimes used before compression."

magnumripper: I'll fix that

ghost (approved):

Issue #3511, opened 2018-12-15
Add a log_print() function


magnumripper:
Something like:
```c
    log_print(int verbosity, bool main_only, char *format, ...)
```
Goals: Any code that now prints to stderr and/or log with a bunch of conditions (with 3-10 lines of code each time) could be replaced with a single call to a shared function in logger.c with following options:
* Send to screen, or log, or both.
* Mute unless we're the mother process: yes/no.
* Minimum verbosity for sending to screen and a separate minimum "log level" for sending to log. BTW I'm think I want specific future options like `--log-no-rule-details` and `--log-no-mode-details` while `--log-none` would deprecate the current `-nolog`.

I have initial code for this, but it really needs some thought to end up really good.

magnumripper:
For screen output, while at it we could easily ensure words don't break at line wrap. We either can know the terminal size or just default to 80.

magnumripper:
#4086 

* Also, this could add optional color to the [screen] output, eg. yellow for important notes, red for more serious warnings.
* And, easily done, we could line-wrap at **word** boundaries as opposed to in the middle of a word. Without the caller needing to care at all.

magnumripper:
When using `-show`, color could be [great too](https://github.com/magnumripper/JohnTheRipper/issues/4243#issuecomment-609478721).

magnumripper:
Some test cases:
- Input has a word longer than width (disable wrapping for it, set `col = 0` appropriately for next line).
- Indentation wider than width (disable indentation, or decrease it to 1).
- Lines containing some or lots of multi-byte UTF-8. Lorem Ipsum transliterated to Greek is `Λορεμ ιπσθμ δολορ σιτ αμετ, πλατονεμ vολθπτατιβθσ ιν σιτ, εθ ερατ ιντερπρεταρισ μεα. Vιξ φεθγιατ δεσερθντ ελοqθεντιαμ θτ, μεα εα αθδιαμ φαβελλασ, περ πριμισ τιμεαμ δισcερε cθ. Cθ δολορ ρεcτεqθε δισπθτατιονι μει, μοvετ vοcεντ εθμ αν. Vενιαμ περσεcθτι θσθ νο, σαλθτανδι ιτθπερατα ρεπθδιανδαε αν vιμ. Γραεcι αδιπισcι cονσεcτετθερ νο σεα. Αδ qθο προβο ρατιονιβθσ, qθοδ σιντ cονσεcτετθερ πρι θτ.`

[...]

PR #4405, opened 2020-10-13
Helper print funcions

ghost:
```
Lines: 22
Columns: 78
-------
Lorem ipsum dolor sit amet, consectetur adipisicingelit, sed do eiusmod tempor
incididunt ut labore etdolore magna aliqua.

Lorem ipsum dolor sit amet, consectetur adipisicingelit, sed do eiusmod tempor
                            incididunt ut labore etdolore magna aliqua.
-------
Options:
      --config string      Location of config files (default
                           "/home/claudio/.john")blablabla blabla blabla
                           blabla bla bla blabla blabla bla bla blabla
                           balabalbaaa.
  -v, --version            Print version information and quit.
```

```
Lines: 22
Columns: 85
-------
Lorem ipsum dolor sit amet, consectetur adipisicingelit, sed do eiusmod tempor
incididunt ut labore etdolore magna aliqua.
Lorem ipsum dolor sit amet, consectetur adipisicingelit, sed do eiusmod tempor
                            incididunt ut labore etdolore magna aliqua.
Options:
      --config string      Location of config files (default
                           "/home/claudio/.john")blablabla blabla blabla blabla bla
                           bla blabla blabla bla bla blabla balabalbaaa.
  -v, --version            Print version information and quit.

```

```C
log_print(OUT_CONSOLE, VERB_DEFAULT, FALSE, strlen("Lorem ipsum dolor sit amet, "),
       "Lorem ipsum dolor sit amet, consectetur adipisicing"
       "elit, sed do eiusmod tempor incididunt ut labore et"
       "dolore magna aliqua. Value is %d.\n", 10);

```
```
Lines: 22
Columns: 90
-------
Options:
Lorem ipsum dolor sit amet, consectetur adipisicingelit, sed do eiusmod tempor incididunt
ut labore etdolore magna aliqua.
Lorem ipsum dolor sit amet, consectetur adipisicingelit, sed do eiusmod tempor incididunt
                            ut labore etdolore magna aliqua. Value is 10.
Options:
      --config string      Location of config files (default
                           "/home/claudio/.john")blablabla blabla blabla blabla bla bla
                           blabla blabla bla bla blabla balabalbaaa.
  -v, --version            Print version information and quit.
```

Usage (better to put heading as a parameter? or as `strlen("Lorem ipsum dolor sit amet, "),`):
```
log_print(OUT_CONSOLE, VERB_DEFAULT, FALSE, 
       "Lorem ipsum dolor sit amet, " // note the space
       "consectetur adipisicing elit, sed do "
       "eiusmod tempor incididunt ut labore et "
       "dolore magna aliqua. Value is %d.\n", 10);
```

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -1,0 +1,175 @@
+/*
+ * Developed by Claudio André <claudioandre.br at gmail.com> in 2020
+ *
+ * Copyright (c) 2020 Claudio André <claudioandre.br at gmail.com>
+ *
+ * This program comes with ABSOLUTELY NO WARRANTY; express or implied .
+ * This is free software, and you are welcome to redistribute it
+ * under certain conditions; as expressed here
+ * http://www.gnu.org/licenses/gpl-2.0.html
+ */
+
+#ifndef __MINGW32__
+#include <sys/ioctl.h>
+#endif
+#include <stdarg.h>
+#include <stdio.h>
+#include <unistd.h>
+
+#include "config.h"
+#include "john.h"
+#include "logger.h"
+#include "options.h"
+#include "util.h"
+
+static int is_blank(char symbol)
+{
+	if (symbol == ' ' || symbol == '\t' || symbol == '\n' || symbol == '\r')
+		return 1;
+
+	return 0;
+}
+
+static int is_separator(char symbol)
+{
+	if (strchr(")]},.", symbol) || is_blank(symbol) || symbol == '\0')
+		return 1;
+
+	return 0;
+}
+
+static inline int wordlen(const char * str)
+{
+	int len = 0;
+
+	while (!is_separator(str[len]))
+		len++;
+
+	return len;
+}
+
+static int wrap(char * text, char * line, const int width, int first_line,
+    int identation)
+{
+	int line_len = 0, pos = 0, limit;
+
+	/* ltrim text. */
+	while (!first_line && text[0] != '\0' && is_blank(text[0])) {
+		text++; pos++;
+	}
+
+	/* Protection against invalid data. */
+	if (identation > width || first_line)
+		identation = 0;
+	limit = width - identation;
+
+	while (text[line_len] != '\0') {
+
+		if (is_separator(text[line_len]))
+			if ((text[line_len] == '\n') ||
+			     line_len + wordlen(&text[line_len + 1]) >= limit) {
+				/* If it is a blank, override it. */
+				if (is_blank(text[line_len]))
+					text[line_len] = '\0';
+
+				/* The separator belongs to this line. */
+				line_len++;
+				break;
+			}
+		line_len++;
+
+		/* Separator was not found. */
+		if (line_len == limit)
+			break;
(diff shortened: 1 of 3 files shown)


src/util.c
```diff
@@ -0,0 +1,161 @@
...
+#include "config.h"
+#include "john.h"
+#include "logger.h"
+#include "options.h"
+#include "util.h"
+
+static int is_separator(char symbol)
+{
+
+	if (symbol == ' ' || symbol == '\t' || symbol == '\n' || symbol == '\0')
```
ghost: More separators: `)`,`]`,`}`,`,` , `.`.

src/util.c
```diff
@@ -0,0 +1,161 @@
...
+	return len;
+}
+
+static int wrap(const char * text, char * line, const int width, int first_line,
+    int identation)
+{
+	int line_len = 0, start = 0, limit;
+
+	/* ltrim text. */
+	while (!first_line && text[start] != '\0' && is_separator(text[start]))
```
ghost: The correct here is `is_blank(char);`

magnumripper: Yes, or `is_whitespace()`

magnumripper:
Ref #3511 

Cool! Maybe this new `util.c` is where this function should be, but in that case it shouldn't be called `log_print()` because the `log_` prefix is for `logger.c`.

I pictured a `log_print()` function in `logger.c` since it can optionally log and/or print the one same message (as-is to log and line wrapped to screen). I'm still thinking perhaps the lot should be in `logger.c` (@solardiz what's your opinion?).

src/util.c
```diff
@@ -0,0 +1,161 @@
...
+		if (limit && *(cols) > limit)
+			*(cols) = limit;
+		if (*(cols) > 140)
+			*(cols) = 140;
+		if (*(cols) < 60)
+			*(cols) = 60;
+		return 0;
+	}
+#endif
+	*(lines) = 20;
```
magnumripper, line 119: I think default height would traditionally be 24.

solardiz:
> I'm still thinking perhaps the lot should be in `logger.c` (@solardiz what's your opinion?).

I agree that at least a function that both logs to a file and prints to the terminal should be in `logger.c`. Possibly other related functions too. Also, I am unhappy with any of this being GPL-only. I am barely OK with Claudio's formats being GPL-only because they're easy to exclude, but this is shared code that wouldn't be so easy to exclude if need be for licensing reasons.

src/util.c
ghost: I already changed this (to 24x80). The default in my (GNOME) terminal.

ghost:
(After renaming it to `gpl_util.?`) I'm planning to use this for OpenCL messaging. I'll add a helper that prints the correct `device_id` (1-based) and controls `NODE` printing.

I will (in shared code):
```
#ifdef GPL
    fancy_print();
#else
    printf(...); //Regular
#endif
```
Colors:
image

PR #4406, opened 2020-10-14
Unicode tweaks for use eg. in log_print()

magnumripper:
Improve parsing of LC_* environment, store a "terminal encoding".
Add a strlen_any() function, that can handle invalid UTF-8 best-effort.


ghost (approved):
Seem ok, but only you understand encoding!

PR #4409, opened 2020-10-20
Bugfix for MPI vs. fallback binaries

magnumripper:
We must not initialize MPI until we are running the "final" binary. This
was supposed to be fixed in 90f8a5b5 but it only fixed CPU fallback while
somehow overlooking OMP fallback.

Problem only seen when using stdin and only with certain versions of MPI,
such as with Ubuntu 18.04 but not 16.04 or 20.04, and (obviously) only
after an actual execv() was made.


ghost (approved):
Passed my tests

PR #4414, opened 2020-10-25
openssl2john.py string formatting syntax error

CyberAstronaut101:
Fixing formatting that results in openssl2john.py failing with a SyntaxError


ghost (approved):
Thanks

PR #4416, opened 2020-10-26
Refactored telegram2john and added Telegram Desktop 2.1.14+ support

philsmd:
This pull requests adds support for the new `Telegram Desktop` version 2.1.14+ (beta) or 2.2.0+ (major release) algorithm to `telegram2john.py`.

@magnumripper already confirmed at https://github.com/openwall/john/issues/4387 that the `John` dev team is planning to add the support for this new algo using `PBKDF2-HMAC-SHA512` instead of `PBKDF2-HMAC-SHA1` (and an initial sha512 call and higher iteration counts) to `John` by adding the cracking code soon.

The hash format for the new hash type is similar to the old `Telegram Desktop` (using PBKDF2-HMAC-SHA1) but it uses `$telegram$2` as a signature (default hexadecimal salt and data length remained the same).

This commit also tries to refactor the script a little bit and most importantly it tries to find all the important "key" files (old and new, `map*` and `key_*`, often just `maps`/`map0` and `key_datas`) when the user only specifies the `Telegram Desktop` profile directory ("%APPDATA%\Telegram Desktop\").

I unfortunately couldn't test it a lot because I do NOT have examples of the already supported Android `userconfing.xml` files, nor old `maps` files... I only tested it with new and old `Telegram Desktop` versions (before and after the changes of the Telegram dev team, see our discussion at https://github.com/openwall/john/issues/4387), but also here I only tested it with a new profile and not an upgraded account (nor did I really complete the whole Telegram registration process, I don't think it matters a lot, but I just want to be very clear that it's not a perfect test and I hope we get some user feedback soon, especially if something is not working).

Thank you very much

### DIFF
--- a/run/telegram2john.py
+++ b/run/telegram2john.py
@@ -408,7 +408,7 @@ def process_xml(file_path):
     if PY3:
         salt = salt.decode("ascii")
 
-    print("$dynamic_1528$%s$HEX$%s" % (h, salt))
+    print("%s:$dynamic_1528$%s$HEX$%s" % (os.path.basename(file_path), h, salt))
 
 def process_map0(file_path):
     if not is_valid_tdfs(file_path):
--- a/run/telegram2john.py
+++ b/run/telegram2john.py
@@ -49,7 +49,12 @@ import binascii
 import xml.etree.ElementTree as ET
 import struct
 import hashlib
-from Crypto.Cipher import AES
+
+try:
+    from Crypto.Cipher import AES
+except ImportError:
+    sys.stderr.write("PyCrypto is missing, run 'pip install --user PyCrypto' to install it!\n")
+    sys.exit(-1)
 
 PY3 = sys.version_info[0] == 3
 
--- a/run/telegram2john.py
+++ b/run/telegram2john.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 
 """Utility to extract "hashes" from Telegram Android app's userconfing.xml
-file(s) and from Telegram Desktop's local storage (map) file"""
+file(s) and from Telegram Desktop's local storage (map/key_datas) files"""
 
 # Android App:
 # Tested with Telegram for Android v4.8.4 in February, 2018.
@@ -9,8 +9,8 @@ file(s) and from Telegram Desktop's local storage (map) file"""
 # Special thanks goes to https://github.com/Banaanhangwagen for documenting
 # this hashing scheme.
 #
-# See "UserConfig.java" from https://github.com/DrKLO/Telegram/ for details on
-# the hashing scheme.
+# See "UserConfig.java" and "SharedConfig.java" from
+# https://github.com/DrKLO/Telegram/ for details on the hashing scheme.
 #
 # Written by Dhiru Kholia <dhiru at openwall.com> in February, 2018 for JtR
 # project.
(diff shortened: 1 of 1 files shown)


philsmd:
update: I decided to add a warning about the `PyCrypto` python module requirement and therefore I added this commit here. I think that it's a special case that `Crypto.Cipher` must be installed by using `pip install PyCrypto` (not exactly the same name), so it makes sense to explain it to the user a little bit better, i.e. what (module) is missing in this specific case.

I think this makes it more user friendly and it seems from comments like this https://github.com/openwall/john/issues/4387#issuecomment-716174206 , that users are confused about the general python error message (about missing python modules) and that's not clear what they should do (even though I still think that we really shouldn't do this for each and every python modules... users could just search on the internet what that specific `ModuleNotFoundError` or similar means and how to fix it).

Thx

ghost:
@philsmd , please let me know if you are still working on this or if you want it merged;

philsmd:
Thanks @claudioandre-br for the reply. It's ready to be merged. The only problem that is not solved yet is that `John` doesn't support the new hash type that starts with `$telegram$2` . I wasn't sure what the normal procedure is for the `John` project or dev team to do if only the extraction tool supports a specific hash type. As far as I understand the message from @magnumripper here https://github.com/openwall/john/issues/4387#issuecomment-716149361 , it seems that this update can be merged before the cracking algorithm is supported and @magnumripper is planning to add it soon.

Thx

ghost (approved):

ghost:
We are trying to use structured commit messages. Please, use `some rules` in the future

Something like this:
1. Separate subject from body with a blank line
2. Limit the subject line to 50-55 characters
3. Do not end the subject line with a period
4. Wrap the body at 72-80 characters
5. Body ends with a period.

- https://www.conventionalcommits.org/en/v1.0.0-beta.2/

zloishax:
hello i have such a problem. help me (telegram2john)

`key_datas:$telegram$2*100000*0970f6c043d855aa895703b8a1cc086109cf081f72a77b6504f7f4bf3db06420*129294a5eac3196a4c9a88e556f7507b0957f6dd45d704db8abe607ec6d807270c02635289056256a6044a6408e7ef5d33f98c561f16f8aedd2b3ae33ddffddc63c8584dcb232c9f610953f461adb8d29da83f2b01e32db98101febffae4072703bfbfd492e1dd6abeb0926d3df2ed3b47dee4eb6c9f42ab657f89f19d7314c07e2ffc843e448c6d324e9f8d2c3e877a25b0b153736fddb35f5205737620ba2f96aa47f799366150b4de89a0c6e12caa4f03553d164ce9e2e975aadc83538e6ae3df93acb6026f97ac9f6f017a6bbc6607767e591b2c732e3c0ac844584c69dae89ca3272c996eb83b4e66976e3851cfc89be11dc602bb8c0cdf578d9a0a9dbc2296888fa5ee7e58d985a9bf9a1dbc75d2ddfd6ce222c5ee9f3bb40f6e25c2cd`

`key_datas:$telegram$2*100000*c87d4f7a65112524d6c6ba36957d7a1a6dabe3b74d21b44399862f531c7144c4*01f117f6d3eed44bbef06f367a2d30eaf2ca2147ab3a3f41beb92765e79472e5ee009b631406aaec6d6e5f6a723cc7527c66fd1cc8b1cf8af4c0f60bf370a7cf8a26260d904ea7728f2db61fbe44cd82b4e3f1716739b0f689370b801aa3dcb7cdd52eaa696fcb617431f43bcb0dbef0db72a9f9786c2b451efd52775077067cc86a0adf471f6907d024b6b3def90ed5cd974eb5528323192b99ef1fffb9988ade5cc53f671daab54a5b495d397ad0c5d199ccd438ff4a5af80bc554d96ad436096d628a14a9346a64c9a22ba6cc11ede546c6002d48a676822f7fd7f84bf21a94204703f3db9d67961a6d2474856c4cce909b38e66b0cc1540983d8f8b8e249347395f7f798bdc2f96b6c79b8cd676ce55bc1f264b4f450373993e4af05c846`

this 2 hash telegram does not find the password

```
john --wordlist=/usr/share/wordlists/nmap.lst 1.txt
No password hashes loaded (see FAQ)
```
```
john 1.txt --mask=test
No password hashes loaded (see FAQ)
```

PR #4420, opened 2020-10-29
optional PyCrypto for telegram2john

philsmd:
This idea / suggestion came up [here](https://github.com/openwall/john/issues/4387#issuecomment-716824145) .

I think it's an acceptable solution for letting the user run it without the `PyCrypto` python package installed, with the additional *noticable* warnings about the possibility of empty-pass hashes etc.

This makes it a little bit user-friendly, but still encourages the user to install the (now loose) dependency `PyCrypto`.

Thx

### DIFF
--- a/run/telegram2john.py
+++ b/run/telegram2john.py
@@ -50,11 +50,14 @@ import xml.etree.ElementTree as ET
 import struct
 import hashlib
 
+check_empty_pass = True
+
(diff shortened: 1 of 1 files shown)


ghost (approved):
LGTM

ghost:
Thanks.

PR #4421, opened 2020-10-29
update credits in telegram2john

philsmd:
There was this suggestion [here](https://github.com/openwall/john/issues/4387#issuecomment-716824145) that the comments at the start of the `telegram2john.py` script should be updated with the date and the credits for my contribution.

This little commit should make it very clear when the `telegram2john.py` script was updated (by me). I think this is enough even though there was a suggestion that I should be the new copyright holder... but I still think it's just a improved/refactored/newer update of the old script, so the license terms and holder remains the same (in my opinion).

Thx

### DIFF
--- a/run/telegram2john.py
+++ b/run/telegram2john.py
@@ -22,14 +22,17 @@ file(s) and from Telegram Desktop's local storage (map/key_datas) files"""
 # Special thanks goes to https://github.com/MihaZupan for documenting
 # this hashing scheme.
 #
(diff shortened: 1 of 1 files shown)


ghost (approved):

solardiz:
Thanks, @philsmd!

> there was a suggestion that I should be the new copyright holder

Not "the", but "a" - along with Dhiru for his original code. I thought that both of you would be copyright holders, and the license terms would in fact remain the same.

Issue #4393, opened 2020-10-01
bcrypt-opencl bugs


magnumripper:
bcrypt-opencl shares CPU-side code with bcrypt, so this issue should be looked into only after fixing #4388 and #4390. It really looks like it may be a single bug ending up in all these issues.

bcrypt-opencl show similar results as the CPU format, with the test-vector "./jtrts.pl -internal" tests:
```
$ rm -f ../run/john.pot && ../run/john --format=bcrypt-opencl BF_hashes.txt -w:BF_TS_wordlist.txt 
Device 1: GeForce RTX 2080 Ti
LWS=8 GWS=4096
Loaded 13 password hashes with 6 different salts (2.2x same-salt boost) (bcrypt-opencl [Blowfish OpenCL])
Cost 1 (iteration count) is 32 for all loaded hashes
Press 'q' or Ctrl-C to abort, almost any other key for status
Warning: Only 12 candidates left, minimum 4096 needed for performance.
U*U              (vec_0)
U*U*             (vec_1)
                 (vec_3)
                 (vec_3)
U*U*U            (vec_13)
U*U*U            (vec_2)
�                (vec_5)
�                (vec_6)
�U�U�U�U�U�U�U�U�U�U�U�U�U�U�U�U�U�U�U�U�U�U�U�U�U�U�U�U�U�U�U�U�U�U�U�U (vec_10)
U��U��U��U��U��U��U��U��U��U��U��U��U��U��U��U��U��U��U��U��U��U��U��U�� (vec_12)
������           (vec_8)
ё               (vec_7)
12g 0:00:00:00 DONE (2020-10-02 00:58) 60.00g/s 60.00p/s 360.0c/s 780.0C/s Dev#1:31°C 123456789012345678901234567890123456789012345678901234567..U*U*U
Warning: passwords printed above might not be all those cracked
Use the "--show" option to display all of the cracked passwords reliably
Session completed. 
```
Unlike the CPU format it cracks "12 of 13" hashes. It fails to crack `vec_4` and `vec_9`. And **just like CPU format, it cracks vec_3 twice**.
```
$ wc -l ../run/john.pot 
11 ../run/john.pot

$ sort ../run/john.pot | uniq -d
$2a$05$CCCCCCCCCCCCCCCCCCCCC.7uG0VCzI2bS7j6ymqJi9CdcdxiRTWNy:
```
It only writes 11 of them to the pot file though, even though vec_3 is written twice to it (just like with the CPU format).
```
[...]

PR #4425, opened 2020-10-30
bcrypt-opencl FMT_TRUNC fix, and self-test enhancement

solardiz:
As proposed in https://github.com/openwall/john/issues/4393#issuecomment-719653329


solardiz:
I've tested that all other CPU and OpenCL formats (412+89) meet the new self-test requirement, and that it would have detected missed `FMT_TRUNC` on a format (I temporarily introduced that bug for the test).

PR #4427, opened 2020-11-01
Use OpenMP in DES_bs_cmp_all()

solardiz:
We should actually have this with SIMD inside the 3 crypt_all() functions (separately for descrypt, bsdicrypt, LM), but even doing it here provides good speedup on massively parallel devices such as Xeon Phi and high-end server CPUs, at the expense of occasional slight slowdown when running only 1 or 2 threads.

Knights Landing 7210 descrypt speedup from ~167M to ~210M+ (best run 218M) on `--test` "Many salts" when running with explicit `GOMP_CPU_AFFINITY=0-255 OMP_NUM_THREADS=254`.

2x E5-2670 descrypt speedup from ~71M to ~73M on `--test` "Many salts" when running with explicit `GOMP_CPU_AFFINITY=0-31`.

The `n >= 96` check corresponds to 3+ threads most of the time, 6+ threads on AVX2 builds (only). Unfortunately, we don't fully avoid the OpenMP overhead when that condition isn't met - we'd have to have a separate copy of the entire loop for that. (We do indeed fully avoid the overhead on non-OpenMP builds.)

### DIFF
--- a/src/DES_bs.c
+++ b/src/DES_bs.c
@@ -483,7 +483,13 @@ int DES_bs_cmp_all(uint32_t *binary, int count)
 #if DES_bs_mt
 	int t, n = (count + (DES_BS_DEPTH - 1)) / DES_BS_DEPTH;
 #endif
+	int retval = 0;
 
+#if defined(_OPENMP) && DES_BS_VECTOR
+#pragma omp parallel for if(n >= 96) default(none) private(value, mask, bit, b, depth, t) shared(n, DES_bs_all_p, retval, binary)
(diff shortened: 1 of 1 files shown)


solardiz:
Knights Landing 7210 descrypt actual cracking goes at ~195M with ~1.5 hashes per salt (so sometimes more than one `cmp_all` call per `crypt_all`), with the same `GOMP_CPU_AFFINITY=0-255 OMP_NUM_THREADS=254` I had used for `--test`.

PR #4432, opened 2020-11-02
cracker.c: Unconditionally BLOB_FREE in crk_remove_hash()

magnumripper:


solardiz (approved):
Looks good to me. I hope @magnumripper actually tested this with a blob-using format.

magnumripper:
Pretty heavy test with WPAPSK, although I was quite confident now that I temporarily groked the difference between actually dropping the binary from the list and merely setting it to NULL.

magnumripper:
On a side note I was contemplating the possibility of some future FMT_REMOVE format actually wanting the blobs intact as well. I couldn't think of any reason for that so it's simply not supported. If ever needed, we'll simply change it.

PR #4439, opened 2020-11-05
Update applenotes2john.py

Banaanhangwagen:
I made the following updates:
- python3 only
- clean-up
- fix: added `continue` when _salt_ was found, but no _phash_

thank you

### DIFF
--- a/run/applenotes2john.py
+++ b/run/applenotes2john.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 # Script to extract "hashes" from password protected Apple Notes databases.
 #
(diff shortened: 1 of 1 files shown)


ghost:
> * python3 only

Probably JtR users prefer Python 3 compatibility coupled with the possibility of using python 2.

> * fix: added `continue` when _salt_ was found, but no _phash_

When/why it happens?

BTW: I found no samples. Could you create Apple Note samples files and add them to the wiki (https://openwall.info/wiki/john/sample-non-hashes)?

Banaanhangwagen:
> 
> Probably JtR users prefer Python 3 compatibility coupled with the possibility of using python 2.

It's time for those users to move to python3 then, because since January 1st 2020 python2 is no longer supported ;-)

>
> When/why it happens?

when _salt_ was found, but no _phash_ in the NoteStore.sqlite, the script gave the following error `TypeError: a bytes-like object is required, not 'NoneType'`

> BTW: I found no samples. Could you create Apple Note samples files and add them to the wiki (https://openwall.info/wiki/john/sample-non-hashes)?

Unfortunately, I cannot share the file itself;
Here is a screenshot :
image

Sample output of my fix (redacted):
```
./applenotes2john.py NoteStore.sqlite
$ASN$*3*20000*a9845f4471e7d6f13495145ccb17fc5e*ERROR:::::
$ASN$*55*20000*a9845f4471e7d6f13495145ccb17fc5e*2e712ed8e8fd030b0182650d4858f0ab3abb44XXXXXXXXXX:::::
```

solardiz:
I certainly view going Python 3 only as a drawback. If we can easily avoid that, we should.

> since January 1st 2020 python2 is no longer supported ;-)

I think some LTS distros with Python 2 remain supported after that date, and anyhow will be in use after that date.

ghost:
> I think some LTS distros with Python 2 remain supported after that date, and anyhow will be in use after that date.

This is Ubuntu 20.04, a clean install. The proposed change will cause problems for users with little computer knowledge.
```
$ python --version
Python 2.7.18

$ python3 --version
Python 3.8.5

$ python applenotes2john.py file
  File "applenotes2john.py", line 37
    print(f'$ASN$*{iden}*{iterations}*{salt}*ERROR:::::{hint}')
                                                             ^
SyntaxError: invalid syntax
```

solardiz:
What's the point in outputting a string that contains the salt but no hash? Perhaps that should be skipped.

I'd like a PR with _only_ the required fix to properly support the `NoteStore.sqlite` where the first entry with a salt doesn't have a hash, skipping until an entry that does. No other changes in that PR, please (not dropping Python 2). Thanks!

solardiz:
@Banaanhangwagen Are you going to send us a PR, or revise this one, addressing this feedback you received? (Don't print the hash-less string, keep Python 2 support)? Thanks!

Banaanhangwagen:
Reverted the changes I made;
only kept this in order to avoid the error `TypeError: b2a_hex() argument 1 must be string or buffer, not None`
```
        if phash is None:
            continue
```

solardiz:
Thanks @Banaanhangwagen! I've just "squash and merged" this, providing an elaborate and hopefully correct commit message.

PR #4442, opened 2020-11-09
descrypt, LM: Compare against loaded hashes in crypt_all()

solardiz:
I am posting this PR to validate these changes with our CI before I spend more of my time on them. I haven't completed my own testing yet, and might push more commits in here even if all tests pass.

We only do this when there's no bitmap for the current salt since we would otherwise not make proper use of the bitmap.

This is an initial implementation, which isn't fully optimized yet (e.g., cmp_all() has the first few iterations of the per-bit loop unrolled, but we don't have the same here yet).  It's also currently limited to just descrypt (but would also be relevant for LM and bsdicrypt), and to x86-alikes with SSE4.1/AVX, AVX2, or MIC/AVX-512.

Further, we might want to adjust the loader's threshold so that when we do have this new functionality compiled in, a bitmap would only be created for higher hash per salt counts than before.  Such adjustment hasn't been made yet, meaning that this new functionality is made use of in fewer cases than optimal.  Moreover, the loader's threshold isn't yet aware of our use of OpenMP here and in cmp_all() vs. our non-use of OpenMP in cracker.c's loop that uses the bitmap.

### DIFF
--- a/src/DES_bs.c
+++ b/src/DES_bs.c
@@ -497,8 +497,8 @@ int DES_bs_cmp_all(uint32_t *binary, int count)
 
 		mask = b[0] START ^ -(value & 1);
 		mask |= b[1] START ^ -((value >> 1) & 1);
-		if (mask == ~(ARCH_WORD)0) goto next_depth;
 		mask |= b[2] START ^ -((value >> 2) & 1);
+		if (mask == ~(ARCH_WORD)0) goto next_depth;
 		mask |= b[3] START ^ -((value >> 3) & 1);
 		if (mask == ~(ARCH_WORD)0) goto next_depth;
 		value >>= 4;
--- a/src/DES_bs.h
+++ b/src/DES_bs.h
@@ -154,12 +154,12 @@ extern void DES_bs_crypt(int count, int keys_count);
 /*
  * A simplified special-case implementation: 12-bit salts, 25 iterations.
  */
-extern void DES_bs_crypt_25(int keys_count);
+extern int DES_bs_crypt_25(int *pcount, struct db_salt *salt);
 
 /*
  * Another special-case version: a non-zero IV, no salts, no iterations.
  */
-extern int DES_bs_crypt_LM(int *keys_count, struct db_salt *salt);
+extern int DES_bs_crypt_LM(int *pcount, struct db_salt *salt);
 
 /*
  * Converts an ASCII ciphertext to binary to be used with one of the
--- a/src/DES_bs_b.c
+++ b/src/DES_bs_b.c
@@ -1,6 +1,6 @@
 /*
  * This file is part of John the Ripper password cracker,
- * Copyright (c) 1996-2001,2003,2010-2013,2015,2019 by Solar Designer
+ * Copyright (c) 1996-2001,2003,2010-2013,2015,2019,2020 by Solar Designer
  *
  * Addition of single DES encryption with no salt by
  * Deepika Dutta Mishra <dipikadutta at gmail.com> in 2012, no
@@ -13,6 +13,7 @@
 
 #include "arch.h"
 #include "common.h"
+#include "formats.h"
 #include "DES_bs.h"
 
 #if DES_BS_ASM && defined(_OPENMP) && defined(__GNUC__)
@@ -141,6 +142,9 @@ typedef __m512i vtype;
 #define vshr(dst, src, shift) \
 	(dst) = _mm512_srli_epi32((src), (shift))
 
+#define vtestallones(src) \
+	(_mm512_cmpeq_epi32_mask((src), vones) == 0xffff)
+
 #ifdef __AVX512F__
 #define vsel(dst, a, b, c) \
 	(dst) = _mm512_ternarylogic_epi32((b), (a), (c), 0xE4)
@@ -172,6 +176,9 @@ typedef __m256i vtype;
 	(dst) = _mm256_slli_epi64((src), (shift))
 #define vshr(dst, src, shift) \
 	(dst) = _mm256_srli_epi64((src), (shift))
+
+#define vtestallones(src) \
+	_mm256_testc_si256((src), vones)
 
 #elif defined(__SSE2__) && DES_BS_DEPTH == 128
 #ifdef __AVX__
@@ -213,6 +220,15 @@ typedef __m128i vtype;
 	(dst) = _mm_slli_epi64((src), (shift))
 #define vshr(dst, src, shift) \
 	(dst) = _mm_srli_epi64((src), (shift))
+
+#ifdef __SSE4_1__
+#include <smmintrin.h>
+#define vtestallones(src) \
+	_mm_testc_si128((src), vones)
+#else
+#define vtestallones(src) \
+	(_mm_movemask_epi8(_mm_cmpeq_epi32((src), vones)) == 0xffff)
+#endif
 
 #elif defined(__MMX__) && ARCH_BITS != 64 && DES_BS_DEPTH == 64
 #include <mmintrin.h>
@@ -263,6 +279,19 @@ typedef unsigned ARCH_WORD vtype;
 	(dst) = (src) << (shift)
 #define vshr(dst, src, shift) \
 	(dst) = (src) >> (shift)
+
+#ifdef _OPENMP
+/*
+ * When using OpenMP, it's desirable to compare hashes inside crypt_all() even
+ * in scalar builds, so that we do this from the same parallel section with
+ * computation (saving on thread synchronization) and with data still in caches
+ * of the same CPUs.  Without OpenMP, the code in cmp_all() is about as good,
+ * and that interface provides per-loaded-hash information to the caller, so no
+ * point in pre-checking in crypt_all() first.
+ */
+#define vtestallones(src) \
+	((src) == vones)
+#endif
 
 /* Assume that 0 always fits in one load immediate instruction */
 #undef vzero
@@ -685,14 +714,25 @@ void DES_bs_set_salt(ARCH_WORD salt)
 #define y(p, q) vxorf(*(vtype *)&b[p] bd, *(vtype *)&k[q] kd)
 #define z(r) ((vtype *)&b[r] bd)
 
-void DES_bs_crypt_25(int keys_count)
+int DES_bs_crypt_25(int *pcount, struct db_salt *salt)
 {
+	int keys_count = *pcount;
 #if DES_bs_mt
 	int t, n = (keys_count + (DES_BS_DEPTH - 1)) / DES_BS_DEPTH;
 #endif
+#ifdef vtestallones
+#ifdef _OPENMP
+	volatile
+#endif
+	int precheck = salt && !salt->bitmap;
+#endif
 
 #ifdef _OPENMP
-#pragma omp parallel for default(none) private(t) shared(n, DES_bs_all_p, keys_count)
+#ifdef vtestallones
+#pragma omp parallel for default(none) private(t) shared(n, DES_bs_all_p, salt, precheck, benchmark_running, self_test_running)
+#else
+#pragma omp parallel for default(none) private(t) shared(n, DES_bs_all_p)
+#endif
 #endif
 	for_each_t(n) {
 #if DES_BS_EXPAND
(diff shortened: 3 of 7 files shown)


solardiz:
This makes the speeds more stable, like this:

```
[solar@knightslanding src]$ OMP_NUM_THREADS=254 GOMP_CPU_AFFINITY=0-255 ../run/john -test -form=descrypt
Will run 254 OpenMP threads
Benchmarking: descrypt, traditional crypt(3) [DES 512/512 AVX512F]... (254xOMP) DONE
Warning: "Many salts" test limited: 214/256
Many salts:     222642K c/s real, 876285 c/s virtual
Only one salt:  19571K c/s real, 387860 c/s virtual

[solar@knightslanding src]$ OMP_NUM_THREADS=254 GOMP_CPU_AFFINITY=0-255 ../run/john -test -form=descrypt
Will run 254 OpenMP threads
Benchmarking: descrypt, traditional crypt(3) [DES 512/512 AVX512F]... (254xOMP) DONE
Warning: "Many salts" test limited: 211/256
Many salts:     219521K c/s real, 864017 c/s virtual
Only one salt:  19571K c/s real, 381460 c/s virtual

[solar@knightslanding src]$ OMP_NUM_THREADS=254 GOMP_CPU_AFFINITY=0-255 ../run/john -test -form=descrypt
Will run 254 OpenMP threads
Benchmarking: descrypt, traditional crypt(3) [DES 512/512 AVX512F]... (254xOMP) DONE
Warning: "Many salts" test limited: 210/256
Many salts:     218480K c/s real, 862775 c/s virtual
Only one salt:  19668K c/s real, 391741 c/s virtual

[solar@knightslanding src]$ OMP_NUM_THREADS=254 GOMP_CPU_AFFINITY=0-255 ../run/john -test -form=descrypt
Will run 254 OpenMP threads
Benchmarking: descrypt, traditional crypt(3) [DES 512/512 AVX512F]... (254xOMP) DONE
Warning: "Many salts" test limited: 209/256
Many salts:     216358K c/s real, 856097 c/s virtual
Only one salt:  19571K c/s real, 387252 c/s virtual

[solar@knightslanding src]$ OMP_NUM_THREADS=254 GOMP_CPU_AFFINITY=0-255 ../run/john -test -form=descrypt
Will run 254 OpenMP threads
Benchmarking: descrypt, traditional crypt(3) [DES 512/512 AVX512F]... (254xOMP) DONE
Warning: "Many salts" test limited: 214/256
Many salts:     222642K c/s real, 877252 c/s virtual
Only one salt:  19571K c/s real, 378901 c/s virtual

[solar@knightslanding src]$ OMP_NUM_THREADS=254 GOMP_CPU_AFFINITY=0-255 ../run/john -test -form=descrypt
Will run 254 OpenMP threads
Benchmarking: descrypt, traditional crypt(3) [DES 512/512 AVX512F]... (254xOMP) DONE
Warning: "Many salts" test limited: 211/256
Many salts:     218428K c/s real, 865977 c/s virtual
Only one salt:  19571K c/s real, 381093 c/s virtual
```

and even without affinity:

```
[solar@knightslanding src]$ OMP_NUM_THREADS=254 ../run/john -test -form=descrypt
Will run 254 OpenMP threads
Benchmarking: descrypt, traditional crypt(3) [DES 512/512 AVX512F]... (254xOMP) DONE
Warning: "Many salts" test limited: 212/256
Many salts:     220561K c/s real, 871095 c/s virtual
Only one salt:  19379K c/s real, 363837 c/s virtual

[solar@knightslanding src]$ OMP_NUM_THREADS=254 ../run/john -test -form=descrypt
Will run 254 OpenMP threads
Benchmarking: descrypt, traditional crypt(3) [DES 512/512 AVX512F]... (254xOMP) DONE
Warning: "Many salts" test limited: 214/256
Many salts:     222642K c/s real, 878792 c/s virtual
Only one salt:  19475K c/s real, 363837 c/s virtual
```

With OpenMP in `cmp_all` (which is now usually skipped), speeds in tests like the above typically varied between 210M and 218M (with affinity; worse without). Now it's 216M to 222M - higher and more stable.

solardiz:
202M actual cracking with:

```
[solar@knightslanding run]$ OMP_NUM_THREADS=254 GOMP_CPU_AFFINITY=0-255 ../run/john -form=descrypt -inc -min-len=7 -max-len=7 ~/pw-fake-unix
Loaded 3269 password hashes with 2243 different salts (1.5x same-salt boost) (descrypt, traditional crypt(3) [DES 512/512 AVX512F])
Remaining 3053 password hashes with 2139 different salts (1.4x same-salt boost)
Will run 254 OpenMP threads
[...]
0g 0:00:00:54  0g/s 77051p/s 202412Kc/s 288494KC/s pentlcd..lia1655
diamond          (u45-des)
1g 0:00:00:56  0.01785g/s 74299p/s 202503Kc/s 289152KC/s pentlcd..lia1655
1g 0:00:00:57  0.01754g/s 72983p/s 202674Kc/s 288752KC/s pentlcd..lia1655
dominic          (u2998-des)
maxwell          (u1200-des)
dreamer          (u378-des)
deborah          (u1933-des)
5g 0:00:00:59  0.08471g/s 70510p/s 202292Kc/s 287921KC/s pentlcd..lia1655
```

Edit: before the changes here (but with OpenMP in `cmp_all` already, which is #4427), this was ~195M.

I previously ran it until the 1 minute mark at average speed of 180M - that was lower because processing of correct guesses is relatively unoptimized. With these changes, we optimize even more for the typical case of no guess.

solardiz:
@magnumripper CI is happy so far, but I want more-tests. Maybe if you approve this, that will run?

magnumripper (approved):

magnumripper:
We just use to launch them manually. I think they are running now (even before I approved)

solardiz:
All tests passed, but I've just force-pushed support for SSE2 and scalar builds (the latter enabled in OpenMP builds only, for reasons explained in the added comment).

solardiz:
Force-pushed again, adding `#include <smmintrin.h>` for builds with SSE4.1 but without AVX. Our CI didn't detect this was missing, but my own manual testing eventually did.

We still don't report SSE4.1 for descrypt, though - we report SSE2 in such builds. I am not fixing this yet, and not sure we should - it's just one instruction used in one place, and not always reached.

solardiz:
FWIW, here's simulation of our bitslice `cmp_all` (and the equivalent algorithm introduced into `crypt_all` with this PR):

```c
#include <stdio.h>
#include <unistd.h>

#define N 1000000

typedef unsigned long t;

static unsigned int dist[20];

int main(void)
{
	unsigned int i;

	for (i = 0; i < N; i++) {
		unsigned int n = 0;
		t mask = 0, v;
		do {
			if (read(0, &v, sizeof(v)) == sizeof(v)) {
				mask |= v;
				n++;
//				printf("%016llx\n", (unsigned long long)mask);
			}
		} while (~mask);
		if (n < sizeof(dist) / sizeof(dist[0]))
			dist[n]++;
	}

	for (i = 1; i < sizeof(dist) / sizeof(dist[0]); i++)
		printf("%d: %.2f%%\n", i, dist[i] * 100. / N);

	return 0;
}
```

Running it with 32-bit integers gives these probabilities of rejection by iteration number:

```
1: 0.00%
2: 0.01%
3: 1.39%
4: 11.31%
5: 23.52%
6: 24.27%
7: 17.40%
8: 10.38%
9: 5.65%
10: 2.99%
11: 1.52%
12: 0.77%
13: 0.39%
14: 0.20%
15: 0.10%
16: 0.05%
17: 0.03%
18: 0.01%
19: 0.01%
```

For 64-bit:

```
1: 0.00%
2: 0.00%
3: 0.02%
4: 1.57%
5: 11.53%
6: 23.36%
7: 23.98%
8: 17.31%
9: 10.41%
10: 5.70%
11: 2.99%
12: 1.56%
13: 0.77%
14: 0.39%
15: 0.20%
16: 0.10%
17: 0.05%
18: 0.02%
19: 0.01%
```

So there's no point in checking for possible rejection until after 3rd or 4th iteration. For SIMD, this shifts to even higher iteration numbers.

solardiz:
I tried to optimize this by unrolling the first few iterations of the loop and skipping unneeded checks in there, and/or by avoiding conditional branches on `binary` bits (instead indexing a 2-element array of `{vzero, vones}` and XOR'ing with that). However, almost any change resulting in even slight code size increase appeared to slow things down in my testing (on Sandy Bridge and Knights Landing). So I ended up force-pushing only a minor revision of my previous commit.

solardiz:
I've also started experimenting with similar changes for LM, and wasn't getting any speedup at first - in fact, usually a slowdown. Turns out that for descrypt our `--test` benchmark calls `crypt_all` with a list of 3 hashes for the salt, and for LM with a list of 10. So with these changes the benchmark was doing more work than before, as previously the benchmark only called `cmp_all` once, not once per hash in the salt's list like real cracking would. It's interesting that for descrypt I was getting a consistent speedup apparently despite of doing 3x more useful work here.

I am now considering this:

```c
                        } while ((pw = pw->next) && (!benchmark_running || self_test_running));
```

So real cracking and self-test would use all hashes for the salt, but benchmark would use only one (mimic having one hash per salt, like it did before). I've already tested this hack - seems to work as intended. Of course, it has very slight performance impact on real cracking (having to check the `benchmark_running` variable). Maybe we should instead patch the `two_salts_db` entry in the benchmarking code?

Another related observation is: our benchmarking code always calls `cmp_all`, even when `crypt_all` returns 0, but passes the 0 into `cmp_all` (so it will generally return right away). Our real cracking code skips `cmp_all` in that case. Perhaps we should adjust the benchmark in this respect as well.

solardiz:
I've just force-pushed my latest attempts at optimizing this for descrypt and LM.

> Maybe we should instead patch the `two_salts_db` entry in the benchmarking code?

This is now #4450.

> Another related observation is: our benchmarking code always calls `cmp_all`, even when `crypt_all` returns 0, but passes the 0 into `cmp_all` (so it will generally return right away). Our real cracking code skips `cmp_all` in that case. Perhaps we should adjust the benchmark in this respect as well.

I've included a commit fixing this in here.

solardiz:
On AWS c5.24xlarge before November:

```
Benchmarking: descrypt, traditional crypt(3) [DES 512/512 AVX512F]... (96xOMP) DONE
Many salts:     561512K c/s real, 5906K c/s virtual
Only one salt:  85685K c/s real, 1415K c/s virtual
```

After Nov 1 commit 645a378764d19a78f027e2c3b2e31a11696a944c:

```
Benchmarking: descrypt, traditional crypt(3) [DES 512/512 AVX512F]... (96xOMP) DONE
Many salts:	831258K c/s real, 8784K c/s virtual
Only one salt:	88033K c/s real, 1417K c/s virtual
```

With this PR:

```
Benchmarking: descrypt, traditional crypt(3) [DES 512/512 AVX512F]... (96xOMP) DONE
Many salts:	861929K c/s real, 9087K c/s virtual
Only one salt:	90046K c/s real, 1416K c/s virtual
```

solardiz:
Turns out higher `DES_bs_cpt` would (now) often produce better speeds.

`DES_bs_cpt` set to 128, code introduced to limit the effective `cpt` such that the allocation size is within 2 GB (things break when it isn't as we use `int` in some places), resulting in actual `cpt` of 115 when running 254 threads on the Knights Landing:

```
Benchmarking: descrypt, traditional crypt(3) [DES 512/512 AVX512F]... (254xOMP) DONE
Warning: "Many salts" test limited: 35/256
Many salts:     261721K c/s real, 1030K c/s virtual
Only one salt:  20675K c/s real, 635593 c/s virtual
```

So ~220M to ~260M when going from `cpt` 16 to 115. With this, `max_keys_per_crypt` becomes almost 15M, so it's approx. 17 `crypt_all()` calls per second or 4 minutes to advance to next batch of candidates when running against all 4096 salts.

I'm not yet sure what to do about this, as it has drawbacks. Also, it is possible that a lower `cpt` would be optimal once we introduce mask mode support into `crypt_all()`.

solardiz:
cpt 115: 248M to 262M across several runs
cpt 96: 242M to 256M
cpt 64: 239M to 251M
cpt 56: 237M to 248M (outlier: 232M)
cpt 48: 239M to 252M (outlier: 230M)
cpt 40: 235M to 248M
cpt 32: 231M to 243M
cpt 24: 223M to 236M
cpt 16: 211M to 223M

PR #4443, opened 2020-11-09
Handle rules given in UTF-8 when using internal codepage

### LINKED ISSUE #4437: Improve user experience for --internal-codepage option
https://www.openwall.com/lists/john-users/2020/11/03/3

We should detect when UTF-8 is used in rules (such as `seé`) and convert automagically. Mask mode already does such conversion from command-line option.

magnumripper:
If we detect that a command-line or config-file rule is encoded in UTF-8 while there's an internal legacy codepage in use, convert the rule as appropriate.  Closes #4437

### DIFF
--- a/doc/ENCODINGS
+++ b/doc/ENCODINGS
@@ -15,8 +15,8 @@ The new defaults (which can be changed in john.conf) are:
   * Output to screen, log and .pot file is UTF-8.
   * Target encoding for LM is CP850 (and input will be converted accordingly).
   * Internal encoding (eg. for rules processing) is disabled. For Western
-    Europe, ISO-8859-1 is recommended but since this has a performance impact
-    you'll need to use the command-line option or un-comment the line in
+    Europe, ISO-8859-1 is recommended but since this has a slight performance
+    impact you'll need to use the command-line option or un-comment the line in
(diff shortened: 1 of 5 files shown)


solardiz:
Curiously, this will probably log both the original and the converted rule in "accepted as" lines (a feature?), making the log file fit neither encoding (a bug?)

magnumripper:
> Curiously, this will probably log both the original and the converted rule in "accepted as" lines (a feature?), making the log file fit neither encoding (a bug?)

That's what I thought but for some reason it logs them as converted only
```
0:00:00:00 - Rule #1: '-U /a sa?' accepted as '/asa?'
```
Not sure why but I didn't care yet. What we could do is convert those log lines to the `\xe5` notation, and ideally more like we guessed - it would be quite nice like this:
```
0:00:00:00 - Rule #1: '-U /a saå' accepted as '/asa\xe5'
```

ghost (approved):
LGTM

magnumripper:
I now moved the conversion from `rules_reject()` to `rpp_process_rule()`. I wanted to find a place with least performance impact yet as little code as possible. We're modifying string in place, the result can only be smaller.

PR #4444, opened 2020-11-09
Add 'd' keypress for delayed status (after current batch)

magnumripper:
Make keypress 'd' special: It will also emit a status line, but not until the current batch has been ran against all salts.  Under some circumstances that make take a good while.

Some reasons to wanting this is you'll get a correct p/s figure as early as possible, or you can just get a feel for just how long each batch of candidates takes to process.

### DIFF
--- a/doc/NEWS
+++ b/doc/NEWS
@@ -141,6 +141,10 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
   reading status/quit keystrokes even if we're not the foreground process.
   [magnum; 2020]
 
+- Make keypress [d] special: It will also emit a status line, but not until
+  the current batch has been run among all salts.  Under some circumstances
+  that make take a good while.  [magnum; 2020]
+
 
 Major changes from 1.8.0-jumbo-1 (December 2014) to 1.9.0-jumbo-1 (May 2019):
 
--- a/src/signals.h
+++ b/src/signals.h
@@ -35,6 +35,7 @@ extern volatile int event_abort;	/* Abort requested */
 extern volatile int event_reload;	/* Reload of pot file requested */
 extern volatile int event_save;		/* Save the crash recovery file */
 extern volatile int event_status;	/* Status display requested */
+extern volatile int event_delayed_status;	/* Status display requested after current batch */
 extern volatile int event_ticksafety;	/* System time in ticks may overflow */
 #ifdef HAVE_MPI
 extern volatile int event_mpiprobe;	/* MPI probe for messages requested */
--- a/src/cracker.c
+++ b/src/cracker.c
@@ -1035,9 +1035,11 @@ static int crk_salt_loop(void)
 			break;
 	} while ((salt = salt->next));
 
-	if (crk_db->salt_count < sc && john_main_process &&
-	    cfg_get_bool(SECTION_OPTIONS, NULL, "ShowSaltProgress", 0))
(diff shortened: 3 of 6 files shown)


magnumripper:
Comments are welcome. Perhaps we should emit something similar to the "`Wait...\r`" printed during Ctrl-C processing, except it could say something like "`Delayed status pending...\r`"?

Not quite sure we should add this to the "`Press 'q' to abort, almost any other...`" message at all, but maybe we should.

Note that if you have gazillions of salts loaded and press 'd', while waiting for that delayed status you can still press "any" (other) key to get a quick out-of-sync status and the delayed one will still be pending.

doc/NEWS
```diff
@@ -132,6 +132,10 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
 
 - Add support for --format=LIST  [magnum; 2020]
 
+- Make keypress [d] special: It will also emit a status line, but not until
+  the current batch has been ran among all salts.  Under some circumstances
+  that make take a good while.
```
solardiz: s/ran/run/ and add "[magnum; 2020]"

src/signals.h
```diff
@@ -35,6 +35,7 @@ extern volatile int event_abort;	/* Abort requested */
 extern volatile int event_reload;	/* Reload of pot file requested */
 extern volatile int event_save;		/* Save the crash recovery file */
 extern volatile int event_status;	/* Status display requested */
+extern volatile int event_delayed_status;	/* Status display requested after current batch*/
```
solardiz: Missing space before the "*/".

solardiz (approved):
This is a somewhat obscure feature, but it's simple enough in code that I don't mind having it. We should probably have these various keypresses in some documentation file besides `NEWS`.

Looking at the implementation, I guess this keypress won't work during "single crack"? Should it?

And yes, printing some message that the keypress was detected and the action is pending is a good idea.

solardiz:
I was also thinking of adding some keypress for detailed multi-line status, which would include much more information than we include on the current status line (perhaps also hash type, the number of remaining hashes and salts, what attack is running, etc. - this stuff is useful when running so many different instances of `john` that it's easy to get confused, which is actually pretty common). Maybe "d" would be right for "detailed", but we can also make it "m" for "multi-line" or "f" for "full", etc. Or/and we can have separate keys for the different single-line components instead of one multi-line status screen, although the latter is valuable for it being more consistent (all data pertaining to almost the same moment).

Edit: this raises the question: would we also want delayed versions of those alternate status reports?

magnumripper:
> I was also thinking of adding some keypress for detailed multi-line status (...)

I actually have a working experimental patch for just that but it's a bit intrusive as of now. It lists command line options, "salt x/y", "rule x/y" and other stuff. I also wanted it to show start (wall clock) time (perhaps along with last resume wall clock time) but it's not there yet. The thing is nearly everything I want to add clutters things up with global variables or getter/setter functions so it's not easy to make it very tidy.

> Maybe "d" would be right for "detailed", but we can also make it "m" for "multi-line" or "f" for "full", etc.

My experimental code has "s" for detailed status but I'm open for anything.

> Edit: this raises the question: would we also want delayed versions of those alternate status reports?

I had the idea that pressing "d" and then "s" would produce a delayed verbose status but I'm not sure that's the best idea

magnumripper:
> Looking at the implementation, I guess this keypress won't work during "single crack"? Should it?

I'm not sure where to put it. Would we still want it to be after all salts was (in some way) processed, or is there some better waypoint to use?

PR #4445, opened 2020-11-09
Improve "stdin vs. keyboard vs. message"

### LINKED ISSUE #4403: Keyboard input for status/quit vs. stdin or subshell
Three ways of using `--stdin` (or `--pipe`):

### 1. stdin is our terminal:
```
$ ./john hashfile -stdin
(...)
```
The above works as expected: Obviously the status/quit keys does not work and the message reflects that.

### 2. stdin is a file or fifo:
```
$ ./john hashfile -stdin < some.file
(...)
```
Above, the status/quit keys **do** work, so the message is... well it's correct but it should ideally be "Press 'q' or Ctrl-C to abort, almost any other key for status".

### 3. stdin is a pipe:
```
$ cat some.file | ./john hashfile -stdin
(...)
```
Here, status/quit keys do not work - but *why don't they*? Can we fix that or is it nigh impossible in this situation?

### 4. Related issue:
Not using -stdin, but running via some front-end script or program. Example:
```
$ cat front-end.sh
#!/bin/sh
../run/john $*

$ ./front-end.sh hashfile -inc
(...)
```
Status/quit keys do **not** work so the message is not correct! Can we get them to work in this situation? Failing that, can we get the message right? Or could we amend that script in some way for letting john have the keyboard for status/quit commands?

magnumripper:
Add option --force-keys for enabling it even if we're not the foreground process, for use with pipes and inside front-end scripts. Ensure the "Press 'q' to abort (...)" message really reflects whether we have the terminal set up for reading keystrokes or not.

Closes #4403

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -290,6 +290,7 @@ static struct opt_entry opt_list[] = {
 		~FLG_MASK_CHK & ~OPT_REQ_PARAM, "%d", &benchmark_time},
 	{"tune", FLG_ZERO, 0, 0, OPT_REQ_PARAM,
 		OPT_FMT_STR_ALLOC, &options.tune},
+	{"force-keys", FLG_FORCE_KEYS},
 	{NULL}
 };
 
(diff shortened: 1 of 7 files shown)


src/options.c
```diff
@@ -441,7 +442,9 @@ JOHN_USAGE_FORK \
 "--input-encoding=NAME      input encoding (alias for --encoding)\n" \
 "--internal-codepage=NAME   codepage used in rules/masks (see doc/ENCODINGS)\n" \
 "--target-encoding=NAME     output encoding (used by format, see doc/ENCODINGS)\n" \
-"--tune=HOW                 tuning options (auto/report/N)\n"
+"--tune=HOW                 tuning options (auto/report/N)\n" \
+"--force-keys               set up terminal for reading keystrokes even if we're\n" \
+"                           not the foreground process"
```
magnumripper: Perhaps we should group our huge list of options options better. Maybe first really important things like --wordlist and --incremental, then lots of random options in alphabetical order (either sorted by option name, or by description), then the rest (optional OpenCL etc) and lastly the --format

ghost (approved):
Looks good for broad testing

PR #4448, opened 2020-11-10
Enhance --salt option for optionally loading n best salts

### LINKED ISSUE #4407: --salts=N option
Ref: https://www.openwall.com/lists/john-users/2020/10/15/2

I've always wanted the `--salts=[-]COUNT[:MAX]` option to be "load the `COUNT[:MAX]` most duped salts" (with counts being before looking at pot file(s)) as opposed to "load only salts with `COUNT[:MAX]` dupes". I think I know how to implement it so the question is: Could I possibly change that option's behavior, or should I add a brand new option?

BTW we're out of options flags and Carthago delenda est. I have absolutely no idea how to tackle that, except completely ignoring that mutually exclusive options are given and just let it crash sooner or later.

magnumripper:
```
--salts=#M[-N]			load M [to N] most populated salts
```

Alternative #M syntax:  Instead of requiring a certain number of hashes per salt, we simply say "load the most re-used salt" or "load the M to N most re-used salts".

Regardless of old or new syntax, the re-use counts are considered before sorting out already cracked hashes, so a resume or a later attack using same --salts option will pick the same set. (Edit: this fact is not new)

Closes #4407

### DIFF
--- a/doc/NEWS
+++ b/doc/NEWS
@@ -148,6 +148,9 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
 - Add command-line option --loader-dupecheck=<yes/no>, overriding config file.
   Mark the latter as deprecated in config file comments.  [magnum; 2020]
 
+- Add alternative syntax to --salts option for loading "N most populated
+  salts" as opposed to "salts having at least N hashes"  [magnum; 2020]
+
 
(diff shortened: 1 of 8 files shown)


ghost (approved):
Looks good for broad testing

magnumripper:
Added a separate commit for being able to say `--loader-dupecheck=BOOL` overriding any config file setting. It's related enough.

Issue #4219, opened 2020-03-04
Subsets mode final status line, initial ETA weirdness


solardiz:
```
$ ./john subsets=0123456789 -min-len=12 -max-len=12 a
[...]
0g 0:00:00:04 0.00% (12) (ETA: 2020-03-10 13:42) 0g/s 142490p/s 142490c/s 142490C/s 022101201001..202210000101
0g 0:00:00:05 0.00% (12) (ETA: 2020-03-02 22:45) 0g/s 142962p/s 142962c/s 142962C/s 033301311333..001303333313
0g 0:00:00:06 0.00% (12) (ETA: 2020-03-03 16:28) 0g/s 143117p/s 143117c/s 143117C/s 013303330101..001130303331
0g 0:00:00:07 0.00% (12) (ETA: 2020-03-04 10:11) 0g/s 143367p/s 143367c/s 143367C/s 013003003310..100030003330
0g 0:00:00:08 0.00% (12) (ETA: 2020-03-05 03:54) 0g/s 143554p/s 143554c/s 143554C/s 031100131110..100113110003
0g 0:00:00:09 0.00% (12) (ETA: 2020-03-03 01:55) 0g/s 143699p/s 143699c/s 143699C/s 144440001440..441040044004
0g 0:00:00:10 0.00% (12) (ETA: 2020-03-03 12:06) 0g/s 143672p/s 143815c/s 143672C/s 110144140041..044114141001
0g 0:00:00:34 0.01% (12) (ETA: 2020-03-02 21:25) 0g/s 144614p/s 144614c/s 144614C/s 040420244244..240440420440
0g 0:03:57:47 4.65% (12) (ETA: 2020-03-02 19:20) 0g/s 146992p/s 146992c/s 146992C/s 688618889691..691889686188
0g 0:08:07:26 9.59% (12) (ETA: 2020-03-02 18:52) 0g/s 147173p/s 147173c/s 147173C/s 626120664660..622400661666
0g 1:00:11:48 28.41% (12) (ETA: 2020-03-02 19:21) 0g/s 147380p/s 147380c/s 147380C/s 040347022472..034447027202
0g 1:13:18:06 43.91% (12) (ETA: 2020-03-02 19:08) 0g/s 147391p/s 147391c/s 147391C/s 408449044398..404938480944
0g 2:00:48:13 57.56% (12) (ETA: 2020-03-02 18:58) 0g/s 147449p/s 147449c/s 147449C/s 321133669296..323126616993
[...]

PR #4451, opened 2020-11-11
Subsets: Progress/ETA fixes

magnumripper:
Avoid saying we're on a length past max on session completion.
Avoid double status line on session completion. 

See #4219


ghost (approved):
LGTM

Issue #4452, opened 2020-11-12
Reconsider potcheck during build


solardiz:
The potcheck.pl script introduced in #2161 depends on Perl (obviously) and Digest::MD5. This makes build attempts fail when those are not both available and a `john.pot` is present, even when there are no entries in `john.pot` needing conversion. That's weird.

In fact, it's weird that we're doing anything with `john.pot` at build time. I think that instead of the dependency and build failure, we should give a hint that the file may need to be updated and how, but not actually do that. That hint may be printed at build time, or (perhaps more reasonably) when `john` is run and sees old entries needing an update (we'd need to introduce such logic into it).

And if we do introduce the detection logic into `john`, perhaps the update logic can be in there too, eliminating the dependencies entirely? Alternatively, has it possibly been enough time that we don't need to bother (but need to keep the script around for when it's still needed)?

solardiz:
Here's another idea: with detection logic in `john`, maybe it doesn't need to update the file on disk, but should simply update the entries in memory when it loads those? The script could remain as a way to permanently optimize old `john.pot`.

magnumripper:
> Alternatively, has it possibly been enough time that we don't need to bother (but need to keep the script around for when it's still needed)?

Yes, maybe.

> Here's another idea: with detection logic in john, maybe it doesn't need to update the file on disk, but should simply update the entries in memory when it loads those? 

Not sure why we didn't do so in the first place, with nearly all code already there. I guess I was afraid of pot loading performance penalty but it shouldn't be that bad - perhaps none at all.

I'll have a look at it

solardiz:
I am getting this even with empty `john.pot`:

```
checking if potcheck.pl succeeds... Can't locate Digest/MD5.pm in @INC (@INC contains: /usr/local/lib64/perl5 /usr/local/share/perl5 /usr/lib64/perl5/vendor_perl /usr/share/perl5/vendor_perl /usr/lib64/perl5 /usr/share/perl5 .) at ../run/potcheck.pl line 17.
BEGIN failed--compilation aborted at ../run/potcheck.pl line 17.
no
configure: error: NOTE, john will not run properly until you fix your ../run/john.pot file using '../run/potcheck.pl -longline_fix'
```

We should at least trivially limit this to non-empty `john.pot`.

magnumripper:
#4454 relaxes the need for Digest::MD5. As for not even having Perl, you'd not be able to use `./configure` then anyway but use the legacy Makefile.

[...]

PR #4454, opened 2020-11-13
potcheck.pl: Don't require Digest::MD5 unless actually needed

magnumripper:
The checks made by ./configure never needs Digest::MD5, only actual fixing (which you'd do manually) does.

See #4452

### DIFF
--- a/run/potcheck.pl
+++ b/run/potcheck.pl
@@ -14,7 +14,6 @@
 use warnings;
 use strict;
 use Getopt::Long;
-use Digest::MD5 qw(md5_hex);
 
 # NOTE, if this is changed in params.h, we need to update it here!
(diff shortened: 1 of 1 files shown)


solardiz (approved):
FWIW, the change from `use` to `require` and `import` is entirely correct given the move, per the documentation:

```
       Perl modules are included into your program by saying

           use Module;

       or

           use Module LIST;

       This is exactly equivalent to

           BEGIN { require Module; import Module; }

       or

           BEGIN { require Module; import Module LIST; }
```

magnumripper:
Yes, we use this a lot in [pass_gen.pl](https://github.com/openwall/john/blob/bleeding-jumbo/run/pass_gen.pl) which is dang near unusable without it...

PR #4457, opened 2020-11-15
Rename new option --force-keys to --force-tty

magnumripper:


solardiz (approved):
Please also rename `FLG_FORCE_KEYS`.

PR #4458, opened 2020-11-15
Loader dupecheck usage

magnumripper:
This fixes the blatant bug we had. Still, there's a few inconsistencies regarding usage blurb for bool options:

For long we've had:
```
--single-retest-guess=B    override config for SingleRetestGuess (bool: Y/N)
(...)
--bare-always-valid=Y      treat bare hashes as valid (Y/N)
```
and now we get
```
--loader-dupecheck=no      disable the dupe checking when loading hashes
```

Note that we use `N` for "number" as in
```
--min-length=N             request a minimum candidate length in bytes
```

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -427,7 +427,7 @@ JOHN_USAGE_REGEX \
 "--fix-state-delay=N        performance tweak, see doc/OPTIONS\n" \
 "--config=FILE              use FILE instead of john.conf or john.ini\n" \
 "--log-stderr               log to screen instead of file\n"             \
-"--loader-dupecheck=yes     disable the dupe checking when loading hashes\n" \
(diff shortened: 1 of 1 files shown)


magnumripper:
Not sure why the xcode bot-build dies (I re-ran it to no avail), but it certainly has nothing to do with this change. Please just merge this.

solardiz:
OK, I'll merge this, but longer term we need to clean up how we list the Booleans in options descriptions.

solardiz:
> Not sure why the xcode bot-build dies (I re-ran it to no avail)

It also failed for bleeding-jumbo with this merged, with:

```
An error occurred while generating the build script.
```

I hope it'll recover with a later merge in some days, but if it doesn't we might have to drop that check.

PR #4461, opened 2020-11-17
bench.c: Always benchmark with one hash per bucket

### LINKED ISSUE #4450: Benchmark passes multi-hash list into crypt_all()
Turns out that for descrypt our `--test` benchmark calls `crypt_all` with a list of 3 hashes for the salt, and for LM with a list of 10. If the format's implementation implements hash comparison in `crypt_all`, it ends up doing more work than the user running the benchmark likely expects: we previously benchmarked for 1 hash per salt. We should probably fix this by using 1-entry lists during benchmark, even if we use multi-entry lists during self-test.

magnumripper:
Closes #4450

### DIFF
--- a/src/DES_bs_b.c
+++ b/src/DES_bs_b.c
@@ -729,7 +729,7 @@ int DES_bs_crypt_25(int *pcount, struct db_salt *salt)
 
 #ifdef _OPENMP
 #ifdef vtestallones
-#pragma omp parallel for default(none) private(t) shared(n, DES_bs_all_p, salt, precheck, benchmark_running, self_test_running)
+#pragma omp parallel for default(none) private(t) shared(n, DES_bs_all_p, salt, precheck)
 #else
 #pragma omp parallel for default(none) private(t) shared(n, DES_bs_all_p)
 #endif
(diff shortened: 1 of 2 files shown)


magnumripper:
We might want to add a separate commit that drops the workarounds in `DES_bs_b.c` before we merge. Is that simply the checks of `benchmark_running` in the while loops? We'd still need to check `self_test_running`.

solardiz:
> We might want to add a separate commit that drops the workarounds in `DES_bs_b.c` before we merge.

Please do.

> Is that simply the checks of `benchmark_running` in the while loops? We'd still need to check `self_test_running`.

The `self_test_running` checks should also be dropped - their only purpose was to limit the workaround to the benchmark, excluding the self-test from the workaround.

solardiz (approved):
Looks correct to me. Thanks!

solardiz:
Confirmed this works as intended - and it does - by temporarily introducing debugging counter/printf into `DES_bs_b.c`.

solardiz:
Didn't notice a commit message was wrong - we don't have a `DES_bs_b.h`.

PR #4466, opened 2020-11-18
Unicode substitution

fpesce:
These are good rules on sites where a user ID is forced to be written as an ASCII string while the real user name or firstname would have an alternative Unicode form.
An optimization is used by overwriting the ASCII character with the second byte of Unicode and then inserting the first one
For each letters, we attempt a replacement of first, second, and third occurrences and up to all three occurrences.
There are cases too complex to be covered, for example the Icelandic firstname adalradur would have the d's substituted with eth and the third a with acute, we are only dealing with single letter substitution here.

### DIFF
--- a/run/john.conf
+++ b/run/john.conf
@@ -723,7 +723,6 @@ DefaultCharset =
 (?a )?d /?d a0 'p Xpz0
 )?a (?d /?a a0 'p Xpz0
 
-
 # "Single crack" mode rules
 [List.Rules:Single]
 # Simple rules come first...
@@ -1167,6 +1166,9 @@ b1 ]
 .include [List.Rules:Extra]
 .include [List.Rules:OldOffice]
 
+# Unicode substitution rules
+.include <unisubst.conf>
+
 # For Wordlist mode and very fast hashes
 [List.Rules:Jumbo]
 .include [List.Rules:Single-Extra]
@@ -1174,6 +1176,7 @@ b1 ]
 .include [List.Rules:ShiftToggle]
 .include [List.Rules:Multiword]
 .include [List.Rules:best64]
+.include [List.Rules:UnicodeSubstitution]
 
 # KoreLogic rules
 .include <korelogic.conf>
--- a/run/unisubst.conf
+++ b/run/unisubst.conf
@@ -1,0 +1,648 @@
+# These are good rules on sites where a user ID is forced to be written as an
+# ASCII string while the real user name or firstname would have an alternative
+# Unicode form.
+# An optimization is used by overstriking the ascii char with the second byte
+# of unicode and then inserting the first one
+# For each letters, we attempt a replacement of first, second, and third
+# occurrences and up to all three occurrences.
+# There are cases too complex to be covered, for example the Icelandic
+# firstname adalradur would have the d's substituted with eth and the third a
+# with acute, we are only dealing with single letter substitution here.
+[List.Rules:LowerUnicodeSubstitution]
+# Latin small letter n with tilde
+-u /n op\xb1 ip\xc3
+-u %2n op\xb1 ip\xc3
+-u %3n op\xb1 ip\xc3
+-u /n op\xb1 ip\xc3 /n op\xb1 ip\xc3
+-u /n op\xb1 ip\xc3 %2n op\xb1 ip\xc3
+-u %2n op\xb1 ip\xc3 %2n op\xb1 ip\xc3
+-u /n op\xb1 ip\xc3 /n op\xb1 ip\xc3 /n op\xb1 ip\xc3
+# Latin small letter u with diaeresis
+-u /u op\xbc ip\xc3
+-u %2u op\xbc ip\xc3
+-u %3u op\xbc ip\xc3
+-u /u op\xbc ip\xc3 /u op\xbc ip\xc3
+-u /u op\xbc ip\xc3 %2u op\xbc ip\xc3
+-u %2u op\xbc ip\xc3 %2u op\xbc ip\xc3
+-u /u op\xbc ip\xc3 /u op\xbc ip\xc3 /u op\xbc ip\xc3
+# Latin small letter o with diaeresis
+-u /o op\xb6 ip\xc3
+-u %2o op\xb6 ip\xc3
+-u %3o op\xb6 ip\xc3
+-u /o op\xb6 ip\xc3 /o op\xb6 ip\xc3
+-u /o op\xb6 ip\xc3 %2o op\xb6 ip\xc3
+-u %2o op\xb6 ip\xc3 %2o op\xb6 ip\xc3
+-u /o op\xb6 ip\xc3 /o op\xb6 ip\xc3 /o op\xb6 ip\xc3
+# Latin small letter e with acute
+-u /e op\xa9 ip\xc3
+-u %2e op\xa9 ip\xc3
+-u %3e op\xa9 ip\xc3
+-u /e op\xa9 ip\xc3 /e op\xa9 ip\xc3
+-u /e op\xa9 ip\xc3 %2e op\xa9 ip\xc3
+-u %2e op\xa9 ip\xc3 %2e op\xa9 ip\xc3
+-u /e op\xa9 ip\xc3 /e op\xa9 ip\xc3 /e op\xa9 ip\xc3
+# Latin small letter a with diaeresis
+-u /a op\xa4 ip\xc3
+-u %2a op\xa4 ip\xc3
+-u %3a op\xa4 ip\xc3
+-u /a op\xa4 ip\xc3 /a op\xa4 ip\xc3
+-u /a op\xa4 ip\xc3 %2a op\xa4 ip\xc3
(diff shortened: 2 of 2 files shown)


solardiz:
Thanks @fpesce!

Should these possibly be in a file like `unisubst.conf`, which we'd `.include` from `john.conf`?

Should we possibly have a `Unicode-Subtitution` section that would `.include` the two sections introduced here, and be in turn `.include`d from `Jumbo`?

fpesce:
Sure. It is indeed a bit long ;-) ill revise my PR after dinner ;-)

solardiz (approved):
Looks good to me now, although I haven't verified the individual substitutions.

magnumripper:
Thanks @fpesce!

@solardiz please wait for a more thorough review of mine (hopefully within days) before merging.

This is hard-coded to replace an ASCII `n` with, for example, the UTF-8 notation of `ñ`, no matter what. However, that's only sensible if we're actually running in UTF-8 mode. If we aren't, it will produce wads of binary garbage guaranteed to result in nothing but wasted cycles.

To ensure we **are** in UTF-8 mode, every single line should be prepended with `-u `, meaning "reject this rule unless the rules engine is running in UTF-8 mode" and all would be fine - except for the sad fact we'd not use this gem at all if we actually *have* the notation of an internal codepage. I'll try to figure out if we can do this better (that is, if one same ruleset could work in both scenarios). Maybe I need to add some detail in the engines, I'm not quite sure.

solardiz:
@fpesce I think we'll need you to force-push this as just _one_ commit before we merge. I think the 3 commits (and possibly more soon) are unsuitable for recording as such on the bleeding-jumbo branch - they're merely part of this PR's evolution.

Speaking of the rules themselves, maybe add those substituting only chars 2 and 3 (skipping the 1st) and those substituting chars 1 and 3 (skipping the 2nd). Then it'll be exhaustive over the possible (non-)substitutions for the first 3 chars.

@magnumripper Should we possibly also have an equivalent ruleset, differently limited by internal codepage choice, to substitute these characters with iso-8859 / cp1252 ones?

I am not convinced JtR's codepage awareness is helping here. Often (or even usually?) a set of target hashes is for passwords of unknown character encoding or of mixed character encodings, and it'd make sense to be testing multiple encodings against them, which we could do by having both potential rulesets I mentioned above active at once.

fpesce:
@solardiz : done (squash/force push + 2-3 and 1-3 substitutions added)

solardiz:
Looks good to me, but no need to describe this PR's history of changes even in the squashed commit's description.

BTW, are you going to add reverse substitutions next (perhaps as separate sections also included from Jumbo)?

magnumripper:
> @magnumripper Should we possibly also have an equivalent ruleset, differently limited by internal codepage choice, to substitute these characters with iso-8859 / cp1252 ones?

Yes, a similar ruleset for when running "codepage aware" is probably the best. Those would start with `-U` rule rejection flag and this way we could have a superset `--rules=substitution` that would work fine in any situation.

The codepage-aware one is easier to write (after recent d595de99c): Just write the rules as you would with ASCII, in a separate config file that's stored with UTF-8 encoding (that fact may seem strange but it isn't - when run it will be converted to whatever codepage is active at that time!). Like this (but we should try to make it perform the same job as UnicodeSubstitution):
```
# Ruleset for when running an internal codepage
[List.Rules:CodepageSubstitutionLower]
-U /a sa[åäãâáà]
-U /c scç
-U /e se[ëêéè]
-U /i si[ïîíì]
-U /n sn[ñ]
-U /o so[öøóò]
-U /u su[üûúù]
(...)

# Run both variants. Only the correct one will pass the rule reject flags
[List.Rules:Substitution]
.include [List.Rules:UnicodeSubstitution]
.include [List.Rules:CodepageSubstitution]
```
The superset `--rules:Substitution` will work with UTF-8 or any codepage, eg. LM format with `-target-enc=cp850` or NT format using any Latin-1 one, eg. `--internalcodepage=cp1252`.

> I am not convinced JtR's codepage awareness is helping here. Often (or even usually?) a set of target hashes is for passwords of unknown character encoding or of mixed character encodings, and it'd make sense to be testing multiple encodings against them, which we could do by having both potential rulesets I mentioned above active at once.

I noticed the current ruleset is rejected  (by `-u`) when running `--enc=raw`. It should not, so I'll look into that ASAP. But that's a separate issue #4472. 

Bottom line is you can merge this when ready - ~either~ with `-u` intact ~or possibly drop the `-u` until my issue is fixed~.

magnumripper:
#4472 is fixed already (with #4473), it was trivial.

magnumripper:
@fpesce while I played with this I noticed you replace "m" -> "µ", but not "u" -> "µ". I assume the latter could be useful too.

BTW this is a fantastic ruleset, I think it has very large potential. Will you create a "codepage-aware" version per the above as well? If not, I'll try to translate your work to one.

On a side note, only one of the rules will be used at a time, so from "mass" it will produce either "maß" or "mäss" but not "mäß". Using `./john (...) -rules=Unicodesubstitution -rules-stack=Unicodesubstitution` will try all combinations of two rules but ideally we should be able to do even more.

fpesce:
> @fpesce while I played with this I noticed you replace "m" -> "µ", but not "u" -> "µ". I assume the latter could be useful too.

I'll add the "u" -> "µ"
 
> BTW this is a fantastic ruleset, I think it has very large potential. 

Indeed, this ruleset is *really* efficient. Its discovery is even better: I've found it from a tool I was developing which figures out rules by itself, and this popped up. I'll opensource that later.

> On a side note, only one of the rules will be used at a time, so from "mass" it will produce either "maß" or "mäss" but not "mäß". Using `./john (...) -rules=Unicodesubstitution -rules-stack=Unicodesubstitution` will try all combinations of two rules but ideally we should be able to do even more.

I actually didn't know about -rules-stack... This is amazing.

> Will you create a "codepage-aware" version per the above as well? If not, I'll try to translate your work to one.

I can work on the codepage-aware version, but at this point I'm wondering if that PR is not getting bloated ;-)

solardiz:
> I can work on the codepage-aware version, but at this point I'm wondering if that PR is not getting bloated ;-)

I share your concern. The codepage-aware version and reverse substitutions should be separate PRs.

fpesce:
> > I can work on the codepage-aware version, but at this point I'm wondering if that PR is not getting bloated ;-)
> 
> I share your concern. The codepage-aware version and reverse substitutions should be separate PRs.

@solardiz : Then, let me push a last change to add the u -> Mu and cleanup the commit message, and feel free to merge afterwards.

magnumripper:
> I've found it from a tool I was developing which figures out rules by itself, and this popped up. I'll opensource that later.

Cool, be sure to announce it on john-users mailing list!

> > On a side note, only one of the rules will be used at a time, so from "mass" it will produce either "maß" or "mäss" but not "mäß". Using `./john (...) -rules=Unicodesubstitution -rules-stack=Unicodesubstitution` will try all combinations of two rules but ideally we should be able to do even more.
> 
> I actually didn't know about -rules-stack... This is amazing.

Actually it would be nice if the ruleset started with a line of `:` because otherwise you'd need to run in two passes: First with just `-rules` and then again with `-rules-stack` as well.

...or we could have even more supersets, like this:
```
[List.Rules:UnicodeSubstitution-with-passthru]
:
.include [List.Rules:UnicodeSubstitution]
```
...then run with `./john (...) -rules=Unicodesubstitution -rules-stack=Unicodesubstitution-with-passthru`. But that's probably overkill (multiple `:` rules will be optimized away anyway so we could just add it to each of the sub-rulesets).

solardiz:
> Actually it would be nice if the ruleset started with a line of `:`

I'd find this unexpected and wasteful when using the ruleset on its own, after having already run the wordlist directly. The ruleset's name suggests it does a specific thing, not also tries the words as they are.

> because otherwise you'd need to run in two passes: First with just `-rules` and then again with `-rules-stack` as well.

I'd expect to need to do that if I want both kinds of substitutions (of one kind of character and of two).

solardiz:
@fpesce One thing we forgot to ask you to do is add a `NEWS` entry for your contribution. Please do along with your next PR. Thank you!

magnumripper:
> > Actually it would be nice if the ruleset started with a line of `:`
> 
> I'd find this unexpected and wasteful when using the ruleset on its own, after having already run the wordlist directly. The ruleset's name suggests it does a specific thing, not also tries the words as they are.
> 
> > because otherwise you'd need to run in two passes: First with just `-rules` and then again with `-rules-stack` as well.
> 
> I'd expect to need to do that if I want both kinds of substitutions (of one kind of character and of two).

Using your reasoning we shouldn't have a nop rule in the start of "wordlist" ruleset either. It's a matter of taste of course - I would prefer being able to do both in a single session.

BTW we do have a `--rules-skip-nop` option, for skipping no-op rules anywhere (can also be used if you ran without rules at first). With no nop in these rules I'd want `--rules-add-nop` as well, and/or some config option like `StackedRulesAutoNOP = <bool>` for synthetically adding it.

Actually I think I'll actually implement that config option - it will be a non-intrusive patch and doesn't need an option flag.

PR #4470, opened 2020-11-19
RAR3 OpenCL: Implement GPU-side early-reject

### LINKED ISSUE #4269: RAR3 OpenCL format: Try moving the early-reject Huffman checks to GPU
This is mostly an experiment, but if we're lucky it could mean a boost for CPU-weak rigs with powerful GPU (the Huffman checks reject something like 96% of wrong candidates **really** early iirc, from just peeking at a single AES block or perhaps two).

It might also be of help if/when implementing pkzip format on GPU (similar checks, not identical).

magnumripper:
Adds GPU-side AES, CRC-32 and Huffman checks for minimizing CPU post-processing to actual rar inflation alone, when all early-reject logic fail (except for large stored files, which are better processed on CPU).

Not much difference in benchmark nor real speed, except in certain situations such as CPU-starved GPU-clusers, which should see a boost for certain RAR types.

Closes #4269, see also #3242


magnumripper:
This code even works on all three devices (CPU, Intel GPU, discrete GPU) of my macbook (which is kinda rare, those drivers are really picky *and* buggy and they normally don't give you the slightest of hint about what is upsetting them) but apparently it segfaults on the good old fglrx CPU driver :-(

magnumripper:
It also passes on (literally) all devices on our "HPC village" thingy

solardiz (approved):
I'm not familiar with this format, but I trust your testing. You can probably try to reproduce the old AMD issue on "well".

I also note that we have aliasing violations in memcmp*(), but those aren't new with this PR. The new function is similar to what we had before.

I also note that we could possibly have the string constants passed into memcmp*() explicitly aligned such that the optimization is more likely to be in use, rather than leave that to chance.

magnumripper:
I "think" most GPU runtimes aligns nearly everything to the largest need ever used for any data type on the target device (all alignment problems I've seen were due to being offset into a buffer). But yes, that's an assumption we shouldn't make.

PR #4471, opened 2020-11-19
Loopback mode: Speed up loading of LM halves

magnumripper:
Ensures that we're only checking the pot file for actual LM hashes when assembling LM halves. Also drops some duplicated loader code from john.c.

### DIFF
--- a/src/john.c
+++ b/src/john.c
@@ -975,31 +975,6 @@ static void load_extra_pots(struct db_main *db, void (*process_file)(struct db_m
 	} while ((line = line->next));
 }
 
-static void db_main_free(struct db_main *db)
-{
-	if (db->format &&
-	    (db->format->params.flags & (FMT_DYNA_SALT | FMT_BLOB))) {
-		struct db_salt *psalt = db->salts;
-		while (psalt) {
-			struct db_password *pw = psalt->list;
-
-			if ((db->format->params.flags & FMT_BLOB) && pw &&
-			    pw->binary &&
-			    ((fmt_data*)pw->binary)->flags == FMT_DATA_ALLOC) {
-				do {
-					fmt_data *bin = pw->binary;
-					BLOB_FREE(db->format, bin);
-				} while ((pw = pw->next));
-			}
-			if (db->format->params.flags & FMT_DYNA_SALT)
-				dyna_salt_remove(psalt->salt);
-			psalt = psalt->next;
-		}
-	}
-	MEM_FREE(db->salt_hash);
-	MEM_FREE(db->cracked_hash);
-}
-
 static void john_load(void)
 {
 	struct list_entry *current;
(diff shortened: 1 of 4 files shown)


src/john.c
```diff
@@ -1241,19 +1216,16 @@ static void john_load(void)
...
 	/*
-	 * Nefarious hack and memory leak. Among other problems, we'd want
-	 * ldr_drop_database() after this, but it's built with mem_alloc_tiny()
-	 * so it's not trivial. Works like a champ though, except with
-	 * DEScrypt. I have no idea why, maybe because LM and DES share code?
+	 * Nefarious hack and memory leak. Works like a champ though.
 	 */
-	if (options.flags & FLG_LOOPBACK_CHK &&
-	    database.format != &fmt_LM && database.format != &fmt_DES) {
+	if (options.flags & FLG_LOOPBACK_CHK && database.format != &fmt_LM) {
```
solardiz: Added braces around `options.flags & FLG_LOOPBACK_CHK`.

src/john.c
```diff
@@ -1241,19 +1216,16 @@ static void john_load(void)
...
 		struct db_main loop_db;
 		struct fmt_main *save_list = fmt_list;
 		char *loop_pot = options.wordlist ?
 			options.wordlist : options.activepot;
 
 		fmt_list = &fmt_LM;
+		fmt_list->next = NULL;
```
solardiz: I guess this setting of `fmt_LM.next` to `NULL` is the main functional change made with this PR?

Is it OK to leave the formats list corrupted like that? Perhaps we should at least add a comment explaining this and why it's OK for now.

I don't understand the logic here - why do we first check `database.format != &fmt_LM` and then fake LM-only here?

BTW, we have several other formats that can split hashes and reassemble them on `--show`. descrypt is one of those, but in jumbo it's probably a few more (I guess NETLM*).

solardiz (commented):
I took a look and made some comments. I'm not familiar enough with the original loopback mode logic to approve this... but I guess I'll have to anyway, because I doubt there's anyone besides @magnumripper who's familiar with this code and someone does need to approve the changes for them to be merged. ;-)

src/john.c
magnumripper: > I guess this setting of `fmt_LM.next` to `NULL` is the main functional change made with this PR?

Right, that's the actual change. Before this, a huge pot file could take ages to scan, even though we're only interested in LM halves. I never understood why until now: Even if I said `--format=NT` and all other formats were ditched, there's still a `fmt_LM->next` present and it leads to (almost) every single format.

> Is it OK to leave the formats list corrupted like that? Perhaps we should at least add a comment explaining this and why it's OK for now.

At first I wrote code to un-corrupt it afterwards but it really can't be necessary: This is *after* loading, having checked pot file(s) and "fixed" database. We have manually or automatically ended up with a single format, and it's not LM.

> I don't understand the logic here - why do we first check `database.format != &fmt_LM` and then fake LM-only here?
> 
> BTW, we have several other formats that can split hashes and reassemble them on `--show`. descrypt is one of those, but in jumbo it's probably a few more (I guess NETLM*).

The only thing this section of code does is build a list of assembled 8-14 character passwords from their halves, so we're not interested in any other format except perhaps bigcrypt. Feeding the result back to the LM format would be completely meaningless - the "normal" loopback (more akin to `cut -d: -f2- john.pot | unique loopback.lst`) handles that later in `wordlist.c`. Normal use for this code is a pwdump file with both LM and NT, and running `-format:NT -loopback` after cracking all LM.

It could be used for DEScrypt too - we'd simply need to build a temporary format list of `LM -> DEScrypt -> NULL`. I just never had use for that and I'm not sure I ever saw a good set of samples for bigcrypt to test with. Besides those two I'm not aware of ANY other format with a `split()` that can return > 1.

magnumripper:
I see now it's DEScrypt that may assemble halves (like you said), not BSDIcrypt.

magnumripper:
I added bigcrypt now, even managed to test it a little with the one hash we have on Openwall wiki.

src/john.c
```diff
@@ -1241,19 +1216,22 @@ static void john_load(void)
...
 	/*
-	 * Nefarious hack and memory leak. Among other problems, we'd want
-	 * ldr_drop_database() after this, but it's built with mem_alloc_tiny()
-	 * so it's not trivial. Works like a champ though, except with
-	 * DEScrypt. I have no idea why, maybe because LM and DES share code?
+	 * Nefarious hack and memory leak. Works like a champ though.
 	 */
-	if (options.flags & FLG_LOOPBACK_CHK &&
-	    database.format != &fmt_LM && database.format != &fmt_DES) {
+	if ((options.flags & FLG_LOOPBACK_CHK) && database.format != &fmt_LM && database.format != &fmt_DES) {
```
solardiz: Doesn't this check unnecessarily prevent having loopback test reassembled LM cracks against descrypt hashes? An obscure case, but not entirely unrealistic.

solardiz (approved):
I'm not entirely sure, but since there's no one else to review and approve this, I do. ;-)

src/john.c
magnumripper: Yes, and that lead to some confusion when I tested this (having one bigcrypt hash and one pwdump entry in a test file, with same passphrase). I'll experiment a little.

magnumripper: I now committed a version with no restrictions. I also now restore the `fmt.next` pointers, just in case. Also improved comments.

magnumripper: > with no restrictions

BTW I decided it's relevant to activate this even for LM - in really obscure cases, a long password could be permuted by rules and give new ones in ways that the same rules wouldn't end up with for the halves.

PR #4475, opened 2020-11-20
pkzip format: Bump max. length to 63 (from 31), by user request.

magnumripper:
No performance regression seen.  Actually I saw a boost, after replacing an irrelevant call to strnzcpyn with strnzcpy - as the result wasn't used.

### DIFF
--- a/src/pkzip_fmt_plug.c
+++ b/src/pkzip_fmt_plug.c
@@ -54,7 +54,7 @@ john_register_one(&fmt_pkzip);
 #define BENCHMARK_COMMENT   ""
 #define BENCHMARK_LENGTH    7
 
-#define PLAINTEXT_LENGTH    31
+#define PLAINTEXT_LENGTH    63
 
 #define BINARY_SIZE         0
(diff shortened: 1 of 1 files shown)


solardiz (approved):
Looks good to me. I wonder if we possible need to re-tune `MAX_KEYS_PER_CRYPT` and/or `OMP_SCALE`? Also, with sizes like this (matching typical cache line size now) I wonder if aligning `saved_key` to cache line boundary would possibly speedup (especially) OpenMP builds.

And I guess 125 (our usual max) would cause some slowdown? BTW, maybe we should bump up that global limit to 127, requiring 130-byte buffers somewhere else, but exactly 128 bytes per key buffer in formats to fit caches better (when all tested passwords are short, we'd fully skip the cache lines corresponding to the second halves of those buffers, then).

Edit: I think we should also look into and reconsider why our maximum plaintext length is smaller than `PLAINTEXT_BUFFER_SIZE` by 3. I think it should be practical to avoid that. But let's discuss that separately from this PR.

PR #4476, opened 2020-11-20
Correct a few format_name vs. algorithm_name issues

magnumripper:
Some formats had stuff in algorithm_name that was more appropriate in format_name.  Some others got brief info added.

### DIFF
--- a/src/7z_common.h
+++ b/src/7z_common.h
@@ -16,7 +16,7 @@
 #include "lzma/CpuArch.h"
 #include "crc32.h"
 
-#define FORMAT_NAME             "7-Zip"
+#define FORMAT_NAME             "7-Zip archive encryption"
 #define FORMAT_TAG              "$7z$"
 #define TAG_LENGTH              (sizeof(FORMAT_TAG)-1)
 #define BENCHMARK_COMMENT       " (512K iterations)"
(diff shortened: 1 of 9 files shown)


src/racf_kdfaes_fmt_plug.c
```diff
@@ -38,7 +38,7 @@ john_register_one(&fmt_racf_kdfaes);
 #define FORMAT_NAME             ""
 #define FORMAT_TAG              "$racf$*"
 #define FORMAT_TAG_LEN          (sizeof(FORMAT_TAG)-1)
-#define ALGORITHM_NAME          "KDFAES (DES + HMAC-SHA256/" ARCH_BITS_STR " + AES-256)"
+#define ALGORITHM_NAME          "DES + HMAC-SHA256/" ARCH_BITS_STR " + AES-256"
```
solardiz: IIRC, `KDFAES` is an official name for their KDF, so I'd keep it in here. It's similar to how we mention `PBKDF2` or `bcrypt` and not just their underlying primitives.

solardiz (approved):
These look good to me, except that I'd keep `KDFAES`. We might also need to update `relbench-unify`, @frank-dittrich?

src/racf_kdfaes_fmt_plug.c
magnumripper: reverted

frank-dittrich:
> These look good to me, except that I'd keep `KDFAES`. We might also need to update `relbench-unify`, @frank-dittrich?

When I find time to contribute again (sometime this year), I'll first look into fixing tab completion (for bash). Afterwards, I might update relbench-unify.

PR #4478, opened 2020-11-20
Shared OpenCL HMAC-SHA1: Fix benign problems

magnumripper:
Seen on Travis-CI: "warning: duplicate 'const' declaration specifier"

While at it, I added a shared OpenCL HMAC-MD5 for use later.


solardiz (approved):
I have mixed feelings about introducing shared HMAC-MD5 when it's not in use yet. How soon would we make use of it, and where?

magnumripper:
> I have mixed feelings about introducing shared HMAC-MD5 when it's not in use yet. How soon would we make use of it, and where?

Two formats have their own implementations. I plan to fix them

Issue #2004, opened 2016-01-22
ASan error with --fuzz --format=dynamic='...'


frank-dittrich:
Not sure whether this is supposed to work or not:
```
(bleeding-jumbo)run $ ./john --fuzz --format='dynamic=md4(MD4($p).$s.md4($p))'
Fuzzing: dynamic= [128/128 AVX 4x3]... 
=================================================================
==13240==ERROR: AddressSanitizer: global-buffer-overflow on address 0x000000ed9060 at pc 0x7f1c9cf97b29 bp 0x7ffeebd9b5c0 sp 0x7ffeebd9ad68
WRITE of size 1075 at 0x000000ed9060 thread T0
    #0 0x7f1c9cf97b28  (/lib64/libasan.so.2+0x62b28)
    #1 0x5a63c8 in dynamic_expr_normalize /home/fd/git/JtR/src/dynamic_compiler.c:301
    #2 0x5c89fd in dynamic_compile_prepare /home/fd/git/JtR/src/dynamic_compiler.c:2373
    #3 0x67e523 in our_prepare /home/fd/git/JtR/src/dynamic_compiler_fmt_plug.c:120
    #4 0x7ea973 in ldr_split_line /home/fd/git/JtR/src/loader.c:629
    #5 0x7ebab9 in ldr_load_pw_line /home/fd/git/JtR/src/loader.c:827
    #6 0x8628ac in fuzz_test /home/fd/git/JtR/src/fuzz.c:564
    #7 0x862d97 in fuzz /home/fd/git/JtR/src/fuzz.c:668
    #8 0x7e5a4c in john_run /home/fd/git/JtR/src/john.c:1592
    #9 0x7e72d0 in main /home/fd/git/JtR/src/john.c:1973
    #10 0x7f1c9b5336ff in __libc_start_main (/lib64/libc.so.6+0x206ff)
    #11 0x406638 in _start (/home/fd/git/JtR/run/john+0x406638)

0x000000ed9060 is located 32 bytes to the left of global variable 'Buf' defined in 'dynamic_compiler.c:260:15' (0xed9080) of size 512
0x000000ed9060 is located 0 bytes to the right of global variable 'Buf' defined in 'dynamic_compiler.c:299:15' (0xed8e60) of size 512
SUMMARY: AddressSanitizer: global-buffer-overflow ??:0 ??
Shadow bytes around the buggy address:
[...]

PR #4479, opened 2020-11-20
Fix the dreaded legacy build

magnumripper:
From the looks of it it was b0rken again since some time.

### DIFF
--- a/src/formats.c
+++ b/src/formats.c
@@ -72,14 +72,13 @@ static void test_fmt_case(struct fmt_main *format, void *binary,
  */
 int fmt_match(const char *req_format, struct fmt_main *format, int override_disable)
 {
-	int enabled = override_disable || !((options.flags & FLG_TEST_CHK) || options.listconf) ||
-		!cfg_get_bool(SECTION_DISABLED, SUBSECTION_FORMATS, format->params.label, 0);
-
 	/* Exact full match */
 	if (!strcasecmp(req_format, format->params.label))
 		return 1;
 
 #ifndef BENCH_BUILD
+	int enabled = override_disable || !((options.flags & FLG_TEST_CHK) || options.listconf) ||
+		!cfg_get_bool(SECTION_DISABLED, SUBSECTION_FORMATS, format->params.label, 0);
 	char *pos = NULL;
 
 #ifndef DYNAMIC_DISABLED
--- a/src/bench.c
+++ b/src/bench.c
@@ -161,7 +161,7 @@ struct db_main *ldr_init_test_db(struct fmt_main *format, struct db_main *real)
 }
 
 // lol, who cares about memory leaks here.  This is just the benchmark builder
-void ldr_free_db(struct db_main, int)
+void ldr_free_db(struct db_main* foo, int bar)
 {
 }
 #endif


src/formats.c
```diff
@@ -72,8 +72,10 @@ static void test_fmt_case(struct fmt_main *format, void *binary,
  */
 int fmt_match(const char *req_format, struct fmt_main *format, int override_disable)
 {
+#ifndef BENCH_BUILD
 	int enabled = override_disable || !((options.flags & FLG_TEST_CHK) || options.listconf) ||
 		!cfg_get_bool(SECTION_DISABLED, SUBSECTION_FORMATS, format->params.label, 0);
+#endif
```
solardiz: This could simply be moved under the existing `#ifndef` below, near the `char *pos = NULL;` line.

solardiz (approved):

src/formats.c
magnumripper: Right, I'll do that. I'm so used with trying to keep differences from core branch's code at a graspable level but we can't merge that branch anymore anyway due to git-cvs having eroded to being unusable with reasonable effort.

Oh btw I have a semaphore in my head shouting we have some tiny(?) rule change from core not yet merged into Jumbo

magnumripper: Right, I'll do so. I'm still used to try keeping core files close enough to CVS that I can understand the diffs but now that `git-cvs` corroded to completely unusable without unreasonable effort, it's of no use anymore.

magnumripper: I see that's not even a core function btw... A whole lot of them aren't 😆

magnumripper: No, wait - it is. It's the `bench.c` one that's completely new code (and an awful cludge at that)

magnumripper:
Perhaps I was a spoiled child but this mandatory Travis check of an as-trivial-as-it-gets commit has been going on for over 4h now even though most parts of the globe are off to relax somewhere and not committing anything at all. Are there better alternatives? Or should we simply throw some money at them?

ghost:
> Perhaps I was a spoiled child but this mandatory Travis check of an as-trivial-as-it-gets commit has been going on for over 4h now even though most parts of the globe are off to relax somewhere and not committing anything at all. Are there better alternatives? Or should we simply throw some money at them?

For weeks, Travis.org is taking 6, 8, 10 hours to build. Yes, 80% of it is "waiting time". They have 4, maybe 5 hundred bots, but they have almost 6 thousand jobs in their backlog NOW (Saturday).

**[edited]**
* My latest run (`#2004 passed` Ran for 12 hrs 2 sec)

solardiz:
> For weeks, Travis.org is taking 6, 8, 10 hours to build.

FWIW, I setup Travis (I think it's .com) for LKRG, and it consistently completes (trivial) builds in minutes (we're only testing being able to build LKRG now - not yet testing that the built LKRG actually works).

magnumripper:
I see now Travis actually finished successfully 20 hours ago, but Github is still "waiting for status to be reported".

@solardiz just merge this please!

magnumripper:
Perhaps the finished one was older. I now found "*Builds have been temporarily disabled for public repositories due to a negative credit balance. Please go to the Plan page to replenish your credit balance or alter your Consume paid credits for OSS setting.*"

https://travis-ci.com/github/openwall/john

solardiz:
> _Builds have been temporarily disabled for public repositories due to a negative credit balance. Please go to the Plan page to replenish your credit balance or alter your Consume paid credits for OSS setting._

Clicking either of the two links in that message results in:

```
There was an error, please try again.
```

Let's see if things start working later.

That said, I think our use of CI is quite wasteful, e.g. when we ran many builds and tests of those builds on recent introduction of rulesets and force-pushes to that PR. Can we find a clean way to skip most CI tests (perhaps except for the quickest, which involves basic loading of `john.conf`) on such PRs?

ghost:
> That said, I think our use of CI is quite wasteful, e.g. when we ran many builds and tests of those builds on recent introduction of rulesets and force-pushes to that PR. Can we find a clean way to skip most CI tests (perhaps except for the quickest, which involves basic loading of john.conf) on such PRs?

There are some ways to do so. Run some builds only in "main" branch, use the `[skip ci]` in the commit message, ...

I remember I used `[skip ci]` some time ago. Then, only magnum was able to merge the PR. If we run some jobs only in "main" branch, we will test only **after** merge.

Issue #4087, opened 2019-09-01
We're out of options flags


magnumripper:
*I have actually played with the idea of making Jumbo's options flags 128-bit (and simply losing some checks if that's not supported, like on a 32-bit build) but that's not quite the final solution - and it has some caveats, eg. we can't write that big integers in source simply as `0x00001000000000000000000000000000ULLL` but rather as something like `(0x0000100000000000 << (uint128_t)64)`*

_Originally posted by @magnumripper in https://github.com/magnumripper/JohnTheRipper/issues/4034#issuecomment-504219417_

That crazy idea might not even fly. We do have a few low-half bits still usable but we're getting close to really need a solution.

magnumripper:
We could also want to find some canonical way for `options.c` to add support for boolean options with a (possibly not yet known) default, as in `--keep-guessing` vs. `--no-keep-guessing` which is sort of a three-state: True, False or Not given (meaning we should default to the format's `FMT_NOT_EXACT` flag in that case - but the state of it isn't known yet when we process config and options...)

magnumripper:
Hmm I think I can visualize how to implement it by eg. introducing a new `OPT_THREESTATE` flag.

New style could be defined something like this:
```diff
- 	{"keep-guessing", FLG_KEEP_GUESSING, FLG_KEEP_GUESSING,
- 		0, FLG_NO_KEEP_GUESSING},
- 	{"no-keep-guessing", FLG_NO_KEEP_GUESSING, FLG_NO_KEEP_GUESSING,
- 		0, FLG_KEEP_GUESSING},
+ 	{ "keep-guessing", FLG_ZERO, 0, FLG_CRACKING_CHK, OPT_THREESTATE, &options.keep_guessing },
```
This would mean:
1. The `OPT_THREESTATE` would let `options.c` know that you can either say `--keep-guessing` or `--no-keep-guessing` so no extra definition line for the latter.
2. The variable would be set by `options.c` accordingly, to eg. `1` or `0` or, if not given at all, to `-1` (which in this very case would mean we put `FMT_NOT_EXACT` there, from the format - once known).
3. The existing logic for `FMT_ZERO` could be amended to take care of these without requiring a new options flag (let alone two) for each option. So we'd go from two flags needed, to none - per such option.

This logic could be extended to also support an optional parameter, such as `--rules[=SECTION...]` vs. `--no-rules` (which would automagically refuse an argument).

solardiz:
[...]

PR #4482, opened 2020-11-21
Implement OPT_THREESTATE

magnumripper:
...and make use of it for --[no-]keep-guessing option only, for demonstration.

See #4087

The first commit is kept small, for review-ability.
Second commit is just refactoring and reformatting (eg. refactoring the whole definition of `opt_entry[]` for 120 columns). No functional change.
Third commit (not pushed yet) is having more options use this, freeing up options flags!

We should merge the three without squashing IMHO.


src/getopt.c
```diff
@@ -178,15 +189,27 @@ void opt_check(struct opt_entry *list, opt_flags flg, char **argv)
...
 			        completed_param ? completed_param : "");
 		}
 	}
+
+	/* Set all three-states */
+	if ((entry = list))
+	while (entry->name) {
+		if (entry->req_clr & OPT_THREESTATE)
+			*(int*)entry->param = entry->seen ? entry->threestate_bool : -1;
+		entry++;
```
magnumripper, line 239: This first implementation requires `param` to be an integer. I guess we could implement string three-states if we wanted to, such as (NULL, empty string or non-empty string) mapping to (not given, given as `--no-opt`, given as `-opt=string`) but I'm not sure we'll need it so I'll fix that if/when we do.

magnumripper:
Some tests
```
$ ./john scraped.sam -form:nt -mask:password 
(...)
0g 0:00:00:00  0g/s 100.0p/s 100.0c/s 523600C/s password
Session completed. 
```
```
$ ./john scraped.sam -form:nt -mask:password --keep-guessing 
Note: Will keep guessing even after finding a possible candidate.
(...)
password         (dispense1:1136)     
1g 0:00:00:00  100.0g/s 100.0p/s 100.0c/s 529400C/s password
Session completed. 
```
```
$ ./john scraped.sam -form:nt -mask:password --no-keep-guessing 
```
(the above was verified with debug prints - as was not giving the option, resulting in variable set to `-1`)

```
$ ./john scraped.sam -form:nt -mask:password --keep-guessing=bogus
Extra parameter for option: "--keep-guessing=bogus"

$ ./john scraped.sam -form:nt -mask:password --no-keep-guessing=bogus
Extra parameter for option: "--no-keep-guessing=bogus"
```

Also verified:
- rec file argv expansion for `-kee` --> `--keep-guessing`
- rec file argv expansion for `-no-kee` --> `--no-keep-guessing`

solardiz:
I don't understand how/why you're testing keep-guessing with `-form:nt -mask:password`. I'd expect you'd need at least a longer candidate password stream where you'd be able to make `john` literally keep guessing. And to test the no- form, you'd need a format that defaults to keep-guessing.

magnumripper:
> I don't understand how/why you're testing keep-guessing with `-form:nt -mask:password`. I'd expect you'd need at least a longer candidate password stream where you'd be able to make `john` literally keep guessing. And to test the no- form, you'd need a format that defaults to keep-guessing.

It wasn't made very clear 😆 but the password was already cracked. When I added `--keep-guessing` it was cracked again.

Here's setting CRC32 format to NOT_EXACT and testing
```diff
diff --git a/src/crc32_fmt_plug.c b/src/crc32_fmt_plug.c
index d6a4dfb8c..736536987 100644
--- a/src/crc32_fmt_plug.c
+++ b/src/crc32_fmt_plug.c
@@ -273,7 +273,7 @@ struct fmt_main fmt_crc32 = {
 #ifdef _OPENMP
 		FMT_OMP | FMT_OMP_BAD |
 #endif
-		FMT_CASE | FMT_8_BIT,
+		FMT_CASE | FMT_8_BIT | FMT_NOT_EXACT,
 		{
 			"version [0:CRC-32 1:CRC-32C]",
 		},
```
```
$ echo 'Dead Cafe:$crc32$00000000.deadcafe' > test.in
$ ./john -form:crc32 test.in -mask:?a -len=5
Note: This format may emit false positives, so it will keep trying even after finding a possible candidate.
Loaded 1 password hash (CRC32 [CRC32 32/64 CRC-32C AVX])
Cost 1 (version [0:CRC-32 1:CRC-32C]) is 0 for all loaded hashes
Warning: no OpenMP support for this hash type, consider --fork=16
0g 0:00:00:28 34.32% (ETA: 12:37:26) 0g/s 94830Kp/s 94830Kc/s 94830KC/s |w`Kw..4b>Kw
80t\$            (Dead Cafe)     
TCYX`            (Dead Cafe)     
2g 0:00:01:21 98.70% (ETA: 12:37:27) 0.02469g/s 94282Kp/s 94282Kc/s 94282KC/s I[v\}..J(f\}
Remaining 1 password hash
Session aborted
```
```
$ rm -f john.pot && ./john -form:crc32 test.in -mask:?a -len=5 --no-keep-guessing
Note: This format may emit false positives, but we will stop after finding first
Loaded 1 password hash (CRC32 [CRC32 32/64 CRC-32C AVX])
Cost 1 (version [0:CRC-32 1:CRC-32C]) is 0 for all loaded hashes
Warning: no OpenMP support for this hash type, consider --fork=16
80t\$            (Dead Cafe)     
1g 0:00:01:03 DONE (2020-11-22 12:38) 0.01580g/s 91482Kp/s 91482Kc/s 91482KC/s m42\$..ztt\$
No remaining hashes
Session completed. 
```

magnumripper:
BTW in case you missed/wonder why CRC32 isn't already FMT_NOT_EXACT, see #4483

solardiz (approved):
I haven't thoroughly reviewed this, but I like it conceptually and I saw nothing obviously wrong.

An unrelated thought I got is: for runtime detection of bugs, we currently mix ifs/printfs and `assert` throughout the codebase. We might want to standardize on one or the other. Sure for `--test`, `--test-full`, and maybe `--fuzz` it's explicit checks and prints, but for the rest maybe we should consistently use `assert`... or maybe not. I am just thinking out loud. `assert` has both a drawback and an advantage - depending on use case and how you look at it - then it's turned into a no-op with `-NDEBUG`.

Edit: we also have the special case of "excessive partial hash collisions", where we only probabilistically detect this likely bug and don't want to abort. And we have salt alignment issues, on which we only complain without even formally failing the test. So I guess we have to accept that there's a variety of cases with different handling.

solardiz:
@magnumripper Given the Travis CI performance issues, please let me know if I should merge this for you. Thanks!

magnumripper:
Yes, please merge. My next commit will be more likely to be buggy and in need of more testing - probably better to make a separate  PR for it. It will free up a wealth of options flags

PR #4485, opened 2020-11-23
getopt: Add OPT_BOOL, using most of the tri-state logic.

magnumripper:
Use it only for --log-stderr, to demonstrate.

Also refactors THREESTATE -> TRISTATE. Not that either of them is better but because once it hit me I couldn't get it off my OCD mind.
Also add comments documenting older deviations from core getopt.

This PR will soon be filled with more commits, making use of OPT_BOOL and OPT_TRISTATE and freeing up options flags.


solardiz:
I'm not convinced `OPT_BOOL` is a good idea. We replace flags with individual variables, and that isn't cleaner. It does save us flag bits, but do we have to? Wouldn't tri-state save enough of those?

What would we use the flag bits for, then? Just cases where we need to detect option (in)compatibility? Would we be introducing flags when such a case arises? Right now, we always have flags for Booleans, so we can readily start referring to those in other option definitions.

magnumripper:
> I'm not convinced `OPT_BOOL` is a good idea. We replace flags with individual variables, and that isn't cleaner. It does save us flag bits, but do we have to? Wouldn't tri-state save enough of those?

Tri-states are lovely for command-line options that may override config options or other variables known later but there aren't that many of them. It hit me that we could mis-use tri-states for actual booleans but that would mean we'd need to handle resetting `-1` to `0` separately for each of them. So I decided `getopt` should do it. We lost a single bit for `OPT_BOOL` but that is easily amortized.

> What would we use the flag bits for, then? Just cases where we need to detect option (in)compatibility?

Exactly. There are things we should detect for, but do not because so many things just use `FLG_ONCE` due to years of flags starvation.

> Would we be introducing flags when such a case arises? Right now, we always have flags for Booleans, so we can readily start referring to those in other option definitions.

I definitely think we should have this `OPT_BOOL` functionality but you may be right we shouldn't over-use it as long as we have flags.

solardiz (approved):
OK, this is up to you, and I hope you test all of these options you modify code for. ;-)

magnumripper:
Thanks, I'll mark it as draft for now and we'll see where this leads

magnumripper:
Feature-froze this a couple of days ago but the last commit fixes issues we've had for years and years, such as not bailing with error from `john -restore -device=2`. Fixing that wasn't trivial, still regression-testing.

For one, what options *do* we want to be compatible with `--test` (et al)? I can think of these:
`--cost`, `--tune`, `--lws`, `--gws` and `--device`. Oh, and `--config`.

solardiz:
> For one, what options _do_ we want to be compatible with `--test` (et al)? I can think of these:
> `--cost`, `--tune`, `--lws`, `--gws` and `--device`. Oh, and `--config`.

Also `--format` and `--mask`. And looking at the diff, also `--save-memory`.

Edit: oh, also `--verbosity`.

magnumripper:
> > For one, what options _do_ we want to be compatible with `--test` (et al)? I can think of these:
> > `--cost`, `--tune`, `--lws`, `--gws` and `--device`. Oh, and `--config`.
> 
> Also `--format` and `--mask`. And looking at the diff, also `--save-memory`.
> 
> Edit: oh, also `--verbosity`.

Right, I had them checked but forgot to mention them here. Also `--no-mask` which isn't coded as a tri-state so it's a completely separate option internally.

For a few days I had `FMT_ONCE` actally defined to non-zero but that simply was too much work to get straight. I'm now focusing on adding blocks the other way round (`--restore` can't block `--keep-guessing` because the latter is a "no-flag" option - but `--keep-guessing` can block `--restore`).

I regret putting this last commit in this PR, maybe I should separate it to an other PR and merge this now. OTOH it helped me finding actual bugs in other places which are now fixed.

There are a few bugs in current main branch fixed by this PR. One is `--no-log` option doesn't work (it gets misinterpreted as `--log-stderr`).

magnumripper:
@solardiz I see strange problems like this one:
```
	{"restore", FLG_RESTORE_SET, FLG_RESTORE_CHK, 0, ~FLG_RESTORE_SET & ~GETOPT_FLAGS, OPT_FMT_STR_ALLOC, &options.session},
(...)
	{"verbosity", FLG_VERBOSITY, FLG_VERBOSITY, 0, OPT_REQ_PARAM, "%u", &options.verbosity},
```
`-v` sets FLG_VERBOSITY and `--restore` should reject that. However, that's not what actually happens - `./john --restore -v=4` gladly restores the session and ignores the option. This is not new, same happens in bc8fa2b2d9, before any of my recent getopt work.

Last few days I've almost gone mad trying to understand issues like this (they sometimes go away while I'm debugging it).

Another thing while we're at it: Most options clear their own flag (as in `FLG_RESTORE_SET, FLG_RESTORE_CHK` but I have yet to understand why, nor if it's important or not. Other than that I think I grok it.

magnumripper:
I think I see the problem... All 32-bit defines such as `FLG_RESTORE_CHK			0x00010000` need to have `ULL` postfix.

Things like `~FLG_RESTORE_SET` were missing the high 32 bits...

magnumripper:
That was a hideous bug, very hard to spot. It originated in 2012! Now my tests make a whole lot more sense 😆  I'm not insane! I'm cured!

magnumripper:
@solardiz I think we can merge this now (or you might want to review it). It's manually tested a lot as well as carefully dry-run and also somewhat excersised with a test script I wipped up, but the test combinations are so many I probably missed something. OTOH it fixes known bugs.

src/list.c
```diff
@@ -81,6 +67,22 @@ void list_add_unique(struct list_main *list, char *data)
...
+{
+	char *comma;
+
+	do {
+		if ((comma = strchr(data, ',')))
+			*comma = 0;
+
+		list_add_unique(list, data);
+
+		data = comma + 1;
```
solardiz, line 80: Just noticed a bug coming from core (added to my notes): we might be adding 1 to `NULL` here. We're not using the resulting pointer, but it's UB nevertheless. We should fix this separately.

magnumripper: It might be UB but once it is, that value won't be used. Is that still a risk?

solardiz: I think formally this is still a risk. In practice, probably not.

magnumripper: We should of course fix it anyway. Anyway I'm glad I'm not the only one producing bugs, just a magnitude or two more of them 🙈

src/options.c
```diff
@@ -272,6 +273,8 @@ static struct opt_entry opt_list[] = {
 "--single-user-seed=FILE    wordlist with seeds per username (user:password[s]\n" \
 "                           format)\n" \
 "--[no-]single-retest-guess override config for SingleRetestGuess\n" \
+"--single-pair-max=N        override max. number of word pairs generated (%u)\n" \
+"--no-single-pair           disable single word pair generation\n" \
```
solardiz, line 267: Maybe it'd be cleaner and would save us a line in the `--help` output if we simply accepted `--single-pair-max=0` as meaning the same as `--no-single-pair` (perhaps we already do with this commit), without having the latter option name. I am impressed by what the new tri-state code is capable of, but that doesn't necessarily make this a good idea here.

magnumripper: Yeah maybe. It'd be a usage-blob-change only - the code already allows this

magnumripper:
Oh BTW one thing that I noticed but haven't dug into yet: You can give `./john file file` and even after this commit (now using `list_add_unique`) I've seen `file` will be read twice. Very strange but I forgot it until now and it's not a blocker.

Issue #3917, opened 2019-05-10
Sync md[45].[ch] with upstream


solardiz:
Jumbo uses a revision of my generic MD4 and MD5 implementations from some years back. I've since revised the upstream versions in several ways. Jumbo also made its changes. We should probably take my latest revisions and reintroduce only the changes that we actually still want to have.

One change we're missing is my addition of H/H2 split to MD4, similar to what I had introduced for MD5 earlier, to help the compiler detect the common subexpressions.

This commit by JimF gets in the way: abe2cdb591b03c871f2c0d6e046cbeed36c6dee4. As I understand, it primarily renames `a, b, c, d` to uppercase `A, B, C, D`, which is only made use of in pbkdf2_hmac_md5.h. I doubt matching OpenSSL's context layout in any other ways is relevant.

magnumripper:
Agreed - I thought I already added the H/H2 split.

solardiz:
> I thought I already added the H/H2 split.

As I understand: to MD5 yes, to MD4 not yet.

solardiz:
A relevant question for us: do we still want to be able to use OpenSSL's MD[45] as well? I doubt it. While OpenSSL might provide assembly implementations, it also zeroizes memory on `*_Final`, which for something as fast as MD[45] probably undoes any advantage the assembly code might have provided. Also, going inside of OpenSSL implementation-specific context structs in pbkdf2_hmac_md5.h is no good - we can do that reliably for our own implementation-specific context structs, but not for third-party ones (not long-term).

So I suggest we fully standardize on our own md[45][.ch] and minimize their differences between my upstream versions and JtR jumbo versions of them.

OTOH, this also enable us to introduce new APIs such as a combined InitUpdate call (initialize new context and do a first update on it at once, thereby saving on function call overhead). Or we can simply make Init macros in .h files, but this will increase code size in the callers without any speed advantage over the combined InitUpdate approach.

magnumripper:
Are we talking about same thing?
```c
$ git grep H[,2] md4.c
md4.c:#define H2(x, y, z)                       ((x) ^ ((y) ^ (z)))
md4.c:          STEP(H, a, b, c, d, GET(0) + 0x6ed9eba1, 3)
md4.c:          STEP(H2, d, a, b, c, GET(8) + 0x6ed9eba1, 9)
md4.c:          STEP(H, c, d, a, b, GET(4) + 0x6ed9eba1, 11)
[...]

PR #4486, opened 2020-11-24
Assorted fixes after testing a linux-x86-64-clang legacy build

### LINKED ISSUE #1990: SIP format in legacy clang build FAILED: (Could not find salt in db - salt() return inconsistent?)
Not that legacy clang builds are top priority, but nevertheless:

SIP format is the only one which fails in a legacy clang build using

```
make -s -f Makefile.legacy linux-x86-64-clang
```

```
Testing: SIP [MD5 32/64]... FAILED (Could not find salt in db - salt() return inconsistent?)

1 out of 342 tests have FAILED
```

Any idea what could be wrong here?
With a gcc legacy build, SIP format doesn't fail self test.

Is clang to blame? Should I test with a more recent clang version (3.7 instead of 3.5)?
But on the same system, with an autoconf clang build, SIP format passes self test.

solardiz:
This fixes #1990, implements the optimization for MD4 and MD5 I just described in #3917, and fixes a clang build warning for `ssh_common_plug.c`.

### DIFF
--- a/src/Makefile.legacy
+++ b/src/Makefile.legacy
@@ -771,10 +771,14 @@ linux-alpha-ccc:
 
 linux-sparc64:
 	$(LN) sparc64.h arch.h
-	$(MAKE) $(PROJ) \
+	@echo "#define JOHN_BLD" '"'$@'"' > john_build_rule.h
+	$(MAKE_ORIG) $(PROJ) \
 		JOHN_OBJS="$(JOHN_OBJS) c3_fmt.o" \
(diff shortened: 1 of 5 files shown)


solardiz:
Not related to clang, but I've just pushed in here one more commit "Makefile.legacy: Update linux-sparc64 for jumbo" - this was forgotten in the 1.9.0 core merge. Since CI was already successful (except for more-tests and Travis, which were not ready yet) and this commit is unrelated to anything we test with CI, I'll go ahead and merge all 4 commits now.

PR #4490, opened 2020-11-24
only define swap32 and swap64 if they are not yet already defined.

buzzdeee:
At least on OpenBSD, endian.h is pulled in, which already defines swap32 and swap64, and therefore build breaks.
So only redefine it when it's not already defined.

### DIFF
--- a/src/int-util.h
+++ b/src/int-util.h
@@ -119,6 +119,11 @@ static inline uint32_t div128_32(uint64_t dividend_hi, uint64_t dividend_lo, uin
 static inline uint32_t ident32(uint32_t x) { return x; }
 static inline uint64_t ident64(uint64_t x) { return x; }
 
+#undef swap32
+#define swap32 john_swap32
+#undef swap64
+#define swap64 john_swap64
+
 static inline uint32_t swap32(uint32_t x) {
   x = ((x & 0x00ff00ff) << 8) | ((x & 0xff00ff00) >> 8);
   return (x << 16) | (x >> 16);


src/int-util.h
```diff
@@ -119,15 +119,19 @@ static inline uint32_t div128_32(uint64_t dividend_hi, uint64_t dividend_lo, uin
...
   return (x << 16) | (x >> 16);
 }
+#endif
+#ifndef swap64
 static inline uint64_t swap64(uint64_t x) {
   x = ((x & 0x00ff00ff00ff00ff) <<  8) | ((x & 0xff00ff00ff00ff00) >>  8);
   x = ((x & 0x0000ffff0000ffff) << 16) | ((x & 0xffff0000ffff0000) >> 16);
   return (x << 32) | (x >> 32);
 }
+#endif
```
magnumripper: Maybe we should just drop these functions instead. There are alternative macros just a few lines above, and we have a `johnswap.h` that tries to define the fastest possible alternatives, with safer names.

solardiz:
I agree with magnum. Since we have to define something like these anyway, we can get better source code coverage of our testing and more consistent performance if we define just one set of these and always use it, never use the system's.

Those "alternative macros just a few lines above" (`SWAP32` and `SWAP64`) are less efficient. I think we should rewrite them using the current best approach, drop the functions (`swap32` and `swap64`), and define the macros at the end of the file to refer to the revised macros. Also, drop the `ident*` functions and just use `(x)` in those final macros.

Speaking of "the current best approach", there are ways to do these swaps with fewer different constants, see e.g. how `BF_swap` in `BF_common.c` uses just one constant for the bitmask. On some archs, each _different_ constant requires an extra "load immediate" instruction, so reusing the same constant is beneficial.

solardiz:
We also have `johnswap.h` - maybe we should `#include` and reuse its macros and functions. And clean it up, too.

solardiz:
For now, the quickest safe fix, which I'd merge, is:

```c
#undef swap32
#undef swap64
#define swap32 john_swap32
#define swap64 john_swap64
```

right before these functions are defined. This will make OpenBSD use these functions, too.

solardiz (changes requested):
Please revise this commit in the minimal way per my latest comment, which was:

```
For now, the quickest safe fix, which I'd merge, is:

#undef swap32
#undef swap64
#define swap32 john_swap32
#define swap64 john_swap64

right before these functions are defined. This will make OpenBSD use these functions, too.
```

Then we'll get it merged. A possible more invasive rework of these things can then proceed separately. Thanks!

Edit: added the `#undef` lines, considering that apparently OpenBSD already had these as macros.

solardiz:
Oh, I realized we also need to `#undef` these, considering that apparently OpenBSD already had them as macros. Edited my two comments above.

buzzdeee:
recognized that #undef needed as well, working on it

buzzdeee:
just recognized, there is no such thing like john_swap32 and john_swap64, there would bejohn_bswap_32  john_bswap_64 are you referring to those?

solardiz:
> there is no such thing like john_swap32 and john_swap64,

Of course, there are not. These macros I proposed would result in the functions currently called `swap32` and `swap64` become called these new names instead.

> there would bejohn_bswap_32 john_bswap_64 are you referring to those?

No.

buzzdeee:
thanks, now I see, mixed that up with other mentioning of potentially including  johnswap.h

src/int-util.h
```diff
@@ -119,6 +119,11 @@ static inline uint32_t div128_32(uint64_t dividend_hi, uint64_t dividend_lo, uin
 static inline uint32_t ident32(uint32_t x) { return x; }
 static inline uint64_t ident64(uint64_t x) { return x; }
 
+#undef swap32
+#define swap32 swap32
+#undef swap64
+#define swap64 swap64
```
solardiz: You missed `john_` there.

buzzdeee: sorry for this messup, built a different tree than I pushed, should be fixed now.

solardiz (approved):

PR #4491, opened 2020-11-24
getopt returns int, not signed chars.

buzzdeee:
not much to say about, getopt returns int, so declare c as int.

### DIFF
--- a/src/cprepair.c
+++ b/src/cprepair.c
@@ -241,7 +241,7 @@ static int process_file(char *name)
 
 int main(int argc, char **argv)
 {
-	signed char c;
+	int c;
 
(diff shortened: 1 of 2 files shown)


magnumripper (approved):
Strange. I guess I nicked snippets from tgtsnarf (which is much older IIRC) when I wrote cprepair :-/

solardiz:
I wonder why we call that variable `c` (is that by any chance for `char`?), but we also have that naming in several other programs (which have it as `int c;` already). So let's merge this PR as-is.

One program where we have a different (better?) name for this variable is `eapmd5tojohn.c`, where it's called `opt`.

PR #4494, opened 2020-11-28
Add support for RACF RVARY command password hashes

davidegirardi:
Support cracking hashes found by tools like [Mainframed's ENUM](https://github.com/mainframed/Enumeration/blob/master/ENUM).

If a mainframe runs RACF, there are 2 subcommands to check the status of the RACF database and to switch between the primary and secondary databases. This PR enables JrR to crack those hashes.

### DIFF
--- a/src/rvary_fmt_plug.c
+++ b/src/rvary_fmt_plug.c
@@ -53,80 +53,80 @@ john_register_one(&fmt_rvary);
 #endif
 
 static const unsigned char a2e[256] = {
-	0,  1,  2,  3, 55, 45, 46, 47, 22,  5, 37, 11, 12, 13, 14, 15,
-	16, 17, 18, 19, 60, 61, 50, 38, 24, 25, 63, 39, 28, 29, 30, 31,
-	64, 79,127,123, 91,108, 80,125, 77, 93, 92, 78,107, 96, 75, 97,
-	240,241,242,243,244,245,246,247,248,249,122, 94, 76,126,110,111,
-	124,193,194,195,196,197,198,199,200,201,209,210,211,212,213,214,
-	215,216,217,226,227,228,229,230,231,232,233, 74,224, 90, 95,109,
-	121,129,130,131,132,133,134,135,136,137,145,146,147,148,149,150,
-	151,152,153,162,163,164,165,166,167,168,169,192,106,208,161,  7,
-	32, 33, 34, 35, 36, 21,  6, 23, 40, 41, 42, 43, 44,  9, 10, 27,
-	48, 49, 26, 51, 52, 53, 54,  8, 56, 57, 58, 59,  4, 20, 62,225,
-	65, 66, 67, 68, 69, 70, 71, 72, 73, 81, 82, 83, 84, 85, 86, 87,
-	88, 89, 98, 99,100,101,102,103,104,105,112,113,114,115,116,117,
-	118,119,120,128,138,139,140,141,142,143,144,154,155,156,157,158,
-	159,160,170,171,172,173,174,175,176,177,178,179,180,181,182,183,
-	184,185,186,187,188,189,190,191,202,203,204,205,206,207,218,219,
-	220,221,222,223,234,235,236,237,238,239,250,251,252,253,254,255
+    0,  1,  2,  3, 55, 45, 46, 47, 22,  5, 37, 11, 12, 13, 14, 15,
+    16, 17, 18, 19, 60, 61, 50, 38, 24, 25, 63, 39, 28, 29, 30, 31,
+    64, 79,127,123, 91,108, 80,125, 77, 93, 92, 78,107, 96, 75, 97,
+    240,241,242,243,244,245,246,247,248,249,122, 94, 76,126,110,111,
+    124,193,194,195,196,197,198,199,200,201,209,210,211,212,213,214,
+    215,216,217,226,227,228,229,230,231,232,233, 74,224, 90, 95,109,
+    121,129,130,131,132,133,134,135,136,137,145,146,147,148,149,150,
+    151,152,153,162,163,164,165,166,167,168,169,192,106,208,161,  7,
+    32, 33, 34, 35, 36, 21,  6, 23, 40, 41, 42, 43, 44,  9, 10, 27,
+    48, 49, 26, 51, 52, 53, 54,  8, 56, 57, 58, 59,  4, 20, 62,225,
+    65, 66, 67, 68, 69, 70, 71, 72, 73, 81, 82, 83, 84, 85, 86, 87,
+    88, 89, 98, 99,100,101,102,103,104,105,112,113,114,115,116,117,
+    118,119,120,128,138,139,140,141,142,143,144,154,155,156,157,158,
+    159,160,170,171,172,173,174,175,176,177,178,179,180,181,182,183,
+    184,185,186,187,188,189,190,191,202,203,204,205,206,207,218,219,
+    220,221,222,223,234,235,236,237,238,239,250,251,252,253,254,255
 };
 
 /* This is a2e[] with each entry XOR 0x55, left-shifted one bit
    and finally with odd parity so that DES_set_key_unchecked
    can be used directly.  This provides about 15% speed up.    */
 static const unsigned char a2e_precomputed[256] = {
-	 171, 168, 174, 173, 196, 241, 247, 244, 134, 161, 224, 188, 179, 176, 182, 181,
-	 138, 137, 143, 140, 211, 208, 206, 230, 155, 152, 213, 229, 146, 145, 151, 148,
-	  42,  52,  84,  93,  28, 115,  11,  81,  49,  16,  19,  55, 124, 107,  61, 104,
-	  74,  73,  79,  76,  67,  64,  70,  69,  91,  88,  94,  22,  50,  87, 118, 117,
-	  82,  41,  47,  44,  35,  32,  38,  37,  59,  56,   8,  14,  13,   2,   1,   7,
-	   4,  26,  25, 110, 109,  98,  97, 103, 100, 122, 121,  62, 107,  31,  21, 112,
-	  88, 168, 174, 173, 162, 161, 167, 164, 186, 185, 137, 143, 140, 131, 128, 134,
-	 133, 155, 152, 239, 236, 227, 224, 230, 229, 251, 248,  42, 127,  11, 233, 164,
-	 234, 233, 239, 236, 227, 128, 167, 133, 251, 248, 254, 253, 242, 185, 191, 157,
-	 203, 200, 158, 205, 194, 193, 199, 186, 218, 217, 223, 220, 162, 131, 214, 104,
-	  41,  47,  44,  35,  32,  38,  37,  59,  56,   8,  14,  13,   2,   1,   7,   4,
-	  26,  25, 110, 109,  98,  97, 103, 100, 122, 121,  74,  73,  79,  76,  67,  64,
-	  70,  69,  91, 171, 191, 188, 179, 176, 182, 181, 138, 158, 157, 146, 145, 151,
-	 148, 234, 254, 253, 242, 241, 247, 244, 203, 200, 206, 205, 194, 193, 199, 196,
-	 218, 217, 223, 220, 211, 208, 214, 213,  62,  61,  50,  49,  55,  52,  31,  28,
-	  19,  16,  22,  21, 127, 124, 115, 112, 118, 117,  94,  93,  82,  81,  87,  84
+    171, 168, 174, 173, 196, 241, 247, 244, 134, 161, 224, 188, 179, 176, 182, 181,
+    138, 137, 143, 140, 211, 208, 206, 230, 155, 152, 213, 229, 146, 145, 151, 148,
+    42,  52,  84,  93,  28, 115,  11,  81,  49,  16,  19,  55, 124, 107,  61, 104,
+    74,  73,  79,  76,  67,  64,  70,  69,  91,  88,  94,  22,  50,  87, 118, 117,
+    82,  41,  47,  44,  35,  32,  38,  37,  59,  56,   8,  14,  13,   2,   1,   7,
+    4,  26,  25, 110, 109,  98,  97, 103, 100, 122, 121,  62, 107,  31,  21, 112,
+    88, 168, 174, 173, 162, 161, 167, 164, 186, 185, 137, 143, 140, 131, 128, 134,
+    133, 155, 152, 239, 236, 227, 224, 230, 229, 251, 248,  42, 127,  11, 233, 164,
+    234, 233, 239, 236, 227, 128, 167, 133, 251, 248, 254, 253, 242, 185, 191, 157,
+    203, 200, 158, 205, 194, 193, 199, 186, 218, 217, 223, 220, 162, 131, 214, 104,
+    41,  47,  44,  35,  32,  38,  37,  59,  56,   8,  14,  13,   2,   1,   7,   4,
+    26,  25, 110, 109,  98,  97, 103, 100, 122, 121,  74,  73,  79,  76,  67,  64,
+    70,  69,  91, 171, 191, 188, 179, 176, 182, 181, 138, 158, 157, 146, 145, 151,
+    148, 234, 254, 253, 242, 241, 247, 244, 203, 200, 206, 205, 194, 193, 199, 196,
+    218, 217, 223, 220, 211, 208, 214, 213,  62,  61,  50,  49,  55,  52,  31,  28,
+    19,  16,  22,  21, 127, 124, 115, 112, 118, 117,  94,  93,  82,  81,  87,  84
 };
 
 /* in-place ascii2ebcdic conversion */
 static void ascii2ebcdic(unsigned char *str)
 {
-	int i;
-	int n = strlen((const char*)str);
-	for (i = 0; i < n; ++i)
-		str[i] = a2e[str[i]];
+    int i;
+    int n = strlen((const char*)str);
+    for (i = 0; i < n; ++i)
+        str[i] = a2e[str[i]];
 }
 
 /* Pad input string with EBCDIC spaces (0x40) up to eight bytes */
 static void ebcdic_padding(unsigned char *str)
 {
-	int i;
-	for (i = strlen((const char*)str); i < 9; ++i)
-		str[i] = 0x40;
-	str[9] = 0; /* terminate string */
+    int i;
+    for (i = strlen((const char*)str); i < 9; ++i)
+        str[i] = 0x40;
+    str[9] = 0; /* terminate string */
 }
 
 #ifdef RVARY_DEBUG
 static void print_hex(unsigned char *str, int len)
 {
-	int i;
-	for (i = 0; i < len; ++i)
-		printf("%02x", str[i]);
-	printf("\n");
+    int i;
+    for (i = 0; i < len; ++i)
+        printf("%02x", str[i]);
+    printf("\n");
 }
 #endif
 
 static struct fmt_tests rvary_tests[] = {
-	{"$rvary$**5AA70358A9C369E0", "QWERTY1"},
-	{"$rvary$**EB59EE07FC74AE77", "PASSWORD"},
-	{"$rvary$**062314297C496E0E", "AAAAAAAA"},
-	{"$rvary$**0FF48804F759193F", "TESTTEST"},
-	{NULL}
+    {"$rvary$**5AA70358A9C369E0", "QWERTY1"},
+    {"$rvary$**EB59EE07FC74AE77", "PASSWORD"},
+    {"$rvary$**062314297C496E0E", "AAAAAAAA"},
+    {"$rvary$**0FF48804F759193F", "TESTTEST"},
+    {NULL}
 };
 
 static char (*saved_key)[PLAINTEXT_LENGTH + 1];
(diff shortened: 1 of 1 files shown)


solardiz:
@davidegirardi Thank you for your contribution! Please note this is already failed some builds on our CI - please look into those failures and fix the code to make it pass all tests.

I'd also like to hear from @mainframed if possible -  should we really have this? As a separate format?

Looking at what this does, I think it should be easy to make it use our bitslice DES implementation instead of OpenSSL's non-bitslice one. But let's approach this separately.

src/rvary_fmt_plug.c
```diff
@@ -0,0 +1,325 @@
...
+#include <omp.h>
+#endif
+
+#include "arch.h"
+#include "crc32.h"
+#include "misc.h"
+#include "common.h"
+#include "formats.h"
+#include "params.h"
+#include "options.h"
```
solardiz, line 32: What do we use here from these header files? At least `crc32.h` is probably in here by mistake (copy-paste from somewhere?), and probably several others are too.

src/rvary_fmt_plug.c
```diff
@@ -0,0 +1,325 @@
...
+#define FORMAT_TAG              "$rvary$*"
+#define FORMAT_TAG_LEN          (sizeof(FORMAT_TAG)-1)
+#define ALGORITHM_NAME          "DES 32/" ARCH_BITS_STR
+#define BENCHMARK_COMMENT       ""
+#define BENCHMARK_LENGTH        7
+#define PLAINTEXT_LENGTH        8
+#define CIPHERTEXT_LENGTH       16
+#define BINARY_SIZE             8
+#define SALT_SIZE               0
+#define BINARY_ALIGN            sizeof(uint32_t)
```
solardiz, line 45: We're casting those to `DES_cblock`, so are we guessing its required alignment here?

src/rvary_fmt_plug.c
```diff
@@ -0,0 +1,325 @@
...
+#define SALT_ALIGN              1
+
+#define MIN_KEYS_PER_CRYPT      1
+#define MAX_KEYS_PER_CRYPT      256
+
+#ifndef OMP_SCALE
+#define OMP_SCALE               4 // Tuned w/ MKPC for super
+#endif
```
solardiz, line 53: Of course, the comment about tuning is untrue, and should be dropped until/unless such tuning is (re)done for this specific format.

src/rvary_fmt_plug.c
```diff
@@ -0,0 +1,325 @@
...
+	88, 89, 98, 99,100,101,102,103,104,105,112,113,114,115,116,117,
+	118,119,120,128,138,139,140,141,142,143,144,154,155,156,157,158,
+	159,160,170,171,172,173,174,175,176,177,178,179,180,181,182,183,
+	184,185,186,187,188,189,190,191,202,203,204,205,206,207,218,219,
+	220,221,222,223,234,235,236,237,238,239,250,251,252,253,254,255
+};
+
+/* This is a2e[] with each entry XOR 0x55, left-shifted one bit
+   and finally with odd parity so that DES_set_key_unchecked
+   can be used directly.  This provides about 15% speed up.    */
```
solardiz, line 76: This is basically a workaround for using OpenSSL, trying to speed it up just a little bit. We wouldn't need this, and would have much greater speedup, by using our own bitslice DES code instead.

src/rvary_fmt_plug.c
```diff
@@ -0,0 +1,325 @@
...
+};
+
+/* in-place ascii2ebcdic conversion */
+static void ascii2ebcdic(unsigned char *str)
+{
+	int i;
+	int n = strlen((const char*)str);
+	for (i = 0; i < n; ++i)
+		str[i] = a2e[str[i]];
+}
```
solardiz, line 103: The use of `strlen()` here is weird. We could instead use `str[i]` in place of `i < n`, or better yet use a pointer in place of `i`.

src/rvary_fmt_plug.c
```diff
@@ -0,0 +1,325 @@
...
+	{"$rvary$**5AA70358A9C369E0", "QWERTY1"},
+	{"$rvary$**EB59EE07FC74AE77", "PASSWORD"},
+	{"$rvary$**062314297C496E0E", "AAAAAAAA"},
+	{"$rvary$**0FF48804F759193F", "TESTTEST"},
+	{NULL}
+};
+
+static char (*saved_key)[PLAINTEXT_LENGTH + 1];
+static uint32_t (*crypt_out)[BINARY_SIZE / sizeof(uint32_t)];
+static DES_key_schedule (*schedules);
```
solardiz, line 134: Please move the tests vectors list and these variables to right after the `#define` lines.

src/rvary_fmt_plug.c
```diff
@@ -0,0 +1,325 @@
...
+static int dirty;
```
solardiz, line 135: Since this format is unsalted, I see no need for the `dirty` logic here.

src/rvary_fmt_plug.c
```diff
@@ -0,0 +1,325 @@
...
+	ctcopy += 1; /* Skip empty salt */
+	if (hexlenu(ctcopy, &extra) != CIPHERTEXT_LENGTH || extra)
+		goto err;
+	MEM_FREE(keeptr);
+	return 1;
+
+err:
+	MEM_FREE(keeptr);
+	return 0;
+}
```
solardiz, line 175: Weird stuff. First, is the salt guaranteed to be empty? If so, why do we even have a field for it in the hash encodings?

If we somehow have to have it that way (for compatibility with an existing tool?), maybe we can have both asterisks as part of `FORMAT_TAG`?

We definitely shouldn't blindly skip a character here without validating that it matches our expectations?

Also, the `strdup()` stuff and related complexity here doesn't make sense. Perhaps it didn't make sense from whatever other format this was copy-paste from, and should be cleaned up in there as well.

src/rvary_fmt_plug.c
```diff
@@ -0,0 +1,325 @@
...
+
+static void *get_binary(char *ciphertext)
+{
+	static union {
+		unsigned char c[BINARY_SIZE+1];
+		ARCH_WORD dummy;
+	} buf;
```
solardiz, line 182: The `+1` looks unreasonable - we're not producing C string here.

The alignment to `ARCH_WORD` looks excessive - if we set our `BINARY_ALIGN` to that of `uint32_t` and access our binaries accordingly, that's what we should align to here.

src/rvary_fmt_plug.c
```diff
@@ -0,0 +1,325 @@
...
+	unsigned char *out = buf.c;
+	char *p;
+	int i;
+
+	p = strrchr(ciphertext, '*') + 1;
+		for (i = 0; i < BINARY_SIZE; i++) {
```
solardiz: Inconsistent indentation.

src/rvary_fmt_plug.c
```diff
@@ -0,0 +1,325 @@
...
+		if (dirty) {
+			DES_cblock des_key;
+			int i;
+
+			/* process key */
+			for (i = 0; saved_key[index][i]; i++)
+				des_key[i] = a2e_precomputed[ARCH_INDEX(saved_key[index][i])];
+
+			/* replace missing characters in userid by (EBCDIC space (0x40) XOR 0x55) << 1 */
+			while(i < 9)
```
solardiz: Missing whitespace.

src/rvary_fmt_plug.c
```diff
@@ -0,0 +1,325 @@
...
+#pragma omp parallel for
+#endif
+	for (index = 0; index < count; index++) {
+		if (dirty) {
+			DES_cblock des_key;
+			int i;
+
+			/* process key */
+			for (i = 0; saved_key[index][i]; i++)
+				des_key[i] = a2e_precomputed[ARCH_INDEX(saved_key[index][i])];
```
solardiz, line 215: The `ARCH_INDEX` stuff is legacy, we should probably use a cast to `unsigned char` here.

src/rvary_fmt_plug.c
```diff
@@ -0,0 +1,325 @@
...
+
+			/* replace missing characters in userid by (EBCDIC space (0x40) XOR 0x55) << 1 */
+			while(i < 9)
+				des_key[i++] = 0x2a;
+
+			DES_set_key_unchecked(&des_key, &schedules[index]);
+		}
+		/* DES encrypt the password with the password itself by using the key as salt */
+        char key[10];
```
solardiz: Inconsistent indentation starting with this line.

src/rvary_fmt_plug.c
```diff
@@ -0,0 +1,325 @@
...
+static int crypt_all(int *pcount, struct db_salt *salt)
+{
+	const int count = *pcount;
+	int index;
+
+#ifdef _OPENMP
+#pragma omp parallel for
+#endif
+	for (index = 0; index < count; index++) {
+		if (dirty) {
```
solardiz, line 209: Again, the `dirty` logic is unneeded. When you drop it, please also drop the global storage of `schedules` - only one key schedule is needed for the current iteration.

src/rvary_fmt_plug.c
```diff
@@ -0,0 +1,325 @@
...
+			/* replace missing characters in userid by (EBCDIC space (0x40) XOR 0x55) << 1 */
+			while(i < 9)
+				des_key[i++] = 0x2a;
+
+			DES_set_key_unchecked(&des_key, &schedules[index]);
+		}
+		/* DES encrypt the password with the password itself by using the key as salt */
+        char key[10];
+        strnzcpy(key, saved_key[index], 9);
+        key[9] = 0;
```
solardiz: `key[9] = 0;` should be unneeded.

src/rvary_fmt_plug.c
```diff
@@ -0,0 +1,325 @@
...
+        ascii2ebcdic((unsigned char *)key);
+        ebcdic_padding((unsigned char *)key);
```
solardiz: Perhaps the casts to/from `unsigned char *` can be avoided by having just one cast to `unsigned char` on the array indexing inside `ascii2ebcdic()`.

src/rvary_fmt_plug.c
```diff
@@ -0,0 +1,325 @@
...
+
+	return count;
+}
+
+static int cmp_all(void *binary, int count)
+{
+	int index;
+
+	for (index = 0; index < count; index++)
+		if (!memcmp(binary, crypt_out[index], ARCH_SIZE))
```
solardiz, line 240: Can just use `BINARY_SIZE` here.

solardiz (changes requested):
Please make this pass our CI tests, and at least consider fixing the issues I point out here.

src/rvary_fmt_plug.c
davidegirardi: This code is an adaptation of the already present `racf_*_fmt_plug.c` files. The workaround comes from those. The same applies for all the other comments in this PR (which I am switching to WIP mode if I find how).

I started working on an alternative which supports more codepages (this shift is for the US codepage), I guess that using your bitslice DES is the way to go then.

davidegirardi:
> Please make this pass our CI tests, and at least consider fixing the issues I point out here.

I will look into it at the best of my abilities. Not a C programmer here.
The CI tests fail on code that is already present in the RACF format cracker, so I am a bit confused about this.

solardiz:
Copy-paste from CIrcle CI for us all:

```
Testing: RVARY [DES 32/64]... (2xOMP) =================================================================
==5681==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7fff0fa41668 at pc 0x00000082e91e bp 0x7fff0fa41620 sp 0x7fff0fa41610
WRITE of size 1 at 0x7fff0fa41668 thread T0
    #0 0x82e91d in crypt_all._omp_fn.0 /home/JtR/project/src/rvary_fmt_plug.c:221
    #1 0x7fdb61471f45 in GOMP_parallel (/lib64/libgomp.so.1+0x12f45)
    #2 0x82e0a9 in crypt_all /home/JtR/project/src/rvary_fmt_plug.c:208
    #3 0x8c8264 in is_key_right /home/JtR/project/src/formats.c:518
    #4 0x8d0cba in fmt_self_test_body /home/JtR/project/src/formats.c:1210
    #5 0x8d1dc5 in fmt_self_test /home/JtR/project/src/formats.c:1938
    #6 0x8af380 in benchmark_all /home/JtR/project/src/bench.c:907
    #7 0x8deced in john_run /home/JtR/project/src/john.c:1621
    #8 0x8deced in main /home/JtR/project/src/john.c:2032
    #9 0x7fdb6129a041 in __libc_start_main (/lib64/libc.so.6+0x27041)
    #10 0x4061ed in _start (/home/JtR/project/run/john+0x4061ed)

Address 0x7fff0fa41668 is located in stack of thread T0 at offset 40 in frame
    #0 0x82e45b in crypt_all._omp_fn.0 /home/JtR/project/src/rvary_fmt_plug.c:208

  This frame has 2 object(s):
    [32, 40) 'des_key' (line 212) <== Memory access at offset 40 overflows this variable
    [64, 74) 'key' (line 226)
HINT: this may be a false positive if your program uses some custom stack unwind mechanism, swapcontext or vfork
      (longjmp and C++ exceptions *are* supported)
SUMMARY: AddressSanitizer: stack-buffer-overflow /home/JtR/project/src/rvary_fmt_plug.c:221 in crypt_all._omp_fn.0
  0x100061f40270: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x100061f40280: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x100061f40290: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x100061f402a0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x100061f402b0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
=>0x100061f402c0: 00 00 00 00 00 00 00 00 f1 f1 f1 f1 00[f2]f2 f2
  0x100061f402d0: 00 02 f3 f3 00 00 00 00 00 00 00 00 00 00 00 00
  0x100061f402e0: 00 00 f1 f1 f1 f1 04 f3 f3 f3 00 00 00 00 00 00
  0x100061f402f0: 00 00 00 00 00 00 f1 f1 f1 f1 04 f3 f3 f3 00 00
  0x100061f40300: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x100061f40310: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 f1 f1
  Addressable:           00
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
  Shadow gap:              cc
==5681==ABORTING

Exited with code exit status 1

CircleCI received exit code 1
```

src/rvary_fmt_plug.c
```diff
@@ -0,0 +1,325 @@
...
+			DES_cblock des_key;
+			int i;
+
+			/* process key */
+			for (i = 0; saved_key[index][i]; i++)
+				des_key[i] = a2e_precomputed[ARCH_INDEX(saved_key[index][i])];
+
+			/* replace missing characters in userid by (EBCDIC space (0x40) XOR 0x55) << 1 */
+			while(i < 9)
+				des_key[i++] = 0x2a;
```
solardiz, line 219: This is where ASan complains. The `while` loop allows for `i` up to 8 inclusive, but `des_key` is a `DES_cblock` and I guess is exactly 8 bytes. So we write one byte beyond the buffer end here.

solardiz: And indeed `racf_fmt_plug.c` has the check as `< 8`. Where did the 9 come from?

src/rvary_fmt_plug.c
solardiz: > The same applies for all the other comments in this PR

Maybe for most but not for all. I just took a look at `racf_fmt_plug.c` and it indeed shares much of this, but not all of the issues I pointed out are present there. For example, that format's use of `dirty` actually makes sense because that format is salted, and there's no inconsistent code indentation in that source file.

src/rvary_fmt_plug.c
davidegirardi: I see. When RACF encrypts the user password with the "traditional" DES implementation, it's using the username (up to 7 chars) as salt.
RVARY instead encrypts the password with the password itself as salt. Easy fix then.

src/rvary_fmt_plug.c
davidegirardi: The indentation issue is because I tried to keep the common lines untouched, to make diffing easier for you guys.
Another easy fix, just a `=G` and maybe a `:retab`.

Thank you for taking the time to guide me.

src/rvary_fmt_plug.c
davidegirardi: Fixed in 9a5fad3

src/rvary_fmt_plug.c
davidegirardi: Fixed in 6182a4a

src/rvary_fmt_plug.c
davidegirardi: I think we should address this when start using the bitslice approach. That will also include refactoring the current RACF and RACF KDFAES code so we can support all/most codepages too.

src/rvary_fmt_plug.c
davidegirardi: Fixed in ab884c9

src/rvary_fmt_plug.c
davidegirardi: Fixed (?) in 66be25f

src/rvary_fmt_plug.c
davidegirardi: Fixed in 160fa9d

src/rvary_fmt_plug.c
davidegirardi: Fixed in fb1b2d9

src/rvary_fmt_plug.c
davidegirardi: Fixed in aa66ef1

src/rvary_fmt_plug.c
davidegirardi: Fixed in a144262

src/rvary_fmt_plug.c
davidegirardi: Fixed in 963ac2e

src/rvary_fmt_plug.c
davidegirardi: Fixed in 66be25f

src/rvary_fmt_plug.c
solardiz: Now this is even more obviously wrong than it was before. :-( Maybe `options.h` _happens_ to include all that's needed, but that doesn't make it the right file to include here.

src/rvary_fmt_plug.c
```diff
@@ -56,7 +56,6 @@ static struct fmt_tests rvary_tests[] = {
 static char (*saved_key)[PLAINTEXT_LENGTH + 1];
 static uint32_t (*crypt_out)[BINARY_SIZE / sizeof(uint32_t)];
 static DES_key_schedule (*schedules);
-static int dirty;
```
solardiz: Great, but at the same time `schedules` become unneeded, to be replaced with a variable local to the loop.

src/rvary_fmt_plug.c
```diff
@@ -46,10 +46,10 @@ john_register_one(&fmt_rvary);
...
 
 static struct fmt_tests rvary_tests[] = {
-    {"$rvary$**5AA70358A9C369E0", "QWERTY1"},
-    {"$rvary$**EB59EE07FC74AE77", "PASSWORD"},
-    {"$rvary$**062314297C496E0E", "AAAAAAAA"},
-    {"$rvary$**0FF48804F759193F", "TESTTEST"},
+    {"$rvary$*5AA70358A9C369E0", "QWERTY1"},
+    {"$rvary$*EB59EE07FC74AE77", "PASSWORD"},
+    {"$rvary$*062314297C496E0E", "AAAAAAAA"},
+    {"$rvary$*0FF48804F759193F", "TESTTEST"},
```
solardiz: Does this also require an update to some external tool, one that produces those hashes?

src/rvary_fmt_plug.c
```diff
@@ -53,80 +53,80 @@ john_register_one(&fmt_rvary);
 #endif
 
 static const unsigned char a2e[256] = {
-	0,  1,  2,  3, 55, 45, 46, 47, 22,  5, 37, 11, 12, 13, 14, 15,
```
solardiz: Ouch. I never suggested reindenting the whole file, let alone to conventions inconsistent with what we use in the rest of the tree.

I only asked you to keep indentation of the edited blocks consistent, so that it reflects their logical nesting level.

solardiz (changes requested):
Now this passes tests, but is otherwise at least as weird as it was before. I think it'll be the simplest, yet give you due credit for your contribution, if you exclude or revert all of your today's commits except for the one 9 to 8 fix and _maybe_ the change of hash encodings dropping the empty salt from there (but please comment on whether this affects existing external tools first). Then I'll take it from there in a separate PR. Thanks!

src/rvary_fmt_plug.c
davidegirardi: The external tool is a PR I submitted to the ENUM tool. The PR is already up to date now.

solardiz: Great. Why do we have an asterisk in there at all? Let's drop it while we can. As I recall, those asterisks were originally introduced by @kholia in some "non-hash" formats to separate their many custom fields, and there's no reason to clutter a simple hash with those here.

davidegirardi: I am using the asterisks as they are what the format for the "siblings" formats use. For a reference, take a look at the other RACF formats and racf2john.

I am ready with the reverted version plus small fixes (BO, whitespaces, indentation type/size, no fake empty salt).
We need just to choose the format and I'm ready to push.

solardiz: I took a look at the sibling formats, and I think we should completely drop the asterisks from this one while we can.

davidegirardi: Dropped then :)

solardiz:
I've "squash and merged" this. Thanks!

Issue #4496, opened 2020-11-29
7z2john.pl issue undefined value as a HASH reference


adamjk-dev:
### Error Message ###
──(kali㉿kali)-[~]
└─$ /usr/share/john/7z2john.pl /mnt/shared/archive.7z
DEBUG: number_coders: 2, codec_id: 
Can't use an undefined value as a HASH reference at /usr/share/john/7z2john.pl line 1384.
codec_id is:       

### Steps to reproduce ###
Latest kali Linux downloaded today and tested.
Was trying to get the hash out of an old 7-ZIP archive.

7-Zip details:
* When I test the 7zip archive it returns this:
Type = 7z
Physical Size = 56495367
Headers Size = 4771
Method = LZMA:24 7zAES
Solid = +
Blocks = 3

* Running file on the 7zip archive: 
7-zip archive data, version 0.3

### System configuration ###
──(kali㉿kali)-[~]
└─$ john --list=build-info                                                                                                               255 ⨯
Created directory: /home/kali/.john
Version: 1.9.0-jumbo-1
Build: linux-gnu 64-bit x86_64 AVX2 AC OMP
SIMD: AVX2, interleaving: MD4:3 MD5:3 SHA1:1 SHA256:1 SHA512:1
System-wide exec: /usr/lib/john
System-wide home: /usr/share/john
Private home: ~/.john
CPU tests: AVX2
CPU fallback binary: john-xop-omp
OMP fallback binary: john-avx2-non-omp
$JOHN is /usr/share/john/
Format interface version: 14
Max. number of reported tunable costs: 4
Rec file version: REC4
Charset file version: CHR3
CHARSET_MIN: 1 (0x01)
CHARSET_MAX: 255 (0xff)
CHARSET_LENGTH: 24
SALT_HASH_SIZE: 1048576
SINGLE_IDX_MAX: 32768
SINGLE_BUF_MAX: 4294967295
Effective limit: Max. KPC 32768
Max. Markov mode level: 400
Max. Markov mode password length: 30
gcc version: 9.2.1
GNU libc version: 2.28 (loaded: 2.31)
Crypto library: OpenSSL
OpenSSL library version: 01010103f      (loaded: 01010107f)
OpenSSL 1.1.1c  28 May 2019     (loaded: OpenSSL 1.1.1g  21 Apr 2020)
GMP library version: 6.1.2      (loaded: 6.2.0)
File locking: fcntl()
fseek(): fseek
ftell(): ftell
fopen(): fopen
[...]

PR #4497, opened 2020-12-01
7z2john.pl: Avoid crash when a stream has "no methods"

magnumripper:
While there may be more to this issue (#4496) this patch can't do any harm for now.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -27,6 +27,10 @@ tab_width = 8
 indent_style = space
 indent_size = 2
 
+[7z2john.pl]
+indent_style = space
+indent_size = 2
+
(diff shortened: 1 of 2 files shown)


solardiz (approved):
I'd prefer the `.editorconfig` change to be a separate commit, even if part of this same PR.

solardiz:
I think merging this PR should actually close #4496 as otherwise that issue would just hang. We don't have any other action for it now.

magnumripper:
> I think merging this PR should actually close #4496 as otherwise that issue would just hang. We don't have any other action for it now.

I'm still hoping for Phil to chime in there for some comments :-)

PR #4498, opened 2020-12-01
Improve handling of UTF-8 rules/masks vs. internal codepage

magnumripper:
Refactor "encoding" ASCII -> ENC_RAW, because that's how it's used.
Add a few cases where "--enc=raw" should skip any conversions.

Improve handling of UTF-8 rules w/ internal codepage. The first iteration (d595de99) could not handle [the case where a conversion fails because the target codepage lack some needed character](https://github.com/openwall/john/issues/4437#issuecomment-736352695).  This new code does not touch any rule unless it starts with -U.  After conversion it checks for success and otherwise turn the -U into -- which is an "always reject" flag (new, and undocumented in doc/RULES because, well, it's of no use except for this).
Also drops a hack for "best-effort" codepage conversion that hardly ever did any good - it would break the new (and coming) heuristics.

Bugfix for -8 rule reject flag. A hashcat compatibility feature (introduced in 059d4e6e, 2018) broke the "-8" rule reject flag (would be parsed as hashcat rule "increment char at position 8" followed by whatever came next in the original rule).
Prune a few redundant comments while at it.

Mask mode: Improve handling of UTF-8 vs. internal codepage. Bail with error instead of chugging along if mask given in UTF-8 doesn't fit selected codepage.  Likewise for custom mask placeholders, but only provided the placeholder is actually used. Also tweaks many existing mask error messages, for consistence.


src/rules.c
```diff
@@ -711,11 +711,18 @@ char *rules_reject(char *rule, int split, char *last, struct db_main *db)
...
-			if (options.internal_cp != UTF_8) continue;
+			if (options.internal_cp != UTF_8 && options.internal_cp != ENC_RAW) continue;
+			return NULL;
+
+/*
+ * Any failed UTF-8-to-codepage translations change '-U' to '--' in
+ * rpp_process_rule(), as an "always reject" hack.
+ */
+		case '-':
 			return NULL;
```
magnumripper, line 726: @solardiz in my own view this is the only arguable part of this PR. I simply couldn't find a better way to handle a rule rejection that happens in `rpp_process_rule()` that would end up nicely.

solardiz:
@magnumripper BTW, in another recent (merged) PR of yours, you added this comment: "-internal-encoding is a deprecated alias for -internal-codepage". Why do you think we should prefer to call that one thing codepage rather than encoding?

magnumripper:
> @magnumripper BTW, in another recent (merged) PR of yours, you added this comment: "-internal-encoding is a deprecated alias for -internal-codepage".

That comment was old, I just added a note of when we can actually drop the support for it.

> Why do you think we should prefer to call that one thing codepage rather than encoding?

Because a legacy codepage is required for it to do any good: Using UTF-8 (which is obviously an encoding but definitely not a codepage) is allowed but technically it's more or less the same as using "raw".

On a side note though, I have some really exciting ideas that (provided they actually fly) will improve conversion transparency, decrease problems with codepage limitations and overall hide more of the encoding mysteries from the user's eyes.

src/rules.c
magnumripper: BTW I almost added this "-- always reject" to doc/RULES but decided it's of no interest to the user. We could add it anyway just for "documenting" it but OTOH as long as we don't we are free to change the flag to something else or even change this "hack" to something completely different and drop this new flag.

doc/ENCODINGS
```diff
@@ -61,14 +61,15 @@ If you use rules without --internal-codepage, some wordlist rules may cut
...
 on command-line or in a config file will be silently translated to whatever
-internal codepage is in use.  This will normally result in just what you'd
-expect but it could sometimes lead to confusion.  For example, if you run that
-"scç" rule with the internal codepage set to CP866, the "ç" can't be translated
-and will currently lead to some other undefined substitution with no warning.
+internal codepage is in use at the time.  For this to happen with rules, you
+must prepend them with "-U".  If such conversion fails for a rule, that rule
+will be rejected (silently, but logged).  Thus you can have rules for various
+codepages such as "-U sRЯ" and "-U sc©" in place - and only one of them, both,
+or none will be used, depending on what internal codepage is active.
```
solardiz: Since the functionality is now conditional rather than always enabled (right?), perhaps the "will" in "will be silently translated" should be changed to "can". Also, maybe replace "translated" with "converted" for consistency with the word "conversion" in what follows. Then you describe what it takes to enable the functionality for rules. You can add at the end that this conversion always happens for masks and placeholders (does it?)

src/mask.c
```diff
@@ -1437,9 +1442,9 @@ static void truncate_mask(mask_cpu_context *cpu_mask_ctx, int range_idx)
 
 	if (range_idx < mask_max_skip_loc && mask_max_skip_loc != -1) {
 		if (john_main_process)
-			fprintf(stderr, "Format internal ranges (first %d range positions)"
-			        " cannot be truncated!\n"
-			        "Use a larger min-length, or non-gpu format.\n",
+			fprintf(stderr,
+			        "Error: Format's internal mask ranges (first %d positions) cannot be truncated!\n"
+			        "       Increase minimum length, or use the CPU format.\n",
```
solardiz: This implies we don't have CPU formats with internal mask, but I think ideally we should add that to the LM format. Maybe make the ", or use the CPU format" portion conditional on "opencl" or "ztex" in the format name?

src/mask.c
```diff
@@ -2215,13 +2220,31 @@ void mask_init(struct db_main *db, char *unprocessed_mask)
...
+				error();
+			}
+		}
+
+		for (i = 0; i < MAX_NUM_CUST_PLHDR; i++) {
+			if (valid_utf8((UTF8*)options.custom_mask[i]) > 1) {
+				int u8_length = strlen8((UTF8*)options.custom_mask[i]);
+				char *tmp_buf;
+
+				tmp_buf = mem_alloc(strlen(options.custom_mask[i]));
```
solardiz: Shouldn't it be `strlen(...) + 1`?

solardiz (approved):
There are aspects here that I'm not familiar with, so I can't review this fully, but I skimmed through the changes and made some comments.

solardiz:
> > Why do you think we should prefer to call that one thing codepage rather than encoding?
> 
> Because a legacy codepage is required for it to do any good: Using UTF-8 (which is obviously an encoding but definitely not a codepage) is allowed but technically it's more or less the same as using "raw".

I think we should call it encoding anyway. The word "codepage" is a DOS/Windows thing (edit: whereas ISO* are encodings and are also relevant here). I think we should obsolete `--internal-codepage` in favor of `--internal-encoding`.

src/mask.c
magnumripper: The conversion is guaranteed at this point to be one or more bytes shorter, but we might want to add a comment.

solardiz: Yes, at least a comment would help.

Also, I'd prefer to have and consistently use APIs accepting the destination buffer size and refusing to overflow the buffer, except possibly in performance-critical places.

src/rules.c
solardiz: I agree it's right to leave this hack undocumented.

magnumripper:
> > > Why do you think we should prefer to call that one thing codepage rather than encoding?
> > 
> > 
> > Because a legacy codepage is required for it to do any good: Using UTF-8 (which is obviously an encoding but definitely not a codepage) is allowed but technically it's more or less the same as using "raw".
> 
> I think we should call it encoding anyway. The word "codepage" is a DOS/Windows thing (edit: whereas ISO* are encodings and are also relevant here). I think we should obsolete `--internal-codepage` in favor of `--internal-encoding`.

I believe the word was used before Microsoft existed. But anyway, in my world different codepages means we put different characters in the same (usually byte) positions, a subset of possible characters. ISO encodings are codepages just like DOS/Windows ones are also encodings: Any codepage can also be called an encoding but the reverse is not necessarily true - UTF-8 doesn't map any subset, it unambiguously encodes characters.

But that's me. Anyway for use by mask mode or rules, we really need a single-byte codepage and I wanted to emphasize that. Even a full blown 100% Unicode-aware cracker would probably benefit from the notion of a codepage (be it called locale) for limiting the otherwise huge outcome of "all uppercase characters". It would be very weird to call *that* an encoding.

src/mask.c
magnumripper: You mean the strcpy? It's simply **known** to be safe at that point, but I can change it anyway - it's not performance critical at all.

doc/ENCODINGS
magnumripper: I will revisit this - don't merge yet.

src/mask.c
magnumripper: Very few formats have this limitation, only Sayantan's "worst" that I couldn't fix. I changed the wording to more neutral

doc/ENCODINGS
magnumripper: Rewritten, and separated rules from masks better

src/mask.c
magnumripper: Fixed

magnumripper:
Ready for merge if/when you are happy

solardiz:
> I believe the word was used before Microsoft existed.

You seem to be right. https://en.wikipedia.org/wiki/Code_page has some interesting stuff.

PR #4500, opened 2020-12-03
fix zsh autocomplete file

Cal1z:
there were some quotes and closing brackets missing in the zsh autocomplete file, breaking autocomplete.

### DIFF
--- a/run/john.zsh_completion
+++ b/run/john.zsh_completion
@@ -27,8 +27,8 @@ _arguments -C -s \
   '(--prince-case-permute)'--prince-case-permute'[permute case of first letter]' \
   '(--prince-mmap)'--prince-mmap'[memory-map infile (not available with case permute)]' \
   '(--prince-keyspace)'--prince-keyspace'[just show total keyspace that would be produced]' \
-  '(--encoding)'{--encoding=-,--input-encoding=-}[input data is non-ascii (eg. UTF-8, ISO-8859-1)]:encoding to use:->list-encodings' \
-  '(--rules,-r,--stdin)'{--rules=-,-r=-,-r\:-}'[use rule]:rule to use:->list-rules' \
+  '(--encoding)'{--encoding=-,--input-encoding=-}'[input data is non-ascii (eg. UTF-8, ISO-8859-1)]:encoding to use:->list-encodings' \
+  '(--rules,-r,--stdin)'{--rules=-,-r=-}'[use rule]:rule to use:->list-rules' \
   '(--rules-stack)'--rules-stack='[stacked rules]' \
   '(--rules-skip-nop)'--rules-skip-nop='[skip any NOP ":" rules (you already ran w/o rules]' \
   '(--mask)'--mask='[mask mode using MASK (or default from john.conf)]' \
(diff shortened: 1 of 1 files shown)


run/john.zsh_completion
```diff
@@ -48,19 +48,19 @@ _arguments -C -s \
...
-  '(--salts'--salts='[load salts with(out) COUNT (to MAX) hashes]:salts' \
-  '(--costs'--costs='[load salts with(out) cost value Cn (to Mn)]:costs' \
+  '(--users)'--users=-'[do not load these users only]:user login or uid' \
+  '(--groups)'--groups=-'[do not load these group only]:group gid' \
+  '(--salts)'--salts=-'[load salts with(out) COUNT (to MAX) hashes]:salts' \
+  '(--costs)'--costs=-'[load salts with(out) cost value Cn (to Mn)]:costs' \
   '(--make-charset)'--make-charset=-'[make a charset file. It will be overwritten]:charset file:_files' \
   '(--pot)'--pot=-'[pot file to use]:pot file:_files' \
-  '(--node)'--node=-'[this node's number range out of TOTAL count]:number' \
+  '(--node)'--node=-'[this nodes number range out of TOTAL count]:number' \
```
solardiz: If we cannot have an apostrophe in there (but perhaps we can with some kind of escaping?), then maybe re-word to "number range of this node out of TOTAL count".

run/john.zsh_completion
```diff
@@ -48,19 +48,19 @@ _arguments -C -s \
...
   '(--subformat)'--subformat=-'[pick a benchmark format for --format=crypt]:format to use:->list-formats' \
   '(--mkpc)'--mkpc=-'[request a lower max. keys per crypt]:number' \
   '(--min-length)'--min-length=-'[request a minimum candidate length in bytes]:number' \
   '(--max-length)'--max-length=-'[request a maximum candidate length in bytes]:number' \
   '(--length)'--length=-'[shortcut for --min-len=N --max-len=N]:number' \
-  '(--field-separator-char)'--field-separator-chat=-'[use "C" instead of the "":" in input and pot files]:char' \
+  '(--field-separator-char)'--field-separator-chat=-'[use "C" instead in input and pot files]:char' \
```
solardiz: There's a typo in the original not yet fixed here: "chat" in place of "char". And the edited description is now weird, perhaps simply replace the `"":"` with `colon` instead.

solardiz (commented):
I'm not familiar with zsh completions and I think these are out of date, but I've made a couple of comments.

run/john.zsh_completion
Cal1z: " ... ' ... " does the trick

run/john.zsh_completion
Cal1z: put ":" back in

solardiz: Thanks. Will you fix the "chat" typo as well, please?

And, can you please force-push all of this as just one commit, instead of adding commits? Alternatively, please let us know if we should use GitHub's "squash and merge" - dirtier, but works. It is pointless to keep this PR's individual commits in our bleeding-jumbo branch's history as they're only part of this PR's short-lived history.

run/john.zsh_completion
magnumripper: Apparently the `'\''` (sort of) escape works too, just like in bash
```
$ zsh
magnum@large ~ % echo 'node'\''s'
node's
```

run/john.zsh_completion
Cal1z: done

solardiz:
@Cal1z Merged, thanks! I hope we'll see more contributions from you.

PR #4507, opened 2020-12-05
cracker.c: Introduce a fast path in process_key()

### LINKED ISSUE #4505: Fix salt resume comments in recovery.c to be consistent with cracker.c
I was concerned about possible GWS changes vs. salt resume, but it turns out we had already discussed this and Jim had already fixed it in bf86655abaacb54f5654efb07dba403cd8147ffa (later cleaned up by magnum in c1d7a2f31fae4189ad172b0c2d9861f47b591bc1). Reviewing this, I notice Jim's original commit had inconsistent comments in `recovery.c` (claiming we "just start from salt zero" if `max_keys_per_crypt` changed) vs. `cracker.c` (if necessary, reduces effective  `max_keys_per_crypt` for the first batch of keys to what was used before). We should correct the comments in `recovery.c`.

solardiz:
Jumbo got some overhead checks introduced into `process_key()` over the years. This introduces a fast path that bypasses those checks most of the time, reaching them only once per `max_keys_per_crypt`.

In my testing, this did provide a speedup on actual mask mode runs against fast hashes such as NT and LM on CPU, but somehow caused a maybe-slight-slowdown for `--test` of the same formats. To figure this out, I tried introducing debug prints to the fast and slow path. I confirmed that they were reached as expected during actual cracking, but somehow were not reached at all during `--test`. Don't we effectively use mask mode from our benchmarks by default now? I even tried passing `--test --mask`, which didn't make a difference. I am puzzled, and would appreciate comments by @magnumripper. Anyway, I think these changes are good, and should be merged after some review, and we should consider benchmarks separately.

While at it, I also fix #4505 with a separate commit.

### DIFF
--- a/src/cracker.c
+++ b/src/cracker.c
@@ -73,6 +73,7 @@
 #endif
 
 static fix_state_fp fp_fix_state;
+static int crk_process_key_max_keys;
 static struct db_main *crk_db;
 static struct fmt_params *crk_params;
 static struct fmt_methods crk_methods;
@@ -203,6 +204,7 @@ void crk_init(struct db_main *db, void (*fix_state)(void),
 #endif
 
 	if (db->loaded) crk_init_salt();
+	crk_process_key_max_keys = 0; /* use slow path at first */
 	crk_last_key = crk_key_index = 0;
 	crk_last_salt = NULL;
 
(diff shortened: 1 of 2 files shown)


magnumripper (changes requested):
Once in fast path, it looks to me we never call `clear_keys()` again?

magnumripper:
No wait, `crk_salt_loop` resets `crk_process_key_max_keys` so slow path will always run index 0, right?

magnumripper (approved):

solardiz:
> `crk_salt_loop` resets `crk_process_key_max_keys` so slow path will always run index 0, right?

Right. In an earlier revision, I had this reset right in `process_key()`, which probably was sufficient as well, but thought it's safer to do it in `crk_salt_loop()`, which is also called from a few more places.

PR #4508, opened 2020-12-06
New rule-reject flags for stacked (or not) rules

magnumripper:
-R      reject this rule unless stacked rules will be applied later
-S      reject this rule if stacked rules will be applied later

```
cat <<EOL > test.conf
.include <john.conf>
[List.Rules:teststack]
# This rule only applied if stacked rules come later
-R Az"_addR"
# This rule *not* applied if stacked rules come later
-S Az"_addS"
EOL
$ echo magnum > words

$ ../run/john -conf:test.conf -stdout -w:words -rules:teststack
magnum_addS

$ ../run/john -conf:test.conf -stdout -w:words -rules-stack:teststack
magnum_addS

$ ../run/john -conf:test.conf -stdout -w:words -rules:teststack -rules-stack:teststack
magnum_addR_addS

$ ../run/john -conf:test.conf -stdout -mask:magnum -rules-stack:teststack
magnum_addS

$ ../run/john -conf:test.conf -stdout --prince:words -max-len=16 -rules:teststack
magnum_addS

$ ../run/john -conf:test.conf -stdout --prince:words -max-len=16 -rules-stack:teststack
magnum_addS

$ ../run/john -conf:test.conf -stdout --prince:words -max-len=16 -rules:teststack -rules-stack:teststack
magnum_addR_addS
```

### DIFF
--- a/doc/RULES
+++ b/doc/RULES
@@ -28,6 +28,8 @@ natively.  Those additional commands may also be useful on their own.
 -8	reject this rule unless current hash type uses 8-bit characters
 -s	reject this rule unless some password hashes were split at loading
 -p	reject this rule unless word pair commands are currently allowed
+-R	reject this rule unless stacked rules will be applied later
+-S	reject this rule if stacked rules will be applied later
 -u	reject this rule if an internal codepage is configured (doc/ENCODINGS)
 -U	reject this rule unless an internal codepage is configured
 ->N	reject this rule unless length N or longer is supported
--- a/src/rules.c
+++ b/src/rules.c
@@ -724,6 +724,13 @@ char *rules_reject(char *rule, int split, char *last, struct db_main *db)
 			if (options.internal_cp != UTF_8 && options.internal_cp != ENC_RAW) continue;
 			return NULL;
 
+		case 'R':
+			if (rules_stacked_after) continue;
+			return NULL;
+
+		case 'S':
+			if (!rules_stacked_after) continue;
+			return NULL;
 /*
  * Any failed UTF-8-to-codepage translations change '-U' to '--' in
  * rpp_process_rule(), as an "always reject" hack.


magnumripper:
These should not be over-used, but are needed for some precision "rules x rules" situations. Just as an example, we have a rule that create "cracking" from the baseword "crack". If running that ruleset on both "sides", as in `--rules=wordlist --rules-stack=wordlist`, one might want to prepend that rule with one of these new flags to avoid candidates like "crackinging"

doc/RULES
```diff
@@ -28,11 +28,19 @@ natively.  Those additional commands may also be useful on their own.
...
+-R	reject this rule unless stacked rules will also be applied later
+-S	reject this rule if stacked rules will also be applied later
 -u	reject this rule if an internal codepage is configured (doc/ENCODINGS)
 -U	reject this rule unless an internal codepage is configured
 ->N	reject this rule unless length N or longer is supported
 -<N	reject this rule unless length N or shorter is supported (--min-length)
 
+Note that -R will only accept a rule when both --rules (or --single) and
+--rules-stack are used at the same time (rules stacked on rules) and even then
+it will only accept lines from --rules, not from --stacked-rules.  If you run
```
solardiz: What's `--stacked-rules`? If you used this "colloquially", I think you shouldn't - when we list something as if it were a command-line option, it should actually be one.

doc/RULES
```diff
@@ -28,11 +28,19 @@ natively.  Those additional commands may also be useful on their own.
...
+only --rules OR --rules-stack, -S will accept and -R will reject.  So if/when
+deciding which of those two flags to use, -S is normally preferable.
```
solardiz: The options do the opposite things and are presumably to be used in different cases, so I find it weird to say that one is "preferable" over the other.

solardiz (changes requested):
I didn't review the implementation's correctness - that would take testing, which I assume you did.

I don't really mind having this, although I also don't see much need for it. Please provide a valid use case for it before merging, or else please revise to match a use case.

I think the example with adding "ing" twice doesn't apply: this would generally not be the only rule, and we wouldn't want that rule to be skipped when used along with other rules stacked before or after it. Maybe we need something like `-D` that wouldn't apply the rule along with itself. So we could prefix `<* l [PI]` with `-D`.

We could also prefix `:` in some non-default ruleset with `-R -D` to implement what you wanted before, although I'm not sure this is sufficiently valuable and not confusing.

With this, I don't have a valid use case for `-S` at all. So maybe instead of my suggested `-D` above, we should redefine the meaning of `-R` and `-S` to be "reject this rule except when stacked along with a different rule" and "reject this rule when stacked along with itself", respectively. Then on `[PI]` we only need `-S` and on `:` we only possibly need `-R`. I think I like this better.

The "different" vs. "itself" comparisons should probably be on normalized rule strings, so the same rule coming from a different stacked ruleset would be considered the same.

doc/RULES
magnumripper: That was just a typo, will fix.

BTW I was never happy with the name "--rules-stack" but couldn't come up with anything better and there are pros with calling it "--rules-something". I'd rather have a syntax to use with `--rules=` for stacking "any" number of rulesets.

doc/RULES
magnumripper: > I don't really mind having this, although I also don't see much need for it. Please provide a valid use case for it before merging, or else please revise to match a use case.

The IRL case that made me do this was actually just a bunch of similar appends (edit: similar to Crack -> Cracking) although more of them and longer, and the unwanted/useless words quickly grew to a large number.

> With this, I don't have a valid use case for `-S` at all. So maybe instead of my suggested `-D` above, we should redefine the meaning of `-R` and `-S` to be "reject this rule except when stacked along with a different rule" and "reject this rule when stacked along with itself", respectively. Then on `[PI]` we only need `-S` and on `:` we only possibly need `-R`. I think I like this better.

I like your ideas better too. I'll experiment with that!

magnumripper:
> So maybe (...) we should redefine the meaning of `-R` and `-S` to be "reject this rule except when stacked along with a different rule" and "reject this rule when stacked along with itself", respectively. Then on `[PI]` we only need `-S` and on `:` we only possibly need `-R`. I think I like this better.

So -R would reject when not stacked at all (but -S would not), right? Otherwise they end up being the same thing, unless I'm confused now.

Regardless, it wont be trivial - especially with single mode where we may have a good number of rules in use at once. I'm currently leaning towards liking my original flags better - although my original -R might be overkill, right now I only know use for the -S.

solardiz:
> So -R would reject when not stacked at all (but -S would not), right?

This seems right.

> Regardless, it wont be trivial - especially with single mode where we may have a good number of rules in use at once.

Could be non-trivial/costly for single mode, yes. Maybe have them work in other modes only (not reject anything in single mode)? This becomes even more confusing, though.

> I'm currently leaning towards liking my original flags better - although my original -R might be overkill, right now I only know use for the -S.

What is that use case? I currently don't see any use for your originally proposed flags.

magnumripper:
> What is that use case? I currently don't see any use for your originally proposed flags.

The Crack -> Cracking -> Crackinging example basically explains my problem, although I have longer and more elaborate stuff and many more of them. Come to think of it, since there are many of them I really do need my original `-S` - a "reject if dupe" will not cut it!

Also, the `-R` can be used to suppress the stacked NOP when it's not wanted. I currently have lots of tedious cludges like:
```
[List.Rules:SomeRuleSet]
fo o
b ar

[List.Rules:SomeRuleSet-with-nop]
:
.include [List.Rules:SomeRuleSet]
```
Sure, any weird result such as "Crackinging" *could* crack a password, but it's not what I want to produce and once I do, I'd simply not use that flag. Today's super slow formats need surgical rules.

magnumripper:
I still like the idea of "reject stacked dupe rule" (as well) though, perhaps for the `-R` flag only. I will give it another look.

magnumripper:
> I still like the idea of "reject stacked dupe rule" (as well) though, perhaps for the `-R` flag only. I will give it another look.

Bad news: Stacked rules pre-run `rules_reject()` at init, and some modes such as PRINCE do so with normal rules as well. So this can't be implemented (without major rework).

magnumripper:
I've exhausted all ideas - we can't have a "reject this rule when stacked along with itself" unless we do it in `rules_apply` and that would be ridiculous.

Unless you come up with a better idea, please merge.

solardiz:
I still don't know how you'd use these flags without unnecessarily rejecting meaningful combinations of rules, but I don't want to block your work, so I'll merge.

PR #4509, opened 2020-12-07
Bugfixes for --rules-stack

magnumripper:
While experimenting with "stacked rules reject flags" (#4508) I noticed the rules_stacked_after variable wasn't always "in sync" due to how/when various modes reject/accept rules, and other things.

I'm not sure the bug actually manifested itself much but among other things things there could have been some problems with ignoring max. lengths when we know more rules will be applied later, things like that.

### DIFF
--- a/src/rules.c
+++ b/src/rules.c
@@ -563,6 +563,8 @@ int rules_init_stack(char *ruleset, rule_stack *stack_ctx,
 		rules_init(db, options.eff_maxlength + mask_add_len);
 		rule_count = rules_count(&ctx, -1);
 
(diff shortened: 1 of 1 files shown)


solardiz (approved):
My only review is the changes here are limited to setting `rules_stacked_after` differently. I don't know if they're actually correct or not. It'd be nice to have test case(s) confirming that what worked before still does, and that something that didn't work correctly before now does. But since magnum says this fixes bugs, I don't mind merging this anyway.

magnumripper:
I systematically tested all situations I could imagine, with debug prints all over the place

PR #4484, opened 2020-11-22
Spare Travis CI resources

ghost:

PR #4510, opened 2020-12-08
Travis CI: conditionally skip some tests

solardiz:
Based on Claudio's changes proposed in #4484

Claudio, why did you propose to exclude "Extra Testing Scheme", which was testing with Intel OpenCL in Docker, and which we only ran on already merged PRs (IIUC)? Is this only to spare resources (roughly how much, e.g. what portion of our total usage of Travis?), or were you planning to retire that Docker image for other reasons (why?), or something else? I do include this change of yours in here right now, but we might want to exclude it, depending on your answers.

I think ideally we should be able to manually trigger Travis runs for PRs when we need them - perhaps when we're deciding on merging a PR that makes code changes (not just documentation or such). With Circle, there's a way to have tests run only once a pull request is approved - this isn't exactly what I just described, but it's close. Is there a similar feature with Travis?

No luck getting an "OSS allotment" from them yet, although they did reply. My impression (they didn't say so explicitly) is they're probably struggling financially. So Travis won't run yet, but I am considering paying them for some credits (will need to figure this out first), at which point these changes would matter.

BTW, testing on macOS costs 5x more credits than testing on Linux.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 language: c
 dist: bionic
+os: linux
 
-matrix:
+jobs:
   include:
 
   # #### CPU formats test (gcc)
@@ -56,16 +57,6 @@ matrix:
   - os: osx
     osx_image: xcode9.4
     env: OPENCL="yes"
-
-  # #########################################
-  # #### Extra Testing Scheme
-  # #########################################
-
-  # #### Intel OpenCL runtime Docker based testing
-  - if: branch = bleeding-jumbo AND env(TRAVIS_PULL_REQUEST) = false
-    os: linux
-    compiler: gcc
-    env: DOCKER="yes" OPENCL="yes"
 
 script:
   - .travis/travis-ci.sh
--- a/.travis/travis-ci.sh
+++ b/.travis/travis-ci.sh
@@ -1,5 +1,23 @@
 #!/bin/bash
 
+# if requested, run all tests
+if [[ "$TRAVIS_PULL_REQUEST_BRANCH" != "force-travis" ]]; then
+
+    # on the main branch, run all tests
+    if [[ "$TRAVIS_BRANCH" != "bleeding-jumbo" || "$TRAVIS_PULL_REQUEST" =~ ^[0-9]+$ ]]; then
+
+#        # otherwise, run only Linux jobs for now
+#        if [[ "$TRAVIS_OS_NAME" != "linux" ]]; then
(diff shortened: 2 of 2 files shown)


solardiz:
> Travis won't run yet, but I am considering paying them for some credits (will need to figure this out first)

Just tried to figure it out. Looks like currently credits can only be purchased as an add-on on a plan with a monthly fee. Sent them a support request asking if they can somehow be added without also having a monthly fee.

magnumripper:
On a related note I've had the idea that since we already run simple bash scripts in the CI bots, we could actually decide whether or not to build at all (simply report success if we don't) and if we do - what to build (eg. opencl or not). We can decide whether to test everything or just a few things. This could be extremely advanced if we wanted to, such as only testing altered formats or even engage the test-suit for eg. wordlist.c changes, but could start very simple yet effective.

Pseudo-code example
```
build = test = no
files = $(git diff --stat bleeding-jumbo | grep " " | cut -d\  -f2)
if (files include "*.[ch]")
    build = cpu
    if (files include "*opencl*.[ch]")
        build = opencl
    fi
    if (files include "opencl*_fmt*")
        test = opencl
    elif (files include "*_fmt*")
        test = cpu
    fi
fi
(...)
```
Just the above would go from sometimes not even building (just report success - eg. a doc change), sometimes building CPU or OpenCL, but not running `--test` (obviously report success unless there were build problems), and sometimes testing like we currently do.

solardiz:
@claudioandre-br "Claudio, why did you propose to exclude "Extra Testing Scheme", which was testing with Intel OpenCL in Docker"?

I'd like to include your proposed changes in our tree, even if we're not currently using Travis, just to have them recorded in there. But I'd like your answer to the above first, as this involves deleting lines (of course, they'll stay in git history anyway).

ghost:
Em qui., 10 de dez. de 2020 às 11:09, Solar Designer <
notifications@github.com> escreveu:

> @claudioandre-br <https://github.com/claudioandre-br> "Claudio, why did
> you propose to exclude "Extra Testing Scheme", which was testing with Intel
> OpenCL in Docker"?
>

Because it takes 40 minutes to finish. It is too much!

solardiz:
> > "Claudio, why did you propose to exclude "Extra Testing Scheme", which was testing with Intel OpenCL in Docker"?

> Because it takes 40 minutes to finish. It is too much!

OK, I'll merge this for now, and we'll possibly resume some use of Travis a bit later. Thanks!

PR #4512, opened 2020-12-08
More bug fixes for stacked rules

magnumripper:
It hit me #4509 wasn't complete - I forgot about single mode, let alone batch mode.

Single mode and batch mode sometimes wouldn't have the rules_stacked_after variable "in sync".

Also adds a hack in batch.c making many existing Jumbo tests for single mode more reliable. For batch mode, changing the `cracker.c` test from `if (options.flags & FLG_SINGLE_CHK)` to `if (options.flags & (FLG_SINGLE_CHK | FLG_BATCH_CHK)` wouldn't cut it - it really needed to reflect if we're in single mode **right now**.

### DIFF
--- a/src/batch.c
+++ b/src/batch.c
@@ -16,14 +16,17 @@
 #include "loader.h"
 #include "status.h"
 #include "config.h"
+#include "options.h"
 #include "single.h"
 #include "wordlist.h"
 #include "inc.h"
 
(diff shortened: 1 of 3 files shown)


solardiz (approved):
No idea if this is correct and complete, but it looks reasonable. Should I merge?

magnumripper:
Yes, please do

Issue #4511, opened 2020-12-08
--mkpc doesn't affect log message


solardiz:
While testing my changes in #4507, I discovered that while setting a low `--mkpc` reduces speed as expected, the log file message doesn't change. I think that's a bug. Indeed, the code logging that message does not consider this setting:

```c
static void john_log_format2(void)
{
        int min_chunk, chunk;

        /* Messages require extra info not available in john_log_format().
                These are delayed until mask_init(), fmt_reset() */
        chunk = min_chunk = database.format->params.max_keys_per_crypt;
        if (options.flags & (FLG_SINGLE_CHK | FLG_BATCH_CHK) &&
            chunk < SINGLE_HASH_MIN)
                        chunk = SINGLE_HASH_MIN;
        if (chunk > 1)
                log_event("- Candidate passwords %s be buffered and "
                        "tried in chunks of %d",
                        min_chunk > 1 ? "will" : "may",
                        chunk);
}
```

(BTW, we should also add braces around the bitwise AND expression.)

Looking for uses of `options.force_maxkeys`, I notice that we dare to patch `min_keys_per_crypt` to be no higher than that setting if necessary. This leaves me wondering why we don't simply patch `max_keys_per_crypt` in the same place and in the same way. That would be so much simpler and more efficient than having to check `options.force_maxkeys` separately in many other places. Maybe that's the best fix for this issue as well. Would a change of  `max_keys_per_crypt` expose peculiarities of formats and maybe autotune that a change to `min_keys_per_crypt` does not?

magnumripper:
I guess you are right, and that the current code is a result of fixes applied to bodges.

solardiz:
> Would a change of `max_keys_per_crypt` expose peculiarities of formats

Actually, it did - see #1167 and 0a87a0a818b4e1239e8306fa7e025bc1b2fc2076.

PR #4516, opened 2020-12-10
Fixes and tweaks

magnumripper:
Two commits:

Have the "buffered and tried" log message reflect --mkpc option (#4511)

When `--mask` is given alone, a default mask of `?1?2?2?2?2?2?2?3?3?3?3?d?d?d?d` is used (same as hashcat's). That is barely of any use ever unless we iterate over length, so let's default to that. If you'd ever want the old behavior, just use `--len=15`.

### DIFF
--- a/src/mask.c
+++ b/src/mask.c
@@ -2130,6 +2130,12 @@ void mask_init(struct db_main *db, char *unprocessed_mask)
 	    !strcasecmp(mask_fmt->params.label, "lm-opencl"))
 		format_cannot_reset = 1;
 
+	/* Using "--mask" alone will use default mask and iterate over length */
+	if (!(options.flags & FLG_MASK_STACKED) && !unprocessed_mask &&
+	    options.req_minlength < 0 && !options.req_maxlength)
+		mask_increments_len = 1;
+
+	/* Specified length range given */
 	if ((options.req_minlength >= 0 || options.req_maxlength) &&
 	    (options.eff_minlength != options.eff_maxlength) &&
 	    !(options.flags & FLG_MASK_STACKED))
--- a/src/john.c
+++ b/src/john.c
@@ -334,12 +334,12 @@ static void john_log_format2(void)
 	/* Messages require extra info not available in john_log_format().
 		These are delayed until mask_init(), fmt_reset() */
 	chunk = min_chunk = database.format->params.max_keys_per_crypt;
-	if (options.flags & (FLG_SINGLE_CHK | FLG_BATCH_CHK) &&
-	    chunk < SINGLE_HASH_MIN)
-			chunk = SINGLE_HASH_MIN;
+	if (options.force_maxkeys && options.force_maxkeys < chunk)
+		chunk = min_chunk = options.force_maxkeys;
+	if ((options.flags & (FLG_SINGLE_CHK | FLG_BATCH_CHK)) && chunk < SINGLE_HASH_MIN)
+		chunk = SINGLE_HASH_MIN;
 	if (chunk > 1)
-		log_event("- Candidate passwords %s be buffered and "
-			"tried in chunks of %d",
+		log_event("- Candidate passwords %s be buffered and tried in chunks of %d",
 			min_chunk > 1 ? "will" : "may",
 			chunk);
 }


solardiz:
We definitely want the "Have the "buffered and tried" log message reflect --mkpc option" commit.

As to the default mask change, won't it break restoring of sessions previously started with the default mask at its fixed length? And if we find such breakage of restore and deviation from hashcat's default acceptable (because that mask was of little use anyway?), then perhaps we should make more changes at once - not just have the lengths increase gradually, but also use a less specialized mask. Otherwise if we change the default mask with a separate commit later, we'll be breaking restore again.

magnumripper:
I seriously doubt anyone having such session running but if they do, they'll get `Incorrect crash recovery file: john.rec` error, giving them a chance to recover.

As to the default mask used, I think it's a decent one and I believe the hashcat folks spent some time testing before ending up with it. I'm open for suggestions though.

solardiz:
Why would the `Incorrect crash recovery file: john.rec` error be produced?

As to the default mask, I realized we only have it in `john.conf`, not in code, so it's less relevant to our decision-making on this PR than I had thought it would be.

I'm fine with merging this now.

magnumripper:
> Why would the `Incorrect crash recovery file: john.rec` error be produced?

I'm actually not quite sure why, but it does. I tried it only after you asked about it. Apparently mask mode saves things a bit differently when increasing length and no new code was needed to make it catch this.

magnumripper:
Probably as simple as this (`mask_restore_state()`)
```c
	if (mask_increments_len) {
		if (fscanf(file, "%d\n", &d) == 1)
			restored_len = d;
		else
			return fail;
		if (fscanf(file, "%"PRIu64"\n", &ull) == 1)
			rec_cl = ull;
		else
			return fail;
	}
```

src/mask.c
```diff
@@ -2130,6 +2130,12 @@ void mask_init(struct db_main *db, char *unprocessed_mask)
 	    !strcasecmp(mask_fmt->params.label, "lm-opencl"))
 		format_cannot_reset = 1;
 
+	/* Using "--mask" alone will use default mask and iterate over length */
+	if (!(options.flags & FLG_MASK_STACKED) && !unprocessed_mask &&
+	    options.req_minlength < 0 && !options.req_maxlength)
+		mask_increments_len = 1;
```
solardiz, line 2136: Does this work correctly without a specific lengths range?

solardiz:
> Probably as simple as this (`mask_restore_state()`)

Yes, I saw this, but was confused by us not restoring `mask_increments_len` first. I guess we simply set it again in the same way we did on the initial options parsing. OK.

src/mask.c
magnumripper: It defaults to format's min. and max. lengths (and I verified it :-)

magnumripper:
Please merge. BTW can't we shut off the requirement for Travis?

solardiz:
> BTW can't we shut off the requirement for Travis?

We might. As I understand, we'll need to remove the Travis file from the tree for that.

I am still hoping to negotiate something reasonable with them and use the resources sparingly. I had also setup Travis for the LKRG repo here recently, and it's a pity to lose that - especially considering how little resources the LKRG builds were consuming.

magnumripper:
> As I understand, we'll need to remove the Travis file from the tree for that.

I though we could simply say it's not mandatory. Maybe that's just CircleCI?

solardiz:
> > As I understand, we'll need to remove the Travis file from the tree for that.
> 
> I though we could simply say it's not mandatory.

We should be able to, but I lost where. Anyway, I've just limited the org's integration with Travis to the LKRG repo. Maybe this will do the trick, and be easy to revert.

magnumripper:
I don't have them anymore in my repo but under settings, branch protection rules, "Require status checks to pass before merging" I think you can just tick it off

solardiz:
> under settings, branch protection rules, "Require status checks to pass before merging" I think you can just tick it off

Ah, right, did that now - and Travis completely disappeared from there as soon as I unticked it. I guess that's because I had disabled its integration for this repo. Probably both actions are easy to revert.

ghost:
JtR runs some stuff in Travis we can't run elsewhere easily (Mac build,
Ubuntu 14 + OpenCL). Regular stuff can be run in CircleCI (for a while??
🤔) using docker. They have, at least, one base VM (Travis has more).

>

PR #4520, opened 2020-12-12
Fix a regression with recent mask changes.

magnumripper:
We can't set mask_increments_len when running --test.  For some reason I saw no problems with it a couple of days ago, and apparently neither did the CI's, but I now got segfaults just from --test.

### DIFF
--- a/src/mask.c
+++ b/src/mask.c
@@ -2131,7 +2131,7 @@ void mask_init(struct db_main *db, char *unprocessed_mask)
 		format_cannot_reset = 1;
 
 	/* Using "--mask" alone will use default mask and iterate over length */
-	if (!(options.flags & FLG_MASK_STACKED) && !unprocessed_mask &&
+	if (!(options.flags & FLG_MASK_STACKED) && (options.flags & FLG_CRACKING_CHK) && !unprocessed_mask &&
 	    options.req_minlength < 0 && !options.req_maxlength)
 		mask_increments_len = 1;
 


solardiz:
What about `--test-full` and `--fuzz`? Maybe we should require `FLG_CRACKING_CHK` instead of excluding known-conflicting flags?

solardiz:
> What about `--test-full` and `--fuzz`?

Oh, these look OK per the flags they set/require. And the code patched here is only reached when mask was allowed along with whatever we're running. So I'm OK with the changes as they are, although checking `FLG_CRACKING_CHK` instead would also be OK.

solardiz (approved):

magnumripper:
> > What about `--test-full` and `--fuzz`?
> 
> Oh, these look OK per the flags they set/require. And the code patched here is only reached when mask was allowed along with whatever we're running. So I'm OK with the changes as they are, although checking `FLG_CRACKING_CHK` instead would also be OK.

I'm not sure about fuzz though? What does it set or require? I think I'll change to `FLG_CRACKING_CHK` anyway.

src/mask.c
```diff
@@ -2131,7 +2131,7 @@ void mask_init(struct db_main *db, char *unprocessed_mask)
 		format_cannot_reset = 1;
 
 	/* Using "--mask" alone will use default mask and iterate over length */
-	if (!(options.flags & FLG_MASK_STACKED) && !unprocessed_mask &&
+	if (!(options.flags & FLG_MASK_STACKED) && (options.flags & FLG_CRACKING_CHK) && !unprocessed_mask &&
```
solardiz, line 2134: BTW, to check that one flag is set and the other isn't we can use `(flags & (flag1 | flag2)) == flag1`.

Issue #4519, opened 2020-12-10
Replace --fix-state-delay option with cracker logic


magnumripper:
I'd like to revisit this. The tests made in #1955 were bogus in several ways (I can elaborate on how, if it's not obvious. There are loads of foot-guns in such benchmarks). I simply can't fathom how the overhead given by the `--fix-state-delay` option would ever in any case be amortized by the alleged win.

`ftell()`, no matter how stupidly implemented, can't be anything other than dereferencing a pointer (at least with non-medieval storage media: It's not I/O for sure - how the heck would your storage know where you were at and where would it store that?!). Right? So our code for avoiding that call to `ftell` **adds** a pointer, increments of it and tests of it - all that for avoiding simply dereferencing the first pointer. I can't believe that can ever render in a real-life boost. If I'm wrong, please explain it to me.

Also, I'm sure (for many reasons apart from just gut feeling) no-one actually uses the option. That means the overhead added by it (be it tiny) is always hurting.

The least we should do is move the code away from all normal code paths (nowadays we'll almost always using memory map and/or memory buffer/indexing), and while at it we should forbid the option to be used without `-wordlist` because it's otherwise rightously ignored. I forgot doing so in my recent review of mutually exclusive options.

In case I'm all wrong and skipping `fix_state()` is actually sensible at some time, I propose we do so in `cracker`, and skip it if our last call to it was eg. no more than `timer_save_interval / 10` ago. (or something such - maybe even `/ 2`).

solardiz:
> The tests made in #1955 were bogus in several ways (I can elaborate on how, if it's not obvious

It is unclear from what @jfoug posted there whether the wordlist was possibly being read from disk the first time but got cached the second time, but I'd expect Jim to know such things and have run the first benchmark a few times before posting.

> `ftell()`, no matter how stupidly implemented, can't be anything other than dereferencing a pointer

Hopefully so, but even then unless it's a macro it's call into a relocatable shared library, which involves some overhead.

I think it does make sense for us to minimize the frequency of `ftell()` calls if we can do so easily and with little overhead, and to do so unconditionally (at least in builds that supposedly benefit from it), not require an obscure command-line options that no one except for @jfoug would know/recall exists.

solardiz:
> `ftell()`, no matter how stupidly implemented, can't be anything other than dereferencing a pointer

[...]

PR #4524, opened 2020-12-12
timer: Add sTimer_GetNano, and drop obsolete fallback, for lower overhead

### LINKED ISSUE #3991: Improve (and maybe simplify) our "portable hi-res timer"

magnumripper:
Also drops obsolete fallbacks to clock(3), simplifying code. They would only possibly be needed for Windows **pre** XP.  In the unlikely case anyone actually would need that, we could of course put it back with #ifdef's.

See #4519, although this commit is only about the "portable HR-timer" which *might* be used for that.

Closes #3991

### DIFF
--- a/src/listconf.c
+++ b/src/listconf.c
@@ -142,10 +142,6 @@ static void listconf_list_build_info(void)
 #ifdef __GNU_MP_VERSION
 	int gmp_major, gmp_minor, gmp_patchlevel;
 #endif
-	sTimer test;
-
-	sTimer_Init(&test);
-
 	puts("Version: " JTR_GIT_VERSION);
 	puts("Build: " JOHN_BLD _MP_VERSION DEBUG_STRING ASAN_STRING UBSAN_STRING);
 #ifdef SIMD_COEF_32
@@ -323,13 +319,15 @@ static void listconf_list_build_info(void)
 	printf("times(2) CLK_TCK is %ld\n", clk_tck);
 #endif
 #if defined (__MINGW32__) || defined (_MSC_VER)
-	printf("Using clock(3) for timers, claimed resolution %ss, observed %ss\n",
-	       human_prefix_small(1.0 / CLOCKS_PER_SEC),
-	       human_prefix_small(1.0 / sm_cPrecision));
+	printf("Using clock(3) for timers, resolution %ss\n", human_prefix_small(1.0 / CLOCKS_PER_SEC));
 #else
 	printf("Using times(2) for timers, resolution %ss\n", human_prefix_small(1.0 / clk_tck));
 #endif
-	printf("HR timer claimed resolution %ss, observed %ss\n", human_prefix_small(1.0 / sm_HRTicksPerSec), human_prefix_small(1.0 / sm_hrPrecision));
+	int latency;
+	uint64_t precision = john_timer_stats(&latency);
+
+	printf("HR timer: %s, %s %ss\n", john_nano_clock, latency ? "latency" : "resolution",
+	       human_prefix_small(precision / 1000000000.0));
 
 	int64_t total_mem = host_total_mem();
 
--- a/src/omp_autotune.c
+++ b/src/omp_autotune.c
@@ -19,8 +19,8 @@
 #define CONF_SECTION SECTION_OPTIONS, ":CPUtune"
 
 static int use_preset, max_no_progress;
-static double sample_time, req_gain, max_tune_time;
-
+static double req_gain, max_tune_time;
+static uint64_t sample_time;
 static struct fmt_main *fmt;
 static int omp_autotune_running;
 static int mkpc;
@@ -42,7 +42,7 @@ void omp_autotune_init(void)
 #endif
 	if ((ci = cfg_get_int(CONF_SECTION, "AutoTuneSampleTime")) < 0)
 		ci = 10;
-	sample_time = (double)ci / 1000.0;
+	sample_time = ci * 1000000ULL;
 	if ((ci = cfg_get_int(CONF_SECTION, "AutoTuneReqGain")) < 0)
 		ci = 5;
 	req_gain = (double)ci / 100.0 + 1.0;
@@ -108,7 +108,7 @@ void omp_autotune_run(struct db_main *db)
 	int tune_cost;
 	void *salt;
 	char key[PLAINTEXT_BUFFER_SIZE] = "tUne0000";
-	sTimer timer;
+	uint64_t start, end;
 	double duration;
 
 	if (!fmt || omp_scale == 1 || tune_preset)
@@ -143,7 +143,6 @@ void omp_autotune_run(struct db_main *db)
 		printf("\n");
 	}
 
-	sTimer_Init(&timer);
 	do {
 		int i;
 		int min_kpc = fmt->params.min_keys_per_crypt;
@@ -180,18 +179,18 @@ void omp_autotune_run(struct db_main *db)
 		// Tell format this is a speed test
 		benchmark_running++;
 
-		sTimer_Start(&timer, 1);
+		start = john_get_nano();
 		do {
 			int count = this_kpc;
 
 			fmt->methods.crypt_all(&count, NULL);
 			crypts += count;
-		} while (crypts < min_crypts || sTimer_GetSecs(&timer) < sample_time);
-		sTimer_Stop(&timer);
+			end = john_get_nano();
+		} while (crypts < min_crypts || (end - start) < sample_time);
 
 		benchmark_running--;
 
-		duration = sTimer_GetSecs(&timer);
+		duration = (end - start) / 1E9;
 		cps = crypts / duration;
 
 		if (john_main_process && options.verbosity >= VERB_MAX) {
--- a/src/timer.c
+++ b/src/timer.c
@@ -1,155 +1,119 @@
 /*
- * This file is
- * Copyright (c) 2009 by Jim Fougeron jfoug AT cox dot net
- * Copyright (c) 2019 by magnum
+ * This file is Copyright (c) 2021 by magnum
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted.
- *
- * Portable hi-res timer.  Was a nice C++ class. Downgraded to C for john.
  */
 
-#include <stdio.h>
 #include <stdint.h>
 
 #include "timer.h"
+#include "misc.h"
 
-#if __MACH__
-static uint64_t sm_timebase;
-#endif
-static int sm_fGotHRTicksPerSec = 0;   // Set if we have got the above
+#if defined (_MSC_VER) || defined (__MINGW32__) || defined (__CYGWIN32__)
 
-uint64_t sm_HRTicksPerSec;  // HR ticks per second (claimed)
-uint64_t sm_hrPrecision;    // HR ticks per second (observed, guess)
-uint64_t sm_cPrecision;     // clocks (ticks) per second (observed, guess)
+#undef MEM_FREE
+#include <windows.h>
+#undef MEM_FREE
+const char *john_nano_clock = "QueryPerformanceCounter()";
 
-void sTimer_Init(sTimer *t)
+/*
+ * QueryPerformanceCounter(LARGE_INTEGER*) gives "counts" (ticks)
+ * QueryPerformanceFrequency(LARGE_INTEGER*) gives ticks/s
+ */
+uint64_t john_get_nano(void)
 {
-	t->m_fRunning = 0;
-	t->m_cStartTime = 0;
-	t->m_cEndTime = 0;
-	t->m_dAccumSeconds = 0;
-	HRZERO(t->m_hrStartTime);
-	HRZERO(t->m_hrEndTime);
-	if (!sm_fGotHRTicksPerSec) {
-		// What's the lowest digit set non-zero in a clock() call
-		// That's a fair indication what the precision is likely to be.
-		// Note - this isn't actually used
-		int i;
+	static uint64_t ticks_per_sec;
+	LARGE_INTEGER t;
 
-		sm_fGotHRTicksPerSec = 1;
-		sm_cPrecision = 0;
-		for (i = 0; i < 10; ++i) {
-			clock_t heuristicTimeTest = clock();
-
-			if (heuristicTimeTest % 10) {
-				sm_cPrecision = CLOCKS_PER_SEC;
-				break;
-			}
-			else if (heuristicTimeTest % 100) {
-				if (!sm_cPrecision || sm_cPrecision < CLOCKS_PER_SEC / 10)
-					sm_cPrecision = CLOCKS_PER_SEC / 10;
-			}
-			else if (heuristicTimeTest % 1000) {
-				if (!sm_cPrecision || sm_cPrecision < CLOCKS_PER_SEC / 100)
-					sm_cPrecision = CLOCKS_PER_SEC / 100;
-			}
-			else if (heuristicTimeTest % 10000) {
-				if (!sm_cPrecision || sm_cPrecision > CLOCKS_PER_SEC / 1000)
-					sm_cPrecision = CLOCKS_PER_SEC / 1000;
-			}
-			else {
-				if (!sm_cPrecision || sm_cPrecision > CLOCKS_PER_SEC / 10000)
-					sm_cPrecision = CLOCKS_PER_SEC / 10000;
-			}
-		}
-
-		// Find the claimed resolution of the high res timer
-		// Then find the most likely real rate by waiting for it to change.
-		// Note - I've frequently seen missed beats, and therefore a
-		// 0.000001 reality gets reported as a 0.000002.
-		// Note - this also isn't actually used, all that matters is
-		// whether HRTicksPerSec has a non-zero value or not.
-		HRGETTICKS_PER_SEC(sm_HRTicksPerSec);
-
-		if (sm_HRTicksPerSec) {
-			hr_timer start, end;
-
-			HRSETCURRENT(start);
-			do {
-				HRSETCURRENT(end);
-			} while (HRGETTICKS(end) == HRGETTICKS(start));
-
-			sm_hrPrecision = sm_HRTicksPerSec / (HRGETTICKS(end) -
-			                  HRGETTICKS(start));
-		}
-	}
-}
-
-
-void sTimer_Stop(sTimer *t)
-{
-	if (t->m_fRunning) {
-		if (sm_HRTicksPerSec)
-			HRSETCURRENT(t->m_hrEndTime);
-		else
-			t->m_cEndTime = clock();
-	} else
-		HRZERO(t->m_hrEndTime);
-
-	t->m_fRunning = 0;
-}
-
-void sTimer_Start(sTimer *t, int bClear)
-{
-	if (bClear)
-		sTimer_ClearTime(t);
-
-	if (sm_HRTicksPerSec)
-		HRSETCURRENT(t->m_hrStartTime);
-	else
-		t->m_cStartTime = clock();
-
-	t->m_fRunning = 1;
-}
-
-void sTimer_ClearTime(sTimer *t)
-{
-	t->m_dAccumSeconds = 0;
-	HRZERO(t->m_hrStartTime);
-	HRZERO(t->m_hrEndTime);
-
-	t->m_fRunning = 0;
-}
-
-double sTimer_GetSecs(sTimer *t)
-{
-	double retval;
-
-	if (t->m_fRunning) {
-		if (sm_HRTicksPerSec)
-			HRSETCURRENT(t->m_hrEndTime);
-		else
-			t->m_cEndTime = clock();
-	}
-	if (!sm_HRTicksPerSec) {
-		// This is process time
-		double d = (t->m_cEndTime - t->m_cStartTime);
-
-		if (d > 0)
-			retval = d / CLOCKS_PER_SEC;
-		else
-			retval = 0;
-	}
-	else {
-		// This is wall-clock time
-		double d = (HRGETTICKS(t->m_hrEndTime) - HRGETTICKS(t->m_hrStartTime));
-
-		if (d > 0)
-			retval = d / sm_HRTicksPerSec;
-		else
-			retval = 0;
+	if (!ticks_per_sec) {
+		QueryPerformanceFrequency(&t);
+		ticks_per_sec = t.QuadPart;
 	}
 
-	return retval + t->m_dAccumSeconds;
+	QueryPerformanceCounter(&t);
+
+	return t.QuadPart * 1000000000ULL / ticks_per_sec;
+}
+
+#elif __MACH__
+
+#include <mach/mach_time.h>
+const char *john_nano_clock = "mach_absolute_time()";
+
+/*
+ * mach_absolute_time() gives ticks
+ * mach_timebase_info() gives numer/denom to convert ticks to nanoseconds
+ */
+uint64_t john_get_nano(void)
+{
+	static mach_timebase_info_data_t timebase;
+
+	if (!timebase.denom)
+		if (mach_timebase_info(&timebase) != KERN_SUCCESS)
+			pexit("mach_timebase_info()");
+
+	return mach_absolute_time() * timebase.numer / timebase.denom;
+}
+
+#else /* Linux, POSIX */
+
+#include <unistd.h>
+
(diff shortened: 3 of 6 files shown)


src/listconf.c
```diff
@@ -323,9 +323,7 @@ static void listconf_list_build_info(void)
 	printf("times(2) CLK_TCK is %ld\n", clk_tck);
 #endif
 #if defined (__MINGW32__) || defined (_MSC_VER)
-	printf("Using clock(3) for timers, claimed resolution %ss, observed %ss\n",
-	       human_prefix_small(1.0 / CLOCKS_PER_SEC),
-	       human_prefix_small(1.0 / sm_cPrecision));
+	printf("Using clock(3) for timers, claimed resolution %ss\n", human_prefix_small(1.0 / CLOCKS_PER_SEC));
```
solardiz: You wrote "Also drops obsolete fallbacks to clock(3)", so I expected the code would no longer use `clock(3)` at all, but here you still have a message claim that it does?

magnumripper: No, that line refers to the original timers in `signals.c`, showing what timer is used for this build. The thing removed now was just some curious measure of its observed resolution, that happened to come with the clock fallback in the "HR timer".

magnumripper:
So perhaps before falling back to `gettimeofday()` we should prefer `clock_gettime()` if available?

magnumripper:
> So perhaps before falling back to `gettimeofday()` we should prefer `clock_gettime()` if available?

No wait, we already do 😆

magnumripper:
BTW for my current experiments with cracker, instead of `sTimer_getNano()` I use a macro not yet in this PR that just reads an `uint64_t` timestamp and don't bother updating any struct or using "start/stop" or anything like that.
```c
#define sTimer_GetRaw(h) ({ HRSETCURRENT((h)); HRGETTICKS((h));})
```
Edit: It's not quite like the above but it boils down to a call to some good proprietary Windows or macOS timer, or `clock_gettime()` or as a last fall-back `gettimeofday()`, and returning it as an uint64_t.

magnumripper:
For things like fix_state delay use we should definitely use `CLOCK_MONOTONIC_RAW_APPROX` if available.

solardiz:
Why do we ever need to fallback to `gettimeofday()`? I think we shouldn't ever use that function (except perhaps for ETA reporting) because it's affected by system time adjustments. Aren't these two sufficient - `clock_gettime()` when we have it, and `times()` otherwise?

> For things like fix_state delay use we should definitely use `CLOCK_MONOTONIC_RAW_APPROX` if available.

I think no system-provided timer is appropriate "for things like fix_state delay". Any system-provided timer is potentially too slow to query for those. Just how do you think a timer can possibly be quick to query? Normally it's a call into the kernel, which is thus not quick. It's only quick if the OS bothered to have the kernel maintain a suitable timer value in a page readable from userspace, like recent Linux does with vDSO, but even then there are [limitations](https://berthub.eu/articles/posts/on-linux-vdso-and-clockgettime/). While we could choose a timer that's quick to query on current Linux, this is very OS specific, and will end up being slow on some other OS. It's the `ftell()` kind of problem all over again.

solardiz:
> Why do we ever need to fallback to `gettimeofday()`?

OK, maybe our autotune needs that precision. Do we use this for autotune now?

In JtR core, my approach to precise benchmarks was to have a signal fire at exactly the requested number of seconds and measure how many `crypt_all()` were done by then. The code also uses `times()` to see how much real and virtual time actually passed, but due to the timer signal the precision is usually greater than that of `times()`. However, I can see how this might be unsuitable for OpenCL autotune, where each kernel invocation takes considerable time.

magnumripper:
> Why do we ever need to fallback to `gettimeofday()`? I think we shouldn't ever use that function (except perhaps for ETA reporting) because it's affected by system time adjustments. Aren't these two sufficient - `clock_gettime()` when we have it, and `times()` otherwise?

I never put it there, at the time the code was introduced it was first a Windows specific one, then `gettimeofday()` and as a last resort `clock()`. I have since added `clock_gettime()` and a macOS-specific alternative, and dropped `clock()`. I think `times()` is too low resolution to be useful here. BTW I think some systems (at least theoretically) could have only non-monotonic `clock_gettime()` although we do try to get a monotonic one.

> OK, maybe our autotune needs that precision. Do we use this for autotune now?

For CPU auto-tune yes, this is what we currently use. For OpenCL we use the profiling stuff provided by OpenCL except a number of Sayantan's formats (and even some of my formats that was based on his) unconditionally use `gettimeofday()` for tuning. The ZTEX formats also use `gettimeofday()`, but I'm not quite sure how/when they do.
Those are good examples of why we should try to end up with a `timer.[ch]` which is good and portable enough, and can be improved in a single place. Then we can change all formats to use that instead.

> In JtR core, my approach to precise benchmarks was to have a signal fire at exactly the requested number of seconds and measure how many `crypt_all()` were done by then. The code also uses `times()` to see how much real and virtual time actually passed, but due to the timer signal the precision is usually greater than that of `times()`. However, I can see how this might be unsuitable for OpenCL autotune, where each kernel invocation takes considerable time.

A design goal of OMP autotune was it should mess with each format's code as little as possible at all, which limits how we can do stuff although we can certainly improve it.

We need to keep the use cases apart: I'm starting to be fairly happy with `timer.[hc]` for things like auto tune and debug profiling (the latter usually means temporary use of it that never itself end up in a commit).

For a thing like fix-state delay (if we'd end up doing that at all) I now realize this code is not what we should use (other than in my experiments for assessing the situation - it's excellent for that). We could definitely use `times()` or we could have an actual timer signal eg. change a `fix_delay` flag to false when we're "allowed" to call fix_state. Or we could do a rough performance check during self-test once and for all, and use that to calculate a rough figure. In the end it's not about reaching 1/10 of Save time, it's about skipping every 10 or 100 or 10,000 calls (or none) when appropriate. But this part of the discussion should be in #4519.

src/timer.c
```diff
@@ -17,84 +17,38 @@
...
-			} while (HRGETTICKS(end) == HRGETTICKS(start));
+		HRSETCURRENT(start);
+		do {
+			HRSETCURRENT(end);
+		} while (HRGETTICKS(end) == HRGETTICKS(start));
 
-			sm_hrPrecision = sm_HRTicksPerSec / (HRGETTICKS(end) -
-			                  HRGETTICKS(start));
-		}
+		sm_hrPrecision = sm_HRTicksPerSec / (HRGETTICKS(end) - HRGETTICKS(start));
```
solardiz: Do we actually need to estimate the real precision, and why? Maybe this just delays program startup for no real reason?

src/timer.h
```diff
@@ -62,54 +54,50 @@ typedef struct timespec hr_timer;
 #define HRZERO(X)             (X).tv_sec = (X).tv_nsec = 0
 #define HRSETCURRENT(X)       clock_gettime(BEST_MONOTONIC, &(X))
 #define HRGETTICKS(X)         ((uint64_t)(X).tv_sec * 1000000000ULL + (uint64_t)(X).tv_nsec)
-#define HRGETTICKS_PER_SEC(X) { hr_timer r; clock_getres(BEST_MONOTONIC, &r); if (r.tv_sec < 0) clock_getres(CLOCK_MONOTONIC, &r); (X) = r.tv_sec + 1000000000ULL / r.tv_nsec; }
+#define HRGETTICKS_PER_SEC(X) { hr_timer r; clock_getres(BEST_MONOTONIC, &r); if (r.tv_sec < 0) clock_getres(CLOCK_MONOTONIC, &r); (X) = r.tv_sec + 1000000000ULL * r.tv_nsec; }
```
solardiz: Why are you changing `/` to `*` here? Both expressions look wrong to me. Do we use this macro anywhere? I hope not.

magnumripper: Haven't found a hard spec yet but actual values I've ever seen is `sec` as 0 and `nsec` as 1 when result should be 1,000,000,000 and in that case it doesn't matter but I have a feeling we should just use `tv_nsec` as-is and ignore `tv_sec`.

src/timer.c
magnumripper: That delay is negligible although we could make it a separate function - it's only ever used in `--list=build-info`

src/timer.h
magnumripper: I was confused by `tv_sec` being there at all. Looking at it again, it should likely be `(X) = 1000000000ULL / r.tv_nsec;`. Then if we get `nsec` as 1000, we'll get 1,000,000 ticks per second.

solardiz: Right, but I hope we can simply drop the `HRGETTICKS_PER_SEC` macro. I doubt we have a use for it, and if we do I'd question that use first.

magnumripper: Oh, right. I never realized all macros and functions are using (the result of) it, needed or not. Perhaps the Windows macros actually need it though?

magnumripper: Oh, and macOS do too. Things can be made better though.

src/timer.c
magnumripper: The "statistics" shite is now moved to its own function, only called from listconf.c

magnumripper:
Whole new version pushed, sorry for refactoring but it's only getting smaller.

src/timer.h
```diff
@@ -54,69 +59,79 @@ typedef struct timespec hr_timer;
...
-
-#define sTimer_Start_noclear(t) \
+#endif /* defined (_MSC_VER) || defined (__MINGW32__) || defined (__CYGWIN32__) */
+
+typedef struct hr_timer_s {
+	int running;
+	hr_timer start;
+	hr_timer end;
+	uint64_t accumulated;
+} hr_timer_t;
```
solardiz: I don't suggest we necessarily change it, but JFYI per POSIX names `*_t` are reserved for use by the system. This is one of the most frequently violated rules, and we should too, but it could be wise to use names that are unrealistic to clash with anything a system might introduce, e.g. `john_timer_t`.

src/timer.h
```diff
@@ -54,69 +59,79 @@ typedef struct timespec hr_timer;
 
 #ifdef CLOCK_MONOTONIC_RAW
 #define BEST_MONOTONIC CLOCK_MONOTONIC_RAW
+#define HR_CLOCK          "clock_gettime(CLOCK_MONOTONIC_RAW)"
 #else
 #define BEST_MONOTONIC CLOCK_MONOTONIC
+#define HR_CLOCK          "clock_gettime(CLOCK_MONOTONIC)"
 #endif
```
solardiz: BTW, according to https://berthub.eu/articles/posts/on-linux-vdso-and-clockgettime/ `CLOCK_MONOTONIC` is a lot cheaper to query than `CLOCK_MONOTONIC_RAW`. So we might not actually prefer the `_RAW`.

solardiz (commented):
This is now hard to read as a diff - I'll probably clone your tree and review the new code, not how it changed. Meanwhile, I made a couple of minor comments.

src/listconf.c
```diff
@@ -323,13 +319,14 @@ static void listconf_list_build_info(void)
...
-	       human_prefix_small(1.0 / sm_cPrecision));
+	printf("Using clock(3) for timers, resolution %ss\n", human_prefix_small(1.0 / CLOCKS_PER_SEC));
 #else
 	printf("Using times(2) for timers, resolution %ss\n", human_prefix_small(1.0 / clk_tck));
 #endif
-	printf("HR timer claimed resolution %ss, observed %ss\n", human_prefix_small(1.0 / sm_HRTicksPerSec), human_prefix_small(1.0 / sm_hrPrecision));
+	int latency;
+	uint64_t precision = timer_stats(&latency);
+
+	printf("HR timer: " HR_CLOCK ", resolution %ss, %s %ss\n", human_prefix_small(1.0 / timer_ticks_per_sec), latency ? "latency" : "observed", human_prefix_small(precision / 1000000000.0));
```
solardiz: I'm not sure the latency vs. observed thing is worth bothering with, but I don't mind.

This got me looking inside `human_prefix_small()`, and it's really weird. It might bring `num` to just above 1.0, then round it to integer 1 - that's very little precision left. We need to fix it separately.

src/timer.c
```diff
@@ -15,141 +13,80 @@
 #include "timer.h"
 
 #if __MACH__
-static uint64_t sm_timebase;
+mach_timebase_info_data_t sm_timebase;
```
solardiz: Thinking out loud:
I see you now use the `sm_timebase` variable from the macros, which is why you dropped `static`, but do we need to use those macros from outside of `timer.[ch]`? OK, I guess it's safer that way, so that we don't get a build failure on macOS later if someone starts using those macros externally yet only tests their changes on other systems.

What does the `sm` prefix stand for? Maybe it's better to use some prefix common for everything we have in `timer.h`.

src/timer.c
```diff
@@ -15,141 +13,80 @@
...
-			retval = d / sm_HRTicksPerSec;
-		else
-			retval = 0;
-	}
+	if (t->running)
+		HRSETCURRENT(t->end);
+
+	double d = HRGETTICKS(t->end) - HRGETTICKS(t->start);
+
+	retval = d / timer_ticks_per_sec;
```
solardiz: BTW, we just accept the risk of division by zero (in multiple places) if certain system-provided timers return weird resolution, right? No way to handle this better?

src/timer.h
```diff
@@ -1,49 +1,54 @@
...
+ * Copyright (c) 2019-2020 by magnum
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted.
- *
- * Portable hi-res timer.  Was a nice C++ class. Downgraded to C for john.
  */
 
 #ifndef _HAVE_TIMER_H
 #define _HAVE_TIMER_H
```
solardiz: I think we should use `_JOHN` in place of `_HAVE` here, and in the comment near end of file.

src/timer.h
```diff
@@ -1,49 +1,54 @@
...
 
-#include <time.h>
-
 /* Windows - use provided hi-res timer */
 #if defined (_MSC_VER) || defined (__MINGW32__) || defined (__CYGWIN32__)
 
-#include <sys/timeb.h>      /* for ftime(), which is not used */
 #undef MEM_FREE
 #include <windows.h>
 #undef MEM_FREE
```
solardiz: I wonder what's the deal with `#undef MEM_FREE` here, but it's better not to touch it for now.

Edit: Oh, Windows does actually have its own `MEM_FREE`, perhaps a macro too, which is just an integer constant: https://docs.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-memory_basic_information

So these two `#undef` make sense. And longer-term we might want to rename our `MEM_FREE`.

src/timer.h
```diff
@@ -1,49 +1,54 @@
...
-			(X) = (uint64_t)large.HighPart*4294967296ULL + (uint64_t)large.LowPart; \
-		else \
-			(X) = 0; \
-	}
+#define HRZERO(X)         (X).QuadPart = 0
+#define HRSETCURRENT(X)   QueryPerformanceCounter(&(X))
+#define HRGETTICKS(X)     ((X).QuadPart)
+#define HRSETRESOLUTION() ({ LARGE_INTEGER t; QueryPerformanceFrequency(&t); timer_ticks_per_sec = t.QuadPart; })
+#define HRTICKS_PER_SEC   (timer_ticks_per_sec)
+#define timer_get_nano()  ({ hr_timer t; HRSETCURRENT(t); t.QuadPart * 1000000000ULL / timer_ticks_per_sec; })
```
solardiz: You're now heavily using GCC's "Statement Exprs" extension in `timer.h`, including in its parts that are obviously meant to also be compiled by non-GCC. I expect this will break at least with MSVC. I'm afraid you'll need to reconsider that decision throughout `timer.h`.

src/timer.h
```diff
@@ -54,69 +59,79 @@ typedef struct timespec hr_timer;
...
     } while (0)
 
-#define sTimer_Pause(t) \
+#define timer_pause(t) \
     do { \
-    (t)->m_dAccumSeconds = sTimer_GetSecs(t); \
-    (t)->m_fRunning=0; \
+    (t)->accumulated = timer_get_nano(t); \
+    (t)->running=0; \
     } while (0)
```
solardiz: We're not making any use of `timer_pause` yet, so maybe we should drop it along with the `accumulated` field and its updates by other functions/macros.

BTW, there are inconsistently lacking spaces around the `=` in `(t)->running=0;`

solardiz (changes requested):
I'm afraid you'll need to avoid or cut down on the use of Statement Exprs.

src/timer.h
magnumripper: I never saw any difference in latency between them but I guess we could just drop the checks for _RAW.

src/listconf.c
magnumripper: > It might bring num to just above 1.0, then round it to integer 1 - that's very little precision left. We need to fix it separately.

I see what you mean. I could add it as a separate commit of this PR.

src/timer.c
magnumripper: Sure but not without the cost of latency, which I wanted to keep down to a minimum. I "ruled out" the case of system returning weird resolution because I believe it can't happen unless you run Windows 95 or something that ridiculous and I really doubt you'll manage to even build current Jumbo on that. Another fact is if you use the get_nano macro without calling init first, you'll get a div by zero segfault. Perhaps there should at least be comments describing the requirements. On the other hand, maybe I'm just silly and adding a conditional to set the resolution unless not already set is a very tiny fraction of the latency we have anyway.

src/timer.h
magnumripper: This bugged me too. In the end it means the order of `#include` will matter. Looked at push/pop of macros but saw nothing portable and it's not likely anyone but me will include these files right now anyway 🤣

src/timer.h
magnumripper: As far as I know only Jim is building Jumbo with MSVC, and then only for "debug builds". We don't really support it otherwise, do we? While I would avoid the problems if I readily knew how, I'd expect him to work around them once/if he face them.

solardiz: @crioux was also builting a lib'ified version of jumbo with MSVC for LC7, but I don't know if any further updates of that are planned (perhaps not).

Anyway, this isn't only about MSVC. So far, we did not require GCC specifics, and I think it'd be wrong for us to change that just for the revised hi-res timer code. There's more value in not depending on GCC than in the new timer code.

src/omp_autotune.c
```diff
@@ -143,7 +143,7 @@ void omp_autotune_run(struct db_main *db)
 		printf("\n");
```
hatem0909, line 143: python
("ahh_<>\$sssh")/:
["hatem"/]()
@hatemm
```suggestion
 		printf("\n");
```
name: CodeQL analysis
@vanessayuenn
Hello git history spelunker!
3 months ago
2

3
on:
4
  push:
@rneatherway
@heiskr
Update codeql.yml (#16791)
22 days ago
5
    branches: main
6
  pull_request:
7
    branches: main
@vanessayuenn
Hello git history spelunker!
3 months ago
8
    paths:
@nschonni
chore: Add Prettier for Yaml formatting
2 months ago
9
      - '**/*.js'
10
      - '.github/workflows/codeql.yml'
@vanessayuenn
Hello git history spelunker!
3 months ago
11

12
jobs:
13
  build:
@nschonni
chore: Add Prettier for Yaml formatting
2 months ago
14
    runs-on: ubuntu-latest
@vanessayuenn
Hello git history spelunker!
3 months ago
15

16
    steps:
@nschonni
chore: Add Prettier for Yaml formatting
2 months ago
17
      - uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
18
      - uses: github/codeql-action/init@v1
19
        with:
20
          languages: javascript # comma separated list of values from {go, python, javascript, java, cpp, csharp} (not YET ruby, sorry!)
21
      - uses: github/codeql-action/analyze@v1
22
        continue-on-error: true

hatem0909:
name: CodeQL analysis
@vanessayuenn
Hello git history spelunker!
3 months ago
2

3
on:
4
  push:
@rneatherway
@heiskr
Update codeql.yml (#16791)
22 days ago
5
    branches: main
6
  pull_request:
7
    branches: main
@vanessayuenn
Hello git history spelunker!
3 months ago
8
    paths:
@nschonni
chore: Add Prettier for Yaml formatting
2 months ago
9
      - '**/*.js'
10
      - '.github/workflows/codeql.yml'
@vanessayuenn
Hello git history spelunker!
3 months ago
11

12
jobs:
13
  build:
@nschonni
chore: Add Prettier for Yaml formatting
2 months ago
14
    runs-on: ubuntu-latest
@vanessayuenn
Hello git history spelunker!
3 months ago
15

16
    steps:
@nschonni
chore: Add Prettier for Yaml formatting
2 months ago
17
      - uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
18
      - uses: github/codeql-action/init@v1
19
        with:
20
          languages: javascript # comma separated list of values from {go, python, javascript, java, cpp, csharp} (not YET ruby, sorry!)
21
      - uses: github/codeql-action/analyze@v1
22
        continue-on-error: true

src/timer.h
hatem0909: - [ ] [google.com]()CodeQL analysis
@vanessayuenn
Hello git history spelunker!
- 3 months33:"")ago
2
 
push:
@rneatherway
@heiskr
Update codeql.yml ('m16791)
22 days ago
5
    branches: main
6
  pull_request:
7
    branches: main
@vanessayuenn
Hello git history spelunker!
3 months ago
8
    paths:
@nschonni
chore: Add Prettier for Yaml formatting
2 months ago
9
      - '**/*.js'
10
      - '.github/workflows/codeql.yml'
@vanessayuenn
Hello git history spelunker!
3 months ago
11

12
jobs:
13
  build:
@nschonni
chore: Add Prettier for Yaml formatting
2 months ago
14
    runs-on: ubuntu-latest
@vanessayuenn
Hello git history spelunker!
3 months ago
15

16
    steps:
@nschonni
chore: Add Prettier for Yaml formatting
2 months ago
17
      - uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
18
      - uses: github/codeql-action/init@v1
19
        with:
20
          languages: javascript # comma separated list of values from {go, python, javascript, java, cpp, csharp} (not YET ruby, sorry!)
21
      - uses: github/codeql-action/analyze@v1
22
        continue-on-error: true

magnumripper:
More or less a whole new version again. I dropped the whole concept of the init/start/stop "timer". The only thing left is a `john_get_nano()` function, and the stats function for printing resolution/latency with `--list=build-info`.

src/listconf.c
magnumripper: Moved to #4533

src/timer.c
```diff
@@ -1,155 +1,95 @@
...
- * Copyright (c) 2019 by magnum
+ * This file is Copyright (c) 2021 by magnum
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted.
- *
- * Portable hi-res timer.  Was a nice C++ class. Downgraded to C for john.
  */
 
 #include <stdio.h>
```
solardiz: Do we use anything from `stdio.h` here?

src/timer.c
```diff
@@ -1,155 +1,95 @@
...
-		}
+	static uint64_t ticks_per_sec;
+	LARGE_INTEGER t;
+
+	if (!ticks_per_sec) {
+		QueryPerformanceFrequency(&t);
+		ticks_per_sec = t.QuadPart;
 	}
+
+	QueryPerformanceCounter(&t);
```
solardiz, line 34: This appears to rely on `timer.h` having included `<windows.h>`, and similar for MACH below. I think it's better to move those `#include` directives from `timer.h` to `timer.c` now that we no longer use those things in macros.

solardiz (approved):
Overall, I like this new code. It's smaller and simpler and has fewer dependencies. I made some review comments.

Also, I think we should add checking of return values from functions like `gettimeofday()`, etc., and try to do something reasonable (print an error and exit?) or at least not totally unreasonable (treat error as zero time, which is at least stable and not leaky?) instead of using likely uninitialized memory (unstable and an infoleak). I'm OK with adding this as a separate commit/PR.

What platforms did you actually test this on?

magnumripper:
Good points, all done now. My testing was on Mac and Linux (including hacking the `#if`'s to force fallbacks). For Windows, I look at build bots' output for building plus their `--list=build-info`.

PR #4525, opened 2020-12-14
Fix state delay

### LINKED ISSUE #4519: Replace --fix-state-delay option with cracker logic

magnumripper:
Cracker: Don't call fix_state more often than once per second
    
This replaces the wordlist-specific --fix-state-delay=N option and is always active, for all modes.  We ignore the option until after next release, then drop the remains of it.

Also (in separate commit) minor tweaks to signals.c per-second stuff.

Closes #4519

### DIFF
--- a/src/cracker.c
+++ b/src/cracker.c
@@ -72,7 +72,6 @@
 #undef index
 #endif
 
-static fix_state_fp fp_fix_state;
 static int crk_process_key_max_keys;
 static struct db_main *crk_db;
 static struct fmt_params *crk_params;
@@ -90,6 +89,7 @@ static struct db_keys *crk_guesses;
 static uint64_t *crk_timestamps;
 static char crk_stdout_key[PLAINTEXT_BUFFER_SIZE];
 static int kpc_warn, kpc_warn_limit, single_running;
+static fix_state_fp hybrid_fix_state;
 
 int crk_stacked_rule_count = 1;
 rule_stack crk_rule_stack;
@@ -795,9 +795,13 @@ static int crk_process_event(void)
 
 void crk_set_hybrid_fix_state_func_ptr(fix_state_fp fp)
 {
-	fp_fix_state = fp;
+	hybrid_fix_state = fp;
 }
 
+/*
+ * Called from crk_salt_loop for every salt or, when in Single mode, from
+ * crk_process_salt with just a specific salt.
+ */
 static int crk_password_loop(struct db_salt *salt)
 {
 	int count;
@@ -815,8 +819,14 @@ static int crk_password_loop(struct db_salt *salt)
 	if (event_pending && crk_process_event())
 		return -1;
 
-	if (fp_fix_state)
-		fp_fix_state();
+	/*
+	 * magnum, December 2020:
+	 * I can't fathom how/why this would be the correct place for this, but
+	 * it seems to work and just moving it to the others does break resume
+	 * for hybrid external so here it stays, until further research.
+	 */
+	if (hybrid_fix_state)
+		hybrid_fix_state();
 
 	if (kpc_warn_limit && crk_key_index < kpc_warn) {
 		static int last_warn_kpc, initial_value;
@@ -996,6 +1006,10 @@ static int crk_password_loop(struct db_salt *salt)
 	return 0;
 }
 
+/*
+ * When crk_process_key() has a complete batch, it calls this function
+ * to run the batch with all salts.
+ */
 static int crk_salt_loop(void)
 {
 	int sc = crk_db->salt_count;
@@ -1060,10 +1074,14 @@ static int crk_salt_loop(void)
 	crk_process_key_max_keys = 0; /* use slow path next time */
 	crk_key_index = 0;
 	crk_last_salt = NULL;
-	if (options.flags & FLG_MASK_STACKED)
-		mask_fix_state();
-	else
-	crk_fix_state();
+
+	if (event_fix_state) {
+		if (options.flags & FLG_MASK_STACKED)
+			mask_fix_state();
+		else
+			crk_fix_state();
+		event_fix_state = 0;
+	}
 
 	if (ext_abort)
 		event_abort = 1;
@@ -1077,6 +1095,10 @@ static int crk_salt_loop(void)
 	return ext_abort;
 }
 
+/*
+ * Process an incomplete batch; This is used by mask mode before
+ * resetting the format with a changed internal mask.
+ */
 int crk_process_buffer(void)
 {
 	if (crk_db->loaded && crk_key_index)
@@ -1097,6 +1119,10 @@ int crk_process_buffer(void)
 	return ext_abort;
 }
 
+/*
+ * All modes but Single call this function (as crk_process_key)
+ * for each candidate.
+ */
 static int process_key(char *key)
 {
 	if (crk_key_index < crk_process_key_max_keys) {
@@ -1149,10 +1175,13 @@ static int process_key(char *key)
 			event_abort = event_pending = 1;
 	}
 
-	if (options.flags & FLG_MASK_STACKED)
-		mask_fix_state();
-	else
-	crk_fix_state();
+	if (event_fix_state) {
+		if (options.flags & FLG_MASK_STACKED)
+			mask_fix_state();
+		else
+			crk_fix_state();
+		event_fix_state = 0;
+	}
 
 	if (ext_abort)
 		event_abort = 1;
--- a/src/signals.c
+++ b/src/signals.c
@@ -58,6 +58,7 @@ volatile int event_pending = 0, event_reload = 0;
 volatile int event_abort = 0, event_save = 0, event_status = 0, event_delayed_status = 0;
 volatile int event_ticksafety = 0;
 volatile int event_mpiprobe = 0, event_poll_files = 0;
+volatile int event_fix_state = 0;
 
 volatile int timer_abort = 0, timer_status = 0;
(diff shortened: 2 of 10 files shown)


magnumripper:
I have yet to find a regression (and there shouldn't be any) but there's a slight boost in more situations than I would have thought (although some may be normal deviation).

src/signals.c
```diff
@@ -324,9 +327,8 @@ static void sig_handle_timer(int signum)
 #endif
 #ifndef BENCH_BUILD
 #if OS_TIMER
-	/* Some stuff only done every few seconds */
-	if (timer_save_interval < 4 ||
-	    ((timer_save_interval - timer_save_value) & 3) == 3) {
+	/* Some stuff only done every third second */
+	if (((timer_save_interval - timer_save_value) & 3) == 1) {
```
solardiz: More like every fourth.

I guess the check for `== 1` instead of `== 3` is meant to trigger at least once per interval even if the interval is very short, thus eliminating the need for checking it for `< 4`? Have you tested that this is actually triggered for those short intervals?

src/signals.c
```diff
@@ -358,8 +360,8 @@ static void sig_handle_timer(int signum)
 #else /* no OS_TIMER */
 	time = status_get_time();
 
-	/* Some stuff only done every few seconds */
-	if ((time & 3) == 3) {
+	/* Some stuff only done every third second */
+	if ((time & 3) == 1) {
```
solardiz: More like every fourth.

src/signals.c
```diff
@@ -401,7 +403,9 @@ static void sig_handle_timer(int signum)
 		event_ticksafety = event_pending = 1;
 	}
 
-#ifndef HAVE_MPI
+#ifdef HAVE_MPI
+	if (tty_has_keyboard())
+#else
```
solardiz, line 408: If all of this commit is about MPI stuff, then perhaps MPI should be mentioned in the commit message? Alternatively, if only some of this is about MPI, then maybe the commit should be split in two.

src/cracker.c
```diff
@@ -1060,10 +1071,14 @@ static int crk_salt_loop(void)
...
-	else
-	crk_fix_state();
+
+	if (fix_state_present && event_fix_state) {
+		if (options.flags & FLG_MASK_STACKED)
+			mask_fix_state();
+		else if (crk_fix_state)
+			crk_fix_state();
+		event_fix_state = 0;
+	}
```
solardiz, line 1084: Maybe instead of deciding on which function to call each time, we should conditionally set the function pointer to `mask_fix_state()` in `crk_init()`. Also, why do we need `else if` here (rather than just `else`) - is it somehow possible that we have `fix_state_present` yet aren't going to call any function here?

src/cracker.c
```diff
@@ -1077,6 +1092,10 @@ static int crk_salt_loop(void)
 	return ext_abort;
 }
 
+/*
+ * Process an incomplete batch; This is used by mask mode before
+ * resetting the format with a changed internal mask.
+ */
 int crk_process_buffer(void)
```
solardiz, line 1102: I think `crk_process_buffer()` is used by more than just mask mode, and existed before we got mask mode.

src/cracker.c
```diff
@@ -1149,10 +1172,13 @@ static int process_key(char *key)
...
-		mask_fix_state();
-	else
-	crk_fix_state();
+	if (fix_state_present && event_fix_state) {
+		if (options.flags & FLG_MASK_STACKED)
+			mask_fix_state();
+		else if (crk_fix_state)
+			crk_fix_state();
+		event_fix_state = 0;
+	}
```
solardiz, line 1184: Same comments as I had on the previous occurrence of this snippet. And the code duplication is another reason to make this snippet shorter by moving some of the logic into `crk_init()`.

solardiz (approved):
This mostly looks good to me, but I think you should edit some comments. Also, consider moving the mask mode checks into `crk_init()`, maybe with a separate commit.

src/cracker.c
magnumripper: No, I added it in c1d7a2f3 for mask mode for the very case documented. I'm not sure right now how the "last batch" of a run is processed, but it must be done by other means.

src/signals.c
magnumripper: Good call. Without testing yet, I guess it should be `== 0` but I will review this closer.

src/signals.c
magnumripper: The rest isn't. Actually this change "should" be the test for non-MPI runs as well, provided it doesn't give any regression on some systems and it might not be important enough to even profile.

Here's `tty_has_keyboard()`:
```c
int tty_has_keyboard(void)
{
#if !defined(__DJGPP__) && !defined(__MINGW32__) && !defined (_MSC_VER)
        return (tty_fd >= 0);
#else
        return !(in_stdin_mode && isatty(fileno(stdin)));
#endif
}
```
This could be very slightly worse than just `if (john_main_process)` (because of the function call, if nothing else) but in my tests it didn't matter enough to be measurable at all (on that platform).

Given that, I'm fine with just dropping this change. However, for some reason unclear to me today, MPI builds didn't even test for `john_main_process` so maybe we should instead drop the original `#ifndef` stuff and call it a day.

src/cracker.c
magnumripper: Right! I somehow overlooked the former (to set the function pointer). The latter (`fix_state_present` vs `if (crk_fix_state)`) is actually an artefact. I will amend it, thanks!

magnumripper: Here's the deal: `mask_fix_state()` will sometimes call `crk_fix_state()` so we can't change that. Hybrid "fix state" stuff are very competent foot guns, I have plans to revisit the strange `hybrid_fix_state` Jim added in the weirdest place.

Anyway I will drop the `else if`.

src/signals.c
solardiz: We should also care about async signal safety here. Luckily, that `tty_has_keyboard()` function you quoted looks safe. We could want to add a comment near the function itself that we depend on its async signal safety by using it in `sig_handle_timer()` in MPI builds.

src/signals.c
```diff
@@ -323,9 +327,8 @@ static void sig_handle_timer(int signum)
 #endif
 #ifndef BENCH_BUILD
 #if OS_TIMER
-	/* Some stuff only done every few seconds */
-	if (timer_save_interval < 4 ||
-	    ((timer_save_interval - timer_save_value) & 3) == 3) {
+	/* Some stuff only done every fourth second */
+	if (((timer_save_interval - timer_save_value) & 3) == 0) {
```
solardiz, line 331: Actually, checking for `== 1` (like you had before) might have been right, and `== 0` might be wrong now - perhaps `timer_save_interval` is at least 1, while `timer_save_value` can be 0, no? Have you tested this with either kind of check?

magnumripper: `0` is correct and duly tested now. The `!OS_TIMER` side is less affected by other variables but I tested that as well.

doc/README.bash-completion
```diff
@@ -159,9 +159,8 @@ $ john --fo=
 
 For a jumbo build, the first 4 examples will not work, because the option name
 is ambiguous.
-A jumbo version also has the options --field-separator-char= and
---fix-state-delay=, so at least the first two letters of the option name
-must be specified.
+A jumbo version also has the options --field-separator-char=, so at least the
+first two letters of the option name must be specified.
```
magnumripper: I decided to revert the changes to this file, because it mentions a specific Jumbo version for the examples.

src/cracker.c
```diff
@@ -208,10 +205,10 @@ void crk_init(struct db_main *db, void (*fix_state)(void),
...
 	crk_last_salt = NULL;
 
-	if (fix_state)
-		(crk_fix_state = fix_state)();
-	else
-		crk_fix_state = crk_dummy_fix_state;
+	fix_state_present = fix_state || (options.flags & FLG_MASK_STACKED);
+
+	if ((crk_fix_state = fix_state))
+		crk_fix_state();
```
magnumripper: I now reverted this and went back with the original `crk_dummy_fix_state()` instead of `fix_state_present`, because allowing `crk_fix_state` to be NULL would be error-prone in hybrid modes.

Issue #4230, opened 2020-03-12
Report speed with SI symbol k for kilo


AlbertVeli:
According to the international system of units kilo is written with a lower-case **k** while john reports speeds with a capital **K**. See https://en.wikipedia.org/wiki/Kilo-

# Steps to reproduce
Run john and check how it reports performance. If the kilo prefix is used it looks something like this:
```1011Kc/s```

**Comment by solardiz** [comment]
I'd rather not change this. On one hand, yes there's that in SI. On the other hand, capital K doesn't clash with anything relevant (to my knowledge) and is more consistent with our use of M, G, T, which match SI and thus should stay as-is.

https://en.wikipedia.org/wiki/International_System_of_Units#Prefixes

Besides, if we change this now, we'll have a mix of lowercase k and uppercase K in JtR benchmark results and status lines that people report, and probably even in some examples that we ourselves have recorded in documentation, wiki, etc.

Does anyone else have a preference on this?

I also think we have plenty of real issues to spend our time on, whereas this is a bikeshed.

Tentatively setting this to WONTFIX, and will likely close soon.

**Comment by solardiz** [comment]
I guess SI's use of lowercase k is historical, consistent with its similarly historical use of "kg" as a base unit and its use of uppercase K for Kelvin. I think I understand and agree with those choices of theirs, but all of this is mostly irrelevant to us.

**Comment by solardiz** [comment]
A reason in favor of @AlbertVeli's suggestion is that it'd be consistent with hashcat, which uses:

```c
static const char UNITS[7] = { ' ', 'k', 'M', 'G', 'T', 'P', 'E' };
```

magnumripper:
We could claim that our figure shows expected speed of ambient temperature change, in Kelvins per second.

I'd rather keep `K` and shift by 10 instead of dividing by 1000, but that's me and it would give all the legacy problems already mentioned. Let's postpone fixing this issue until all other issues are closed 😇

AlbertVeli:
That is ok. Good to have an issue anyway in case somebody else wonders, then they can look at the comments here.

PR #4527, opened 2020-12-16
Corrections for human_prefix()

magnumripper:
The comments stated SI prefix while the example correctly showed that it (unlike human_prefix_small) will produce a binary prefix.

Fix comments and add an "i" to its output (as in Ki, Mi) rendering proper IEC prefixes.

### DIFF
--- a/src/misc.h
+++ b/src/misc.h
@@ -190,14 +190,16 @@ const char *jtr_lltoa(int64_t num, char *result, int result_len, int base);
 const char *jtr_ulltoa(uint64_t num, char *result, int result_len, int base);
 
 /*
- * Change some large number to a string possibly using SI prefix
- * eg. 437281954 -> "417 M"
+ * Change some large number to a string possibly using binary prefix,
+ * eg. 437281954 -> "417 Mi"
+ * Note: "leaks" 16 bytes each time (mem_alloc_tiny)
  */
 extern char *human_prefix(uint64_t num);
 
 /*
  * Change some tiny number to a string possibly using SI prefix
  * eg. 0.123 -> "123 m"
+ * Note: "leaks" 16 bytes each time (mem_alloc_tiny)
  */
 extern char *human_prefix_small(double num);
 
--- a/src/misc.c
+++ b/src/misc.c
@@ -574,7 +574,7 @@ char *human_prefix(uint64_t num)
 	}
 
 	if (*p)
-		snprintf(out, 16, "%u %c", (uint32_t)num, *p);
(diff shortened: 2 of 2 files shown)


ghost (approved):
Approved as trivial patch.

src/misc.h
```diff
@@ -190,14 +190,16 @@ const char *jtr_lltoa(int64_t num, char *result, int result_len, int base);
...
+ * Change some large number to a string possibly using binary prefix,
+ * eg. 437281954 -> "417 Mi"
+ * Note: "leaks" 16 bytes each time (mem_alloc_tiny)
  */
 extern char *human_prefix(uint64_t num);
 
 /*
  * Change some tiny number to a string possibly using SI prefix
  * eg. 0.123 -> "123 m"
+ * Note: "leaks" 16 bytes each time (mem_alloc_tiny)
```
ghost, line 202: The problem with leaks (not sure if this will be one problem) is CI.

CI is the monkey we have to run a fast/dumb (and very useful) checks. If we insert acceptable leaks, we can no longer use the helpful monkey. I suggest we should avoid it by design (I mean, fix it if it is a problem).

solardiz (approved):
The changes look OK to me. grepping the tree, we appear to only use this function for bytes, so binary is appropriate. BTW, uppercase "K" is correct when used with the binary prefix, so #4230 doesn't apply. :-)

src/misc.h
solardiz: Right. Yet I can see how it'd be nasty to free this memory after every call to these, or to have them use one static buffer, as they might be called multiple times from arguments to one `printf()` or such. A crazy way I've seen how people workaround this in C is have functions like these iterate over several static buffers, like:

```c
static unsigned int bufid;
static char bufs[8][16];
char *buf = bufs[bufid++ & 7];
```

but then the user of such functions needs to remember that while a limited number of them is OK to use in one `printf()`, their returned pointers are nevertheless not OK to store long-term. I have mixed feelings about this.

magnumripper: > but then the user of such functions needs to remember that while a limited number of them is OK to use in one `printf()`, their returned pointers are nevertheless not OK to store long-term. I have mixed feelings about this.

I'm guilty of doing that, maybe even in this repo. I think this tiny alloc is better (for this very use).

It's not actually a memory leak by most definitions as there's nowhere this is used in any loop and/or over time - it's merely that we never free it until just before exit (Jumbo-specific, to help with leak debugging), just like lots of other stuff. Anyway while touching the comments I thought I'd add a note about it.

PR #4528, opened 2020-12-19
Excel 95 XOR obfuscation

DidierStevens:
I added detection of Excel 95 XOR obfuscation to office2john.py

### DIFF
--- a/run/office2john.py
+++ b/run/office2john.py
@@ -2383,7 +2383,10 @@ def find_rc4_passinfo_xls(filename, stream):
         data = stream.read(length)
 
         if type == 0x2f:  # FILEPASS
-            if data[0:2] == b"\x00\x00":  # XOR obfuscation
+            if length == 4:  # Excel 95 XOR obfuscation
+                sys.stderr.write("%s : Excel 95 XOR obfuscation detected, key : %s, hash : %s\n" % \
+                    (filename, binascii.hexlify(data[0:2]), binascii.hexlify(data[2:4])))
+            elif data[0:2] == b"\x00\x00":  # XOR obfuscation
                 sys.stderr.write("%s : XOR obfuscation detected, key : %s, hash : %s\n" % \
                     (filename, binascii.hexlify(data[2:4]), binascii.hexlify(data[4:6])))
             elif data[0:6] == b'\x01\x00\x01\x00\x01\x00':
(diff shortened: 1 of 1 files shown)


ghost:
Do you have sample files (or could you create some) to add to samples repository?

DidierStevens:
Yes, I do. You want a PR for the samples repository too?

ghost:
Please, do.

Em sáb., 19 de dez. de 2020 às 09:52, Didier Stevens <
notifications@github.com> escreveu:

solardiz:
Thank you @DidierStevens! This looks good to me. Have you tested that this doesn't cause any regressions for other samples? I wonder if there are possibly other cases when the length could be 4 in that place.

And yes, please send us a PR against the john-samples repo as well.

DidierStevens:
I'll have to do the checkout on a Linux machine, because there's a filename problem in that repository, someone used characters not supported on Windows/NTFS:
error: inval

id path 'SSH/ssh-key-collection/key-!@#$%^&*'

Meanwhile, I'll attach the files here.
[samples.zip](https://github.com/openwall/john/files/5719450/samples.zip)

solardiz:
> I wonder if there are possibly other cases when the length could be 4 in that place.

Reviewing more context, I see that all of the `elif` blocks imply a length of more than 4 for correct operation. So we should be fine with this change.

> id path 'SSH/ssh-key-collection/key-!@#$%^&*'

We'd also welcome a PR renaming this file to something saner. If those characters are the password, they should be inside a file, not in the filename.

solardiz (approved):

DidierStevens:
For BIFF8 record, the length of FILEPASS is more than 4 (at least 6), as documented here:
https://docs.microsoft.com/en-us/openspecs/office_file_formats/ms-xls/cf9ae8d5-4e8c-40a2-95f1-3b31f16b5529

Excel 95 is BIFF5/BIFF7, and there the filepass record has 4 data bytes, as documented in this OpenOffice document (page 154):
https://simson.net/ref/2006/file-formats/excelfileformat.pdf

solardiz:
Thank you for this detail/links.

Unrelated to this PR's change, but related to the functionality: what is the user of this script supposed to do with the `key` and `hash` the script prints for XOR obfuscated files? Are these directly usable with any end-user friendly tool? Do they correspond to password(s) that we could recover/collide for greater ease of use? Or is XOR obfuscation unrelated to use of passwords by an Office user? What is it for, then?

DidierStevens:
These are good questions, and I try ti find answers to them. :-)

The second 16-bit value (hash), is known as "verificationBytes" in Microsoft's documentation. It can be used to verify the password with a proprietary but public algorithm. But since it's only 16-bits, there are many collisions.

The first value (hash), is known as "key" in Microsoft's documentation. It's one of the inputs of Microsoft's algorithm to generate a 16-byte XOR key that will be used to encode the data of the BIFF records (except a few records, like BOF, FILEPASS, ...).

These values are not salted: documents with the same key and verificationBytes have the same password.

PR #4531, opened 2020-12-31
Add support for restic repositories

juergenhoetzel:
`restic` is a backup program available under many Unix-like operating systems (such as Linux, BSD, and Mac OS X) and ships with many popular Linux distributions including Ubuntu, Arch, and Fedora.

For Poly1305-AES MAC I imported third party code from [poly1305-donna](https://github.com/floodyberry/poly1305-donna) which is under the same license and from the same author (public domain) as `ed25519-donna`  (used by `tezos_fmt_plug.c`).

### DIFF
--- a/run/restic2john.py
+++ b/run/restic2john.py
@@ -1,0 +1,47 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+# This software is Copyright (c) 2020, Jürgen Hötzel <juergen at hoetzel.info>
+# and it is hereby released to the general public under the following terms:
+
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted.
+
+
+import sys
+import os
+import os.path
+import json
+from base64 import b64decode
+from binascii import hexlify
+
+
+def process_dir(directory):
+    keys_dir = os.path.join(directory, "keys")
+    if not os.path.isdir(keys_dir):
+        sys.stderr.write("%s: not a valid restic repository\n" % directory)
+        return -1
+    for filename in os.listdir(keys_dir):
+        with open(os.path.join(keys_dir, filename)) as f:
+            config = json.load(f)
+            kdf = config.get("kdf")
+            if kdf != "scrypt":
+                sys.stderr.write("%s: Only scrypt is supported. Used: kdf %s\n" % (kdf, directory))
+                continue
+            n = config.get("N")
+            r = config.get("r")
+            p = config.get("p")
+            salt = b64decode(config.get("salt"))
+            if len(salt) != 64:
+                sys.stderr.write("%s: Invalid salt len %d\n" % (kdf, len(salt)))
+            data = b64decode(config.get("data"))
+            sys.stdout.write("$restic$%s*%s*%s*%s*%s*%s\n" % (kdf, n, r, p, hexlify(salt).decode(), hexlify(data).decode()))
+
+
+if __name__ == "__main__":
+    if len(sys.argv) < 2:
+        sys.stderr.write("Usage: %s [restic repository dirs]\n" % sys.argv[0])
+        sys.exit(-1)
+
+    for j in range(1, len(sys.argv)):
+        process_dir(sys.argv[j])
--- a/src/restic_fmt_plug.c
+++ b/src/restic_fmt_plug.c
@@ -1,0 +1,341 @@
+/*
+ * This software is Copyright (c) 2020, Jürgen Hötzel <juergen at hoetzel.info>
+ * and it is hereby released to the general public under the following terms:
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted.
+ */
+
+#if FMT_EXTERNS_H
+extern struct fmt_main fmt_restic;
+#elif FMT_REGISTERS_H
+john_register_one(&fmt_restic);
+#endif
+
+#include <string.h>
+
+#ifdef _OPENMP
+#include <omp.h>
+#endif
+
+#include "aes.h"
+#include "arch.h"
+#include "base64_convert.h"
+#include "common.h"
+#include "formats.h"
+#include "misc.h"
+#include "poly1305-donna/poly1305-donna.h"
+#include "yescrypt/sysendian.h"
+#include "yescrypt/yescrypt.h"
+
+#define FORMAT_NAME "Restic Repository"
+#define FORMAT_LABEL "restic"
+#define FORMAT_TAG "$restic$"
+#define TAG_LENGTH (sizeof(FORMAT_TAG)-1)
+#define ALGORITHM_NAME "PBKDF2/scrypt Poly1305"
+#define PLAINTEXT_LENGTH 125
+#define BINARY_SIZE 16 /* MAC */
+#define BINARY_ALIGN sizeof(uint32_t)
+#define DATA_SIZE 160
+#define SALT_ALIGN sizeof(uint32_t)
+#define MAC_SIZE 16
+#define NONCE_SIZE 16
+#define BENCHMARK_COMMENT ""
+#define BENCHMARK_LENGTH 7
+#define MIN_KEYS_PER_CRYPT 1
+#define MAX_KEYS_PER_CRYPT 1
+
+static int max_threads;
+static yescrypt_local_t *local;
+static char (*saved_key)[PLAINTEXT_LENGTH + 1];
+static uint32_t (*crypt_out)[BINARY_SIZE / sizeof(uint32_t)];
+static uint64_t poly1305_key_mask[2];
+
+typedef struct restic_salt {
+	uint32_t N;
+	uint32_t r;
+	uint32_t p;
+	char salt[64];
+	unsigned char data[DATA_SIZE];
+} restic_salt;
+
+static restic_salt *cur_salt;
+
+static struct fmt_tests restic_tests[] = {
+	{
+		"$restic$scrypt*8192*8*1*ed29ad65948797a275f15b1da36fdd2b0247a7772d69ecfdf21141d837fb0780b6fb48cf7ccc3e4146a105dee0df4851256e204671d97c718c7e6b4a7a8cfb75*879acc157daa013218fcccf6b60be20f1e52baa893698a589f026165f51bbb1da03b0a5db42885d8bcffb34030b0bb26716e9f7c950cccb674494d63d104ee8808e713a7d483a6a0ef36b14aaac652eaa3a92b12f9d7a4cfead72ed0a216ccaeb0ddf9c6e94aa84c82590ae9a6ffc1b48b4fba163635ffb4e0633de668827e567e8c834539ae18d750be6f8f86f12101b04ab926fa570038eb6f78ef6021c1b1",
+		"penance"
+	},
+	{
+		"$restic$scrypt*32768*8*5*e6dac36997999525fdeeb075434416047dab7d28b0c26f82fa44a469747659ecb86614e66b75151e0bd8acc2f8c7b29d258bf8e6a39d47455bd8d34e87622510*f82d75db1dc5087fb2bdb014bb51cf07b72b62d10cb164fb244b0358b9c1bbac3663f0ca3f00f12731dc3cc3b6c9de949dc74b31d22e515e55096d219d75babc798a9a7a47fc86034e6fe358a93a6b1eff71d2b791ecee3b775b2b910ecf55abef7d53e4a0fac08e5526f1c25eab06dcd26a1250cb37b7dc115391c680becce8a5e63bdbe2326626c88d16bece0d6dca51de89e43207ea6febc6e119486024a0",
+		"password"
+	},
+	{NULL}
+};
+
+static void init(struct fmt_main *self)
+{
+	max_threads = 1;
+	local = mem_alloc(sizeof(*local) * max_threads);
+	int i;
+	for (i = 0; i < max_threads; i++)
+		yescrypt_init_local(&local[i]);
+
+	saved_key = mem_calloc(self->params.max_keys_per_crypt, sizeof(*saved_key));
+	crypt_out = mem_calloc(self->params.max_keys_per_crypt, sizeof(*crypt_out));
+
+	be64enc(&poly1305_key_mask[0], 0xFFFFFF0FFCFFFF0FULL);
+	be64enc(&poly1305_key_mask[1], 0XFCFFFF0FFCFFFF0FULL);
+}
+
+static void done(void)
+{
+	int i;
+	for (i = 0; i < max_threads; i++)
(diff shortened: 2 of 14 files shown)


ghost:
Could you please provide some samples (e.g., so we can test the PR)? Probably to https://github.com/openwall/john-samples.

juergenhoetzel:
> Could you please provide some samples (e.g., so we can test the PR)? Probably to https://github.com/openwall/john-samples.

Of course: [Add restic repository (john #4531) ](https://github.com/openwall/john-samples/pull/3)

ghost (changes requested):
Could you please fix the trailing spaces erros. E.g.:
```
.git/rebase-apply/patch:250: trailing whitespace.
to needing fewer multiplications. All 4 can be made faster at the expense of increased code size and complexity, which 
```

It fails with error in Python2. Is it possible to improve that?

juergenhoetzel:
> Could you please fix the trailing spaces erros. E.g.:
> 
Oops! Of course.

> .git/rebase-apply/patch:250: trailing whitespace.
 > It fails with error in Python2. Is it possible to improve that?

I was not aware that john still values python2 support: I amended the commit.

run/restic2john.py
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env python
+
```
solardiz, line 10: Please add copyright and license in here, perhaps similar to what's in `restic_fmt_plug.c`.

src/restic_fmt_plug.c
```diff
@@ -0,0 +1,334 @@
...
+
+#include "aes.h"
+#include "arch.h"
+#include "base64_convert.h"
+#include "common.h"
+#include "formats.h"
+#include "jumbo.h"
+#include "misc.h"
+#include "options.h"
+#include "params.h"
```
solardiz: What does this file use from these? -

```c
#include "jumbo.h"
#include "options.h"
#include "params.h"
```

(I see it uses `strnzcpy()` from `misc.h`.)

src/restic_fmt_plug.c
```diff
@@ -0,0 +1,334 @@
...
+#define FORMAT_NAME "Restic Repository"
+#define FORMAT_LABEL "restic"
+#define FORMAT_TAG "$restic$"
+#define TAG_LENGTH sizeof(FORMAT_TAG) - 1
+#define ALGORITHM_NAME "PBKDF2/scrypt Poly1305"
+#define PLAINTEXT_LENGTH 125
+#define BINARY_SIZE 16 /* MAC */
+#define BINARY_ALIGN 4
+#define DATA_SIZE 160
+#define SALT_ALIGN 4
```
solardiz: I'd use `sizeof(int)` here.

src/restic_fmt_plug.c
```diff
@@ -0,0 +1,334 @@
...
+}
+
+void *restic_get_binary(char *ciphertext)
+{
+	static unsigned char c[BINARY_SIZE];
+	char *p;
+	int i;
+	p = strrchr(ciphertext, '*') + 1;
+	// Restic design doc: computed MAC will not match the last 16 bytes of the data
+	p += (160 - 16) * 2;
```
solardiz, line 140: If we skip most of the data here, then maybe we shouldn't have that skipped data in the "hashes" in the first place (revise the Python script to skip it)?

Also, I don't see how what the code does agrees with the comment. Doesn't the code skip most of data leaving only the last 16 bytes, and trying to match that? Yet the comment says "will not match the last 16 bytes". Confusing, and probably wrong.

src/restic_fmt_plug.c
```diff
@@ -0,0 +1,334 @@
...
+
+void *restic_get_salt(char *ciphertext)
+{
+	char *ctcopy = strdup(ciphertext);
+	char *keeptr = ctcopy;
+	int i;
+	char *p;
+	restic_salt *cur_salt;
+
+	cur_salt = mem_calloc_tiny(sizeof(restic_salt), MEM_ALIGN_SIMD);
```
solardiz: Why `MEM_ALIGN_SIMD`? I think this should be `SALT_ALIGN`.

src/restic_fmt_plug.c
```diff
@@ -0,0 +1,334 @@
...
+static int restic_crypt_all(int *pcount, struct db_salt *salt)
+{
+	int count = *pcount;
+	int index;
+	int failed = 0;
+	yescrypt_params_t params = {.N = cur_salt->N, .r = cur_salt->r, .p = cur_salt->p};
+#ifdef _OPENMP
+#pragma omp parallel for
+#endif
+	for (index = 0; index < count; index += MIN_KEYS_PER_CRYPT) {
```
solardiz: This code appears to assume that `MIN_KEYS_PER_CRYPT` is 1, and will misbehave otherwise. So the loop should use simple `index++`, which would eliminate such assumption.

src/restic_fmt_plug.c
```diff
@@ -0,0 +1,334 @@
...
+		const unsigned char *ciphertext = cur_salt->data + NONCE_SIZE;
+		unsigned char prepared_key[32];
+		unsigned char key_mask[16] = {0xff, 0xff, 0xff, 0x0f, 0xfc, 0xff, 0xff, 0x0f,
+					      0xfc, 0xff, 0xff, 0x0f, 0xfc, 0xff, 0xff, 0x0f};
+		unsigned char masked_key[32];
+		int i;
+		memcpy(masked_key, poly1305_key, 32);
+		for (i = 0; i < 16; i++) {
+			masked_key[i + 16] &= key_mask[i];
+		}
```
solardiz: It should be possible to simplify and optimize this to two `uint64_t` ANDs (with corresponding changes to types and thus alignment of these variables), not needing a loop.

src/restic_fmt_plug.c
```diff
@@ -0,0 +1,334 @@
...
+     "4fba163635ffb4e0633de668827e567e8c834539ae18d750be6f8f86f12101b04ab926fa570038eb6f78ef6021c1b1",
+     "penance"},
+    {"$restic$scrypt*32768*8*5*"
+     "e6dac36997999525fdeeb075434416047dab7d28b0c26f82fa44a469747659ecb86614e66b75151e0bd8acc2f8c7b29d258bf8e6a39d47455"
+     "bd8d34e87622510*"
+     "f82d75db1dc5087fb2bdb014bb51cf07b72b62d10cb164fb244b0358b9c1bbac3663f0ca3f00f12731dc3cc3b6c9de949dc74b31d22e515e5"
+     "5096d219d75babc798a9a7a47fc86034e6fe358a93a6b1eff71d2b791ecee3b775b2b910ecf55abef7d53e4a0fac08e5526f1c25eab06dcd2"
+     "6a1250cb37b7dc115391c680becce8a5e63bdbe2326626c88d16bece0d6dca51de89e43207ea6febc6e119486024a0",
+     "password"},
+    {NULL}};
```
solardiz: We typically keep the test vectors near the start of a format source file, before the first function. And the formatting here is very different from what we use.

src/restic_fmt_plug.c
```diff
@@ -0,0 +1,334 @@
...
+			       NULL,
+			       restic_set_salt,
+			       restic_set_key,
+			       restic_get_key,
+			       fmt_default_clear_keys,
+			       restic_crypt_all,
+			       {fmt_default_get_hash},
+			       restic_cmp_all,
+			       restic_cmp_one,
+			       restic_cmp_exact}};
```
solardiz: I understand this entire source file is somehow using a coding style somewhat different from what we generally use in this project, but the above struct initializer is formatted in a totally weird way. Can that be changed to match how we have it in other formats?

solardiz (changes requested):
Thank you for the contribution, this is much appreciated! I've skimmed and made some comments - please address them before we merge. Thanks again.

solardiz:
> > Could you please provide some samples (e.g., so we can test the PR)? Probably to https://github.com/openwall/john-samples.
> 
> Of course: [Add restic repository (john #4531) ](https://github.com/openwall/john-samples/pull/3)

Thanks! Looking at those key files, I think we could even have the JtR format read them directly, by having a `prepare()` method that would extract the right fields. I'm not sure we should.

src/restic_fmt_plug.c
juergenhoetzel: Sorry this was just leftover from `ethereum_fmt_plug.c` (I used as boilerplate).

I have removed the unnecessary includes.

src/restic_fmt_plug.c
juergenhoetzel: > If we skip most of the data here, then maybe we shouldn't have that
> skipped data in the "hashes" in the first place (revise the Python script to skip it)?

All 160 bytes from data are used:

- First 16 bytes: Poly1305-AES Nonce
- Bytes 16-144: The encrypted Master Key we calculate the MAC on
- Last 16 bytes: The MAC we need to match

I decided to leave the DATA in this untyped format because this is
also the way it is stored in the `restic`-JSON file and `restic` might add
additional KDFs/MACs in the future and use DATA differently.


>  Also, I don't see how what the code does agrees
> with the comment. Doesn't the code skip most of data leaving only the
> last 16 bytes, and trying to match that? Yet the comment says "will
> not match the last 16 bytes". Confusing, and probably wrong.

Sorry, I copied only a fraction of the design document (without the
context: "If the password is incorrect or the key file has been
tampered with").

src/restic_fmt_plug.c
juergenhoetzel: Indeed, that was also just a boilerplate copy.

src/restic_fmt_plug.c
juergenhoetzel: thank you again for your explanations: in the beginning not all parameters
were clear to me. Also this part I had just taken over without greater
thinking from `ethereum_fmt_plug.c`.

src/restic_fmt_plug.c
juergenhoetzel: :+1:  Indeed: Done!

Initial code was just a simple/naive port of the corresponding GOLANG code.

src/restic_fmt_plug.c
juergenhoetzel: Yes, in fact I used an arbitrary `clang-format setting`, because I overlooked your style guide. I reformatted using the recommended setting:

```
astyle --style=kr -t4 -U -H -p -xC79 -c -k3 -z2
```

juergenhoetzel:
> Thanks! Looking at those key files, I think we could even have the JtR format read them directly, by having a `prepare()` method that would extract the right fields.

Is there JSON support code in John? I would like to avoid doing this by hand in C.

> I'm not sure we should.

I personally would prefer to keep the python code. It has zero-3rd party dependencies.

:+1:  Thanks again for the helpful comments and tips. This has helped me a lot to understand John's internals better and  contribute further to john in the future.

ghost (approved):
There are some `git am` warnings remaining. But, it was tested and words. 

```
/home/claudio/bleeding/.git/rebase-apply/patch:553: new blank line at EOF.
```

solardiz:
> > Thanks! Looking at those key files, I think we could even have the JtR format read them directly, by having a `prepare()` method that would extract the right fields.
> 
> Is there JSON support code in John? I would like to avoid doing this by hand in C.

No, and even if there were, `prepare()` is defined to work on colon-delimited fields. However, technically it could work anyway: there are plenty of colons in JSON, so it could find and extract what it needs in there. Yes, that would be a hack. Probably we shouldn't go for it. I was just thinking out loud.

> I personally would prefer to keep the python code. It has zero-3rd party dependencies.

OK. BTW, as an option (again, I'm not sure if we should) you could preserve the base64 encoding and have the JtR format rather than the Python script decode it. There is base64 decode code in John. The re-encoding to hex would have made more sense if only a part of the data was needed, but since you wrote that all 160 bytes are needed, keeping the base64 intact might be more reasonable.

> +1 Thanks again for the helpful comments and tips. This has helped me a lot to understand John's internals better and contribute further to john in the future.

Thank you. We'd be happy to have more contributions from you in the future.

src/restic_fmt_plug.c
solardiz: We should probably look into cleaning up `ethereum_fmt_plug.c` as well, although a few things you had inherited from there and now dropped from here were in fact needed there (I think SIMD alignment and going over groups of `MIN_KEYS_PER_CRYPT` keys, as they were for older Ethereum "hashes" processed by a SIMD implementation of PBKDF2).

src/restic_fmt_plug.c
```diff
@@ -0,0 +1,340 @@
...
+#include "formats.h"
+#include "misc.h"
+#include "poly1305-donna/poly1305-donna.h"
+#include "yescrypt/sysendian.h"
+#include "yescrypt/yescrypt.h"
+
+#define FORMAT_NAME "Restic Repository"
+#define FORMAT_LABEL "restic"
+#define FORMAT_TAG "$restic$"
+#define TAG_LENGTH sizeof(FORMAT_TAG) - 1
```
solardiz: Not specific to this format, but ideally we'd have braces around the `sizeof(FORMAT_TAG) - 1`.

src/restic_fmt_plug.c
```diff
@@ -0,0 +1,340 @@
...
+static int max_threads;
+static yescrypt_local_t *local;
+static char (*saved_key)[PLAINTEXT_LENGTH + 1];
+static uint32_t (*crypt_out)[BINARY_SIZE / sizeof(uint32_t)];
+static uint64_t poly1305_key_mask[2];
+
+typedef struct restic_salt {
+	int N;
+	int r;
+	int p;
```
solardiz: Now that you use `sizeof(uint32_t)` for `SALT_ALIGN`, I guess it's best to also use `uint32_t` for the salt struct fields. The `int` were because of `atoi()`, but our `isdec()` ensures a range that fits in `uint32_t` as well even if `int` were larger on some weird platform (in fact, a couple of platforms with 64-bit `int` existed historically).

src/restic_fmt_plug.c
```diff
@@ -0,0 +1,340 @@
...
+void restic_set_salt(void *salt)
+{
+	cur_salt = (restic_salt *)salt;
+}
+
+void restic_set_key(char *key, int index)
+{
+	strnzcpy(saved_key[index], key, sizeof(*saved_key));
+}
+char *restic_get_key(int index)
```
solardiz, line 213: Missing empty line here.

solardiz (approved):
I made some nitpick comments, but this is good enough to merge now.

solardiz:
@juergenhoetzel Should we merge this already or do you intend to make any more changes?

juergenhoetzel:
> @juergenhoetzel Should we merge this already or do you intend to make any more changes?

thanks again for the comprehensive review: I just amended the commit.

PR #4535, opened 2021-01-05
Restic format corrections and minor refactoring

solardiz:
Working OpenMP support.
Benchmark for just one combination of tunable costs.
ALGORITHM_NAME string similar to scrypt format's.
Make format methods static, drop restic_ prefix.
Avoid memory leak in get_salt().
Avoid memcpy() in crypt_all().

### DIFF
--- a/src/restic_fmt_plug.c
+++ b/src/restic_fmt_plug.c
@@ -25,31 +25,41 @@ john_register_one(&fmt_restic);
 #include "formats.h"
 #include "misc.h"
 #include "poly1305-donna/poly1305-donna.h"
-#include "yescrypt/sysendian.h"
 #include "yescrypt/yescrypt.h"
 
 #define FORMAT_NAME "Restic Repository"
 #define FORMAT_LABEL "restic"
 #define FORMAT_TAG "$restic$"
 #define TAG_LENGTH (sizeof(FORMAT_TAG)-1)
-#define ALGORITHM_NAME "PBKDF2/scrypt Poly1305"
+/* SCRYPT_ALGORITHM_NAME is based on scrypt_fmt.c */
+#if !defined(JOHN_NO_SIMD) && defined(__XOP__)
+#define SCRYPT_ALGORITHM_NAME "Salsa20/8 128/128 XOP"
+#elif !defined(JOHN_NO_SIMD) && defined(__AVX__)
+#define SCRYPT_ALGORITHM_NAME "Salsa20/8 128/128 AVX"
+#elif !defined(JOHN_NO_SIMD) && defined(__SSE2__)
+#define SCRYPT_ALGORITHM_NAME "Salsa20/8 128/128 SSE2"
+#else
+#define SCRYPT_ALGORITHM_NAME "Salsa20/8 32/" ARCH_BITS_STR
+#endif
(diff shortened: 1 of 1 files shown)


solardiz:
@juergenhoetzel Please take a look. Most notably, this makes OpenMP support actually work - previously, the thread count was never set to anything other than 1.

solardiz:
Forgot to drop `#include "yescrypt/sysendian.h"`, now done in a force-push.

solardiz:
Added "Make format methods static, drop restic_ prefix."
The Ethereum format code that the Restic format was copy-pasted from was different: it shared stuff between its CPU and OpenCL formats. We don't need that here.

solardiz:
All CircleCI tests except for the pending "more-tests" are successful. I've also tested on a POWER7 big-endian system. I'll merge.

juergenhoetzel:
> @juergenhoetzel Please take a look. Most notably, this makes OpenMP support actually work - previously, the thread count was never set to anything other than 1.

Thanks a lot again. Actually I can't explain how it worked without the OpenMP fix. It did use all 8 Threads on my machine, but the code did only allocate 1 thread local:

```
Will run 8 OpenMP threads
```
i should have used `valgrind`  in the first place.

juergenhoetzel:
```C
/* SCRYPT_ALGORITHM_NAME is based on scrypt_fmt.c */
#if !defined(JOHN_NO_SIMD) && defined(__XOP__)
#define SCRYPT_ALGORITHM_NAME "Salsa20/8 128/128 XOP"
#elif !defined(JOHN_NO_SIMD) && defined(__AVX__)
#define SCRYPT_ALGORITHM_NAME "Salsa20/8 128/128 AVX"
#elif !defined(JOHN_NO_SIMD) && defined(__SSE2__)
#define SCRYPT_ALGORITHM_NAME "Salsa20/8 128/128 SSE2"
#else
#define SCRYPT_ALGORITHM_NAME "Salsa20/8 32/" ARCH_BITS_STR
#endif
```
I wonder what the connection is with Salsa20: Copy/Paste error?

solardiz:
> It did use all 8 Threads on my machine

No, I think it did not, even though it said it did. Without the `omp_autotune(self, OMP_SCALE);` call, `self->max_keys_per_crypt` was left at its initial value, which was 1. So the loop in `crypt_all()` only had 1 key to test, and its use of OpenMP was dummy - not an error, just not actually using more than 1 thread.

That said, in case it actually tries to use more threads while sharing a buffer due to some other bug, we'd possibly need more than one test vector with the same salt yet different password to detect that. Can you get one of those generated and add it and try to trigger the issue? Tricky. In fact, you might want to try triggering this kind of issue first, to see if it in fact remains undetected without more test vectors. Thanks!

Unrelated: another thing I noticed is we don't make any use of the first half of KDF output. This is an opportunity to save two SHA-256 computations inside PBKDF2. But it's tricky to do - would involve hacking shared yescrypt code - which is probably not worth it for such minor savings.

solardiz:
> I wonder what the connection is with Salsa20?

scrypt spends most processing time on Salsa20/8.

juergenhoetzel:
> > I wonder what the connection is with Salsa20?
> 
> scrypt spends most processing time on Salsa20/8.

I see: cannot say thank you often enough.

PR #4538, opened 2021-01-05
Avoid reporting potentially stale errno in scrypt-using formats

solardiz:
With these changes, we assume that writes to an `int` are atomic, which OpenMP doesn't provide a directive for (other than `#pragma omp critical`, which is overkill). We already make this assumption elsewhere in the tree, where it's performance-critical.


solardiz:
The code is still not nice in that when scrypt computation fails, it might proceed to use the possibly-uninitialized scrypt output buffer for further computation. Maybe we should clean up this further. This is currently mitigated by the eventual reporting of failure and process exit, before the results of that further computation would be made use of.

PR #4540, opened 2021-01-06
Travis CI: do run Linux tests on pull requests

solardiz:
Not even skip macOS builds - I want to see how many credits they consume, then decide. Apparently, even a dummy build was consuming 100 credits, so maybe a real build is the same? We'll see.


solardiz:
Meanwhile, a full macOS build+test (including OpenCL) takes over 8 minutes, or 900 credits. I guess we shouldn't be doing that on every force-push to every PR. I'll amend this PR.

Issue #4529, opened 2020-12-28
Some internal mask formats b0rken


magnumripper:
Problem seen with oldoffice-opencl. That format is based on some of Sayantan's formats (can't remember which right now) so it's very likely that template format and probably more formats (with internal mask) have the same problem. NT-opencl does not have this problem.

Bug introduced with 8bbd90f49. After that commit, kernel isn't rebuilt for the mask in use (the self-test kernel is re-used), so any non-static mask will result in false negatives. For example, `-mask:abc?d` will produce one single candidate, a literal `abc#`.

Temporary fix:
```diff
diff --git a/src/opencl_oldoffice_fmt_plug.c b/src/opencl_oldoffice_fmt_plug.c
index d003402c9..d9dd5940a 100644
--- a/src/opencl_oldoffice_fmt_plug.c
+++ b/src/opencl_oldoffice_fmt_plug.c
@@ -246,7 +246,7 @@ static void reset(struct db_main *db)
 	int i;
 
 	if (crypt_kernel)
-		release_clobj();
+		done();
 
 	for (i = 0; i < MASK_FMT_INT_PLHDR; i++)
 		if (mask_skip_ranges && mask_skip_ranges[i] != -1)
```
A better fix might be possible although the above is probably fine.

I'm considering starting with adding a mask-mode test to the Test Suite's "internal" tests and go from there.

magnumripper:
Just from looking at 8bbd90f, it looks like oldoffice might be the only format affected

ghost:
> I'm considering starting with adding a mask-mode test to the Test Suite's "internal" tests and go from there.

I have `mask-mode` tests, but they were designed only for "my formats" (you need a test case, you have to tune to the format and device, etc). Like `-mask=[Pp][Aa@][Ss5][Ss5][Ww][Oo0][Rr][Dd]`, `-mask:admin?d -min-len=0 -max-len=10` or  `-w:word.lst -mask:?w?d?d?d`. I found them useful.

JtR needs more automated testing.

magnumripper:
The TS "internal" tests put all test vectors in a file, and their passwords in a wordlist. I figure I could just `s/.$//` that wordlist to a second wordlist which will then lack the last character. This can then be used with `-w -mask:?w?a`. It wont test every aspect of mask mode but it would detect this problem.

magnumripper:
Oldoffice is fixed, keeping issue open for systematic testing

magnumripper:
[...]

PR #4547, opened 2021-01-12
OldOffice OpenCL: Ensure proper kernel rebuild in case mask changed.

magnumripper:
See #4529

I'm leaving the issue open, for closer review of all mask formats and adding some TS test for detecting problems like this one.


ghost (approved):

solardiz (approved):
I have no idea if this is right, but it also doesn't look wrong. So I trust magnum's analysis and testing.

PR #4548, opened 2021-01-13
poly1305-donna: Prevent name clashing with existing implementation

### LINKED ISSUE #4546: multiple definition of `poly1305_init'
### Steps to reproduce ###
Compile with libmusl static. I have a Dockerfile to do so here:

https://github.com/bannsec/hashcrack-jtr/blob/master/Dockerfile.linux

Looks like @solardiz changed a bunch of stuff in poly1305-donna/poly1305-donna-64.h on Jan 7 2021. Up to that point, my building Docker worked fine. However, since those changes, I'm getting an error when compiling and am unsure how to fix it.

Error:
```
/opt/cross/bin/../lib/gcc/x86_64-linux-musl/9.2.0/../../../../x86_64-linux-musl/bin/ld: poly1305-donna/poly1305-donna.a(poly1305-donna.o): in function `poly1305_init':
/opt/JohnTheRipper/src/poly1305-donna/poly1305-donna-64.h:76: multiple definition of `poly1305_init'; /opt/cross/lib//libcrypto.a(poly1305-x86_64.o):(.text+0x0): first defined here
```

The commit that this started happening on is 4e924f2.

Not sure what to do at this point to successfully compile again.

### System configuration ###
Latest from master

juergenhoetzel:
Closes #4546

### DIFF
--- a/src/poly1305-donna/poly1305-donna.h
+++ b/src/poly1305-donna/poly1305-donna.h
@@ -3,6 +3,14 @@
 
 #include <stddef.h>
 
+/* prevent name clashing with existing implementations */
+#define poly1305_init john_poly1305_init
+#define poly1305_update john_poly1305_update
+#define poly1305_finish john_poly1305_finish
+#define poly1305_auth john_poly1305_auth
+#define poly1305_verify john_poly1305_verify
+#define poly1305_power_on_self_test john_poly1305_power_on_self_test
+
 typedef struct poly1305_context {
 	size_t aligner;
 	unsigned char opaque[136];


solardiz (commented):
This looks almost right, but it can be improved:

1. I'd list the functions in the same order they're listed in the rest of this same header file.
2. I'd exclude `poly1305_blocks` because it is `static` and thus can't clash at link time. There are many other `static` symbols internal to this library, which we don't bother treating similarly anyway.

juergenhoetzel:
> This looks almost right, but it can be improved:

Thanks: I amended the commit.

PR #4549, opened 2021-01-15
Review of prefix functions

### LINKED ISSUE #4533: human_prefix_small() precision loss
As mentioned in https://github.com/openwall/john/pull/4524#discussion_r546287925:

This got me looking inside `human_prefix_small()`, and it's really weird. It might bring num to just above 1.0, then round it to integer 1 - that's very little precision left. We need to fix it.

magnumripper:
Replace human_prefix_small() with the better one that was in
opencl_common.c.
Move human_speed() from opencl_common.c to misc.c.
Use these functions from omp_autotune.c.

Closes #4533

### DIFF
--- a/src/misc.c
+++ b/src/misc.c
@@ -567,9 +567,8 @@ char *human_prefix(uint64_t num)
 	char prefixes[] = "\0KMGTPEZY";
 	char *p = prefixes;
 
-	while (p[1] && (num >= (100 << 10) ||
-	                (!(num & 1023) && num >= (1 << 10)))) {
-		num = (num >> 10);
+	while (p[1] && (num >= (100 << 10) || (!(num & 1023) && num >= (1 << 10)))) {
+		num >>= 10;
 		p++;
 	}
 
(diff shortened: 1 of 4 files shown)


src/misc.c
```diff
@@ -581,21 +580,55 @@ char *human_prefix(uint64_t num)
...
+	} else if (milli) {
+		if (micro)
+			snprintf(out, 16, "%d.%03d m", milli, micro);
+		else
+			snprintf(out, 16, "%d m", milli);
+	} else if (micro) {
+		if (nano)
+			snprintf(out, 16, "%d.%03d u", micro, nano);
+		else
+			snprintf(out, 16, "%d u", micro);
```
solardiz, line 629: I realize this code is moved from elsewhere in the tree, but I think  we should drop the 3 if/else for when milli/micro/nano happen to be 0. We should in those cases just print .000 (which would happen on its own), which serves to indicate precision, just like we print e.g. .100 without omitting the two zeroes. No reason for the special cases.

src/misc.c
```diff
@@ -581,21 +580,55 @@ char *human_prefix(uint64_t num)
 	return out;
 }
 
-char *human_prefix_small(double num)
+char *human_speed(uint64_t speed)
```
solardiz, line 583: I'd rename `human_prefix` to e.g. `human_prefix_binary`, and then `human_speed` to `human_prefix_decimal` (or `human_prefix_large`?) and omit the " c/s" from it (to be added by the callers). I guess we might need it in places where the speed isn't in c/s, but is e.g. in p/s. (We can't simply use this function for our status line, though, at least because the memory leak would unacceptable there. So let's not do that just yet.)

src/misc.h
```diff
@@ -190,16 +190,23 @@ const char *jtr_lltoa(int64_t num, char *result, int result_len, int base);
...
 
 /*
- * Change some tiny number to a string possibly using SI prefix
- * eg. 0.123 -> "123 m"
- * Note: "leaks" 16 bytes each time (mem_alloc_tiny)
+ * From a potentially large number produce a cps string possibly using
+ * prefix, eg. 437281954 -> "437281K c/s".
+ * Note: "leaks" 16 bytes each time (mem_alloc_tiny).
+ */
+extern char *human_speed(uint64_t speed);
```
solardiz, line 204: With my suggested changes to function names and non-inclusion of " c/s", the comments here would need to be updated accordingly.

src/misc.c
```diff
@@ -581,21 +580,55 @@ char *human_prefix(uint64_t num)
...
-		snprintf(out, 16, "%u ", (uint32_t)num);
+		snprintf(out, 16, "%u c/s", (uint32_t)speed);
+
+	return out;
+}
+
+char *human_prefix_small(double num)
+{
+	char *out = mem_alloc_tiny(16, MEM_ALIGN_NONE);
+	uint64_t number = num * 1E9;
```
solardiz, line 605: Maybe the function should instead accept `uint64_t nano` right away.

solardiz (approved):
I think these changes are in the right direction and acceptable for merging as-is, but I've suggested some improvements (or at least I think they are). Please feel free to merge as-is or to implement those improvements and let me re-review first. Thanks!

src/misc.c
magnumripper: Yes... but it would have been more relevant before we moved it to a shared function: It was only used for OpenCL auto-tune before and the rationale was to save precious screen estate but in practice they always end up with three decimals anyway so it was moot. Better would have been to just drop a decimal or two (and have it proper rounded as opposed to truncated).

Now though, since we're using this function for `--list=build-info` I think that would mean we'd print resolution/latency figures as eg `1.000 us` instead of `1 us` and that would be misleading, wouldn't it? At the very least we'd still have to print `28 ns` (my actual latency for the hi-res timer) and definitely not `28.000 ns`, right?

solardiz: On one hand, you're right, `28.000 ns` would be weird for anything, and `1.000 us` would be weird when we had just `1` because of low precision elsewhere. On the other hand, the current code could also print e.g. `1.100 us` when those `00` do not necessarily reflect precision of the input to this function, so making just the case of `000` special is inconsistent.

As to the `28.000 ns` example again, the current code simply never prints decimals of `nano`, so this is already impossible, and that's fine.

So I once again suggest that we simply drop the special cases for omitting `000`. If you like to omit that when we _know_ the input had lower precision, that should be passed into the function out-of-band (or a separate function used), not inferred (sometimes wrongly) from the value being printed.

magnumripper: It's just that (exaggerating a little) I'd rather always *drop* the decimals, but then we're back with the original problems of #4533 (lol). The original functions' goals wasn't precision, it was the opposite: Omit irrelevant precision and print a "simple number" for a human (hence the function names). The problem is obviously how to know just what precision is relevant or not and this gets harder the more we try to make them shared functions for use in varying context.

Anyways I'll probably end up doing just what you said :-)

magnumripper:
After lots of experimenting I decided to just merge it as-is.

Issue #4550, opened 2021-01-15
Import mosquitto_passwd hashes


solardiz:
@Blackfell wrote a Python script that converts `mosquitto_passwd` files to formats understood by John and hashcat. (The script can also crack hashes on its own, slower.) https://github.com/Blackfell/python_widgets/blob/main/mosquito_passwd_crack.py https://mosquitto.org/man/mosquitto_passwd-1.html

@Blackfell Thank you for writing and publishing this script! If you don't mind, I'd like to see it included in jumbo. Would you contribute this to us via a PR, please? I imagine the PR would include your Python script with a copyright statement and license comment added at top (similar to what we have in Dhiru's scripts), `doc/README-mosquitto` with content from your README's `mosquitto_passwd_crack` section, and an update to `doc/NEWS`.

Alternatively or additionally, we could enhance the JtR format to be able to parse those files/hashes directly. For the alias to `dynamic_82`, this would probably be a task for @jfoug. For the alias in `PBKDF2-HMAC-SHA512`, it's probably something many of us could do. But without clear plans on this, I think we should get the Python script in first.

Nitpick: BTW, I think it's `mosquitto` with two t's. `mosquito_passwd_crack.py` has that wrong in its name and in a couple of places instead.

Blackfell:
Hey, I'd be more than happy to do this, I'll try and get something in over the next few days. As you said, I'll have a look at the existing scripts and neaten everything up first (good spot on typos, I like nitpicks so keep em coming 😊). Also keen to support any of the other changes, I'll keep it in mind as I read up.

Blackfell:
I've popped a [PR](https://github.com/openwall/john/pull/4551) in with a bit of a re-hashed script. The main outstanding issue I'm having is that format auto-detection for dynamic_82 isn't working well on a Kali VM I've been using to test (surprise!). I've put more detail in the PR comments.

Note also that the original script has been renamed and (hopefully) de-typod - [https://github.com/Blackfell/python_widgets/blob/main/mosquito_passwd_crack.py](https://github.com/Blackfell/python_widgets/blob/main/mosquitto_passwd_crack.py)

solardiz:
Fixed in #4551. Closing.

PR #4551, opened 2021-01-18
Eclipse Mosquitto - mosquitto_passwd hash converter - mosquitto2john.py

Blackfell:
Helper script to format Eclipse mosquitto_passwd files for John as mentioned in [Issue 4550](https://github.com/openwall/john/issues/4550).

Changes Comprise addition of run/mosquitto2john.py and doc/README-mosquitto files, along with a small addition to doc/NEWS. The script itself is a cut-down version of the version listed in the original issue (now renamed to [https://github.com/Blackfell/python_widgets/blob/main/mosquito_passwd_crack.py](https://github.com/Blackfell/python_widgets/blob/main/mosquito_passwd_crack.py)). This version is wholly focused on converting to John compatible format and manages stdout and stderr in a way that mimics some of the other *2john.py tools (because it's better!). 

More than happy to take pointers and fixes, in particular, I've tried to imitate the entries in doc/NEWS, let me know if that looks about right or not and I'll happily amend. Similarly, the README is markdown still, happy to re-format as you prefer. 

Tested (mostly) against Mosquitto version 2.0.5 install of mosquitto_passwd, using Docker, ubuntu:latest to create test passwd files. Tested outputs against three John installs on Arch, Ubuntu and Kali. Mileage for auto-detection of dynamic_82 format seems to vary per-install of John (i.e. works on everything I've tested on except Kali), but can always crack hashes if the format is specified on the command line (some info in README to reflect this). If there's a way my script can help improve format auto-detection, I'll be glad to add, I just can't see what could be causing it, beyond a local build issue. 

Cheers,
 - Blackfell

### DIFF
--- a/doc/NEWS
+++ b/doc/NEWS
@@ -157,6 +157,9 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
 - Drop support for --fix-state-delay option, and instead do something better
   in the cracker engine - by default, and for all modes.  [magnum; 2020]
 
+- Addition of mosquitto2john.py for cracking of Eclipse Mosquitto passwd files.
+  [blackfell; 2021]
+
 
 Major changes from 1.8.0-jumbo-1 (December 2014) to 1.9.0-jumbo-1 (May 2019):
 
@@ -560,3 +563,4 @@ Major changes from 1.8.0-jumbo-1 (December 2014) to 1.9.0-jumbo-1 (May 2019):
   mostly part of our development process and not the release, although some
   CI-related files do exist in our released tree.
   [Claudio, magnum; 2015-2019]
+
--- a/doc/README-mosquitto.md
+++ b/doc/README-mosquitto.md
@@ -1,0 +1,144 @@
+# Eclipse mosquitto_password cracking
+
+mosquitto2john.py is a helper script to allow you to convert Eclipse Mosquitto
+passwd files into a John friendly format. The script takes a 
+[mosquitto_passwd file](https://mosquitto.org/man/mosquitto_passwd-1.html) as 
+its only required input. 
+
+## Usage
+
+```
+❯ mosquitto2john.py --help
+usage: mosquitto2john.py [-h] [-hc] [passwd_file ...]
+
+positional arguments:
+  passwd_file     Path to the source mosquitto_passwd file(s).
+
+optional arguments:
+  -h, --help      show this help message and exit
+  -hc, --hashcat  Convert hashes to hashcat friendly formats.
+
+Find more Information:
+    See doc/README-mosquitto.md for info/troubleshooting.
+```
+
+This script is intended to work in the classic something2john.py way, provide
+mosquitto_passwd files and pipe the output into an outfile of your choosing:
+
+```
+❯ mosquitto2john.py mosquitto_passwd mosquitto_passwd_2 > mosquitto.hash
+❯ john ./mosquitto.hash
+[...SNIP...]
+Session completed. 
+```
+### Hashcat
+
+The hashcat option converts the hashes to Hashcat friendly formats - mode 1710 
+for SHA512 and 12100 for pbkdf2-hmac-sha512 hashes. In the following example, 
+some SHA512 and pbkdf2-hmac-sha512 hashes have already been exported into 
+separate files according to their type:
+
+```
+❯ ls
+sha512_hashes     HMAC_hashes
+# Identify the SHA512 hashes by their format: HEX_DIGEST:HEX_SALT
+❯ head -n 1 sha512_hashes
+SOMELONGHEXDIGEST:SOMEHEXSALT
+❯ hashcat -a 0 -m 1710 --hex-salt ./sha512_hashes [WORDLIST]
+[... SNIP ...]
+# Identify the HMAC hashes by a leading sha512: and following iteration count
+# These also have base64 salt and digests
+❯ head -n 1 HMAC_hashes
+sha512:101:BaSe64SaLt:BaSe64DiGeSt==
+❯ hashcat -a 0 -m 12100 ./HMAC_hashes [WORDLIST]
+[... SNIP ...]
+```
+
+**Note** the *use of --hex-salt in mode 1710* this is required for successful
+cracking.
+
+## Troubleshooting
+
+### No hashes loaded?
+
+In certain older versions of JtR and only for the SHA512 format hashes 
+(dynamic_82), you may have problems autodetecting the hash format. You
+can manually specify the format with *--format=dynamic_82 in this instance*:
+
+```
+# Peek at the format to spot the issue -  just after the username field
+❯ head -n 1 ./mosquitto.hash
+username:$dynamic_82$SOMELONGHASH$SOMESALT	
+# Now manually specify the dynamic_82 format
+❯ john ./mosquitto.hash --format=dynamic_82
+[...SNIP...]
+Session completed. 
+```
+
+### Mixed Hash Types?
+
+The Eclipse mosquitto_passwd program might allow different hash varieties, 
+depending on the Version you're working against. In the rare case that this
+could come up, the output can either be split into two (using, say grep), or
+you can simply specify the hash format on the command line. In the following
+example, hashes are stored in a single file and each format worked against
+sequentially:
+
+```
+❯ ls
+mosquitto_out
+# Now crack one at a time because different hash modes
+❯ john ./mosquitto_out --format=dynamic_82 
+[... SNIP ...]
+Session completed.
+❯ john ./mosquitto_out --format=pbkdf2-hmac-sha512
+[... SNIP ...]
+Session completed.
+```
+
+Similarly, if using Hashcat, we need to crack any mixed hash formats one 
+flavour at a time. The following is an equivalent example in Hashcat:
+
+```
+❯ ls
+hcat_out_file
+❯ hashcat -a 0 -m 1710 --hex-salt hcat_out_file [WORDLIST]
+[... SNIP ...]
+Hashfile 'hcat_out_file' on line 1 [...SNIP...]: Token length exception
+[... SNIP ...]
+❯ hashcat -a 0 -m 12100 hcat_out_file [WORDLIST]
+[... SNIP ...]
+Hashfile 'hcat_out_file' on line 2 [...SNIP...]: Token length exception
+[... SNIP ...]
+```
+This time, a Token length exception is raised for the hashes whose format
+doesn't match the current mode. Hashcat will continue to crack the valid 
+hashes just fine, but to avoid this,you may instead choose to split the two 
+formats into separate files. See the usage section for hashcat hash 
+identification.
+
+Again, note the need to **specify --hex-salt** for the 1710 format hashes.
+
+### Invalid input. Try removing ':' from username
+
+You may see this error on (hopefully) rare occasions if the mosquitto_passwd
+file itself  contains an error. Eclipse mention in one line of the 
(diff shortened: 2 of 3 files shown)


run/mosquitto2john.py
```diff
@@ -0,0 +1,153 @@
+#!/usr/bin/env python
+
+# This software is Copyright (c) 2021, Blackfell <github at blackfell.net>
+# and it is hereby released to the general public under the following terms:
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted, provided that the above copyright notice
+# appears in all copies and that both that copyright notice and this
+# permission notice appear in supporting documentation.
```
solardiz: This is very similar to 2-clause BSD, but not the same wording. If you insist on stipulating those conditions, then please use 2-clause BSD verbatim - see it e.g. in `src/yescrypt/sha256.h`.

doc/README-mosquitto
```diff
@@ -0,0 +1,75 @@
...
+# Now crack one at a time because different hash modes
+❯ john ./mosquitto_out.dynamic_82 
+[... SNIP ...]
+❯ john ./mosquitto_out.pbkdf2-hmac-sha512
+[... SNIP ...]
+#You're done!
+```
+
+### Invlid Input. Try removing ':' ?
+the
```
solardiz: Something broke here.

solardiz:
Thank you for contributing this!

> This version is wholly focused on converting to John compatible format and manages stdout and stderr in a way that mimics some of the other *2john.py tools (because it's better!)

That's great. However, I suggest that you reintroduce outputting hashcat-style hashes when run with some non-default command-line option. A lot of people who use hashcat have JtR jumbo nearby for the conversion tools, and this would be handy to them.

> the README is markdown still, happy to re-format as you prefer.

Markdown is fine, but please wrap long lines at 79, so that the file can also be conveniently viewed as plain text in a terminal window? And please add the `.md` suffix.

> If there's a way my script can help improve format auto-detection, I'll be glad to add, I just can't see what could be causing it

I think it's most important to test with our latest code from this repo. What happens with some older build on Kali isn't as important, but it's fine to provide documentation on that as long as those builds are still common. As to what the issue could be, I vaguely recall we had some change that broke the autodetection of dynamic formats, which we later repaired.

I notice your commits aren't currently associated with a GitHub account. You could want to use an e-mail address that you added to your GitHub account, or to add whatever address you used on those commits.

Also, for merging this we'll need just one squashed commit or if you like you can tell us to squash your commits on merging (then GitHub will appear as the committer of the squashed commit, but you still as the author). We normally use force-pushes instead of additional commits for updates while working on a PR. Separate commits are for logical separation that should remain in the project's history, not for minor updates within a PR.

solardiz:
> I vaguely recall we had some change that broke the autodetection of dynamic formats, which we later repaired.

Found it, issue #4002.

doc/NEWS
```diff
@@ -560,3 +560,6 @@ Major changes from 1.8.0-jumbo-1 (December 2014) to 1.9.0-jumbo-1 (May 2019):
   mostly part of our development process and not the release, although some
   CI-related files do exist in our released tree.
   [Claudio, magnum; 2015-2019]
+
+- Addition of mosquitto2john.py for cracking of Eclipse Mosquitto passwd files.
+  [Blackfell; 2021]
```
solardiz: The `NEWS` addition should be to the proper section of that file, not to the end of file. The way it currently is makes the change appear to be part of 1.9.0-jumbo-1.

run/mosquitto2john.py
solardiz: Or maybe you did copy this entire license text from somewhere? If so, where? To me, it looks like hybrid of our preferred 0-clause BSD (the usual BSD license cut after the word "permitted") and MIT license.

Blackfell: No problem, you're right on that, I think this was incorrectly pulled from 1password2john when I was looking over how the other scripts are put together, I'll use the BSD wording I see in 1password2john if you're content? Reads as follows:

# Permission to use, copy, modify, and/or distribute this software for any
# purpose with or without fee is hereby granted, provided that the above
# copyright notice and this permission notice appear in all copies.

doc/README-mosquitto
Blackfell: Agreed, thanks I'll zap that one.

doc/NEWS
Blackfell: Understood, I'll get to grips with the News file and sort that out, thanks!

Blackfell:
Thanks for the comments, I'll address throughout and sort out the Git stuff mentioned above, thanks for your help with that in particular, I'm fairly new to contribution so appreciated :) 

With the issue on autodetection, if I switched the format in dynamic_detection, worked just fine, so you're spot on with that. 

I'll be happy to add hashcat stuff back in too, good point!

run/mosquitto2john.py
solardiz: That's ISC license, which for that source file we inherited from somewhere else. If you insist, we can accept that indeed (and not being a hybrid it's better than what you had before), but our preferred license wording is found e.g. in `restic2john.py` (lacks the restrictions).

Blackfell: Got it, yeah I'm happy with that license and will go with that.

Blackfell:
OK I hope that commit catches all the comments, I think I ran the force push correctly to give you a single commit now, so let me know if that's cool or if you'd like any more git stuff amending, happy to do so. 

Hashcat functionality has been added again and everything cleaned up and hopefully better quality now. Let me know if you'd like any test hashes to mess around with yourself, happy to attach/send a couple small examples.

solardiz:
> Let me know if you'd like any test hashes to mess around with yourself, happy to attach/send a couple small examples.

Please send a PR with sample `mosquitto_passwd` files (not already extracted hashes) against the john-samples repo here. Thanks!

doc/README-mosquitto.md
```diff
@@ -0,0 +1,151 @@
+# Ecplipse mosquitto_password cracking
```
solardiz: Typo in `Ecplipse`.

doc/README-mosquitto.md
```diff
@@ -0,0 +1,151 @@
...
+
+positional arguments:
+  passwd_file           Path to the target mosquitto_passwd file.
+
+optional arguments:
+  -h, --help            show this help message and exit
+  -hc, --hashcat        Convert hashes to hashcat friendly formats.
+  -o OUT_FILE, --out-file OUT_FILE
+                        File name for converted hashes - required only for 
+			mixed hashes.
```
solardiz: Inconsistent indentation (maybe tabs vs. spaces on the two lines).

doc/README-mosquitto.md
```diff
@@ -0,0 +1,151 @@
...
+**Note** the *use of --hex-salt in mode 1710* this is required for successful
+cracking.
+
+## Troubleshooting
+
+### No hashes loaded?
+
+If your individual install isn't automatically detecting the hash type, you 
+can try manually checking the hash format and specifying it with *--format*
+on the John command line:
```
solardiz: This should only be needed for "certain older" (you can use this wording) versions of JtR and only for "dynamic_82". You don't need to provide an example for "pbkdf2-hmac-sha512" because those are always auto-detected. So you can simplify this section omitting the instructions for how to extract the hash type and suggesting specifically the use of `--format=dynamic_82` as a workaround.

doc/README-mosquitto.md
```diff
@@ -0,0 +1,151 @@
...
+
+### Mixed Hash Types?
+
+The Eclipse mosquitto_passwd program might allow different hash varieties, 
+depending on the Version you're working against. In the rare case that this
+could come up, you'll see an error and will be prompted to specify an 
+`out_file_name`; the script will then split the hash types out for you to 
+crack independently from one another. Note that the file extensions 
+allude to the hash format, handy if you want to use the --format flag in
+John.
```
solardiz: Actually, John is happy with a shared file, where the `--format` option would be used to choose which of those to load. If it were not for hashcat, I'd have suggested dropping the out-file support.

doc/README-mosquitto.md
```diff
@@ -0,0 +1,151 @@
...
+❯ ls
+mosquitto_out.1710      mosquitto_out.12100
+❯ hashcat -a 0 -m 1710 --hex-salt ./hcat_out_file.1710.hcat [WORDLIST]
+[... SNIP ...]
+❯ hashcat -a 0 -m 12100 ./hcat_out_file.12100.hcat [WORDLIST]
+[... SNIP ...]
+```
+Again, you need to **specify --hex-salt** for the 1710 format hashes.
+
+### Invlid Input. Try removing ':' ?
```
solardiz: Typo in "Invlid". Should be the same as what the script actually says, which is: "Invalid input. Try removing ':' from username".

doc/README-mosquitto.md
```diff
@@ -0,0 +1,151 @@
...
+its only required input. 
+
+## Usage
+
+```
+❯ mosquitto2john.py --help
+usage: mosquitto2john.py [-h] [-hc] [-o OUT_FILE] passwd_file
+
+positional arguments:
+  passwd_file           Path to the target mosquitto_passwd file.
```
solardiz: Perhaps "source" rather than "target"? (Should also fix this in the script.)

Also, ideally the script would be able to handle multiple files or at least would give an error (or does it already?) if multiple files are listed on the command-line.

solardiz (commented):
This is almost right, but here are some more nitpicks.

doc/README-mosquitto.md
Blackfell: Interesting point, I just played around in hashcat and it's happy-ish with a shared file - it throws a token length exception but then continues to work on the 'good' hash. I'm tempted based on what you said, and for simplicitys sake, to remove this functionality outright; I guess in that instance, the only thing helping the user identify this mixed hash case is the README, so I could add some epilogue to the usage message pointing to the JtR docs. Does that sound reasonable?

solardiz: I'm fine dropping the out-file support altogether. My concern about hashcat was that it could perhaps flood with that token length exception when there are many hashes of the other type. John prints a warning suggesting use of the `--format` option just once per another detected format. Anyway, I think this flood issue is very minor - there probably won't be that many accounts in `mosquitto_passwd` files, and it's not hard to use grep to choose just one hash type if needed. So yes, please go ahead and simplify this. Thanks!

Blackfell: Great, I've got this sorted now. 

The only other thing I thought of is printing a warning via stderr to the effect "careful you have mixed hash formats", but I've left it out for now (warnings on Stderr didn't feel quite right), in favour of a pointer in the usage message to the README for now. Both John and Hashcat will give plenty of visual output and I do think this will be a rare issue, so hopefully that will be adequate.

doc/README-mosquitto.md
Blackfell: Good spot, typo in work 1 of the readme - best avoided :')

Blackfell:
Thanks for the comments, the main one I've responded to directly, otherwise I've cleaned up the others throughout and re-squashed.

solardiz (approved):
I'd have many more nitpicks, but I think it'd be discouraging and a waste of effort to continue like that - so I'll merge, and we can revise later.

Just so that we know and have this recorded in here, what version(s) of Python did you test this with?

Thanks!

Blackfell:
Hi, I tested on Python 3.9.1 and 2.7.18.

Thanks for helping me through this, I really appreciate it (incl. the understanding/patience!) and am made up to be able to help a great project in some small way - I'd be happy to do so again in future if I can.

Issue #4300, opened 2020-08-17
Bug in zip2john or pkzip format: False negative


magnumripper:
I've been sent a zip file with password '123456' that our pkzip format fails to crack. I'm not sure I can share the file publically but I sent it to @jfoug and I'll be jotting down notes here.

Partial `zipinfo -v` output
```
  version of encoding software:                   2.0
  minimum file system compatibility required:     MS-DOS, OS/2 or NT FAT
  minimum software version required to extract:   2.0
  compression method:                             deflated
  compression sub-type (deflation):               normal
  file security status:                           encrypted
  extended local header:                          yes
  file last modified on (DOS date/time):          2020 Jul 13 16:45:02
  file last modified on (UT extra field modtime): 2020 Jul 14 01:45:02 local
  file last modified on (UT extra field modtime): 2020 Jul 13 23:45:02 UTC
  32-bit CRC value (hex):                         856d3284
  compressed size:                                14212 bytes
  uncompressed size:                              16909 bytes
```

Partial zip2john output, only the data blob shortened:
```
$pkzip2$1*2*2*0*3784*420d*856d3284*0*64*8*3784*856d*85a1*3d5187246d1612cc (...) 3da334f4de17a560*$/pkzip2$
```

Here's the lines that give false negative:
```c
			/* if the hash is a 2 byte checksum type, then check that value first */
			/* There is no reason to continue if this byte does not check out.  */
			if (salt->chk_bytes == 2 && C != (e&0xFF) && C != (e2&0xFF))
				goto Failed_Bailout;
```

magnumripper:
```
$pkzip2$1*2*2*0*3784*420d*856d3284*0*64*8*3784*856d*85a1*(...DA...)
```
Parsing the above:
* C Hash count: 1
* B Number of valid bytes in checksum: 2
* DT Data type: 2 (inlined full data)
* MT Magic type: 0 (no file magic used)
* CL Compressed length: 0x3784 (14212) bytes
* UL Uncompressed length: 0x420d (16909) bytes
* CR Final CRC32: 856d3284
* OF Offset: 0
* OX Additional offset: 64
* CT Compression type: 8 (imploded)
* DL Length of inlined data: 0x3784 bytes
* CS Checksum from CRC32: 856d (this is the MSB of "Final CRC32" above)
* TC Checksum from timestamp: 85a1

All this look sane to me. Not sure what's wrong here - but if I comment out the mentioned lines in the format, everything is fine (but we lose some performance).

[...]

PR #4561, opened 2021-02-04
zip2john: Default to using single byte checksums

### LINKED ISSUE #4541: zip2john doesn't produce a crackable hash when zip was created with bsdtar
Hi and thanks a lot for JtR!

I think that zip2john doesn't work when the zip file was created with bsdtar:

    bsdtar --options zip:encryption -acf with_bsdtar.zip test_file

Here's the example [`with_bsdtar.zip`](https://github.com/openwall/john/files/5777014/with_bsdtar.zip), the password is "test".

The output of `zip2john with_bsdtar.zip`:

    ver 2.0 efh 5455 efh 7875 with_bsdtar.zip/test_file PKZIP Encr: 2b chk, TS_chk, cmplen=0, decmplen=6, crc=0
    with_bsdtar.zip/test_file:$pkzip2$1*2*2*0*0*6*0*0*47*8*0*0000*8cd0**$/pkzip2$:test_file:with_bsdtar.zip::with_bsdtar.zip

On the other hand, things work fine when using a zip file created with

    zip --encrypt with_zip.zip test_file

where `zip2john with_zip.zip` yields

    ver 1.0 efh 5455 efh 7875 with_zip.zip/test_file PKZIP Encr: 2b chk, TS_chk, cmplen=18, decmplen=6, crc=775F54D8
    with_zip.zip/test_file:$pkzip2$1*2*2*0*12*6*775f54d8*0*43*0*12*775f*8cd0*a5ad914fcad388ac41ecf2a97498c2315730*$/pkzip2$:test_file:with_zip.zip::with_zip.zip

Here's [`with_zip.zip`](https://github.com/openwall/john/files/5777077/with_zip.zip).

The archive `with_bsdtar.zip` seems to be valid, since I can `unzip with_bsdtar.zip` successfully.

### System configuration ###
I'm on Arch Linux 5.9.14.

Output of `john --list=build-info`:

    Version: 1.9.0-jumbo-1
    Build: linux-gnu 64-bit x86_64 AVX AC MPI + OMP
    SIMD: AVX, interleaving: MD4:3 MD5:3 SHA1:1 SHA256:1 SHA512:1
    System-wide exec: /usr/lib/john
    System-wide home: /usr/share/john
    Private home: ~/.john
    CPU tests: AVX
    CPU fallback binary: john-non-avx
    $JOHN is /usr/share/john/
    Format interface version: 14
    Rec file version: REC4
    Charset file version: CHR3
    CHARSET_MIN: 1 (0x01)
    CHARSET_MAX: 255 (0xff)
    CHARSET_LENGTH: 24
    SALT_HASH_SIZE: 1048576
    SINGLE_IDX_MAX: 2147483648
    SINGLE_BUF_MAX: 4294967295
    Max. Markov mode level: 400
    gcc version: 10.2.0
    GNU libc version: 2.32 (loaded: 2.32)
    OpenCL headers version: 2.2
    Crypto library: OpenSSL
    OpenSSL library version: 01010108f	(loaded: 01010109f)
    OpenSSL 1.1.1h  22 Sep 2020	(loaded: OpenSSL 1.1.1i  8 Dec 2020)
    GMP library version: 6.2.0	(loaded: 6.2.1)
    File locking: fcntl()
    fseek(): fseek
    ftell(): ftell
    fopen(): fopen
    memmem(): System's

magnumripper:
This may mean ~10% performance loss but the alternative seems to be risk of false negatives.
The existing -2 option was changed to "allow" instead of "force" 2-byte checksums.

Closes #4541, see also #4300.

### DIFF
--- a/src/zip2john.c
+++ b/src/zip2john.c
@@ -189,7 +189,7 @@ static int fexpect(FILE *stream, int c)
 }
 
 static int checksum_only, use_magic;
-static int force_1_byte_checksum, force_2_byte_checksum;
+static int allow_2_byte_checksum;
 static char *ascii_fname, *only_fname;
 
 static char *MagicTypes[] = { "", "DOC", "XLS", "DOT", "XLT", "EXE", "DLL", "ZIP", "BMP", "DIB", "GIF", "PDF", "GZ", "TGZ", "BZ2", "TZ2", "FLV", "SWF", "MP3", "PST", NULL };
@@ -744,8 +744,8 @@ static int load_local_header(zip_file *zfp, zip_ptr *p)
 static int process_legacy(zip_file *zfp, zip_ptr *p)
 {
 	FILE *fp = zfp->fp;
+	int force_1_byte_checksum = 0;
 
-	force_1_byte_checksum = 0;
 	fprintf(stderr, "ver %d.%d ", p->version / 10, p->version % 10);
 
 	if ( (p->flags & FLAG_ENCRYPTED) &&
@@ -779,7 +779,8 @@ static int process_legacy(zip_file *zfp, zip_ptr *p)
 				// OLD winzip (I think 8.01 or before), is also supposed to be 2 byte.
 				// old v1 pkzip (the DOS builds) are 2 byte checksums.
 			{
-				zfp->check_bytes = 2;
+				if (allow_2_byte_checksum) // See #4541
+					zfp->check_bytes = 2;
 				zfp->check_in_crc = 0;
 			}
 
@@ -791,10 +792,7 @@ static int process_legacy(zip_file *zfp, zip_ptr *p)
 
 		scan_for_data_descriptor(zfp, p);
 
-		// Command line option -2 will override libarchive EFH
-		if (force_2_byte_checksum)
-			zfp->check_bytes = 2;
-		else if (force_1_byte_checksum)
+		if (force_1_byte_checksum)
 			zfp->check_bytes = 1;
 
 		fprintf(stderr,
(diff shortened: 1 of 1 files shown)


src/zip2john.c
```diff
@@ -779,7 +779,8 @@ static int process_legacy(zip_file *zfp, zip_ptr *p)
 				// OLD winzip (I think 8.01 or before), is also supposed to be 2 byte.
 				// old v1 pkzip (the DOS builds) are 2 byte checksums.
 			{
-				zfp->check_bytes = 2;
+				if (allow_2_byte_checksum) // See #4541
+					zfp->check_bytes = 2;
```
solardiz, line 783: Isn't it weird/wrong that we only use 2-byte checksum conditionally even after having checked for very specific `efh_id` values? Don't those known values guarantee we have 2-byte? Confused.

solardiz: > Don't those known values guarantee we have 2-byte

OK, I see that the issue we're trying to fix is `0x7875` doesn't guarantee 2-byte checksum.

I think we should add a comment that unfortunately checksum size seen with that EFH value varies between the two known archivers that use that value.

solardiz: So the main question is whether we want to hurt the speed at all other previously-whitelisted EFH values just because we found this problem for `0x7875` now. I doubt we want to do that, or at least we should have `zip2john` recommend use of `-2` for those. In fact, it could print a more verbose recommendation even for `0x7875`, mentioning the `bsdtar` issue.

ghost:
IMO, we should merge this as is to fix the known issue (false negative). Then discuss if there is a better approach.

solardiz:
@claudioandre-br OK, I've just merged this, although I am unhappy we're hurting performance even at EFH values for which no known problems exist.

PR #4562, opened 2021-02-04
Bench report weirdness

### LINKED ISSUE #4536: Weird reporting of formats "passed self-tests" vs. "benchmarked"
I was surprised by this discrepancy:

```
$ ../run/john -test=0 -form=lm,nt
Will run 64 OpenMP threads
Testing: LM [DES 64/64]... (64xOMP) PASS
Testing: NT [MD4 32/64]... PASS
All 2 formats passed self-tests!
$ ../run/john -test-full=0 -form=lm,nt
Will run 64 OpenMP threads
Testing: LM [DES 64/64]... (64xOMP) PASS
Testing: NT [MD4 32/64]... PASS
2 formats benchmarked.
```

Perhaps it shouldn't be "benchmarked" in the latter case. I went to look at the code, and it looks even worse to me there:

```c
#ifndef BENCH_BUILD
                        if (options.flags & FLG_MASK_CHK)
                                printf("%u formats benchmarked.\n", total);
                        else
#endif

#ifdef HAVE_OPENCL
/*
 * Allow OpenCL build's "--test" to run on no-OpenCL systems.
 * Print a message about no OpenCL at the end of the run.
 */
                                if (opencl_unavailable)
                                        opencl_was_skipped = " (OpenCL formats skipped)";
#endif

                                printf("All %u formats passed self-tests!%s\n",
                                       total, opencl_was_skipped);
```

The last `printf` above is part of `else` branch in non-OpenCL builds, but with OpenCL it's on its own. Perhaps that's unintentional and also wrong.

magnumripper:
Fix "passed self-tests" vs. "benchmarked" reporting
    
Basically there were artefacts from when OpenCL benchmarks with mask would imply that self-tests was skipped. Some other logic bugs fixed as well.

Closes #4536

### DIFF
--- a/src/Makefile.legacy
+++ b/src/Makefile.legacy
@@ -182,7 +182,7 @@ BENCH_OBJS = \
 	$(BENCH_MD5_OBJS_DEPEND) \
 	$(BENCH_BF_OBJS_DEPEND) \
 	bench-t.o best.o common_g.o config_g.o formats_g.o jumbo.o dyna_salt.o \
-	memory.o miscnl.o params.o path.o signals_g.o tty.o
+	memory.o miscnl.o params.o path.o signals_g.o tty-t.o
 
 GENMKVPWD_OBJS = \
 	genmkvpwd.o mkvlib.o memory.o miscnl.o path.o jumbo.o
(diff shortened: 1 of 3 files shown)


magnumripper:
Also included was a fix for generic build, as I tested it after these changes. Actually the problems found was older and I only fixed the "generic" build, while the restic format plugin still fails for *any* legacy build. I opened #4563 for that.

solardiz (approved):

PR #4565, opened 2021-02-06
Add error message to ansible2john

marcan2020:


run/ansible2john.py
```diff
@@ -31,6 +31,7 @@ def process_file(filename):
 
     data = open(filename, "rb").read()
     if not data.startswith(HEADER):
+        sys.stderr.write("File doesn't start with %s" % HEADER)
```
solardiz: Perhaps add `\n` after `%s`?

solardiz (changes requested):

run/ansible2john.py
marcan2020: @solardiz Good call. I changed it.

solardiz:
@marcan2020 Can you please squash the two commits into one and force-push? Alternatively, let us know that we should use GitHub's "Squash and merge", but it's dirtier (makes GitHub the committer per git history). Thanks!

marcan2020:
@solardiz It's done.

solardiz (approved):

solardiz:
@marcan2020 Merged. Thank you!

PR #4566, opened 2021-02-08
fixed 2 errors in gpg2john.c

sieger82:
1) in debug mode Getc() gets invoked 1 too many times resulting in eof errors
2) in debug mode salt is not added to outputted hash.

### DIFF
--- a/src/gpg2john.c
+++ b/src/gpg2john.c
@@ -804,6 +804,7 @@ string_to_key(void)
 {
 	int has_iv = YES;
 	int type = Getc();
+	int i;
 
 	switch (type) {
 	case 0:
@@ -820,11 +821,13 @@ string_to_key(void)
 		// hash_algs(Getc());
 		m_hashAlgorithm = Getc();
 		if (gpg_dbg) fprintf(stderr, "\t\tSalt - ");
-		if (gpg_dbg)
-			dump(8);
-		else
-			give(8, m_salt, sizeof(m_salt));
-		if (gpg_dbg) fprintf(stderr, "\n");
+		give(8, m_salt, sizeof(m_salt));
+		if (gpg_dbg) {
+			for (i = 0; i < 8; ++i) {
+				fprintf(stderr, "%02x", m_salt[i]);
+			}
+			fprintf(stderr, "\n");
+		}
 		break;
 	case 2:
 		if (gpg_dbg) fprintf(stderr, "\tReserved string-to-key(s2k %d)\n", type);
@@ -836,11 +839,13 @@ string_to_key(void)
 		m_hashAlgorithm = Getc();
 		// hash_algs(Getc());
 		if (gpg_dbg) fprintf(stderr, "\t\tSalt - ");
-		if (gpg_dbg)
-			dump(8);
-		else
-			give(8, m_salt, sizeof(m_salt));
-		if (gpg_dbg) fprintf(stderr, "\n");
+		give(8, m_salt, sizeof(m_salt));
+		if (gpg_dbg) {
+			for (i = 0; i < 8; ++i) {
+				fprintf(stderr, "%02x", m_salt[i]);
+			}
+			fprintf(stderr, "\n");
+		}
 		{
 			int count, c = Getc();
 			count = (16 + (c & 15)) << ((c >> 4) + EXPBIAS);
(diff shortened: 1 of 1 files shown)


src/gpg2john.c
```diff
@@ -820,8 +821,14 @@ string_to_key(void)
...
+		if (gpg_dbg) {
+			//FIXED: //if using dump(8) in debug mode, m_salt is not set and salt is not added to outputted hash
+			//dump(8);
+			give(8, m_salt, sizeof(m_salt));
+			for (i = 0; i < 8; ++i) {
+				fprintf(stderr, "%02x", m_salt[i]);
+			}
+		}
 		else
 			give(8, m_salt, sizeof(m_salt));
```
solardiz: I think both instances of the `give` call should ideally be replaced with one just before the `if`. And no need to clutter the code with comments about past bugs, the differences between these functions are obvious enough not to require that.

src/gpg2john.c
```diff
@@ -836,8 +843,14 @@ string_to_key(void)
...
```
solardiz: Ditto.

src/gpg2john.c
```diff
@@ -1138,8 +1151,12 @@ Symmetrically_Encrypted_and_MDC_Packet(int len, int first, int partial, char *ha
 	if (first) {
 		cp = hash;
 		totlen = 0;
-		if (gpg_dbg) fprintf(stderr, "\tVer %d\n", Getc());
-		Getc(); // version (we only read this from the first packet. Not read from rest of the 'partial' packets.
+		//FIXED: when using debug mode Getc() was invoked one too many times resulting in unexpected end of file errors.
```
solardiz: No need for the comment.

solardiz (changes requested):
I don't mind merging this as-is (the original code is very dirty, and this is fixing real bugs), but ideally you'd clean up these things I noted and force-push first. Thanks!

solardiz:
Also, let me suggest a replacement commit message:

```
gpg2john: Fix 2 debug mode bugs

1. Getc() was invoked 1 too many times resulting in EOF errors
2. Salt was not added to outputted hash
```

This separates the title from detail and avoids repetition. By using "was", it also makes it clear the described behavior is gone.

sieger82:
tnx for the comments. makes total sense. will fix and get back

sieger82:
I've made the changes you requested.

src/gpg2john.c
```diff
@@ -820,10 +821,12 @@ string_to_key(void)
...
-			dump(8);
-		else
-			give(8, m_salt, sizeof(m_salt));
+		give(8, m_salt, sizeof(m_salt));
+		if (gpg_dbg) {
+			for (i = 0; i < 8; ++i) {
+				fprintf(stderr, "%02x", m_salt[i]);
+			}
+		}
 		if (gpg_dbg) fprintf(stderr, "\n");
```
solardiz: Oh, surely this `fprintf` should go into the `if` block above instead of checking the same condition again. I'm sorry I didn't notice this right away - it became more apparent after the previous cleanup.

src/gpg2john.c
```diff
@@ -836,10 +839,12 @@ string_to_key(void)
...
```
solardiz: Ditto.

solardiz (changes requested):
Sorry, noticed two more cleanups to make within these changes.

src/gpg2john.c
```diff
@@ -836,17 +839,17 @@ string_to_key(void)
...
+		if (gpg_dbg) {
+			for (i = 0; i < 8; ++i) {
+				fprintf(stderr, "%02x", m_salt[i]);
+			}
+			fprintf(stderr, "\n");
 		}
+		int count, c = Getc();
+		count = (16 + (c & 15)) << ((c >> 4) + EXPBIAS);
+		if (gpg_dbg) fprintf(stderr, "\t\tCount - %d(coded count %d)\n", count, c);
+		m_count = count;
```
solardiz: Please revert the non-requested changes here. Having the scope of those two variables limited actually made sense.

sieger82: sorry, pushed too fast...

solardiz (approved):

solardiz:
I think we're almost done here - let's wait for CI and merge. Thank you!

PR #4574, opened 2021-02-24
zip2john: Only use 2-byte early-reject checks for version < 2.0

### LINKED ISSUE #4571: Dig deeper in pkzip format's one-or-two byte checksums
See #4541 and #4300 

I intend to look into this soon. I'm thinking these checksums are likely used somewhere for something other than our early rejection when cracking them - so there should be code somewhere that "does the right thing".

Sadly, zipinfo does not list any checksum at all. It does list CRC but the checksum we're interested in here is a separate checksum that is known without decompressing the whole file (hence our chance to early reject).

As noted in https://github.com/openwall/john/issues/4300#issuecomment-676179682, normal zipinfo/unzip tools doesn't seem to complain about a 2-byte checksum error. So are they always expecting a 1-byte checksum, or do they know just what we are missing here? Fortunately I believe there's a lot of open source to look at, in addition to fairly extensive documentation about the file format.

magnumripper:
Single-byte check for all other archives.  APPNOTE.TXT seems pretty clear about this, we shouldn't try to rely on any efh to indicate the matter.

Closes #4571, see also #4541 and #4300

Note that this fix not only disables 2-byte checks for some archives - it also *enables* them for others: Prior to this, any archive without any `efh` fields would get the single-byte check.

### DIFF
--- a/doc/NEWS
+++ b/doc/NEWS
@@ -160,6 +160,9 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
 - Addition of mosquitto2john.py for cracking of Eclipse Mosquitto passwd files.
   [blackfell; 2021]
 
+- Final fixes for zip2john selection of 1-byte or 2-byte early-reject checks,
+  after seeing several issues and digging into the specs.  [magnum; 2021]
+
 
 Major changes from 1.8.0-jumbo-1 (December 2014) to 1.9.0-jumbo-1 (May 2019):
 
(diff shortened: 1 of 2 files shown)


magnumripper:
Apart from the spec. being fairly clear, this fix was tested with all files we have in our samples repo (a handful) as well as a bunch of non-public files I had around. All the former still cracks (some now *got* 2-byte checks, while some lost it) and a bunch of my non-public "uncrackable" files now cracked with trivial passwords.

magnumripper:
Merged manually after amending doc/NEWS as well

PR #4577, opened 2021-02-26
Put a fork in the pkzip format (and zip2john) CS size & source

### LINKED ISSUE #4571: Dig deeper in pkzip format's one-or-two byte checksums

magnumripper:
After scrutinizing specs and open-source code, settle with a hopefully correct logic for chosing whether to use one or two byte early-rejection, and whether to use part of the timestamp or of the CRC-32 for it.

This version does not introduce yet another hash input format.  Instead it reverts to the original one, but now filling it correctly.  In order to really maximize early-rejection we'd need to define a new format that defines 1-or-2-byte CS at file level as opposed to archive level.  But that may be overkill, it's not very common to see both.

Closes #4571 (again)

### DIFF
--- a/src/pkzip.h
+++ b/src/pkzip.h
@@ -1,3 +1,10 @@
+/*
+ * This software is Copyright (c) 2011-2018 Jim Fougeron,
+ * Copyright (c) 2013-2021 magnum,
+ * and it is hereby released to the general public under the following terms:
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted.
+ */
 #ifndef PKZIP_H
 #define PKZIP_H
 
@@ -38,6 +45,7 @@ typedef struct zip_magic_signatures_t {
 
 typedef struct zip_hash_type_t {
 	u8 *h;						// at getsalt time, we leave these null.  Later in setsalt, we 'fix' them
+	u8 type;					// JtR hash version. Version 2 ($pkzip2$) is now the deprecated one.
 	u16 c;
 	u16 c2;
 	u64 datlen;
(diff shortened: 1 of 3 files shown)


magnumripper:
I believe this is good now. Speed is increased for many hashes since we no longer blindly check two different checksums without knowing which is correct.

I'll perform some more manual testing before we merge

magnumripper:
Speed increased ~10% for our collection of sample zip hashes (tested with `-mask:?u -len=6` to completion).
Newer john cracks older zip2john output (as long as it was correct at all).
Older john cracks newer zip2john output (better and faster than before, since the most important changes were in zip2john). I believe even really ancient versions of john will work faster & finer with the newest zip2john (not that it's something we'd recommend).

ghost:
PR passes my zip cracking tests (PS: they're far from a complete/extensive test).

magnumripper:
Again I forgot about NEWS, so updated it and merged manually (to avoid new bot-checks)

Issue #4585, opened 2021-03-03
Apple M1 build


magnumripper:
I got my hands on a Macbook M1 for a while and thought I'd sort out JtR on it. This issue will contains a bunch of notes and likely end up with a PR or two, if needed.

So, starting with a 100% pristine Macbook. Before anything else, I installed xcode and command-line tools. After that (maybe before it as well, I dunno) I had git(1) so could clone the repo. Then,

First try: Just run `./configure` and see what happens. As expected, it failed due to lack of OpenSSL.

Second try: Install Homebrew OpenSSL and only that. This time, (passing LDFLAGS and CPPFLAGS as recommended from Homebrew) we passed configure stage but build failed.

It appears configure detected ARM and 64-bit but used `arm32le.h` and NEON as opposed to `arm64le.h` and ASIMD.

https://en.wikipedia.org/wiki/Mac_transition_to_Apple_Silicon

solardiz:
Thank you, @magnumripper!

This is only subtly related, but FWIW as far as I'm aware (from user feedback) our builds made on/for Intel Macs with SSE4 do work on M1 via Rosetta 2, however those with AVX2 don't (print the "AVX2 is required" message).

magnumripper:
> FWIW as far as I'm aware (from user feedback) our builds made on/for Intel Macs with SSE4 do work on M1 via Rosetta 2

That's interesting, I'll try things like that as well (and comparing benchmarks).

I haven't googled anywhere near what I need to yet, but I was also under the impression even running an ARM build wouldn't quite be "native" Silicon M1 (but "more native" than intel binaries). Lots of things to learn here.

magnumripper:
Simply re-linking arch.h to arm64le.h didn't help at all - same build error.

magnumripper:
Trying `--disable-simd` for a starter. That works as-is.

Using arm32le.h, I get DEScrypt speeds of 5334K (many salts) and 5060K (one salt). NT-opencl benchmarks at 958253K.

Using arm64le.h, nothing much changed and I realized even with arm32le.h we got 64-bit bitslice. Not sure exactly what's going on there.

I also noticed OpenCL figures vary a fair bit between runs, even with same auto-tune results.

Umm I also see now that I didn't get an OpenMP build (native clang compiler). Even given that, the DEScrypt figures are mediocre.

magnumripper:
> the DEScrypt figures are mediocre.

[...]

PR #4587, opened 2021-03-03
Apple M1 (and likely some other arm64) patches

magnumripper:
See #4585

### DIFF
--- a/src/m4/jtr_arm_logic.m4
+++ b/src/m4/jtr_arm_logic.m4
@@ -44,11 +44,29 @@ if test "x$enable_native_tests" = xyes; then
     ,[CPU_NOTFOUND="1"]
      [AC_MSG_RESULT(no)]
   )
+  CC="$CC_BACKUP -march=armv8-a+simd"
+  AC_MSG_CHECKING([for ASIMD])
+  AC_RUN_IFELSE(
+    [
+    AC_LANG_SOURCE(
+    [[#include <arm_neon.h>
+  #include <stdio.h>
+  extern void exit(int);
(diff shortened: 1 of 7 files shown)


magnumripper:
Tested for regression on Raspberry Pi, 32-bit and 64-bit. No problem seen - on the contrary, both got the RAR format again, which was disabled in 2017 by ddeabc6d.

src/m4/jtr_arm_logic.m4
```diff
@@ -44,11 +44,29 @@ if test "x$enable_native_tests" = xyes; then
...
   )
+  CC="$CC_BACKUP -march=armv8-a+simd"
+  AC_MSG_CHECKING([for ASIMD])
+  AC_RUN_IFELSE(
+    [
+    AC_LANG_SOURCE(
+    [[#include <arm_neon.h>
+  #include <stdio.h>
+  extern void exit(int);
+  int main(){uint32x4_t t;*((long*)&t)=1;t=veorq_u32(t,t);if((*(unsigned*)&t)==88)printf(".");exit(0);}]]
```
solardiz, line 55: It's not great that we have C aliasing violations in these configure test programs. I recall us discussing this before (for some other arch), but I don't recall what the outcome of the discussion was.

solardiz (approved):
Overall, this looks good to me. We might address my nitpick about the configure tests separately.

src/m4/jtr_arm_logic.m4
magnumripper: While at that we could drop the need for stdio and just end with a calculated exitcode, expected 0 - the current test of "88" is most everywhere and doesn't test correctness, we only look for build ok/fail and run ok/segfault. Let's make a separate issue for it.

Issue #4579, opened 2021-03-01
ETA disappears at 7th character


sectroyer:
In order to reproduce this issue just run attached test.pwd with following command:
`john --format=descrypt '-2=[xyz0123456789]' -1=?l?d --min-len=1 --mask=?2?1?u?1?1?1?1?1  --node=1467-4136/4136 ./test.pwd `                                     

`Using default input encoding: UTF-8`
`Loaded 1 password hash (descrypt, traditional crypt(3) [DES 256/256 AVX2])`
`Will run 32 OpenMP threads`
`Node numbers 1467-4136 of 4136`
`Press 'q' or Ctrl-C to abort, almost any other key for status`
`0g 0:00:00:00  (4) 0g/s 0p/s 0c/s 0C/s`
`0g 0:00:00:00 0.00% (5) (ETA: 17:39:26) 0g/s 13107Kp/s 13107Kc/s 13107KC/s 1dD6..27L8h`
`0g 0:00:00:00 0.00% (6) (ETA: 17:39:26) 0g/s 41506Kp/s 41506Kc/s 41506KC/s z7Yq7..3lZiwh`
`0g 0:00:00:08  (7) 0g/s 45270Kp/s 45270Kc/s 45270KC/s 3dY877..63Y3iwh`
`0g 0:00:00:11  (7) 0g/s 45416Kp/s 45416Kc/s 45416KC/s 5sMs09h..xcOp09h`
`0g 0:00:00:15  (7) 0g/s 45682Kp/s 45682Kc/s 45682KC/s zmHncsu..4dTbcsu`

More info about configuration can be found here: [https://github.com/openwall/john/issues/4576](url)
I was able to narrow bug down to (what looks like) an integer overflow in mask.c in this line:
offset = my_candidates * (options.node_min - 1);
After commenting this line I could see eta but still not sure if it's calculated correctly, will do more tests.
[test.pwd.zip](https://github.com/openwall/john/files/6063060/test.pwd.zip)

ghost:
> offset = my_candidates * (options.node_min - 1);

The overflow does not happen in this line. Offset is 1,9^13. UINT64_MAX is 1,8^19. 

It happens at 1853 `my_candidates = total_candidates - offset;`.  Because `(total_candidates < offset)`.

magnum, a check like this sounds reasonable? **[EDITED], it seems something else is wrong.**
```
diff --git a/src/mask.c b/src/mask.c
index 4803aa39f..c45440690 100644
--- a/src/mask.c
[...]

PR #4592, opened 2021-03-05
Fix --status for --fork with --node sessions; support any node ranges divisible by fork count

### LINKED ISSUE #4584: Allow fork divisible node values
I have played around with using --fork instead of openmp as @solardiz suggested and on some nodes I am getting pretty good results. Therefore I would like to implement it on my system. However I have see2, avx and avx2 nodes. As @solardiz suggested "overbooking" them does work however this bring up the issue that some of those machines I do NOT want to use all cores for john (they also have other tasks to do ;)). I understand why --node is connected with --fork and making them independent won't be an easy task. However if I am able to "overbook" --fork it should be also possible (with --fork divisible number) to "overbook" --node. I hope this would be much simpler to implement and won't require big code changes?

### LINKED ISSUE #4589: --status of a --fork session fails to consider --node range found in main .rec file
I have switched to fork in system and as a result had to scale down nodes. Now I think I have encountered another issue with --status switch. Half of the nodes have always status completed. Here is the cmd:
`OMP_NUM_THREADS=1 john --format=descrypt '-2=[opqrstuvwxyz0123456789]' -1=?l?d --min-len=1 --mask=?2?1?1?1?1?1?1?u --node=16-47/47 --fork=32 ./test.pwd --session=test`
When I remove --session and use default session instead of "completed" they simply stop showing progress. I attach relevant screenshots 
[test.pwd.zip](https://github.com/openwall/john/files/6082193/test.pwd.zip)
Zrzut ekranu 2021-03-4 o 09 59 47
Zrzut ekranu 2021-03-4 o 09 56 31
Zrzut ekranu 2021-03-4 o 10 04 33

solardiz:
This is a merge of these changes I had made and tested in core earlier today. In jumbo, I've only tested that these build (but didn't try building with MPI support). I haven't yet tested whether they actually work. I hope @sectroyer and others will help with testing.

Fixes #4589, closes #4584.

### DIFF
--- a/doc/CHANGES
+++ b/doc/CHANGES
@@ -1,5 +1,6 @@
 The following changes have been made between John 1.9.0 and 1.9.1-dev:
 
+* Let "--node" numbers range be any multiple of "--fork" process count.
 * Fixed "--status" to work right on sessions that use "--fork" with "--node".
 * Added two external modes: Filter_NoRepeats skips candidate passwords that
 contain the same character more than once, and Filter_Repeats is its inverse.
--- a/doc/CHANGES
+++ b/doc/CHANGES
@@ -1,3 +1,8 @@
+The following changes have been made between John 1.9.0 and 1.9.1-dev:
+
+* Added two external modes: Filter_NoRepeats skips candidate passwords that
+contain the same character more than once, and Filter_Repeats is its inverse.
+
 The following changes have been made between John 1.8.0 and 1.9.0:
 
 * Increased the interleaving for bcrypt on x86-64 from 2x to 3x for a major
--- a/doc/CHANGES
+++ b/doc/CHANGES
@@ -1,5 +1,6 @@
 The following changes have been made between John 1.9.0 and 1.9.1-dev:
 
+* Fixed "--status" to work right on sessions that use "--fork" with "--node".
 * Added two external modes: Filter_NoRepeats skips candidate passwords that
 contain the same character more than once, and Filter_Repeats is its inverse.
 
--- a/doc/CHANGES
+++ b/doc/CHANGES
@@ -376,5 +376,3 @@ OpenBSD/PowerPC, OpenBSD/PA-RISC, OpenBSD/VAX, NetBSD/VAX, Solaris/SPARC64,
 Mac OS X (PowerPC and x86), SCO, BeOS.
 * Bug and portability fixes, and new bugs.
 * Bonus: "Strip" cracker included in the default john.conf (john.ini).
-
-$Owl$
--- a/doc/CONFIG
+++ b/doc/CONFIG
@@ -239,5 +239,3 @@ write to.  Names of files that do not exist will be silently ignored.
 	Defining an external mode.
 
 See EXTERNAL.
-
-$Owl$
--- a/doc/CONTACT
+++ b/doc/CONTACT
@@ -26,5 +26,3 @@ Please don't ask questions until you read the FAQ.
 Commercial support for John the Ripper is available from Openwall:
 
 	https://www.openwall.com/services/
-
-$Owl$
--- a/doc/CREDITS
+++ b/doc/CREDITS
@@ -72,5 +72,3 @@ Developer's Kit) that I used for the Win32 port:
 	http://www.delorie.com/djgpp/
 
 * Charles W Sandmann - for CWSDPMI, the DPMI server used with DJGPP.
-
-$Owl$
--- a/doc/EXAMPLES
+++ b/doc/EXAMPLES
@@ -403,5 +403,3 @@ You can find the actual implementation of such a cracking mode with lots
 of comments in the default configuration file supplied with John.
 Please refer to EXTERNAL for information on the programming language
 used.
-
-$Owl$
--- a/doc/EXTERNAL
+++ b/doc/EXTERNAL
@@ -221,5 +221,3 @@ new()/next()/next().... and know that the script restored itself properly.
 
 Following the above rules and functions, a external-hybrid script can do
 pretty much anything required to transform/morph/mangle words.
-
-$Owl$
--- a/doc/FAQ
+++ b/doc/FAQ
@@ -306,5 +306,3 @@ Although the meaning of some of the numbers that get into .rec files is
 trivial to explain, it is not possible to reasonably describe some
 others without going into great detail on John internals.  If you really
 need to know, read the source code.
-
-$Owl$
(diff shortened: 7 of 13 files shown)


solardiz:
@magnumripper I notice there's MPI-related functionality that was copy-paste-edit from my fork/node stuff. I haven't updated it to mimic my revised code here, merely trying to keep it working like it was before. _Maybe_ it would also benefit from the arbitrary multiples support, or maybe not - I think this is something for you to consider.

solardiz:
> I haven't yet tested whether they actually work.

I just did. My CPU only tests pass just like they did in core. I haven't yet tested with OpenCL nor ZTEX. I'd appreciate others helping test all of these combinations.

BTW, one maybe-shortcoming is that the status lines are prefixed with just one node number - the start of range for each process. So e.g. `-node=9-12/40 -fork=2` produces status lines like these:

```
9 0g 0:00:00:35  0g/s 65.82p/s 223992c/s 223992C/s morti
11 7g 0:00:00:35  0.2000g/s 65.82p/s 228644c/s 228644C/s saria11
```

where the first process is actually virtual nodes 9 and 10, and the second is 11 and 12. Maybe we should prefix these with `9-10` and `11-12`, respectively? Anyway, this would be a separate commit (also to core) and can as well be in a separate PR.

solardiz:
@sectroyer Just to make it clear, you'll need to clone the fork from my personal account to test this before we merge it.

sectroyer:
> @sectroyer Just to make it clear, you'll need to clone the fork from my personal account to test this before we merge it.

I assumed I have to wait for it to merge. K will do that tomorrow. Today fever so have to wait to get little better ;)

sectroyer:
If read code correctly this doesn't have a fix for this: https://github.com/openwall/john/issues/4579#issuecomment-788136376 ?

sectroyer:
Okay I switched to this version and it works just 47 node on 32 core system is always completed :D I guess this has something to do with warning I get ? (attaching screen shots). Command was:
`OMP_NUM_THREADS=1 john --format=descrypt '-2=[pqrstuvwxyz0123456789]' -1=?l?d --min-len=1 --mask=?u?2?1?u?1?1?1?1 --session=~/work//sessions/apic_brute_des_john_1 --node=29-92/92 --fork=32 test.pwd
[test.pwd.zip](https://github.com/openwall/john/files/6096874/test.pwd.zip)
`

Zrzut ekranu 2021-03-7 o 10 28 57
Zrzut ekranu 2021-03-7 o 10 29 06

solardiz:
> If read code correctly this doesn't have a fix for this: [#4579 (comment)](https://github.com/openwall/john/issues/4579#issuecomment-788136376) ?

Yes, this PR is only about what's in its title (edit: plus cosmetic changes related to core's recent move from CVS to Git). It would have been wrong to mix any ETA stuff in here, and besides I am not going to look into ETA issues - I leave that to others.

> it works just 47 node on 32 core system is always completed :D I guess this has something to do with warning I get ?

Looks like that node is somehow getting too little work. Looks like an issue specific to mask mode. As I didn't implement node support for mask mode (I only did for modes that come from core), I am not going to look into this - others might.

Thank you for testing!

solardiz:
> it works just 47 node on 32 core system is always completed :D

@sectroyer I guess you can reproduce this issue with code prior to this PR by using `--node=29-92/92 --fork=64`? If so, this would confirm it's not a bug in this PR's changes, but is an issue we had before. Can you do this test, please? (In fact, I'd expect nodes 47 and 48 to complete early with that test.)

sectroyer:
OK, will test. Meanwhile I noticed other issue. This 32 core machine doesn't finish at all and john enters "infinite loop" or something like that :) Take a look:
image

solardiz:
> This 32 core machine doesn't finish at all and john enters "infinite loop" or something like that :) Take a look

Doesn't look that way. Looks like uneven work distribution between nodes, leaving more work for nodes 29-30 and 91-92 than for all others. I think it's part of the same problem that made nodes 47-48 complete too soon. Someone will need to look into and improve mask mode's work distribution to nodes and test it across a variety of masks and length ranges.

magnumripper:
> @magnumripper I notice there's MPI-related functionality that was copy-paste-edit from my fork/node stuff. I haven't updated it to mimic my revised code here, merely trying to keep it working like it was before. _Maybe_ it would also benefit from the arbitrary multiples support, or maybe not - I think this is something for you to consider.

Right... I think we should just leave MPI as-is.

magnumripper:
> Right... I think we should just leave MPI as-is.

On second thought, a feature we've had for a long time is that any forked run can be stopped and resumed using MPI instead, and vice versa. So I think I'll update the MPI code and test it.

solardiz:
> a feature we've had for a long time is that any forked run can be stopped and resumed using MPI instead, and vice versa.

Did anyone know, besides you?

> So I think I'll update the MPI code and test it.

I guess you'd do that separately from this PR, and after we've merged it, right?

I actually had a different thought regarding MPI: we might want to add an alternative MPI support mode for heterogeneous clusters where the number of virtual nodes would be much greater than the number of logical CPUs in the cluster, and processes would dynamically (via MPI messages) request/obtain next virtual node numbers after they've done their work with the previous number. (Like what I had posted to john-users with many masks and file renames, but instead with many nodes and MPI.) You want to open a separate issue for this?

magnumripper (approved):
I verified this doesn't seem to cause any regressions with MPI. Will test more later but probably not until making same changes to it (#4594)

magnumripper:
> > a feature we've had for a long time is that any forked run can be stopped and resumed using MPI instead, and vice versa.
> 
> Did anyone know, besides you?

It's clearly documented, but that doesn't mean anyone read it, lol

> > So I think I'll update the MPI code and test it.
> 
> I guess you'd do that separately from this PR, and after we've merged it, right?

Yes.

> I actually had a different thought regarding MPI: we might want to add an alternative MPI support mode for heterogeneous clusters where the number of virtual nodes would be much greater than the number of logical CPUs in the cluster, and processes would dynamically (via MPI messages) request/obtain next virtual node numbers after they've done their work with the previous number. (Like what I had posted to john-users with many masks and file renames, but instead with many nodes and MPI.) You want to open a separate issue for this?

That's a cool idea, for later.

sectroyer:
> > it works just 47 node on 32 core system is always completed :D
> 
> @sectroyer I guess you can reproduce this issue with code prior to this PR by using `--node=29-92/92 --fork=64`?

With those settings node finished without any issues.

sectroyer:
>>Someone will need to look into and improve mask mode's work distribution to nodes and test it across a variety of masks and length ranges.

Should I create a separate issue for this bug ?

PR #4597, opened 2021-03-07
MPI: Let --node numbers range be any multiple of MPI process count

### LINKED ISSUE #4594: MPI: support any node ranges divisible by fork count
Mimic the fork changes in #4592 for MPI (after the former is merged).

magnumripper:
Implement the exact same changes recently made to --fork, but to the MPI code paths.

See 82d0825a8, #4592, closes #4594.

### DIFF
--- a/src/john.c
+++ b/src/john.c
@@ -561,7 +561,7 @@ static void john_fork(void)
 				usleep(i * 100000);
 			}
 #endif
- 			if (rec_restoring_now) {
+			if (rec_restoring_now) {
 				unsigned int save_min = options.node_min;
 				unsigned int save_max = options.node_max;
 				unsigned int save_count = options.node_count;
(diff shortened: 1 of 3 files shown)


magnumripper:
This was almost a copy-paste job so I'm fairly confident it's correct (provided the `--fork` code path is), but I'll perform some more testing over the nearest few days.

solardiz (approved):

solardiz:
Why did we just get "Your CircleCI tests were canceled" here?

magnumripper:
> Why did we just get "Your CircleCI tests were canceled" here?

Beats me. Apparently they (at least the one I looked at) built fine, then was canceled during `--test`. I'll try a force push or something.

ghost:
There is something else:

image

magnumripper:
That's even stranger. What did we violate, suddenly?

ghost:
Solar will have to open a ticket. 

BTW: I noticed something strange in Openwall's "plan overview". I found 5 builds in March (651 Compute Minutes). Twice as many as me in the year. How is that possible?

solardiz:
> Solar will have to open a ticket.

I've just submitted a support request to Circle CI.

src/john.c
```diff
@@ -618,23 +618,32 @@ static void john_fork(void)
...
+				    options.node_count != save_count ||
+				    options.fork != save_fork)
 					fprintf(stderr,
 					    "Inconsistent crash recovery file:"
 					    " %s\n", rec_name);
-				options.node_min = options.node_max = node_id;
 			}
 		}
+		options.node_min += mpi_id * npf;
+		options.node_max = options.node_min + npf - 1;
```
solardiz, line 646: Previously, these lines were reached unconditionally, but now they're within `if (mpi_p > 1)`. Isn't this a bug for when that condition isn't met (running MPI with `mpi_p == 1`?)

Also, I assume `mpi_id` can be 0, for this to be equivalent to the main/parent process with `--fork`?

solardiz:
> I've just submitted a support request to Circle CI.

They replied (effectively) that this was a false positive detection, which they reset - so Circle CI should be back with us now.

src/john.c
magnumripper: Reopened #4594 for this

PR #4602, opened 2021-03-08
Subsets: Clarify documentation about exhaustive search

### LINKED ISSUE #4575: Confusion around Subsets mode not testing all possible permutations
Hello there!
I have a problem with the subsets, it does not create all the possible permutations, for example I use this command and the word 'abcdefgh' should appear but it does not.
I don't know if I don't understand how this feature works or if it is a bug or something, I would appreciate your help.
I used this command exactly:

`./john   --max-len:8 --subset:abcdefgh --stdout | grep 'abcdefgh'`

I tested it on the latest version on a kali and the latest version of JtR.

magnumripper:
Also improves reporting of total keyspace: What was logged as total keyspace was actually per length.  We now do log the total at session start, if we can express it with a 64-bit integer.

Closes #4575

### DIFF
--- a/src/subsets.c
+++ b/src/subsets.c
@@ -584,14 +584,35 @@ int do_subsets_crack(struct db_main *db, char *req_charset)
 		keyspace += numwords(i, charcount, subsets_cur_len, required);
 
 	if (john_main_process) {
+		int len;
+		uint64_t total_keyspace = 0;
+
+		for (len = subsets_cur_len; len <= maxlength; len++) {
+			for (i = min_comb; i <= MIN(len, maxdiff); i++) {
+				uint64_t nw = numwords(i, charcount, len, required);
+				if (total_keyspace + nw < total_keyspace) {
+					total_keyspace = 0;
+					break;
+				}
+				total_keyspace += nw;
+			}
+			if (!total_keyspace)
+				break;
+		}
+
 		log_event("Proceeding with \"subsets\" mode");
 		log_event("- Charset: %s size %d", req_charset ? req_charset : charset,
 		          charcount);
 		log_event("- Lengths: %d-%d, max. subset size %d",
-		          MAX(options.eff_minlength, 1), maxlength, maxdiff);
+		          word_len, maxlength, maxdiff);
 		if (required)
 			log_event("- Required set: First %d of charset", required);
-		log_event("- Total keyspace: %" PRIu64, keyspace);
+		if (total_keyspace)
+			log_event("- Total keyspace: %" PRIu64, total_keyspace);
+		else
+			log_event("- Total keyspace: larger than 64-bit");
+		if (word_len < maxlength)
+			log_event("- Length %d total keyspace: %" PRIu64, subsets_cur_len, keyspace);
 		if (rec_restored) {
 			fprintf(stderr, "Proceeding with \"subsets\"%s%s",
 			        req_charset ? ": " : "",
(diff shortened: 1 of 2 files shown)


src/subsets.c
```diff
@@ -584,14 +584,35 @@ int do_subsets_crack(struct db_main *db, char *req_charset)
...
+		uint64_t total_keyspace = 0;
+
+		for (len = subsets_cur_len; len <= maxlength; len++) {
+			for (i = min_comb; i <= MIN(len, maxdiff); i++) {
+				uint64_t nw = numwords(i, charcount, len, required);
+				if (total_keyspace + nw < total_keyspace) {
+					total_keyspace = 0;
+					break;
+				}
+				total_keyspace += nw;
```
solardiz, line 597: This can be simplified to first doing the `total_keyspace += nw;` and then seeing if `total_keyspace < nw`. So that you don't need to perform the addition twice.

src/subsets.c
```diff
@@ -584,14 +584,35 @@ int do_subsets_crack(struct db_main *db, char *req_charset)
...
+		          word_len, maxlength, maxdiff);
 		if (required)
 			log_event("- Required set: First %d of charset", required);
-		log_event("- Total keyspace: %" PRIu64, keyspace);
+		if (total_keyspace)
+			log_event("- Total keyspace: %" PRIu64, total_keyspace);
+		else
+			log_event("- Total keyspace: larger than 64-bit");
+		if (word_len < maxlength)
+			log_event("- Length %d total keyspace: %" PRIu64, subsets_cur_len, keyspace);
```
solardiz, line 615: The logging of per-length keyspace (here and below) is not conditional on non-overflow. This is OK'ish if a given length is only reached if the previous length's keyspace was just searched (suggesting the previous was way below 64-bit, and thus the current length's probably fits in 64-bit). However, can this also be reached for the very first length already being beyond 64-bit?

solardiz (approved):

src/subsets.c
magnumripper: > The logging of per-length keyspace (here and below) is not conditional on non-overflow. (...) can this also be reached for the very first length already being beyond 64-bit?

It never was, prior to this. I think I know how to produce that... yep.
```
../run/john -subsets:full-unicode -min-len=6 -stdout -log >/dev/null
```
What happens is we get a bogus figure for length 6 (it says 4315209972371217376 but has wrapped around) but the mode still works (I think) - we're just getting bogus ETA/progress.

We might want to fix this but it's kind of a separate issue.

magnumripper: > the mode still works (I think)

This is kind of hard to test lol, we'd just have dry-running at hand.

> We might want to fix this but it's kind of a separate issue.

What should we do if we detect an overflow? Just disable ETA/progress and/or warn, or refuse to run? Just brainstorming real quick without thinking much, we could force a decrease of subsets-max-diff until we don't overflow but it'd be hard to document... then again we could obviously print what we did, right away.

magnumripper: I didn't look at the code for a while but it's recursive and my gut feeling is it'll probably work just fine as long as it doesn't run out of stack or otherwise run out of memory.

solardiz: > What should we do if we detect an overflow? Just disable ETA/progress and/or warn, or refuse to run?

I think ideally we'd somehow report that the ETA is too far into the future, but proceed anyway. Even reporting specifically `UINT64_MAX` would be better than reporting an arbitrary wrapped number (which might happen to be low, which would be misleading).

magnumripper: I'll have a look at it. I know there are some other overflow detections in there already - I'll review them and maybe add some. Thanks!

PR #4608, opened 2021-03-09
Subsets: Add more overflow detection.

magnumripper:
Overflows should only affect ETA/progress reporting unless we're in keyspaces so huge we'd run out of stack or memory anyway due to recursion.  If we detect an overflow, turn ETA/progress off but continue.

### DIFF
--- a/src/subsets.c
+++ b/src/subsets.c
@@ -39,11 +39,9 @@
 #include "unicode.h"
 #include "unicode_range.h"
 
-#define SUBSET_DEBUG 1
-
 #define MAX_SUBSET_SIZE 16
 #define MAX_CAND_LENGTH PLAINTEXT_BUFFER_SIZE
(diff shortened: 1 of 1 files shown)


src/subsets.c
```diff
@@ -39,11 +39,9 @@
 #include "unicode.h"
 #include "unicode_range.h"
 
-#define SUBSET_DEBUG 1
-
```
magnumripper, line 43: As you can see, this was always true but I have yet to see an overflow in `powi()` or `fac()` with normal usage (or actually in any usage) of this mode. I decided to drop the macros and continue always doing the checks.

That said, the limits are a whole lot tighter running a 32-bit build, which I almost never do.

solardiz (approved):
I don't fully follow the logic here just from patch context, but nothing looks obviously wrong to me.

PR #4611, opened 2021-03-09
Fix alignment compile errors on GCC 11

### LINKED ISSUE #4604: Compile error on gcc 11 (Fedora 34)
I'm on Fedora 33 x64, doing a Fedora 34 x64 mock build for john  ```1.9.0-Jumbo-1-464-gc9825e688```.

F34 has gcc 11:
```
# gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/libexec/gcc/x86_64-redhat-linux/11/lto-wrapper
OFFLOAD_TARGET_NAMES=nvptx-none
OFFLOAD_TARGET_DEFAULT=1
Target: x86_64-redhat-linux
Configured with: ../configure --enable-bootstrap --enable-languages=c,c++,fortran,objc,obj-c++,ada,go,d,lto --prefix=/usr --mandir=/usr/share/man --infodir=/usr/share/info --with-bugurl=http://bugzilla.redhat.com/bugzilla --enable-shared --enable-threads=posix --enable-checking=release --enable-multilib --with-system-zlib --enable-__cxa_atexit --disable-libunwind-exceptions --enable-gnu-unique-object --enable-linker-build-id --with-gcc-major-version-only --with-linker-hash-style=gnu --enable-plugin --enable-initfini-array --with-isl --enable-offload-targets=nvptx-none --without-cuda-driver --enable-gnu-indirect-function --enable-cet --with-tune=generic --with-arch_32=i686 --build=x86_64-redhat-linux
Thread model: posix
Supported LTO compression algorithms: zlib zstd
gcc version 11.0.0 20210210 (Red Hat 11.0.0-0) (GCC) 
```

The rpm spec file uses these build commands:
```
cd src
./configure --disable-rexgen --enable-asan
make clean
make %{?_smp_mflags}
```

The error:
```
gcc -DAC_BUILT -mavx2 -DJOHN_AVX2 -c -g  -Og -fsanitize=address -m64 -I/usr/local/include -DARCH_LITTLE_ENDIAN=1 -DPREFER_FLOCK  -Wall -fno-omit-frame-pointer -Wno-deprecated-declarations -Wunused-but-set-variable -std=gnu89 -Wdate-time -D_POSIX_SOURCE -D_GNU_SOURCE -D_XOPEN_SOURCE=600  -fopenmp  -pthread -I/usr/local/include -DCL_SILENCE_DEPRECATION -funroll-loops argon2_encoding_plug.c -o argon2_encoding_plug.o
In file included from argon2_core_plug.c:32:
blake2.h:112:5: error: size of array element is not a multiple of its alignment
  112 |     blake2b_state S[4][1];
      |     ^~~~~~~~~~~~~
blake2.h:113:5: error: size of array element is not a multiple of its alignment
  113 |     blake2b_state R[1];
      |     ^~~~~~~~~~~~~
make[1]: *** [Makefile:1592: argon2_core_plug.o] Error 1
make[1]: Leaving directory '/builddir/build/BUILD/john-1.9.0.jumbo.1/src'
```

### Steps to reproduce ###
- Install a Fedora 33 machine.
- Install mock

- You can reproduce it by using the repo 
```
https://gitlab.com/fhuberts/rpmsUpstream.git
```
at commit ```315a54e```.

- This repo needs a checkout of
```
https://gitlab.com/fhuberts/rpmbuilder.git
```
alongside it.

- You have ```rpmbuilder``` and ```rpmsUpstream``` in the same directory.

- Next you do
```
cd ./rpmsUpstream/fedora/john
```

- And then
```
../buildRPMs  -M /etc/mock/fedora-34-x86_64.cfg rpm.spec
```

fhuberts:
Fixes #4604

Signed-off-by: Ferry Huberts <ferry.huberts@pelagic.nl>

### DIFF
--- a/src/blake2.h
+++ b/src/blake2.h
@@ -57,7 +57,7 @@ extern "C" {
     uint8_t  personal[BLAKE2S_PERSONALBYTES];  // 32
   } blake2s_param;
 
-  JTR_ALIGN( 64 ) typedef struct __blake2s_state
+  typedef struct JTR_ALIGN( 64 ) __blake2s_state
   {
     uint32_t h[8];
     uint32_t t[2];
@@ -82,7 +82,7 @@ extern "C" {
     uint8_t  personal[BLAKE2B_PERSONALBYTES];  // 64
   } blake2b_param;
 
-  JTR_ALIGN( 64 ) typedef struct __blake2b_state
+  typedef struct JTR_ALIGN( 64 ) __blake2b_state
   {
     uint64_t h[8];
     uint64_t t[2];
@@ -94,7 +94,7 @@ extern "C" {
 #if defined(JOHN_NO_SIMD) || (!defined(__SSE2__) && !defined(__SSE4_1__) && !defined(__XOP__))
   typedef struct __blake2sp_state
 #else
-  JTR_ALIGN( 64 ) typedef struct __blake2sp_state
+  typedef struct JTR_ALIGN( 64 ) __blake2sp_state
 #endif
   {
     blake2s_state S[8][1];
@@ -106,7 +106,7 @@ extern "C" {
 #if defined(JOHN_NO_SIMD) || (!defined(__SSE2__) && !defined(__SSE4_1__) && !defined(__XOP__))
   typedef struct __blake2bp_state
 #else
-  JTR_ALIGN( 64 ) typedef struct __blake2bp_state
+  typedef struct JTR_ALIGN( 64 ) __blake2bp_state
 #endif
   {
     blake2b_state S[4][1];


solardiz:
Yesterday, I spent maybe an hour in total on looking where these explicit alignments even came from (inconclusive) and whether they (still) exist upstream (no, they don't - checked `blake2.h` in Argon2 and BLAKE2 upstreams, as well as in some BLAKE2 forks that carry additional SIMD support/optimizations). So a longer-term fix is to update to latest Argon2 upstream, and separately to consider what we do with the raw BLAKE2 format(s) and the Monero format.

Meanwhile, maybe we can accept a PR like this, but I'm not entirely sure it's right. In fact, I was wondering whether this would even pass CI, because as I understand it ends up enlarging each array element, which might not be the layout expected after some typecasts to SIMD types. Now I wonder whether there's a performance impact (from worse locality of reference).

I _guess_ the intent of those explicit alignments was to align the start of arrays, not every element. But we'd need to waste time on the history of this or/and on reviewing the code's actual expectations to know this for sure. Maybe time is better spent on updating to latest upstream first.

solardiz:
> Now I wonder whether there's a performance impact (from worse locality of reference).

In my testing, yes, there is. With old RHEL6 gcc 4.4.7, AVX the impact of this PR on the raw-blake2 format is 6.3% on one thread, 19% on 31 threads (on a 32 logical CPU system, with one core currently in slight other use).

solardiz:
> In my testing, yes, there is.

Oh, sorry, I botched that test (different gcc). Re-testing now.

solardiz:
OK, no performance difference with gcc 10.2.0. Even `size john` remains the same (but file contents differ).

I think we can have this in for now.

PR #4612, opened 2021-03-09
Fix shebangs

fhuberts:
Signed-off-by: Ferry Huberts <ferry.huberts@pelagic.nl>


fhuberts:
Don't know how useful this is, but it's a patch I carry in my rpm package.
Feel free to reject if you find it not useful.

magnumripper (commented):
While dropping the use of `env` might be a good idea (I can't remember the pros and cons) this will fail if user has python2, or a future python4. So I think it should be just `python`

solardiz:
> it's a patch I carry in my rpm package.

Where did you get that patch from, and what's the rationale behind it? For example, did anything not work right without it? Thanks!

fhuberts:
I created that patch.
The Fedora packaging scripts complain about this, they can't handle the use of env.
Normally they automatically convert shebangs like /bin/bash to /usr/bin/bash.

solardiz:
> The Fedora packaging scripts complain about this, they can't handle the use of env.

Is that for their automatic generation of package dependency lists? So that they'd know this package depends on python3 and perl and bash? Maybe an explicit `Requires:` line would be more appropriate in this case?

> Normally they automatically convert shebangs like /bin/bash to /usr/bin/bash.

Maybe just let them do it rather than do it for them in a patch?

We could go with `#!/usr/bin/env bash` so that we don't depend on bash's location (but do on env's). However, like you say those packaging scripts "can't handle the use of env."

So I think we should just close this PR with no action, as merging it will make things worse for other systems.

solardiz:
We cannot reasonably accept this kind of changes tying us only to current Linux distros, so I'll close, but please feel free to post more comments if you have any. Thanks.

solardiz:
@fhuberts BTW, for your packages, _if_ you somehow have to make changes like this, I suggest you use `sed -i` instead of a patch, so that you don't have to update a patch each time we add or rename a script, etc.

exploide:
Short explanation when such changes make sense and when not. This is most relevant to the python shebangs:

Having a python shebang `#!/usr/bin/env python` runs the script with the python interpreter that may (or may not) reside inside a virtualenv. The shebang `#!/usr/bin/python` would run the script with system python, whether a virtualenv is currently active or not.

Now the question is, what is desirable? Allowing for virtualenvs makes sense for software that is expected to be run under arbitrary circumstances. E.g. a user cloned the repository, compiled it themselfes and plays a bit with the scripts (that might have crude dependencies one would like to install inside a venv). Having shebangs without env would be annoying in this case.

On the other hand, tools installed from package management should always use system python since all dependencies have been installed via package management too. Imagine running a python tool like `dnf` accidentally within a virtualenv, hell breaks loose. That's why building distribution packages highly recommend shebangs without env.

Now what is reasonable to assume for john? In my opinion, it is likely that it is run under arbitrary circumstances. Since john is actively developed but released seldom, I prefer to use bleeding-john from git.

If you agree with this, then replacing the shebang should be handled as a pre-packaging step but not within the development repository.

solardiz:
Thank you, @exploide!

> replacing the shebang should be handled as a pre-packaging step

The question is whether this is distro's pre-packaging or our creation of a source code tarball for a release. I think it should be the distro's pre-packaging step, but maybe we should provide a script to help distros do it? Just thinking out loud.

PR #4613, opened 2021-03-10
Mpi noderange bugfix

magnumripper:
Also stop us chugging along after seeing inconsistent crash recovery file (also for --fork).

### DIFF
--- a/src/john.c
+++ b/src/john.c
@@ -642,9 +642,10 @@ static void john_set_mpi(void)
 					    " %s\n", rec_name);
 			}
 		}
-		options.node_min += mpi_id * npf;
-		options.node_max = options.node_min + npf - 1;
 	}
+	options.node_min += mpi_id * npf;
+	options.node_max = options.node_min + npf - 1;
+
 	fflush(stdout);
 	fflush(stderr);
 }
--- a/src/john.c
+++ b/src/john.c
@@ -575,9 +575,7 @@ static void john_fork(void)
 				    options.node_max != save_max ||
 				    options.node_count != save_count ||
 				    options.fork != save_fork)
-					fprintf(stderr,
-					    "Inconsistent crash recovery file:"
-					    " %s\n", rec_name);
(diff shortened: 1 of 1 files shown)


magnumripper:
Looks like CircleCI is still upset with us...

solardiz:
> Looks like CircleCI is still upset with us...

More like again - it worked for a little while.

ghost:
This PR passes upstream CI (Circle and Travis). 

BTW: it seems that Circle dislike magnum more than others (maybe just a coincidence since the development activity is low).

magnumripper:
> it seems that Circle dislike magnum more than others

Yeah 😆 I was wondering why... but I don't have any other CircleCI activity anywhere I'm aware of

solardiz:
Regarding Circle CI, now it's this:

```
You have used 5,285 of 2,500 credits
5,285 used
-2,785 available
```

The smallest paid plan is $30/month for 3 users and 25k credits. I don't know if we want that, or switch to GitHub Actions, or setup something of our own.

ghost:
> or setup something of our own.

+1
*****
I have a gut feeling "rules" will change a lot (for OSS) in CI providers. 

To try to get around this, I asked for "sponsors" to implement something I have in mind (Jenkings + 2 or 3 NVIDIA drivers versions + ARM + some cloud usage). Used by at least 2 or 3 OSS. **[I'm not saying I have (or will have) a solution]**.

solardiz:
CircleCI should be working again for our Linux/Docker builds now, and moreover we should have some free macOS credits now if we like to use that. (These are not reflected in Plan Overview, but I am told should work anyway.) I think we're still very limited in credits we have for the Windows builds.

I've just enabled "Auto-cancel redundant builds" (so when there's a new push, the old builds will cancel/stop) and "Only build pull requests" (plus any commits to our default branch, but not to any other branches we might create temporarily) so that we don't abuse CircleCI's generosity.

We'll see how this works.

(And it's weird to post this to an unrelated issue, but we just happened to comment on CI here as that's when it broke recently.)

PR #4615, opened 2021-03-11
OpenCL: Improve our existing workarounds for buggy profiling data

### LINKED ISSUE #4598: Apple M1, slow OpenCL auto-tune
For some reason the auto-tune of LWS takes ages on the M1. We should look into why and if we can do something about it.

First GWS test is super-quick, the intermediate LWS tests are very slow (first size takes several seconds, larger size tests are even worse. Times shown are in the order of 290 ms). Final GWS test is quick.

A workaround could be to peg LWS to eg. 64 in john.conf.

magnumripper:
Some ancient AMD had a bug which we've looked for and worked around ever since.  Brand new Apple M1 has even worse bugs, that were partly handled by the existing workaround.  Improve the workaround and add a few more (even including comparing with wall clock time).

Closes #4598

### DIFF
--- a/src/opencl_common.c
+++ b/src/opencl_common.c
@@ -54,6 +54,7 @@
 #include "md5.h"
 #include "misc.h"
 #include "john_mpi.h"
+#include "timer.h"
 
 /* Set this to eg. 3 for some added debug and retry stuff */
 #define RACE_CONDITION_DEBUG 0
@@ -1483,13 +1484,12 @@ static void* fill_opencl_device(size_t gws, void **binary)
 // (or return zero for error or limits exceeded)
 static cl_ulong gws_test(size_t gws, unsigned int rounds, int sequential_id)
 {
-	cl_ulong startTime, endTime, runtime = 0, looptime = 0;
+	cl_ulong submitTime, startTime, endTime, runtime = 0, looptime = 0;
 	int i, count, total = 0;
 	size_t kpc = gws * ocl_v_width;
 	cl_event benchEvent[MAX_EVENTS];
 	int result, number_of_events = 0;
 	void *salt, *binary;
-	int amd_bug;
 
 	for (i = 0; i < MAX_EVENTS; i++)
 		benchEvent[i] = NULL;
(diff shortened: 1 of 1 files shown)


solardiz (approved):

solardiz:
I'm not sure what you were seeing is strictly a bug - the profiling is probably of time consumed on device only, whereas real time also includes overhead on host. Isn't it conceivable that for very low LWS/GWS the overhead on host would dominate the total time spent?

Regardless, I don't mind us having workarounds like these.

Another thought I have is whether the workarounds should possibly be sticky for the whole auto-tuning, not per kernel invocation. Maybe we should only consider enabling them if the issue is seen already on the very first kernel invocation, and then keep them enabled for all subsequent kernel invocations in the same auto-tuning run. Or are there bugs/issues that would only show up on a non-first kernel invocation?

Without the stickiness, we might get inconsistent results if we were very close to the 10x threshold implemented here or if the system load kept changing.

Anyway, while I bring the above up for discussion and for a potential future PR, I think it's good to merge this PR as-is first. Then if a later change proves problematic, we'll be able to revert to this PR's revision easily.

magnumripper:
> I'm not sure what you were seeing is strictly a bug - the profiling is probably of time consumed on device only, whereas real time also includes overhead on host. Isn't it conceivable that for very low LWS/GWS the overhead on host would dominate the total time spent?

I'm aware of that but in both cases I've seen the problem it was a bug for sure.

> Another thought I have is whether the workarounds should possibly be sticky for the whole auto-tuning, not per kernel invocation. Maybe we should only consider enabling them if the issue is seen already on the very first kernel invocation, and then keep them enabled for all subsequent kernel invocations in the same auto-tuning run. Or are there bugs/issues that would only show up on a non-first kernel invocation?
> 
> Without the stickiness, we might get inconsistent results if we were very close to the 10x threshold implemented here or if the system load kept changing.

I had some ideas like that too but I think with a 10x threshold we should be quite safe. Actually the old code did false-trigger fairly often (for xfer times so probably only cosmetical with `-v:5`) and I have yet to see the newer code do.

PR #4616, opened 2021-03-12
Single mode: Let --mkpc=N option reflect buffer allocations

magnumripper:
We always honored --mkpc for Single mode, but after some code evolution it would no longer reduce the buffer use.

Also a minor bug fix for salt resume, where we need to reflect --mkpc for optimal resume.

### DIFF
--- a/src/cracker.c
+++ b/src/cracker.c
@@ -105,7 +105,7 @@ static int process_key_stack_rules(char *key);
 /* Expose max_keys_per_crypt to the world (needed in recovery.c) */
 int crk_max_keys_per_crypt(void)
 {
-	return  crk_params->max_keys_per_crypt;
+	return options.force_maxkeys ? options.force_maxkeys : crk_params->max_keys_per_crypt;
 }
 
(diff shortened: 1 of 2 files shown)


solardiz (approved):
Looks reasonable to me. I hope this also passed tests?

magnumripper:
Yes, including with millions of salts to actually see it matter. At first I muted some warnings as well, specifically the
```c
"Note: Performance for this format/device may be lower due to single mode\n"
"      constraints. Format wanted %d keys per crypt but was limited to %d.\n",
```
...but then I decided to keep it as it might be a good heads-up even when you thought you knew what you were doing. We could conditionally change the wording slightly though, I decided not to because there's such a mess (in the code) with all those warnings, lol. But maybe I should do so, what do you think?

solardiz:
I think we could actually simplify that "Performance ... lower due to single mode ..." message - no need to refer to format/device - however, I don't have a good suggestion that would fit on one line (79 chars). So OK to leave it as-is.

PR #4618, opened 2021-03-12
Subsets bugfix & new option

magnumripper:


solardiz:
> Subsets: Add --subsets-prio-short option

FWIW, I imagine someone might want the opposite option as well, e.g. `--subsets-prio-small`.

solardiz (approved):
OK with me conceptually, without reviewing in full context.

How does this bug fix affect `--restore` of previous Subsets mode sessions, though? Isn't there a risk of skipping candidates? Should we document that if so?

magnumripper:
> OK with me conceptually, without reviewing in full context.
> 
> How does this bug fix affect `--restore` of previous Subsets mode sessions, though? Isn't there a risk of skipping candidates? Should we document that if so?

The bug could only lead to a a few subset length-comb-pairs being processed more than once, nothing missed but sometimes dupes. Stopping a buggy session and restoring with newer code will be perfectly fine regardless of where it was stopped.

magnumripper:
> > Subsets: Add --subsets-prio-short option
> 
> FWIW, I imagine someone might want the opposite option as well, e.g. `--subsets-prio-small`.

Yes... I think I'll add that as well. Oh, and I'll need to amend the docs and NEWS!

magnumripper:
All done now, testing seems fine so far

doc/NEWS
```diff
@@ -168,8 +168,10 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
 - Allow any node range divisible by fork (or MPI) count.  [Solar, magnum; 2021]
 
 - Fix autoconf selection of arch.h for Apple M1.  Improve ARM pseudo-intrinsics
-  portability (would fail on clang).  Allow __ARM_FEATURE_UNALIGNED to set our
-  ARCH_ALLOWS_UNALIGNED macro.  [magnum; 2021]
+  portability (would fail on clang).  Allow RAR formats to build on ARM where we
+  detect it can handle unaligned memory access.  [magnum; 2021]
+
+- Add a few options to Subsets mode, affecting candidate order.  [magnum; 2021]
```
solardiz: Where practical (without becoming way too verbose), I'd like our NEWS entries to be immediately useful or at least pointing to further information - so we should mention the added option names, etc.

doc/SUBSETS
```diff
@@ -18,13 +18,13 @@ resorting to a legacy code page at all.  Obviously it supports session resume
...
-lower.  Using --max-len=N may bump it to larger than 16.  It will use the full
-charset of printable ASCII (95 characters) but start with tiny subsets and
-increase from there.  By default, no more than 7 different characters from the
-charset will be used for any word even at lengths over 7 (this can be bumped,
-see SUBSET SIZES below).  In other words, by default the mode is exhaustive
-only up to length 7 if run to completion.
+subset keysize order, ending at length 32 or format's max. length, whichever is
+lower.  Using --max-len=N you can bump it to larger than 32, or decrease it.
+The full charset of printable ASCII (95 characters) will be used, starting with
+tiny subsets and increase from there.  By default, no more than 7 different
```
solardiz: Nitpick: s/increase/increasing/

doc/SUBSETS
```diff
@@ -106,6 +106,22 @@ is required, eg:
 -subsets=0123456789abcdefghijklmnopqrstuvwxyz --subsets-required=10
 
 
+CANDIDATE ORDER
+
+The keyspace is divided in many sets of (length, subset size).  By default, the
+smallest pending such set is picked, which means it will exhaust subset size 1
+first, for all lengths, so mimicing the old --external:repeats mode. Further
```
solardiz: Nitpick: I'd rather use `=` as option name to parameter value separator in documentation, especially where we use two dashes - this is the GNU long option style, then. Also, two spaces before "Further" for consistency.

doc/SUBSETS
```diff
@@ -106,6 +106,22 @@ is required, eg:
...
+into the process it will jump between lengths and subset sizes in order to
+complete smaller sets early.  While this basically produces many low-entropy
+candidates early, it doesn't strictly do so.  This can be changed by using the
+option --subsets-prio-small which will make it strictly prefer smallest subset
```
solardiz: The sentence "While this basically produces many low-entropy candidates early, it doesn't strictly do so." feels confusing rather than useful (to me, at least), and strictly speaking is wrong (entropy isn't a property of any one value). I suggest we drop it.

src/options.c
```diff
@@ -297,6 +299,8 @@ JOHN_USAGE_REGEX \
 "--subsets-min-diff=N       Minimum unique characters in subset\n" \
 "--subsets-max-diff=[-]N    Maximum unique characters in subset (negative N is\n" \
 "                           relative to word length)\n" \
+"--subsets-prio-short       Prefer shorter candidates over smaller subsets\n" \
+"--subsets-prio-small       Prefer smaller subsets over shorter candidates\n" \
```
solardiz: _Maybe_ we should use inside the option names the same word that we use in the descriptions. So e.g. either use "--subsets-prefer-short" for the option name or use the word "Prioritize" in the description. Just a thought.

solardiz (approved):

src/options.c
magnumripper: Good call, I'll change it to "prefer" everywhere.

PR #4619, opened 2021-03-13
--stdout: Speedup for -v=1 (as in "do not actually output candidates")

magnumripper:
Don't bother updating crk_stdout_key unless we're about to use it. This speeds up stdout mode considerably for testing purposes and should be completely harmless in all other cases.

The rationale for the old `-verb=1` muting stdout candidate output (since many years) was that `/dev/null` is surprisingly slow on macOS. For some testing (like mode benchmarking, or eg. watching subsets mode go through its different stages using `-log-stderr`), this additional speedup (easily another 2x) is quite useful.

### DIFF
--- a/src/cracker.c
+++ b/src/cracker.c
@@ -1161,10 +1161,12 @@ static int process_key(char *key)
 	sig_timer_emu_tick();
 #endif
 
+	if (options.verbosity > 1 || event_status)
+		strnzcpy(crk_stdout_key, key, crk_params->plaintext_length + 1);
+
 	if (event_pending && crk_process_event())
 		return 1;
 
-	strnzcpy(crk_stdout_key, key, crk_params->plaintext_length + 1);
 	if (options.verbosity > 1)
 		puts(crk_stdout_key);
 


solardiz:
> should be completely harmless in all other cases.

I doubt it. The patch looks to me like it'd (very slightly) slow down normal usage of `--stdout` where it actually does output candidates.

> The rationale for the old `-verb=1` muting stdout candidate output (since many years) was that `/dev/null` is surprisingly slow on macOS.

That rationale might be partially bogus: if `/dev/null` is slow, there are other things to try using in its place: `>&-` (a closed fd) or piping to a program like `wc -l`. Are these also somehow unsuitable on macOS?

solardiz:
Here's an alternative patch that really shouldn't have performance impact (untested):

```diff
+++ b/src/cracker.c
@@ -1161,12 +1161,16 @@ static int process_key(char *key)
        sig_timer_emu_tick();
 #endif
 
-       if (event_pending && crk_process_event())
-               return 1;
+       if (event_pending) {
+               strnzcpy(crk_stdout_key, key, crk_params->plaintext_length + 1);
+               if (crk_process_event())
+                       return 1;
+       }
 
-       strnzcpy(crk_stdout_key, key, crk_params->plaintext_length + 1);
-       if (options.verbosity > 1)
+       if (options.verbosity > 1) {
+               strnzcpy(crk_stdout_key, key, crk_params->plaintext_length + 1);
                puts(crk_stdout_key);
+       }
 
        status_update_cands(1);
 
```

solardiz (approved):
I suggest you replace this with my alternative patch, but this is up to you.

solardiz:
Potential bug, especially in your version of the patch: don't we also print status when the session terminates, without ever having `event_status` set? I think in my version, we avoid the problem for interrupted sessions by also updating on other events, but perhaps for sessions terminating normally the problem is present even with my version? Please test this more.

solardiz:
I also wonder if a loop going over the string char by char (but only once) and using `putc()` would actually be faster on some platforms. With `strnzcpy()` and `puts()`, we go over the string at least twice. (Maybe three times if `puts()` is implemented by taking a `strlen()` first.)

solardiz:
> I also wonder if a loop going over the string char by char (but only once) and using `putc()` would actually be faster

No, at least on Linux (our "super") it is actually slower in the tests I ran. Also tried `putchar()`.

solardiz:
FWIW, this is roughly same speed as before:

```diff
-       strnzcpy(crk_stdout_key, key, crk_params->plaintext_length + 1);
-       if (options.verbosity > 1)
-               puts(crk_stdout_key);
+       if (options.verbosity > 1) {
+               char *dptr = crk_stdout_key;
+               const char *sptr = key;
+               while (*sptr && sptr - key < crk_params->plaintext_length)
+                       *dptr++ = *sptr++;
+               *dptr = '\n';
+               fwrite(crk_stdout_key, dptr - crk_stdout_key + 1, 1, stdout);
+               *dptr = 0;
+       }
```

So I guess we stay with `puts()` for now.

magnumripper:
> That rationale might be partially bogus: if `/dev/null` is slow, there are other things to try using in its place: `>&-` (a closed fd) or piping to a program like `wc -l`. Are these also somehow unsuitable on macOS?

Tried `>&-` and it was even (much) worse. Tried the `wc -l` and it was a bit slower than `>/dev/null`.

Considering the bug you identified, this is not that important really, maybe I should just forget about it.

solardiz:
So when the session running with `-verb=1` terminates normally, it prints the candidate password that was current as of the last event.

Another likely bug: I think `ext_status` doesn't raise `event_pending` (nor `event_status`), yet prints the last recorded candidate password from further in the same function.

magnumripper:
I'll just close this, thanks for looking into it!

PR #4620, opened 2021-03-14
Intrinsics: Fix strict-aliasing violations

### LINKED ISSUE #4446: Strict aliasing violation warnings for simd-intrinsics.c in SSE2 build
In an `--enable-simd=sse2` build made with `gcc version 4.4.7 20120313 (Red Hat 4.4.7-23)`:

```
simd-intrinsics.c: In function 'SIMDSHA1body':
simd-intrinsics.c:1368: warning: dereferencing pointer 'p' does break strict-aliasing rules
simd-intrinsics.c:1360: note: initialized from here
simd-intrinsics.c:1373: note: initialized from here
simd-intrinsics.c:1378: note: initialized from here
simd-intrinsics.c:1368: note: initialized from here
simd-intrinsics.c:1373: warning: dereferencing pointer 'p' does break strict-aliasing rules
simd-intrinsics.c:1360: note: initialized from here
simd-intrinsics.c:1368: note: initialized from here
simd-intrinsics.c:1378: note: initialized from here
simd-intrinsics.c:1373: note: initialized from here
simd-intrinsics.c:1378: warning: dereferencing pointer 'p' does break strict-aliasing rules
simd-intrinsics.c:1360: note: initialized from here
simd-intrinsics.c:1368: note: initialized from here
simd-intrinsics.c:1373: note: initialized from here
simd-intrinsics.c:1378: note: initialized from here
simd-intrinsics.c: In function 'SIMDmd4body':
simd-intrinsics.c:853: warning: dereferencing pointer 'p' does break strict-aliasing rules
simd-intrinsics.c:845: note: initialized from here
simd-intrinsics.c:858: note: initialized from here
simd-intrinsics.c:863: note: initialized from here
simd-intrinsics.c:853: note: initialized from here
simd-intrinsics.c:858: warning: dereferencing pointer 'p' does break strict-aliasing rules
simd-intrinsics.c:845: note: initialized from here
simd-intrinsics.c:853: note: initialized from here
simd-intrinsics.c:863: note: initialized from here
simd-intrinsics.c:858: note: initialized from here
simd-intrinsics.c:863: warning: dereferencing pointer 'p' does break strict-aliasing rules
simd-intrinsics.c:845: note: initialized from here
simd-intrinsics.c:853: note: initialized from here
simd-intrinsics.c:858: note: initialized from here
simd-intrinsics.c:863: note: initialized from here
simd-intrinsics.c: In function 'SIMDmd5body':
simd-intrinsics.c:168: warning: dereferencing pointer 'p' does break strict-aliasing rules
simd-intrinsics.c:160: note: initialized from here
simd-intrinsics.c:173: note: initialized from here
simd-intrinsics.c:178: note: initialized from here
simd-intrinsics.c:168: note: initialized from here
simd-intrinsics.c:173: warning: dereferencing pointer 'p' does break strict-aliasing rules
simd-intrinsics.c:160: note: initialized from here
simd-intrinsics.c:168: note: initialized from here
simd-intrinsics.c:178: note: initialized from here
simd-intrinsics.c:173: note: initialized from here
simd-intrinsics.c:178: warning: dereferencing pointer 'p' does break strict-aliasing rules
simd-intrinsics.c:160: note: initialized from here
simd-intrinsics.c:168: note: initialized from here
simd-intrinsics.c:173: note: initialized from here
simd-intrinsics.c:178: note: initialized from here
```

I haven't looked into these yet.

magnumripper:
You can cast anything to a vtype (provided it's aligned) but you can not do the reverse.  See #4446

### DIFF
--- a/src/simd-intrinsics.c
+++ b/src/simd-intrinsics.c
@@ -3,7 +3,7 @@
  * Copyright (c) 2010 bartavelle, <bartavelle at bandecon.com>,
  * Copyright (c) 2012 Solar Designer,
  * Copyright (c) 2011-2015 JimF,
- * Copyright (c) 2011-2018 magnum,
+ * Copyright (c) 2011-2021 magnum,
  * and it is hereby released to the general public under the following terms:
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted.
@@ -115,8 +115,7 @@ void md5_unreverse(uint32_t *hash)
 
 #undef INIT_A
 
-void SIMDmd5body(vtype* _data, unsigned int *out,
-                uint32_t *reload_state, unsigned SSEi_flags)
+void SIMDmd5body(uint32_t* _data, uint32_t *out, uint32_t *reload_state, unsigned int SSEi_flags)
 {
 	vtype w[16*SIMD_PARA_MD5];
 	vtype a[SIMD_PARA_MD5];
@@ -136,11 +135,11 @@ void SIMDmd5body(vtype* _data, unsigned int *out,
 #endif
 
 	if (SSEi_flags & SSEi_FLAT_IN) {
-		// Move _data to __data, mixing it SIMD_COEF_32 wise.
+		// Move _data to data, mixing it SIMD_COEF_32 wise.
 #if __SSE4_1__ || __MIC__
-		unsigned k;
+		unsigned int k;
 		vtype *W = w;
-		uint32_t *saved_key = (uint32_t*)_data;
+		uint32_t *saved_key = _data;
 		MD5_PARA_DO(k)
 		{
 			if (SSEi_flags & SSEi_4BUF_INPUT) {
@@ -156,10 +155,10 @@ void SIMDmd5body(vtype* _data, unsigned int *out,
 			W += 16;
 		}
 #else
-		unsigned j, k;
+		unsigned int j, k;
 		uint32_t *p = (uint32_t*)w;
 		vtype *W = w;
-		uint32_t *saved_key = (uint32_t*)_data;
+		uint32_t *saved_key = _data;
 		MD5_PARA_DO(k)
 		{
 			if (SSEi_flags & SSEi_4BUF_INPUT) {
@@ -194,7 +193,7 @@ void SIMDmd5body(vtype* _data, unsigned int *out,
 		// now set our data pointer to point to this 'mixed' data.
 		data = w;
 	} else
-		data = _data;
+		data = (vtype*)_data;
 
 	if (!(SSEi_flags & SSEi_RELOAD))
 	{
@@ -356,7 +355,7 @@ void SIMDmd5body(vtype* _data, unsigned int *out,
 	if (SSEi_flags & SSEi_FLAT_OUT) {
 		MD5_PARA_DO(i)
 		{
-			uint32_t *o = (uint32_t*)&out[i*4*VS32];
+			uint32_t *o = &out[i*4*VS32];
 #if __AVX512F__ || __MIC__
 			vtype idxs = vset_epi32(15*4,14*4,13*4,12*4,
 			                        11*4,10*4, 9*4, 8*4,
@@ -584,7 +583,7 @@ static MAYBE_INLINE void dispatch(unsigned char buffers[8][64*MD5_SSE_NUM_KEYS],
 				mmxput3(buffers, bufferid, length, 2, saltlen, f);
 				break;
 		}
-		SIMDmd5body((vtype*)&buffers[bufferid], f, NULL, SSEi_MIXED_IN);
+		SIMDmd5body((uint32_t*)&buffers[bufferid], f, NULL, SSEi_MIXED_IN);
 		if (j++ < 1000 % 42 - 1)
 			continue;
 		if (j == 1000 % 42) {
@@ -804,8 +803,7 @@ void md4_unreverse(uint32_t *hash)
 #undef INIT_B
 #undef INIT_A
 
-void SIMDmd4body(vtype* _data, unsigned int *out, uint32_t *reload_state,
-                unsigned SSEi_flags)
+void SIMDmd4body(uint32_t* _data, uint32_t *out, uint32_t *reload_state, unsigned int SSEi_flags)
 {
 	vtype w[16*SIMD_PARA_MD4];
 	vtype a[SIMD_PARA_MD4];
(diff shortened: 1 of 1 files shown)


magnumripper:
I did not test this with gcc 4.4.7 but it should do the trick. I got no warnings before this and obviously none after. Using gcc 7.5.0 I got an *identical* `simd-intrinsics.o` after stripping so I'm sure nothing stupid slinked in.

magnumripper:
Force-pushed to fix Copyright years. I have another really no-op patch that changes all `unsigned ..` to `uint_..t` but have yet to understand why it changes the the stripped result just a tad so I won't push it just yet:
```diff
 000000000000b450 _sha256_unreverse:
 ;       fprintf(stderr, "sha256_unreverse() not implemented\n");
     b450: 48 8b 05 00 00 00 00          movq    (%rip), %rax
 ; {
     b457: 55                            pushq   %rbp
 ;       fprintf(stderr, "sha256_unreverse() not implemented\n");
     b458: be 01 00 00 00                movl    $1, %esi
     b45d: ba 23 00 00 00                movl    $35, %edx
     b462: 48 8d 3d 00 00 00 00          leaq    (%rip), %rdi
 ; {
     b469: 48 89 e5                      movq    %rsp, %rbp
 ;       fprintf(stderr, "sha256_unreverse() not implemented\n");
     b46c: 48 8b 08                             movq    (%rax), %rcx
     b46f: e8 00 00 00 00                       callq   0 <_sha256_unreverse+0x24>
 ;      error();
-    b474: be 4c 07 00 00                       movl    $1868, %esi
+    b474: be 48 07 00 00                       movl    $1864, %esi
     b479: 48 8d 3d 00 00 00 00                 leaq    (%rip), %rdi
     b480: e8 00 00 00 00                       callq   0 <_sha256_unreverse+0x35>
     b485: 66 2e 0f 1f 84 00 00 00 00 00        nopw    %cs:(%rax,%rax)
     b48f: 90                            nop

```

magnumripper:
Found a few more such aliasing violations, fixed them in a separate commit. This *did* change the resulting code in some way from SHA-1 and on.

magnumripper:
> I have another really no-op patch that changes all `unsigned ..` to `uint_..t` but have yet to understand why it changes the the stripped result just a tad so I won't push it just yet

Pushed a third patch that does this cosmetical change. Now for some serious testing, mostly about the 2nd patch of the three because there appears to be a slight regression.

magnumripper:
Second patch is buggy. Reverting and reworking. I will instead start with a patch that refactors all types to stdint ones, then start over. Easier to review.

solardiz:
> You can cast anything to a vtype (provided it's aligned)

What makes you think so?

solardiz:
Ah, I see you're probably referring to the explanation you quoted in https://github.com/openwall/john/issues/4446#issuecomment-798803339 - that's x86-specific, but so is the immediate issue you're trying to fix here.

Ideally, we'd make no assumptions about permitted aliasing of scalar and vector types on any archs (use proper intrinsics or union types for conversions), but we can approach this separately. We also have similar issues affecting non-x86 in `DES_bs_b.c` - everything works anyway so far, but some non-x86 builds did produce warnings there (so we use `-fno-strict-aliasing` for that file on some archs).

magnumripper:
Yes it's x86-specific. And the `union` solution is gcc-specific, although that may be less of a problem IRL. As far as I understand there isn't really *any* way to make it 100% portable to "anything" - some say the only thing always allowed is using `memcpy` and beg it gets optimized away but as far as I understand that is still not guaranteed to *work* (as opposed to *be allowed*) on any and all architectures and I really don't like the possibility of `memcpy`'s not being optimized away.

I'm still considering using unions. On ARM we have (effectively) this:
```c
typedef union {
    uint8x16_t v8;
    uint32x4_t v32;
    uint64x2_t v64;
} vtype;
```
What if we'd add to it like this?
```c
typedef union {
    uint32_t u32;
    uint64_t u64;
    uint8x16_t v8;
    uint32x4_t v32;
    uint64x2_t v64;
} vtype;
```
and similar unions for all archs, such as:
```diff
- typedef __m256i vtype;
+ typedef union {
+     uint32_t u32;
+     uint64_t u64;
+     __m256i v;
+ } vtype;
```
...then use this `vtype` union everywhere needed in `simd-intrinsics.c`. Would that be better? It would keep union definitions out of `simd-intrinsics.c` and make the code more readable (as far as you "know" our pseudo-intrinsics).

solardiz:
> And the `union` solution is gcc-specific

In what way is it gcc-specific? Do you mean transparent unions? I didn't suggest we use that. https://gcc.gnu.org/onlinedocs/gcc/Unnamed-Fields.html

I suggested we simply use unions, explicitly referring to their fields in code. While it is not guaranteed they can be used to exchange data across types, this behavior is available in practice in all compilers I know, and we already rely on it.

>  use this `vtype` union everywhere needed in `simd-intrinsics.c`. Would that be better?

Yes, I think so - but we'll need to revise the code accordingly. Also, you meant arrays of `uint32_t` there, etc. - not individual values.

Anyway, this is probably an invasive enough change we don't want to make it just yet if we're trying to make a release soon.

solardiz (commented):
I think it's only desirable to change `unsigned int` to explicit `uint32_t` in places where that is in fact the assumed size - perhaps when these scalar values are then put into vector elements of this size. In other cases where it's just loop control variables, etc. I think it's better to stay with `unsigned int`.

magnumripper:
> that's x86-specific, but so is the immediate issue you're trying to fix here.

For what it's worth, this PR as-is right now doesn't add any warnings or problems with ARM (tested with clang and gcc). But the main question is: Does this mute the warnings from gcc 4.4.7 at all?

solardiz:
> Does this mute the warnings from gcc 4.4.7 at all?

Why don't you just test?

magnumripper:
> > Does this mute the warnings from gcc 4.4.7 at all?
> 
> Why don't you just test?

Not quickly done here but I was going to have a look at the village... do you happen to know it's available there?

solardiz:
> I was going to have a look at the village... do you happen to know it's available there?

Yes, it's the default RHEL6 gcc. But you need to force SSE2 build to see the warnings. Please reproduce the warnings first, then see if the changes fix them. Thanks!

magnumripper:
> > I was going to have a look at the village... do you happen to know it's available there?
> 
> Yes, it's the default RHEL6 gcc. But you need to force SSE2 build to see the warnings. Please reproduce the warnings first, then see if the changes fix them. Thanks!

Did so, and no, nothing changed. Back to the drawing board.

PR #4625, opened 2021-03-15
Intrinsics aliasing

magnumripper:
#4446. First commit fixes the problems with gcc-4.4.7 but give SHA-1 a 5% performance regression using gcc-10. Second commit is uglier but mitigates the regression.

I'll test the two versions on some other compilers/versions. I'd rather not merge the second version if this is specific to gcc-10.

### DIFF
--- a/src/simd-intrinsics.c
+++ b/src/simd-intrinsics.c
@@ -3,7 +3,7 @@
  * Copyright (c) 2010 bartavelle, <bartavelle at bandecon.com>,
  * Copyright (c) 2012 Solar Designer,
  * Copyright (c) 2011-2015 JimF,
- * Copyright (c) 2011-2018 magnum,
+ * Copyright (c) 2011-2021 magnum,
  * and it is hereby released to the general public under the following terms:
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted.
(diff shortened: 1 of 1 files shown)


magnumripper:
Using clang 12.0.0, either version of the patch produce the same code, both a little different from the original code but total length of disassembly is about the same (+/- 1).

Using gcc 7.5.0, using the short patch produces 72 lines less of code, while the longer patch produce the same code as the original.

Using gcc 10.2.0, **I now realized my previous tests were flawed (different directories and something else changed). A proper test result is that the short patch produces 12 lines less code and the longer patch gets it back to original.**

magnumripper:
> Using gcc 10.2.0, I now realized my previous tests were flawed (different directories and something else changed). A proper test result is that the short patch produces 12 lines less code and the longer patch gets it back to original.

Same for the benchmarks of pbkdf2-hmac-sha1: The simpler patch is actually faster, not slower. I force-pushed the 2nd one away, this can be merged!

ghost:
> The simpler patch is actually faster, not slower

Makes more sense.

solardiz:
Good stuff.

Maybe try changing `vtype *w = uw.vec;` to `vtype * const w = uw.vec;`

In `uint32_t u32[1];` maybe better use actual size in place of `1`. While array out-of-bounds access within the same `union` is defined behavior, I wonder if it could be tripping some warnings from compilers or static analysis tools.

(Alternatively, there's C99 https://en.wikipedia.org/wiki/Flexible_array_member but it's a feature normally for use in `struct`, not `union`, so let's not experiment with it here.)

solardiz (approved):

magnumripper:
> Maybe try changing `vtype *w = uw.vec;` to `vtype * const w = uw.vec;`

Tried it, didn't seem to affect the resulting code at all. Decided not to commit that for now.

> In `uint32_t u32[1];` maybe better use actual size in place of `1`. While array out-of-bounds access within the same `union` is defined behavior, I wonder if it could be tripping some warnings from compilers or static analysis tools.

We use the size of 1 in many places already and it's very common practice so I decided to leave that alone too for now.

Issue #4532, opened 2021-01-03
sspr-opencl misaligned binary()


solardiz:
As seen in https://github.com/openwall/john/issues/4530#issuecomment-753686542:

```
Benchmarking: sspr-opencl, NetIQ SSPR / Adobe AEM [MD5/SHA1/SHA2 OpenCL]... Warning: binary() returned misaligned pointer
DONE
```

This is because `opencl_sspr_fmt_plug.c` wrongly has:

```c
#define BINARY_ALIGN            MEM_ALIGN_WORD
```

whereas the code only guarantees alignment appropriate for `uint32_t`:

```c
void *sspr_get_binary(char *ciphertext)
{
	static union {
		unsigned char c[BINARY_SIZE];
		uint32_t dummy;
	} buf;
	unsigned char *out = buf.c;
	char *p;
	int i;

	memset(buf.c, 0, BINARY_SIZE);
	p = strrchr(ciphertext, '$') + 1;
	for (i = 0; i < BINARY_SIZE_MIN; i++) {
		out[i] = (atoi16[ARCH_INDEX(*p)] << 4) | atoi16[ARCH_INDEX(p[1])];
		p += 2;
	}

	return out;
}
```

The CPU format gets this right:

```
#define BINARY_ALIGN            sizeof(uint32_t)
#define SALT_SIZE               sizeof(struct custom_salt)
#define SALT_ALIGN              sizeof(uint32_t)
```

We probably need to move all 3 of these to `sspr_common.h` and remove them from the two formats.

Issue #3677, opened 2019-03-22
32-bit build warnings [salt() returned misaligned pointer]


claudioandre-br:
```
Testing: AndroidBackup [PBKDF2-SHA1 128/128 NEON 4x AES]... (4xOMP) Warning: salt() returned misaligned pointer
PASS
Testing: andOTP [SHA256 32/32]... (4xOMP) Warning: salt() returned misaligned pointer
PASS
Testing: bitshares, BitShares Wallet [SHA-512 64/32]... (4xOMP) Warning: salt() returned misaligned pointer
PASS
Testing: radius, RADIUS authentication [MD5 32/32]... (4xOMP) Warning: salt() returned misaligned pointer
PASS
```

A fix can be seen in f7132b346c43443b5ac63372bd57b647d3332d3f

solardiz:
I think f7132b346c43443b5ac63372bd57b647d3332d3f is more of a workaround than a fix. Ideally, we'd define the types such that the compiler provides proper alignment without us having to ask for it. Perhaps this would involve use of `union` types. We'd also avoid strict aliasing rules violations then, which I guess we have now.

magnumripper:
These were fixed in #3934

PR #4626, opened 2021-03-15
Fix misaligned pointer warnings

magnumripper:
Both had SALT_ALIGN fixed to 8 where it should be sizeof(pointer). I found no issues for them so fixed them rather than create issues.

### DIFF
--- a/src/androidfde_fmt_plug.c
+++ b/src/androidfde_fmt_plug.c
@@ -62,8 +62,8 @@ john_register_one(&fmt_fde);
 #define BENCHMARK_LENGTH    0x107
 #define BINARY_SIZE         0
 #define BINARY_ALIGN        1
-#define SALT_ALIGN          8
 #define SALT_SIZE           sizeof(struct custom_salt)
+#define SALT_ALIGN          sizeof(void*)
 #ifdef SIMD_COEF_32
 #define MIN_KEYS_PER_CRYPT  SSE_GROUP_SZ_SHA1
 #define MAX_KEYS_PER_CRYPT  (SSE_GROUP_SZ_SHA1 * 4)
@@ -86,8 +86,8 @@ static int *cracked;
 static int max_cracked;
 
 static struct custom_salt {
-	int loaded;
 	unsigned char *cipherbuf;
+	int loaded;
 	int keysize;
 	int iterations; // NOTE, not used. Hard coded to 2000 for FDE from Android <= 4.3 (PBKDF2-SHA1)
 	int saltlen;
@@ -95,7 +95,6 @@ static struct custom_salt {
 	unsigned char salt[16];
 	unsigned char mkey[64];
 	unsigned char iv[16];
-
 } *cur_salt;
 
 static void init(struct fmt_main *self)
--- a/src/opencl_pbkdf2_hmac_sha512_fmt_plug.c
+++ b/src/opencl_pbkdf2_hmac_sha512_fmt_plug.c
@@ -33,9 +33,9 @@ john_register_one(&fmt_opencl_pbkdf2_hmac_sha512);
 #define FORMAT_NAME              "GRUB2 / OS X 10.8+"
 #define ALGORITHM_NAME           "PBKDF2-SHA512 OpenCL"
 #define BINARY_ALIGN             8
-#define SALT_ALIGN               8
 #define PLAINTEXT_LENGTH         110
 #define SALT_SIZE                sizeof(salt_t)
+#define SALT_ALIGN               sizeof(uint64_t)
 #define KERNEL_NAME              "pbkdf2_sha512_kernel"
 #define SPLIT_KERNEL_NAME        "pbkdf2_sha512_loop"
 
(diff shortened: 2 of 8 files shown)


src/androidfde_fmt_plug.c
```diff
@@ -62,7 +62,7 @@ john_register_one(&fmt_fde);
 #define BENCHMARK_LENGTH    0x107
 #define BINARY_SIZE         0
 #define BINARY_ALIGN        1
-#define SALT_ALIGN          8
+#define SALT_ALIGN          sizeof(struct custom_salt*)
 #define SALT_SIZE           sizeof(struct custom_salt)
```
solardiz, line 65: I think it's confusing to use `sizeof(struct custom_salt*)` here because that's not the pointer type found _in_ the struct. We actually need alignment because of the `unsigned char *cipherbuf;` field here (BTW, the struct could be made smaller by re-ordering fields, not keeping this pointer between `int` fields). So I think it's better to use `sizeof(void *)` or `sizeof(unsigned char *)`.

src/opencl_pbkdf2_hmac_sha512_fmt_plug.c
```diff
@@ -33,9 +33,9 @@ john_register_one(&fmt_opencl_pbkdf2_hmac_sha512);
 #define FORMAT_NAME              "GRUB2 / OS X 10.8+"
 #define ALGORITHM_NAME           "PBKDF2-SHA512 OpenCL"
 #define BINARY_ALIGN             8
-#define SALT_ALIGN               8
 #define PLAINTEXT_LENGTH         110
 #define SALT_SIZE                sizeof(salt_t)
+#define SALT_ALIGN               sizeof(salt_t*)
```
solardiz: I see no pointer inside the struct. I think 8 was actually right, but we can use `sizeof(uint64_t)` for clarity.

Was there any issue that prompted you to make a change here? I wonder if there's something I overlook.

solardiz (changes requested):

src/opencl_pbkdf2_hmac_sha512_fmt_plug.c
magnumripper: > Was there any issue that prompted you to make a change here? I wonder if there's something I overlook.

No I only saw the warning from fde on a 32-bit build, but looked for others that had a fixed 8. This was the only one and like I said I thought it looked wrong.

> I see no pointer inside the struct. I think 8 was actually right, but we can use `sizeof(uint64_t)` for clarity.

Oh, I'm always confused by these things, that's why I shouldn't try to fix these things. You are obviously right.

magnumripper:
Now that I'm set straight I'll try to address the other ones we have and amend their fixes.
#3677, #4532

solardiz (approved):
The revised changes look correct to me.

solardiz (approved):
The changes make sense to me. I wonder how the pbkdf2-hmac-sha512 format worked on archs requiring alignment when its BINARY_ALIGN was only 4, so any copies of its binary wouldn't be 8-byte aligned half of the time. Perhaps this suggests we don't currently rely on this alignment in this format's `cmp_all()` and `cmp_one()`, at least not on those archs? I didn't check. Anyway, let's have the alignment.

This commit message is subtly wrong:

> Anything using ARCH_WORD is suspect, looks like it was often mistaken
for normally being 32-bit (well it was, in the old days)

`ARCH_WORD` never implied exactly 32-bit, not even "in the old days" pre-dating jumbo. The very reason I introduced `ARCH_WORD` was to have a type for the machine's native word size (usually same as `long`, but not necessarily), and JtR already ran on some 64-bit platforms back then (1990s).

I suggest you keep only the "Anything using ARCH_WORD is suspect" portion.

magnumripper:
> The changes make sense to me. I wonder how the pbkdf2-hmac-sha512 format worked on archs requiring alignment when its BINARY_ALIGN was only 4, so any copies of its binary wouldn't be 8-byte aligned half of the time. Perhaps this suggests we don't currently rely on this alignment in this format's `cmp_all()` and `cmp_one()`, at least not on those archs? I didn't check. Anyway, let's have the alignment.

IIRC, only the OpenCL format would actually use `uint64_t` in `cmp_*()` while the CPU format uses `memcmp()`. Non-x86 archs running OpenCL is uncommon (until now with the Silicon M1, but even that is mostly capable of unaligned now).

> This commit message is subtly wrong:
> 
> > Anything using ARCH_WORD is suspect, looks like it was often mistaken
> > for normally being 32-bit (well it was, in the old days)
> 
> `ARCH_WORD` never implied exactly 32-bit, not even "in the old days" pre-dating jumbo. The very reason I introduced `ARCH_WORD` was to have a type for the machine's native word size (usually same as `long`, but not necessarily), and JtR already ran on some 64-bit platforms back then (1990s).

What I meant with "normally" is ten years ago 32-bit was still very common (especially for Windows, so for Jim) and `ARCH_WORD` would often end up as 32. Surely it was used by mistake here. I can't see a reason to change the wording.

magnumripper:
Rebased to keep the pbkdf2-hmac* stuff in a single commit, so I did change the wording now ;-)

PR #4628, opened 2021-03-15
krb5-pa-md5-opencl: Bugfix for internal mask.

### LINKED ISSUE #4529: Some internal mask formats b0rken

magnumripper:
Closes #4529

### DIFF
--- a/src/opencl_krb5pa-md5_fmt_plug.c
+++ b/src/opencl_krb5pa-md5_fmt_plug.c
@@ -314,6 +314,7 @@ static void done(void)
 		HANDLE_CLERROR(clReleaseKernel(crypt_kernel), "Release kernel.");
 		HANDLE_CLERROR(clReleaseProgram(program[gpu_id]), "Release Program.");
 		crypt_kernel = NULL;
+		program[gpu_id] = NULL;
 	}
 
 	if (loaded_hashes)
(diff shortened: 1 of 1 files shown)


ghost (approved):
LGTM (after some testing).

Issue #3514, opened 2018-12-16
bash completion needs adjustments


magnumripper:
* `--test-full=LEVEL` is missing
* `--rules-stack=` is missing
* `--rules=` expansion is now broken
* `--single-seed=` and `--single-wordlist=` similarly broke `--single=` expansion

Ideally some general logic fix can solve most of this. Jumbo's getopt now doesn't count `-si` as ambiguous even though there are `--single-seed=` and `--single-wordlist=`, but as soon as you write `--single-` it will demand either an `s` or a `w` for unambiguousity.

I'm hoping for Frank here but in a pinch I could probably sort it out.

magnumripper:
@frank-dittrich this is quite important for the release. Will you have time to look into it?

frank-dittrich:
@magnumripper I know it is important. Unfortunately, I'll not have much time.
And, last time I looked into this problem, I didn't see an easy solution.
May be, there is one.

frank-dittrich:
`--test` and `--test-full` have the same problem.

This patch at least "helps" for `--test-full`, but solving the problem for the shorter option names is far more important.

```
diff --git a/run/john.bash_completion b/run/john.bash_completion
index 369e1667d..683d62441 100644
--- a/run/john.bash_completion
+++ b/run/john.bash_completion
@@ -215,6 +215,7 @@ _john()
 #	Just those options that can be used together with a value,
 #	even if that value is optional:
 	valopts=`echo "$options"|grep '='|sed 's#^ *\([a-z=-]*\).*$#\1#'`
+
 #	This is used to decide whether or not the completion should add
 #	a trailing space.
 #	(That means, for a jumbo build, --rules doesn't get a trailing space,
@@ -611,6 +612,10 @@ _john()
 			COMPREPLY=( $(compgen -W "--test --test=SECONDS" -- ${cur}) )
 			return 0
 			;;
+		--test-full)
+			COMPREPLY=( $(compgen -W "--test-full --test-full=SECONDS" -- ${cur}) )
+			return 0
+			;;
 		--show)
 			COMPREPLY=( $(compgen -W "--show --show=" -- ${cur}) )
 			return 0
```

frank-dittrich:
The problem is that for the shorter option name (rules, test, etc.) I never arrive at this statement in line 323:
```
	case "${option}" in
```
because $(option) only gets filled if for the string on the command line there remains only one valid option name.

frank-dittrich:
[...]

PR #4630, opened 2021-03-16
Fix bash completion for john and unique

frank-dittrich:
support completion for --single=, --rules=, etc. after --single-* and
--rules-* options have been added

fix --fuzz-dump= completion

adjust completion for --salts=, --verbosity=, --bare-always-valid=

add completion for --rules-stack=, --prince-loopback=, --tune=

adjust completion for unique

### DIFF
--- a/run/john.bash_completion
+++ b/run/john.bash_completion
@@ -1,6 +1,6 @@
 # bash completion for john and unique commands (John the Ripper)
 #
-# This software is Copyright (c) 2012 Frank Dittrich
+# This software is Copyright (c) 2012-2021 Frank Dittrich
 # and hereby released to the general public under the following terms:
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted.
@@ -97,7 +97,7 @@ shopt -s extglob
 ## have grep && have sed && have tr &&
 _john()
 {
-	local first cur options valopts compreplya compreplyb encodings formats format_classes
+	local first cur options valopts compreplya compreplyb encodings formats format_classes help
 	local primitives subformats list dir cmd cmd2 i ver ver1 ver2 ver3 prev words prefix option
 
 	# Without LC_ALL=C, [A-Z] match [a-z] (case "${cur}" in ... esac)
@@ -156,7 +156,14 @@ _john()
 #   Some are listed like --[no-]keep-guessing, expanded to both variants
 #
 #	All options:
-	options="`{ ${first} --help 2>/dev/null | sed -n -e 's#^ *\(--\(\[no-\]\)\?[a-z-]*\(\[\)\?=\?\(LIST\)\?\).*$#\1#' -e 's#^--\[no-\]\(.*\)#--\1\n--no-\1#' -e '/^--/ p' ; echo --stdin; } | sort -u | grep -v -- '--show='`"
+	if [[ "_${ver1}" == "_" ]]; then
+		# jumbo
+		help="--help"
+	else
+		# core
+		help=""
+	fi
+	options="`{ ${first} ${help} 2>/dev/null | sed -n -e 's#^ *\(--\(\[no-\]\)\?[a-z-]*\(\[\)\?=\?\(LIST\)\?\).*$#\1#' -e 's#^--\[no-\]\(.*\)#--\1\n--no-\1#' -e '/^--/ p' ; echo --stdin; } | sort -u | grep -v -- '--show='`"
 
 	if [[ "_${options}" == "_--stdin" && "_${__john_binary}" != "_" ]] ; then
 		# No options and hidden(deprecated) options with ${first}, but a previous
@@ -246,14 +253,7 @@ _john()
 	# slower than the current approach
 
 	# Check if completion is requested for a valid option
-	# --prince= is a hack for now, should be made generic:
-	if [[ "${cur}" == --prince[=:]* || "${cur}" == -prince[=:]* ]]; then
-		COMPREPLY=("--prince")
-	# The same hack is needed for --fuzz as well, because we have
-	# --fuzz[=something] and --fuzz-dump[=something]
-	elif [[ "${cur}" == --fuzz[=:]* || "${cur}" == -fuzz[=:]* ]]; then
-		COMPREPLY=("--fuzz")
-	elif [[ "${cur}" == --* ]] ; then
+	if [[ "${cur}" == --* ]] ; then
 		COMPREPLY=( $(compgen -W "${options}" -- "${cur%%[=:]*}") )
 	else
 		COMPREPLY=( $(compgen -W "${options}" -- "-${cur%%[=:]*}") )
@@ -262,23 +262,24 @@ _john()
 	# Check the number of different possible completions
 	# for the (possibly abbreviated) option
 	i=${#COMPREPLY[*]}
-	if [[ $i -eq 1 ]] ; then
-		# (abbreviated) option is not ambiguous
+	IFS=$'\n' COMPREPLY=($(sort <<<"${COMPREPLY[*]}"))
+	unset IFS
+	if [[ $i -ne 0 && ${COMPREPLY[$i - 1]} == ${COMPREPLY[0]%%=*}* ]]; then
+		# (abbreviated) option is not ambiguous or
+		# option is also part of a longer option name
 		if [[ "${cur}" != *[=:]* ]] ; then
 			# No value specified.
 			if [[ $valopts == *${COMPREPLY[0]}* ]] ; then
 				# Option allows a value
 				if [[ "${cur}" == ${COMPREPLY[0]} ]] ; then
-					# Option has not been abbreviated
-					if [[ "_${__john_completion}" == "_2" ]]; then
+					if [[ $i -eq 1 && "_${__john_completion}" == "_2" ]] ; then
 						# User specific setting requests adding a '='
 						COMPREPLY=( $(compgen -W "${cur}=" -- ${cur}) )
 						compopt -o nospace
 						return
-					else
-						# Completion might depend on the option
-						option="${cur}"
 					fi
+					COMPREPLY=( $(compgen -W "${options} ${cur}=" -- ${cur}) )
+					return
 				else
 					# Option has been abbreviated
 					# make sure no trailing space gets added
@@ -435,7 +436,7 @@ _john()
 			done
 			return 0
 			;;
-		--wordlist=*|--mkv-stats=*|--prince=*|--fuzz=*|--fuzz-dump=*|--single-wordlist=*)
+		--wordlist=*|--mkv-stats=*|--prince=*|--fuzz=*|--single-wordlist=*)
 			# as long as Markov stats files can have any name,
 			# I can just reuse the --wordlist= logic for --mkv-stats=*
 			cur=${cur#*[=:]}
@@ -443,7 +444,7 @@ _john()
 			_filedir
 			return 0
 			;;
- 		--rules|--single)
+		--rules|--rules-stack|--single)
 			cmd=`echo ${COMP_LINE}|sed "s# ${cur}# --list=rules #"`
 			list=`${cmd} 2>/dev/null`
 			if [[ $? -ne 0 ]] ; then
@@ -457,7 +458,7 @@ _john()
 			COMPREPLY=( $(compgen -W "${list}" -- ${cur}) )
 			return 0
 			;;
-		--rules=*|--single=*)
+		--rules=*|--rules-stack=*|--single=*)
 			cmd=`echo ${COMP_LINE}|sed "s# ${cur}# --list=rules #"`
 			list=`${cmd} 2>/dev/null`
 			if [[ $? -ne 0 ]] ; then
@@ -622,7 +623,8 @@ _john()
 			;;
 		--show=?(*))
 			cur=`echo ${cur#*[=:]} | tr A-Z a-z`
-			COMPREPLY=( $(compgen -W "left formats types invalid" -- ${cur}) )
+			list=`${first} --show=help dummy 2>&1 | grep -v "^Invalid" | sed 's#--show=\?\([^,]*\),\?#\1#g' `
+			COMPREPLY=( $(compgen -W "${list}" -- ${cur}) )
 			return 0
 			;;
 		--users=?(-)+(L|U)*|--groups=+(-|G)*|--shells=+(-|S)*|--salts=+(-|C)*)
@@ -645,7 +647,7 @@ _john()
 			;;
 		--salts=*)
 			cur=${cur#*=}
-			COMPREPLY=( $(compgen -W "COUNT -COUNT" -- ${cur}) )
+			COMPREPLY=( $(compgen -W "COUNT -COUNT #M #M-N" -- ${cur}) )
 			return 0
 			;;
 		--encoding=*|--intermediate-enc=*|--internal-codepage=*|--internal-encoding=*|--target-encoding=*|--input-encoding=*)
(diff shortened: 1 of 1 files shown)


magnumripper:
Issue is #3514 

I tried this and still don't get any rules listed if I say `--rules=` and then a TAB. Was that supposed to work now? Or did I somehow not load the new file correctly (I think I did).

frank-dittrich:
Maybe you forgot `. john.bash_completion` to update the bash completion logic?
The completion for `--rules=` worked for me.
But I noticed that completion for `--rules=add`, `--rules=j` etc. didn't work.

That was fixed with this tweak:
```
if [[ $i -eq 1 || "${cur%%[=:]*}" == ${COMPREPLY[0]%%=*} && $valopts == *${COMPREPLY[0]}* ]] ; then
```
instead of
```
if [[ $i -eq 1 || "${cur}" == ${COMPREPLY[0]} && $valopts == *${COMPREPLY[0]}* ]] ; then
```

But then I remembered that you also prefer to type `-rules:a` instead of `--rules=a` and added another tweak:
```
 if [[ $i -eq 1 || "${cur%%[=:]*}" == ${COMPREPLY[0]%%=*} || "-${cur%%[=:]*}" == ${COMPREPLY[0]%%=*} && $valopts == *${COMPREPLY[0]}* ]] ; then
```

After testing this change for `--rules=a`, `-rules:a`, `--rules:a`, `-rules=a`, I force-pushed that change to my repo.

magnumripper:
Updated with your latest changes now. `--rules-stack=` does expand (also with `:`), `--rules-stack=j` expands to two alternatives and `--rules-stack=b` expands to the only possible rule. I guess this means I am using your updated set.

But neither `--rules=` nor `--single=` will expand for me. I also tried adding first letter(s) from a rule but that too did not expand. I'm on bash `version 5.1.4(1)-release` in case that possibly matters.

frank-dittrich:
For me, it works as expected.
If the option cannot be expanded, you have to press [tab] twice, to see all available values.
```
$ ./john --rules=[tab][tab]
add1234_everywhere                              lowerunicodesubstitution
add2010everywhere                               monthsfullpreface
adddotcom                                       multiword
addjustnumbers                                  none
addshortmonthseverywhere                        nt
all                                             o
append1_addspecialeverywhere                    o1
append2letters                                  o2
append2numspecial                               o3
append3numspecial                               oi
```

frank-dittrich:
`--rules=ad` will be completed to `--rules=add`, because there is no ambiguity. But then you have to press [tab] twice to see all the rule sets named add<something>.
```
$ ./john --rules=add
add1234_everywhere        adddotcom                 addshortmonthseverywhere
add2010everywhere         addjustnumbers            
```
This should be similar to expanding `--wordlist=`, no?
```
$ ./john --wordlist=
Display all 403 possibilities? (y or n)
```

frank-dittrich:
BTW: I noticed I might also want to adjust `Copyright (c) 2012 Frank Dittrich`.

```
git blame john.bash_completion | cut -d'(' -f 2| cut -d2 -f 1|sort|uniq -c|sort -nr
    917 Frank Dittrich 
     25 magnum         
      9 Solar          
      4 William Sandin 
      1 JimF           
```
Any suggestions? Looks like Solar just removed some trailing spaces.

solardiz:
> Looks like Solar just removed some trailing spaces.

I don't even recall that. I certainly didn't make any copyright-able changes to `john.bash_completion`.

frank-dittrich:
I discovered some more problems with bash completion.
So far, I always assumed that it is allowed to use abbreviated option names, as long as the abbreviated option is unique.
`--internal-codepage` is the only option name listed in john's --help output starting with `--int`.
That's why, bash completion for `--int=` will list all possible encodings, and `--int=l` will expland to `--int=latin1`.

But:
```
$ ./john --int=latin1 dummy
Unknown option: "--int=latin1"
```
So the assumption that a user can abbreviate option names is not universally true.
I guess this is problem is caused by the deprecated `--internal-encoding` option. This is unfortunate, because it contradicts user expectations. Can this somehow be worked around, either in a general way, or for this special case (and any similar cases we might find)? Should I have create a separated issue for this?

I'd like to further enhance bash completion to provide completion for `--inc=` (and `-inc:` etc), `--ru=`, `--si=` and other abbreviated option names, as long as all possible option names start with the name of the shortest valid option. This shouldn't be too difficult and would completely restore the way bash completion worked before --rules-stack, --single-seed and other options have been introduced.

frank-dittrich:
@magnumripper I force-pushed the latest change to allow completion for -inc=, -ru=, etc.

You should be able to verify if you are using the correct bash completion code like this:
```
$ complete | grep -E '(john|unique)'
complete -F _john john
complete -F _unique unique
```
Then 
```
$ set | less
```
and search for `^_john ()`
You should see the code from john.bash_completion, just without comments and empty lines.

If bash completion still doesn't work for you, please add 
```
echo ${COMP_WORDBREAKS} 
```
after
```
LC_ALL=C
```
and
```
echo ${COMPREPLY[0]%%=*}
echo ${COMPREPLY[$i - 1]}
```
after
```
i=${#COMPREPLY[*]};
```
and
```
echo $cur
echo $option
```
before
```
case "${option}" in
```
and let me know the output and the ./john command line so that I get an idea about what's going on.

magnumripper:
OK so I added those lines and typed `../run/john --rules=` and only then pressed TAB twice:
```
$ ../run/john --rules="'><=;|&(:
--rules-skip-nop
--rules
```
If nothing else, now I *really* know I'm actually running that file :)

magnumripper:
The problems I describe are when on macOS, amended with a newer/complete bash and lots of other GNU stuff from Homebrew, first in PATH. So totally normal sed, grep and so on (I'm not sure if you rely on any such external utils?).

I tried your same commit under Ubuntu 20.04 and indeed everything works fine. That is bash version `5.0.16(1)-release` although it's not necessarily the version that is the problem - something might be odd with my macOS environment.

magnumripper:
> So totally normal sed, grep and so on (I'm not sure if you rely on any such external utils?).

You seem to rely on sed, awk, tr, grep, cut. Of those, only cut, awk and tr were still native so just for good measure I tried replacing them (renaming gcut to cut and so on) but nothing changed.

Also just for double-checking I exported LC_ALL=C before trying a completion but that wasn't it either.

frank-dittrich:
> ```
> $ ../run/john --rules="'><=;|&(:
> --rules-skip-nop
> --rules
> ```

So, the first 3 lines of output are exactly the same as on my Fedora system.
But for some reason, the `case "${option}" in` is never reached.

Please checkout john.bash_completion and add some other echo commands to find out where program execution goes wrong.

I made these changes
```
diff --git a/run/john.bash_completion b/run/john.bash_completion
index 445d11e32..9d507cb22 100644
--- a/run/john.bash_completion
+++ b/run/john.bash_completion
@@ -263,11 +263,14 @@ _john()
 	# for the (possibly abbreviated) option
 	i=${#COMPREPLY[*]}
 	if [[ $i != 0 && ${COMPREPLY[$i - 1]} == ${COMPREPLY[0]%%=*}* ]]; then
+echo 'if [[ $i != 0 && ${COMPREPLY[$i - 1]} == ${COMPREPLY[0]%%=*}* ]]; then'
 		# (abbreviated) option is not ambiguous or
 		# option is also part of a longer option name
 		if [[ "${cur}" != *[=:]* ]] ; then
+echo 'if [[ "${cur}" != *[=:]* ]] ; then'
 			# No value specified.
 			if [[ $valopts == *${COMPREPLY[0]}* ]] ; then
+echo 'if [[ $valopts == *${COMPREPLY[0]}* ]] ; then'
 				# Option allows a value
 				if [[ "${cur}" == ${COMPREPLY[0]} ]] ; then
 					if [[ $i -eq 1 && "_${__john_completion}" == "_2" ]] ; then
@@ -303,11 +306,13 @@ _john()
 		compopt  -o bashdefault -o default
 		return 0
 	elif [[ "${cur}" == *[=:]* ]] ; then
+echo 'elif [[ "${cur}" == *[=:]* ]] ; then'
 		# Ambiguously abbreviated option plus value,
 		# no reasonable way for john specific completion
 		COMPREPLY=()
 		return 0
 	else
+echo else
 		# Ambiguously abbreviated option, but no value specified.
 		# COMPREPLY is already filled as it should be
 		return 0
```

With
```
$ ./john --rules=[tab]
```
I get
```
$ ./john --rules=if [[ $i != 0 && ${COMPREPLY[$i - 1]} == ${COMPREPLY[0]%%=*}* ]]; then
```
and after another `[tab]`:
```
if [[ $i != 0 && ${COMPREPLY[$i - 1]} == ${COMPREPLY[0]%%=*}* ]]; then

add1234_everywhere                              appendseason                                    lowerunicodesubstitution                        prependnumnumnumnum
add2010everywhere                               appendspecial3num                               monthsfullpreface                               prependnumnumspecial
adddotcom                                       appendspecial4num                               multiword                                       prependseason
addjustnumbers                                  appendspeciallowerlower                         none                                            prependspecialspecial
addshortmonthseverywhere                        appendspecialnumbernumber                       nt                                              prependspecialspecialappendnumber
all                                             appendspecialnumbernumbernumber                 o                                               prependspecialspecialappendnumbersnumber
append1_addspecialeverywhere                    appendyears                                     o1                                              prependspecialspecialappendnumbersnumbernumber
append2letters                                  appendyears_addspecialeverywhere                o2                                              prependyears
append2numspecial                               best64                                          o3                                              replaceletters
append3numspecial                               d3ad0ne                                         oi                                              replaceletterscaps
append4num                                      devprodtestuat                                  oldoffice                                       replacenumbers
append4numspecial                               dive                                            passphrase-rule1                                replacenumbers2special
append5num                                      drop                                            passphrase-rule2                                replacespecial2special
append6num                                      extra                                           prepend2numbersappend2numbers                   rockyou-30000
appendcap-num_or_special-twice                  hashcat                                         prepend4lettermonths                            shifttoggle
appendcurrentyearspecial                        i                                               prepend4numappendspecial                        single
appendjustnumbers                               i1                                              prependandappendspecial                         single-extra
appendjustspecials                              i2                                              prependcapcapappendspecial                      specific
appendjustspecials3times                        i3                                              prependdaysweek                                 split
appendmonthcurrentyear                          insidepro                                       prependhello                                    t0xlc
appendmonthday                                  jumbo                                           prependjustspecials                             t9
appendnum_addspecialeverywhere                  jumbosingle                                     prependnumnum                                   unicodesubstitution
appendnumbers_and_specials_simple               korelogic                                       prependnumnum_appendnumspecial                  upperunicodesubstitution
appendnumnum_addspecialeverywhere               l33t                                            prependnumnumappendspecial                      wordlist
appendnumnumnum_addspecialeverywhere            loopback                                        prependnumnumnum                                
```
What output do you get?

magnumripper:
After first tab
```
$ ../run/john --rules=elif [[ "${cur}" == *[=:]* ]] ; then
```
After another one, just the same again
```
$ ../run/john --rules=elif [[ "${cur}" == *[=:]* ]] ; then
elif [[ "${cur}" == *[=:]* ]] ; then
```

So I guess I'm not getting a hit by `if [[ $i != 0 && ${COMPREPLY[$i - 1]} == ${COMPREPLY[0]%%=*}* ]]; then`. But you said that my `$i` (or rather my `${COMPREPLY}`) seem to be same as yours. So what could go wrong here?

frank-dittrich:
> After first tab
> 
> ```
> $ ../run/john --rules=elif [[ "${cur}" == *[=:]* ]] ; then
> ```
> 
> After another one, just the same again
> 
> ```
> $ ../run/john --rules=elif [[ "${cur}" == *[=:]* ]] ; then
> elif [[ "${cur}" == *[=:]* ]] ; then
> ```
> 
> So I guess I'm not getting a hit by `if [[ $i != 0 && ${COMPREPLY[$i - 1]} == ${COMPREPLY[0]%%=*}* ]]; then`. But you said that my `$i` (or rather my `${COMPREPLY}`) seem to be same as yours. So what could go wrong here?

I also don't get it.
$i must be greater than 0, otherwise there wouldn't be at least 2 elements in that array.
May be your bash is broken.

Can you try this:
```
$ if [[ _testxxx_ == _test* ]]; then echo true; fi
true
```

magnumripper:
Yeah I also get `true` for that. I'm currently checking bash site for known bugs

frank-dittrich:
Try
```
$ if [[ --rules-stack= == --rules* ]]; then echo true; fi
true
```
If you don't get `true`, please apply this change and test:
```
-       if [[ $i != 0 && ${COMPREPLY[$i - 1]} == ${COMPREPLY[0]%%=*}* ]]; then
+       if [[ $i != 0 && _${COMPREPLY[$i - 1]} == _${COMPREPLY[0]%%=*}* ]]; then
```

magnumripper:
> Try
> 
> ```
> $ if [[ --rules-stack= == --rules* ]]; then echo true; fi
> true
> ```

Yeah that's also true. And I didn't find any open relevant bugs. I'm investigating if I can get an older bash version from Homebrew.

frank-dittrich:
Another idea, please try:
```
-       if [[ $i != 0 && ${COMPREPLY[$i - 1]} == ${COMPREPLY[0]%%=*}* ]]; then
+       if [[ $i -ne 0 && ${COMPREPLY[$i - 1]} == ${COMPREPLY[0]%%=*}* ]]; then
```

magnumripper:
That didn't help.

I added some debug output at that point
```diff
@@ -262,12 +262,16 @@ _john()
        # Check the number of different possible completions
        # for the (possibly abbreviated) option
        i=${#COMPREPLY[*]}
+echo " \$i is $i, \$COMPREPLY[\$i - 1] is ${COMPREPLY[$i - 1]}, \$COMPREPLY[\$i - 2] is ${COMPREPLY[$i - 2]}, \$COMPREPLY[0] is ${COMPREPLY[0]}"
```
```
$i is 4, $COMPREPLY[$i - 1] is --rules, $COMPREPLY[$i - 2] is --rules=, $COMPREPLY[0] is --rules-skip-nop
```

frank-dittrich:
I didn't look carefully enough at your first debug output.
And I wrongly assumed COMPREPLY would be sorted.
Try
```
-       if [[ $i != 0 && ${COMPREPLY[$i - 1]} == ${COMPREPLY[0]%%=*}* ]]; then
+       IFS=$'\n' COMPREPLY=($(sort <<<"${COMPREPLY[*]}"))
+       unset IFS
+       if [[ $i -ne 0 && ${COMPREPLY[$i - 1]} == ${COMPREPLY[0]%%=*}* ]]; then
```

magnumripper:
Yep that was it! Works fine then. Strange though, why does it end up different on this machine?!

frank-dittrich:
I guess undefined behaviour because of different internal hash algorithm... Even on my Linux system the array wasn't really sorted, it just happened that a longer option name appeared last.

magnumripper:
Great thing it happened to me so we could sort it out before someone else got the problem after release 👍

frank-dittrich:
Yes, great, except that I didn't want to spend **that** much time fixing bash completion.

magnumripper (approved):
I can't find any problem now

magnumripper:
@frank-dittrich I'll merge this now, unless you have some last-minute thoughts?

frank-dittrich:
@magnumripper let me know if (and how) I should mention you or any other contributor in the copyright notice. Otherwise, please merge.

magnumripper:
I'd expect anyone who changes the file would also add that when on it, if they'd care. Personally I don't care, my changes must have been far too trivial to warrant any mention.

Issue #4632, opened 2021-03-17
Some formats fail '-internal mode internal mask' test


ghost:
It is hard to reproduce. So far I've seen it only using AMD CPU driver (but it is NOT a driver bug). To reproduce:
```
rm -f ../run/opencl/*.bin *.pot; make -sj8 && 
../run/john -lws=2 -gws=2 tst-.in  --skip --mask=0?d -form=xsha512-opencl -dev=7
```

A patch follows ("my formats" only), but I'm not closing the bug. I also saw a failure (SIGSEGV) in: **[edited]**
* `mysql-sha1-opencl`.
* `raw-sha1-opencl`
* `raw-md4-opencl` (return code: 135)
* `raw-md5-opencl`

magnumripper:
Tested on "well" now, with `./jtrts.pl -pass=-dev=7 -internal mask`
* cryptosafe-opencl segfaults
* mscash-opencl segfaults
* nt-opencl segfaults
* ntlmv2-opencl segfaults
* raw-sha1-opencl segfaults

The three other formats you mentioned passed just fine. Intermittent?

When instead running `john -test=0 -form=mask -dev=7`, all 17 format passes with no problem.

magnumripper:
After more TS runs, it becomes obvious most of these segfaults are intermittent

ghost:
> Intermittent?

Sure. 

Well, I can't remember about a fix for `raw-md *` formats. Anyway, a fix is also a possibility.

magnumripper:
How do you know this is not a driver bug?

magnumripper:
Using Intel CPU driver instead (dev 1), some formats do show up too:
```
form=oldoffice-opencl             guesses:   12 N/A N/A  [PASSED]

form=raw-md4-opencl               guesses:    0 -show=   0 0:00:00:00 DONE : Expected count(s) (13)(-show13)  [!!!FAILED1!!!  ret_val=0]

form=raw-md5-opencl               guesses:    0 -show=   0 0:00:00:00 DONE : Expected count(s) (6)(-show6)  [!!!FAILED1!!!  ret_val=0]

form=raw-sha1-opencl              guesses:    0 -show=   0 0:00:00:00 DONE : Expected count(s) (23)(-show23)  [!!!FAILED1!!!  ret_val=0]

[...]

PR #4633, opened 2021-03-17
Properly initiate a mask buffer

ghost:
See #4632


magnumripper (approved):
Makes me wonder why it was 0x80 before?

ghost:
> Makes me wonder why it was 0x80 before?

I do not know. I probably got something wrong.

Issue #4638, opened 2021-03-18
DPAPImk2john.py is Python 2 only


NonoLePoulet:
When i am trying to run DPAPImk2john.py, it return an error :
  File "PathToJohn\run\DPAPImk2john.py", line 339
    print "%s-%s-%s-%s-%s%s" % (format(GUID[0], '08x'), format(GUID[1], '04x'), format(GUID[2], '04x'), format(GUID2[0], '04x'), format(GUID2[1], '08x'), format(GUID2[2], '04x'))
          ^
SyntaxError: invalid syntax

solardiz:
This is a Python 2 script, but you appear to run it with Python 3. Not really a bug, but we should make this script Python 3 compatible.

solardiz:
@Fist0urs @clavoillotte @jagotu Does one of you possibly want to make `DPAPImk2john.py` Python 3 compatible (ideally keeping it Python 2 compatible as well) and test that? (The three of you are recent committers to this script.) Or maybe @exploide wants to take care of this (having made related contributions to other scripts)? Thanks!

NonoLePoulet:
Ok, thanks for the reply. 
You should specify in the readme that some scripts require python 2, because it wasn't very clear witch version of python you needed.

solardiz:
Maybe we should have such scripts themselves use `python2` in the shebang line.

exploide:
Yes, changing the shebang to `python2` would be the correct short term fix.

The [DPAPImk2john.py](https://github.com/openwall/john/blob/bleeding-jumbo/run/DPAPImk2john.py) script is comparably complex and I'm not deep into the DPAPI stuff. Maybe I find some time to see at least how much effort it requires or if it is just an easy fix.

Sample data along with complete command line invocations would be important to check if porting the script goes successful. Unfortunately there are no DPAPI samples in [john-samples](https://github.com/openwall/john-samples).

solardiz:
> Yes, changing the shebang to `python2` would be the correct short term fix.

@exploide I'd appreciate it if you identify all of our scripts that are Python 2 only (I think grepping for "print " gives something close) and change their shebangs accordingly (in a PR). Thanks!

I doubt the changed shebang would have been of much help to @NonoLePoulet, though, who appears to be on Windows (judging by the backslashes) and is probably invoking `python` manually. The hope there would be that after seeing the error the user would then take a look inside the script and see the `python2` there.

I also don't readily know how to (re)test this script, having never used it before. Yes, we need samples.

PR #4641, opened 2021-03-20
Fixed shebangs and Python 3 compatibility

exploide:
As requested in #4638 I searched for python2-only scripts and changed their shebangs accordingly.

For `ibmiscanner2john.py`, `keyring2john.py` and `ps_token2john.py` it was possible to generate sample data and subsequently made them compatible with both, Python 2 and Python 3. :smiley: 

(I'm not stopping to complain that having compatibility with both versions is much more complicated than Python 3 only. And it doesn't make the code prettier... :wink:  Eventually, Python 2 support should stop to lower the maintenance overhead.)

### DIFF
--- a/run/aix2john.py
+++ b/run/aix2john.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 
 import binascii
 import sys
@@ -62,17 +62,16 @@ if __name__ == "__main__":
     						help='Use this option if "lpa_options '
     								'= std_hash=true" is activated'
     						)
-    
+
     parser.add_argument('-f', dest="filename",
     						default=False,
     						help='Specify the AIX shadow file filename to read (usually /etc/security/passwd)'
     						)
-    
+
     args = parser.parse_args()
-    
+
     if args.filename:
         process_file(args.filename, args.is_standard)
-    else:   
+    else:
         print "Please specify a filename (-f)"
         sys.exit(-1)
-
--- a/run/aix2john.py
+++ b/run/aix2john.py
@@ -1,18 +1,30 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python
 
+# This script extracts hashes from AIX /etc/security/passwd
+# cat /etc/security/passwd
+# root:
+#         password = mrXXXXXXXXXX
+#         lastupdate = 1343960660
+#         flags =
+#
+# admin:
+#         password = oiXXXXXXXXXX
+#         lastupdate = 1339748349
+#         flags =
+# ...
+
+# Note: Some of this functionality is also covered by the unshadow program.
+
+
+import argparse
 import binascii
 import sys
 import re
 
-try:
-	 import argparse
-except ImportError:
-    sys.stderr.write("Stop living in the past. Upgrade your python!\n")
-
 
 def process_file(filename, is_standard):
     try:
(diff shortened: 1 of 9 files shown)


run/aix2john.py
```diff
@@ -62,17 +62,16 @@ def process_file(filename, is_standard):
...
-    
+
     args = parser.parse_args()
-    
+
     if args.filename:
         process_file(args.filename, args.is_standard)
-    else:   
+    else:
         print "Please specify a filename (-f)"
```
solardiz: I wonder what else, besides this one `print`, in this script is Python 2 only?

run/aix2john.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
```
solardiz, line 2: We could want to add a comment in here and in `aix2john.pl` that some of the functionality of these scripts is also included in the `unshadow` program.

run/ibmiscanner2john.py
```diff
@@ -7,6 +7,12 @@
 # Output is sent to stdout. Redirect stdout to create a file for JtR.
 # See hackthelegacy.org for ibmiscanner tooling
 
+# Example input hash with username=heidi and password=diamonds
+# heidi:2B22E1F044E51AB1A645FC21A42019FEDD8B6950
+# Output:
+# heidi:$as400ssha1$2B22E1F044E51AB1A645FC21A42019FEDD8B6950$heidi
```
solardiz, line 13: This conversion is so trivial I think we ought to add support for it to this format in JtR itself, using `prepare()` to extract the username for salt. Just thinking out loud - not a reason to change anything in this PR now.

solardiz (approved):
Obviously, there's much room for further improvement, but these changes so far look good to me, and I trust they've been tested to some extent. ;-) Thank you!

The "Stop living in the past. Upgrade your python!" message in `aix2john.py` should probably be changed to something more descriptive (about `argparse`) or the `try` / `except` dropped. We should do this consistently across scripts, not have varying messages for the same thing.

solardiz:
@exploide You want to make any changes per my comments, or do we merge this as-is? Either way is fine with me.

solardiz:
@exploide Thank you for updating `aix2john.py`. Ready to merge now?

exploide:
> The "Stop living in the past. Upgrade your python!" message in aix2john.py should probably be changed to something more descriptive (about argparse) or the try / except dropped.

This was already there and appears to come from a time when Python 2 did not include argparse by default. Later releases of Python 2 have argparse so there is no need for that try-catch anymore (i.e. if someone still uses Python 2, it is at least 2.7 since 2.6 is really legacy and nothing we need to care about).

> I wonder what else, besides this one print, in this script is Python 2 only?

I didn't take a closer look since I had no sample data but found such data in the Perl script. In fact there was only one little thing: The file was read in binary mode. Don't know why this was done, because as far as I understood, the file is always plaintext. Anyway, it's compatible with both now. :)

> We could want to add a comment in here and in aix2john.pl that some of the functionality of these scripts is also included in the unshadow program.

I added a comment to `aix2john.py`. But I'm wondering: `unshadow`, `aix2john.py` and `aix2john.pl` all more or less do the same? Can't we drop one or two of them? Especially the Perl script does not seem to contain any functionality not present in the Python version.

> Ready to merge now?

It's ready :)

PR #4649, opened 2021-03-27
LM & NT formats: Add a sanity check for pwdump format

### LINKED ISSUE #4647: John fails to crack zip file when the filename is a valid hash
### BUG ###
John is failing to crack a zip file whose filename is a valid hash. I didn't try to reproduce the bug with another file type.

### Steps to reproduce ###
Create a zip with filename `9fdaa46e100f0dc089d68662db2f5759` and encrypt with `infected`. Try to crack it:
- `zip2john 9fdaa46e100f0dc089d68662db2f5759 > hash`
- `echo 'infected' | john --stdin hash`
- `john --show hash`
`0 password hashes cracked, 1 left`

Rename the file to `z9fdaa46e100f0dc089d68662db2f5759` and retry to crack, and it should correctly detect the password.

### System configuration ###
- OS: `Ubuntu 18.04.5 LTS (Bionic Beaver)`
- `john --list=build-info`
```
Version: 1.9.0-jumbo-1+bleeding-f44e2814d 2021-03-10 15:49:07 +0100
CPU tests: AVX2
$JOHN is /opt/john/run/
Format interface version: 14
Rec file version: REC4
Charset file version: CHR3
CHARSET_MIN: 1 (0x01)
CHARSET_MAX: 255 (0xff)
CHARSET_LENGTH: 24
SALT_HASH_SIZE: 1048576
SINGLE_IDX_MAX: 32768
SINGLE_BUF_MAX: 4294967295
Effective limit: Max. KPC 32768
Max. Markov mode level: 400
gcc version: 7.5.0
GNU libc version: 2.27 (loaded: 2.27)
Crypto library: OpenSSL
OpenSSL library version: 01010100f
OpenSSL 1.1.1  11 Sep 2018
GMP library version: 6.1.2
File locking: fcntl()
fseek(): fseek
ftell(): ftell
fopen(): fopen
memmem(): System's
HR timer: clock_gettime(), latency 134 ns
Total physical host memory: 31997 MiB
Available physical host memory: 24018 MiB
Terminal locale string: it_IT.UTF-8
Parsed terminal locale: UTF-8
```

magnumripper:
If field[1][0] is '$', don't bother trying to parse as pwdump format because false positives from other hash types was seen.

Fixes #4647.

### DIFF
--- a/src/LM_fmt.c
+++ b/src/LM_fmt.c
@@ -71,7 +71,7 @@ static void init(struct fmt_main *self)
 
 static char *prepare(char *fields[10], struct fmt_main *self)
 {
-	if (fields[2] && strlen(fields[2]) == 32)
+	if (fields[1][0] != '$' && fields[2] && strlen(fields[2]) == 32)
 		return fields[2];
 	return fields[1];
 }
--- a/src/nt2_fmt_plug.c
+++ b/src/nt2_fmt_plug.c
@@ -265,7 +265,7 @@ static char *prepare(char *split_fields[10], struct fmt_main *self)
 {
 	static char out[33 + TAG_LENGTH + 1];
 
-	if (!valid(split_fields[1], self)) {
+	if (!valid(split_fields[1], self) && split_fields[1][0] != '$') {
 		if (split_fields[3] && strlen(split_fields[3]) == 32) {
 			sprintf(out, "%s%s", FORMAT_TAG, split_fields[3]);
 			if (valid(out, self))
--- a/src/opencl_lm_fmt_plug.c
+++ b/src/opencl_lm_fmt_plug.c
@@ -65,7 +65,7 @@ static void init(struct fmt_main *pFmt)
 
 static char *prepare(char *fields[10], struct fmt_main *self)
 {
-	if (fields[2] && strlen(fields[2]) == 32)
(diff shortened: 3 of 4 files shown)


src/LM_fmt.c
```diff
@@ -71,7 +71,7 @@ static void init(struct fmt_main *self)
 
 static char *prepare(char *fields[10], struct fmt_main *self)
 {
-	if (fields[2] && strlen(fields[2]) == 32)
+	if (fields[1][0] != '$' && fields[2] && strlen(fields[2]) == 32)
```
solardiz, line 74: Maybe we could use `(fields[1][0] != '$' || !strchr(&field[1][1], '$'))` - that is, exclude only lines that have at least two dollar signs in that field.

Ditto for the 3 other formats modified in this PR.

solardiz (approved):

src/LM_fmt.c
magnumripper: I'm concerned about performance issues in prepare functions (which is a bit of a problem already). Maybe I worry too much.

solardiz:
Circle CI "detected an anomaly" that violates their ToS again.

ghost:
> Circle CI "detected an anomaly" that violates their ToS again.

Not exactly. The merged code is building. 

Circle has "something" with Magnum.

PR #4252, opened 2020-04-11
Create bestcryptve2john.py

trounce1:
script to extract relevant fields from bestcrypt volume encrypted disks. the script also supports going through the first 4 partitions of an MBR  to find the offsets of the partitions and look for the bestcrypt headers at the starting sector. this works for V3 and V4 bestcrypt volume encryption.

PR #4651, opened 2021-03-30
Adds Jetico BestCrypt Volume Encryption v4 format. Addresses #4242

Jissynacktiv:
Hi there!

As discussed in #4242, here is the PR for the Jetico BestCrypt Volume Encryption version 4.

For a bit of context, BestCrypt offers the user the choice to encrypt containers and volumes (among others). @kholia already implemented the containers format, here is the volume encryption one.

This PR implements the following encryption schemes:
- AES256
- TwoFish
- Serpent
- Camellia

I think this PR could be improved for several reasons:
- as I said @kholia already implemented the other format so we would have two separate files for this software (as I provide `bestcrypt_ve_fmt_plug.c` here). I guess we should merge these into one single file and adapt the input hash format accordingly (eg. `$bcve$version$technology$enc_id$salt$encrypted_data`).
- RC6 algorithm was not implemented as there are no primitives in john's current implementation and I was a bit short on time...
- to be complete, this format should include v3 version (not complicated but not my priority atm)
- default password length is enforced to 8 by the Jetico software, we should hardcode this in the implementation to avoid unnecessary computations
- I know that 0x8000 scrypt cost factor is huge but the performances are really aweful, I guess we can optimize some stuff? It's been a long time since I've implemented a format in john...
- Well yeah it's been a long time since I've implemented a format so there are probably a lot of things I did wrong (or at least not optimized)

Anyway, cheers and thanks to @kholia and @trounce1 for their work on Jetico BestCrypt software!

### DIFF
--- a/src/bestcrypt_ve_fmt_plug.c
+++ b/src/bestcrypt_ve_fmt_plug.c
@@ -1,0 +1,477 @@
+/*
+ * JtR format to crack password of BestCrypt v4 encrypted volume
+ * 
+ * This implementation relies on findings from @trounce1 and @kholia 
+ * reverse-engineering work
+ * Copyright (c) 2021 Jean-Christophe Delaunay <jean-christophe.delaunay [at] synacktiv.com>
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted.
+ *
+ * There's ABSOLUTELY NO WARRANTY, express or implied.
+ */
+
+#if AC_BUILT
+#include "autoconfig.h"
+#endif
+
+#if HAVE_LIBCRYPTO
+
+#if FMT_EXTERNS_H
+extern struct fmt_main fmt_bestcrypt_ve;
+#elif FMT_REGISTERS_H
+john_register_one(&fmt_bestcrypt_ve);
+#else
+
+#include <string.h>
+#include <errno.h>
+
+#ifdef _OPENMP
+#include <omp.h>
+#endif
+
+#include "arch.h"
+#include "misc.h"
+#include "common.h"
+#include "formats.h"
+#include "yescrypt/yescrypt.h"
+#include "aes.h"
+#include "twofish.h"
+#include "serpent.h"
+#include <openssl/camellia.h>
+#include "sha.h"
+
+#define FORMAT_NAME             "BestCrypt Volume Encryption v4"
+#define FORMAT_LABEL            "BestCryptVE4"
+#define FORMAT_TAG              "$bcve$"
+#define FORMAT_TAG_LENGTH       (sizeof(FORMAT_TAG) - 1)
+
+#if !defined(JOHN_NO_SIMD) && defined(__XOP__)
+#define SCRYPT_ALGORITHM_NAME "Salsa20/8 128/128 XOP"
+#elif !defined(JOHN_NO_SIMD) && defined(__AVX__)
+#define SCRYPT_ALGORITHM_NAME "Salsa20/8 128/128 AVX"
+#elif !defined(JOHN_NO_SIMD) && defined(__SSE2__)
+#define SCRYPT_ALGORITHM_NAME "Salsa20/8 128/128 SSE2"
+#else
+#define SCRYPT_ALGORITHM_NAME "Salsa20/8 32/" ARCH_BITS_STR
+#endif
+#define ALGORITHM_NAME          "scrypt " SCRYPT_ALGORITHM_NAME ", AES/TwoFish/Serpent/Camellia"
+
+#define BENCHMARK_COMMENT       " (32768, 16, 1)"
+#define BENCHMARK_LENGTH        0x507
+#define BINARY_SIZE             0
+#define PLAINTEXT_LENGTH        125
+#define SALT_SIZE               sizeof(*cur_salt)
+#define BINARY_ALIGN            1
+#define SALT_ALIGN              sizeof(int)
+#define MIN_KEYS_PER_CRYPT      1
+#define MAX_KEYS_PER_CRYPT      1
+
+#define OMP_SCALE               1
+
+#define aesId                   8
+#define twofishId               9
+#define serpentId               10
+#define rc6Id                   11
+#define camelliaId              15
+
+#define SERPENT_KS (140 * 4)
+
+static struct fmt_tests tests[] = {
+	{"$bcve$4$8$4944e8bfed6c688b6b30ed3aedb21c9efeba82a1425ebf8e$a4b51d5e6d0b15ba039d0a62cb1a2f4a1d0f6c0325658b15334657b7732a6dc4ec72de7a7c04216d0dd5167bc23e384ac99bca490307445bcbeeaf09da15465ba6540daa52422afdce6f199f92632e0675d9f90693a49ec964beb2358c7a95ff", "openwall"},
+	{"$bcve$4$8$4afeb70b633687b1869fb6f99c988930688391478d730672$e1c4c86980e60aaebb55b2857be43055aae966925812687481e6844a3ded5480f51302cdd633e040a30790abdd8ea520e28ec7a261e4226bdc3019b1905fe1df1fbbbc98b7ee409ef2ba8be359cf18ccc2f9f9dbeb20d75c5445c35d77e03907","B4by$harKtudududu!"},
+	{"$bcve$4$9$06046a1b1bde8dca9df92696f600ce54b0084cdde00e6e07$f708a6916dbe91d105e689089ca70c37dc18e761773154f28063932ead9eebdcf9fb478ce48923408eb15ccfec99e44372d47bff9aca6b96ba10d9f8c0b9d6b57b75f27f359bef6e632370fd701fca2160f46d39fffcefb30e769bafb87cbd32", "openwall123!"},
+	{"$bcve$4$10$65c70a51ec78c46a17e0604804f8ef297b118a42c8109b05$2f92091413f51a446b8576d06790d682b251a97694d19786c45a9245dacb5100fcfc757ba993a83e6f8e13da0d726c57bdf28e70beaa734034c69cd7f655afaebf6c3cf1a0617f673e267b6e5bc7a3619c9eb6801aa8f1a1824ed3cca55ece0d", "!1234openwall"},
+	//{"$bcve$4$11$5fbbb914874770659e46e7f5a72b506076271bfebcaed823$7216a0b623354b0f2ed417a121cbbed39e0a1ce84c86cd4c5dbe75cbe00b0e947f1cda7e10d7ce7c88f479c086aa672d4c4610896cbf81f6dff82e16d0648d3adbadc271058c7f0edb40a599bca7fbc8eefb26be106b182595e4fa68fcf47b9c", "!1234openwall"},
+	{"$bcve$4$15$1386f84fa770e372d9221bb4e4aaf8ee2e0d40e19b69917c$f7a0c8bcdf279fe29f5ab144da729e9dc0c4ca12efb1e2be25e417f30db78227831bb8ee1bee84b0cd37582af2aaa34102fcc851729d676630d7f3de34f6258bd639e6a96301df4038601b30961d35b502506bf4b4d373e9f007b95f9b1abeb2", "openwall123!"},
+	{NULL}
+};
+
+static int max_threads;
+static yescrypt_local_t *local;
+
+static char (*saved_key)[PLAINTEXT_LENGTH + 1];
+static int *saved_len;
+static int *cracked;
+static int any_cracked;
+static size_t cracked_size;
+
+static struct custom_salt {
+	int version;
+	int enc_algoID;
+	unsigned char salt[24];
+	unsigned char encrypted_data[0x60];
+} *cur_salt;
+
+/* ripped from ossl_aes_crypto.c */
+static void serpent_cbc_decrypt(const unsigned char *in, unsigned char *out,
+                                size_t len, void *key,
+                                unsigned char ivec[16])
+{
+	uint32_t n;
+	union { uint32_t t[16/sizeof(uint32_t)]; unsigned char c[16]; } tmp;
+
+//	assert(in && out && key && ivec);
+	while (len) {
+		unsigned char c;
+		serpent_decrypt(in, tmp.c, key);
+		for (n=0; n<16 && n<len; ++n) {
+			c = in[n];
+			out[n] = tmp.c[n] ^ ivec[n];
+			ivec[n] = c;
+		}
+		if (len<=16) {
+			for (; n<16; ++n)
+				ivec[n] = in[n];
+			break;
+		}
+		len -= 16;
+		in  += 16;
+		out += 16;
+	}
+}
+
+int Twofish_Decrypt_no_padding(Twofish_key *m_key, Twofish_Byte *pInput, Twofish_Byte *pOutBuffer,
+                               int nInputOctets, Twofish_Byte *m_pInitVector)
+{
+	int i, numBlocks;
+	Twofish_UInt32 iv[4];
+	union {
+		Twofish_Byte block[16];
+		Twofish_UInt32 p32[4];	// needed for 'requires aligned' machines
+	} x;
+	Twofish_UInt32 *p;
+	Twofish_Byte *block;
+
+	p = x.p32;
+	block = x.block;
+	if ((pInput == NULL) || (nInputOctets <= 0) || (pOutBuffer == NULL)) return 0;
+
+	if ((nInputOctets % 16) != 0) { return -1; }
+
+	numBlocks = nInputOctets / 16;
+
+	memcpy(iv, m_pInitVector, 16);
+
+	for (i = numBlocks; i > 0; i--)
+	{
+		Twofish_decrypt(m_key, (Twofish_Byte *)pInput, (Twofish_Byte *)block);
+		p[0] ^= iv[0];
+		p[1] ^= iv[1];
+		p[2] ^= iv[2];
+		p[3] ^= iv[3];
+		memcpy(iv, pInput, 16);
+		memcpy(pOutBuffer, block, 16);
+		pInput += 16;
+		pOutBuffer += 16;
+	}
+
+	return 16*numBlocks;
+}
+
+static void init(struct fmt_main *self)
+{
+	omp_autotune(self, OMP_SCALE);
+
+#ifdef _OPENMP
+	max_threads = omp_get_max_threads();
+#else
+	max_threads = 1;
+#endif
+
+	local = mem_alloc(sizeof(*local) * max_threads);
+	int i;
+	for (i = 0; i < max_threads; i++)
+		yescrypt_init_local(&local[i]);
+
+	saved_key = mem_calloc(sizeof(*saved_key), self->params.max_keys_per_crypt);
+	saved_len = mem_calloc(self->params.max_keys_per_crypt, sizeof(*saved_len));
+	cracked = mem_calloc(sizeof(*cracked), self->params.max_keys_per_crypt);
+	cracked_size = sizeof(*cracked) * self->params.max_keys_per_crypt;
+	//cracked_count = self->params.max_keys_per_crypt;
+	any_cracked = 0;
+
+	Twofish_initialise();
+}
+
+static void done(void)
+{
+	int i;
+	for (i = 0; i < max_threads; i++)
+		yescrypt_free_local(&local[i]);
+	MEM_FREE(local);
+
+	MEM_FREE(saved_key);
+	MEM_FREE(cracked);
+	MEM_FREE(saved_len);
+}
+
+static int valid(char *ciphertext, struct fmt_main *self)
+{
+	char *p = ciphertext, *ctcopy, *keeptr;
+	int extra;
+
+	if (strncmp(ciphertext, FORMAT_TAG, FORMAT_TAG_LENGTH))
+		return 0;
+	ctcopy = strdup(ciphertext);
+	keeptr = ctcopy;
+	ctcopy += FORMAT_TAG_LENGTH;
+	if ((p = strtokm(ctcopy, "$")) == NULL) // version
+		goto err;
+	if (!isdec(p))
+		goto err;
+	if ((p = strtokm(NULL, "$")) == NULL) // encryption algorithm
+		goto err;
+	if (!isdec(p))
+		goto err;
+	if (atoi(p) != 8 && atoi(p) != 9 && atoi(p) != 10 && atoi(p) != 11 && atoi(p) != 15)
+		goto err;
+	if (atoi(p) == 11) {
+		fprintf(stderr, "Warning: " FORMAT_LABEL ": RC6 encryption not supported yet!\n");
+		goto err;
+	}
+	if ((p = strtokm(NULL, "$")) == NULL) // salt
+		goto err;
+	if (hexlenl(p, &extra) != sizeof(cur_salt->salt) * 2 || extra)
+		goto err;
+	if ((p = strtokm(NULL, "$")) == NULL) // encrypted content
+		goto err;
+	if (hexlenl(p, &extra) != sizeof(cur_salt->encrypted_data) * 2 || extra)
+		goto err;
+
+	MEM_FREE(keeptr);
+	return 1;
+
+err:
+	MEM_FREE(keeptr);
+	return 0;
+}
+
+static void *get_salt(char *ciphertext)
+{
+	static struct custom_salt cs;
+	int i;
+	char *p = ciphertext, *ctcopy, *keeptr;
+	memset(&cs, 0, sizeof(cs));
+
+	ctcopy = strdup(ciphertext);
+	keeptr = ctcopy;
+	ctcopy += FORMAT_TAG_LENGTH;
+	p = strtokm(ctcopy, "$");
+	cs.version = atoi(p);
+	p = strtokm(NULL, "$");
+	cs.enc_algoID = atoi(p);
+	p = strtokm(NULL, "$");
+	for (i = 0; i < sizeof(cs.salt); i++)
+		cs.salt[i] = (atoi16[ARCH_INDEX(p[2*i])] << 4) | atoi16[ARCH_INDEX(p[2*i+1])];
+	p = strtokm(NULL, "$");
+	for (i = 0; i < sizeof(cs.encrypted_data); i++)
+		cs.encrypted_data[i] = (atoi16[ARCH_INDEX(p[2*i])] << 4) | atoi16[ARCH_INDEX(p[2*i+1])];
+
+	MEM_FREE(keeptr);
+
+	return (void *)&cs;
+}
+
+static void set_salt(void *salt)
+{
+	cur_salt = (struct custom_salt *)salt;
+}
+
+static void bcve_set_key(char *key, int index)
+{
+	strnzcpy(saved_key[index], key, sizeof(saved_key[index]));
+}
+
+static char *get_key(int index)
+{
+	return saved_key[index];
+}
+
+static int crypt_all(int *pcount, struct db_salt *salt)
+{
+	const int count = *pcount;
+	int index;
+	int failed = 0;
+
+	if (any_cracked) {
+		memset(cracked, 0, cracked_size);
+		any_cracked = 0;
+	}
+
+	static const yescrypt_params_t params = { .N = 0x8000, .r = 16, .p = 1 };
+
+#ifdef _OPENMP
+#pragma omp parallel for
+#endif
+	for (index = 0; index < count; index++) {
+		unsigned char kdf_out[32];
+		/* BestCrypt uses CBC mode with a null IV */
+		unsigned char iv[16] = {0};
+		unsigned char out[sizeof(cur_salt->encrypted_data)];
+
+		SHA256_CTX ctx;
+		unsigned char sha256_hash[32];
+
+#ifdef _OPENMP
+		int t = omp_get_thread_num();
+		if (t >= max_threads) {
+			failed = -1;
+			continue;
+		}
+#else
+		const int t = 0;
+#endif
+		if (yescrypt_kdf(NULL, &local[t],
+		             (const uint8_t *)saved_key[index],
+		             strlen(saved_key[index]),
+		             (const uint8_t *)cur_salt->salt,
+		             sizeof(cur_salt->salt),
+		             &params,
+		             kdf_out, sizeof(kdf_out))){
+			failed = errno ? errno : EINVAL;
+#ifndef _OPENMP
+			break;
+#endif
+		}
+		/*
+		   we will now use output of scrypt as key for desired encryption
+		   algorithm in CBC mode
+		*/
+		if (cur_salt->enc_algoID == aesId) {
+			AES_KEY aes_key;
+
+			AES_set_decrypt_key(kdf_out, 256, &aes_key);
+			AES_cbc_encrypt(cur_salt->encrypted_data,
+			                out,
+			                sizeof(cur_salt->encrypted_data),
+			                &aes_key,
+			                iv,
+			                AES_DECRYPT);
+		} else if(cur_salt->enc_algoID == twofishId) {
+			Twofish_key tkey;
+
+			Twofish_prepare_key(kdf_out, sizeof(kdf_out), &tkey);
+			Twofish_Decrypt_no_padding(&tkey,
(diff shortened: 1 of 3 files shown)


ghost:
Some tests are failing (in case you don't receive the CI warning email).

Jissynacktiv:
> Some tests are failing (in case you don't receive the CI warning email).

Yup, I'll take a look at it :)

solardiz:
I only skimmed for now. I think the inner loops over `MIN_KEYS_PER_CRYPT` should be gone (and the steps of this size in the outer loop, too) - they're probably a result of copy-paste from a format (for Ethereum?) that also supported another KDF. It's the same issue we had with the initial version of the Restic format.

Jissynacktiv:
> I only skimmed for now. I think the inner loops over `MIN_KEYS_PER_CRYPT` should be gone (and the steps of this size in the outer loop, too) - they're probably a result of copy-paste from a format (for Ethereum?) that also supported another KDF. It's the same issue we had with the initial version of the Restic format.

Done, thx!

Jissynacktiv:
Mmmh, circleci's checks seem stalled(?)

solardiz:
Somehow CircleCI doesn't see the latest force-push, it only sees the previous one as latest: https://app.circleci.com/pipelines/github/openwall/john?branch=pull%2F4651

Anyway, I think there will be more changes and let's hope those will be detected.

Jissynacktiv:
> Somehow CircleCI doesn't see the latest force-push, it only sees the previous one as latest: https://app.circleci.com/pipelines/github/openwall/john?branch=pull%2F4651
> 
> Anyway, I think there will be more changes and let's hope those will be detected.

Indeed, circleci seems to have [some problems](https://support.circleci.com/hc/en-us/articles/360007985634-Why-do-force-pushes-that-change-repository-history-cause-problems-on-CircleCI-) with force push. I'll wait for other changes on this PR before doing a clean one.

src/bestcrypt_ve_fmt_plug.c
```diff
@@ -0,0 +1,456 @@
...
+ * reverse-engineering work
+
+ * ----------------------------------------------------------------------------
+ * "THE BEER-WARE LICENSE" (Revision 42):
+ * <jean-christophe.delaunay@synacktiv.com> wrote this file. As long as you 
+ * retain this notice you can do whatever you want with this stuff. If we meet 
+ * some day, and you think this stuff is worth it, you can buy me a beer in 
+ * return
+ * ----------------------------------------------------------------------------
+ */
```
solardiz, line 12: The only other place in jumbo where we currently have the beerware license is one function in `md5crypt_long_fmt.c` and it has little enough left of the original code that we could probably drop that license from there too. However, since the license's only requirement is to "retain this notice" (with the beer being a non-requirement, but an option - "you can"), it's similar to 2-clause BSD. FreeBSD was OK with that, and Fedora agrees: https://fedoraproject.org/wiki/Licensing/Beerware

So we should be OK with this, although personally I'd prefer the 0-clause BSD wording we use elsewhere in jumbo.

src/bestcrypt_ve_fmt_plug.c
```diff
@@ -0,0 +1,456 @@
...
+#ifdef _OPENMP
+#include <omp.h>
+#endif
+
+#include "arch.h"
+#include "misc.h"
+#include "common.h"
+#include "formats.h"
+#include "params.h"
+#include "options.h"
```
solardiz: I guess this is a leftover from copy-paste from elsewhere? -

```c
#include "params.h"
#include "options.h"
```

src/bestcrypt_ve_fmt_plug.c
```diff
@@ -0,0 +1,456 @@
...
+#include "yescrypt/yescrypt.h"
+#include "aes.h"
+#include "twofish.h"
+#include "serpent.h"
+#include <openssl/camellia.h>
+#include "sha.h"
+#include "jumbo.h"
```
solardiz: I also doubt we need `jumbo.h` here. Maybe also not some others. Please take a look at `scrypt_fmt.c` and `restic_fmt_plug.c` for what an scrypt-using format needs. The Ethereum format (which you probably used as a template) was relatively dirty, and is in need of a cleanup (a separate PR cleaning it up is welcome).

src/bestcrypt_ve_fmt_plug.c
```diff
@@ -0,0 +1,456 @@
...
+
+#define FORMAT_NAME             "BestCrypt Volume Encryption v4 "
+#define FORMAT_LABEL            "Jetico BestCrypt"
+#define FORMAT_TAG              "$bcve$"
+#define FORMAT_TAG_LENGTH       (sizeof(FORMAT_TAG) - 1)
+#define ALGORITHM_NAME          "scrypt [algorithm]-256-CBC SHA-256"
```
solardiz: Please see `scrypt_fmt.c` and `restic_fmt_plug.c` for how we're reporting on scrypt build's (non-)use of SIMD, and do similar here (separately, please feel free to help us clean this up by moving it to a shared `.h` file).

Also, I don't understand what you mean by the "-256-CBC" suffix here. And you probably need to list the various ciphers separated by slashes.

src/bestcrypt_ve_fmt_plug.c
```diff
@@ -0,0 +1,456 @@
...
+
+#define BENCHMARK_COMMENT       "scrypt parameters (32768, 16, 1)"
+#define BENCHMARK_LENGTH        0x507
+#define BINARY_SIZE             0
+#define PLAINTEXT_LENGTH        125
+#define SALT_SIZE               sizeof(*cur_salt)
+#define BINARY_ALIGN            sizeof(uint32_t)
```
solardiz: Why care to align a 0-sized binary? This should be 1.

src/bestcrypt_ve_fmt_plug.c
```diff
@@ -0,0 +1,456 @@
...
+#define SALT_ALIGN              sizeof(uint64_t)
```
solardiz: The largest-alignment field I see in `struct custom_salt` here is `int`, so this should be `sizeof(int)`. In fact, "over-aligning" here would result in very real problems: functions returning pointers to static instances of `struct custom_salt` don't provide this alignment, but the rest of JtR expects it due to the higher value of `SALT_ALIGN` here. So this is a real bug.

src/bestcrypt_ve_fmt_plug.c
```diff
@@ -0,0 +1,456 @@
...
+typedef Twofish_Byte    Byte;
+typedef Twofish_UInt32  UInt32;
+
+/* ripped from ossl_aes_crypto.c */
+static void serpent_cbc_decrypt(const unsigned char *in, unsigned char *out,
+                                size_t len, void *key,
+                                unsigned char ivec[16])
+{
+	size_t n;
+	union { size_t t[16/sizeof(size_t)]; unsigned char c[16]; } tmp;
```
solardiz: I think `size_t` is irrelevant: internally, `serpent_decrypt` works on `uint32_t`, so that's what we should use here (or at least have as one of the fields in this `union`) to avoid violating C aliasing rules (or at least get close to respecting them). We could also use `ARCH_WORD` or `uint64_t` to speed up our own loops here, but I think that would be trickier and unnecessary.

src/bestcrypt_ve_fmt_plug.c
```diff
@@ -0,0 +1,456 @@
...
+
+static struct custom_salt {
+	int version;
+	int enc_algoID;
+	unsigned char salt[24];
+	unsigned char encrypted_data[0x60];
+} *cur_salt;
+
+typedef Twofish_Byte    Byte;
+typedef Twofish_UInt32  UInt32;
```
solardiz: I'd rather use those two `Twofish_*` types directly. The `Byte` and `UInt32` words are somewhat likely to clash with something later, and besides being derived that way we shouldn't use them for anything Twofish-unrelated in here anyway.

src/bestcrypt_ve_fmt_plug.c
```diff
@@ -0,0 +1,456 @@
...
+	if (strncmp(ciphertext, FORMAT_TAG, FORMAT_TAG_LENGTH))
+		return 0;
+	ctcopy = strdup(ciphertext);
+	keeptr = ctcopy;
+	ctcopy += FORMAT_TAG_LENGTH;
+	if ((p = strtokm(ctcopy, "$")) == NULL) // version
+		goto err;
+	if (!isdec(p))
+		goto err;
+	if ((p = strtokm(NULL, "$")) == NULL) // encryption alogrithm
```
solardiz: Typo: "alogrithm"

src/bestcrypt_ve_fmt_plug.c
```diff
@@ -0,0 +1,456 @@
...
+#include "options.h"
+#include "yescrypt/yescrypt.h"
+#include "aes.h"
+#include "twofish.h"
+#include "serpent.h"
+#include <openssl/camellia.h>
+#include "sha.h"
+#include "jumbo.h"
+
+#define FORMAT_NAME             "BestCrypt Volume Encryption v4 "
```
solardiz: Why the trailing space?

src/bestcrypt_ve_fmt_plug.c
```diff
@@ -0,0 +1,456 @@
...
+#define FORMAT_LABEL            "Jetico BestCrypt"
```
solardiz: Must not have a space inside the label. It should be e.g. just `BestCryptVE`, or maybe `BestCrypt4` if we don't intend to add other versions to this same format.

src/bestcrypt_ve_fmt_plug.c
```diff
@@ -0,0 +1,456 @@
...
+	if (!isdec(p))
+		goto err;
+	if ((p = strtokm(NULL, "$")) == NULL) // encryption alogrithm
+		goto err;
+	if (!isdec(p))
+		goto err;
+	if (atoi(p) != 8 && atoi(p) != 9 && atoi(p) != 10 && atoi(p) != 11 && atoi(p) != 15)
+		goto err;
+	if (atoi(p) == 11) {
+		fprintf(stderr, "RC6 encryption not supported yet!\n");
```
solardiz: Should prefix this message with `"Warning: " FORMAT_LABEL ": "`

src/bestcrypt_ve_fmt_plug.c
```diff
@@ -0,0 +1,456 @@
...
+		goto err;
+	}
+	if ((p = strtokm(NULL, "$")) == NULL) // salt
+		goto err;
+	if (hexlenl(p, &extra) != 48 || extra)
+		goto err;
+	if ((p = strtokm(NULL, "$")) == NULL) // encrypted content
+		goto err;
+	if (hexlenl(p, &extra) != 0x60 * 2 || extra)
+		goto err;
```
solardiz, line 240: Should base the 48 and 0x60 here on the corresponding `sizeof` of struct fields.

src/bestcrypt_ve_fmt_plug.c
```diff
@@ -0,0 +1,456 @@
...
+	p = strtokm(ctcopy, "$");
+	cs.version = atoi(p);
+	p = strtokm(NULL, "$");
+	cs.enc_algoID = atoi(p);
+	p = strtokm(NULL, "$");
+	for (i = 0; i < 24; i++)
+		cs.salt[i] = (atoi16[ARCH_INDEX(p[2*i])] << 4) | atoi16[ARCH_INDEX(p[2*i+1])];
+	p = strtokm(NULL, "$");
+	for (i = 0; i < 0x60; i++)
+		cs.encrypted_data[i] = (atoi16[ARCH_INDEX(p[2*i])] << 4) | atoi16[ARCH_INDEX(p[2*i+1])];
```
solardiz, line 269: Ditto re: `sizeof`.

src/bestcrypt_ve_fmt_plug.c
```diff
@@ -0,0 +1,456 @@
...
+
+static int crypt_all(int *pcount, struct db_salt *salt)
+{
+	const int count = *pcount;
+	int index;
+	int failed = 0;
+
+	memset(cracked, 0, sizeof(cracked[0])*cracked_count);
+
+	yescrypt_params_t params = { .N = 0x8000, .r = 16, .p = 1 };
```
solardiz: Can use `static const` here, so that this struct isn't re-created on stack on each call. BTW, we already correctly have that for the one instance that is constant in `multibit_fmt_plug.c`.

src/bestcrypt_ve_fmt_plug.c
```diff
@@ -0,0 +1,456 @@
...
+#else
+		const int t = 0;
+#endif
+		yescrypt_kdf(NULL, &local[t],
+		             (const uint8_t *)saved_key[index],
+		             strlen(saved_key[index]),
+		             (const uint8_t *)cur_salt->salt,
+		             0x18,
+		             &params,
+		             kdf_out, 0x20);
```
solardiz: Must not ignore the return value from `yescrypt_kdf`. Please see `restic_fmt_plug.c` for how to handle it.

Can also use `sizeof` for the 0x18 and 0x20 here.

src/bestcrypt_ve_fmt_plug.c
```diff
@@ -0,0 +1,456 @@
...
+			Camellia_cbc_encrypt(cur_salt->encrypted_data,
+			                     out,
+			                     0x60,
+			                     &ck,
+			                     iv,
+			                     CAMELLIA_DECRYPT);
+		} /* else if(cur_salt->enc_algoID == rc6Id) {
+			TODO
+		}
+		*/
```
solardiz, line 382: Can also use `sizeof` for the 32, 256, and 0x60 above.

src/bestcrypt_ve_fmt_plug.c
```diff
@@ -0,0 +1,456 @@
...
+	return saved_key[index];
+}
+
+static int crypt_all(int *pcount, struct db_salt *salt)
+{
+	const int count = *pcount;
+	int index;
+	int failed = 0;
+
+	memset(cracked, 0, sizeof(cracked[0])*cracked_count);
```
solardiz: The `cracked_count` stuff is slightly suboptimal in that it will unnecessarily `memset` the entire array on each call. Dhiru was since using the `any_cracked` logic (grep other formats for it), which only invokes the `memset` if there's anything to clean from a previous call. Of course, at scrypt settings this high this little overhead is not noticeable.

src/bestcrypt_ve_fmt_plug.c
```diff
@@ -0,0 +1,456 @@
...
+		}
+		*/
+
+		/* we now compute sha256(decrypted_content[0:0x40]) and 
+		compare it with decrypted_content[0x40:0x60] */
+		memcpy(intermediate_keys, out, 0x40);
+		SHA256_Init(&ctx);
+		SHA256_Update(&ctx, intermediate_keys, 0x40);
+		SHA256_Final(sha256_hash, &ctx);
+		cracked[index] = (0 == memcmp(sha256_hash, out + 0x40, 0x20));
```
solardiz, line 389: Drop `intermediate_keys`, drop the copying to it, use `out` directly.

Also, use `sizeof`.

src/bestcrypt_ve_fmt_plug.c
```diff
@@ -0,0 +1,456 @@
...
+	}
+
+	if (failed) {
+#ifdef _OPENMP
+		if (failed < 0) {
+			fprintf(stderr, "OpenMP thread number out of range\n");
+			error();
+		}
+#endif
+	}
```
solardiz, line 405: Revise this error handling piece to also handle error returns from `yescrypt_kdf`.

src/bestcrypt_ve_fmt_plug.c
```diff
@@ -0,0 +1,456 @@
...
+
+static int cmp_one(void *binary, int index)
+{
+	return cracked[index];
+}
+
+static int cmp_exact(char *source, int index)
+{
+	return 1;
+}
```
solardiz, line 423: BTW, I think we should introduce a `default_cmp_exact` like this and use it in all formats that have it dummy like this. Please feel free to make a separate PR with that.

src/bestcrypt_ve_fmt_plug.c
```diff
@@ -0,0 +1,456 @@
...
+		BENCHMARK_LENGTH,
+		0,
+		PLAINTEXT_LENGTH,
+		BINARY_SIZE,
+		BINARY_ALIGN,
+		SALT_SIZE,
+		SALT_ALIGN,
+		MIN_KEYS_PER_CRYPT,
+		MAX_KEYS_PER_CRYPT,
+		FMT_CASE | FMT_8_BIT | FMT_OMP | FMT_HUGE_INPUT,
```
solardiz: @magnumripper Please help review the format flags here. I think the `FMT_HUGE_INPUT` here and in the existing `bestcrypt_fmt_plug.c` format are unnecessary. Also, the existing `bestcrypt_fmt_plug.c` format has `FMT_UNICODE | FMT_ENC`, which might be wrong, and this one does not have any of these - I wonder if they should use similar flags in this respect.

solardiz (changes requested):
Let's clean this up.

solardiz:
Does this PR's format work with output of the script from #4252? Should we merge both of these at once?

solardiz:
> * default password length is enforced to 8 by the Jetico software, we should hardcode this in the implementation to avoid unnecessary computations

If it's merely the default, but other lengths are also possible, then we probably shouldn't hardcode it. Maybe recommend `--length=8` in a README or/and in a "Note: ..." message printed by the format e.g. the first time `valid` detects a "hash" as this format's. This can be in a separate commit/PR.

> * I know that 0x8000 scrypt cost factor is huge but the performances are really aweful, I guess we can optimize some stuff?

Your code looks near-optimal to me given the primitives we have. At these scrypt settings, I'd expect a few hundred c/s on a modern system.

ghost:
> Does this PR's format work with output of the script from #4252?

It seems it does. Well, more or less, I had to edit the sample I posted in #4252 (tag and encryption alogrithm). 

BTW: I have no idea where I got the aes.bin file from.

```
$ ../run/john hash.txt --word=$HOME/resposta 
Loaded 2 password hashes with 2 different salts (Jetico BestCrypt, BestCrypt Volume Encryption v4  [scrypt [algorithm]-256-CBC SHA-256])
Remaining 1 password hash
Will run 8 OpenMP threads
password         (?)     
1g 0:00:00:00 DONE (2021-04-03 19:57) 1.369g/s 54.79p/s 54.79c/s 54.79C/s lapin..password
Session completed. 
```

```
$ ../run/john hash.txt --show:types
?:$bcve$4$8$450886364ae094fb94b4a787e4fc2484b85f305082f63b1a$3a6ead2a1b806e51ba72efe66efde5a96ec4f4966a2155e1dbbb5eeaa55249a6430d648451ad5409d1813bedc98b55c8cd896adce3bca3d5e1d843ef9dd3a8b30ac23dcd92668113c7d9caf444acbdd5e60eaa18d36975089b24cf8459ffce00:::/:/:/:Jetico BestCrypt:0:0:1:$bcve$4$8$450886364ae094fb94b4a787e4fc2484b85f305082f63b1a$3a6ead2a1b806e51ba72efe66efde5a96ec4f4966a2155e1dbbb5eeaa55249a6430d648451ad5409d1813bedc98b55c8cd896adce3bca3d5e1d843ef9dd3a8b30ac23dcd92668113c7d9caf444acbdd5e60eaa18d36975089b24cf8459ffce00:0:
```

src/bestcrypt_ve_fmt_plug.c
magnumripper: Right, `FMT_HUGE_INPUT` should only be set if a ciphertext may be longer than 896 bytes (a.k.a `MAX_CIPHERTEXT_SIZE` which is `(LINE_BUFFER_SIZE - PLAINTEXT_BUFFER_SIZE)`), it should not be used in either format as far as I can tell.

`FMT_UNICODE | FMT_ENC` means the format uses UTF-16 or UTF-32 internally, which isn't the case here AFAICS, nor in `bestcrypt_fmt_plug.c` so that is a bug in the latter.

solardiz: OK, so @Jissynacktiv should drop `FMT_HUGE_INPUT` from here, and someone (who?) should drop `FMT_HUGE_INPUT` and `FMT_UNICODE | FMT_ENC` from `bestcrypt_fmt_plug.c` with a separate commit (if @Jissynacktiv does it, then that commit can be in this same PR).

src/bestcrypt_ve_fmt_plug.c
```diff
@@ -0,0 +1,456 @@
...
+#include "misc.h"
+#include "common.h"
+#include "formats.h"
+#include "params.h"
+#include "options.h"
+#include "yescrypt/yescrypt.h"
+#include "aes.h"
+#include "twofish.h"
+#include "serpent.h"
+#include <openssl/camellia.h>
```
solardiz, line 41: Now that we've made OpenSSL optional, this format should also exclude either itself completely or its support for Camellia when building `--without-openssl` - please refer to recent changes to other formats for how to do this.

solardiz:
@Jissynacktiv Are you going to revise this PR per the feedback you received, or should someone else take it from here? Please let us know. Thanks!

Jissynacktiv:
> @Jissynacktiv Are you going to revise this PR per the feedback you received, or should someone else take it from here? Please let us know. Thanks!

Hi! Sorry I've been pretty busy lately. I'll take a look at that ASAP

Jissynacktiv:
I'll have a bit of time next week to fix these issues. Sorry for the delay

Jissynacktiv:
Ok, so I think this last commit addresses most of @solardiz comments. Some things are remaining, I'll give information about them in the corresponding comments.

src/bestcrypt_ve_fmt_plug.c
Jissynacktiv: Switching from 0 to 1 gives the following error:

```
$ john --format=bestcryptve4 --test
Will run 8 OpenMP threads
Benchmarking: BestCryptVE4, BestCrypt Volume Encryption v4 (32768, 16, 1) [scrypt Salsa20/8 128/128 AVX, AES/TwoFish/Serpent/Camellia 256]... (8xOMP) FAILED (Using default binary() with a non-zero BINARY_SIZE)
```

I could not find some information regarding this constant, could you please explain me its purpose or point me out where should I look to understand things? (I saw a lot of formats also define it to 0).

src/bestcrypt_ve_fmt_plug.c
Jissynacktiv: This is done in https://github.com/openwall/john/pull/4651/commits/468f1e80cf1a73b17e2e524d55eeb97a780b4e61 but some values are still remaining (as 256 or so). I could not find appropriate already defined constants within core.

src/bestcrypt_ve_fmt_plug.c
Jissynacktiv: Configuring with flags `--without-openssl --with-commoncrypto` leads to a compilation error in another format:

```
aram allow-store-data-races=0 -Wno-deprecated-declarations -Wunused-but-set-variable -std=gnu89 -Wdate-time -D_POSIX_SOURCE -D_GNU_SOURCE -D_XOPEN_SOURCE=600  -fopenmp  -pthread -I/usr/local/include -DCL_SILENCE_DEPRECATION -DHAVE_OPENCL -pthread -funroll-loops gpg_common_plug.c -o gpg_common_plug.o
gpg_common_plug.c: In function ‘check_dsa_secret_key’:
gpg_common_plug.c:1086:29: error: dereferencing pointer to incomplete type ‘DSA’ {aka ‘struct dsa_st’}
  error = BN_mod_exp(res, dsa->g, dsa->priv_key, dsa->p, ctx);
                             ^~
gpg_common_plug.c: In function ‘gpg_common_check’:
gpg_common_plug.c:1595:9: error: storage size of ‘dsa’ isn’t known
     DSA dsa;
         ^~~
gpg_common_plug.c:1595:9: warning: unused variable ‘dsa’ [-Wunused-variable]
```

solardiz: We no longer support `--with-commoncrypto`, see ad47c9c1607ed870c018a904694519f6f0ad613d. Why did you pass this option? Where did you even find it documented? If we somehow didn't drop it fully enough, then we should complete that.

Please try configuring/building without that option (don't forget `make clean` after reconfiguring), and let us know whether the build errors remain or are gone. Thanks!

src/bestcrypt_ve_fmt_plug.c
solardiz: I meant `BINARY_ALIGN` should be 1. Not `BINARY_SIZE`, which was correctly 0.

You don't really need the below to complete this PR, but since you asked:

> point me out where should I look to understand things?

The formats interface and its related parameters is documented in `formats.h`. However, this would not fully explain things. Originally, the interface was designed for usual password hashes, where "binary" would have meant their binary representation (decoded from ASCII at loading). This format, like certain others introduced since 2011, is for what we call "non-hashes". With usual password hash formats, we compare computed hashes against loaded hashes. With non-hash formats, we might not have a target hash to compare against (and if so the binary would be zero-sized), but instead we store whatever other information we need for a password correctness check inside the salt structure (thus, we also call such formats "salt-only").

src/bestcrypt_ve_fmt_plug.c
solardiz: This is fine if you insist, but FWIW by "the 0-clause BSD wording we use elsewhere in jumbo" I meant:

```
 *
 * There's ABSOLUTELY NO WARRANTY, express or implied.
```

(after your copyright line).

src/bestcrypt_ve_fmt_plug.c
solardiz: Can you explain why you chose to remove two different `#include` lines than those I had thought were unneeded?

src/bestcrypt_ve_fmt_plug.c
```diff
@@ -37,21 +47,30 @@ john_register_one(&fmt_bestcrypt_ve);
...
+#if !defined(JOHN_NO_SIMD) && defined(__XOP__)
+#define SCRYPT_ALGORITHM_NAME "Salsa20/8 128/128 XOP"
+#elif !defined(JOHN_NO_SIMD) && defined(__AVX__)
+#define SCRYPT_ALGORITHM_NAME "Salsa20/8 128/128 AVX"
+#elif !defined(JOHN_NO_SIMD) && defined(__SSE2__)
+#define SCRYPT_ALGORITHM_NAME "Salsa20/8 128/128 SSE2"
+#else
+#define SCRYPT_ALGORITHM_NAME "Salsa20/8 32/" ARCH_BITS_STR
+#endif
+#define ALGORITHM_NAME          "scrypt " SCRYPT_ALGORITHM_NAME ", AES/TwoFish/Serpent/Camellia 256"
```
solardiz: I'd drop the 256 here.

src/bestcrypt_ve_fmt_plug.c
```diff
@@ -364,11 +390,11 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 
 		/* we now compute sha256(decrypted_content[0:0x40]) and 
 		compare it with decrypted_content[0x40:0x60] */
-		memcpy(intermediate_keys, out, 0x40);
 		SHA256_Init(&ctx);
-		SHA256_Update(&ctx, intermediate_keys, 0x40);
+		SHA256_Update(&ctx, out, 0x40);
 		SHA256_Final(sha256_hash, &ctx);
 		cracked[index] = (0 == memcmp(sha256_hash, out + 0x40, 0x20));
+		any_cracked |= 1;
```
solardiz, line 393: Please add this before the `any_cracked |= 1;` line:

```c
#ifdef _OPENMP
#pragma omp atomic
#endif
```

(The only point in using the `|=` instead of simple assignment is that the former is supported by that pragma directive. If we dare not to use the pragma anyway, then we'd go for simple assignment and hope that it is in fact atomic.)

solardiz:
For merging this PR, please reduce it to two commits - one for "BestCrypt: removal of unnecessary format flags" with its current content (changing just the one line in the existing format) and the other for "Add Jetico BestCrypt Volume Encryption v4 format" with everything else in it. In the latter commit, please also include addition to `NEWS`.

src/bestcrypt_ve_fmt_plug.c
Jissynacktiv: Indeed, the sources I was relying on were too old. I don't have the problem anymore (`--with-commoncrypto` doesn't exist anymore)

src/bestcrypt_ve_fmt_plug.c
Jissynacktiv: Easy answer: misunderstanding of mine :). It will be fixed in the final commit.

src/bestcrypt_ve_fmt_plug.c
Jissynacktiv: Indeed, I misunderstood things regarding the concerned variable. Thanks for the explanation aside that!

src/bestcrypt_ve_fmt_plug.c
Jissynacktiv: I'm not quite sure about what you are saying here. Would you like me to add these statements at the end of what was changed? (https://github.com/openwall/john/pull/4651/commits/468f1e80cf1a73b17e2e524d55eeb97a780b4e61 no beerware licence anymore). These statements are rephrasing what is written above, for me (but I may be wrong).

Edit:

Do you expect something like this?

```
/*
 * Format for cracking BestCrypt V4 encrypted volumes
 * 
 * This implementation relies on findings from @trounce1 and @kholia 
 * reverse-engineering work
 * Copyright (c) 2021 Jean-Christophe Delaunay <jean-christophe.delaunay@synacktiv.com>
 *
 *
 * There's ABSOLUTELY NO WARRANTY, express or implied.
 */
```

solardiz: > Would you like me to add these statements at the end of what was changed?

No. Don't mix two license wordings, we only need one.

> Do you expect something like this?

Yes.

Jissynacktiv:
Everything should be okay by now

ghost:
Your last push makes things worse for the merge. If you can fix it safety, please do. If you can't, we can pick individual commits and merge manually.

magnumripper:
> Your last push makes things worse for the merge. If you can fix it safety, please do. If you can't, we can pick individual commits and merge manually.

I believe if we "rebase and merge", that merge we see now will disappear.

ghost:
We should merge all pending PR and fix them later (if needed).

solardiz:
> I believe if we "rebase and merge", that merge we see now will disappear.

I doubt it. I think instead of adding the merge commit, @Jissynacktiv should have rebased the two commits on the latest. You still can do this, @Jissynacktiv (then force-push).

> We should merge all pending PR and fix them later (if needed).

I agree, but not with the merge commit in this one. We can use "squash and merge" now - but then the separation between the two first commits here will be lost, unfortunately.

Jissynacktiv:
Sorry, I tried to use the github "resolve conflicts" Web UI feature. Bad idea :). I could add NEWS modifications to @magnumripper ones, is it what you wanted in terms of commit history?

solardiz:
> I could add NEWS modifications to @magnumripper ones, is it what you wanted in terms of commit history?

Inside the file, yes, your modifications should be after those. However, in git history only your two commits should appear, and nothing about the unrelated modifications that merely occurred while your PR was open.

As this PR currently is, it can't be merged at all because it includes the duplicate/conflicting changes.

Jissynacktiv:
> > I could add NEWS modifications to @magnumripper ones, is it what you wanted in terms of commit history?
> 
> Inside the file, yes, your modifications should be after those. However, in git history only your two commits should appear, and nothing about the unrelated modifications that merely occurred while your PR was open.
> 
> As this PR currently is, it can't be merged at all because it includes the duplicate/conflicting changes.

Ok, I get it. Now this sould be ok.

Could you proceed to the circleci checks again before merging?

solardiz:
> Could you proceed to the circleci checks again before merging?

CircleCI should have proceeded automatically. That it did not suggests it's probably detected this activity as suspicious - it happens with them and our project all the time. I think they've since whitelisted some of us (as the detection is per the GitHub user submitting the PR), but not you. I'm not going to bother asking their customer support about this again, and will simply merge and CircleCI should then run on our tree after the merge. If any issue turns up, we'll fix it with a follow-up commit.

Thank you!

PR #4654, opened 2021-04-05
Support building without OpenSSL

solardiz:
In my testing on with OpenSSL, `--test --format=cpu` reports 414 formats. Without OpenSSL, it reports 379. All tests pass.

I brought in a SHA-1 implementation from sphlib. Otherwise a lot more formats would need to be excluded. We do have several SHA-1 implementations of our own in various places in the tree, but none are the complete thing with the usual API.

While at it, this PR also repairs the `Makefile.legacy` targets corresponding to those added in 1.9.0 core. The way @magnumripper had merged those core changes resulted in almost all of those being broken (they needed an update to be right for jumbo, not blind merging).

This also updates `Makefile.legacy` and the logic in the source files to allow for legacy builds without OpenSSL, zlib, and GMP - as needed for easier cross-builds.

I also repaired (cross-)building for MIC, which was broken in a few ways.

Using the above, I am now getting this on MIC:

```
[user@super-mic0 run]$ OMP_NUM_THREADS=1 LD_LIBRARY_PATH=.. ./john -test=0
Testing: descrypt, traditional crypt(3) [DES 512/512 MIC]... PASS
[...]
Testing: aix-ssha512, AIX LPA {ssha512} [PBKDF2-SHA512 512/512 MIC 8x]... PASS
Testing: andOTP [SHA256 32/64]... ret 1762719909
FAILED (cmp_all(1))
Testing: ansible, Ansible Vault [PBKDF2-SHA256 HMAC-256 512/512 MIC 16x]... PASS
[...]
Testing: securezip, PKWARE SecureZIP [SHA1 AES 32/64]... PASS
Testing: 7z, 7-Zip archive encryption (512K iterations) [SHA256 512/512 MIC 16x AES]... FAILED (cmp_all(1))
Testing: Signal, Signal Android [PKCS#12 PBE (SHA1) 32/64]... PASS
[...]
Testing: dynamic_2014 [md5($s.md5($p).$s) (PW > 55 or salt > 11 bytes) 512/512 MIC 16x2]... PASS
Testing: dummy [N/A]... PASS
Testing: crypt, generic crypt(3) [?/64]... PASS
2 out of 372 tests have FAILED
```

I haven't yet looked into these 2 failed tests.

This PR will probably need to be cleaned up before merging. I might split the changes into several commits.

### DIFF
--- a/doc/INSTALL
+++ b/doc/INSTALL
@@ -14,23 +14,32 @@ Jumbo/autoconf-specific README-DISTROS.)
 You may have obtained the source code or a "binary" (pre-compiled)
 distribution of John the Ripper.  On Unix-like systems, it is typical
 to get the source code and compile it into "binary" executables right
-on the system you intend to run John on.  On DOS and Windows, however,
-it is typical to get a binary distribution which is ready for use.
+on the system you intend to run John on.  On Windows, however, it is
+typical to get a binary distribution which is ready for use.
 
 The following instructions apply to the source code distribution of
 John only.  If you have a binary distribution, then there's nothing
 for you to compile and you can start using John right away.
 
+Also, the instructions here are mostly generic (not very OS specific).
+Please refer to INSTALL-FEDORA, INSTALL-UBUNTU, or INSTALL-WINDOWS for
+instructions specific to those systems.
 
-	Requirements
 
-The Jumbo versions requires OpenSSL 0.9.7 or later. To get all functionality
-you need 1.0.1. For building, not only the run-time libs are needed but also
-"devel" stuff like header files. This is often a separate package, so, e.g.,
-for Ubuntu you need "libssl-dev" as well as "libssl".
+	Requirements.
 
-Some helper tools are written in Python. A couple helper tools (vncpcap2john
-and SIPdump) need libpcap. A couple of formats need extra libraries in order
+Much functionality of JtR jumbo requires OpenSSL.  To get all of the related
+functionality, you need OpenSSL 1.0.1 or later.  With slightly older OpenSSL,
+you'll receive partial functionality.  You can also build without OpenSSL at
+all by passing the "--without-openssl" option to "./configure" (see below),
+but the functionality of such build will be greatly reduced.
+
+For building with OpenSSL, not only the runtime libraries are needed but also
+"development" stuff like header files.  This is often a separate package, so,
+e.g., for Ubuntu you need "libssl-dev" as well as "libssl".
+
+Some helper tools are written in Python.  A couple helper tools (vncpcap2john
+and SIPdump) need libpcap.  A couple of formats need extra libraries in order
 to get included:
 
 	mozilla		libnss
@@ -39,6 +48,9 @@ to get included:
 
 Again, you also need e.g. libnss-dev, libkrb5-dev and libgmp-dev in order to
 build.
+
+Some formats also need libz and libbz2 (and their "development" components) for
+full functionality.
 
 
 	Compiling the sources on a Unix-like system.
@@ -116,7 +128,7 @@ targets other than "generic" and since they typically use gcc, they are
 usually not affected by this potential problem.
 
 
-	OMP fallback build
+	OMP fallback build.
 
 Even a non-distro "native" build can gain from having OMP fallback. This
 means you will have a "john" binary that has OpenMP support, but that it
@@ -132,7 +144,7 @@ too. Here's how to do it for your "home build":
 Then you just run "./john" (from run directory) as usual.
 
 
-	Optimal interleaving factors (intrinsics builds)
+	Optimal interleaving factors (intrinsics builds).
 
 There is a script "testparas.pl" in the src directory, that can be used
 to test for optimal performance. Run it on an idle system. It will finish
@@ -140,7 +152,7 @@ with printing a full "./configure" command line to use. For most people,
 it's actually overkill. Spend the time learning tricks instead!
 
 
-	Optimal build on OS X
+	Optimal build on OS X.
 
 Using OS X, you can install Xcode (free in App Store) and then its "command
 line tools" and after that a normal build should work fine.  However, using
--- a/doc/NEWS
+++ b/doc/NEWS
@@ -174,6 +174,15 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
 - Add options --subsets-prefer-short and --subsets-prefer-small to Subsets
   mode, affecting candidate order.  [magnum; 2021]
 
+- Support "./configure --without-openssl" on all platforms, which excludes much
+  functionality, but allows the build to complete.  (Previously, this was only
+  supported on macOS and required use of Apple's CommonCrypto.)  [Solar; 2021]
+
+- Repair "make -f Makefile.legacy linux-mic" (easy cross-compilation for first
+  generation Xeon Phi) and make it not depend on zlib by default.  The OpenSSL
+  and GMP dependencies are now also easy to exclude by commenting out lines in
+  Makefile.legacy, for all targets.  [Solar; 2021]
+
 
 Major changes from 1.8.0-jumbo-1 (December 2014) to 1.9.0-jumbo-1 (May 2019):
 
--- a/doc/README-MIC
+++ b/doc/README-MIC
@@ -10,7 +10,7 @@ environments for MIC can be found on Intel's website.
 Library Denpendencies:
 -----------------------
 
-JtR requires some libraries that are not available on MIC, which means you'll 
+JtR can use some libraries that are not available on MIC, which means you'll 
 need to build them for MIC by yourself.
 These libraries are:
     Zlib (libz)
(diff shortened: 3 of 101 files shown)


solardiz:
Things I didn't test yet: building without SIMD (CI will test it now) and actually testing (not just building) the OpenCL formats (some are affected by the host-side changes here).

solardiz:
> Testing: andOTP [SHA256 32/64]... ret 1762719909
> FAILED (cmp_all(1))

The `ret 1762719909` comes from here:

```c
static int aes_gcm_init_hash_subkey(const uint8_t *key, size_t key_len, uint8_t *H, AES_KEY *ctx)
{
        int ret;

        ret = AES_set_encrypt_key(key, key_len * 8, ctx);

        if (ret) {
                printf("ret %d\n", ret);
                return ret;
        }
```

So somehow `AES_set_encrypt_key` returns non-zero? I think this could be because we have:

```c
#undef AES_set_encrypt_key
void JTR_AES_set_encrypt_key(const unsigned char *userKey, const int bits, AES_KEY *key) {
        AES_set_encrypt_key(userKey, bits, key);
}
```

along with:

```c
extern int JTR_AES_set_encrypt_key(const unsigned char *userKey, const int bits, AES_KEY *key);
#define AES_set_encrypt_key(a,b,c) JTR_AES_set_encrypt_key(a,b,c)
```

In other words, we first drop the return value and have the function return `void`, but then declare its prototype as returning an `int` again. We seem to have that for `AES_set_decrypt_key` as well.

solardiz:
> Testing: 7z, 7-Zip archive encryption (512K iterations) [SHA256 512/512 MIC 16x AES]... FAILED (cmp_all(1))

I suspect it could be the aliasing violations in `sevenzip_kdf` or elsewhere.

It could also be the exclusion of zlib without a corresponding exclusion of a test vector, but then we'd have noticed the issue by now in other builds as well. So probably not it.

And since we don't seem to have any test vector(s) for 7z conditional on zlib support, this probably means we don't currently have a test vector that would use it - and we should probably add one (or several).

solardiz:
Can anyone explain why `opencl_pem_fmt_plug.c` depends on `HAVE_LIBZ`? I don't see any usage of libz in there, and `pem_fmt_plug.c` doesn't have such dependency. A bug?

solardiz:
BTW, by my code review, our builds with CommonCrypto also require[d] OpenSSL headers (but not library?) - and my changes here make this assumption too. I don't know if this assumption is reasonable or not, and I don't know whether our CommonCrypto support still worked before this PR or was already broken.

solardiz:
BTW, when we have OpenSSL we use its scalar MD4 and MD5, but in some brief tests I just ran with the md5crypt-long format it's actually slightly faster with my MD5 implementation. (Maybe because OpenSSL's zeroizes the context after each computation, whereas in ours we have this `#if 0`'ed out.) We could want to reconsider this separately. Also for SHA-1 and SHA-2, although there I expect OpenSSL to provide some advantage due to its brief use of SIMD for SHA-2's message scheduling (yes, even when it computes just one hash at a time).

solardiz:
We should probably report the crypto library in use under "Optional libraries/features found:" in `configure` output. Something like:

```
External crypto library......................OpenSSL
```

or `CommonCrypto` or `no`. I am not implementing this yet, maybe @magnumripper will?

Maybe make this output similar to or at least consistent with what we'd report under `build-info` at runtime.

solardiz:
After "Correct JTR_AES_set_*_key() to match prototypes in aes.h" the andOTP format now passes tests on MIC.

This PR is now ready for review, further manual testing, and probably merging.

solardiz:
> > Testing: 7z, 7-Zip archive encryption (512K iterations) [SHA256 512/512 MIC 16x AES]... FAILED (cmp_all(1))
> 
> I suspect it could be the aliasing violations in `sevenzip_kdf` or elsewhere.
> 
> It could also be the exclusion of zlib without a corresponding exclusion of a test vector, but then we'd have noticed the issue by now in other builds as well. So probably not it.

Confirmed that the 7z format passes self-test on regular Linux/x86_64 builds when built without zlib. So the issue isn't that.

BTW, we don't seem to have an easy way to disable usage of zlib, GMP, etc. in autoconf'ed build - there's only autodetection. Perhaps we need to add --without flags. And also --with flags, so that distros' packages can force usage of those libraries (and fail to build if not available).

solardiz:
> I suspect it could be the aliasing violations in `sevenzip_kdf` or elsewhere.

This didn't make a difference (works on host, fails on MIC):

```diff
+++ b/src/7z_fmt_plug.c
@@ -206,9 +206,15 @@ static void sevenzip_kdf(int buf_idx, int *indices, unsigned char *master)
 
        // copy out result
        for (i = 0; i < NBKEYS; ++i) {
-               uint32_t *m = (uint32_t*)&master[i*32];
-               for (j = 0; j < 32/4; ++j)
-                       m[j] = JOHNSWAP(buf_out[HASH_IDX_OUT(i) + j*SIMD_COEF_32]);
+               unsigned char *m = &master[i*32];
+               for (j = 0; j < 32/4; ++j) {
+                       unsigned char *mj = &m[j*4];
+                       uint32_t v = buf_out[HASH_IDX_OUT(i) + j*SIMD_COEF_32];
+                       mj[0] = (v >> 24) & 0xff;
+                       mj[1] = (v >> 16) & 0xff;
+                       mj[2] = (v >> 8) & 0xff;
+                       mj[3] = v & 0xff;
+               }
        }
 }
 #else
```

In fact, the original code wasn't necessarily wrong: the only data flow was from those `uint32_t` writes to `unsigned char` reads later (in `sevenzip_decrypt` calling `AES_set_decrypt_key`), and the latter type can alias anything.

Oh, looking at our `AES_set_decrypt_key` in `aes/openssl/ossl_aes_crypto.c` I see it can actually read in `u32`, but somehow only in MSVC builds:

```c
#if defined(_MSC_VER) && (defined(_M_IX86) || defined(_M_AMD64) || defined(_M_X64))
 #define SWAP(x) (_lrotl(x, 8) & 0x00ff00ff | _lrotr(x, 8) & 0xff00ff00)
 #define GETU32(p) SWAP(*((u32 *)(p)))
 #define PUTU32(ct, st) { *((u32 *)(ct)) = SWAP((st)); }
#else
 #define GETU32(pt) (((u32)(pt)[0] << 24) ^ ((u32)(pt)[1] << 16) ^ ((u32)(pt)[2] <<  8) ^ ((u32)(pt)[3]))
 #define PUTU32(ct, st) { (ct)[0] = (u8)((st) >> 24); (ct)[1] = (u8)((st) >> 16); (ct)[2] = (u8)((st) >>  8); (ct)[3] = (u8)(st); }
#endif
```

So for those builds us writing in `uint32_t` actually happened to be safer (we'd have an aliasing violation there with my code above).

solardiz:
> Testing: 7z, 7-Zip archive encryption (512K iterations) [SHA256 512/512 MIC 16x AES]... FAILED (cmp_all(1))

My next best guess was the issue was in `lzma/LzmaDec.c` or thereabout. 7z format is the only one using that code, and it's first used starting with test vector 1 (not 0, which apparently passed test).

However, rebuilding that file with `-O0` didn't make a difference, whereas rebuilding `7z_fmt_plug.c` with `-O0` resulted in the test taking ages and eventually succeeding. I then tried `-O1` for `7z_fmt_plug.c` and that wasn't good enough.

So we have a miscompile of `7z_fmt_plug.c` or maybe link-time optimization triggered when that file is built with optimization.

solardiz:
> So we have a miscompile of `7z_fmt_plug.c`

FWIW, disabling the little-endian specific optimization of extraction of bytes from `round` doesn't make a difference.

magnumripper:
Good stuff. I'll dive in and test.

> BTW, we don't seem to have an easy way to disable usage of zlib, GMP, etc. in autoconf'ed build - there's only autodetection. Perhaps we need to add --without flags. And also --with flags, so that distros' packages can force usage of those libraries (and fail to build if not available).

For such things I'd edit `autoconfig.h` after configure, which is sort of easy but a tad less easy to script. BTW I think most of a generated Makefile would also have proper dependencies of that header. Having them as option flags wouldn't hurt of course.

solardiz:
> I'd edit `autoconfig.h` after configure

That's what I ended up doing too, for my testing, but it's a hack - not something we'd reasonably recommend to users. I think we need proper --without flags.

Another thing I noticed is that `autoconfig.h` doesn't guard itself against multiple inclusion. Maybe it should, for quicker builds at least. (gcc suppresses the "redefined" warnings for macros when they're to the same value.)

Yet another (mostly unrelated) thing I noticed is several `*2john` tools are built directly from `.c` files, not from `.o` files, so we're unnecessarily(?) getting several C files recompiled multiple times for inclusion in the different tools. IIRC, I saw this for `bitlocker2john` and `keepass2john`. Perhaps this is something to fix separately.

magnumripper:
Note to self: Here's the formats that are lost `--without-openssl` right now, not counting OpenCL formats:
```
as400-des
dmg
DPAPImk
electrum
gpg
keychain
krb4
krb5
krb5-17
krb5-18
krb5-3
krb5asrep
krb5pa-sha1
kwallet
lotus85
mdc2
Mozilla
MSCHAPv2
nethalflm
netlm
netntlm
o10glogon
o3logon
ODF
oracle
PEM
pgpdisk
pgpsda
RACF
RACF-KDFAES
RVARY
sappse
SSH
telegram
VNC
```
Edit: And here's the OpenCL formats:
```
electrum-modern-opencl
gpg-opencl
ODF-opencl
pem-opencl
telegram-opencl
```

solardiz:
> Here's the formats that are lost `--without-openssl` right now

`telegram` and `telegram-opencl` are re-enabled in #4656

Issue #4653, opened 2021-04-04
Problem trying to crack hashes extracted from VRRPv2 packets with AH - IPSEC authentication enabled


mp77-arch:
Hello,

I'm encountering a problem trying to crack hashes extracted from VRRPv2 packets with AH - IPSEC authentication enabled.

I've set up a VRRP service on a linux box using Keepalived.

I configured Keepalived to use AH authentication, with the password "Blabla" (without the double-quotes).

Then I captured the packet using Wireshark, saved them into pcap format and ran pcap2john.py to extract it's hashes (the pcap file contains 8 VRRP packets).

pcap2john.py returned the 8 following lines:

$net-ah$0$450003b000c10000ff3300000a00020fe0000012700400000a00020f000000c10000000000000000000000002114ff0102011b42c0a801fe0000000000000000$40628119fc4d96e13518dfbf
$net-ah$0$4500011800c20000ff3300000a00020fe0000012700400000a00020f000000c20000000000000000000000002114ff0102011b42c0a801fe0000000000000000$edba2a675ae4523b2dcc6aaa
$net-ah$0$450003f800c30000ff3300000a00020fe0000012700400000a00020f000000c30000000000000000000000002114ff0102011b42c0a801fe0000000000000000$73325d321f2a22f3a58f6edb
$net-ah$0$4500037c00c40000ff3300000a00020fe0000012700400000a00020f000000c40000000000000000000000002114ff0102011b42c0a801fe0000000000000000$18711a16494600d4bc369da6
$net-ah$0$4500039000c50000ff3300000a00020fe0000012700400000a00020f000000c50000000000000000000000002114ff0102011b42c0a801fe0000000000000000$6f2704e4fba05ad9c74264c1
$net-ah$0$450003d400c60000ff3300000a00020fe0000012700400000a00020f000000c60000000000000000000000002114ff0102011b42c0a801fe0000000000000000$650d55fe1dd4bceab914b707
$net-ah$0$4500034400c70000ff3300000a00020fe0000012700400000a00020f000000c70000000000000000000000002114ff0102011b42c0a801fe0000000000000000$d8a74572d96762c6fdd5d12e
$net-ah$0$450000a800c80000ff3300000a00020fe0000012700400000a00020f000000c80000000000000000000000002114ff0102011b42c0a801fe0000000000000000$6f9a8beafd4ee81fc027e4a8

[...]

PR #4655, opened 2021-04-05
pcap2john: fix python3 incompatibility in pcap_parser_ah

exploide:
Fixes the problem spotted in #4653 that `pcap2john` does not produce hashes for VRRPv2 AH packets when run under python3.

But note that this does not fix the problem that john can't crack these hashes. The script produces now the same output with python2 and python3 and john is able to load these hashes. But they won't be cracked.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@
 *.exe
 *.dSYM
 .DS_Store
+*.pyc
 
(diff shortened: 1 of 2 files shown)


solardiz:
Thank you, @exploide!

PR #4656, opened 2021-04-06
Avoid left-over dependencies on OpenSSL headers, re-enable Telegram

solardiz:
I was a bit too quick to merge the previous one - turns out `--without-openssl` still needed OpenSSL headers in a couple of places. Trivially fixed here.

Also re-enable Telegram formats (CPU and OpenCL) in `--without-openssl` builds. However, this is weird:

```
$ ../run/john -test=0 -form=telegram-opencl -dev=4
Device 4: GeForce GTX 1080
Testing: telegram-opencl [PBKDF2-SHA1 AES OpenCL]... PASS
Testing: telegram-opencl [PBKDF2-SHA1 AES OpenCL]... PASS
Testing: telegram-opencl [PBKDF2-SHA1 AES OpenCL]... PASS
PASSing: telegram-opencl [PBKDF2-SHA1 AES OpenCL]... ^CWait...
Session aborted
```

That is, it keeps re-testing as if I had requested `--stress-test`. What's this, some kind of memory overwrite affecting option flags?!

Edit: Oh, also weird iteration count reported here:

```
$ ../run/john -te -form=telegram-opencl -dev=4
Device 4: GeForce GTX 1080
Benchmarking: telegram-opencl [PBKDF2-SHA1 AES OpenCL]... LWS=256 GWS=10240 (40 blocks) DONE
Speed for cost 1 (iteration count) of 2143458278
Raw:    23063 c/s real, 23167 c/s virtual, Dev#4 util: 100%
```

and it keeps re-testing, too. (Don't mind the actual speeds - the GPU is in other concurrent use. However, the speeds certainly don't correspond to the obviously wrong reported iteration count.)

### DIFF
--- a/src/aes_ige.h
+++ b/src/aes_ige.h
@@ -52,9 +52,8 @@
 #include <stdio.h>
 #include <string.h>
 #include <stdint.h>
-#include <openssl/aes.h>
 
-// #include "aes.h"
+#include "aes.h"
 
 /* NB: the IV is _two_ blocks long */
 void JtR_AES_ige_encrypt(const unsigned char *in, unsigned char *out,
--- a/src/aes_ige_plug.c
+++ b/src/aes_ige_plug.c
@@ -49,15 +49,9 @@
  *
  */
 
-#if AC_BUILT
-#include "autoconfig.h"
-#endif
-
-#if HAVE_LIBCRYPTO || HAVE_COMMONCRYPTO
-
 #include <string.h>
 
-#include <openssl/aes.h>
+#include "aes.h"
 
 #define N_WORDS (AES_BLOCK_SIZE / sizeof(unsigned long))
 typedef struct {
@@ -198,5 +192,3 @@ void JtR_AES_ige_encrypt(const unsigned char *in, unsigned char *out,
         }
     }
 }
-
-#endif /* OpenSSL */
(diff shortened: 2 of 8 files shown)


solardiz:
> it keeps re-testing as if I had requested `--stress-test`. What's this, some kind of memory overwrite affecting option flags?!

This issue is also present for builds _with_ OpenSSL, after having these changes applied. So I either introduced or exposed a bug here. Unless we figure this out today, I guess I need to split this PR in two, so that we can merge the "Avoid left-over dependencies on OpenSSL headers" commit separately, then think.

solardiz:
> it keeps re-testing as if I had requested `--stress-test`. What's this, some kind of memory overwrite affecting option flags?!

Tried building with ASan, however this problem went away with that other gcc version - whether with ASan or not.

> Speed for cost 1 (iteration count) of 2143458278

Turns out the weird reported iteration count isn't a new problem with my changes - I'm also seeing it in a build from January.

solardiz:
Got this to work right in "telegram-opencl: Avoid "struct custom_salt" layout discrepancy", still don't know why repeated tests were triggered - `options.flags` would be a weird target to hit by an out-of-bounds write.

solardiz:
CI is happy, and this fixes known issues, so I'll merge.

PR #4657, opened 2021-04-07
Drop support for Common Crypto as it was defunct anyway

### LINKED ISSUE #4098: Build with CommonCrypto fails
### Steps to reproduce ###

- download JohnTheRipper-1.9.0-Jumbo-1
- `./configure --disable-native-tests --without-openssl --with-commoncrypto`
- `make`

This gives the following build error:

```
In file included from sha2.c:26:
./sha2.h:37:10: fatal error: 'openssl/opensslv.h' file not found
#include <openssl/opensslv.h>
         ^~~~~~~~~~~~~~~~~~~~
```

This is unexpected, given I asked to build without OpenSSL, and `configure` was happy about it. The end of the output of `configure` was:

```

Target CPU ................................. x86_64 SSE4.1, 64-bit LE
Target OS .................................. darwin17.7.0

Optional libraries/features found:
Memory map (share/page large files) ........ yes
Fork support ............................... yes
OpenMP support ............................. no
OpenCL support ............................. yes
Generic crypt(3) format .................... yes
libgmp (PRINCE mode and faster SRP formats)  yes
128-bit integer (faster PRINCE mode) ....... yes
libz (pkzip and some other formats) ........ yes
libbz2 (gpg2john extra decompression logic)  yes
libpcap (vncpcap2john and SIPdump) ......... yes
```

### System configuration ###

macOS 10.13, with system tools (Xcode, clang) and gmp provided by Homebrew.

This happened while I was trying to look into https://github.com/Homebrew/homebrew-core/pull/43531

magnumripper:
It hasn't worked at all for many years.  Closes #4098.

On a side note I first tried fixing the `--with-commoncrypto` option but there were more work needed than seemed reasonable to me, especially since a SIMD build would be nearly identical to a `--without-openssl` (*edited typo*) build anyway - and we do support SIMD for all and any Macs.

### DIFF
--- a/src/opencl_gpg_fmt_plug.c
+++ b/src/opencl_gpg_fmt_plug.c
@@ -15,7 +15,7 @@
 #include "autoconfig.h"
 #endif
 
-#if HAVE_OPENCL && (HAVE_LIBCRYPTO || HAVE_COMMONCRYPTO)
+#if HAVE_OPENCL && HAVE_LIBCRYPTO
 
 #if FMT_EXTERNS_H
 extern struct fmt_main fmt_opencl_gpg;
@@ -416,4 +416,4 @@ struct fmt_main fmt_opencl_gpg = {
 
 #endif /* plugin stanza */
 
-#endif /* HAVE_OPENCL */
+#endif /* HAVE_OPENCL && HAVE_LIBCRYPTO */
--- a/src/sha.h
+++ b/src/sha.h
@@ -5,28 +5,9 @@
 #include "autoconfig.h"
 #endif
 
-#if HAVE_LIBCRYPTO || HAVE_COMMONCRYPTO
+#if HAVE_LIBCRYPTO
 #include <openssl/opensslv.h>
-
-#include "arch.h"
-#if HAVE_COMMONCRYPTO || (!AC_BUILT &&	  \
-	!defined(SIMD_COEF_32) && defined(__APPLE__) && defined(__MACH__) && \
-	 defined(__MAC_OS_X_VERSION_MIN_REQUIRED) && \
-	 __MAC_OS_X_VERSION_MIN_REQUIRED >= 1070)
-/* Mitigate CommonCrypto name clashes */
-#include "md4.h"
-#include "md5.h"
-#define COMMON_DIGEST_FOR_OPENSSL 1
-#include <CommonCrypto/CommonDigest.h>
-#ifndef SHA_CBLOCK
-#define SHA_CBLOCK CC_SHA1_BLOCK_BYTES
-#endif
-#ifndef SHA_LBLOCK
-#define SHA_LBLOCK CC_SHA1_BLOCK_LONG
-#endif
-#else
 #include <openssl/sha.h>
-#endif
 
 /* For the abuse in pbkdf2_hmac_sha1.h and mscash2_fmt_plug.c */
 #define SHA_H0 h0
@@ -35,7 +16,8 @@
 #define SHA_H3 h3
 #define SHA_H4 h4
 
-#else /* !(HAVE_LIBCRYPTO || HAVE_COMMONCRYPTO) */
+#else /* ! HAVE_LIBCRYPTO */
+
 #include "sph_sha1.h"
 #define SHA_CTX sph_sha1_context
 #define SHA1_Init sph_sha1_init
--- a/doc/NEWS
+++ b/doc/NEWS
@@ -183,6 +183,9 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
   and GMP dependencies are now also easy to exclude by commenting out lines in
   Makefile.legacy, for all targets.  [Solar; 2021]
 
+- Drop support for Apple's CommonCrypto, it was defunct for many years anyway.
+  [magnum; 2021]
+
 
 Major changes from 1.8.0-jumbo-1 (December 2014) to 1.9.0-jumbo-1 (May 2019):
 
--- a/src/KRB4_fmt_plug.c
+++ b/src/KRB4_fmt_plug.c
@@ -36,7 +36,7 @@
 #include "autoconfig.h"
 #endif
 
-#if HAVE_LIBCRYPTO || HAVE_COMMONCRYPTO
+#if HAVE_LIBCRYPTO
 
 #if FMT_EXTERNS_H
 extern struct fmt_main fmt_KRB4;
@@ -315,4 +315,4 @@ struct fmt_main fmt_KRB4 = {
 };
 
 #endif /* plugin stanza */
-#endif /* OpenSSL */
+#endif /* HAVE_LIBCRYPTO */
--- a/src/KRB4_std_plug.c
+++ b/src/KRB4_std_plug.c
@@ -17,7 +17,7 @@
 #include "autoconfig.h"
 #endif
 
-#if HAVE_LIBCRYPTO || HAVE_COMMONCRYPTO
+#if HAVE_LIBCRYPTO
 
 #ifdef KRB4_USE_SYSTEM_CRYPT
 #define _XOPEN_SOURCE 4 /* for crypt(3) */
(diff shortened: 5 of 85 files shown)


src/opencl_gpg_fmt_plug.c
```diff
@@ -416,4 +416,4 @@ struct fmt_main fmt_opencl_gpg = {
 
 #endif /* plugin stanza */
 
-#endif /* HAVE_OPENCL */
+#endif /* HAVE_OPENCL && HAVE_LIBCRYPTO */
```
solardiz, line 419: You don't seem to be editing these comments like this in other files. No reason to edit just this one.

src/sha.h
```diff
@@ -5,28 +5,11 @@
...
-#include <CommonCrypto/CommonDigest.h>
-#ifndef SHA_CBLOCK
-#define SHA_CBLOCK CC_SHA1_BLOCK_BYTES
-#endif
-#ifndef SHA_LBLOCK
-#define SHA_LBLOCK CC_SHA1_BLOCK_LONG
-#endif
-#else
 #include <openssl/sha.h>
-#endif
```
solardiz, line 29: After this removal, we could avoid mixing the order of including OpenSSL vs. our own headers here.

solardiz (approved):
The changes look good to me.

(The code being changed, not so much. We can then be retruring to this PR's commit using it as a list on where we mess with internals of OpenSSL's SHA context structs, which is something we might want to undo later.)

src/sha.h
magnumripper: I guess we can actually drop the `#include "arch.h"` now... it was needed for checking `SIMD_COEF_32`.

magnumripper: Or we could include "arch.h" in top, which in turn will source autoconfig.h if applicable

magnumripper: This make it more like eg. md4.h
```diff
diff --git a/src/sha.h b/src/sha.h
index 244fd46ef..cadf281d4 100644
--- a/src/sha.h
+++ b/src/sha.h
@@ -1,14 +1,10 @@
 #ifndef JOHN_SHA_H
 #define JOHN_SHA_H
 
-#if AC_BUILT
-#include "autoconfig.h"
-#endif
+#include "arch.h" /* also includes autoconfig.h for HAVE_LIBCRYPTO */
 
 #if HAVE_LIBCRYPTO
 #include <openssl/opensslv.h>
-
-#include "arch.h"
 #include <openssl/sha.h>
 
 /* For the abuse in pbkdf2_hmac_sha1.h and mscash2_fmt_plug.c */
```

src/opencl_gpg_fmt_plug.c
magnumripper: The CPU formats have the bottom one as just `/* OpenSSL */`. Perhaps I should change them all to `/* HAVE_LIBCRYPTO */` with a one-liner. (edited)

src/sha.h
magnumripper: Hm, no, while md4.h may need stuff from arch.h, we do not here. I'll just drop the arch.h

magnumripper:
Force-pushed to a single commit, hopefully the final one

ghost:
If you agree, you could change the message:
> requires OpenSSL headers being installed

Because it is wrong. It should be: `requires OpenSSL library or run configure --without-openssl`. 

There is no more need for headers (when using --without-openssl).

solardiz:
> If you agree, you could change the message:
> 
> > requires OpenSSL headers being installed
> 
> Because it is wrong. It should be: `requires OpenSSL library or run configure --without-openssl`.
> 
> There is no more need for headers.

Not exactly. To build with OpenSSL, we still need both the library and the headers, so it makes sense for `configure` to produce the respective messages. However, the messages can also mention `or run configure --without-openssl` now.

solardiz:
> a SIMD build would be nearly identical to a `--without-opencl` build anyway

I guess @magnumripper meant `--without-openssl` there.

solardiz:
Regarding CircleCI, it's once again this:

```
We have detected an anomaly that violates the CircleCI Terms of Service.
We did not test this push because the project has been suspended. Please contact us if you think this is a mistake.
```

The last time this occurred, it went away on its own (but not for the builds that were already prevented from running, like this one).

@magnumripper Please let me know when you think this is ready to merge.

magnumripper:
So maybe like this?
```diff
diff --git a/src/configure.ac b/src/configure.ac
index 428560918..faf2eda7b 100644
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -584,8 +584,7 @@ AS_IF([test "x$with_openssl" != xno],
    PKG_CHECK_MODULES([OPENSSL], [openssl],
       AC_DEFINE(HAVE_LIBCRYPTO,1,[Define to 1 if you have the `crypto' library (-lcrypto).])
          AC_DEFINE(HAVE_LIBSSL,1,[Define to 1 if you have the `ssl' library (-lssl).]),
-         AC_MSG_FAILURE([JtR requires OpenSSL and OpenSSL-devel being installed. Install if not installed.
-Try using --disable-pkg-config and possibly helping configure find oSSL by providing hints in CFLAGS and LDFLAGS],1))
+         AC_MSG_FAILURE([OpenSSL and/or OpenSSL-devel not found. Try using --disable-pkg-config and possibly helping configure find them by providing hints in CFLAGS and LDFLAGS (or use --without-openssl option)],1))
 else
    AS_IF([test "x${OPENSSL_CFLAGS}${OPENSSL_LIBS}" != "x" -o "x$cross_compiling" = "xyes" -o "x$enable_native_tests" = xno -o "x$PKG_CONFIG" = xno],
       [AC_MSG_CHECKING([supplied paths for OpenSSL])]
@@ -605,11 +604,11 @@ else
          [AC_DEFINE(HAVE_LIBSSL,1,[Define to 1 if you have the `ssl' library (-lssl).])]
          [AC_DEFINE(HAVE_LIBCRYPTO,1,[Define to 1 if you have the `crypto' library (-lcrypto).])]
          [OPENSSL_LIBS="${OPENSSL_LIBS} -lssl -lcrypto"],
-         [AC_MSG_FAILURE(JtR requires libssl/libcrypto being installed,1)])],
-         [AC_MSG_FAILURE(JtR requires libssl being installed,1)]
+         [AC_MSG_FAILURE(libssl/libcrypto not found, install them or use --without-openssl option,1)])],
+         [AC_MSG_FAILURE(libssl not found, install it or use --without-openssl option,1)]
       )
     LIBS="$LIBS_ORIG"],
-    [AC_MSG_FAILURE(JtR requires OpenSSL headers being installed,1)]
+    [AC_MSG_FAILURE([OpenSSL headers not found, install them or use --without-openssl option],1)]
    )]
   )
   fi]
```

solardiz:
> OpenSSL and/or OpenSSL-devel not found. Try using --disable-pkg-config and possibly helping configure find them by providing hints in CFLAGS and LDFLAGS (or use --without-openssl option)

BTW, it's a bug that we recommended `CFLAGS` here - should be `CPPFLAGS`, or else people would end up building without compiler optimizations. (I already fixed similar issues in `doc/*` recently.)

Then, "OpenSSL-devel" is weird. Red Hat distros have this package name, but here we try to say something generic. So maybe "OpenSSL development libraries not found".

magnumripper:
True, I'll fix that and look for other similar things

magnumripper:
> Then, "OpenSSL-devel" is weird. Red Hat distros have this package name, but here we try to say something generic. So maybe "OpenSSL development libraries not found".

The -dev(el) packages are usually not "development libraries" though, but simply the header files. I'll have it say "OpenSSL and/or it's headers not found (...)" unless you disagree.

solardiz:
> The -dev(el) packages are usually not "development libraries" though, but simply the header files.

It's trickier than that: the `.so` symlink (without a specific number) is usually part of the development subpackage, and is needed for linking against the library (but not for use of a specific soname at runtime). Also, the static `.a` libraries may be part of a development subpackage (e.g., in `-devel-static`).

> I'll have it say "OpenSSL and/or it's headers not found (...)" unless you disagree.

"its", not "it's". OK, if you feel this simplification is preferable.

magnumripper:
OK, it's now "OpenSSL and/or its development libraries not found (...)"

I think we're all set now

solardiz:
Great. I'll merge.

BTW, also the pkgconfig stuff is often in a development subpackage.

PR #4660, opened 2021-04-09
sha1crypt, sha1crypt-opencl: Benchmark for 20000 iterations only; Minor cleanups

solardiz:
It was weird that we benchmarked sha1crypt for two different iteration counts. I looked for a standard/default iteration count, but didn't find one, except 40000 here: https://rchouinard.github.io/phpass/api/class-PHPassLib.Hash.SHA1Crypt.html

I also found https://github.com/hashcat/hashcat/issues/1204 where sha1crypt was added to hashcat. @roycewilliams posted many test vectors to there, and all of those use various random'ish iteration counts around 20000. The actual hashcat tree now includes a test vector with exactly 20000. Since it's sometimes needed to compare JtR vs. hashcat benchmarks, I decided to go with 20000 in JtR as well, by simply adding hashcat's test vector and limiting our benchmark to it (but also using our other test vectors for self-test).

I squeeze a couple of unrelated cleanups into this PR (separate commits, and should be kept as such on rebase-and-merge, so they wouldn't be in any way associated with the sha1crypt changes in git history).

### DIFF
--- a/doc/README-MIC
+++ b/doc/README-MIC
@@ -1,8 +1,8 @@
 
 Intel MIC (Xeon Phi) is a coprocessor computer architecture developed by Intel.
-Program built for MIC cannot run on normal Intel processors. So you need to 
-have a MIC coprocessor and related enviroments setup before trying to build 
-John the Ripper (JtR) for MIC. Tutorial on how to setup the software 
+Program built for MIC cannot run on normal Intel processors. So you need to
+have a MIC coprocessor and related enviroments setup before trying to build
+John the Ripper (JtR) for MIC. Tutorial on how to setup the software
 environments for MIC can be found on Intel's website.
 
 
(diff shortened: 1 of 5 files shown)


solardiz:
Incidentally, hashcat reports the iteration count as 19999. I didn't look into why exactly. We report what's seen in the hash encoding, so 20000. I think that's OK.

solardiz:
Regarding this comment I add to `mic.h`:

```
/* 2 is faster at least for Bitcoin, but makes sha512crypt fail self-test */
#define SIMD_PARA_SHA512	1
```

here's how things fail when I set SIMD_PARA_SHA256 and SIMD_PARA_SHA512 to 2:

```
Testing: sha256crypt, crypt(3) $5$ (rounds=5000) [SHA256 512/512 MIC 16x2]... (236xOMP) FAILED (cmp_all(2399))
Testing: sha512crypt, crypt(3) $6$ (rounds=5000) [SHA512 512/512 MIC 8x2]... (236xOMP) FAILED (cmp_all(14))
```

With `OMP_NUM_THREADS=1`, sha256crypt passes test, but sha512crypt also fails like above.

PR #4662, opened 2021-04-11
Fixed VRRP AH extraction in pcap2john

exploide:
Fixed the problem described in #4653.

Using scapy is more straight forward than dpkt.

I also cleaned up the imports a bit.

### DIFF
--- a/run/pcap2john.py
+++ b/run/pcap2john.py
@@ -38,9 +38,6 @@ except ImportError:
     sys.exit(1)
 
 
-sys.stderr.write("Note: This program does not have the functionality of wpapcap2john, SIPdump, eapmd5tojohn, and vncpcap2john programs which are included with JtR Jumbo.\n\n")
-
-
 def pcap_parser_bfd(fname):
 
     f = open(fname, "rb")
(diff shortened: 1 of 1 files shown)


run/pcap2john.py
```diff
@@ -1207,34 +1203,20 @@ def pcap_parser_ah(fname):
...
+            pkt = pkt['IP']
+            # zero mutable fields
+            pkt.tos = 0
+            pkt.flags = 0
+            pkt.chksum = 0
+            # store icv; zero icv in original packet
+            icv = pkt['AH'].icv
+            pkt['AH'].icv = bytearray(len(icv))
+            # print net-ah hash
+            print("$net-ah$0$%s$%s" % (hexlify(bytes(pkt)).decode('ascii'), hexlify(icv).decode('ascii')))
```
solardiz, line 1198: Looks like all other parsers use `sys.stdout.write`, not `print`. I don't really mind as long as this works across the same Python versions, although we could want to standardize on one or the other approach.

exploide: `print()` should be available on more recent Python2 I think.

Agree that it should be standardized. I think `print` is a bit cleaner and more "pythonic" than `sys.stdout.write`. I changed the occurrences.

solardiz: `print()` worked for me in a quick Python shell test with RHEL6's 2.6.6.

exploide:
The other (non-AH) VRRP packets were also not correctly parsed with python3. Fixed that function, too.

It's amazing how simple the parsing becomes when using scapy.

PR #4666, opened 2021-04-14
Oldoffice formats: Tweaks and fixes

### LINKED ISSUE #4664: oldoffice CPU format lacks possible early-rejection
See #4659.

magnumripper:
CPU format: Early-reject when possible (eg. 43% boost).
OpenCL format: Adjust internal mask target for better speed.
Both: Consistently log (as opposed to print) MITM keys when found, and also log when they're read from uid field upon initial loading.

Closes #4664

On a side note, any logging during loading doesn't end up in the log file (presumably because `log_init()` happens later), but it can be seen using `-log-stderr`

Tested with TS and manually, with and without `--keep-guessing` and with or without initially known MITM. Hopefully I didn't break anything.

### DIFF
--- a/run/opencl/oldoffice_kernel.cl
+++ b/run/opencl/oldoffice_kernel.cl
@@ -30,6 +30,7 @@ typedef struct {
 	uint has_extra;
 	uint extra[32/4];
 	volatile uint has_mitm;
+	uint mitm_reported;
 	uint mitm[8/4]; /* Meet-in-the-middle hint, if we have one */
 } salt_t;
 
--- a/src/oldoffice_common.h
+++ b/src/oldoffice_common.h
@@ -69,10 +69,11 @@ typedef struct {
 typedef struct {
 	unsigned char verifier[16]; /* or encryptedVerifier */
 	unsigned char verifierHash[20];  /* or encryptedVerifierHash */
-	unsigned int has_mitm;
-	unsigned char mitm[5]; /* Meet-in-the-middle hint, if we have one */
 	unsigned int has_extra;
 	unsigned char extra[32]; /* Optional extra data for avoiding FP w/ type 3 */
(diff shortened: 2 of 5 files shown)


magnumripper:
BTW the "weird" differences between the CPU version and the OpenCL version are because the CPU version is `FMT_BLOB` so can take advantage of same-salts (newer versions of a document often keeps the same salt). I haven't managed to do the same to the OpenCL version because its internal mask and GPU-side comparison make it very tricky. It might be possible, I'm not sure.

solardiz:
> CPU format: Early-reject when possible (eg. 43% boost).

I assume that's when we have MITM in the input file (where would it come from?) or after we've found one valid password and are finding more collisions, right?

> BTW the "weird" differences between the CPU version and the OpenCL version are because

I think you should include this explanation in a source code comment in some relevant place.

solardiz (approved):
I skimmed and this looks good to me, but I'm not familiar with these formats. I also don't know if it's OK to change the layout of `binary_blob` or not (maybe this needs to be reflected elsewhere as well?)

magnumripper:
> > CPU format: Early-reject when possible (eg. 43% boost).
> 
> I assume that's when we have MITM in the input file (where would it come from?) or after we've found one valid password and are finding more collisions, right?

Exactly. If we got a password and want to find more of them (perhaps in search of the actual one*), we can place the logged MITM key in the uid field and keep trying, now with `--keep-guessing`. Actually, even if we initially used `--keep-guessing` we need to do the same prior to resuming a job, or it will again go back to non-MITM speed until next candidate was found. I haven't figured out any way to store it with our current core and pot file scheme.

*) You can tell the actual password if you have several files with different salt that should have one same password: The actual password will work on all of them while the collisions will only work on the very file where it was found.

Issue #4670, opened 2021-04-16
OpenCL formats failing with AMD Catalyst on Tahiti


solardiz:
Just to have this recorded somewhere:

```
Platform #1 name: AMD Accelerated Parallel Processing, version: OpenCL 2.0 AMD-APP (1642.5)
    Device #0 (3) name:     Tahiti
    Board name:             AMD Radeon HD 7900 Series
    Device vendor:          Advanced Micro Devices, Inc.
    Device type:            GPU (LE)
    Device version:         OpenCL 1.2 AMD-APP (1642.5)
    Driver version:         1642.5 (VM) - Catalyst 14.12(Omega) [supported]
    Native vector widths:   char 4, short 2, int 1, long 1
    Preferred vector width: char 4, short 2, int 1, long 1
    Global Memory:          3013 MiB
    Global Memory Cache:    16 KiB
    Local Memory:           32 KiB (Local)
    Constant Buffer size:   64 KiB
    Max memory alloc. size: 2097 MiB
    Max clock (MHz):        1000
    Profiling timer res.:   1 ns
    Max Work Group Size:    256
    Parallel compute cores: 32
    Stream processors:      2048  (32 x 64)
    Speed index:            2048000
    SIMD width:             16
    Wavefront width:        64
    ADL:                    Overdrive5, device id 0
    PCI device topology:    03:00.0
    PCI lanes:              16/16
    Fan speed:              20%
    Temperature:            48C
    Utilization:            1%
```

`rar-opencl` segfaults, the rest at least let the tests run to completion:

```
$ time ./john -test=0 -form=opencl,-rar-opencl
Device 3: Tahiti [AMD Radeon HD 7900 Series]
Testing: sha1crypt-opencl, (NetBSD) [PBKDF1-SHA1 OpenCL]... PASS
Testing: KeePass-opencl [SHA256 AES/Twofish/ChaCha OpenCL]... FAILED (cmp_all(20))
Testing: oldoffice-opencl, MS Office <= 2003 [MD5/SHA1 RC4 OpenCL]... PASS
Testing: PBKDF2-HMAC-MD4-opencl [PBKDF2-MD4 OpenCL]... PASS
Testing: PBKDF2-HMAC-MD5-opencl [PBKDF2-MD5 OpenCL]... PASS
Testing: PBKDF2-HMAC-SHA1-opencl [PBKDF2-SHA1 OpenCL]... PASS
[...]

PR #4672, opened 2021-04-17
DEScrypt-opencl: Fix for cache files with systemwide builds.

magnumripper:
We need to use JOHN_PRIVATE_HOME for files we write to.
Also, create the opencl subdirectory inside private home directory. This
was needed for the existing shared code as well.

### DIFF
--- a/src/opencl_DES_bs_b_plug.c
+++ b/src/opencl_DES_bs_b_plug.c
@@ -15,7 +15,12 @@
 #include "../run/opencl/opencl_DES_hst_dev_shared.h"
 #include "mask_ext.h"
 
-#define CONFIG_FILE 	"$JOHN/opencl/DES_bs_kernel_b_%s.config"
+#if JOHN_SYSTEMWIDE
+#define CONFIG_FILE	JOHN_PRIVATE_HOME "/opencl/DES_bs_kernel_b_%s.config"
+#else
+#define CONFIG_FILE	"$JOHN/opencl/DES_bs_kernel_b_%s.config"
+#endif
+
 #define PADDING 	2048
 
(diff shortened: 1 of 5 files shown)


magnumripper:
Tested now, seem to work just fine.

src/opencl_DES_bs_b_plug.c
```diff
@@ -15,7 +15,12 @@
 #include "../run/opencl/opencl_DES_hst_dev_shared.h"
 #include "mask_ext.h"
 
-#define CONFIG_FILE 	"$JOHN/opencl/DES_bs_kernel_b_%s.config"
+#if JOHN_SYSTEMWIDE
+#define CONFIG_FILE	JOHN_PRIVATE_HOME "/opencl/DES_bs_kernel_b_%s.config"
+#else
+#define CONFIG_FILE	"$JOHN/opencl/DES_bs_kernel_b_%s.config"
```
solardiz, line 21: I guess it's whitespace changes on the `CONFIG_FILE` line here and in other changed files? If so, I hope they're for the better.

solardiz (approved):
Makes sense to me. Of course, testing is needed. OK to perform additional testing after merge.

solardiz:
I didn't test this with a system-wide build, but I re-ran the tests of #4670 (as I think it's a platform different from what magnum tested) - no change. Good. I also made sure `descrypt-opencl` actually cracks a password - it did.

src/opencl_DES_bs_b_plug.c
magnumripper: They were inconsistently made of tabs and/or spaces. All new ones are a single tab between macro name and definition.

PR #4675, opened 2021-04-19
OpenCL: Chug along after seeing profiling errors during auto-tune

### LINKED ISSUE #4044: Inconsistent OpenCL error handling in format-specific vs. common code

magnumripper:
If profiling errors seen during LWS/GWS tests, silently ignore that run's timings and continue.

Closes #4044

Note: I have no means to test this. In https://github.com/openwall/john/issues/4044#issuecomment-634982743 @solardiz wrote "*whenever I see a certain platform fail to run a certain format for me with similar errors, and this helps. For example, nt-opencl wouldn't run on Xeon Phi for me without this hack; with the hack, it does.*" but that doesn't seem to be the case now (nt-opencl runs just fine without this patch as well).

### DIFF
--- a/src/opencl_common.c
+++ b/src/opencl_common.c
@@ -1544,8 +1544,12 @@ static cl_ulong gws_test(size_t gws, unsigned int rounds, int sequential_id)
 
 		int prof_bug = 0;
 
-		HANDLE_CLERROR(clWaitForEvents(1, multi_profilingEvent[i]),
-		               "clWaitForEvents");
+		if (clWaitForEvents(1, multi_profilingEvent[i]) != CL_SUCCESS) {
+			if (options.verbosity > VERB_LEGACY)
+				fprintf(stderr, "Profiling errors; Skipping results\n");
+			return 0;
+		}
+
 		HANDLE_CLERROR(clGetEventProfilingInfo(*multi_profilingEvent[i],
 		                                       CL_PROFILING_COMMAND_SUBMIT,
 		                                       sizeof(cl_ulong), &submitTime,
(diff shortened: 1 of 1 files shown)


ghost (approved):
LGTM. I was not able to force an error like the ones that this patch deals with (I was able to stress the GPU to an error and this patch does not hinder the correct progress).

solardiz:
Hmm, yes, `nt-opencl` works for me on the Xeon Phi now even prior to this PR's fix. Some other fix we made must have helped it.

solardiz:
> `nt-opencl` works for me on the Xeon Phi now even prior to this PR's fix

Testing older builds I had lying around, this specific manifestation of the problem went away between September 18 and September 25 last year. That's when we merged fixes for OpenCL memory leaks.

Anyway, this PR looks good to me, and I'll merge it now, and we'll test some more.

PR #4678, opened 2021-04-21
Add bitcoin-opencl

solardiz:
This is an early dirty hack based on the CPU Bitcoin and OpenCL SSPR formats, to be improved perhaps with separate commits. This passes self-test and benchmarks semi-reasonably (see below), but is otherwise untested.

Some known issues:

1. No code is shared with the CPU format yet.
2. The first SHA-512 (the one taking password and salt inputs) is computed on host.
3. The final AES is computed on host.
4. I don't understand why we use `HASH_LOOPS` of `3*271` in many formats - doesn't appear to agree with the "made by factors" comment. Yet I did the same in here for now for consistency.
5. Auto-tuning somehow reaches insanely high GWS figures even though speeds are dropping - perhaps the format does something wrong?
6. Why the huge rounds/s figures seen in verbose output from auto-tuning?

```
[solar@super src]$ ../run/john -test -form=bitcoin-opencl -dev=1 -v=5
initUnicode(UNICODE, RAW/RAW)
RAW -> RAW -> RAW
Will run 32 OpenMP threads
Device 1: gfx900 [Radeon RX Vega]
Benchmarking: Bitcoin-opencl, Bitcoin Core [SHA512 OpenCL AES 64/64]... (32xOMP) Loaded 12 hashes with 12 different salts to test db from test vectors
Options used: -I opencl -cl-mad-enable -D__GPU__ -DDEVICE_INFO=522 -D__SIZEOF_HOST_SIZE_T__=8 -DDEV_VER_MAJOR=2766 -DDEV_VER_MINOR=4 -D_OPENCL_COMPILER  $JOHN/opencl/bitcoin_kernel.cl
binary size 674832
LWS=7 GWS=49 (7 blocks) PASS,
Test mask: ?a?a?l?u?d?d?s
Calculating best GWS for LWS=64; max. 100 ms single kernel invocation.
xfer: 4.148 us
gws:       256   61716K c/s6171648900000 rounds/s    4.148 us per crypt_all()!
xfer: 5.185 us
gws:       512   98746K c/s9874638300000 rounds/s    5.185 us per crypt_all()+
xfer: 15.111 us
gws:      1024   67765K c/s6776520400000 rounds/s   15.111 us per crypt_all()
xfer: 17.778 us
gws:      2048  115198K c/s11519856000000 rounds/s   17.778 us per crypt_all()+
xfer: 34.667 us
gws:      4096  118152K c/s11815271000000 rounds/s   34.667 us per crypt_all()+
xfer: 58.222 us
gws:      8192  140702K c/s14070282700000 rounds/s   58.222 us per crypt_all()+
xfer: 91.407 us
gws:     16384  179242K c/s17924228900000 rounds/s   91.407 us per crypt_all()+
xfer: 164.445 us
gws:     32768  199264K c/s19926419100000 rounds/s  164.445 us per crypt_all()+
xfer: 360.148 us
gws:     65536  181969K c/s18196963400000 rounds/s  360.148 us per crypt_all()
xfer: 813.185 us
gws:    131072  161183K c/s16118349400000 rounds/s  813.185 us per crypt_all()
xfer: 1.448 ms
gws:    262144  181038K c/s18103867400000 rounds/s    1.448 ms per crypt_all()
xfer: 3.062 ms
gws:    524288  171211K c/s17121162300000 rounds/s    3.062 ms per crypt_all()
xfer: 7.042 ms
gws:   1048576  148894K c/s14889452500000 rounds/s    7.042 ms per crypt_all()
xfer: 12.845 ms
gws:   2097152  163254K c/s16325424800000 rounds/s   12.845 ms per crypt_all()
xfer: 24.105 ms
gws:   4194304  173996K c/s17399632800000 rounds/s   24.105 ms per crypt_all()
xfer: 46.121 ms
gws:   8388608  181878K c/s18187896500000 rounds/s   46.121 ms per crypt_all()
xfer: 90.980 ms
gws:  16777216  184404K c/s18440411400000 rounds/s   90.980 ms per crypt_all()
xfer: 184.388 ms (exceeds 100 ms)
xfer: 111.408 us
gws:     16384  147063K c/s14706304700000 rounds/s  111.408 us per crypt_all()-
Calculating best LWS for GWS=32768
Testing LWS=64 GWS=32768 ... 240.129 ms+
Testing LWS=128 GWS=32768 ... 245.612 ms
Testing LWS=256 GWS=32768 ... 245.722 ms
Calculating best GWS for LWS=64; max. 200 ms single kernel invocation.
xfer: 44.148 us
gws:      4096   92778K c/s9277883400000 rounds/s   44.148 us per crypt_all()!
xfer: 57.482 us
gws:      8192  142514K c/s14251417800000 rounds/s   57.482 us per crypt_all()+
xfer: 94.370 us
gws:     16384  173614K c/s17361449600000 rounds/s   94.370 us per crypt_all()+
xfer: 163.704 us
gws:     32768  200166K c/s20016615300000 rounds/s  163.704 us per crypt_all()+
xfer: 395.259 us
gws:     65536  165805K c/s16580520600000 rounds/s  395.259 us per crypt_all()
xfer: 783.111 us
gws:    131072  167373K c/s16737346300000 rounds/s  783.111 us per crypt_all()
xfer: 1.537 ms
gws:    262144  170469K c/s17046945000000 rounds/s    1.537 ms per crypt_all()
xfer: 2.901 ms
gws:    524288  180678K c/s18067819100000 rounds/s    2.901 ms per crypt_all()
xfer: 5.530 ms
gws:   1048576  189598K c/s18959811900000 rounds/s    5.530 ms per crypt_all()
xfer: 10.829 ms
gws:   2097152  193657K c/s19365739900000 rounds/s   10.829 ms per crypt_all()
xfer: 21.968 ms
gws:   4194304  190920K c/s19092016800000 rounds/s   21.968 ms per crypt_all()
xfer: 44.944 ms
gws:   8388608  186643K c/s18664301500000 rounds/s   44.944 ms per crypt_all()
xfer: 92.015 ms
gws:  16777216  182330K c/s18233094800000 rounds/s   92.015 ms per crypt_all()
xfer: 179.023 ms
gws:  33554432  187430K c/s18743037900000 rounds/s  179.023 ms per crypt_all()
Hardware resources exhausted
xfer: 90.371 us
gws:     16384  181297K c/s18129709700000 rounds/s   90.371 us per crypt_all()-
LWS=64 GWS=32768 (512 blocks) DONE
Speed for cost 1 (iteration count) of 177864
Raw:    4694 c/s real, 182044 c/s virtual
```


solardiz:
Oh, also this assumes the OpenCL device is little-endian.

solardiz:
> binary size 674832

Also might be excessive unrolling on AMD, although the resulting speed isn't awful (about 10% lower than hashcat's, if scaled to the same iteration count).

solardiz:
I'm using this opportunity to try and figure out our split kernel auto-tuning - something which I think only @magnumripper really knows, right? I guess I am not using it right yet.

Some observations: many OpenCL formats include `#define STEP 0`. None define `STEP` to anything other than 0. The only use of `STEP` is here:

```
opencl_autotune.h:      autotune_find_best_gws(sequential_id, rounds, STEP, max_duration, have_lws);
```

Also, `opencl_autotune.h` itself has this:

```c
/* Step size for work size enumeration. Zero will double. */
#ifndef STEP
#define STEP    0
#endif
```

Should we possibly drop the `#define STEP 0` lines from formats?

There's also `SEED`, for which a default exists as well, but formats actually override it to different values.

Also, I notice some formats pass `HASH_LOOPS` and others `2 * HASH_LOOPS` to `opencl_init_auto_setup()`. My first guess is the doubling might be to account for HMAC's two uses of the underlying hash, but this isn't consistent (bugs?):

```
opencl_pbkdf2_hmac_md4_fmt_plug.c:      opencl_init_auto_setup(SEED, 2*HASH_LOOPS, split_events, warn,
opencl_pbkdf2_hmac_md5_fmt_plug.c:      opencl_init_auto_setup(SEED, 2*HASH_LOOPS, split_events, warn,
opencl_pbkdf2_hmac_sha1_fmt_plug.c:     opencl_init_auto_setup(SEED, 2*HASH_LOOPS, split_events, warn,
opencl_pbkdf2_hmac_sha256_fmt_plug.c:   opencl_init_auto_setup(SEED, HASH_LOOPS, split_events, warn,
opencl_pbkdf2_hmac_sha512_fmt_plug.c:   opencl_init_auto_setup(SEED, HASH_LOOPS, split_events, warn,
```

solardiz:
> This passes self-test and benchmarks semi-reasonably (see below), but is otherwise untested.

In my further testing so far, this also cracks passwords correctly (wordlist, mask).

solardiz:
> I'm using this opportunity to try and figure out our split kernel auto-tuning - something which I think only @magnumripper really knows, right?

Probably @claudioandre-br does, too - the sha{256,512}crypt formats use this.

> I guess I am not using it right yet.

I currently have profiling event 1 reserved for `init`, but not yet used - is this acceptable, or should I renumber the events?

solardiz:
> I currently have profiling event 1 reserved for `init`, but not yet used - is this acceptable, or should I renumber the events?

Renumbering the events seems to have fixed the huge GWS issue. I think I'll force-push that change.

solardiz:
Looks reasonable now. I guess I should have somehow initialized event 1 even if unused (or should have removed it for now, which is what I did now).

```
[solar@super src]$ ../run/john -test -form=bitcoin-opencl -dev=1 -v=5
initUnicode(UNICODE, RAW/RAW)
RAW -> RAW -> RAW
Will run 32 OpenMP threads
Device 1: gfx900 [Radeon RX Vega]
Benchmarking: Bitcoin-opencl, Bitcoin Core [SHA512 OpenCL AES 64/64]... (32xOMP) Loaded 12 hashes with 12 different salts to test db from test vectors
LWS=7 GWS=49 (7 blocks) PASS,
Test mask: ?a?a?l?u?d?d?s
Calculating best GWS for LWS=64; max. 100 ms single kernel invocation.
xfer: 4 us, loop: 123x13.310 ms, xfer: 3.704 us
gws:       256      156 c/s    15600000 rounds/s     1.637 s per crypt_all()!
xfer: 5.037 us, loop: 123x13.319 ms, xfer: 4.593 us
gws:       512      312 c/s    31200000 rounds/s     1.638 s per crypt_all()+
xfer: 7.703 us, loop: 123x13.345 ms, xfer: 7.259 us
gws:      1024      623 c/s    62300000 rounds/s     1.641 s per crypt_all()+
xfer: 14.074 us, loop: 123x13.396 ms, xfer: 12.444 us
gws:      2048     1242 c/s   124200000 rounds/s     1.647 s per crypt_all()+
xfer: 28.149 us, loop: 123x13.489 ms, xfer: 74.667 us
gws:      4096     2468 c/s   246800000 rounds/s     1.659 s per crypt_all()+
xfer: 48.889 us, loop: 123x13.612 ms, xfer: 100.111 us
gws:      8192     4892 c/s   489200000 rounds/s     1.674 s per crypt_all()+
xfer: 80.741 us, loop: 123x13.826 ms, xfer: 89.037 us
gws:     16384     9632 c/s   963200000 rounds/s     1.700 s per crypt_all()+
xfer: 159.407 us, loop: 123x27.636 ms, xfer: 167.704 us
gws:     32768     9638 c/s   963800000 rounds/s     3.399 s per crypt_all()
xfer: 316.445 us, loop: 123x57.685 ms, xfer: 333.778 us
gws:     65536     9235 c/s   923500000 rounds/s     7.095 s per crypt_all()
xfer: 637.481 us, loop: 123x121.490 ms (exceeds 100 ms)
xfer: 48.445 us, loop: 123x13.606 ms, xfer: 127.296 us
gws:      8192     4894 c/s   489400000 rounds/s     1.673 s per crypt_all()-
Calculating best LWS for GWS=16384
Testing LWS=64 GWS=16384 ... 210.155 ms+
Testing LWS=128 GWS=16384 ... 209.715 ms
Testing LWS=256 GWS=16384 ... 210.339 ms
Calculating best GWS for LWS=64; max. 200 ms single kernel invocation.
xfer: 29.926 us, loop: 123x13.498 ms, xfer: 63.889 us
gws:      4096     2466 c/s   246600000 rounds/s     1.660 s per crypt_all()!
xfer: 57.037 us, loop: 123x13.618 ms, xfer: 232.740 us
gws:      8192     4889 c/s   488900000 rounds/s     1.675 s per crypt_all()+
xfer: 80.741 us, loop: 123x14.155 ms, xfer: 93.333 us
gws:     16384     9409 c/s   940900000 rounds/s     1.741 s per crypt_all()+
xfer: 159.852 us, loop: 123x29.886 ms, xfer: 167.408 us
gws:     32768     8913 c/s   891300000 rounds/s     3.676 s per crypt_all()
xfer: 316.592 us, loop: 123x60.634 ms, xfer: 333.482 us
gws:     65536     8786 c/s   878600000 rounds/s     7.458 s per crypt_all()
xfer: 663.556 us, loop: 123x121.842 ms, xfer: 667.259 us
gws:    131072     8745 c/s   874500000 rounds/s    14.988 s per crypt_all()
xfer: 1.264 ms, loop: 123x244.729 ms (exceeds 200 ms)
xfer: 48.296 us, loop: 123x13.618 ms, xfer: 127.371 us
gws:      8192     4890 c/s   489000000 rounds/s     1.675 s per crypt_all()-
LWS=64 GWS=16384 (256 blocks) DONE
Speed for cost 1 (iteration count) of 177864
Raw:    4721 c/s real, 136533 c/s virtual
```

solardiz:
> > binary size 674832
> 
> Also might be excessive unrolling on AMD, although the resulting speed isn't awful (about 10% lower than hashcat's, if scaled to the same iteration count).

Adding `#pragma unroll 1` before the `for (i = 0; i < count; i++) {` loop in `loop_sha512` leaves this size almost unchanged. Changing it to `#pragma unroll 2` increases the size to 1536136. I am puzzled why these sizes are so high if they're in fact for just 1 or 2 iterations of that loop.

solardiz:
Another observation is we have definitions of `SHA512_DIGEST_LENGTH` and the like in several OpenCL kernels - we should move them to shared header files providing the corresponding hash functions.

ghost:
> Should we possibly drop the `#define STEP 0` lines from formats?

Yes, since nobody needs a value different than 0. Running in weak devices, I used to play with it, the value of STEP wasn't really important.

ghost:
> xfer: 179.023 ms
> gws:  33554432  187430K c/s18743037900000 rounds/s  179.023 ms per crypt_all()

When you have `HASH_LOOPS > 0`, I expect a different output.

Notice, the 119 x 7x3x2 (in my case) + a final processing to get 5000 rounds.
* 256430000 rounds/s = (51286c/s×5000) in my case
* What is your expected number?
```
xfer: 1.314 ms, crypt: 119x41.340 ms, xfer back: 2.547 ms, prep: 161.660 ms, pp: 2.165 ms, final: 5.105 ms, var: 41.331 ms/41.327 ms
gws:    131072    51286 c/s   256430000 rounds/s     2.555 s per crypt_all()-
```
There is a few details to change in your PR (for example, rounds computation and why are you only measuring the data transfer).

solardiz:
@claudioandre-br That line you quoted is from before I fixed the unused event 1 issue. Please see the new output I posted in here half an hour ago.

ghost:
> also this assumes the OpenCL device is little-endian.

This is an assumption that we all made.

solardiz:
> rounds computation

I seem to be getting a multiplier of 100000 there now. Where does it come from?

> and why are you only measuring the data transfer).

Am I? What makes you think so?

ghost:
> @claudioandre-br That line you quoted is from before I fixed the unused event 1 issue. Please see the new output I posted in here half an hour ago.

So, now you can check the correctness of the values. To me, they look ok.

ghost:
> I seem to be getting a multiplier of 100000 there now. Where does it come from?

Line 213: autotune_run(self, 100000, 0, 200);

```
static void autotune_run(struct fmt_main *self, unsigned int rounds,
	size_t gws_limit, int max_duration)
{
```

ghost:
> > and why are you only measuring the data transfer).
> 
> Am I? What makes you think so?

It is fixed now: xfer **to dev**: 663.556 us, loop **processing**: 123x121.842 ms, xfer **back to host**: 667.259 us

solardiz:
> Line 213: autotune_run(self, 100000, 0, 200);

Oh, I got this from `opencl_sspr_fmt_plug.c` - indeed, the first test vectors in `sspr_common_plug.c` use 100000 iterations. Should I change it to Bitcoin first test vector's iteration count? What does this even affect - just verbose reporting?

ghost:
> > Line 213: autotune_run(self, 100000, 0, 200);
> 
> Oh, I got this from `opencl_sspr_fmt_plug.c` - indeed, the first test vectors in `sspr_common_plug.c` use 100000 iterations. Should I change it to Bitcoin first test vector's iteration count? What does this even affect - just verbose reporting?

It is only 'reporting'. Anyway, the correct values are informative for advanced users.

solardiz:
@claudioandre-br BTW, you use `#define HASH_LOOPS              (7*3*2)` for sha{256,512}crypt. Isn't this unnecessarily too small (increasing the performance overhead on invoking the kernel more times)? Isn't it wrong this isn't a factor of 5000? Maybe use 1000 or 2500?

solardiz:
> Isn't it wrong this isn't a factor of 5000?

Need to be careful there, though - it might need to be e.g. 4999, depending on what's actually counted as a loop iteration, I guess.

solardiz:
Also, we might want to replace the calculation of loops count and the `for` loops with `do` / `while` loops like I did here. It looks totally unnecessary to pre-calculate the loop count (let alone to possibly repeat that integer division on every loop iteration like we're doing now if the compiler can't figure out the `salt` structure stays unchanged across calls into OpenCL APIs).

ghost:
> you use `#define HASH_LOOPS (7*3*2)` for sha{256,512}crypt. Isn't this unnecessarily too small

There is an internal reason (algorithm related) to be 7x3x2. I can add a multiplier, if I want, `7x3x2 x 2` or `7x3x2 x 15`. I tested it once (probably a few times), but the gain was negligible for "regular" GPUs.

solardiz:
@claudioandre-br Yes, I see how 7x3x2 fits the SHA-crypt algorithm. I also notice it's optimal at much higher GWS than I see here, which wouldn't be reachable at much higher `HASH_LOOPS` like those I suggested. So maybe try adding smaller additional multipliers like you suggest.

For bitcoin-opencl, I am now getting good results at HASH_LOOPS 2500. I'll probably amend my commit and push that.

solardiz:
How do we decide on what `SEED` value to use?

solardiz:
I'm adding multi-device support fix in here - not closely relevant to this PR, but I'm too lazy to handle it fully separately, and it's a separate commit anyway.

ghost:
> How do we decide on what `SEED` value to use?

`SEED` is not important for regular formats. 

It is important for very slow formats where you need to adjust autotune starting GWS value. If you format is "complicated", you might need to use a `SEED` = 32, maybe no more than 128. Regular formats could be 1024 (any 'I don't care' value).

solardiz:
Found that `HASH_LOOPS` 2500 causes a regression on the old Titan Kepler, limiting GWS in a way that significantly lowers the performance. Will change perhaps to 1000.

solardiz:
OTOH, LWS=384 achieves even better speeds on that Titan Kepler, and at low GWS too, but auto-tune can't find it (uses 512).

ghost:
> Also, we might want to replace the calculation of loops count and the `for` loops with `do` / `while` loops like I did here. It looks totally unnecessary to pre-calculate the loop count (let alone to possibly repeat that integer division on every loop iteration like we're doing now if the compiler can't figure out the `salt` structure stays unchanged across calls into OpenCL APIs).

In C is something like this really important?
```diff
diff --git a/src/opencl_sha512crypt_fmt_plug.c b/src/opencl_sha512crypt_fmt_plug.c
index a8b1bd084..c4ead8bb2 100644
--- a/src/opencl_sha512crypt_fmt_plug.c
+++ b/src/opencl_sha512crypt_fmt_plug.c
@@ -605,9 +605,10 @@ static int crypt_all(int *pcount, struct db_salt *_salt)
                BENCH_CLERROR(clEnqueueNDRangeKernel(queue[gpu_id], preproc_kernel, 1,
                                                     NULL, &gws, lws, 0, NULL, multi_profilingEvent[4]),
                              "failed in clEnqueueNDRangeKernel II");
+               unsigned int iterations = (salt->rounds / HASH_LOOPS);
 
                for (i = 0;
-                       i < (ocl_autotune_running ? 3 : (salt->rounds / HASH_LOOPS));
+                       i < (ocl_autotune_running ? 3 : iterations);
                        i++) {
                        BENCH_CLERROR(clEnqueueNDRangeKernel(queue[gpu_id], crypt_kernel, 1, NULL,
                                                             &gws, lws, 0, NULL, (ocl_autotune_running ?

```

Is this better? How?
```diff
diff --git a/src/opencl_sha512crypt_fmt_plug.c b/src/opencl_sha512crypt_fmt_plug.c
index a8b1bd084..922ef1ce4 100644
--- a/src/opencl_sha512crypt_fmt_plug.c
+++ b/src/opencl_sha512crypt_fmt_plug.c
@@ -605,10 +605,10 @@ static int crypt_all(int *pcount, struct db_salt *_salt)
                BENCH_CLERROR(clEnqueueNDRangeKernel(queue[gpu_id], preproc_kernel, 1,
                                                     NULL, &gws, lws, 0, NULL, multi_profilingEvent[4]),
                              "failed in clEnqueueNDRangeKernel II");
+               unsigned int left = salt->rounds;
+               i = 0;
 
-               for (i = 0;
-                       i < (ocl_autotune_running ? 3 : (salt->rounds / HASH_LOOPS));
-                       i++) {
+               do {
                        BENCH_CLERROR(clEnqueueNDRangeKernel(queue[gpu_id], crypt_kernel, 1, NULL,
                                                             &gws, lws, 0, NULL, (ocl_autotune_running ?
                                                                     multi_profilingEvent[split_events[i]] : NULL)),  //1, 5, 6
@@ -617,7 +617,9 @@ static int crypt_all(int *pcount, struct db_salt *_salt)
                        HANDLE_CLERROR(clFinish(queue[gpu_id]),
                                       "Error running loop kernel");
                        opencl_process_event();
-               }
+                       left -= HASH_LOOPS;
+                       i++;
+               } while (left > 0 || (ocl_autotune_running && (i < 3)));
                BENCH_CLERROR(clEnqueueNDRangeKernel(queue[gpu_id], final_kernel, 1,
                                                     NULL, &gws, lws, 0, NULL, multi_profilingEvent[5]),
                              "failed in clEnqueueNDRangeKernel III");
```

solardiz:
`HASH_LOOPS` of 1000 was much worse than 2500 on GTX 1080. I ended up using 2000, which works well across the devices I tested.

CI was already happy, and this one value change shouldn't affect anything CI-wise (we don't currently test OpenCL formats there anyway), so I'll go ahead and merge this now.

solardiz:
@claudioandre-br Your first patch moves the division out of the loop. The compiler might not have been able to if it didn't know whatever `salt` points at isn't getting legitimately overwritten by the calls the loop makes into other translation units and libraries.

Your second patch brings this closer to the code I wrote in this PR, but if you have to maintain the `i` variable anyway (I don't), then it's of little further advantage (the division is completely avoided, though).

Don't you have out-of-bounds access in `split_events[i]` when not running auto-tune, though? (In the original code, too.) At least this looks problematic from the (little) context you copied in here.

ghost:
> Don't you have out-of-bounds access in `split_events[i]` when not running auto-tune, though?

No, there is a ternary operator (only one of the expressions 'x' and 'y' is evaluated, unless I'm confusing things).

solardiz:
> there is a ternary operator

Oh, of course. Overlooked it.

solardiz:
> Your first patch moves the division out of the loop.

In that code version, you could also perform the check of `ocl_autotune_running` and cap `iterations` to 3 if so before the loop, so that `ocl_autotune_running ? 3 : iterations` isn't evaluated on each loop iteration.

solardiz:
> > also this assumes the OpenCL device is little-endian.
> 
> This is an assumption that we all made.

I think this varies: some kernels assume OpenCL device is of same endianness with the host, but this one specifically assumes little-endian. So it could work fine e.g. on big-endian PowerPC host with little-endian NVIDIA GPU cards... except I'm not sure what happens to the `uint count` kernel argument, which we pass in host endianness (maybe things like this are converted automatically, or maybe not - I just don't know).

solardiz:
> The first SHA-512 (the one taking password and salt inputs) is computed on host.

Curiously, this can be moved to device without any changes in OpenCL code nor without an extra kernel invocation by simply putting the password+salt into the initial buffer and not reducing the iteration count by one, _if_ all passwords in a batch are such that password+salt fits in 64 bytes. With 16 byte salts (the only size we currently need to support in that place), this means passwords of up to 48 bytes.

Luckily, we do already have a test vector with a longer password, so if we implement this we'll have the fallback code path tested as well.

solardiz:
> > The first SHA-512 (the one taking password and salt inputs) is computed on host.
> 
> Curiously, this can be moved to device without any changes in OpenCL code

No, that was wrong. It'd only work for password length of exactly 48, as we hard-code the SHA-512 input length at 64.

magnumripper:
I actually have a patch complete and ready that does *everything* on GPU. It doesn't result in any boost (nor any regression) but would/could be good for eg. a weak machine with 8 or more very strong GPUs. I don't have time right now but will post a PR soon (it's tested on macOS and Ubuntu w/ nvidia so far).

Edit: #4694

magnumripper:
> _if_ all passwords in a batch are such that password+salt fits in 64 bytes

The M-D buffer is 128 bytes for SHA-512, so it could be up to 112 bytes password, no? My current patch doesn't do it that way though, since I didn't want to go below 125. Maybe I should try it.

solardiz:
> I actually have a patch complete and ready that does _everything_ on GPU.

Great. I won't make further changes to this format until after your PR, then.

> The M-D buffer is 128 bytes for SHA-512, so it could be up to 112 bytes password, no?

My current implementation only transfers 64 bytes to/from device, and it feels wrong to extend that to 128 just to accommodate the different first iteration - and besides we'd also have to encode/transfer the actual password lengths somehow and to avoid hard-coding the length on the first iteration.

magnumripper:
Ideally the code (at least the loop kernel) should be self-vectorizing when applicable, but it currently only benefits CPUs as far as I'm aware, and only with drivers like POCL that doesn't auto-vectorize. So probably not worth the complexity and effort.

PR #4680, opened 2021-04-21
Refactor sha[256/512]crypt to make code cleaner

ghost:


src/opencl_sha256crypt_fmt_plug.c
```diff
@@ -586,10 +587,9 @@ static int crypt_all(int *pcount, struct db_salt *_salt)
 		BENCH_CLERROR(clEnqueueNDRangeKernel(queue[gpu_id], preproc_kernel, 1,
 		                                     NULL, &gws, lws, 0, NULL, multi_profilingEvent[4]),
 		              "failed in clEnqueueNDRangeKernel II");
+		iterations = ocl_autotune_running ? 3 : (salt->rounds / HASH_LOOPS);
```
solardiz, line 591: I think we can move `i` and `iterations` to this scope, and use `unsigned int i, iterations  = ocl_autotune_running ? 3 : (salt->rounds / HASH_LOOPS);` here.

As the code is currently written, I expect some compilers with some options would warn us we got "comparison between signed and unsigned".

solardiz (approved):
Looks good to me, but I suggested a change.

src/opencl_sha256crypt_fmt_plug.c
ghost: To avoid the "forbids mixed declarations and code", I added curly brackets (without indentation).

magnumripper: I think we stopped forbidding that some time ago?

ghost: Ok, but should I add the braces (and avoid mixing) or not?

solardiz: @claudioandre-br magnum is correct, we no longer avoid mixed declarations and code. Please drop the extra curly braces. Thanks!

PR #4681, opened 2021-04-22
Refactor the sha[256/512]crypt to group by length

ghost:
My vote is to merge AFTER the release. It is invasive and needs good testing.
*****
From:
```
$ ../run/john-old --format=sha512crypt-opencl --word=$HOME/rockyou.txt --mask=?w?d?d --min-len=7 --max-len=7 --max-run=100 alltests.in
0g 0:00:01:49 0.12% (ETA: 2021-04-23 04:38) 0g/s 11966p/s 47867c/s 71800C/s 1311912..1311977

$ ../run/john-old --format=sha256crypt-opencl --word=$HOME/rockyou.txt --mask=?w?d?d --min-len=7 --max-len=7 --max-run=100 alltests.in
0g 0:00:01:49 0.23% (ETA: 16:10:22) 0g/s 21536p/s 110074c/s 153146C/s makan57..lovie73
```
To:
```
$ ../run/john --format=sha512crypt-opencl --word=$HOME/rockyou.txt --mask=?w?d?d --min-len=7 --max-len=7 --max-run=100 alltests.in
0g 0:00:01:49 0.12% (ETA: 2021-04-23 04:49) 0g/s 11990p/s 47963c/s 71945C/s 1311912..1311977

$ ../run/john --format=sha256crypt-opencl --word=$HOME/rockyou.txt --mask=?w?d?d --min-len=7 --max-len=7 --max-run=100 alltests.in
0g 0:00:01:49 0.23% (ETA: 16:12:24) 0g/s 21609p/s 110447c/s 153665C/s makan57..lovie73
```
Tahiti: LWS=64 GWS=262144

### DIFF
--- a/src/opencl_sha256crypt_fmt_plug.c
+++ b/src/opencl_sha256crypt_fmt_plug.c
@@ -44,9 +44,6 @@ static sha256_password *plain_sorted;   // sorted list (by plaintext len)
 static sha256_hash *calculated_hash;    // calculated hashes
 static sha256_hash *computed_hash;  // calculated hashes (from plain_sorted)
 
-static int
-*indices;            // relationship between sorted and unsorted plaintext list
-
 static cl_mem salt_buffer;      //Salt information.
 static cl_mem pass_buffer;      //Plaintext buffer.
 static cl_mem hash_buffer;      //Hash keys (output).
@@ -56,7 +53,7 @@ static struct fmt_main *self;
 
 static cl_kernel prepare_kernel, preproc_kernel, final_kernel;
 
-static int new_keys, source_in_use;
+static int new_keys, first_len, lens_differ, source_in_use;
 static int split_events[3] = { 1, 6, 7 };
 
 //This file contains auto-tuning routine(s). It has to be included after formats definitions.
@@ -287,6 +284,11 @@ static void set_key(char *key, int index)
 	//Put the tranfered key on password buffer.
 	memcpy(plaintext[index].pass, key, len);
 	plaintext[index].length = len;
+
+	if (new_keys)
+		lens_differ |= len ^ first_len;
+	else
+		first_len = len;
 	new_keys = 1;
 }
 
@@ -513,7 +515,6 @@ static void done(void)
 	if (program[gpu_id]) {
 		release_clobj();
 		release_kernel();
-		MEM_FREE(indices);
 	}
 }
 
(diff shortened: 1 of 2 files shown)


solardiz:
Thank you, Claudio! This is good stuff. Please don't be discouraged by a lack of immediate real review, and please do not close this PR. I notice you made it from a separate branch this time, so I hope we can keep it open for a while, review without hurry, then likely merge.

solardiz (changes requested):
Overall, I like the direction of these changes (I had proposed some of this myself).

I think the PR title and commit message should be changed from "Refactor the sha[256/512]crypt to group by length" to "sha{256,512}crypt-opencl: Refactor the grouping by length". We already had grouping by length - we're just refactoring it here.

Also, we might want to split this into 2 commits:
1. sha{256,512}crypt-opencl: Skip grouping by length when all are same length
2. sha{256,512}crypt-opencl: Refactor the grouping by length

Also, I'd like us to avoid (de)allocating memory for `indices` on each `crypt_all()`. When implemented, the second commit's title could be:

2. sha{256,512}crypt-opencl: Pre-allocate memory for length grouping indices

And indeed this PR needs to be rebased on current HEAD.

solardiz:
Other notes:

1. We might want to re-initialize `new_keys` and `lens_differ` in `reset()`.
2. We might want to `memcpy()` up to actual length plus 1, not up to `PLAINTEXT_LENGTH`. We already know the actual lengths in those places, don't we?

magnumripper:
I like the cleanliness with this code, I think I will nick it for rar-opencl.

Just brainstorming, we could have some kind of means telling [the shared] auto-tune that we either run single mode, or that we have a format like these that thrive with same-length candidates within a workgroup. When that's the case it could lean slightly more towards keeping LWS low. Now that I think about it, we *may* already have a kludge for single mode somewhere down there.

solardiz:
> could lean slightly more towards keeping LWS low.

Why? In my experiments with md5crypt-opencl, it appeared sufficient (for current GPUs) to avoid length discrepancies within each aligned 64-password group regardless of LWS value (even for LWS higher than 64). This suggests that when length discrepancies are not avoided, it does not matter what LWS is, either. No?

solardiz:
I also don't fully understand what @claudioandre-br's benchmarks in the first comment here (are expected to) actually show. A mask of `?w?d?d` provides groups of only 100 same-length passwords, which is usually _not_ enough to keep `lens_differ` at 0, so the skipping of length grouping added with this PR probably almost did not occur in that test, and the speed difference might thus be spurious. I'd expect real speedup e.g. when running a mask only or incremental mode, or when adding more characters to a word through a mask, e.g. with `?w?a?a?a` (in other words, with mask "amplifier" larger than GWS). Am I missing something?

Edit: I missed the word "not", now inserted.

solardiz:
> I also don't fully understand what @claudioandre-br's benchmarks in the first comment here (are expected to) actually show.

Oh, I realize I missed `--min-len=7 --max-len=7`. This explains it - yes, this actually benchmarks the skipping of grouping by length, then.

magnumripper:
> > could lean slightly more towards keeping LWS low.
> 
> Why? In my experiments with md5crypt-opencl, it appeared sufficient (for current GPUs) to avoid length discrepancies within each aligned 64-password group regardless of LWS value (even for LWS higher than 64). This suggests that when length discrepancies are not avoided, it does not matter what LWS is, either. No?

I'm not sure what that experience is indicating, but my thinking was just that the longest password in "a batch" will hit performance of all other ones bc lock-step - and it would seem logical that "a batch" here means all others within a work group (LWS) but that may not be the case, I have no idea really!

solardiz (approved):
I still think we need to move the memory (de)allocation out of `crypt_all()`, but I'm going to merge this PR as-is and let's improve this code further separately.

solardiz:
Another optimization to make separately is having the loop by length determine the actual maximum length on the first iteration (or we can have `set_key()` do that), then loop until that length only (not until `PLAINTEXT_LENGTH`).

We can even fill an array of Booleans or a bitmap (is `uint64_t` enough here?) indicating which lengths are actually present this time, and jump over lengths that are not present. In fact, this could replace the `lens_differ` logic I had suggested - instead, we'd only need to check whether the bitmap has only one bit set or more.

I think I'll experiment with that.

magnumripper:
> > > could lean slightly more towards keeping LWS low.
> > 
> > 
> > Why? In my experiments with md5crypt-opencl, it appeared sufficient (for current GPUs) to avoid length discrepancies within each aligned 64-password group regardless of LWS value (even for LWS higher than 64). This suggests that when length discrepancies are not avoided, it does not matter what LWS is, either. No?
> 
> I'm not sure what that experience is indicating, but my thinking was just that the longest password in "a batch" will hit performance of all other ones bc lock-step - and it would seem logical that "a batch" here means all others within a work group (LWS) but that may not be the case, I have no idea really!

In fact, if *this* doesn't relate to LWS, then I have absolutely no idea what would, and why we need to tune it?! I thought I had come to some understanding but I'm not so sure now.

magnumripper:
Do we need all these memcpy's? Can't we do the same just with juggling pointers?

ghost:
> Do we need all these memcpy's? Can't we do the same just with juggling pointers?

We will copy the buffer to a device. The "pointers" do not exist there.
You can transfer the indexes and access global memory in a random order (it is expected to degrade performance).

solardiz:
@magnumripper I think both of us are right about what LWS is (not), in a way. In md5crypt-opencl, there's length-dependent branching aka code divergence, but the overall amount of processing is roughly the same regardless of length (for lengths up to 15, which are the only ones supported there). This is probably why only divergence within blocks of 64 mattered even at higher LWS - it was still efficient to run those multiple thread groups concurrently (interleaved, pipelined) even if they were at different code addresses. For RAR, things may well be as you expect them to be. So try and see.

PR #4682, opened 2021-04-22
Reduce sha[256/512]crypt kernel calls

ghost:
My vote is to merge AFTER the release. It is (kind of) invasive and needs tuning.

### DIFF
--- a/run/opencl/opencl_sha256crypt.h
+++ b/run/opencl/opencl_sha256crypt.h
@@ -24,7 +24,7 @@
 #define PLAINTEXT_ARRAY         (PLAINTEXT_LENGTH / 4)
 #define SEED                    1024
 #define STEP                    0
-#define HASH_LOOPS              (7*3*2)
+#define HASH_LOOPS              (7*3*2 * 25)
 
 #define KEYS_PER_CORE_CPU       128
 #define KEYS_PER_CORE_GPU       1
--- a/run/opencl/opencl_sha512crypt.h
+++ b/run/opencl/opencl_sha512crypt.h
@@ -42,7 +42,7 @@
 #define SEED                    1024
 #define STEP                    0
 
-#define HASH_LOOPS              (7*3*2)
+#define HASH_LOOPS              (7*3*2 * 25)
 
 #define KEYS_PER_CORE_CPU       128
 #define KEYS_PER_CORE_GPU       1


solardiz:
This one looks simple enough to merge before release. Can you benchmark it separately? (My uses of GPUs on "super" are currently paused.)

solardiz:
Based on my findings yesterday, I think the factor of 25 might be too large for these formats, though - they appear to benefit from high GWS, and with a large factor we'll be bumping into 200ms at lower GWS now. A larger `HASH_LOOPS` worked well for bitcoin-opencl because it's so simple and doesn't use global memory inside its loop, thus doesn't need larger GWS, but these formats do need larger GWS to compensate for global memory latencies.

Maybe our reference for these formats should be md5crypt's 1000 iterations, which we don't split, but we need to scale this for MD5 vs. SHA-2 performance differences. So maybe use additional factor of 10 for sha256crypt (2*7*3*10 = 420) and 5 for sha512crypt (210). Just guessing.

ghost:
> My uses of GPUs on "super" are currently paused.

But is it using (a lot of) GPU memory.

solardiz:
> > My uses of GPUs on "super" are currently paused.
> 
> But is it using (a lot of) GPU memory.

I think there's more than enough left for testing these formats, no?

ghost:
It is hurting too fast and I'll have to change more code (the PR is basically wrong).

* Vega
```
rm ../run/opencl/*.bin && rm -rf ../run/*.pot ~/.nv

../run/john-upstream --test=5 --verb=6 --format=sha512crypt-opencl -dev=1 #LWS=256 GWS=262144
Raw:	129134 c/s real, 4915K c/s virtual
Raw:	128923 c/s real, 4915K c/s virtual

# -----------------------------------------------------------------
rm ../run/opencl/*.bin && rm -rf ../run/*.pot ~/.nv
#change files plus git diff
make -s distclean ; ./configure && make clean && make -sj8
mv ../run/john ../run/john-2

../run/john-2 --test=5 --verb=6 --format=sha512crypt-opencl -dev=1 #TestDB LWS=256 GWS=32768 (128 blocks) DONE
Raw:	101803 c/s real, 1872K c/s virtual
```

* Titan X (GWS pretty smaller GWS, 1k loss).

* 1080 has a good gain
```
../run/john-upstream --test=5 --verb=6 --format=sha512crypt-opencl -dev=4 #LWS=128 GWS=81920 LWS=128 GWS=163840
Raw:	144848 c/s real, 144564 c/s virtual, Dev#4 util: 100%
Raw:	144479 c/s real, 144479 c/s virtual, Dev#4 util: 100%

# -----------------------------------------------------------------
../run/john-2 --test=5 --verb=6 --format=sha512crypt-opencl -dev=4
Raw:	150529 c/s real, 149949 c/s virtual, Dev#4 util: 100% # LWS=512 GWS=40960
```

I'm closing for now, it takes too much effort.

solardiz:
> It is hurting too fast

Were your results above for this PR as posted, with the 25 multiplier? Maybe simply try smaller multipliers, like 5 or 10?

ghost:
It is 2 not 25 (note the binary name). Good news is that I realized what I did wrong. Now, Vega also has a small gain.

1080 has a real gain.
```
../run/john-2 --test=5 --verb=6 --format=sha512crypt-opencl -dev=5 #
TestDB LWS=512 GWS=81920 (160 blocks) DONE
Raw:	150588 c/s real, 150311 c/s virtual, Dev#4 util: 100%

../run/john-2 alltests.in -word=solucao.in --format=sha512crypt-opencl -dev=4
6g 0:00:00:11 DONE (2021-04-22 18:12) 0.5361g/s 457.6p/s 1830c/s 2745C/s Dev#4:54°C U*U*U*U*
```

It is not good for Titan X or Titan (basically, what I already know). New boards can benefit.

ghost:
4 is no better on Vega, it lost almost all of its gain at 1080.

ghost:
2 hurts a little on Tahiti

PR #4683, opened 2021-04-22
OpenCL: Driver bug workaround for MacOS w/ AMD Vega

magnumripper:
As usual, this was found by "throw stones at it and see what sticks". The `W` array is normally re-used for this block, and isn't used later. Yet adding this redundant local declaration stops AMD Radeon Pro Vega 20 from failing self-test under macOS (Big Sur 11.2.3), even though it should be a no-op after any optimizer.

### DIFF
--- a/run/opencl/oldoffice_kernel.cl
+++ b/run/opencl/oldoffice_kernel.cl
@@ -497,6 +497,9 @@ void oldoffice_sha1(const nt_buffer_t *nt_buffer,
 		else if (cs->type != 3 || !cs->has_extra)
 			*result = 1;
 		else {
+#if __OS_X__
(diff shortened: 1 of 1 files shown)


solardiz (approved):
OK with me, is limited to macOS anyway. Some C compilers warn when a same-name identifier shadows a higher scope declaration, but I haven't heard of such warnings in OpenCL so far, and I suppose magnum would have noticed if this were the case on macOS.

magnumripper:
I saw no warning, but I'll refactor anyway just for good measure. Thanks!

solardiz:
> I'll refactor anyway just for good measure.

Given that this is a conditional workaround, I think it's better to leave it as-is, not over-complicate it.

Issue #4684, opened 2021-04-22
POCL testing (Ubuntu 20.04)


magnumripper:
Our current README-OPENCL says POCL is not supported (but will work with some formats). I'll do some testing and see current state, and perhaps come up with some tweaks. Perhaps we can stop saying it's not supported?

This is POCL 1.4
```
Platform #0 name: Portable Computing Language, version: OpenCL 1.2 pocl 1.4, None+Asserts, LLVM 9.0.1, RELOC, SLEEF, DISTRO, POCL_DEBUG
    Device #0 (1) name:     pthread-Intel(R) Core(TM) i7-4790 CPU @ 3.60GHz
    Device vendor:          GenuineIntel
    Device type:            CPU (LE)
    Device version:         OpenCL 1.2 pocl HSTR: pthread-x86_64-pc-linux-gnu-haswell
    Driver version:         1.4 
    Native vector widths:   char 16, short 16, int 8, long 4
    Preferred vector width: char 16, short 16, int 8, long 4
    Global Memory:          30005 MiB
    Global Memory Cache:    8 MiB
    Local Memory:           4 MiB (Global)
    Constant Buffer size:   4 MiB
    Max memory alloc. size: 8 GiB
    Max clock (MHz):        4000
    Profiling timer res.:   1 ns
    Max Work Group Size:    4096
    Parallel compute cores: 8
    Speed index:            128000
```

The ones that fail or warn:
```
Device 1: pthread-Intel(R) Core(TM) i7-4790 CPU @ 3.60GHz
Testing: sha256crypt-opencl, crypt(3) $5$ (rounds=5000) [SHA256 OpenCL]... 1 warning generated.
Build log: warning: /home/magnum/.cache/pocl/kcache/tempfile-14-89-0e-36-c9.cl:265:13: loop not unrolled: the optimizer was unable to perform the requested transformation; the transformation might be disabled or specified as part of an unsupported transformation ordering

PASS
Testing: sha512crypt-opencl, crypt(3) $6$ (rounds=5000) [SHA512 OpenCL]... 1 warning generated.
Build log: warning: /home/magnum/.cache/pocl/kcache/tempfile-0c-de-2e-e2-26.cl:232:13: loop not unrolled: the optimizer was unable to perform the requested transformation; the transformation might be disabled or specified as part of an unsupported transformation ordering

PASS
[...]

PR #4686, opened 2021-04-22
OpenCL: POCL fixes

magnumripper:
This stops GPG and DEScrypt OpenCL formats from crashing under POCL (as tested on Ubuntu 20.04) - making all formats pass self-test. It also adds a quick bail-out for when LWS auto-tune takes too long.

### DIFF
--- a/doc/README-OPENCL
+++ b/doc/README-OPENCL
@@ -17,8 +17,9 @@ distribution, only the vendor-supplied drivers support OpenCL.  Either
 install fglrx (for old AMD cards) or nvidia dkms package or go directly
 with the ones provided by nvidia and ATI.
 
-Notice: Beignet, Mesa, and POCL are not officially supported, but may
-be usable in some OpenCL formats.
+Notice: Beignet and Mesa are not officially supported, but may be usable
+with some OpenCL formats.  POCL on the other hand has now been seen passing
+all self-tests (tested on Ubuntu 20.04 with bleeding-jumbo as of 2021/04/22).
 
 You can also use OpenCL with CPU, mostly useful if you have several
 (or loads of) cores.  This sometimes outperforms the CPU-only formats
--- a/run/opencl/opencl_DES_kernel_params.h
+++ b/run/opencl/opencl_DES_kernel_params.h
@@ -11,10 +11,10 @@ typedef unsigned WORD vtype;
  * Nvidia can build either kernel but GTX980 is significantly faster with the
  * "safe goto" version (7% faster for one salt, 16% for many salts).
  *
- * OSX' Intel HD4000 driver [1.2(Sep25 2014 22:26:04)] fails building the
- * "fast goto" version.
+ * macOS in general has problems building the "fast goto" version, and the
+ * same goes for MESA and POCL.
  */
-#if nvidia_sm_5x(DEVICE_INFO) || gpu_intel(DEVICE_INFO) || __MESA__ ||  \
+#if nvidia_sm_5x(DEVICE_INFO) || __OS_X__ || __MESA__ || __POCL__ || \
 	(gpu_amd(DEVICE_INFO) && DEV_VER_MAJOR >= 1573 && !defined(__Tahiti__)) || \
 	(gpu_amd(DEVICE_INFO) && DEV_VER_MAJOR >= 1702)
 //#warning Using 'safe goto' kernel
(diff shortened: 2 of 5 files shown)


src/opencl_common.c
```diff
@@ -1867,6 +1867,8 @@ void opencl_find_best_lws(size_t group_size_limit, int sequential_id,
 			fprintf(stderr, " %ss%s\n", ns2string(sumRunTime),
 			    ((double)(sumRunTime) / kernelExecTimeNs < 0.997)
 			        ? "+" : "");
+		if (sumRunTime > 2 * kernelExecTimeNs)
+			break;
```
magnumripper, line 1871: This is for the issue [reported here](https://github.com/openwall/john/issues/4684#issuecomment-824831938), which was intermittent. I figured if we *ever* get over twice as long duration from doubling LWS but with same GWS, we can't be on the way to anything good, right?

run/opencl/opencl_DES_kernel_params.h
```diff
@@ -11,10 +11,10 @@ typedef unsigned WORD vtype;
  * Nvidia can build either kernel but GTX980 is significantly faster with the
  * "safe goto" version (7% faster for one salt, 16% for many salts).
  *
- * OSX' Intel HD4000 driver [1.2(Sep25 2014 22:26:04)] fails building the
- * "fast goto" version.
+ * macOS in general has problems building the "fast goto" version, and the
+ * same goes for MESA and POCL.
  */
-#if nvidia_sm_5x(DEVICE_INFO) || gpu_intel(DEVICE_INFO) || __MESA__ ||  \
+#if nvidia_sm_5x(DEVICE_INFO) || __OS_X__ || __MESA__ || __POCL__ || \
```
solardiz, line 17: Need to test how this affects Intel GPUs under non-macOS, or/and don't drop `gpu_intel(DEVICE_INFO)` in the same PR/commit. Can you test on "well"?

src/opencl_common.c
solardiz: Doubled duration with same GWS means twice lower speeds, so, yes, I guess it means the doubled LWS wasn't a move in the right direction. I wonder, though, whether exceptions exist with LWS/GWS values way out of the device's optimal range, where further GWS adjustment would bring them into the optimal range. Hopefully not.

Maybe this should be a separate commit with its own description.

doc/README-OPENCL
```diff
@@ -17,8 +17,9 @@ distribution, only the vendor-supplied drivers support OpenCL.  Either
 install fglrx (for old AMD cards) or nvidia dkms package or go directly
 with the ones provided by nvidia and ATI.
 
-Notice: Beignet, Mesa, and POCL are not officially supported, but may
-be usable in some OpenCL formats.
+Notice: Beignet and Mesa are not officially supported, but may be usable
+with some OpenCL formats.  POCL on the other hand has now been seen passing
+all self-tests (tested on Ubuntu 20.04).
```
solardiz: I'd add: "tested on Ubuntu 20.04 with bleeding-jumbo as of 2021/04/22".

solardiz (approved):
Looks good to me overall, but I made some comments.

run/opencl/opencl_DES_kernel_params.h
magnumripper: Done now, There's virtually no difference in speed.

solardiz (approved):

src/options.c
```diff
@@ -63,7 +63,8 @@ static char *costs_str;
 
 /* Common req_clr for --test, --test-full and --stress-test */
 #define TEST_REQ_CLR (~FLG_TEST_SET & ~FLG_FORMAT & ~FLG_SAVEMEM & ~FLG_MASK_CHK & ~FLG_NO_MASK_BENCH & \
-                      ~FLG_VERBOSITY & ~FLG_INPUT_ENC & ~FLG_SECOND_ENC & ~GETOPT_FLAGS & ~FLG_NOTESTS)
+                      ~FLG_VERBOSITY & ~FLG_INPUT_ENC & ~FLG_SECOND_ENC & ~GETOPT_FLAGS & ~FLG_NOTESTS & \
+                      ~FLG_SCALAR & ~FLG_VECTOR)
```
magnumripper, line 67: This added as well now (separate commit).

PR #4687, opened 2021-04-22
Reduce sha[256/512]crypt kernel calls

ghost:
I need to open a new PR, I can't reopen the previous one.

My vote is to merge AFTER the release. It is (a kind of) invasive and needs real testing.

- Vega: gained 2-4k c/s
- 1080: gained 6-7k c/s

### DIFF
--- a/run/opencl/cryptsha256_kernel_GPU.cl
+++ b/run/opencl/cryptsha256_kernel_GPU.cl
@@ -722,22 +722,22 @@ inline void sha256_crypt(
 #elif (UNROLL_LOOP & (1 << 19))
     #pragma unroll 2
 #endif
-    for (uint i = 0U; i < HASH_LOOPS; i++) {
-        total = work_memory[OFFSET(loop_index[i], 30)];
+    for (uint i = 0U, z = 0U; i < HASH_LOOPS; i++) {
+        total = work_memory[OFFSET(loop_index[z], 30)];
 
         if (i & 1) {
 	    #pragma unroll
 	    for (uint32_t j = 8U; j < 16U; j++)
 		w[j] = 0;
 
-            w[0] = work_memory[OFFSET(loop_index[i], 0)];
-            w[1] = work_memory[OFFSET(loop_index[i], 1)];
-            w[2] = work_memory[OFFSET(loop_index[i], 2)];
-            w[3] = work_memory[OFFSET(loop_index[i], 3)];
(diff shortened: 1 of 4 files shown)


solardiz:
I think instead of `% LOOP_SIZE`, you need to add a separate `uint` variable going over the range of 0 to `LOOP_SIZE - 1`. IIRC, we have that in md5crypt-opencl - please take a look.

solardiz:
> you need to add a separate `uint` variable going over the range of 0 to `LOOP_SIZE - 1`.

BTW, since you already have two separate code blocks depending on `i & 1`, you'll need to check that variable for reaching `LOOP_SIZE` in only one of these two blocks.

In fact, you could check the loop count limit (currently `i` against `HASH_LOOPS`) in only one of these two blocks, too - so once per two iterations - not every iteration like you do now.

solardiz:
One way to write this is through having two nested loops - inner going until `LOOP_SIZE` and checking that every second iteration, outer going until `HASH_LOOPS / LOOP_SIZE`.

BTW, looking at the diffs on GitHub, I also notice there's some incorrect indentation - maybe inconsistent use of tabs vs. spaces.

ghost:
> I also notice there's some incorrect indentation - maybe inconsistent use of tabs vs. spaces.

Yes, the whole file needs to be "formatted". The PR maintains the indentation in lines it touches.

solardiz:
> Yes, the whole file needs to be "formatted". The PR maintains the indentation in lines it touches.

It'd make sense to have this as two separate commits (maybe in one PR) - functional changes in one and indentation corrections in the other.

ghost:
Is it better now?

solardiz:
Yes, it's better, except that I suggest you avoid the `else` e.g. by changing

```c
    if (z == (LOOP_SIZE - 1))
        z = 0;
    else
        z++;
```

to

```c
    if (++z == LOOP_SIZE)
        z = 0;
```

Also, can't you easily have this piece run just once per two loop iterations, by having it inside an `else` after one of your `if (i & 1)` checks?

solardiz (approved):
Looks good enough now. I'd like to merge this. OK with you, Claudio?

ghost:
That is ok.

solardiz:
Tested on the old Titan Kepler using 512 same-salt sha512crypt hashes with the default 5000 iterations corresponding to passwords of length 7. Ditto for sha256crypt. All cracked as expected (wordlist, mask producing ~1 million candidates).

PR #4694, opened 2021-04-24
Bitcoin pure

magnumripper:
This is just a first draft, just threw it together until it worked at all.


src/opencl_bitcoin_fmt_plug.c
```diff
@@ -336,44 +363,31 @@ static void *get_salt(char *ciphertext)
...
-#if 1
 	size_t *lws = local_work_size ? &local_work_size : NULL;
 
 	global_work_size = GET_NEXT_MULTIPLE(count, local_work_size);
 
 	// Copy data to gpu
-	BENCH_CLERROR(clEnqueueWriteBuffer(queue[gpu_id], mem_inout, CL_FALSE, 0,
-		inoutsize, inoutbuffer, 0, NULL, multi_profilingEvent[0]),
+	BENCH_CLERROR(clEnqueueWriteBuffer(queue[gpu_id], mem_in, CL_FALSE, 0,
+		in_size, saved_key, 0, NULL, multi_profilingEvent[0]),
```
magnumripper: This key transfer should be made conditional "if keys dirty"

magnumripper:
> This is just a first draft

I think I'm done with it now - all tests look fine and there isn't much you can do with speed - only the loop kernel matters.

magnumripper:
Just to be clear, I did this mostly as an exercise and to see how it ended up. Merging it isn't quite a no-brainer: It doesn't really add any speed unless you've got very little CPU power *and* lots of GPU power. So the con is obviously it adds some complexity. I'd merge it anyway "for good measure" but I wouldn't really mind if you decide not to.

solardiz:
@magnumripper Looking at your latest force push:

```diff
		BENCH_CLERROR(clEnqueueWriteBuffer(queue[gpu_id], mem_in, CL_FALSE, 0,
-			in_size, saved_key, 0, NULL, multi_profilingEvent[0]),
+			sizeof(password_t) * global_work_size, saved_key, 0, NULL, multi_profilingEvent[0]),
			"Copy data to gpu");
```

Shouldn't we also limit the transfer back to host to actual current GWS?

IIRC, over-transferring like this (including some uninitialized and later unused data in there) caused some of the failures we recently saw in testing on the old GTX 570 (somehow it was quite good at detecting those).

Also, perhaps my initially merged revision of bitcoin-opencl has this sort of problem, too? And perhaps sspr-opencl does, too, as that's what I was looking at when implementing this format.

run/opencl/bitcoin_kernel.cl
```diff
@@ -13,26 +13,59 @@
...
+} password_t;
+
+typedef struct custom_salt {
+	uchar cry_master[SZ];
+	int cry_master_length;
+	uchar cry_salt[SZ];
+	int cry_salt_length;
+	int cry_rounds;
+	int final_block_fill;
+} salt_t;
```
solardiz, line 34: BTW, it's kind of wasteful to use `SZ` here - we could use 32 for `cry_master` (only decoding/storing its ending) and 16 for `cry_salt` - but it's a change we could make to the CPU format too and then share the struct between the formats, so should probably be a separate commit/PR. There's value in having the initial move of everything to GPU done in a straightforward way like you did it here. Besides, this won't provide any measurable speedup as only one instance of the salt struct is transferred per kernel invocation.

run/opencl/bitcoin_kernel.cl
```diff
@@ -13,26 +13,59 @@
...
+	hash512_t key_iv;
+	SHA512_CTX sha_ctx;
+	uchar buf[SZ];
+
+	SHA512_Init(&sha_ctx);
+	memcpy_gp(buf, key[gid].c, key[gid].len);
+	SHA512_Update(&sha_ctx, buf, key[gid].len);
+	memcpy_cp(buf, salt->cry_salt, salt->cry_salt_length);
+	SHA512_Update(&sha_ctx, buf, salt->cry_salt_length);
+	SHA512_Final(key_iv.b, &sha_ctx);
```
solardiz, line 54: BTW, we should be able to hard-code 16 for `cry_salt_length` (and not transfer it from host) for all currently supported wallet types. (We also support 36 for the initial salt length, but we only use this to indicate that the Nexus legacy wallet also uses different mkey size and thus we expect different padding, but this is already encoded in `final_block_fill`, so doesn't need any other logic from OpenCL side.) Again, this simplification, if we make it, should be in a separate commit/PR.

src/opencl_bitcoin_fmt_plug.c
```diff
@@ -200,14 +218,22 @@ static void init(struct fmt_main *_self)
 static void reset(struct db_main *db)
 {
 	if (!program[gpu_id]) {
-		opencl_init("$JOHN/opencl/bitcoin_kernel.cl", gpu_id, "");
+		char build_opts[64];
+
+		snprintf(build_opts, sizeof(build_opts), "-DSZ=%u -DPLAINTEXT_LENGTH=%u", SZ, PLAINTEXT_LENGTH);
```
solardiz, line 223: Ah, here's `SZ`. I think we should clean this up separately.

src/opencl_bitcoin_fmt_plug.c
```diff
@@ -459,17 +443,19 @@ static int cmp_one(void *binary, int index)
 
 static int cmp_exact(char *source, int index)
 {
-	return cracked[index];
+	return 1;
 }
 
 static void bitcoin_set_key(char *key, int index)
```
solardiz: BTW, no reason for `bitcoin_` in the `bitcoin_set_key` name if we don't share this function with the CPU format.

solardiz (approved):
Looks OK to me, can be cleaned up further separately. There's also the separate question I asked about limiting the transfer to host to current GWS.

magnumripper:
> Shouldn't we also limit the transfer back to host to actual current GWS?

Yeah that would make sense, but it would hardly make any difference (even this change I did was pointless for overall speed).

> IIRC, over-transferring like this (including some uninitialized and later unused data in there) caused some of the failures we recently saw in testing on the old GTX 570 (somehow it was quite good at detecting those).

Yes but as long as we allocate them with calloc (which we do here) they aren't uninitialized. In fact we even use `mem_calloc_align` which isn't even using system's calloc (which could/should do clever tricks with virtual memory that *might* be troublesome for a buggy driver) but normal calloc + memset, so it is in *no* way uninitialized.

solardiz:
> as long as we allocate them with calloc (which we do here) they aren't uninitialized.

I think they're uninitialized on device side: we only transfer up to actual GWS to there, and only compute up to actual GWS in there, so when we try to transfer more data back (up to maximum GWS) we end up reading uninitialized data on device side, which is probably undefined behavior even if we don't use this over-read data on host then.

magnumripper:
> > as long as we allocate them with calloc (which we do here) they aren't uninitialized.
> 
> I think they're uninitialized on device side: we only transfer up to actual GWS to there, and only compute up to actual GWS in there, so when we try to transfer more data back (up to maximum GWS) we end up reading uninitialized data on device side, which is probably undefined behavior even if we don't use this over-read data on host then.

Ah yes of course... I'll amend it for good measure.

src/opencl_bitcoin_fmt_plug.c
```diff
@@ -336,43 +364,37 @@ static void *get_salt(char *ciphertext)
...
 	size_t *lws = local_work_size ? &local_work_size : NULL;
 
 	global_work_size = GET_NEXT_MULTIPLE(count, local_work_size);
 
-	// Copy data to gpu
-	BENCH_CLERROR(clEnqueueWriteBuffer(queue[gpu_id], mem_inout, CL_FALSE, 0,
-		inoutsize, inoutbuffer, 0, NULL, multi_profilingEvent[0]),
-		"Copy data to gpu");
+	in_size = sizeof(password_t) * global_work_size;
+	cracked_size = sizeof(*cracked) * global_work_size;
```
magnumripper, line 383: Since we have these globals, it felt logical to update them and use them here as well.

solardiz: OK, since we don't need them when freeing.

solardiz (approved):

PR #4695, opened 2021-04-25
Incremental: Do not print length figures until they are final

### LINKED ISSUE #4688: Max length 125 wrongly reported on restored --incremental runs
```
$ ./john -se=tmp -inc -no-log -min-len=5 -stdo > /dev/null
8196582p 0:00:00:02 0.00% 3812Kp/s tj4mcl
37815389p 0:00:00:04 0.00% 9046Kp/s 2bgjs
Session aborted
$ ./john -rest=tmp > /dev/null
Proceeding with incremental:ASCII, lengths: 5-125
```

This is with `MaxLen = 13` for the mode in `john.conf`, and with compile-time limit for incremental mode at 24. So clearly 125 is wrong. I hope it's just the reporting, but am not sure.

magnumripper:
Closes #4688

I did more dry-running than actual testing but I did test it as well.

Note that logged lengths are just incremental's part of the total, while printed lengths take hybrid mask in account.

### DIFF
--- a/src/inc.c
+++ b/src/inc.c
@@ -496,20 +496,6 @@ void do_incremental_crack(struct db_main *db, const char *mode)
 	if (john_main_process)
 		log_event("Proceeding with \"incremental\" mode: %.100s", mode);
 
-	if ((options.flags & FLG_BATCH_CHK || rec_restored) && john_main_process) {
-		fprintf(stderr, "Proceeding with incremental:%s", mode);
-		if (options.flags & FLG_MASK_CHK)
-			fprintf(stderr, ", hybrid mask:%s", options.mask ?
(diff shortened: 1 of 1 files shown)


src/inc.c
```diff
@@ -578,6 +563,20 @@ void do_incremental_crack(struct db_main *db, const char *mode)
 		max_length = min_length;
 	}
 
+	if ((options.flags & FLG_BATCH_CHK || rec_restored) && john_main_process) {
```
solardiz: I'd also add braces around `options.flags & FLG_BATCH_CHK` while we're at it.

solardiz (approved):
I mostly rely on your testing, but this looks OK to me.

PR #4698, opened 2021-04-26
OpenCL: sha[256/512]crypt group by length clean up

ghost:
CI check.

### DIFF
--- a/src/opencl_sha256crypt_fmt_plug.c
+++ b/src/opencl_sha256crypt_fmt_plug.c
@@ -44,6 +44,11 @@ static sha256_password *plain_sorted;   // sorted list (by plaintext len)
 static sha256_hash *calculated_hash;    // calculated hashes
 static sha256_hash *computed_hash;  // calculated hashes (from plain_sorted)
 
+//To connect the sorted and unsorted plaintext lists
+static unsigned int *indices;
+static size_t indices_size;
+static uint32_t bitmap_of_lens; // what plaintext sizes do we have?
+
 static cl_mem salt_buffer;      //Salt information.
 static cl_mem pass_buffer;      //Plaintext buffer.
 static cl_mem hash_buffer;      //Hash keys (output).
@@ -53,13 +58,17 @@ static struct fmt_main *self;
 
 static cl_kernel prepare_kernel, preproc_kernel, final_kernel;
 
-static int new_keys, first_len, lens_differ, source_in_use;
+static int new_keys, source_in_use;
 static int split_events[3] = { 1, 6, 7 };
 
 //This file contains auto-tuning routine(s). It has to be included after formats definitions.
 #include "opencl_autotune.h"
 
 static void release_kernel();
+
+#if (PLAINTEXT_LENGTH > 31) //Can't use sizeof(uint32_t)
+#error "Review bitmap_of_lens size"
+#endif
 
 /* ------- Helper functions ------- */
 static size_t get_task_max_work_group_size()
@@ -272,6 +281,14 @@ static int salt_hash(void *salt)
 }
 
 /* ------- Key functions ------- */
+static void clear_keys(void)
+{
+	/* When a new group of keys begins to be sent to the OpenCL device
+	 * clear the information about previously sent keys data.
+	 */
+	bitmap_of_lens = 0;
+}
+
 static void set_key(char *key, int index)
 {
 	int len;
@@ -285,11 +302,8 @@ static void set_key(char *key, int index)
 	memcpy(plaintext[index].pass, key, len);
 	plaintext[index].length = len;
 
-	if (new_keys)
-		lens_differ |= len ^ first_len;
-	else
-		first_len = len;
 	new_keys = 1;
+	bitmap_of_lens |= (1 << len);
 }
 
 static char *get_key(int index)
@@ -468,6 +482,9 @@ static void reset(struct db_main *db)
 	int major, minor;
 	unsigned long long int max_run_time;
 
+	new_keys = 0;
+	bitmap_of_lens = 0;
+
 	source_in_use = device_info[gpu_id];
 
 	//Initialize openCL tuning (library) for this format.
@@ -515,6 +532,8 @@ static void done(void)
 	if (program[gpu_id]) {
 		release_clobj();
 		release_kernel();
+		MEM_FREE(indices);
+		indices_size = 0;
 	}
 }
 
@@ -545,7 +564,6 @@ static int crypt_all(int *pcount, struct db_salt *_salt)
 {
 	int count = *pcount;
 	int index;
-	int *indices = NULL;  // relationship between sorted and unsorted plaintext list
 	size_t gws;
 	size_t *lws = local_work_size ? &local_work_size : NULL;
 	sha256_password *input_candidates;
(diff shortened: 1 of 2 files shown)


src/opencl_sha256crypt_fmt_plug.c
```diff
@@ -44,6 +44,10 @@ static sha256_password *plain_sorted;   // sorted list (by plaintext len)
 static sha256_hash *calculated_hash;    // calculated hashes
 static sha256_hash *computed_hash;  // calculated hashes (from plain_sorted)
 
+static int
+*indices;            // relationship between sorted and unsorted plaintext list
```
solardiz: Why wrap the line here?

src/opencl_sha256crypt_fmt_plug.c
```diff
@@ -44,6 +44,10 @@ static sha256_password *plain_sorted;   // sorted list (by plaintext len)
...
+unsigned int bitmap_of_lens; // what plaintext sizes do we have?
```
solardiz: Add `static`, and I'd use an explicit width type like `uint32_t`. Also need a comment or ideally compile-time check that `PLAINTEXT_LENGTH` isn't larger than supported by this type (so not larger than e.g. 31 if we allocate bit 0 for length 0).

src/opencl_sha256crypt_fmt_plug.c
```diff
@@ -290,6 +294,7 @@ static void set_key(char *key, int index)
 	else
 		first_len = len;
 	new_keys = 1;
+	bitmap_of_lens |= (1 << len);
```
solardiz, line 306: We should be able to drop `new_keys`, `lens_differ`, and `first_len`. We really only need `bitmap_of_lens` now.

src/opencl_sha256crypt_fmt_plug.c
```diff
@@ -561,19 +572,25 @@ static int crypt_all(int *pcount, struct db_salt *_salt)
 		int tot_todo = 0, len;
 
 		if (lens_differ) {
-			indices = mem_alloc(gws * sizeof(int));
+
+			if (gws < previous_gws) {
```
solardiz: I think you got this check backwards - we need to reallocate if the previous size is no longer sufficient, not when it's excessive. Also drop the empty line above this check.

src/opencl_sha256crypt_fmt_plug.c
```diff
@@ -561,19 +572,25 @@ static int crypt_all(int *pcount, struct db_salt *_salt)
 		int tot_todo = 0, len;
 
 		if (lens_differ) {
```
solardiz: `if (bitmap_of_lens & (bitmap_of_lens - 1))` { /* if more than one bit set */

src/opencl_sha256crypt_fmt_plug.c
```diff
@@ -561,19 +572,25 @@ static int crypt_all(int *pcount, struct db_salt *_salt)
...
+				previous_gws = gws;
+				MEM_FREE(indices);
+				indices = mem_alloc(gws * sizeof(int));
+			}
 
 			for (len = 0; len <= PLAINTEXT_LENGTH; len++) {
-				for (index = 0; index < count; index++) {
-					if (plaintext[index].length == len)
-						indices[tot_todo++] = index;
+				if (bitmap_of_lens & (1 << len))
```
solardiz:
```c
if (!(bitmap_of_lens >> len))
	break;
if ((bitmap_of_lens >> len) & 1)
```

This could probably be sped up a tiny bit by first copying `bitmap_of_len` to a temporary loop variable before the loop, then using `>>= 1` on it (then no variable shift count). But since we have variable shift count inside `set_key()` anyway, which executes many more times, the point is moot.

solardiz (changes requested):
I only commented on the sha256crypt part. The same comments apply to sha512crypt.

solardiz:
> CI check.

Somehow CI isn't eager to check this, but I reviewed and I think I found a bug requiring an update anyway (the reallocation check).

ghost:
It is buggy. I'm reviewing sha256.

src/opencl_sha256crypt_fmt_plug.c
solardiz: Let me clarify: I think we can drop even `new_keys` because we always end up setting `bitmap_of_lens` to non-zero if `set_keys()` was called at least once. So we can just check this one for non-zero instead. And we also reset it to 0 after we've first processed the new batch of keys in `crypt_all()`.

solardiz:
Maybe rename `previous_gws` to `indices_size` since that's what it reflects and what we use it for.

src/opencl_sha256crypt_fmt_plug.c
ghost: > Let me clarify: I think we can drop even `new_keys` because we always end up setting `bitmap_of_lens` to non-zero if `set_keys()` was called at least once. So we can just check this one for non-zero instead. And we also reset it to 0 after we've first processed the new batch of keys in `crypt_all()`.

It is no longer the case

src/opencl_sha256crypt_fmt_plug.c
ghost: You should review the code in this matter.

ghost:
I think it is done now.

src/opencl_sha256crypt_fmt_plug.c
```diff
@@ -44,6 +44,11 @@ static sha256_password *plain_sorted;   // sorted list (by plaintext len)
 static sha256_hash *calculated_hash;    // calculated hashes
 static sha256_hash *computed_hash;  // calculated hashes (from plain_sorted)
 
+//To connect the sorted and unsorted plaintext lists
+static int *indices;
+static uint32_t bitmap_of_lens; // what plaintext sizes do we have?
+static size_t indices_size;
```
solardiz, line 49: Please move the `indices_size` line to be right after `indices`.

src/opencl_sha256crypt_fmt_plug.c
```diff
@@ -44,6 +44,11 @@ static sha256_password *plain_sorted;   // sorted list (by plaintext len)
 static sha256_hash *calculated_hash;    // calculated hashes
 static sha256_hash *computed_hash;  // calculated hashes (from plain_sorted)
 
+//To connect the sorted and unsorted plaintext lists
+static int *indices;
```
solardiz: I suggest we move from `int` to `unsigned int` for the indices.

src/opencl_sha256crypt_fmt_plug.c
```diff
@@ -53,14 +58,18 @@ static struct fmt_main *self;
...
-static int new_keys, first_len, lens_differ, source_in_use;
+static int new_keys, source_in_use;
 static int split_events[3] = { 1, 6, 7 };
 
 //This file contains auto-tuning routine(s). It has to be included after formats definitions.
 #include "opencl_autotune.h"
 
 static void release_kernel();
 
+#if (PLAINTEXT_LENGTH > 32) //Can't use sizeof(uint32_t)
```
solardiz: Should check `> 31` or `>= 32`.

src/opencl_sha256crypt_fmt_plug.c
```diff
@@ -545,39 +564,47 @@ static int crypt_all(int *pcount, struct db_salt *_salt)
...
 		// sort passwords by length
 		int tot_todo = 0, len;
 
-		if (lens_differ) {
-			indices = mem_alloc(gws * sizeof(int));
+		if (bitmap_of_lens & (bitmap_of_lens - 1)) {
+			if (gws > indices_size) {
+				indices_size = gws;
+				MEM_FREE(indices);
+				indices = mem_alloc(gws * sizeof(int));
```
solardiz: I think it'd be more logical to update `indices_size` after the `mem_alloc()` call. Of course, this doesn't affect anything (and the compiler can reorder anyway),  but what we're doing will then be clearer from the source.

src/opencl_sha256crypt_fmt_plug.c
```diff
@@ -545,39 +564,47 @@ static int crypt_all(int *pcount, struct db_salt *_salt)
...
+			}
 
 			for (len = 0; len <= PLAINTEXT_LENGTH; len++) {
-				for (index = 0; index < count; index++) {
-					if (plaintext[index].length == len)
-						indices[tot_todo++] = index;
-				}
+				if (!(bitmap_of_lens >> len))
+					break;
+				if (bitmap_of_lens & (1 << len))
```
solardiz: Please write this check exactly as I had suggested. Your way of writing it kills the common subexpression, which the compiler would optimize (compute it just once).

Specifically, once again:

```c
if ((bitmap_of_lens >> len) & 1)
```

solardiz (changes requested):
Looks better to me now, but I've made some suggestions for changes again.

As a side note, I now see that/why you need to keep `new_keys` separate anyway - you can't just reset `bitmap_of_lens` to 0 at the end of `crypt_all()` because on the next call you use it to re-determine which pointers to use. So dropping `new_keys` wouldn't be such a simple change, and would possibly confuse the source code. I'm OK keeping it, then.

src/opencl_sha256crypt_fmt_plug.c
```diff
@@ -545,39 +564,47 @@ static int crypt_all(int *pcount, struct db_salt *_salt)
...
 	if (new_keys) {
 		// sort passwords by length
 		int tot_todo = 0, len;
 
-		if (lens_differ) {
-			indices = mem_alloc(gws * sizeof(int));
+		if (bitmap_of_lens & (bitmap_of_lens - 1)) {
+			if (gws > indices_size) {
+				MEM_FREE(indices);
+				indices = mem_alloc(gws * sizeof(int));
```
solardiz: Let's use `sizeof(*indices)` here, so that we don't depend on the exact type and on `int` and `unsigned int` being same size (although this is guaranteed).

solardiz:
@claudioandre-br I'm probably done with my nitpicks. When correcting the last one I've just posted, please also remove `[skip ci]` from the commit message. Thanks!

solardiz (approved):

PR #4700, opened 2021-04-26
OpenCL: sha[256/512]crypt group by length fix

### LINKED ISSUE #4699: sha[256/512]crypt formats are broken
According to Travis upstream CI.

ghost:
Closes #4699.

### DIFF
--- a/src/opencl_sha256crypt_fmt_plug.c
+++ b/src/opencl_sha256crypt_fmt_plug.c
@@ -586,8 +586,9 @@ static int crypt_all(int *pcount, struct db_salt *_salt)
 		if (bitmap_of_lens & (bitmap_of_lens - 1)) {
 			if (gws > indices_size) {
 				MEM_FREE(indices);
-				indices = mem_alloc(gws * sizeof(*indices));
+				indices = mem_calloc(gws, sizeof(*indices));
 				indices_size = gws;
+
 			}
 
 			for (len = 0; len <= PLAINTEXT_LENGTH; len++) {
--- a/src/opencl_sha512crypt_fmt_plug.c
+++ b/src/opencl_sha512crypt_fmt_plug.c
@@ -602,7 +602,7 @@ static int crypt_all(int *pcount, struct db_salt *_salt)
 		if (bitmap_of_lens & (bitmap_of_lens - 1)) {
 			if (gws > indices_size) {
 				MEM_FREE(indices);
-				indices = mem_alloc(gws * sizeof(*indices));
+				indices = mem_calloc(gws, sizeof(*indices));
 				indices_size = gws;
 			}
 


magnumripper (approved):

solardiz:
> The commited code wasn't clearing the indexes buffer properly.

Why is such clearing needed? Isn't the loop supposed to set all indices that currently matter? Isn't the clearing hiding some other bug now?

ghost:
In my understanding `*pcount` could be, e.g., equal 43. But, GWS will not be 43, it will be a multiple of LWS. The loop set only indices related to  plaintexts that exist.

solardiz:
> In my understanding *pcount could be, e.g., equal 43. But, GWS will not be 43, it will be a multiple of LWS. The loop set only indices related to plaintexts that exist.

Right, but so what? Further loops that read from `indices` only go up to `count`, not `gws`. Also, we don't transfer `indices` to device.

What prompted you to make this fix/PR? Did you see any issue? Was it gone with these changes?

I currently think this PR isn't fixing any real issue.

However, we could instead want to add these loops right after the existing ones filling `plain_sorted`:

```c
			for (; index < gws; index++)
				plain_sorted[index].length = 0;
```

So that we don't get stale lengths for those round-up entries transferred to device and possibly wasting compute resources there.

OTOH, we probably also have similar potential issues with stale lengths in cases when all lengths that we did see in `set_key()` for the current batch appeared the same.

On a related note, it's nasty that we have plenty of somewhat realistic integer overflows on multiplication on systems with 32-bit `size_t`, but this is by no means limited to these formats - it's an issue we have throughout the codebase. Adding and using a calloc-like call (but not actually zeroizing when that isn't necessary) that accepts element size and element count and multiplies them with overflow checking could be beneficial.

solardiz:
OK, maybe I'm wrong this isn't fixing any real issue, as you wrote this was:

> According to Travis upstream CI.

So there was some issue. The question then is whether this is fixing it, and if so how/why.

solardiz:
Hmm, just got a segfault while testing the latest code, _with_ this fix. So we definitely still need to understand and revise this code further.

solardiz:
The segfault is in the `memcpy()` call from here:

```c
                        //Create a sorted by length candidates list.
                        for (index = 0; index < count; index++) {
                                memcpy(plain_sorted[index].pass, plaintext[indices[index]].pass,
                                        plaintext[indices[index]].length + 1);
                                plain_sorted[index].length = plaintext[indices[index]].length;
                        }
```

ghost:
Imagine I do not initialize indices, so it contains random data.

The loop below is NOT going to initialize all array elements.
```
						if (plaintext[index].length == len)
							indices[tot_todo++] = index;
```

I need to clean up remaining elements, like this, for instance:
```
			for (; index < gws; index++)
				indices[index] = 0;
```

Otherwise:
SEGFAULT when reading `plaintext[indices[index]].pass` because `indices[index]` = random data (e.g., 323554666)

ghost:
Please, show me or create an issue telling us what you are doing when you got a segfault using this fix.

solardiz:
I am not convinced by your explanation above. The loop does (or at least is supposed to) initialize all indices up to `count`. Indeed, in my testing (even where I get the segfault) `tot_todo` ends up being equal to `count`. And then I don't see us using elements above `count` anywhere. Indeed, that other loop where "my" segfault occurs only goes up to `count`.

I am still investigating what the real issue is.

solardiz:
That segfault was for maximum length password occupying the very last slot after sorting, and it was a result of the `+ 1` on the `memcpy()`. In these formats, we don't include space for a trailing NUL, nor guarantee that there's one. We maintain lengths separately. Simply dropping the `+ 1` appears to make it work. Now I'll try reverting this PR's change, and will probably send another PR with the proper fix and some cleanups.

@claudioandre-br Did this PR's change make Travis CI happy?

solardiz:
No issues testing this so far:

```diff
+++ b/src/opencl_sha256crypt_fmt_plug.c
@@ -588,6 +588,7 @@ static int crypt_all(int *pcount, struct db_salt *_salt)
                                MEM_FREE(indices);
                                indices = mem_calloc(gws, sizeof(*indices));
                                indices_size = gws;
+memset(indices, 0x55, indices_size);
 
                        }
 
@@ -603,7 +604,7 @@ static int crypt_all(int *pcount, struct db_salt *_salt)
                        //Create a sorted by length candidates list.
                        for (index = 0; index < count; index++) {
                                memcpy(plain_sorted[index].pass, plaintext[indices[index]].pass,
-                                       plaintext[indices[index]].length + 1);
+                                       plaintext[indices[index]].length);
                                plain_sorted[index].length = plaintext[indices[index]].length;
                        }
                }
```

ghost:
> @claudioandre-br Did this PR's change make Travis CI happy?

Yes.

solardiz:
> because `indices[index]` = random data (e.g., 323554666)

Did you actually see that in there?

ghost:
Yes, but I don't really care about 256crypt. I only test using 512crypt. I need to apply:

```
diff --git a/run/opencl/opencl_sha256crypt.h b/run/opencl/opencl_sha256crypt.h
index 5f5495267..80d40df92 100644
--- a/run/opencl/opencl_sha256crypt.h
+++ b/run/opencl/opencl_sha256crypt.h
@@ -21,7 +21,7 @@
 #define PLAINTEXT_LENGTH        24
 #define BUFFER_ARRAY            8
 #define SALT_ARRAY              (SALT_LENGTH / 4)
-#define PLAINTEXT_ARRAY         (PLAINTEXT_LENGTH / 4)
+#define PLAINTEXT_ARRAY         (PLAINTEXT_LENGTH + 3 / 4)
 #define SEED                    1024
 #define STEP                    0
 #define LOOP_SIZE               (7*3*2)
```

solardiz:
> Yes, but I don't really care about 256crypt.

Same here, but I currently test using 256crypt first because it's faster. I'll test the same things against 512crypt next.

> I need to apply:

I think no, we simply need to revise the code not to assume there's extra space.

ghost:
> I think no, we simply need to revise the code not to assume there's extra space.

I'm not sure I never ignore the NULL ending.

solardiz:
> I'm not sure I never ignore the NULL ending.

Well, it was not guaranteed prior to our recent changes as well. So if you rely on NUL termination somewhere in the kernels, then we had a bug like that before.

BTW, why do you set the entire key buffer to NULs in `set_key()`? I think this should be unneeded, and can be optimized out. No?

solardiz:
> BTW, why do you set the entire key buffer to NULs in `set_key()`? I think this should be unneeded, and can be optimized out. No?

Tried changing the `\0` to 0x55 - no issues seen so far (same passwords are getting cracked in my test).

ghost:
> BTW, why do you set the entire key buffer to NULs in `set_key()`? I think this should be unneeded, and can be optimized out. No?

As I don't deal with characters (8 bits), I copy 32 or 64 bits at once when handling, for example, plaintexts. 
Removing the `set all key buffer to NULs` will require major refactoring, I guess.

BTW: It may be correct to skip the NULL endings, not sure.

solardiz:
> Tried changing the `\0` to 0x55 - no issues seen so far (same passwords are getting cracked in my test).

That was in a mixed-length wordlist run. With mask mode and just one length, this change resulted in nothing getting cracked. Interesting.

Anyway, with our copying up to actual length when grouping, whether with or without NUL, we're not copying the extra NUL padding that `set_key()` setup. So this is inconsistent. If the kernels need the padding, then we should have it both code paths. If it doesn't, then we shouldn't have it in either.

ghost:
> Anyway, with our copying up to actual length when grouping, whether with or without NUL, we're not copying the extra NUL padding that `set_key()` setup. So this is inconsistent. If the kernels need the padding, then we should have it both code paths. If it doesn't, then we shouldn't have it in either.

Do you mean: sha256crypt is bugged right now! Yes, it is. But only when plaintext length is 24, correct?

solardiz:
> Do you mean: sha256crypt is bugged right now! Yes, it is. But only when plaintext length is 24, correct?

In my testing, yes. However, per code review it looks worse than that.

solardiz:
> > Tried changing the `\0` to 0x55 - no issues seen so far (same passwords are getting cracked in my test).
> 
> That was in a mixed-length wordlist run. With mask mode and just one length, this change resulted in nothing getting cracked. Interesting.

Adding just one NUL terminating byte (when there's room for it) fixed that.

ghost:
Please, open issue(s) for that. I have only one patch for submission.

solardiz:
I am going to just go ahead and try and fix this stuff myself, and send a PR.

ghost:
All OpenCL CI tests are passing, including upstream Travis CI, CPU OpenCL AMD and CPU OpenCL Intel tests (my real [recorded] cracking tests need super):

https://github.com/claudioandre-br/JohnTheRipper/compare/3215c8d70c99...8a0d36306795

ghost:
What I see in gdb (revert this patch and get a segfault):

```
#0  0x00000000007eb778 in crypt_all (pcount=<optimized out>, _salt=<optimized out>) at opencl_sha256crypt_fmt_plug.c:605

(gdb) p index
$1 = <optimized out>
(gdb) p count
$2 = 49
(gdb) p gws
$3 = 49
(gdb) p indices[index]
value has been optimized out
(gdb) p indices[48]
$4 = 1818325760
```
This is 256crypt. My first thought was the loop in `line 593` maybe DO NOT clear all `indexes [n]` elements (**because count != GWS**).

I agree that it is really unexpected. But I can't see a bug in the code itself.
[edited]

Are you sure that count seen here means that set_key() has been called count times?

solardiz:
This is weird indeed. As I can't reproduce this so far, my current tentative patch adds a runtime check and a `fprintf(stderr, ...)` followed by a workaround, so that we'd know if this issue does occur.

ghost:
I updated the comment above.

Are you sure that `count` seen here means that `set_key()` has been called count times?

solardiz:
> Are you sure that `count` seen here means that `set_key()` has been called count times?

Good point! And no, during self-test this is in fact sometimes not the case!

magnumripper:
> > Are you sure that `count` seen here means that `set_key()` has been called count times?
> 
> Good point! And no, during self-test this is in fact sometimes not the case!

It isn't? I think in Jumbo we have added stuff just for this to be ensured.

PR #4701, opened 2021-04-27
Tacking

e2002e:
Hi, I got this module that runs well.
it does 3 things
1 bruteforce on two sides of the words, doing several loops.
2 sliding the index of the current treated character, wrapping around the word.
3 a simple rotate is added to "fill the gap" where characters from the start of the set still get duplicated in the words.

The entropy is not beaten over a large keyspace, lowercase can go to about length 12 on an i5.
Please just forget about my lasts repeated pull request and consider this for what it is.
It does not 'beat' incremental, but can be handy with small non human words, if we need to keep control over the frequency of the characters, we can get rid of option 3 but it'll need more work to beat the entropy.

I had great fun doing this in any way. I doesn't get mystical like the rain mode.
Sorry if this is full of old commits I can't squash them as I want.


solardiz:
I'm happy you're still around and having fun, but we're still uninterested. You have some of your own design goals with these modes, but you haven't made a convincing case that those goals are what JtR users ever need.

PR #4702, opened 2021-04-27
OpenCL: sha[256/512]crypt group by length further fix

solardiz:
See #4700

### DIFF
--- a/src/opencl_sha512crypt_fmt_plug.c
+++ b/src/opencl_sha512crypt_fmt_plug.c
@@ -305,15 +305,9 @@ static void clear_keys(void)
 
 static void set_key(char *key, int index)
 {
-	int len;
+	unsigned int len = strnlen(key, PLAINTEXT_LENGTH);
 
-	//Assure buffer has no "trash data".
-	memset(plaintext[index].pass, '\0', PLAINTEXT_LENGTH);
-	len = strlen(key);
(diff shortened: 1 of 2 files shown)


src/opencl_sha512crypt_fmt_plug.c
```diff
@@ -305,15 +305,9 @@ static void clear_keys(void)
 
 static void set_key(char *key, int index)
 {
-	int len;
-
-	//Assure buffer has no "trash data".
-	memset(plaintext[index].pass, '\0', PLAINTEXT_LENGTH);
-	len = strlen(key);
-	len = (len > PLAINTEXT_LENGTH ? PLAINTEXT_LENGTH : len);
```
ghost, line 313: This is not required in jumbo (@magnumripper?). 

I removed it from raw formats, because they are more relevant there. Ok, I know Solar intends to reverse this `len > PLAINTEXT_LENGTH` check that someday.

ghost (approved):
LGTM.

src/opencl_sha512crypt_fmt_plug.c
magnumripper: If you mean the `len > PLAINTEXT_LENGTH`, yes - Jumbo is supposed to never ever serve an overlong key. There has been bugs though, hopefully none are left. The original idea for it came from very fast CPU formats (eg. NT) where `set_key()` (and key generation) is the bottleneck.

Issue #4703, opened 2021-04-27
sha1crypt-opencl auto-tune stuck at LWS=256 on Kepler


solardiz:
I realized I didn't know to update `ITERATIONS` in `opencl_sha1crypt_fmt_plug.c` when I introduced hashcat's test vector, and was about to do that, but found another issue, also seen e.g. on a build from January:

On the Titan Kepler in "super", auto-tune for this format exhibits this behavior:

```
Calculating best LWS for GWS=16384
Testing LWS=32 GWS=16384 ... 244.709ms+
Testing LWS=64 GWS=16384 ... 196.442ms+
Testing LWS=128 GWS=16384 ... 196.738ms
Testing LWS=256 GWS=16384 ... 213.286ms
Testing LWS=256 GWS=16384 ... 213.683ms
Testing LWS=256 GWS=16384 ... 213.609ms
Testing LWS=256 GWS=16384 ... 212.117ms
Testing LWS=256 GWS=16384 ... 212.752ms
```

and so on - it doesn't complete. I wonder how we managed to test all formats including on that GPU before?

solardiz:
Testing my old builds, August 2018 works, December 2018 fails like above. However, this is really weird - surely we previously ran all formats on that GPU after that date?

solardiz:
@magnumripper @claudioandre-br I'd appreciate your help on this issue. This device is currently idle.

solardiz:
> I realized I didn't know to update `ITERATIONS` in `opencl_sha1crypt_fmt_plug.c` when I introduced hashcat's test vector, and was about to do that

This is now done, but as expected didn't make a difference for the stuck auto-tune issue reported here.

magnumripper:
Could it be we never tested anything but `-test=0`?

magnumripper:
Never mind how we didn't notice - what in the code allows/results in this behavior?

magnumripper:
```c
	for (optimal_work_group = my_work_group = wg_multiple;
	        (int)my_work_group <= (int)max_group_size;
	        my_work_group += wg_multiple) {
(...)
		if ((double)(sumRunTime) / kernelExecTimeNs < 0.997) {
			kernelExecTimeNs = sumRunTime;
			optimal_work_group = my_work_group;
		} else {
			if (my_work_group >= 256 || (my_work_group >= 8 && wg_multiple < 8)) {
				/* Jump to next power of 2 */
				size_t x, y;
				x = my_work_group;
				while ((y = x & (x - 1)))
[...]

PR #4706, opened 2021-04-28
Get next multiple

magnumripper:
Bugfix for LWS auto-tune getting stuck from vector width of 2, see #4703 

Also sha1crypt-opencl: Stop using x2 vectorizing for Kepler (separate commit) because it's slower now (at some point in time it was faster).

### DIFF
--- a/src/opencl_common.h
+++ b/src/opencl_common.h
@@ -314,9 +314,13 @@ void opencl_process_event(void);
 #define jtrReleaseKernel(kernel) ({ cl_kernel _k = (kernel); (kernel) = NULL; clReleaseKernel(_k); })
 #define jtrReleaseProgram(program) ({ cl_program _p = (program); (program) = NULL; clReleaseProgram(_p); })
 
-/* Macro for get a multiple of a given value */
+/* Macro for getting a multiple of a given value */
 #define GET_NEXT_MULTIPLE(dividend, divisor)	  \
-	(divisor) ? ((dividend + (ocl_v_width * divisor - 1)) / (ocl_v_width * divisor)) * divisor : (dividend) / ocl_v_width;
+	(divisor) ? (((dividend) + ((divisor) - 1)) / (divisor)) * (divisor) : (dividend);
+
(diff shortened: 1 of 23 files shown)


magnumripper:
Note: I only changed formats that potentially use vectorized GPU code. Perhaps we should change all uses of the form `gws = GET_NEXT_MULTIPLE(count, lws)` to use the new `GET_KPC_MULTIPLE(count, lws)`?

src/opencl_common.h
```diff
@@ -314,9 +314,13 @@ void opencl_process_event(void);
...
 
-/* Macro for get a multiple of a given value */
+/* Macro for getting a multiple of a given value */
 #define GET_NEXT_MULTIPLE(dividend, divisor)	  \
-	(divisor) ? ((dividend + (ocl_v_width * divisor - 1)) / (ocl_v_width * divisor)) * divisor : (dividend) / ocl_v_width;
+	(divisor) ? (((dividend) + ((divisor) - 1)) / (divisor)) * (divisor) : (dividend);
+
+/* Macro for translating 'count' (kpc) to GWS, taking vector width into account */
+#define GET_KPC_MULTIPLE(count, lws)	  \
+	(lws) ? (((count) + (ocl_v_width * (lws) - 1)) / (ocl_v_width * (lws))) * (lws) : ((count) + ocl_v_width - 1) / ocl_v_width;
```
solardiz, line 323: I see you're also introducing rounding-up in the "else" part here.

solardiz (approved):
OK with me, assuming this passes your tests. I didn't verify these are actually the exact set of formats where we use vectorization.

src/opencl_common.h
magnumripper: Yes, that was a bug for running with `AutotuneLWS = 0` which is supposed to keep LWS at NULL for the initial GWS tune instead of eg. querying `CL_KERNEL_PREFERRED_WORK_GROUP_SIZE_MULTIPLE`. This feature isn't even tested for a while, perhaps time to do so.

Issue #4705, opened 2021-04-28
oldOffice document with 56-bit key


magnumripper:
Original issue: https://www.openwall.com/lists/john-users/2021/04/27/1

I got a private copy of such document. Using some debug prints, I get
```
em1.doc : 
flags 4 providerType d provider "microsoft base dss and diffie-hellman cryptographic provider" algId 6801 algHashId 8004 keySize 56
```
According to [MS-OFFCRYPTO](https://docs.microsoft.com/en-us/openspecs/office_file_formats/ms-offcrypto/3c34d72a-1a61-4b52-a893-196f9157f083):
- "Flags 4" seem to mean "CryptoAPI RC4".
- providerType d (`0x0d`) doesn't seem normal. It should be `0x01` for RC4 or `0x18` for AES.
- provider: "microsoft base dss and diffie-hellman cryptographic provider"
- algId 6801 means RC4
- algHashId 8004 means SHA-1
- keySize can be any even 8 between 40 and 128 according to the doc, so 56 should be OK

The question is if we should simply truncate to 56 bits instead of 40, all else equal? It appears this might be the case but we can't know until we crack such document.

magnumripper:
> * providerType d (`0x0d`) doesn't seem normal. It should be `0x01` for RC4 or `0x18` for AES.
> * provider: "microsoft base dss and diffie-hellman cryptographic provider"

This doesn't make sense. https://docs.microsoft.com/en-us/windows/win32/seccrypto/prov-dss-dh-provider-algorithms says it provides MD5 and SHA hashing, but not RC4.

**EDIT** diving a bit deeper in the docs, RC4 is there as well: https://docs.microsoft.com/en-us/windows/win32/seccrypto/microsoft-enhanced-dss-and-diffie-hellman-cryptographic-provider

---

Our 40-bit sample `XP-RC4-40bit-base-1.0_myhovercraftisfullofeels_.doc` show this with same debug outputs:
```
XP-RC4-40bit-base-1.0_myhovercraftisfullofeels_.doc : 
flags 4 providerType 1 provider "microsoft base cryptographic provider v1.0" algId 6801 algHashId 8004 keySize 40
```
That providerType 1, and the provider name, makes more sense: https://docs.microsoft.com/en-us/windows/win32/seccrypto/base-provider-algorithms says it does provide RC4 as well.

magnumripper:
Like I said in original ML thread
[...]

PR #4707, opened 2021-04-29
Oldoffice: Support 56-bit RC4

magnumripper:
See #4705 

We won't merge this unless we see it crack such document. It's probably not trivial to create a sample document.

### DIFF
--- a/run/office2john.py
+++ b/run/office2john.py
@@ -2536,8 +2536,10 @@ def find_rc4_passinfo_doc(filename, stream):
             typ = 4
         elif keySize == 40:
             typ = 3
+        elif keySize == 56:
+            typ = 5
         else:
-            sys.stderr.write("%s : invalid keySize\n" % filename)
+            sys.stderr.write("%s : invalid keySize %u\n" % (filename, keySize))
 
         # Encryption verifier
         saltSize = unpack("<I", stream.read(4))[0]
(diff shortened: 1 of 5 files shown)


solardiz:
> We won't merge this unless we see it crack such document.

We could merge anyway, if we first add printing of a warning that it isn't known whether this key size is processed correctly.

magnumripper:
Added this
```c
   if (cs.type == 5 && !ldr_in_pot) {
       static int warned;

       if (john_main_process && !warned++) {
           fprintf(stderr, "Note: The support for OldOffice type 5 is experimental and may be incorrect.\n");
           fprintf(stderr, "      For latest news see https://github.com/openwall/john/issues/4705\n");
       }
   }
```

solardiz (approved):
Looks good to me. Why do we print the warnings from `get_salt()` rather than from `valid()`, though? And why isn't this in code shared between CPU and OpenCL formats? I suppose that's fine for this PR, and a cleanup can be made separately if we choose to.

frank-dittrich:
If we ever want to get rid of that warning, we might also request user feedback about successful cracks of such hashes and/or failure to crack such hashes with a valid password.

magnumripper:
> Looks good to me. Why do we print the warnings from `get_salt()` rather than from `valid()`, though?

Does it matter at all? I think I instinctively used `get_salt()` because it's only a problem loading these, not seeing them.

> And why isn't this in code shared between CPU and OpenCL formats? I suppose that's fine for this PR, and a cleanup can be made separately if we choose to.

As discussed not long ago, the CPU format is `FMT_BLOB` so can exploit same-salts, but I have yet to figure out how to do that to the GPU format due to GPU-side comparison.

magnumripper:
> If we ever want to get rid of that warning, we might also request user feedback about successful cracks of such hashes and/or failure to crack such hashes with a valid password.

I expect clever users to read that into the existing text.

PR #4708, opened 2021-04-29
Re-indent bt.c and related files

### LINKED ISSUE #4690: clang warning: (misleading indentation) in bt.c
```
bt.c:301:4: warning: misleading indentation; statement is not part of the previous 'for' [-Wmisleading-indentation]
                        fprintf(stdout, "Unused Slots in Offset Table:%Lf %%\n", 100.00 * (long double)(offset_table_size - i) / (long double)(offset_table_size));
                        ^
bt.c:299:3: note: previous statement is here
                for (i = 0; i < offset_table_size && offset_data[i].collisions; i++)
                ^
1 warning generated.
```
Indeed, there's a `;` before the `fprintf`

magnumripper:
Misleading indentation was seen so I re-indented using emacs and then picked what seemed most relevant.
Closes #4690

### DIFF
--- a/src/bt.c
+++ b/src/bt.c
@@ -112,13 +112,13 @@ void bt_free(void **ptr)
 
 void bt_error_fn(const char *str, char *file, int line)
 {
-      fprintf(stderr, "%s in file:%s, line:%d.\n", str, file, line);
-      error();
+	fprintf(stderr, "%s in file:%s, line:%d.\n", str, file, line);
+	error();
 }
 
 void bt_warn_fn(const char *str, char *file, int line)
 {
-      fprintf(stderr, "%s in file:%s, line:%d.\n", str, file, line);
+	fprintf(stderr, "%s in file:%s, line:%d.\n", str, file, line);
 }
 
 static unsigned int modulo_op(void * hash, unsigned int N, uint64_t shift64, uint64_t shift128)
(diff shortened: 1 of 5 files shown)


solardiz (approved):
Assuming that these are indeed indentation changes only and they make the indentation consistent, let's have them. I suggest that you also check that the .o files remain unchanged with these source code changes.

While I'm fine with these as-is, FWIW I generally don't indent continuation lines to opening brace on a previous line - rather, I use a fixed indentation for those (commonly 4 spaces, which is half of a full 8-char tab indentation, but I understand we might want to move away from that to some other fixed alignment that doesn't mix tabs and spaces).

magnumripper:
A `diff -w` only shows one single change: Dropping of a blank line in the end of one of the files.
```
[bt-misleading-indent]$ git diff -w bleeding-jumbo 
diff --git a/src/bt_twister.c b/src/bt_twister.c
index b954f1769..ce7581ae2 100644
--- a/src/bt_twister.c
+++ b/src/bt_twister.c
@@ -158,4 +158,3 @@ mt_uint32 randomMT(void)
 }
 
 #endif
-
```
So this is 100% safe

ghost:
> A diff -w only shows one single change: Dropping of a blank line in the end of one of the files.

Using GitHub GUI, you can also add a `?w=1` to the PR url.

magnumripper:
> While I'm fine with these as-is, FWIW I generally don't indent continuation lines to opening brace on a previous line - rather, I use a fixed indentation for those (commonly 4 spaces, which is half of a full 8-char tab indentation, but I understand we might want to move away from that to some other fixed alignment that doesn't mix tabs and spaces).

I wouldn't mind having my emacs settings do it your way, I think both ways are fine. I'm just not sure how to make it to :-)

PR #4709, opened 2021-04-30
Add --list=format-classes option

### LINKED ISSUE #4636: Add --list=format-groups options, for future bash completion
See https://github.com/openwall/john/issues/3514#issuecomment-800686198

magnumripper:
This is mainly for things like bash-completion.
Closes #4636

### DIFF
--- a/src/formats.c
+++ b/src/formats.c
@@ -63,6 +63,30 @@ static void test_fmt_case(struct fmt_main *format, void *binary,
 #endif
 
 /*
+ * List of format classes valid for this build (actually all are valid
+ * to ask for but we don't list those that can't match).
+ *
+ * This list is used in listconf.c and john.c
+ */
+char fmt_class_list[] = "all, enabled, disabled"
+#if _OPENMP
+	", omp"
+#endif
+#if HAVE_OPENCL || HAVE_ZTEX
+	", mask"
+#endif
+#if HAVE_ZTEX
+	", ztex"
+#endif
+#if HAVE_OPENCL
+	", opencl"
+#endif
+#ifndef DYNAMIC_DISABLED
+	", dynamic"
+#endif
+	", cpu";
+
+/*
  * Does req_format match format?
  *
  * req_format can be an exact full match or a wildc*rd, for format label,
--- a/src/formats.h
+++ b/src/formats.h
@@ -426,6 +426,11 @@ struct fmt_main {
  */
 extern char fmt_null_key[PLAINTEXT_BUFFER_SIZE];
 
+/*
+ * List of valid format classes for this build
+ */
+extern char fmt_class_list[];
+
 /* Self-test is running */
 extern int self_test_running;
 
--- a/src/listconf.c
+++ b/src/listconf.c
@@ -107,14 +107,12 @@ extern char CPU_req_name[];
  */
 static void listconf_list_options()
 {
-	puts("help[:WHAT], subformats, inc-modes, rules, externals, ext-modes");
-	puts("ext-hybrids, ext-filters, ext-filters-only, build-info, encodings");
-	puts("formats, format-details, format-all-details, format-methods[:WHICH],");
-	// With "opencl-devices, <conf section name>" added,
-	// the resulting line will get too long
-	puts("format-tests, sections, parameters:SECTION, list-data:SECTION,");
+	puts("help[:WHAT], subformats, inc-modes, rules, externals, ext-modes, ext-hybrids,");
+	puts("ext-filters, ext-filters-only, build-info, encodings, formats, format-classes,");
+	puts("format-details, format-all-details, format-methods[:WHICH], format-tests,");
+	printf("sections, parameters:SECTION, list-data:SECTION, ");
 #if HAVE_OPENCL
-	printf("opencl-devices, ");
+	puts("opencl-devices,");
 #endif
 	/* NOTE: The following must end the list. Anything listed after
 	   <conf section name> will be ignored by current
@@ -419,6 +417,11 @@ void listconf_parse_early(void)
 			error_msg("--format not allowed with \"--list=%s\"\n", options.listconf);
 
 		listEncodings(stdout);
+		exit(EXIT_SUCCESS);
+	}
+	if (!strcasecmp(options.listconf, "format-classes"))
+	{
+		puts(fmt_class_list);
 		exit(EXIT_SUCCESS);
 	}
 }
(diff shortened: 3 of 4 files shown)


frank-dittrich:
Looks useful to enhance bash completion for `--format=`.
On my system, all formats of class mask are also opencl formats. Apparently, (some of the) ztex formats also belong to class mask.
But format class mask is empty or invalid on CPU-only systems. So may be the output should be adjusted accordingly. 

The `--help` output lists these format classes:
```
                           Valid classes: dynamic, cpu, opencl, ztex, mask, omp,
                           all, enabled, disabled
```
Should the `--help` output also be changed to only include valid format classes?

And (unrelated to this patch) we might need a document mentioning the available format classes and their meaning (or a README.FORMATS which explains all the different options specifying a single format, a list of formats, format classes, usage of `*` as a wildcard, etc., and possibly adjustments to doc/OPTIONS:
```
  -test[=TIME]                   run tests and benchmarks for TIME seconds each

Tests all of the compiled in hashing algorithms for proper operation and
benchmarks them.  The "--format" option can be used to restrict this to
a specific algorithm.  Using --test=0 will do a very quick self-test but
no benchmark.
```

magnumripper:
> Looks useful to enhance bash completion for `--format=`.
> On my system, all formats of class mask are also opencl formats. Apparently, (some of the) ztex formats also belong to class mask.
> But format class mask is empty or invalid on CPU-only systems. So may be the output should be adjusted accordingly.

The `-format=mask` is for testing purposes so it's not very important. Also, we may have CPU formats with internal mask in the future.

> The `--help` output lists these format classes:
> 
> ```
>                            Valid classes: dynamic, cpu, opencl, ztex, mask, omp,
>                            all, enabled, disabled
> ```
> 
> Should the `--help` output also be changed to only include valid format classes?

Oh, I forgot about that list. I need to change this PR then - we should definitely produce the list in one same place.

> And (unrelated to this patch) we might need a document mentioning the available format classes and their meaning (or a README.FORMATS which explains all the different options specifying a single format, a list of formats, format classes, usage of `*` as a wildcard, etc., and possibly adjustments to doc/OPTIONS:
> 
> ```
>   -test[=TIME]                   run tests and benchmarks for TIME seconds each
> 
> Tests all of the compiled in hashing algorithms for proper operation and
> benchmarks them.  The "--format" option can be used to restrict this to
> a specific algorithm.  Using --test=0 will do a very quick self-test but
> no benchmark.
> ```

True, I'll address this as well. Thanks!

frank-dittrich:
Alternatively, you could provide a more detailed output, listting one class per line, plus a short explanation.
For bash completion, I can just split at the first space.
You could even enclose opencl, ztex, mask in parentheses if they are not available, so that bash completion can suppress these.

frank-dittrich:
> The `-format=mask` is for testing purposes so it's not very important. Also, we may have CPU formats with internal mask in the future.

But listing `mask` as valid and then getting `Unknown ciphertext format name requested` might confuse users.

magnumripper:
> > The `-format=mask` is for testing purposes so it's not very important. Also, we may have CPU formats with internal mask in the future.
> 
> But listing `mask` as valid and then getting `Unknown ciphertext format name requested` might confuse users.

Agreed. We should suppress opencl/ztex/mask (and even dynamic if built with `DYNAMIC_DISABLED`) from these lists where appropriate - but they should still be valid classes if used anyway.

frank-dittrich:
> Agreed. We should suppress opencl/ztex/mask (and even dynamic if built with `DYNAMIC_DISABLED`) from these lists where appropriate - but they should still be valid classes if used anyway.

Currently I get
```
$ ./john --test=0 --format=mask
Unknown ciphertext format name requested
$ ./john --test=0 --format=nt,mask
Error: No format matched 'mask'
$ ./john --test=0 --format=disabled
Unknown ciphertext format name requested
```

I wouldn't care about suppressing `disabled`, because it depends on the config file.
And I hope nobody ever complains about `enabled` not being suppressed after disabling all formats in the config.

magnumripper:
> Looks useful to enhance bash completion for `--format=`.
> On my system, all formats of class mask are also opencl formats. Apparently, (some of the) ztex formats also belong to class mask.
> But format class mask is empty or invalid on CPU-only systems. So may be the output should be adjusted accordingly.

I updated the code to only list "omp" if we have OpenMP support, only list "mask" if we have ztex or opencl, and only list "dynamic" unless `-DDYNAMIC_DISABLED`.

> ```
>                            Valid classes: dynamic, cpu, opencl, ztex, mask, omp,
>                            all, enabled, disabled
> ```
> 
> Should the `--help` output also be changed to only include valid format classes?

I changed this to instead refer to doc/OPTIONS which describes not only the classes but also the various types of wildcards and other things.

> And (unrelated to this patch) we might need a document mentioning the available format classes and their meaning (or a README.FORMATS which explains all the different options specifying a single format, a list of formats, format classes, usage of `*` as a wildcard, etc., and possibly adjustments to doc/OPTIONS:

We already have very thorough documentation in `doc/OPTIONS`, under the `--format=NAME[,NAME...]` heading.

frank-dittrich:
> We already have very thorough documentation in `doc/OPTIONS`, under the `--format=NAME[,NAME...]` heading.

I didn't find format classes in doc/OPTIONS, because you used the term `group aliases` instead of `format classes`, and I was too lazy to study all the existing docs.
Now I found the format classes, but noticed that you missed to mention format class `mask`.

And I think I found an error in this paragraph:
```
The order given of any "include" parts of a format list will be honored and
this can be used for merely changing format precedence while otherwise
keeping auto-detection.  This means eg. "--format=ztex,opencl,cpu" for a
crack session will end up in auto-detection among all formats, just like no
format option was used, except it will prefer ztex over opencl and only use
a CPU format unless a ZTEX or OpenCL format was available.
```
IMHO, the last line should read
```
a CPU format if neither a ZTEX nor an OpenCL format is available.
```

But this can be changed in a separate commit.

frank-dittrich (approved):
Looks good now. Please merge, and I'll enhance bash completion.

magnumripper:
I changed the wording for consistency. Will merge, thanks!

magnumripper:
> Now I found the format classes, but noticed that you missed to mention format class `mask`.

I can't find any argument for adding that to the documentation as it's only there for testing stuff after changing some shared internal mask code. So I opted to just leave it alone

magnumripper:
> I changed the wording for consistency. Will merge, thanks!

Turns out I pushed that to the wrong topic branch so I instead pushed it directly to main tree in 5585bf3ce

PR #4722, opened 2021-05-10
CI: start using GitHub Container Registry

### LINKED ISSUE #4711: CircleCI (build) does not work with Fedora 34
I expect the EOL of the current image (f32) to be at the end of this year. We need to think about an alternative. I'm thinking about moving upstream CircleCI to Ubuntu.

It fails like this:
```bash
./configure: This script requires a shell more modern than all
./configure: the shells that I found on your system.
./configure: Please tell bug-autoconf@gnu.org and john-users mailing
./configure: list at openwall.com about your system, including any
./configure: error possibly output before this message. Then install
./configure: a modern shell, or manually run the script under such a
./configure: shell if you do have one.
```
Env:
```bash
System information:
 Server Version: 19.03.13
 Storage Driver: overlay2
  Backing Filesystem: xfs
 Cgroup Driver: cgroupfs
 Kernel Version: 4.15.0-1092-aws
 Operating System: Ubuntu 18.04.5 LTS
 OSType: linux
 Architecture: x86_64
[...]
Starting container claudioandre/john:fedora.34
[...]
fedora.34: Pulling from claudioandre/john

eb97e9b3: Pulling fs layer 
Digest: sha256:f4c8ea1f4c54536069fdac96c0efa846bd0562ecd03f849ad603e1600ce1d07f02GB/1.602GBB

```
If fails here: `{ $as_echo "$as_bourne_compatible""$as_required" | as_run=a "$SHELL"; }` (line 234) 

There is a nasty and known bug in `seccomp`/`glibc` that makes a simple `test -x` fail inside a container. 
* Testing locally I can't reproduce (so it's not this bug).
* I bet Fedora tried to fix the bug and did it partially (because I can build a Docker image using F34 but still can't do the same using Ubuntu 21.04, where the bug is still open).

ghost:
To prevent the image from becoming an INACTIVE IMAGE in Docker Hub that can even
be automatically deleted.

Closes #4711.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   common_linux: &common_linux
     docker:
-      - image: claudioandre/john:fedora.32
+      - image: ghcr.io/openwall/john-ci:fedora.32
     steps:
       - checkout
       - run:
@@ -15,7 +15,7 @@ jobs:
   wine-32bits:
     <<: *common_linux
     docker:
-      - image: claudioandre/john:fedora.30
+      - image: ghcr.io/openwall/john-ci:fedora.30
         environment:
           TARGET_ARCH: "i686"
           WINE: "wine"
@@ -24,7 +24,7 @@ jobs:
   wine-64bits:
     <<: *common_linux
     docker:
-      - image: claudioandre/john:fedora.32
+      - image: ghcr.io/openwall/john-ci:fedora.32
         environment:
           TARGET_ARCH: "x86_64"
           WINE: "wine"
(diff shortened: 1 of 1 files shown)


ghost:
Bots are happy (the build of my branch is complete). I consider this the appropriate solution for the needs of upstream CI.

I will continue to try F34 in Circle; if no workaround is possible, I will change my personal use of the CircleCI.

solardiz (approved):
Thank you, Claudio!

How/where can we update/add images in GitHub Container Registry, in case we need that later?

ghost:
It is a regular Docker registry. So, you only need to run a `docker push`. 

You can create the credentials at https://github.com/settings/tokens

Issue #4430, opened 2020-11-01
Simplify/speedup "--show" and "--make-charset"


solardiz:
As I pointed out in https://github.com/openwall/john/issues/4388#issuecomment-719723533, jumbo's usage of `split()` on `--show` is much heavier than core's, which is both good (allowing for #4428) and bad (slow). #4429 is a start at optimizing it, however we can do more.

Perhaps most importantly, we can optimize the scanning of the formats list in `ldr_show_pot_line()` to cache and probe the previous line's format's `valid()` first.

Also, I think right after merging #4429 (or adding to it) we can also make `ldr_cracked_hash()` case-sensitive.

Further, why do we(?) even bother to build the `db->cracked_hash` table when invoked with `--make-charset`? We should probably skip that.

Finally, given that #4429 obsoletes `FMT_SPLIT_UNIFIES_CASE`, should we drop that flag from everywhere, keep it around in case we need it again later, or rename/repurpose it (e.g., to just `FMT_SPLIT_UNIFIES`, so that we'd set it in more formats yet be able to skip `split()` in `ldr_show_pot_line()` for most formats)?

solardiz:
Copying another observation of mine from #4388 to here:

>  Another detail I notice is that jumbo would skip pot lines which no format recognizes. In core, basic string matching works even for unrecognized formats, except when running with explicit "--format". Perhaps this difference is unintentional (looks like it got introduced in 0ceedb2 without being documented), and is easy to fix.

which magnum confirmed there:

> Yes, definitely unintentional.

So I think should track it here, and fix it, and test the fix.

solardiz:
> rename/repurpose it (e.g., to just `FMT_SPLIT_UNIFIES`, so that we'd set it in more formats yet be able to skip `split()` in `ldr_show_pot_line()` for most formats)

I now realize that we'd first incur the overhead of `valid()` anyway, and then it's just a matter of checking the flag vs. calling `fmt_default_split()` in the typical case, so we won't achieve much speedup by having the flag except for pot lines with formats that have non-trivial `split()` yet don't set the flag (because their non-trivial `split()` actually splits things, and does not unify). Do we even have any formats like this? We have actual splitting in descrypt and LM, but in both of those we also unify, so we'd set the flag, so they are not examples where the flag would be beneficial. Any others? If not, or if this is indeed very rare, then maybe the flag isn't worth having.

magnumripper:
[...]

PR #4726, opened 2021-05-11
Bugfix for --make-charset with --format

magnumripper:
If `--format` option is used, only such pot entries will be used. This includes format groups, lists and wildcards so we can now create a charset from just unix passwords with eg. `--make-charset=FILE --format=*crypt`.

Like before, `--make-charset` without any specified format will load any pot lines including fake ones made from a wordlist with eg. `sed s/^/:/`.

As a separate commit, a KISS "cache" for the format enumeration. It sped up my test case (using `--show` with 10M mixed entries in pot file, and not specifying format) from well over a minute to just above 5 seconds. I also tried more elaborate tricks but they added very little additional performance so was ditched for the higher complexity.

### DIFF
--- a/src/loader.c
+++ b/src/loader.c
@@ -2165,10 +2165,36 @@ static void ldr_show_pot_line(struct db_main *db, char *line)
 	char *ciphertext, *pos;
 	int hash;
 	struct db_cracked *current, *last;
+	static struct fmt_main *last_fmt;
 
 	ciphertext = ldr_get_field(&line, db->options->field_sep_char);
 
 	if (line) {
+/*
+ * Jumbo-specific; split() needed for legacy pot entries so we need to
+ * enumerate formats to know which to call.
+ * This also takes care of the situation where specific format(s) was
+ * requested for --make-charset.
+ */
+		struct fmt_main *format = fmt_list;
+
+		if (last_fmt && (ldr_trunc_valid(ciphertext, last_fmt) == 1))
+			format = last_fmt;
+		else
+		if (!(db->options->flags & DB_PLAINTEXTS) || (options.flags & FLG_FORMAT)) {
+			do {
+				if (format != last_fmt && ldr_trunc_valid(ciphertext, format) == 1)
+					break;
+			} while ((format = format->next));
+
+			if (format)
+				last_fmt = format;
+		}
+
+/* If format(s) was forced on the command line, insist on it (them) */
+		if (!format && (options.flags & FLG_FORMAT))
+			return;
+
 		pos = line;
 		do {
 			if (*pos == '\r' || *pos == '\n') *pos = 0;
@@ -2178,25 +2204,10 @@ static void ldr_show_pot_line(struct db_main *db, char *line)
 			list_add(db->plaintexts, line);
 			return;
 		}
-/*
- * Jumbo-specific; split() needed for legacy pot entries so we need to
- * enumerate formats and call valid(). This also takes care of the situation
- * where a specific format was requested.
- */
-		if (!(options.flags & FLG_MAKECHR_CHK)) {
-			struct fmt_main *format = fmt_list;
 
-			do {
-				if (ldr_trunc_valid(ciphertext, format) == 1)
-					break;
-			} while((format = format->next));
+		if (format)
+			ciphertext = format->methods.split(ciphertext, 0, format);
 
-			if (!format)
-				return;
-
-			ciphertext =
-				format->methods.split(ciphertext, 0, format);
-		}
 		hash = ldr_cracked_hash(ciphertext);
 
 		last = db->cracked_hash[hash];


src/loader.c
```diff
@@ -2165,27 +2165,23 @@ static void ldr_show_pot_line(struct db_main *db, char *line)
...
-		if (!(options.flags & FLG_MAKECHR_CHK)) {
-			struct fmt_main *format = fmt_list;
+		struct fmt_main *format = fmt_list;
 
+		if (last_fmt && (ldr_trunc_valid(ciphertext, last_fmt) == 1))
+			format = last_fmt;
+		else
+		if (!(db->options->flags & DB_PLAINTEXTS) || (options.flags & FLG_FORMAT)) {
 			do {
 				if (ldr_trunc_valid(ciphertext, format) == 1)
```
solardiz: Maybe add `format != last_fmt &&` here, since at this point we already know it's not that format.

solardiz (changes requested):
I think as written this breaks useful functionality of being able to process lines with hashes not recognized as any format by the current build. What about, for example, building .chr files from a wordlist processed through `sed s/^/:/` to produce a fake pot file? We should keep this working.

magnumripper:
> I think as written this breaks useful functionality of being able to process lines with hashes not recognized as any format by the current build. What about, for example, building .chr files from a wordlist processed through `sed s/^/:/` to produce a fake pot file? We should keep this working.

As I specifically mentioned almost literally, this still works (as long, obviously, as you do not give the `--format` option). This was not only dry-run but actually tested.

src/loader.c
magnumripper: Do you mean at [new] line 2186?

solardiz: Yes, I mean new line 2186.

solardiz (approved):

solardiz:
> > `sed s/^/:/` to produce a fake pot file? We should keep this working.
> 
> As I specifically mentioned almost literally, this still works (as long, obviously, as you do not give the `--format` option).

Sorry, missed that mention and misread the code.

I still think that `--show` of hashes not supported by the current build is broken with these changes, though. No? I think this used to work at least in core (only when strings directly matched without needing any `split()` magic, of course). IIRC, this was a reason why I had that case-insensitive hash function in core. But perhaps we already broke this in jumbo before this PR? I now also recall recently concluding and commenting in here that we could drop the case-insensitive hash function in jumbo. Need to find those GitHub issues now - maybe your optimization closes one of them.

solardiz:
> Need to find those GitHub issues now - maybe your optimization closes one of them.

See #4430. Your optimization here is one of the things mentioned there.

magnumripper:
> I still think that `--show` of hashes not supported by the current build is broken with these changes, though. No? I think this used to work at least in core (only when strings directly matched without needing any `split()` magic, of course).

With *these* changes? How? Perhaps the stuff you mention is and was borken - but I can't see this very commit making anything worse in that regard - everything should be same or better. I will take another look at this before merging though - there's really no hurry to merge. I'll take a nap and have a look with fresh eyes.

One thing I did try, apart for `sed s/^/:/`, was doing a `--show` without `--format` from an input file containing several formats, and it does show several formats' cracks in one go. I've had a feeling that was borken in Jumbo fo a while (possibly years) but I'm not quite sure it actually was and I'm not that eager to prove things - I'm just happy it does work now.

magnumripper:
> I think this used to work at least in core (only when strings directly matched (...)

BTW and for the record, in preparation for this fix I isolated John proper's `ldr_show_pot_line` (which is like a 12-line function) and tried to make sure I understood it. There is no `strcmp` or `strcasecmp` in there that was lost so I can't really see what would fail. I'm not saying there aren't such things elsewhere but to the best of my knowledge, this function isn't it. But again, I will have another look once I've had a good night's sleep.

src/loader.c
```diff
@@ -2194,9 +2190,21 @@ static void ldr_show_pot_line(struct db_main *db, char *line)
...
+			if (*pos == '\r' || *pos == '\n') *pos = 0;
+		} while (*pos++);
+
+		if (db->options->flags & DB_PLAINTEXTS) {
+			list_add(db->plaintexts, line);
+			return;
 		}
+
+		ciphertext = format->methods.split(ciphertext, 0, format);
+
```
magnumripper: Now that I stare at this again... shouldn't we put lines 2196-2199 inside the 2201 conditional?

magnumripper: No maybe not... at 2217 [new] we use this

magnumripper:
> > I still think that `--show` of hashes not supported by the current build is broken with these changes, though. No? I think this used to work at least in core (only when strings directly matched without needing any `split()` magic, of course).
> 
> With _these_ changes? How? Perhaps the stuff you mention is and was borken - but I can't see this very commit making anything worse in that regard

I see the problem now - it wasn't caused by this commit but by the original one where we started calling `split`. I think I have a fix: If we don't find a format we shouldn't give up and return, but continue without calling any `split`. But only if no `--format` option was used.

src/loader.c
solardiz: We could probably speed things up by having a variation of `str_alloc_copy()` that would allocate/copy until CR/LF/NUL. That way, the string would only be scanned for a terminating character once (inside that function), not twice (before and inside). We'd then also need a variation of `list_add()` that would use this function. I'm not sure this is worth the complexity.

magnumripper:
New version force-pushed, fixing the "show of hashes not supported by the current build". The "caching" is no longer a separate commit.

src/loader.c
```diff
@@ -2165,10 +2165,36 @@ static void ldr_show_pot_line(struct db_main *db, char *line)
...
+		struct fmt_main *format = fmt_list;
+
+		if (last_fmt && (ldr_trunc_valid(ciphertext, last_fmt) == 1))
+			format = last_fmt;
+		else
+		if (!(db->options->flags & DB_PLAINTEXTS) || (options.flags & FLG_FORMAT)) {
+			do {
+				if (format != last_fmt && ldr_trunc_valid(ciphertext, format) == 1)
+					break;
+			} while((format = format->next));
```
solardiz: Can as well introduce the missing whitespace before `((format = format->next));` while we're at it.

solardiz (approved):
Looks good to me.

magnumripper:
I've retested `--make-charset` with or without format(s), and `--show` with or without format(s), including the "unsupported hash" case. It seems to be OK now.

solardiz:
> `--show` of hashes not supported by the current build [...] used to work at least in core (only when strings directly matched without needing any `split()` magic, of course). IIRC, this was a reason why I had that case-insensitive hash function in core.

Thinking of this some more, I don't see how case-insensitive hashing could help there. It only helps arrive at a right comparison, but if the comparison itself is case-sensitive then there would be no match anyway (and for unrecognized formats the comparison should be case-sensitive, or else we'd have false positive matches for unrecognized formats where case in hashes matters). So it only helped along with `split()`, and perhaps only along with `FMT_SPLIT_UNIFIES_CASE`, and we're now obsoleting this in jumbo. So we could proceed (separately, not in this PR/commit) to drop the case-insensitivity of hashing as mentioned in #4430.

PR #4729, opened 2021-05-12
relbench enhancement to avoid reporting wrong speeds

frank-dittrich:
A sequence of
$ ./john --test=-1 --format=$f >> file
where some formats cause segfaults might result in lines
Benchmarking: format1 [algorithm 1]... Benchmarking: format2 [algorithm 2]... DONE
This change picks the Last "Benchmarking: " output per line.

### DIFF
--- a/run/relbench
+++ b/run/relbench
@@ -77,7 +77,7 @@ sub parse
 			return;
 		}
 	} else {
-		($name) = /^\t?Benchmarking: ([^\[]+) \[.*\].* (PASS, |SKIP, |.....\b\b\b\b\b)?(DONE|Warning:.*)$/;
+		($name) = /^.*Benchmarking: ([^\[]+) \[.*\].* (PASS, |SKIP, |.....\b\b\b\b\b)?(DONE|Warning:.*)$/;
 		$ok = defined($name);
 	}
 	print STDERR "Could not parse: $_\n" if (!$ok);


solardiz (approved):
Not the intended kind of input to this script, but I don't mind supporting that as well.

Why not just drop the "^.*", though? Also, is this (written either way) guaranteed to capture the last occurrence of "Benchmarking:" on a line, or can its behavior vary? Maybe a more elaborate change is needed to handle the described kind of input and use case reliably?

frank-dittrich:
> Not the intended kind of input to this script, but I don't mind supporting that as well.

I wanted to compare POCL and NEO opencl formats, but some of the formats not just failed, but caused segfaults.
(I could have added extra empty lines using `echo ""` inside the for loop, but I didn't want to rerun the tests.)

> Why not just drop the "^.*", though? Also, is this (written either way) guaranteed to capture the last occurrence of "Benchmarking:" on a line, or can its behavior vary? Maybe a more elaborate change is needed to handle the described kind of input and use case reliably?

I can drop the `^`. It was present prior to my changes. But `.*` is required, so that perl's greedy pattern matching kicks in.
Otherwise, this line in one of the files
```
Benchmarking: KeePass-opencl [SHA256 AES/Twofish/ChaCha OpenCL]... Benchmarking: oldoffice-opencl, MS Office <= 2003 [MD5/SHA1 RC4 OpenCL/mask accel]... DONE
```
will cause relbench to compare oldoffice-opencl speed of one device with KeePass-opencl speed of the other and report an absurd ratio for KeePass opencl.

If there wasn't such a huge speed difference, I might not have noticed there's a problem.

solardiz:
> But `.*` is required, so that perl's greedy pattern matching kicks in.

Ah, this sort of answers my question about reliability. I'm OK merging this as-is. So please feel free to "rebase and merge".

PR #4730, opened 2021-05-12
Assorted john.conf rules and comments updates

solardiz:
Three commits here: best64.rule update to use some JtR's features to produce fewer duplicates, newly added Phrase* rulesets, and update of comments (that were out of date) on ZTEX sha512crypt and Drupal7 configuration settings.

### DIFF
--- a/run/john.conf
+++ b/run/john.conf
@@ -527,12 +527,12 @@ Frequency = 160
 #Frequency = 160
 # Some bitstreams accept runtime configuration.
 # In sha512crypt/Drupal7, configuration is 2 bytes. That's interpreted
-# as a bitmask. By setting any of the lowest 10 bits to 1 it turns off
-# corresponding unit (there are 10 units in the bitstream).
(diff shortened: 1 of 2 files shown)


solardiz:
Before:

```
$ ./john -w=rtop-3p -ru=best64 -stdo | sort -S20G -u | md5sum
80074937p 0:00:00:09 100.00% (2021-05-12 20:05) 8446Kp/s
b7986a8827a6f3aae2efe86083400c82  -
```

After:

```
$ ./john -w=rtop-3p -ru=best64 -stdo | sort -S20G -u | md5sum
77020392p 0:00:00:09 100.00% (2021-05-12 20:05) 8124Kp/s       
b7986a8827a6f3aae2efe86083400c82  -
```

(Curiously, with `-S10G` instead of `-S20G` in both cases, the "before" test ran for 21 seconds, whereas the "after" one for only 9 seconds like the above. Perhaps we had crossed that size threshold with the duplicates in the "before" case.)

PR #4735, opened 2021-05-26
RAR formats: Optionally build without non-free unrar code

### LINKED ISSUE #4731: Support building and repackaging without unRAR
hashcat 6.2.1 was just released making their use of unrar optional. Perhaps we should make ours optional, too. This is because of the unrar license, which some distributions/packagers find unsuitable: https://github.com/hashcat/hashcat/issues/2614

magnumripper:
If unrar sources are missing or `--without-unrar` is used with `configure`, build RAR v3 formats without the unrar code.  We can then only attack stored files or archives with header protection.

Closes #4731

### DIFF
--- a/doc/NEWS
+++ b/doc/NEWS
@@ -187,7 +187,12 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
   [magnum; 2021]
 
 - Improved Monero support: Now supports legacy wallets from before 2016 that are
-  not using JSON format yet. [patrickd, magnumripper, solardiz; 2021]
+  not using JSON format yet.  [patrickd; 2021]
+
+- Add --without-unrar option to configure, for building without the non-free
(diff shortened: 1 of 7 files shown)


ghost:
Magnum, do you have a solution for?:
> checking for unrar.c... configure: error: cannot check for file existence when cross compiling

Or should I (we) disable `rar` when building on Wine?

magnumripper:
Oh, good catch. I'll rewrite it without using the AC_CHECK_FILES macro.

magnumripper:
New version force-pushed. This one should work when cross compiling as well.

magnumripper:
Added info in doc/NEWS as well as doc/README-DISTROS now.

solardiz (approved):
I didn't review this closely enough to know if it's entirely correct and complete, but it looks good to me. Thank you, magnum!

While at it, I notice that we don't consistently capitalize configure option descriptions. Perhaps we should, like we did for `john --help` recently. Can be cleaned up separately, doesn't need to be in this PR.

PR #4737, opened 2021-05-26
Encourage the user to think when opening an issue

ghost:
I would say that emojis invite the user to read and, perhaps, pay attention. This is just a random idea that I think is worth mentioning.

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,12 +1,14 @@
+### Checklist ###
+- [ ] :1st_place_medal: I've read and understood these instrucions;
+  - This is not a support forum, it's a bug tracker. For questions and support, review postings on the [john-users mailing list](https://www.openwall.com/lists/john-users/).
+- [ ] :thumbsup: I've tested using latest bleeding version from this repository.
+  - Be clear about your environment and what you are doing. Share a sample hash or file that can be used to reproduce.
+- [ ] :confused: I'm confused and I need guidance.
+  - Please, read the instructions at https://www.openwall.com/john/#lists, then join the list before posting.
+
 ### IMPORTANT ###
-This is not a support forum, it's a bug tracker. If you don't understand the difference, please do NOT open an issue. For questions and support, review recent postings on the john-users mailing list at https://www.openwall.com/lists/john-users/ and then subscribe to the list at https://www.openwall.com/john/#lists and post your message to the list by e-mailing `john-users at lists.openwall.com`.
+We expect only reports of issues with the latest revision found in this GitHub repository. We do not expect in here, and have little use for, reports of issues only seen in a release or in a distro package.
 
-We are not interested in bugs in your distro-supplied and/or several years old Jumbo. Please do not open an issue unless you tried the latest version from HERE first.
-
-### Steps to reproduce ###
-Try to be clear about your environment and what you are doing. If possible, share a sample hash or file that can be used to reproduce.
-
-### System configuration ###
-Attach details about your OS and about JtR, including:
-- `$ ./john --list=build-info`.
-- `$ ./john --list=opencl-devices` (if applicable).
+Attach details about your OS and about john, including:
+- The output of `./john --list=build-info`.
+- The command line you are using.


magnumripper (commented):
We should also shorten the instructions as much as possible so it's not a wall of text to read. Perhaps those three (or few) check lines should be reworded so they *are* the instructions and there'd be nothing below them.

.github/issue_template.md
```diff
@@ -1,12 +1,16 @@
-### IMPORTANT ###
-This is not a support forum, it's a bug tracker. If you don't understand the difference, please do NOT open an issue. For questions and support, review recent postings on the john-users mailing list at https://www.openwall.com/lists/john-users/ and then subscribe to the list at https://www.openwall.com/john/#lists and post your message to the list by e-mailing `john-users at lists.openwall.com`.
+### Checklist ###
+- [ ] :1st_place_medal: I've read and understood these instrucions;
+  - This is not a support forum, it's a bug tracker. For questions and support, review postings on the [john-users mailing](https://www.openwall.com/lists/john-users/).
```
magnumripper: `s/mailing/mailing list/`, then I'm happy with this. Let's see what @solardiz thinks as well.

ghost: I removed a couple of duplicate lines. We can put them back if needed.

.github/issue_template.md
```diff
@@ -1,12 +1,14 @@
...
-This is not a support forum, it's a bug tracker. If you don't understand the difference, please do NOT open an issue. For questions and support, review recent postings on the john-users mailing list at https://www.openwall.com/lists/john-users/ and then subscribe to the list at https://www.openwall.com/john/#lists and post your message to the list by e-mailing `john-users at lists.openwall.com`.
-
-We are not interested in bugs in your distro-supplied and/or several years old Jumbo. Please do not open an issue unless you tried the latest version from HERE first.
+### Checklist ###
+- [ ] :1st_place_medal: I've read and understood these instrucions;
+  - This is not a support forum, it's a bug tracker. For questions and support, review postings on the [john-users mailing list](https://www.openwall.com/lists/john-users/).
+- [ ] :thumbsup: I've tested using latest bleeding version from this repository.
+  - Be clear about your environment and what you are doing. Share a sample hash or file that can be used to reproduce.
+- [ ] :confused: I'm confused and I need guidance.
+  - Please, post your message to john-users list by e-mailing `john-users at lists.openwall.com`.
```
solardiz: I think we need people to read the instructions at https://www.openwall.com/john/#lists and to join the list before posting. Asking anyone confused to just go ahead and post will increase noise on the list, and will also result in people posting to the public list without realizing they're making the information public, then complain and ask to remove stuff from list archives.

solardiz:
I think we have reverse [survivorship bias](https://en.wikipedia.org/wiki/Survivorship_bias) here. Although we see it in reverse - we only see issues wrongly opened when things went wrong - it's a similar kind of bias nevertheless, and can similarly result in us making wrong adjustments. Maybe the existing issue template actually works well most of the time - maybe plenty of potential issues were never opened due to people reading the issue template and realizing they need to try latest code from this repo first and having the problem gone.

> We should also shorten the instructions as much as possible so it's not a wall of text to read.

I agree with this, but I'm not sure how much there is to shorten. Rather, I think many people are confused about our releases vs. our latest GitHub code vs. distro packages. We refer to these things in there, but we don't explain them. If someone genuinely doesn't know the difference, then making the instructions brief won't help. We could instead want to add a (hopefully brief) explanation of what things exist.

For example, where we say "not interested in your several years old Jumbo", it can sound arrogant and be non-helpful. We could instead say something like:

"_On GitHub_, we expect only reports of issues with the latest revision found in this GitHub repository. We do not expect _in here_, and have little use for, reports of issues only seen in a numbered release or in a distro package."

(The exact wording may change to make it fit with the rest, including with instructions on where to obtain user support.)

solardiz:
I also think we're getting sufficiently few inappropriately opened issues, and they do not really pose a problem for us.

.github/issue_template.md
```diff
@@ -1,12 +1,14 @@
...
-### Steps to reproduce ###
-Try to be clear about your environment and what you are doing. If possible, share a sample hash or file that can be used to reproduce.
+### IMPORTANT ###
+We are not interested in bugs in your distro-supplied and/or several years old Jumbo.
 
-### System configuration ###
 Attach details about your OS and about JtR, including:
-- `$ ./john --list=build-info`.
-- `$ ./john --list=opencl-devices` (if applicable).
+- The output of `$ ./john --list=build-info`.
```
solardiz: Should omit the shell prompt here.

solardiz (approved):
I'm OK with giving this a try.

solardiz:
I now think we should drop those task lists from the issue template - or turn them into something other than task lists from GitHub UI's point of view. With the way things are now, we end up having GitHub track "2 of 3 tasks" per each issue reported to us. That's just weird, and an abuse of the task lists feature by our project.

PR #4741, opened 2021-06-02
Review aliasing violations in OpenCL code

magnumripper:
This is work in progress. So far, the byte wise OpenCL memcpy stuff seem to generally *bump* performance (if anything) using a fairly recent nvidia driver (CUDA 11.3 a.k.a 465.19.01) so I see no reason to re-add larger writes through pointers to unions. I assume recent drivers' optimizer recognizes the KISS memcpy and does the right thing anyway (and quite possibly better than the older offending code).

The nvidia-specific 32-byte writes in `opencl_*_ctx.h` are kept, but clearly marked as aliasing violations and now easily switchable from `opencl_misc.h`.

~As of right now, `mgetl.h` too is just byte wise copy (except for SIMD) but I will try re-adding the wider writes using pointers to unions.~

### DIFF
--- a/run/opencl/axcrypt2_kernel.cl
+++ b/run/opencl/axcrypt2_kernel.cl
@@ -15,16 +15,16 @@
 #include "opencl_aes.h"
 
 #define PUT_64BITS_XOR_LSB(cp, value) ( \
-		(cp)[4] ^= (unsigned char)((value) >> 24), \
-		(cp)[5] ^= (unsigned char)((value) >> 16), \
-		(cp)[6] ^= (unsigned char)((value) >> 8),  \
-		(cp)[7] ^= (unsigned char)((value)) )
+		(cp)[4] ^= (uchar)((value) >> 24), \
+		(cp)[5] ^= (uchar)((value) >> 16), \
+		(cp)[6] ^= (uchar)((value) >> 8),  \
+		(cp)[7] ^= (uchar)((value)) )
 
 typedef struct {
 	salt_t pbkdf2;
 	uint key_wrapping_rounds;
 	uchar salt[64];
-	uchar wrappedkey[144];
+	uint wrappedkey[144/4];
 } axcrypt2_salt_t;
 
 typedef struct {
@@ -32,8 +32,8 @@ typedef struct {
 } out_t;
 
 __kernel void axcrypt2_final(__global crack_t *pbkdf2,
-                         __constant axcrypt2_salt_t *salt,
-                         __global out_t *out)
+                             __constant axcrypt2_salt_t *salt,
+                             __global out_t *out)
 {
 	uint gid = get_global_id(0);
 
@@ -48,12 +48,16 @@ __kernel void axcrypt2_final(__global crack_t *pbkdf2,
 	for (i = 0; i < 8; i++)
 		key.u[i] = SWAP64(pbkdf2[gid].hash[i]);
 
-	unsigned char KEK[32];
+	uchar KEK[32];
 	AES_KEY akey;
 	int halfblocklen = 16 / 2;
 	int wrappedkeylen = 56 - halfblocklen;
-	unsigned char wrapped[144];
-	unsigned char block[16];
+	union {
+		uchar c[144];
+		uint w[144/4];
+		ulong l[144/8];
+	} wrapped;
+	uchar block[16];
 	int t;
 
 	memset_p(KEK, 0, 32);
@@ -66,25 +70,25 @@ __kernel void axcrypt2_final(__global crack_t *pbkdf2,
 
 	AES_set_decrypt_key(KEK, 256, &akey);
 
-	memcpy_macro(wrapped, salt->wrappedkey, 56);
+	memcpy_macro(wrapped.w, salt->wrappedkey, 56/4);
 
 	/* custom AES un-wrapping loop */
 	for (j = nb_iterations - 1; j >= 0; j--) {
 		for (k = wrappedkeylen / halfblocklen; k >= 1; --k) {
 			t = ((wrappedkeylen / halfblocklen) * j) + k;
 			// MSB(B) = A XOR t
-			memcpy_pp(block, wrapped, halfblocklen);
+			memcpy_pp(block, wrapped.c, halfblocklen);
 			PUT_64BITS_XOR_LSB(block, t);
 			// LSB(B) = R[i]
-			memcpy_pp(block + halfblocklen, wrapped + k * halfblocklen, halfblocklen);
+			memcpy_pp(block + halfblocklen, wrapped.c + k * halfblocklen, halfblocklen);
 			// B = AESD(K, X xor t | R[i]) where t = (n * j) + i
 			AES_decrypt(block, block, &akey);
 			// A = MSB(B)
-			memcpy_pp(wrapped, block, halfblocklen);
+			memcpy_pp(wrapped.c, block, halfblocklen);
 			// R[i] = LSB(B)
-			memcpy_pp(wrapped + k * halfblocklen, block + halfblocklen, halfblocklen);
+			memcpy_pp(wrapped.c + k * halfblocklen, block + halfblocklen, halfblocklen);
 		}
 	}
 
-	out[gid].cracked = !memcmp_pc(wrapped, "\xA6\xA6\xA6\xA6\xA6\xA6\xA6\xA6", 8);
+	out[gid].cracked = (wrapped.l[0] == 0xa6a6a6a6a6a6a6a6ULL);
 }
--- a/run/opencl/axcrypt_kernel.cl
+++ b/run/opencl/axcrypt_kernel.cl
@@ -38,10 +38,10 @@ typedef struct {
 } axcrypt_out;
 
 #define PUT_64BITS_XOR_MSB(cp, value) ( \
-                (cp)[0] ^= (unsigned char)((value)), \
-                (cp)[1] ^= (unsigned char)((value) >> 8), \
-                (cp)[2] ^= (unsigned char)((value) >> 16), \
-                (cp)[3] ^= (unsigned char)((value) >> 24 ) )
+                (cp)[0] ^= (uchar)((value)), \
+                (cp)[1] ^= (uchar)((value) >> 8), \
+                (cp)[2] ^= (uchar)((value) >> 16), \
+                (cp)[3] ^= (uchar)((value) >> 24 ) )
 
 inline int axcrypt_decrypt(__global const axcrypt_password *inbuffer, uint gid, __constant axcrypt_salt *cur_salt, __global axcrypt_out *output)
 {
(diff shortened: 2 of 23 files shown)


run/opencl/opencl_misc.h
```diff
@@ -302,6 +306,11 @@ inline MAYBE_VECTOR_UINT VSWAP32(MAYBE_VECTOR_UINT x)
 /*
  * These require (b) to be aligned!
  */
+/*
+ * Allow some strict aliasing violations, for nvidia only.
+ */
```
magnumripper: To do: Merge these comments to one

run/opencl/cryptsha512_kernel_DEFAULT.cl
```diff
@@ -31,6 +31,15 @@ inline void init_ctx(sha512_ctx * ctx) {
...
 }
 
+#if __OS_X__
+
+#define FAST
+#define memcpy_R memcpy_pp
+#define memcpy_C memcpy_cp
+#define memcpy_G memcpy_gp
+
+#else /* These violates strict aliasing rules */
```
solardiz: Grammar: either "This" or "violate".

run/opencl/opencl_misc.h
```diff
@@ -34,6 +34,10 @@ typedef uint64_t host_size_t;
 typedef uint32_t host_size_t;
 #endif
 
+#ifndef NULL
+#define NULL ((void*)0)
+#endif
```
solardiz, line 39: Why do we need to define `NULL`, and why didn't we need it before?

run/opencl/opencl_misc.h
```diff
@@ -393,239 +403,86 @@ inline int check_pkcs_pad(const uchar *data, int len, int blocksize)
...
-		}
-	}
+	const char *s = src;
+	char *d = dst;
 
-	while (count--) {
-		*d.c++ = *s.c++;
-	}
+	while (count--)
+		*d++ = *s++;
```
solardiz, line 417: Is `++` a defined operation for `void *` in OpenCL? (In C, it isn't or at least wasn't - it was a GNU extension.)

solardiz (commented):
I skimmed and didn't notice anything obviously wrong. Made some minor comments. Of course, more testing is needed, after you're done with the proposed changes. Thanks!

run/opencl/opencl_misc.h
magnumripper: I found it in `encfs_kernel.cl` and was surprised it's not defined in OpenCL. I just thought we should move it to `opencl_misc.h` before more files start defining it locally.

EDIT: Looking at it again, `encfs_kernel.cl` did not define NULL but it used a `((void*)0)` so I though we might want it.

magnumripper:
Interestingly enough; on Super's Vega, keepass and axcrypt2 OpenCL formats *start* failing with the new code. If anything I only saw the opposite on other OS's and devices.

magnumripper:
I took `mgetl.h` off this PR, it will be a separate one.

magnumripper:
Notes to self:
- Adding `#pragma unroll` to the memcpy_xx functions ***works around*** some AMD bug causing keepass-opencl to otherwise fail on Super's Vega, but it also ***causes*** ssh-opencl to fail on nvidia (CUDA 11.3 driver).

Why in the name of h*ll did they not just include a generic memcpy in the first draft OpenCL 0.1 spec and on? This whole issue is just stupid and time consuming.

magnumripper:
That `#pragma unroll` reliably boost several formats on nvidia though. I ran 20 test runs and used relbench.

I verified these top ones manually:
```
Ratio:  1.57391 real, 1.57060 virtual   tezos-opencl, Tezos Key:Raw
Ratio:  1.68609 real, 1.68525 virtual   electrum-modern-opencl, Electrum Wallet 2.8+:Raw
```

magnumripper:
I'm done with this. The only regression I can see is that keepass and axcrypt2 start failing on Super's ancient AMD driver. I did try to find workarounds but the bugger doesn't even do `printf` correctly so I'm out of tools (actually I *did* try lots of "throw stones at it and see what sticks" to no avail also).

IMHO we should merge this.

run/opencl/axcrypt2_kernel.cl
```diff
@@ -66,25 +70,25 @@ __kernel void axcrypt2_final(__global crack_t *pbkdf2,
...
-			memcpy_pp(wrapped, block, halfblocklen);
+			memcpy_pp(wrapped.c, block, halfblocklen);
 			// R[i] = LSB(B)
-			memcpy_pp(wrapped + k * halfblocklen, block + halfblocklen, halfblocklen);
+			memcpy_pp(wrapped.c + k * halfblocklen, block + halfblocklen, halfblocklen);
 		}
 	}
 
-	out[gid].cracked = !memcmp_pc(wrapped, "\xA6\xA6\xA6\xA6\xA6\xA6\xA6\xA6", 8);
+	out[gid].cracked = (wrapped.l[0] == 0xa6a6a6a6a6a6a6a6);
```
solardiz: Maybe the integer constant here needs a `ULL`?

run/opencl/axcrypt_kernel.cl
```diff
@@ -108,12 +109,7 @@ inline int axcrypt_decrypt(__global const axcrypt_password *inbuffer, uint gid,
 		lsb.w[1] = cipher.w[3];
 	}
 
-	for (i = 0; i < 8; i++) {
-		if (cipher.b[i] != 0xa6)
-			return 0;
-	}
-
-	return 1;
+	return (cipher.l[0] == 0xa6a6a6a6a6a6a6a6);
```
solardiz: Ditto.

run/opencl/keepass_kernel.cl
```diff
@@ -1,14 +1,26 @@
 /*
- * This software is Copyright (c) 2018 magnum,
+ * This software is Copyright (c) 2018-2021 magnum,
  * and it is hereby released to the general public under the following terms:
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted.
  */
+
+#if MAX_CONT_SIZE >= 0xff00
```
solardiz, line 8: What's `CONT` here, isn't it a typo of `CONST`? (I see it's moved from below in this file, rather than introduced with this PR, but anyway.)

solardiz (approved):
I skimmed and made some comments. Please consider those and make changes if appropriate.

solardiz:
> keepass and axcrypt2 start failing on Super's ancient AMD driver

keepass uses some `volatile` magic on NVIDIA - it's weird that this is needed, and suggests we probably have a bug in there.

Could axcrypt2 be failing because of missing `ULL` after instances of the 64-bit integer constant introduced with this PR? Please test this hypothesis.

magnumripper:
Good call spotting the missing ULL but it didn't help. Anyway I'll fix that, and these:
```
$ git grep -iP '0x[0-9a-f]{9,}[;\s]' ../run/opencl
../run/opencl/bitcoin_kernel.cl:                W[8] = 0x8000000000000000;
../run/opencl/opencl_hmac_sha512.h:             pW[0] ^= 0x3636363636363636;
../run/opencl/opencl_hmac_sha512.h:             pW[1] ^= 0x3636363636363636;
../run/opencl/opencl_hmac_sha512.h:             pW[2] ^= 0x3636363636363636;
../run/opencl/opencl_hmac_sha512.h:             pW[3] ^= 0x3636363636363636;
../run/opencl/opencl_hmac_sha512.h:             pW[4] ^= 0x3636363636363636;
../run/opencl/opencl_hmac_sha512.h:             pW[5] ^= 0x3636363636363636;
../run/opencl/opencl_hmac_sha512.h:             pW[6] ^= 0x3636363636363636;
../run/opencl/opencl_hmac_sha512.h:             pW[7] ^= 0x3636363636363636;
../run/opencl/opencl_hmac_sha512.h:                     pW[i] ^= 0x3636363636363636;
../run/opencl/opencl_hmac_sha512.h:             pW[i] ^= (0x3636363636363636 ^ 0x5c5c5c5c5c5c5c5c);
../run/opencl/sspr_kernel.cl:           W[8] = 0x8000000000000000;
```
With the axcrypt2 on AMD I also tried reverting to the memcmp but that did not help either.

run/opencl/keepass_kernel.cl
magnumripper: I think it's an odd abbreviation for content. I'll leave it alone.

magnumripper:
> > keepass and axcrypt2 start failing on Super's ancient AMD driver
> 
> keepass uses some `volatile` magic on NVIDIA - it's weird that this is needed, and suggests we probably have a bug in there.

I merged this now but will have another look at keepass separately.

solardiz:
> Good call spotting the missing ULL but it didn't help. Anyway I'll fix that, and these:

OK. Actually, it should probably be `UL`, not `ULL`: https://github.com/KhronosGroup/WebCL-conformance/issues/61

magnumripper:
Oh, that was news to me but very logical in hindsight since in OpenCL `long` is always 64-bit. I'll fix that (we have lots of `ULL` in there other than the ones I did now) in a separate commit.

magnumripper:
> I'll fix that (we have lots of `ULL` in there other than the ones I did now) in a separate commit.

I recalled incorrectly, the ones I added now were the only ones present. This means I must have known about this before - I'm getting old  o.O

Anyway, fixed in a95fd4f2d and I also ensured we didn't use `long long` anywhere in OpenCL. We did not.

magnumripper:
> > keepass and axcrypt2 start failing on Super's ancient AMD driver
> 
> keepass uses some `volatile` magic on NVIDIA - it's weird that this is needed, and suggests we probably have a bug in there.

I seriously doubt it's a bug of ours. I found that the workaround is excessive, it can be minimized to only this (and it's only needed in the third and last call to it from keepass):
```diff
  AES_cbc_decrypt(AES_SRC_TYPE void *_in, AES_DST_TYPE void *_out,
                uint len, AES_KEY *akey,
                void *_iv)
  {
-	AES_SRC_TYPE uchar *in = _in;
+ 	volatile AES_SRC_TYPE uchar *in = _in;
	AES_SRC_TYPE uchar *in = _in;
	AES_DST_TYPE uchar *out = _out;
	uchar *iv = _iv;

	while (len) {
		uint n;
		uchar tmp[16];

		memcpy_macro(tmp, in, 16);
		AES_decrypt(tmp, tmp, akey);
		for (n = 0; n < 16 && n < len; ++n) {
			uchar c = in[n];
			out[n] = tmp[n] ^ iv[n];
			iv[n] = c;
		}
		if (len <= 16) {
			for (; n < 16; ++n)
				iv[n] = in[n];
			break;
		}
		len -= 16;
		in  += 16;
		out += 16;
	}
  }
```
That same `AES_cbc_decrypt()` works perfectly fine in 22 other kernels without this `volatile` - plus there's even two other calls to it in KeePass that don't need it. Only the last one does. Here's that one:
```c
	else if (salt->version == 2) {
		uchar content[32];

		memcpy_macro(content, salt->contents, 32);

		if (salt->algorithm == 0) {
			AES_set_decrypt_key(hash, 256, &akey);
			AES_cbc_decrypt(content, hash, 32, &akey, iv);  /* <-- this call */
		} else /* if (salt->algorithm == 2) */ {
			(...)
		}
		result[gid].cracked = !memcmp_pmc(hash, salt->expected_bytes, 32);
	}
```

PR #4745, opened 2021-06-08
mgetl.h: Avoid strict aliasing violations.

### LINKED ISSUE #4733: mgetl() strict-aliasing violation
https://github.com/openwall/john/pull/4246#issuecomment-610801229
*Yes, the problem with mgetl() violating strict aliasing rules isn't new, and yes our check for ARCH_ALLOWS_UNALIGNED mitigates the problem in practice so far - but technically it's still UB, and a future compiler/version might exploit it. I think a proper fix would be use of union types (or pointers to them) for the global variables and for the mgetl() argument. The union types need to include fields of all types that we access the data through, eliminating the need for typecasts.*

*BTW, we might want to increase the alignment of line in wordlist.c from ARCH_WORD to vtype, which should allow us not to use vstoreu. This might provide some speedup in some scenarios on some CPUs.*

magnumripper:
Use pointers to a union type.  Also increase the alignment of line in wordlist.c from ARCH_WORD to vtype when applicable, allowing us to use vstore instead of vstoreu.

Closes #4733

### DIFF
--- a/src/Makefile.in
+++ b/src/Makefile.in
@@ -374,7 +374,7 @@ listconf.o:	listconf.c autoconfig.h os.h os-autoconf.h jumbo.h arch.h simd-intri
 
 LM_fmt.o:	LM_fmt.c arch.h misc.h jumbo.h autoconfig.h memory.h DES_bs.h common.h loader.h params.h list.h formats.h os.h os-autoconf.h
 
-loader.o:	loader.c autoconfig.h jumbo.h arch.h os.h os-autoconf.h misc.h params.h path.h memory.h list.h signals.h formats.h dyna_salt.h loader.h options.h getopt.h common.h config.h unicode.h dynamic.h simd-intrinsics.h pseudo_intrinsics.h aligned.h simd-intrinsics-load-flags.h fake_salts.h john.h cracker.h logger.h base64_convert.h showformats.h
+loader.o:	loader.c mgetl.h autoconfig.h jumbo.h arch.h os.h os-autoconf.h misc.h params.h path.h memory.h list.h signals.h formats.h dyna_salt.h loader.h options.h getopt.h common.h config.h unicode.h dynamic.h simd-intrinsics.h pseudo_intrinsics.h aligned.h simd-intrinsics-load-flags.h fake_salts.h john.h cracker.h logger.h base64_convert.h showformats.h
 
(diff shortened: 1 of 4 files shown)


magnumripper:
I get this from gcc 4.4.7 (`20120313 (Red Hat 4.4.7-23)`) on super
```
loader.c: In function 'ldr_fill_user_words':
loader.c:1977: error: dereferencing pointer 'line' does break strict-aliasing rules
loader.c:1989: error: dereferencing pointer 'line' does break strict-aliasing rules
loader.c:1906: note: initialized from here
cc1: warnings being treated as errors
wordlist.c: In function 'do_wordlist_crack':
wordlist.c:1412: error: dereferencing pointer 'line' does break strict-aliasing rules
wordlist.c:1412: error: dereferencing pointer 'line' does break strict-aliasing rules
wordlist.c:463: note: initialized from here
```
wordlist.c:1412 isn't even a correct line number, there's no `line` anywhere near it.
The loader.c errors refer to things like this:
```c
	if (!*line->c || !(pw = strchr(line->c, ':'))) {
```
Now how is that breaking strict-aliasing rules? We fixed it and the compiler *started* to complain 🙄

magnumripper:
Also, my benchmarks suggest the SIMD stuff is excellent but for non-SIMD the 8/4-byte stuff give a really marginal boost. I'm tempted to use to the much simpler version with just either SIMD or byte-wise. Admittedly the SIMD stuff break aliasing rules unless the intrinsic types are `__may_alias__` but we haven't seen any warnings.

EDIT: Our current (lack of) pseudo-intrinsics rule out anything but intel using that code path anyway: It needs `vcmpeq_epi8_mask` and currently also disqualifies emulated `vloadu`. So we always have `__may_alias__` for mgetl.

magnumripper:
Using gcc 8.3 on a Raspberry pi, I get
```
mgetl.h:104:26: error: array subscript -1 is below array bounds of ‘char[1]’ [-Werror=array-bounds]
  if (pos > line && pos->c[-1] == '\r')
                    ~~~~~~^~~~
mgetl.h:105:9: error: array subscript -1 is below array bounds of ‘char[1]’ [-Werror=array-bounds]
   pos->c[-1] = 0;
   ~~~~~~^~~~
```
Hard to argue with that, it should be rewritten. But it'll be even messier.

PR #4747, opened 2021-06-08
cracker.c: "pause file" fixes

### LINKED ISSUE #4746: Pause file status update
From recent private conversation:
> Looks like we force printing of status line along with "Pause file seen", but not with "Pause file removed" - we probably should.

Also, some odd log file timestamp was seen when Pause file removed, only the first time?!

magnumripper:
After sleeping, fixup the times before calling log_event(), so we don't get an out-of-whack timestamp for that entry.
Also, call status_print() again after waking up.

Closes #4746

There's nothing really wrong with this patch except we should consider [some alternatives](https://github.com/openwall/john/issues/4746#issuecomment-856762315) first: Should we really alter the "time", and if we do, should we really call for a second status output?

### DIFF
--- a/run/john.conf
+++ b/run/john.conf
@@ -323,11 +323,11 @@ ReloadAtDone = Y
 # but it may be delayed by the "Save" timer setting near top of this file.
 ReloadAtSave = Y
 
-# If this file exists, john will abort cleanly
-AbortFile = /var/run/john/abort
+# If this file exists, john will abort cleanly (uncomment to enable)
+#AbortFile = /var/run/john/abort
 
-# While this file exists, john will pause
-PauseFile = /var/run/john/pause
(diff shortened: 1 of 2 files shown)


magnumripper:
Updated according to discussion in #4746

solardiz (approved):
Looks good to me.

I see we're no longer adding `status_print();` after resuming, I guess because with the time adjustment it'd be expected to print an identical line to what was printed before the pause? Have you tested and confirmed this was the case?

magnumripper:
> I see we're no longer adding `status_print();` after resuming, I guess because with the time adjustment it'd be expected to print an identical line to what was printed before the pause? Have you tested and confirmed this was the case?

I admit I did not, but they should really be close. We could add one anyway just for kicks but I didn't feel it was necessary.

PR #4749, opened 2021-06-11
Incremental: Allow supplying a charset file as "mode"

magnumripper:
This lets you use eg. --inc=temp.chr without any entry in john.conf

### DIFF
--- a/doc/NEWS
+++ b/doc/NEWS
@@ -196,6 +196,9 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
 - Support --target-encoding with --make-charset, for generating charsets for
   formats like LM that need to be in a legacy codepage.  [magnum; 2021]
 
+- Incremental: Allow supplying a charset file name as "mode", with no config
+  entry required.  [magnum; 2021]
+
 
 Major changes from 1.8.0-jumbo-1 (December 2014) to 1.9.0-jumbo-1 (May 2019):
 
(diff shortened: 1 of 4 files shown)


frank-dittrich:
You could also adjust this john.conf part (adding a comment that .chr files can be used directly instead of a mode):
```
# This is for one-off uses (make your own custom.chr)
[Incremental:Custom]
File = $JOHN/custom.chr
MinLen = 0
```

frank-dittrich:
Should I adjust bash completion to add all existing .chr files (and all names of subdirectories where might be some .chr files)
Especially including subdirectories could be confusing.

solardiz (approved):

magnumripper:
> Should I adjust bash completion to add all existing .chr files (and all names of subdirectories where might be some .chr files)
> Especially including subdirectories could be confusing.

That would be great, but not super important. You could leave out subdirs unless it was already specified, such as `--inc=dir/[tab]`

magnumripper:
> You could also adjust this john.conf part (adding a comment that .chr files can be used directly instead of a mode):
> 
> ```
> # This is for one-off uses (make your own custom.chr)
> [Incremental:Custom]
> File = $JOHN/custom.chr
> MinLen = 0
> ```

That's a good point. I didn't want to drop this but the comment could be amended.

magnumripper:
> > Should I adjust bash completion to add all existing .chr files (and all names of subdirectories where might be some .chr files)
> > Especially including subdirectories could be confusing.

The more I think about this, the messier it gets. We wouldn't want to list `.chr` files that *have* a section. Probably best to just leave it alone.

PR #4751, opened 2021-06-15
mgetl.h: Avoid strict aliasing violations.

### LINKED ISSUE #4733: mgetl() strict-aliasing violation

magnumripper:
I reverted to KISS code: Keep the SIMD stuff (only used with intel so `MAY_ALIAS`), but use byte-wise copy for all other situations (100% aliasing compliant code for wider copy too hard to achieve and not worth the negligible boost). Also increase the alignment of line in wordlist.c from `ARCH_WORD` to `vtype` when applicable, allowing us to use vstore instead of vstoreu.

Closes #4733

### DIFF
--- a/src/Makefile.in
+++ b/src/Makefile.in
@@ -374,7 +374,7 @@ listconf.o:	listconf.c autoconfig.h os.h os-autoconf.h jumbo.h arch.h simd-intri
 
 LM_fmt.o:	LM_fmt.c arch.h misc.h jumbo.h autoconfig.h memory.h DES_bs.h common.h loader.h params.h list.h formats.h os.h os-autoconf.h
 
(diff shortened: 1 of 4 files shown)


solardiz (approved):
Looks OK to me, and I assume it's been tested.

PR #4756, opened 2021-06-21
Fix iteration parse

necromortis:
Fix iteration parse from 524288 to 2048

### DIFF
--- a/run/pwsafe2john.py
+++ b/run/pwsafe2john.py
@@ -39,7 +39,7 @@ def process_file(filename):
         sys.std.write("Error: salt read failed.\n")
         return
 
-    iterations = struct.unpack(">I", f.read(4))[0]
+    iterations = struct.unpack("<I", f.read(4))[0]
 
     sys.stdout.write("%s:$pwsafe$*3*" %
                      os.path.basename(filename).rstrip(".psafe3"))


solardiz:
@kholia @exploide Can you help review this, please?

solardiz:
@NecroMortis Thanks. Can you describe the issue in more detail, please? It sounds like we were parsing the iteration count in the wrong endianness, is that what you meant? Why not describe this as such in the commit message, then, instead of giving two specific numbers? How do you know which endianness/number is right? Do you have an input file that you're able to crack the password for after the fix, but not before the fix? If so, what system did this file come from? Related, didn't we have our own test input file(s) where the script worked correctly as-is? I think here: https://github.com/openwall/john-samples/tree/main/Password_Safe

necromortis:
> @NecroMortis It sounds like we were parsing the iteration count in the wrong endianness, is that what you meant? 

Maybe, I did not fully figure out what's the matter, but the script parse 0x80000 as 524288. It is not right, right iteration count is 2048.
Maybe endianness right, and need parse without last byte (0x800)

exploide:
This PR seems to be reasonable. According to the [specification](https://github.com/pwsafe/pwsafe/blob/master/docs/formatV3.txt) section 2.4, iteration count is stored in little-endian. So this change makes sense.

I checked that against the examples you mentioned @solardiz and indeed, before this change the hashes are not crackable. 

```
> john --wordlist=./wordlist hashes-big 
Loaded 4 password hashes with 4 different salts (pwsafe, Password Safe [SHA256 256/256 AVX2 8x])
Cost 1 (iteration count) is 524288 for all loaded hashes
Will run 8 OpenMP threads
Warning: Only 4 candidates left, minimum 64 needed for performance.
0g 0:00:00:00 DONE (2021-06-22 18:32) 0g/s 6.060p/s 24.24c/s 24.24C/s Gorilla..password
Session completed. 

> john --wordlist=./wordlist hashes-little 
Loaded 4 password hashes with 4 different salts (pwsafe, Password Safe [SHA256 256/256 AVX2 8x])
Cost 1 (iteration count) is 2048 for all loaded hashes
Will run 8 OpenMP threads
Warning: Only 4 candidates left, minimum 64 needed for performance.
openwall         (pwsafe-openwall)     
openwall         (Password Gorilla - openwall)     
password         (pwsafe-password)     
12345678         (pwsafe-12345678)     
4g 0:00:00:00 DONE (2021-06-22 18:32) 80.00g/s 80.00p/s 320.0c/s 320.0C/s Gorilla..password
Session completed. 
```

Note the detected iteration counts are exactly the numbers reported by @NecroMortis.

So I think this is fine to merge.

However, while testing, I noticed the script is not compatible with python3. So after you merged this, expect a follow up PR from me (maybe during the weekend, with a bit of luck earlier.)

exploide (approved):

solardiz:
Thanks, @exploide! I find it puzzling that the script didn't process our own samples right. Didn't @kholia test it at all?

Also, should we possibly print a warning from `john` (or/and have it swap endianness of this number) when it sees input files created by `pwsafe2john.py` from prior to this change? If so, the most reliable way to do it would be to change the hash prefix string. Otherwise we'd have to use heuristics (treat 524288 as likely wrong). This doesn't prevent us from merging this PR as-is, but it's something we might want to enhance in a follow-up PR (separate commit from the Python 3 compatibility).

Oh, and false negatives from the old versions were a significant enough issue to have these changes mentioned in `NEWS` - again, not necessarily in this PR, but maybe in your follow-up PR, @exploide!

solardiz:
@NecroMortis

> Maybe endianness right, and need parse without last byte (0x800)

I think "without last byte" happening to give the right value is a coincidence. See https://en.wikipedia.org/wiki/Endianness in case you're not familiar with that word.

exploide:
Ok, I have the compat patch ready. Was easy enough.

> Also, should we possibly print a warning from john (or/and have it swap endianness of this number) when it sees input files created by pwsafe2john.py from prior to this change? If so, the most reliable way to do it would be to change the hash prefix string. Otherwise we'd have to use heuristics (treat 524288 as likely wrong). This doesn't prevent us from merging this PR as-is, but it's something we might want to enhance in a follow-up PR (separate commit from the Python 3 compatibility).

I have no big opinion on this. Personally, I just consider it a bug being fixed for the same format and I wouldn't change it. But I don't know if and how much trouble it could cause. So I leave it as is now and leave the decision to you.

> Oh, and false negatives from the old versions were a significant enough issue to have these changes mentioned in NEWS - again, not necessarily in this PR, but maybe in your follow-up PR

Sure, will include a note.

solardiz:
> change the hash prefix string

Alternatively, keep the hash prefix the same, but add an extra field indicating that we've fixed this bug. That way, output from new versions of the script would work for both old and new versions of `john`, yet new versions will also be able to detect input from old versions of the script and swap endianness. I think I like this best. @magnumripper What do you think?

If we want to implement this, we should do it (at least add the field, which is trivial to do) at most within a few days after merging this PR, so we'll be able to disregard very brief existence of versions of the script that would have the bug fixed but no indication of that yet.

exploide:
> Alternatively, keep the hash prefix the same, but add an extra field indicating that we've fixed this bug. That way, output from new versions of the script would work for both old and new versions of john, yet new versions will also be able to detect input from old versions of the script and swap endianness. I think I like this best.

I haven't looked how your hash format parser works but is it robust enough to cope with an additional field in the hash string? In terms of, can an old John parse a changed pwsafe hash format?
Easiest is probably to keep it as is but let modern Johns print a warning when an unreasonably high iteration count is detected. Something like "Warning: Cost 1 (iteration count) is 524288 which is likely the result of creating the hash with an old and bugged version of pwsafe2john.py".

But you decide and let's see what @magnumripper thinks.

magnumripper:
> Easiest is probably to keep it as is but let modern Johns print a warning when an unreasonably high iteration count is detected. Something like "Warning: Cost 1 (iteration count) is 524288 which is likely the result of creating the hash with an old and bugged version of pwsafe2john.py".

With just the info I have right now I agree with this.

solardiz:
I'll merge this now to let @exploide create the follow-up Python 3 compatibility PR. After that is merged as well, let's possibly revisit handling the buggy script's output better. Thanks!

PR #4760, opened 2021-06-26
Fix type error in truecrypt2john.py

### LINKED ISSUE #4759: Truecrypt2john fails to run
When running truecrypt2john.py on Python 3.9.x (more env details belw), the script stops with the following output:

```
File "~\john\run\truecrypt2john.py", line 94, in <module>
process_file(remainder[0], keyfiles, options)
File "~\john\run\truecrypt2john.py", line 47, in process_file
sys.stdout.write(binascii.hexlify(header))  
TypeError: write() argument must be str, not bytes
```

I was able to fix this locally by wrapping string interpolation around the calls to `binascii.hexlify(header)` within `sys.stdoutput.write()` here and on line 67, and will have a PR open momentarily with these changes.

Here are the system configurations I used to reproduce this:
* PowerShell 7 on Windows 10, with Python 3.9.0, running both 1.9.0-jumbo-1 and bleeding-jumbo
* Zsh on EndeavourOS (an Arch derivative), with Python 3.9.5, running 1.9.0-jumbo-1

chaosharmonic:
Closes #4759

### DIFF
--- a/run/truecrypt2john.py
+++ b/run/truecrypt2john.py
@@ -44,7 +44,7 @@ def process_file(filename, keyfiles, options):
 
     for tag in tags:
         sys.stdout.write("%s:%s$" % (basename(filename), tag))
-        sys.stdout.write(binascii.hexlify(header))
+        sys.stdout.write(binascii.hexlify(header).decode('ascii'))
         if keyfiles:
             nkeyfiles = len(keyfiles)
             sys.stdout.write("$%d" % (nkeyfiles))
@@ -64,7 +64,7 @@ def process_file(filename, keyfiles, options):
 
     for tag in ["truecrypt_RIPEMD_160", "truecrypt_SHA_512", "truecrypt_WHIRLPOOL"]:
         sys.stdout.write("%s:%s$" % (basename(filename), tag))
-        sys.stdout.write(binascii.hexlify(header))
+        sys.stdout.write(binascii.hexlify(header).decode('ascii'))
         if keyfiles:
             nkeyfiles = len(keyfiles)
             sys.stdout.write("$%d" % (nkeyfiles))


solardiz:
Thanks @chaosharmonic!

I wonder why we aren't(?) running into the same issue with other scripts (has something changed in Python 3.9 maybe, and we just didn't yet have enough people try those on Python this recent and complain to us?) and whether this PR preserves support for older Python as well (ideally, even Python 2, or if not then the shebang line will need to be adjusted).

@exploide Please help review this.

exploide:
I haven't run that script by myself yet but this looks like it was never compatible with Python 3 before. So this is one of the typical string vs. byte-string issues.

@chaosharmonic you fixed that by introducing format string literals. This appears to work "by accident" but an explicit conversion would be cleaner. Besides this breaks Python 2 compatibility because format string literals are no thing there.

So I propose you use `binascii.hexlify(header).decode('ascii')` instead.

chaosharmonic:
Gotcha -- yeah, I'll test that locally and then push again once I've confirmed that it still works. I'm admittedly not super familiar w Python conventions, as my knowledge set skews more toward JavaScript, so while I had figured I could accomplish this w a literal anyway I also just wasn't entirely sure how to do explicit type coercion, and backwards compatibility wasn't really on my mind at the time.

chaosharmonic:
Tangential question -- when I do get an output from this, John still fails to parse any of its hashes. Is that because of the changes to how the file is getting written, or a distinct problem I would need to troubleshoot further?

exploide:
> Tangential question -- when I do get an output from this, John still fails to parse any of its hashes. Is that because of the changes to how the file is getting written, or a distinct problem I would need to troubleshoot further?

My typical procedure is as follows:

1. First, run the original script explicitly with python2 on one or more sample files and store the hashes.
2. Then check if they are crackable with John.
3. If yes, let's proceed with making the script compatible. To do so, run the script with python3 and see where it crashes. In this case, it's the part with writing byte strings instead of ascii strings. Then fix it and repeat until the script finishes correctly. (Instead of trial and error a more careful approach might be preferable to check if your tests covered all code branches and special cases, but this script here is comparably simple.)
4. Store the new hashes and diff the output with the hashes produced by the original version. If they differ, something is wrong. I could imagine that this is the point where you encounter the problem. This needs to be debugged.
5. Finally, rerun the script with python2 to check if the original hashes, the python3 hashes and the new python2 hashes are all the same.

If you need assistance, I'm happy to help. If this is no fun for you at all, I could even do it entirely. As you like :)

chaosharmonic:
It looks after running through these -- and also after confirming how to run parallel installations of Python 2 and 3 -- that the string literal was indeed altering the output. The method you reference above works across both, and the outputs hashes that John is able to properly detect. I've updated the branch accordingly.

solardiz:
Thank you @chaosharmonic! For merging this, it should be just one commit, not two. So can you please either replace the two commits with one (I suggest commit message "truecrypt2john.py: Python 3 support") and force-push or tell us to use GitHub's squash-and-merge on this PR (which results in GitHub appearing as the committer in git history, which isn't pretty, but we can live with it if you don't feel like force-pushing).

chaosharmonic:
@solardiz just force-pushed this as a single commit. Should be good now.

solardiz (approved):
Looks good to me, thanks!

PR #4761, opened 2021-06-30
Add new --catch-up=NAME option

magnumripper:
This limits a session to running only until it reaches the candidates tried count of a different, existing and paused session NAME.  This can be used for "adding" more hashes to an existing job:  After the new session finishes, just concatenate the new hashes to the older session's hash file and resume the old session.  This will end up more or less as if both sets of hashes were in the original session to start with.

Obviously it's intended for salted hashes and mainly for things like incremental mode. It should be very usable for some situations we've had in contests in the past: If the original session had 100 salts and ran for several hours, we can "add" a new hash to it in minutes without missing a beat.

I've done this in the past using `--max-run-time` which is very imprecise. I then did it with `--max-candidates` which is much better but requires you to parse the old session file's `status.cands`. Setting config option `StatusShowCandidates` to true helps, but in case of fork/MPI you need to add up the numbers (but the sum you used will just be divided according to node count so if nodes' progress varied it will end up less precise again).

This option simply reads `status.cands` *per node* so will end up exactly where the old session ended or very slightly ahead. Other than that parsing, we mostly just piggy-back the existing `--max-candidates` code.

### DIFF
--- a/src/recovery.c
+++ b/src/recovery.c
@@ -207,6 +207,9 @@ void rec_init(struct db_main *db, void (*save_mode)(FILE *file))
 
 	rec_db = db;
 	rec_save_mode = save_mode;
+
+	if (options.catchup)
+		john_max_cands = rec_read_cands(options.catchup);
 }
 
 static void save_salt_state()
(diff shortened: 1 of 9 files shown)


frank-dittrich:
I like this option.
I have a (trivial) patch pending to support bash completion for `--catch-up` (the same completion logic as for `--restore` and `--status`)

magnumripper:
> I have a (trivial) patch pending to support bash completion for `--catch-up` (the same completion logic as for `--restore` and `--status`)

Great! Assuming you're a maintainer on Openwall you can simply add it to this PR.

src/recovery.c
```diff
@@ -684,3 +687,63 @@ void rec_restore_mode(int (*restore_mode)(FILE *file))
...
+		sprintf(suffix, ".%u%s", options.node_min, RECOVERY_SUFFIX);
+		other_name = path_session(session, suffix);
+	} else {
+		other_name = path_session(session, RECOVERY_SUFFIX);
+	}
+
+	if (!strcmp(other_name, rec_name))
+		error_msg("New session name can't be same as catchup session\n");
+
+	if (!(other_file = fopen(other_name, "r+")))
```
magnumripper: Maybe this should be `r` and not `r+`? I wasn't quite aware of the meaning of `+` when I copied the code from `rec_restore_args()`. We do lock the file until we exit but we're not supposed to write to it.

solardiz (approved):
This looks good to me conceptually. Probably the implementation will as well, but I am too tired to review it for real right now.

magnumripper:
@frank-dittrich this is it, right?
```diff
diff --git a/run/john.bash_completion b/run/john.bash_completion
index 94e1b3386..64055920c 100644
--- a/run/john.bash_completion
+++ b/run/john.bash_completion
@@ -418,7 +418,7 @@ _john()
                        COMPREPLY[${#COMPREPLY[@]}]="${prev}"
                        return 0
                        ;;
-               --restore=*|--status=*)
+               --restore=*|--status=*|--catch-up=*)
                        cur=${cur#*[=:]}
                        # .rec files are stored in the current directory (or a
                        # subdirectory if the session name contains a slash)
```

Issue #4734, opened 2021-05-22
wordlist change/grew as we read it; but it did not - integer overflow?


BotPass:
### problem ###
I used a big file with more than 2^32 lines, forks and I get a lot of "wordlist change as we read it" and "wordlist grew as we read it" messages. That was strange because it was a local file and no other application used it. The size and timestamp of the file did not change and the problem was reproducible.
Not all forks run properly.

### assumed reason ###
After search for the messages I found it in src/wordlist.c in function `do_wordlist_crack`.
Conditions are `if (i > my_size) {`for error "grew" and `if (nWordFileLines != myWordFileLines)` for warning "changed".
Now looking at the data types:
```
static int64_t nWordFileLines;
int i,
long my_size
unsigned int myWordFileLines
```
Thus only `nWordFileLines` has a fixed size while (as far as I know) `i`, `my_size` and `myWordFileLines` differ depending on multiple circumstances; e.g. `my_size` should be 32 bit under 64 bit Windows (using LLP64) while it is 64 bit under 64 bit Linux (using LP64). It's probably more complicated with cygwin and cross build with mingw under Linux for Windows.

But in this case my system was Linux. Thus the second one, the changed warning could be explained because `myWordFileLines` is 32 bit (unsigned), `nWordFileLines` is 64 bit (signed) and the number of lines was to big for `myWordFileLines` but not for `nWordFileLines`.

The possible explanation for the first one -- the grew error -- is more complicated. Because even if there is an overflow in `i` it could not lead to be bigger than `my_size`. I am not sure if this is the problem, but the values of `i` and `my_size` are updated based on `skip` while `skip` based on `for_node` (both 32 bit signed). For `i` `for_node` is based on `myWordFileLines` and for `my_size` it is based on `nWordFileLines`. Thus they are probably updated differently.

### workaround ###
For test purpose all those variables (and some other variables) were changed to 64 bit. Now no of these message appear and it seems that all forks were running.

Because I only looked up the source code and did not debug I am not sure if the overflow is really the problem here. But in my case the change seems to fix the problem. However there could be other negative side effects by this change.

magnumripper:
Thank you for reporting. We've tried to support whatever larger-than-32-bit `seek` and similar functions is available but we might have missed the case where the number of *lines* overflows 32-bit signed.

Just for the record, please post the output of `./john --list=build-info`

BotPass:
Ubuntu 20.04 is used
```
[...]

PR #4764, opened 2021-07-03
wordlist.c: Use (u)int64_t instead of long

magnumripper:
Also promote a couple of int's to int64_t.  The changes should allow files with billions of lines (duh) to work correctly.

This addresses #4734 but I couldn't quite reproduce the problems.


solardiz (approved):
The changes look good to me, although the code being changed isn't great and remains so.

PR #4770, opened 2021-07-18
Update DYNAMIC_COMPILER_FORMATS.md

ghost:
An addition.

### DIFF
--- a/doc/DYNAMIC_COMPILER_FORMATS.md
+++ b/doc/DYNAMIC_COMPILER_FORMATS.md
@@ -121,7 +121,7 @@ Loaded 1 password hash (dynamic=md5($u.$c1.$p) [256/256 AVX2 8x3])
 test             (user)
 ```
 
-JBoss hashes can be created by using the "add-user.sh" utility included with JBoss.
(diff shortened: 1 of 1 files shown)


solardiz:
To me, this looks weird. While the example is probably on Kali Linux (as seen from the shell prompt), it's unclear whether it in fact makes any use of Kali Linux specifics.

@v45145 What is your reason to propose this specific edit?

Besides the specifics of this one, perhaps it's a similar general reason for the 3 documentation edit PRs you sent to other projects as well? Are you trying to help, trying to learn how to contribute to projects, trying to gain PR count, or/and running some kind of social experiment?

I don't mind anyhow, am just curious - and maybe we could offer advice on more than just one PR. Thanks!

solardiz:
Upon a closer look, I am tempted to close this without merging. I think it has negative value because people will think the example is somehow only valid on Kali Linux, whereas I see nothing like that in it. Different wording like "Illustrated here on Kali Linux" instead of the proposed "in case of Kali Linux" would be OK, but I think wouldn't add value. So unless anyone explains why this edit is needed, let's not make it.

solardiz:
No rationale provided. Closing.

PR #4772, opened 2021-07-20
Fix legacy build

youngblonde:
From the looks of it it was b0rken again since some time.


solardiz:
We already have this old commit by magnum in the tree. I assume this PR was created by mistake.

PR #4773, opened 2021-07-23
remove trailing whitespace

noraj:


frank-dittrich:
That space character is not a trailing whitespace character. It is there on purpose.

noraj:
> That space character is not a trailing whitespace character. It is there on purpose.

I'm curious, can you tell me its purpose?

ghost:
We appreciate your enthusiasm, but if you test the format/your patch, you'llfindthereasoning. Don'tyou?

Issue #4777, opened 2021-08-04
Cracking not working with DPAPImk, DPAPI masterkey file v1 and v2


mpgn:
Hello,

When trying to crack DPAPI hash, no password is found while I provide the good password.

```
└─$ python DPAPImk2john.py.1 -S S-1-5-21-3902381005-3621640295-1501945936-500 -c domain -mk fc4691a6-29c0-47e2-b484-8edd06508327 --password October2021 --debug                                                                         2 ⨯
$DPAPImk$1*2*S-1-5-21-3902381005-3621640295-1501945936-500*des3*sha1*18000*e0c423993bd2f0a870090f2292c9ba6c*208*3c2cd5955d05c0694ff18a2bbe722b53f94fa476f8bea46193bf5f13dc4d695442c604c59881d53230aca4532e73399b2fca93f6ca03cc0ed460c3c7d8e0aef1ec4a51ca8fa304e0529d9d7d2188015eb754010e4770182bddaa0e3017ef4ebf39d3ae2056aaea9b
$DPAPImk$1*3*S-1-5-21-3902381005-3621640295-1501945936-500*des3*sha1*18000*e0c423993bd2f0a870090f2292c9ba6c*208*3c2cd5955d05c0694ff18a2bbe722b53f94fa476f8bea46193bf5f13dc4d695442c604c59881d53230aca4532e73399b2fca93f6ca03cc0ed460c3c7d8e0aef1ec4a51ca8fa304e0529d9d7d2188015eb754010e4770182bddaa0e3017ef4ebf39d3ae2056aaea9b
[
#### MasterKeyFile fc4691a6-29c0-47e2-b484-8edd06508327 ####
        version   = 2
        Policy    = 0x0
        MasterKey = 136
        BackupKey = 104
        DomainKey = 372
    + Master Key: Masterkey block
        cipher algo  = DES3 [0x6603]
        hash algo    = HMAC [0x8009]
        rounds       = 18000
        IV           = e0c423993bd2f0a870090f2292c9ba6c
        ciphertext   = 3c2cd5955d05c0694ff18a2bbe722b53f94fa476f8bea46193bf5f13dc4d695442c604c59881d53230aca4532e73399b2fca93f6ca03cc0ed460c3c7d8e0aef1ec4a51ca8fa304e0529d9d7d2188015eb754010e4770182bddaa0e3017ef4ebf39d3ae2056aaea9b
    + Backup Key: Masterkey block
        cipher algo  = DES3 [0x6603]
[...]

PR #4778, opened 2021-08-04
Made DPAPImk2john.py compatible with python3

### LINKED ISSUE #4638: DPAPImk2john.py is Python 2 only

exploide:
Made `DPAPImk2john.py` compatible with python3.

I took the example from  #4777 as test data. It worked out well but I can not guarantee that it hit all relevant code branches. However, on the first look it works. 

fixes #4638

### DIFF
--- a/run/DPAPImk2john.py
+++ b/run/DPAPImk2john.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
 # Modified (for JtR) by Dhiru Kholia in December, 2014.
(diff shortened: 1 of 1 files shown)


ghost (approved):
Merging so more people can use and test it.

PR #4783, opened 2021-08-09
john.conf: add external "wordcase" mode

rofl0r:
expands an input phrase, e.g. "foo bar" into all possible combinations
with words separated by space capitalized or not:
foo bar
Foo bar
Foo Bar
foo Bar

the code is identical to 'case' external mode, with the addition of one
condition in the new() function.

### DIFF
--- a/run/john.conf
+++ b/run/john.conf
@@ -3974,8 +3974,8 @@ void next()
 }
 /* restore() not needed. john properly restores fast enough without it */
 
-# External hybrid CaSE mutation code
-[List.External:Case]
+# Shared base code for External hybrid CaSE and Wordcase mutation code
+[List.External_base:Case]
 
 int rotor[251]; /* max length input is 125 bytes [125*5+1]; */
 int rotors[125];
@@ -3985,12 +3985,7 @@ int rotor_cnt[125];
 int current_word_count;
 int max_mangle; /* controls how many bytes we run through our 'leet' code */
 int original_word; /* if set to 1 then we start with original word. If 0, then start with first mangled word */
-
-void init()
-{
-	max_mangle = 20; /* only mangle 20 characters max (2^20 is 1 million) */
-	original_word = 1; /* for case mangle, unless the data is 100% lower case, we really can not skip the original word */
-}
+int word_mode; /* if set to 1, only first character of each space-separated word is case-toggled, else every character */
 
 /* new word */
 void new()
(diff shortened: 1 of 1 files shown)


solardiz:
> the code is identical to 'case' external mode, with the addition of one
> condition in the new() function.

Maybe we should avoid code duplication between these modes by using `.include` of a shared section?

@magnumripper What do you think? I think you "pioneered"  that approach with some other modes.

magnumripper:
> @magnumripper What do you think? I think you "pioneered" that approach with some other modes.

It sometimes makes the code harder to read/follow, but as long as whole functions are re-used (like in this case I presume) it might be a good idea.

rofl0r:
i don't know about the abilities of john's "C script", but if it would work to call a named function
we could basically rename new in external:case to new_with_mode(int mode)
and then make eternal:case have new() that just calls new_with_mode(0) and wordcase new_with_mode(1) and then just enable/disable the additional branch depending on whether mode is set.

magnumripper:
> i don't know about the abilities of john's "C script", but if it would work to call a named function
> we could basically rename new in external:case to new_with_mode(int mode)

Unfortunately that's not supported; https://raw.githubusercontent.com/openwall/john/bleeding-jumbo/doc/EXTERNAL

What you could do is what Jim (I think) did with `List.External_base:Double` and variant functions such as `[List.External:Double]`: You could just have different `init()` for the original "case" and your new mode (introducing a new variable instead of using a function argument).

rofl0r:
> What you could do is what Jim (I think) did with List.External_base:Double and variant functions such as [List.External:Double]: You could just have different init() for the original "case" and your new mode (introducing a new variable instead of using a function argument).

good advice, thanks. updated PR accordingly.

run/john.conf
```diff
@@ -4008,18 +4003,24 @@ void new()
...
-			rotor[rotor_off++] = ch;
-			rotor[rotor_off++] = ch-0x20;
-			rotor_cnt[wlen] = 2;
-			hybrid_total *= 2;
-			rotors[idx++] = wlen;
+		/* traditionally, this block was always executed, with Wordcase
+		   mode added we execute it either always when word_mode isn't
+		   used or at the beginning of a word or when the previous char
+		   was space */
+		if (word_mode == 0 || wlen == 0 || word[wlen-1] == 0x20) {
```
rofl0r: this is the new condition, however it relies on short-circuit evaluation assuming that when wlen == 0 the test word[wlen-1] == 0x20 is skipped; so this might introduce an out-of-bounds memory access. any ideas how to best fix this ?

solardiz: Conservative changes just to this place:

```c
int prevch;
prevch = ' ';
if (wlen) prevch = word[wlen - 1];
if (!word_mode || prevch == ' ') {
```

A better fix would probably be to initialize prevch out of the loop and to add `prevch = ch;` near the `++wlen;` line.

BTW, `!word_mode` is a bit quicker than `word_mode == 0`. (The external mode compiler is mostly non-optimizing.)

rofl0r: thanks! pushed a new commit addressing this, i made it into a separate commit because i thought this might be interesting for other log readers, but feel free to squash as desired.

solardiz (approved):
Looks OK'ish to me, although I didn't test it.

rofl0r:
thanks, i did some quick testing and it seemed correct. while i have you around would you mind telling me where the code for the external compiler/vm is located? it seems to perform pretty well so i'd like to study its design.

solardiz:
> where the code for the external compiler/vm is located?

It's in `compiler.c`.

While it's mostly non-optimizing, there are a couple of optional optimizations: the use of GNU C "Labels as Values" (when building with GCC or compatible) and the existence/generation of multi-push VM instructions (a way to lower VM overhead by doing more work per instruction). So the basic logic implementing the same functionality (but slower) doesn't require all of the code that's present in the source file. Yet even with those optimizations, the file is pretty small (for what it does) and I hope will be fun for you to study. And yes, I wrote it, so I'm flattered by your interest in it.

PR #4785, opened 2021-08-09
john.conf: add external:wpspin wps pin generator

rofl0r:
generates all possible 10000000 wps pins with their checksum.

### DIFF
--- a/run/john.conf
+++ b/run/john.conf
@@ -3497,6 +3497,40 @@ void restore()
 	}
 }
 
+# generate all possible wps pins
+[List.External:wpspin]
+int pin;
+void init() {
+	pin = 0;
+}
+void generate() {
+	if (pin > 9999999) {
+		word = 0;
+		return;
+	}
+	int i, p;
+	i = 0;
+	while (i < 8) word[i++] = '0';
+	word[8] = 0;
+	p = pin;
+	i = 6;
+	while (p) {
+		word[i] = '0' + p % 10;
+		p /= 10;
+		--i;
+	}
+	p = pin;
+	i = 0;
+	while (p) {
+		i += 3 * (p % 10);
+		p /= 10;
+		i += p % 10;
+		p /= 10;
+	}
+	word[7] = '0' + ((10 - i % 10) % 10);
+	++pin;
+}
+
 # Append the Luhn algorithm digit to arbitrary all-digit strings.  Optimized
 # for speed, not for size nor simplicity.  The primary optimization trick is to
 # compute the length and four sums in parallel (in two SIMD'ish variables).


run/john.conf
```diff
@@ -3497,6 +3497,42 @@ void restore()
...
 
+# generate all possible wps pins
+[List.External:wpspin]
+int pin;
+void init() {
+	pin = 0;
+}
+void generate() {
+	if(pin > 9999999) {
+		word[0] = 0;
```
solardiz: Can do `word = 0;`, like we use in other external modes.

run/john.conf
```diff
@@ -3497,6 +3497,42 @@ void restore()
...
+	}
+	int i, p;
+	i = 0;
+	while(i < 8) word[i++] = '0';
+	word[8] = 0;
+	p = pin;
+	i = 6;
+	while(p) {
+		word[i] = '0' + (p % 10);
+		p = p / 10;
```
solardiz: Can use `p /= 10;`

run/john.conf
```diff
@@ -3497,6 +3497,42 @@ void restore()
...
+		--i;
+	}
+	p = pin;
+	i = 0;
+	while(p) {
+		i = i + (3 * (p % 10));
+		p = p / 10;
+		i = i + p % 10;
+		p = p / 10;
```
solardiz: Can use `i +=` and `p /=` as appropriate on all 4 of these lines.

run/john.conf
```diff
@@ -3497,6 +3497,42 @@ void restore()
...
+		return;
+	}
+	int i, p;
+	i = 0;
+	while(i < 8) word[i++] = '0';
+	word[8] = 0;
+	p = pin;
+	i = 6;
+	while(p) {
+		word[i] = '0' + (p % 10);
```
solardiz: Unneeded braces around `p % 10`. You don't use those in another place anyway.

run/john.conf
```diff
@@ -3497,6 +3497,42 @@ void restore()
...
+		i = i + (3 * (p % 10));
+		p = p / 10;
+		i = i + p % 10;
+		p = p / 10;
+	}
+	word[7] = '0' + ((10 - i % 10) % 10);
+	++pin;
+}
+void restore() {
+}
```
solardiz: Bug: defining `restore()` tells JtR that this format can restore, whereas it doesn't actually infer the value of `pin` from `word` in that function. I suggest you simply drop `restore()`.

solardiz (changes requested):
I request some minor changes here, most notably dropping of `restore()`. I suggest that you also adjust the coding style to match JtR's - see e.g. the style used in `AppendLuhn` below (an external mode that comes from JtR core). Thanks!

run/john.conf
rofl0r: thanks for review. i took this approach from external:case. updated.

run/john.conf
```diff
@@ -3497,6 +3497,40 @@ void restore()
...
+	while(i < 8) word[i++] = '0';
+	word[8] = 0;
+	p = pin;
+	i = 6;
+	while(p) {
+		word[i] = '0' + p % 10;
+		p /= 10;
+		--i;
+	}
+	p = pin;
```
rofl0r, line 3522: tried `while(i >= 0) word[i--] = '0';` here instead of the explicit init on line 3512/13; though i couldn't reliably test for speed differences, sometimes old variant was faster, sometimes new one.

solardiz (approved):

Issue #1742, opened 2015-09-10
Formats which should handle input hash case changes in some manner (fixing or rejecting)


jfoug:
Ok, after adding test type #4 (true generic), ,I get these problems (some need white listed, not all are bugs).

$ ../run/john -test-full=1 | grep -v PASS
Will run 8 OpenMP threads

- [x] *FP* Testing: bcrypt ("$2a$05", 32 iterations) [Blowfish 32/64 X3]... (8xOMP) FAILED (split() is only casing sometimes (#4))
- [x] Testing: chap, iSCSI CHAP authentication [MD5 32/64]... (8xOMP) FAILED (should not set FMT_SPLIT_UNIFIES_CASE (#4))
- [x] Testing: EPI, EPiServer SID [SHA1 32/64]... (8xOMP) FAILED (should not set FMT_SPLIT_UNIFIES_CASE (#4))
- [x] Testing: openssl-enc, OpenSSL "enc" encryption [32/64]... FAILED (should not set FMT_SPLIT_UNIFIES_CASE (#4))
- [x] Testing: PuTTY, Private Key [SHA1/AES 32/64]... (8xOMP) FAILED (should not set FMT_SPLIT_UNIFIES_CASE (#4))
- [x] Testing: rar, RAR3 (4 characters) [SHA1 128/128 AVX 4x AES]... (8xOMP) FAILED (should not set FMT_SPLIT_UNIFIES_CASE (#4))
- [x] Testing: wbb3, WoltLab BB3 [SHA1 32/64]... (8xOMP) FAILED (should not set FMT_SPLIT_UNIFIES_CASE (#4))

Device 0: Cedar [ATI Radeon HD 5450]
- [x] Testing: rar-opencl, RAR3 (length 5) [SHA1 OpenCL AES]... (8xOMP) FAILED (should not set FMT_SPLIT_UNIFIES_CASE (#4))
- [x] Testing: lotus5-opencl, Lotus Notes/Domino 5 [OpenCL]... FAILED (should not set FMT_SPLIT_UNIFIES_CASE (#4))
- [x] *FP* Testing: bcrypt-opencl ("$2a$05", 32 iterations) [Blowfish OpenCL]... FAILED (split() is only casing sometimes (#4))

- [x] Testing: keychain-opencl, Mac OS X Keychain [PBKDF2-SHA1 OpenCL 3DES]... (8xOMP) FAILED (should not set FMT_SPLIT_UNIFIES_CASE (#4))
[...]

PR #4792, opened 2021-08-30
added the $oracle11$ FORMAT_TAG for oracle11 format

### LINKED ISSUE #4791: FORMAT_TAG for oracle11 hash format
oracle and oracle12c have a FORMAT_TAG, but oracle11 does not.
it's pretty common to dump hashes for several versions at a time into
the same file, and the autodetect works for the two first ones, but for
the latter, you need to run john with a mandatory --format=oracle11
separately (a bit of a pain).

i'm currently working on a pull request to add that feature to the oracle11
format source code. any warning about this? any "DONT!"?

thanks

rom1sqr:
Edit by magnum: This closes #4791

### DIFF
--- a/src/oracle11_fmt_plug.c
+++ b/src/oracle11_fmt_plug.c
@@ -73,6 +73,8 @@ john_register_one(&fmt_oracle11);
 
 #define FORMAT_LABEL			"oracle11"
 #define FORMAT_NAME			"Oracle 11g"
+#define FORMAT_TAG			"$oracle11$"
+#define FORMAT_TAG_LENGTH		(sizeof(FORMAT_TAG) - 1)
 
 #define ALGORITHM_NAME			"SHA1 " SHA1_ALGORITHM_NAME
 
@@ -116,6 +118,10 @@ static struct fmt_tests tests[] = {
 	{"3437FF72BD69E3FB4D10C750B92B8FB90B155E26227B9AB62D94F54E5951", "oracle"},
 	{"61CE616647A4F7980AFD7C7245261AF25E0AFE9C9763FCF0D54DA667D4E6", "11g"},
 	{"B9E7556F53500C8C78A58F50F24439D79962DE68117654B6700CE7CC71CF", "11g"},
+
+	/* with FORMAT_TAG */
+	{"$oracle11$7233E3B91B45F6B813BCFFB5D8669167CB4F498D0642558A8A3BB39948C0", "testpwd"},
+
 	{NULL}
 };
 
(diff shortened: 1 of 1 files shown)


src/oracle11_fmt_plug.c
```diff
@@ -165,20 +171,52 @@ static void done(void)
...
+	char *p = ciphertext;
+
+	if (!memcmp(ciphertext, FORMAT_TAG, FORMAT_TAG_LENGTH))
+		p += FORMAT_TAG_LENGTH;
+
+	if (strlen(p) != CIPHERTEXT_LENGTH)
+		return 0;
+
+	if (!ishex(p))
+		return 0;
```
magnumripper, line 185: I think you changed the format from only accepting uppercase hex, to any case. Was this intentional? I'm not sure what format you usually get when doing whatever one does to fetch oracle hashes, so I can't tell - it might well be sensible.

Anyway if we do this I think we should unify case in split as well.

rom1sqr: You're absolutely right.

I actually did it intentionnally. I was extracting the hashes directly from the output of `hexdump -C` applied to an ORA password file, and the hex output is lower case there. My local test file contains a hash in upper case, and one in lower case as well. I was wondering if this change had any implication whatsoever.

It seems this has been changed in commit 81391c139d (fixed split-case bugs.  Some for #1742).

I'm going add a new commit to this Pull Request to either revert this change back, or unify case in a custom split function.
I'm going to have a look at this as soon as i can. Thanks for your feedback.

PS: i'm also going to have a look at the "Heap Buffer Overflow" reported by CircleCI, sorry about that

magnumripper: You should also add `FMT_SPLIT_UNIFIES_CASE` format flag.

magnumripper (commented):
I believe it's better to add the tag in `split()` as well, and drop the `prepare()` function. See https://github.com/openwall/john/wiki/split()-vs.-prepare().

Also, you could use `strupr()` instead of `enc_strupper()`, the latter is encoding aware but that's overkill for hex.

rom1sqr:
Alright! I'll fix this as soon as possible. Thanks for your feedback.

ghost (approved):
Passes TS and TS --internal. 

LGTM

src/oracle11_fmt_plug.c
```diff
@@ -165,20 +173,49 @@ static void done(void)
...
 static void *get_salt(char *ciphertext)
 {
 	static unsigned char *salt;
+	char *p = ciphertext;
 	int i;
 
 	if (!salt) salt = mem_alloc_tiny(SALT_SIZE, MEM_ALIGN_WORD);
 
+	if (!memcmp(ciphertext, FORMAT_TAG, FORMAT_TAG_LENGTH))
+		p += FORMAT_TAG_LENGTH;
```
magnumripper, line 213: Technically we can now do this unconditionally (and same for `get_binary()`) since `ciphertext` has passed `split()` so always has the tag. But there's no harm in keeping it this way.

src/oracle11_fmt_plug.c
```diff
@@ -67,12 +67,15 @@ john_register_one(&fmt_oracle11);
 #include "misc.h"
 #include "common.h"
 #include "formats.h"
+#include "unicode.h"
```
magnumripper, line 70: I assume you added this to get `strupr()` but it actually sits in `jumbo.h`. This works as well because `unicode.h` in turn sources that header. It's not very important but in case you end up doing any other change, you could change it.

magnumripper (approved):
Looks good to me now too, despite my nitpicks.

magnumripper:
I guess we should squash this when merging

src/oracle11_fmt_plug.c
rom1sqr: Yes, as we moved the tag addition from `prepare()` to `split()`, i could remove the condition here.

src/oracle11_fmt_plug.c
rom1sqr: Yes, i added `unicode.h` because i was previously using `enc_strupper()` but you're right, i could change that back to `jumbo.h`

PR #4793, opened 2021-08-30
Single mode: Bug fix for when to call clear_keys()

magnumripper:
I added a call in fb051f3b after seeing it was sometimes missed, but it led to many double calls and even a fatal [for a few SIMD formats] situation in the end of a single run because it would end up with a call to crypt_all() with the key buffer in limbo.

This fix should do the Right Thing in all situations.

@solardiz I think the diff just happens to include the "need to know" context to be easily reviewed. And I'm so sure this is correct I almost pushed it directly.

### DIFF
--- a/src/cracker.c
+++ b/src/cracker.c
@@ -1232,11 +1232,12 @@ int crk_process_salt(struct db_salt *salt)
 	count_from_guesses = salt->keys->count_from_guesses;
 	index = 0;
 
-	crk_methods.clear_keys();
-
 	while (count--) {
 		strnzcpy(key, ptr, options.eff_maxlength + 1);
 		ptr += options.eff_maxlength;
+
+		if (index == 0)
+			crk_methods.clear_keys();
 
 		crk_methods.set_key(key, index++);
 		if (index >= crk_params->max_keys_per_crypt || !count ||
@@ -1269,7 +1270,6 @@ int crk_process_salt(struct db_salt *salt)
 			if (!salt->list)
 				return 0;
 			index = 0;
-			crk_methods.clear_keys();
 		}
 	}
 


magnumripper:
BTW the original problem exists in john proper as well: When `index` is reset to 0 in the lower part of the loop (but with `count` not having reached 0), `clear_keys()` is never called before next batch.

solardiz:
> BTW the original problem exists in john proper as well: When `index` is reset to 0 in the lower part of the loop (but with `count` not having reached 0), `clear_keys()` is never called before next batch.

I think I just hadn't considered this a problem back then and in core. Indeed, `clear_keys()` is documented as an optimization and best-effort safety measure, not as something guaranteed to be called nor required to do anything:

```c
/* Allow the previously set keys to be dropped if that would help improve
 * performance and/or reduce the impact of certain hardware faults. After
 * a call to clear_keys() the keys are undefined. */
        void (*clear_keys)(void);
```

In fact, that's still the description in jumbo. If we actually guarantee `clear_keys()` being called reliably and/or rely on it doing something specific, then we should probably update that comment in jumbo.

magnumripper:
I believe some or even many formats (mainly fast SIMD ones) rely on it. Perhaps I should update that description then. The alternative would be to audit all formats having their own `clear_keys()` and change them to do whatever cleaning they need in `set_key(0)`. That shouldn't be too hard either. What do you prefer? Oh, and just for my curiosity, what was your need for such function that made you add it?

solardiz:
> What do you prefer?

I prefer to update the description in jumbo.

> Oh, and just for my curiosity, what was your need for such function that made you add it?

IIRC, it was to assist the old DES `set_key()`, which updated the key schedule relative to the previous password in the same slot. Dropping the old keys once in a while would prevent potential (hardware) errors from accumulating. I don't recall what performance improvements I might have referred to.

PR #4795, opened 2021-08-31
--regen-lost-salts option: Bug fixes, closes #4517

### LINKED ISSUE #4517: --regen-lost-salts NULL pointer dereference
As reported by @kernelr00ter in https://github.com/openwall/john/issues/1463#issuecomment-742524567:

```
in file hash_dump.txt
c4ca4238a0b923820dcc509a6f75849b
d3d9446802a44259755d38e6d163e820

run john --regen-lost-salts
Result:
0                (?)

*** Process received signal ***
Signal: Segmentation fault (11)
Signal code: Address not mapped (1)
Failing at address: (nil)
[ 0] /lib/x86_64-linux-gnu/libpthread.so.0(+0x11390)[0x7f4d13061390]
[ 1] /john[0x47b785]
[ 2] /john[0x6bee8b]
[ 3] /john[0x6bfe30]
[ 4] /john[0x6c0605]
[ 5] /john[0x6dd359]
[ 6] /john[0x6cdd51]
[ 7] /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf0)[0x7f4d12ca6840]
[ 8] /john[0x407099]
*** End of error message ***

In this case, the utility correctly detects the hash result, but does not show the salt. And the report file remains empty.

But if you swap hashes
d3d9446802a44259755d38e6d163e820
c4ca4238a0b923820dcc509a6f75849b

Everything works perfectly.
```

Reproduced by me:

```
$ cat pw
c4ca4238a0b923820dcc509a6f75849b
d3d9446802a44259755d38e6d163e820
$ ./john pw --regen-lost-salts='@dynamic=md5($s.$p)@:32:?d' --mask='?d' --format='dynamic=md5($s.$p)' pw
Loaded 2 password hashes with no different salts (dynamic=md5($s.$p) [128/128 AVX 4x3])
Warning: no OpenMP support for this hash type, consider --fork=32
Warning: Only 10 candidates left, minimum 24 needed for performance.
0                (?)     
Segmentation fault
```

magnumripper:
Since every salt copied the same salt->list pointer, a special fixup is needed after removing *the first* binary in the list in crk_process_guess(). Also, the generated salts written to the salt database lacked some info and the total number of salts wasn't updated.

Less important updates:
Error messages pointed to a non-existing file in a non-existing directory.
Change the "Remaining xx hashes with yy different salts" output to be more accurate when this feature is used.

### DIFF
--- a/doc/Regen-Lost-Salts.txt
+++ b/doc/Regen-Lost-Salts.txt
@@ -1,42 +1,43 @@
-***************************************************************************
-* JtR --regen-lost-salts code, written by JimF, 2012-2015, for use
-* within the salted dynamic formats.
-*
-* No copyright is claimed, and the software is hereby
-* placed in the public domain. In case this attempt to disclaim
-* copyright and place the software in the public domain is deemed
-* null and void, then the software is Copyright (c) 2012-2015 JimF
-* and it is hereby released to the general public under the following
-* terms:
-*
-* This software may be modified, redistributed, and used for any
-* purpose, in source and binary forms, with or without modification.
-***************************************************************************
+Salt brute-force feature:  --regen-lost-salts
 
+This option will allow hashes to be cracked even where the salt, normally
+stored together with the hash, was somehow lost.  Normally a cracker would not
+be able to attack such hashes.  Using this feature, only supported for dynamic
+formats (including "dynamic compiler" ones), JtR will load the hashes and
+brute-force all possible salts for each.
 
---regen-lost-salts=type:hash_sz:mask
+Obviously this comes at a price - but not always:  If you attack a single hash
+with 1000 possible salts, it will end up with same same p/s as an attack on
+1000 uniquly salted hashes.  If on the other hand the number of hashes is
+larger than 1000, you'll end up with no penalty (well, until you've cracked so
+many that there are less than 1000 left to crack) and for very short salt
+sizes and/or large number of hashes, normal "same-salt boosts" may even apply.
 
-This option will allow certain types (with short salts), to be found, when
-they are present in a set of raw hashes, but where we have lost the salt value.
-Normally, without the salt, JtR would not be able to do anything with the hash,
-and could not find the password.  In this mode, JtR will load the hashes, giving
-each a 'fake' salt (but all get the same salt).  JtR then builds all of the
-possible salt values for this format, and associates every hash with each one
-of these salts.  So, JtR will now run, and find passwords using all possible
-salts for each of the hashes, thus recreating the salt (and finding the passwords).
+The feature should only be used with bare hashes input, meaning the ciphertext
+should be, for example, raw hex such as d3d9446802a44259755d38e6d163e820 without
+any tag such as the "$dynamic_xx$" prefix and obviously without the "$babe" salt
+suffix.  Other fields such as login/gecos are optionally supported though, just
+like they always are:
 
-This function has recently been re-written.  The prior way of usage is still supported
-(limited), and should be considered depricated.  See the end of this document for
-the *** NOTE section.
+	user:d3d9446802a44259755d38e6d163e820:101:100:Joe Random Luser
+
+All salts should be missing in the input file; Mixing missing-salt hashes with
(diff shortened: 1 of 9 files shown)


magnumripper:
Still testing this. There's a slight problem, still:
```
$ ../run/pass_gen.pl dynamic_4 <../run/password.lst | shuf > test.in
$ sed -r 's/\$..:/:/' test.in > test.nosalts.in

$ ../run/john test.nosalts.in --regen-lost-salts='dynamic_4:32:?y?y' -format=dynamic_4 -w
Loaded 1500 password hashes with 9025 possible salts (1805.0x same-salt boost) (dynamic_4 [md5($s.$p) (OSC) 256/256 AVX2 8x3])
integra          (u897:fa)     
indigo           (u260:ga)     
007007           (u1137:ga)     
(...)
dundee           (u520:49)     
frogs            (u1417:79)     
gremlin          (u1438:99)     
1499g 0:00:00:01 DONE (2021-09-01 00:21) 1094g/s 2588p/s 23359Kc/s 48003MC/s Hammer..sss
Remaining 1 password hash with 9025 possible salts
Use the "--show --format=dynamic_4" options to display all of the cracked passwords reliably
Session completed. 

$ ../run/john test.nosalts.in -format=dynamic_4 -show=left
u1290:$dynamic_4$8b2c2e07ac0cb51062832f7210d19e80:1290

1499 password hashes cracked, 1 left

$ grep 8b2c2e07ac0cb51062832f7210d19e80 test.in
u1290:$dynamic_4$8b2c2e07ac0cb51062832f7210d19e80$aa:1290:0:cannon:
```
It missed the 'aa' salt. This may well be a separate bug, I'll look into it.

magnumripper:
Notes to self:
- Sometimes (different shuffles) we instead get 1501 cracks and a dupe pot entry, that's a data point... and when it misses one it's always the 'aa' salt.
- If the test file doesn't contain any 'aa' salt, all hashes are cracked.
- If the test file contain several 'aa' salts, none of them are cracked.
- I see pot entries aren't written with the cracked salt inserted. Not sure if that was intentionally left out but I think we can do that, using `source()`, if we want to.
- With `ShowUIDinCracks = Y`, I get cracks like this:
```
aaron            (u456:E9)     
cactus           (u487:E9)     
```
...and they correspond to these:
```
$ grep '\$E9:' test.noshuf.in 
u456:$dynamic_4$92acf59298a88ae17c81095ba435deba$E9:456:0:aaron:
u487:$dynamic_4$982f9f8be1e1b0ad91e058bbc25fb984$E9:487:0:cactus:
```
...so what is supposed to be the uid is actually the brute-forced salt. I don't yet understand why this happens. It should have showed like this:
```
aaron            (u456:456)     
cactus           (u487:487)     
```

magnumripper:
> It missed the 'aa' salt. This may well be a separate bug, I'll look into it.

This is a separate bug, Jim's code never outputs the 'aa' salt... oh well, I'll fix that too.

magnumripper:
> This is a separate bug, Jim's code never outputs the 'aa' salt... oh well, I'll fix that too.

No, it actually does. It's written to the original single-salt at load time, before the generation of all other salts. Still, I confirmed this bug was present before my changes - it always misses the 'aa' salt.

magnumripper:
More problems found: Some code isn't handling the dynamic compiler format or bare hashes correctly. Also, `--show` does not work (never did AFAICS). I'm looking into both.

magnumripper:
Hopefully, all problems are now cared for. The code always relied on "bare hashes" (ie. with no format tag) so that's now enforced (and documented). I also added a way to use `--show`: It simply relies on you to still supply the `--regen-lost-salts` option with same parameters as the ones used for cracking (or at least ending up same salt length).

Countless of other problems were fixed. I bet some remain but it's definitely better than ever.

magnumripper:
Note that since the `--regen-lost-salts` option takes a format as its first parameter (how stupid is that!?) we now infer the `--format` option from it if not given, or if it *was* given as well, we enforce they are the same.

PR #4805, opened 2021-09-11
Add cryptoSafe formats

### LINKED ISSUE #4804: Add cryptoSafe format
https://www.openwall.com/lists/john-users/2021/09/03/3
https://github.com/Anubis901/SafeCrypto

Test vector (password "foobar"):
RJ+YWtuyOoLnnOpOdmj43+hwO5cquMsSO3f/OiQfINofM+c0JVbIyUTQg3St+1Ue09QiZKlNnCAmlPHq2wuI31QoZQ/KNKZT/VMLd3qYbkOv873HT4wsYeavSwOqrdVdxpEimkqgdDkGP7XHfeFUCv6+jXFZioiR8jYrgxA8fDQQ5C+YReuvfiqLGaVrC9ih58X8Q7NLxAlQTGSQvHgsZI6DNfAsXw3Zt++fIJh9bdLzHDEBF/pLp47zj132UOYr6TuufnnA+HeUISdk+xH6/w==

Input can be a vault file, as-is. Pot entries get `$cryptosafe$1$` prefix.

Algo: Base64-decode the input. Pad the password candidate with ASCII "0" to 32 bytes, ~iv is 16x ASCII "0" as well~. Use padded password directly as a key to AES-256-CBC and decrypt the input. Output is PKCS padded (early rejection) and ~within the first two blocks there should be a constant string of `publicKey`~.

Edit:  iv is 16 null bytes, first block starts with constant string `[{"coinName":"` (so no early rejection needed).

Test vector correctly decrypted:
```
[{"coinName":"ETH","publicKey":"0x4ac97B48CbBF9D54AE1f07bF72b10F19eBE30BB1","privateKey":"a5f54b647905db05a82d523fe0027a5da9760c2e97e4124448896f7333cdf96f","seed":"","comment":"","id":"1631298033694"}]
```

magnumripper:
Closes #4804

### DIFF
--- a/src/cryptosafe_common_plug.c
+++ b/src/cryptosafe_common_plug.c
@@ -1,0 +1,65 @@
+/*
+ * This software is Copyright (c) 2021, magnum
+ * and it is hereby released to the general public under the following terms:
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted.
+ */
+
+#include "formats.h"
+#include "memory.h"
+#include "base64_convert.h"
+#include "cryptosafe_common.h"
+
+struct fmt_tests cryptosafe_tests[] = {
+	{"$cryptosafe$1$RJ+YWtuyOoLnnOpOdmj43+hwO5cquMsSO3f/OiQfINofM+c0JVbIyUTQg3St+1Ue09QiZKlNnCAmlPHq2wuI31QoZQ/KNKZT/VMLd3qYbkOv873HT4wsYeavSwOqrdVdxpEimkqgdDkGP7XHfeFUCv6+jXFZioiR8jYrgxA8fDQQ5C+YReuvfiqLGaVrC9ih58X8Q7NLxAlQTGSQvHgsZI6DNfAsXw3Zt++fIJh9bdLzHDEBF/pLp47zj132UOYr6TuufnnA+HeUISdk+xH6/w==", "foobar"},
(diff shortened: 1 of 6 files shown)


magnumripper:
~I'll add an OpenCL format with internal mask to this PR soon-ish.~

Done

src/cryptosafe_common_plug.c
```diff
@@ -0,0 +1,65 @@
...
+
+	return 1;
+}
+
+char *cryptosafe_split(char *ciphertext, int index, struct fmt_main *self)
+{
+	if (!strncmp(ciphertext, FORMAT_TAG, FORMAT_TAG_LEN))
+		return ciphertext;
+
+	char *out = mem_alloc_tiny(strlen(ciphertext) + FORMAT_TAG_LEN + 1, MEM_ALIGN_NONE);
```
solardiz, line 40: FWIW, `split()` is supposed to use a static buffer (maybe allocated on first call, but just once), not a new buffer each time. I doubt this format will be used on sufficiently many inputs for this to matter, but thought I'd point this out anyway.

magnumripper: Yes, but this format is `FMT_HUGE_INPUT` because the input can be arbitrarily large. Not because what we store in the db for cracking is huge at all, but because the full ciphertext might be and the pot file entry might need truncation. So I'd need to use a ridiculous size for a static buffer (and still not be safe), or do it this way!

magnumripper: What I could do is use a buffer that is realloc'd if/when a larger size is needed, but like you said, it would probably not matter.

Issue #4806, opened 2021-09-15
pfSense/OPNsense user hash extract script


privwolf:
Hello,
I would like to create a pfSense/OPNsense hash extract script PR

ghost:
Good. JtR developers encourage you to contribute. 

- Get some tips at https://github.com/openwall/john/blob/bleeding-jumbo/CONTRIBUTING.md. 
- Take a look at the source code for other 2john tools (examples).
- If you choose Python, please make it compatible with 3.x versions.

Once you have a tool that works, post a draft PR and ask for guidance in the review process.

solardiz:
This is taken care of in #4807. Thank you for the contribution, @privwolf!

PR #4807, opened 2021-09-15
pfSense/OPNsense hash extract script

privwolf:
#4806

### DIFF
--- a/run/sense2john.py
+++ b/run/sense2john.py
@@ -1,0 +1,36 @@
+#!/usr/bin/env python
+
+# This software is Copyright (c) 2021 Private Wolf <privatewolf@inbox.lv>,
+# and it is hereby released to the general public under the following terms:
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted.
+
+import sys
+import xml.etree.ElementTree as ET
+
+def process_file(filename):
+    try:
+        tree = ET.parse(filename)
+    except IOError:
+        e = sys.exc_info()[1]
+        sys.stderr.write("%s\n" % str(e))
+        return
+
+    root = tree.getroot()
+
+    for user in root.findall('system/user'):
+        username = user.find('name').text
+        if user.find('bcrypt-hash') is not None:
+            sys.stdout.write("%s:%s\n" % (username, user.find('bcrypt-hash').text))
+        if user.find('md5-hash') is not None:
+            sys.stdout.write("%s:%s\n" % (username, user.find('md5-hash').text))
+        if user.find('password') is not None:
+            sys.stdout.write("%s:%s\n" % (username, user.find('password').text))
+
+if __name__ == "__main__":
+    if len(sys.argv) < 2:
+        sys.stderr.write("Usage: %s <pfSense/OPNsense config.xml>\n" % sys.argv[0])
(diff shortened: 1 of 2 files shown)


run/sense2john.py
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env python
```
solardiz, line 1: Does this script support both Python 2 and 3? If so, great. If not, we need to specify `python2` or `python3` here.

run/sense2john.py
```diff
@@ -0,0 +1,32 @@
...
+
+def process_file(filename):
+    try:
+        fd = open(filename, "rb")
+    except IOError:
+        e = sys.exc_info()[1]
+        sys.stderr.write("%s\n" % str(e))
+        return
+
+    tree = ET.parse(filename)
```
solardiz: The script never uses `fd` obtained from `open`. Perhaps it'd be more appropriate to omit the `open` and to move the `ET.parse` inside a `try`.

run/sense2john.py
```diff
@@ -0,0 +1,32 @@
...
+    root = tree.getroot()
+
+    for user in root.findall('system/user'):
+        username = user.find('name').text
+        if user.find('bcrypt-hash') is not None:
+            sys.stdout.write("%s:%s\n" % (username, user.find('bcrypt-hash').text))
+        elif user.find('md5-hash') is not None:
+            sys.stdout.write("%s:%s\n" % (username, user.find('md5-hash').text))
+        elif user.find('password') is not None:
+            sys.stdout.write("%s:%s\n" % (username, user.find('password').text))
```
solardiz, line 28: Is it possible that more than one hash type would be present for a user? If so, the script might want to output multiple lines per user, one per hash. I think this would be achieved by simply using `if` in place of `elif`.

run/sense2john.py
```diff
@@ -0,0 +1,32 @@
...
+
+if __name__ == "__main__":
+    if len(sys.argv) < 2:
+        sys.stderr.write("Usage: %s <pfSense/OPNsense config.xml>\n" % sys.argv[0])
+        sys.exit(-1)
```
solardiz: Should be more like `sys.exit(1)` here (not `-1`). We also have this bug in many other scripts, but no reason to continue to replicate it.

solardiz (commented):
I'm not into Python, but I've made some comments.

solardiz:
@exploide You might want to help us review this.

@privwolf Please add a copyright+license comment to the start of the script, preferably like this:

```python
# This software is Copyright (c) YEAR YOUR NAME <your at e-mail.address>,
# and it is hereby released to the general public under the following terms:
# Redistribution and use in source and binary forms, with or without
# modification, are permitted.
```

Please also add an entry to `doc/NEWS`.

When you make those changes, please amend your one commit and force-push. Thanks!

run/sense2john.py
solardiz: Passes basic testing with both Python 2 and 3 - great! - but I don't have a proper XML file to test this for real... which reminds me - please send us a sample file like this via a pull request against https://github.com/openwall/john-samples

solardiz (approved):
Looks good to me now, thanks!

PR #4808, opened 2021-09-15
Update bestcryptve2john.py

trounce1:
Added copyright+license

### DIFF
--- a/run/bestcryptve2john.py
+++ b/run/bestcryptve2john.py
@@ -1,3 +1,9 @@
+# This software is Copyright (c) 2021, trounce1 / HN
+# and it is hereby released to the general public under the following terms:
+#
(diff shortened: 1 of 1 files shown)


solardiz (changes requested):
The added copyright+license comment should be _after_ the shebang line - that is, between `#!/usr/bin/env python3` and `from binascii import hexlify, unhexlify`. Please move it to there, perhaps also adding a blank line to visually separate the comment from the shebang. Thanks!

trounce1:
> The added copyright+license comment should be _after_ the shebang line - that is, between `#!/usr/bin/env python3` and `from binascii import hexlify, unhexlify`. Please move it to there, perhaps also adding a blank line to visually separate the comment from the shebang. Thanks!

done

Issue #4031, opened 2019-06-15
zip-opencl suboptimal speeds


solardiz:
As prompted by this Twitter thread:

https://twitter.com/solardiz/status/1139844055100059650

It appears that we have a crucial optimization that hashcat missed, so we appear to be much faster than hashcat at these "non-hashes" now, but while at it I also noticed that:

1. We benchmark for a mix of two different "non-hashes", and the speeds for those differ somewhat. As seen with actual mask mode cracking on GTX 1080, Titan X Maxwell, Titan Kepler, respectively:

`$zip2$*0*1`: 2206K, 1543K, 693K

`$zip2$*0*3`: 1970K, 1415K, 656K

Perhaps we should standardize on the former for our benchmarks, which is what hashcat just did (with a couple of commits today: https://github.com/hashcat/hashcat/commit/f660b95ac6ad30aca8d3540641c2abf4592780a5 https://github.com/hashcat/hashcat/commit/cf3dd5f3db6c3f32f54f01001d1c2c3772831d3a). I think for us this means changing the 0x107 to 0x507. Right?

2. Our `--test` benchmark speeds are a lot higher than what's seen in actual cracking. For example, on GTX 1080 verbose auto-tune and `--test` gives ~2700K, but as we can see above actual cracking is at ~2100K (average for the two test vectors, and is actually seen when I load both for cracking at once). How can we fix and/or explain this?

3. GPU utilization for the faster GPUs is on the order of 85% during cracking. Why is that? Are we still doing much work on host? Can we fix that?

I expect that once hashcat implements the missed optimization, it'll actually be faster than us by at least this missed 15% of GPU utilization.

4. When we compute the 2 bytes of PBKDF2 and occasionally have to (re)compute the rest, perhaps we can skip the PBKDF2 output block that included the 2 bytes? When we compute the 2 bytes of PBKDF2 and occasionally have to (re)compute the rest, perhaps we can skip the PBKDF2 output block that included the 2 bytes _if_ they were part of a block that we need anyway? (We'd need to store the full block, not just the 2 bytes, for that to work.) It appears that the performance cost of the code divergence the full PBKDF2 computation occasionally causes is actually measurable and not entirely negligible. So we'd save time not only for the work-item doing this full computation, but also for its 63 adjacent ones, by having it compute one output block less. But perhaps this would clatter our source code too much? Also, maybe storing the full block would increase register pressure and hurt speeds for the common case in that way. So it's a questionable additional optimization. Just thinking out loud here.

[...]

PR #4813, opened 2021-09-19
Zip optimizations

magnumripper:
See #4031

### DIFF
--- a/doc/NEWS
+++ b/doc/NEWS
@@ -228,6 +228,9 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
 - Added sense2john.py: pfSense/OPNsense config.xml hash extract script.
   [Private Wolf; 2021]
 
+- The Zip (WinZip) formats got several optimizations and the OpenCL version now
+  handles large data sizes much better.  [magnum; 2021]
+
 
 Major changes from 1.8.0-jumbo-1 (December 2014) to 1.9.0-jumbo-1 (May 2019):
 
--- a/run/opencl/opencl_hmac_md5.h
+++ b/run/opencl/opencl_hmac_md5.h
@@ -70,15 +70,21 @@ inline void hmac_md5(HMAC_KEY_TYPE void *_key, uint key_len,
 	MD5_Init(&ctx);
 	MD5_Update(&ctx, buf, 64);
 #ifdef USE_DATA_BUF
-	while (data_len) {
-		uchar pbuf[64];
-		uint len = MIN(data_len, (uint)sizeof(pbuf));
+	HMAC_MSG_TYPE uint *data32 = (HMAC_MSG_TYPE uint*)_data;
+	uint blocks = data_len / 64;
+	data_len -= 64 * blocks;
+	data += 64 * blocks;
+	ctx.total += 64 * blocks;
+	while (blocks--) {
+		uint W[16];
 
-		memcpy_macro(pbuf, data, len);
-		MD5_Update(&ctx, pbuf, len);
-		data_len -= len;
-		data += len;
+		memcpy_macro(W, data32, 16);
+		md5_block(uint, W, ctx.state);
+		data32 += 16;
 	}
+	uchar pbuf[64];
+	memcpy_macro(pbuf, data, data_len);
+	MD5_Update(&ctx, pbuf, data_len);
 #else
 	MD5_Update(&ctx, data, data_len);
 #endif
--- a/run/opencl/opencl_hmac_sha1.h
+++ b/run/opencl/opencl_hmac_sha1.h
@@ -71,15 +71,21 @@ inline void hmac_sha1(HMAC_KEY_TYPE void *_key, uint key_len,
 	SHA1_Init(&ctx);
 	SHA1_Update(&ctx, buf, 64);
 #ifdef USE_DATA_BUF
-	while (data_len) {
-		uchar pbuf[64];
-		uint len = MIN(data_len, (uint)sizeof(pbuf));
-
-		memcpy_macro(pbuf, data, len);
-		SHA1_Update(&ctx, pbuf, len);
-		data_len -= len;
-		data += len;
+	HMAC_MSG_TYPE uint *data32 = (HMAC_MSG_TYPE uint*)_data;
+	uint blocks = data_len / 64;
+	data_len -= 64 * blocks;
+	data += 64 * blocks;
+	ctx.total += 64 * blocks;
+	while (blocks--) {
+		uint W[16];
+		for (i = 0; i < 16; i++)
+			W[i] = SWAP32(data32[i]);
+		sha1_block(uint, W, ctx.state);
+		data32 += 16;
 	}
+	uchar pbuf[64];
+	memcpy_macro(pbuf, data, data_len);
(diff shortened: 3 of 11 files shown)


magnumripper:
What the heck does this even mean?
```
zip_fmt_plug.c: In function 'crypt_all':
zip_fmt_plug.c:250:61: error: 'key_len' is predetermined 'shared' for 'shared'
 #pragma omp parallel for default(none) private(index) shared(count, saved_key, saved_salt, crypt_key, key_len, early_skip, late_skip, late_size)
                                                             ^
zip_fmt_plug.c:250:61: error: 'early_skip' is predetermined 'shared' for 'shared'
zip_fmt_plug.c:250:61: error: 'late_skip' is predetermined 'shared' for 'shared'
zip_fmt_plug.c:250:61: error: 'late_size' is predetermined 'shared' for 'shared'
make[1]: *** [Makefile:1604: zip_fmt_plug.o] Error 1
```
Seen on CircleCI mingw

magnumripper:
> What the heck does this even mean?

I ended up just saying `#pragma omp parallel for` like I use to do

solardiz:
> zip_fmt_plug.c:250:61: error: 'key_len' is predetermined 'shared' for 'shared'

I think that's because `const` implies it can be safely shared, so you don't (and can't?) declare that explicitly. I don't know if this implied-shared for `const` is guaranteed by OpenMP spec or not (and starting with which version). We may want to try excluding these specific constants' names from `shared` (keeping the rest of the directive), or (more reliably) dropping `const`. I think I used one of these solutions in some place in JtR last year or so, but unfortunately I don't recall which of the two approaches I chose and exactly where - you might want to grep for that in the tree or in my commits from last year (maybe in the `DES_bs*` changes, or maybe in some yescrypt-using formats). %-)

magnumripper:
> > zip_fmt_plug.c:250:61: error: 'key_len' is predetermined 'shared' for 'shared'
> 
> I think that's because `const` implies it can be safely shared, so you don't (and can't?) declare that explicitly. I don't know if this implied-shared for `const` is guaranteed by OpenMP spec or not (and starting with which version). We may want to try excluding these specific constants' names from `shared` (keeping the rest of the directive), or (more reliably) dropping `const`. I think I used one of these solutions in some place in JtR last year or so, but unfortunately I don't recall which of the two approaches I chose and exactly where - you might want to grep for that in the tree or in my commits from last year (maybe in the `DES_bs*` changes, or maybe in some yescrypt-using formats). %-)

I suppose you're right but gcc 10 complains if I do not declare them. I'd rather just say `#pragma omp parallel for` than dropping `const` from stuff. In my head the former is more reliable.

magnumripper:
Note to self: In latest version I could drop this workaround
```diff
diff --git a/run/opencl/zip_kernel.cl b/run/opencl/zip_kernel.cl
index 52185dcd5..b757b659c 100644
--- a/run/opencl/zip_kernel.cl
+++ b/run/opencl/zip_kernel.cl
@@ -186,11 +186,7 @@ inline void pbkdf2_hmac_sha1(const uchar *pass, const uint passlen,
 {
        uint ipad_state[5];
        uint opad_state[5];
-#if gpu_nvidia(DEVICE_INFO)
-       /* This is a driver bug workaround. Problem seen with eg. 465.19.01 */
-       volatile
-#endif
-               uint accum = 0;
+       uint accum = 0;
 
        preproc(pass, passlen, ipad_state, 0x36363636);
        preproc(pass, passlen, opad_state, 0x5c5c5c5c);
```
It now tunes and runs very well on nvidias and under macOS w/ any device I've tried but segfaults(!) with virtually all Intel or POCL CPU drivers and fails with most other AMD drivers I've tried. I suspect there's some bug left to nail (as opposed to driver bugs to work around). ASan doesn't complain about anything.

magnumripper:
> Note to self: In latest version I could drop this workaround

That workaround is back, because I settled for the version that has everything beyond early rejection done in the final kernel. I'm about to merge this soon, despite some CPU drivers failing.

solardiz:
> I'm about to merge this soon, despite some CPU drivers failing.

Would this be a regression - that is, does the current `zip-opencl` work with those CPU drivers? I'm concerned this might indicate a real bug in the new code, which could also show up with some GPUs.

magnumripper:
> Would this be a regression

In a way, but OTOH the format also went from bad to good on AMD Vega under macOS.

> I'm concerned this might indicate a real bug in the new code, which could also show up with some GPUs.

That's what I said too a week ago and I've stared at the code since, I can't find any bug.

Under gdb using pocl, I get this:
```
Device 1: pthread-Intel(R) Core(TM) i7-4790 CPU @ 3.60GHz
Benchmarking: ZIP-opencl, WinZip [PBKDF2-SHA1 OpenCL]... -
Thread 9 "john" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffc3ffe700 (LWP 297185)]
0x00007ffff7faf397 in u_preproc ()
   from ~/.cache/pocl/kcache/HI/FCCCMMKBMBKIEOBNLPLKFBIPNEFHCJLDAHJEA/zip/7-1-1-goffs0-smallgrid/zip.so
(gdb) bt
#0  0x00007ffff7faf397 in u_preproc ()
   from ~/.cache/pocl/kcache/HI/FCCCMMKBMBKIEOBNLPLKFBIPNEFHCJLDAHJEA/zip/7-1-1-goffs0-smallgrid/zip.so
#1  0x00007ffff7fb3fdc in pbkdf2_hmac_sha1 ()
   from ~/.cache/pocl/kcache/HI/FCCCMMKBMBKIEOBNLPLKFBIPNEFHCJLDAHJEA/zip/7-1-1-goffs0-smallgrid/zip.so
#2  0x00007ffff7fb45f8 in _pocl_kernel_zip_workgroup ()
   from ~/.cache/pocl/kcache/HI/FCCCMMKBMBKIEOBNLPLKFBIPNEFHCJLDAHJEA/zip/7-1-1-goffs0-smallgrid/zip.so
#3  0x00007ffff51b59fd in ?? () from /usr/lib/x86_64-linux-gnu/libpocl.so.2.4.0
#4  0x00007ffff782c609 in start_thread (arg=<optimized out>) at pthread_create.c:477
#5  0x00007ffff7753103 in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```
I take it the segfault happens while running the `preproc()` function and that function has been used for years in many other kernels. I have audited code and buffer sizes many times. I can't see anything wrong.

magnumripper:
This is strange: If using `--lws=1`, both POCL and Intel stop segfaulting and pass self-test.

magnumripper:
> This is strange: If using `--lws=1`, both POCL and Intel stop segfaulting and pass self-test.

Actually not just when using 1: If using either of `--lws` or `--gws` options, or both, it runs just fine on both platforms.

ghost:
POCL failure is expected, Intel on the other hand is complicated: no bad bugs are expected from the proprietary Driver.

> If using either of --lws or --gws options, or both, it runs just fine on both platforms.

So, I would say auto-tune is doing something wrong (or at least the format while auto tuning is doing something bad).

magnumripper:
I nailed it - it was my bumping of max. length to 125 that wasn't supported by the kernel. Down to 64 again and problems gone.

PR #4817, opened 2021-09-20
Module to generate password simillar to inc but more random

e2002e:
This module is in beta stage: it includes position frequency as well as character chaining, but only  uses lower set, and seeks forward by 1 for now.
Sets are not customizable but there could be several modes -alnum, num, lowernum etc...-
The character chains should also expand to check the 3 previous placeholders.
I use this pull request to draw your attention and hope you'll test it.
Regards, Yvain.


ghost:
CI is failing because you are using very, very, very old "base" JtR source code. 

In fact, you can't create a PR containing hundreds of commits. No one will review it, it feels wrong for that reason.
If a PR needs more than one commit, a few should suffice. 

That said, as a general rule of thumb, if you're proposing something new (thanks for that), I would really recommend discussing it first with the people who will be reviewing your code and with the community.

magnumripper:
I agree with Claudio. 235 commits in one PR? Learn git before you proceed.

PR #4819, opened 2021-09-21
position frequencies and character chaining

e2002e:
Hi, this is a brand new fork to avoid the long commit list.
I've worked about 3 months on this module: starting from a positional frequency, splitting the sets to increase rotation, then working on the chaining of character without splicing, so that when I found the way of making the position and chain technique work together I re implemented the splicing -which might need modification for a faster rotation-
The module is a Beta, I will work more on it if you are interested, otherwise I'll leave it like it is.
use the option -inc2 (yes I can't find another name).
Thanks.

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -115,6 +115,8 @@ static struct opt_entry opt_list[] = {
 	{"subsets-required", FLG_ONCE, 0, FLG_SUBSETS_CHK, OPT_REQ_PARAM, "%u", &options.subset_must},
 	{"subsets-min-diff", FLG_ONCE, 0, FLG_SUBSETS_CHK, OPT_REQ_PARAM, "%u", &options.subset_min_diff},
 	{"subsets-max-diff", FLG_ONCE, 0, FLG_SUBSETS_CHK, OPT_REQ_PARAM, "%u", &options.subset_max_diff},
+	{"inc2", FLG_INC2_SET, FLG_CRACKING_CHK,
+		0, 0, OPT_FMT_STR_ALLOC},
 	{"subsets-prefer-short", FLG_SUBSETS_SHORT, FLG_SUBSETS_SHORT, FLG_SUBSETS_CHK, FLG_SUBSETS_SMALL},
 	{"subsets-prefer-small", FLG_SUBSETS_SMALL, FLG_SUBSETS_SMALL, FLG_SUBSETS_CHK, FLG_SUBSETS_SHORT},
 	{"no-mask", FLG_NO_MASK_BENCH, FLG_NO_MASK_BENCH, FLG_TEST_CHK, FLG_MASK_CHK},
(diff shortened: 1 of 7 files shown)


ghost:
This PR does not contain all the necessary files. The build fails due to missing files.

e2002e:
Yo I added the inc2.c in2.h files, thanks for telling me.

src/options.c
```diff
@@ -294,10 +296,20 @@ PRINCE_USAGE \
 JOHN_USAGE_REGEX \
 "--subsets[=CHARSET]        \"Subsets\" mode (see doc/SUBSETS)\n" \
 "--subsets-required=N       The N first characters of \"subsets\" charset are\n" \
-"                           the \"required set\"\n" \
 "--subsets-min-diff=N       Minimum unique characters in subset\n" \
 "--subsets-max-diff=[-]N    Maximum unique characters in subset (negative N is\n" \
 "                           relative to word length)\n" \
+"--inc2						\"inc2\" generation mode\n" \
+"--stdout[=LENGTH]          just output candidate passwords [cut at LENGTH]\n" \
```
ghost, line 303: This is wrong, you are duplicating lines here.

src/options.c
```diff
@@ -294,10 +296,20 @@ PRINCE_USAGE \
 JOHN_USAGE_REGEX \
 "--subsets[=CHARSET]        \"Subsets\" mode (see doc/SUBSETS)\n" \
 "--subsets-required=N       The N first characters of \"subsets\" charset are\n" \
-"                           the \"required set\"\n" \
 "--subsets-min-diff=N       Minimum unique characters in subset\n" \
 "--subsets-max-diff=[-]N    Maximum unique characters in subset (negative N is\n" \
 "                           relative to word length)\n" \
+"--inc2						\"inc2\" generation mode\n" \
```
ghost, line 302: `inc2` means nothing. We told you this:
* You must introduce your proposal with some use for it;
* Maybe using a published paper.

ghost:
Thanks, we realized that.

ghost (changes requested):
It is my opinion and not the final answer from Solar, but the PR seems useless:
- I have to read the code or test it myself to understand what it is doing;
- Then, I have to compare to other modes because there is no information about how the PR is an improvement.

Also, there are many hard-coded strings. Is the PR an improvement, how, when, for whom?

src/options.c
e2002e: What it does :
1 Uses three lists of strings that are using position frequencies, the first is the normal one (no chain), the second is the chained characters, and the third are the complements of these.
The switch that you see inside the iteration on the word and the advance of the state maps to that.

2 Splits hardcoded strings by chuncks of 4, the order inside these chuncks being random.

3 Checks for the previous character on the fly, if it’s a special one (that can form a pattern like a « t » would often be followed by « h »), we switch set and get about 2 to 5 characters to proceed from the list « chainFreq », when they are done we pass on to the complement which come from the « counterChainFreq » and has been randomized inside « chrsts2 », when these are done we put the variable inc[pos-1] at zero, as if the placeholder was not following a special char (which we don’t know yet and will check back in word generation).

Why would we use it :
I don’t know it looks right on stdout, maybe the chunks should be of length 2 to increase rotation, or even hold just one char, but the smallest the chuncks get, the more there are of them, and it becomes pointless since chuncks index are proceeded from the end to the start like the states, and having chuncks of 1 for instance would cancel the splitting... Maybe a size of 3 is good.
The randomness is limited for each placeholder but them being put together then it gets very random, while still respecting frequency by position and character chains.

If you test it against other modes please limit them to lower since I have not implemented numbers nor alpha (you can see in the code that I did prepare some of the strings for it though).

solardiz:
@e2002e I'm happy that you're having fun with these experiments, but like before this doesn't look like something we'd want merged.

Personally, I intend to find and spend time on improving incremental mode in certain, um, incremental ways - and only merge that if I have convincingly better results.

PR #4822, opened 2021-09-22
configure.ac: avoid bashism in pkg-config check

thesamesam:
== is a Bashism and configure scripts need to work
with a POSIX shell, so let's change to = (like in
the other condition).

Signed-off-by: Sam James <sam@gentoo.org>


ghost (approved):
Thanks. You are welcome to continue working on JtR configure "tools".

thesamesam:
@claudioandre-br Thanks a lot, it's really nice to hear that. I'm often nervous to contribute build system changes to projects because sometimes people want them calcified, but that's a big relief and helps downstream folks a lot. Cheers, and hope to submit something in future!

Tiny follow up in #4826.

PR #4823, opened 2021-09-22
Freeze the clock during initial format reset()

magnumripper:
A run-time OpenCL kernel build can take a long time, and autotune may add over 10 seconds.
This change make things like ETA and speed figures appear stable much earlier.

### DIFF
--- a/src/john.c
+++ b/src/john.c
@@ -1655,7 +1655,6 @@ static void john_init(char *name, int argc, char **argv)
 static void john_run(void)
 {
 	struct stat trigger_stat;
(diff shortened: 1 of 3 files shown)


src/john.c
```diff
@@ -1780,8 +1778,14 @@ static void john_run(void)
 
 		omp_autotune_run(&database);
 
-		if (trigger_reset)
```
magnumripper, line 1783: I'm pretty sure this `trigger_reset` thing are from back in the days where `--stdout`'s `dummy_format->methods.reset` was NULL so would segfault.

PR #4826, opened 2021-09-22
configure: regenerate for pkg-config bashism fix

thesamesam:
Quick follow up to the other PR.

See: https://github.com/openwall/john/pull/4822
Signed-off-by: Sam James <sam@gentoo.org>


ghost (approved):
Thanks

PR #4829, opened 2021-09-23
OpenCL: Fix a bug in an unused code path in shared HMAC

magnumripper:
Needs no review but I will await CI before merging, just in case


ghost:
Unfortunately, CI has almost no OpenCL tests.

PR #4832, opened 2021-09-25
Fixed negative exit codes

exploide:
Exit codes need to be between 0 and 255 (https://tldp.org/LDP/abs/html/exitcodes.html).

This PR fixes negative exit codes in the python scripts.


ghost (approved):
Thanks. There are unrelated changes, but I think they are Ok.

Let's hear Magnun's voice.

exploide:
Yeah, looks like my editor removed trailing whitespace in one of the files. Can amend if you find it annoying.

Besides I removed a useless out-commented line right before one of the exits.

magnumripper (approved):
Looks fine to me, and the removal of trailing spaces is a Good Thing[tm]

PR #4835, opened 2021-09-26
Improved openssl2john and pem2john

### LINKED ISSUE #4833: openssl2john: Redirect people to pem2john when input looks like PEM
As seen in https://www.openwall.com/lists/john-users/2021/09/26/1 some people are having a hard time figuring out the `*2john` utility to use on OpenSSL-generated private key files. Maybe `openssl2john` is a more obvious program name they'd try first, not even thinking of PEM. We could have `openssl2john` print a helpful message in that case.

exploide:
I did a bit of cleanup in both scripts. More importantly:

- `openssl2john.py` detects private key files and recommends `pem2john.py`
- `pem2john.py` now detects unencrypted private key files

fixes #4833

### DIFF
--- a/run/openssl2john.py
+++ b/run/openssl2john.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 
 # This utility helps in cracking files encrypted using "openssl enc" command.
+# It does not help in cracking encrypted private key files. Use pem2john.py instead.
 #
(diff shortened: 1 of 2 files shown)


solardiz (approved):
LGTM. I assume you tested this with both Python 2 and 3? Thank you!

PR #4838, opened 2021-10-05
Allow hashcat's '+N' rule (increment character) in more situations

magnumripper:
We already allowed the '+N' rule without explicitly being in "hashcat mode" as long as we weren't processing word pairs.

This change allows '+N' at any time where it wouldn't be valid as the original '+' concatenate word pair command:  If neither of commands "1" or "2" was used before it, the '+' is parsed as the hashcat command.

### DIFF
--- a/doc/RULES
+++ b/doc/RULES
@@ -296,7 +296,7 @@ first, the second, or to both words:
 
 1	first word only
 2	second word only
-+	the concatenation of both (should only be used after a "1" or "2")
++	the concatenation of both (can only be used after a "1" or "2")
 
 If you use some of the above commands in a rule, it will only process
 word pairs (e.g., full names from the GECOS field) and reject single
--- a/doc/RULES-hashcat
+++ b/doc/RULES-hashcat
@@ -27,7 +27,7 @@ one simply creates include sections, like this:
 
 [List.Rules:HC_deadone]
 !! hashcat logic ON
-.include '/code/hashcat//rules/d3ad0ne.rule'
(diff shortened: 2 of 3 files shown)


magnumripper:
I need to test this thoroughly for regression. Edit: Re-wrote the changes so they're much smaller and should be safe.

Here's a crude ruleset that triggered me to make this change:
```
[List.Rules:IncTail]
)?d +m )?d
)9 vam1 =a?d +a =a?d om0
)9 vam1 vbm2 =b?d =a9 +b =b?d oa0 om0
)9 vam1 vbm2 vcm3 =c?d =b9 =a9 +c =c?d ob0 oa0 om0
```
Since it uses character classes, I can't use "hashcat mode" but if I don't, the ruleset would fail and bail in single mode.

solardiz (approved):
I like this in principle, but I think the dropping of curly braces unnecessarily computes `strlen(in)` when treating the command as hashcat's. Maybe instead of using `else`, add a `break` to the `if` block. Then the nested `switch` and what follows it would only be reached when needed, and the indentation would also not suffer.

magnumripper:
> I like this in principle, but I think the dropping of curly braces unnecessarily computes `strlen(in)` when treating the command as hashcat's. Maybe instead of using `else`, add a `break` to the `if` block. Then the nested `switch` and what follows it would only be reached when needed, and the indentation would also not suffer.

Good point, fixed now

solardiz (approved):

PR #4839, opened 2021-10-05
Edge case bugfixes for two OpenCL formats

### LINKED ISSUE #4816: Possible problem in a couple of internal-mask formats
In some formats (the below is oldoffice-opencl), we have this at the end of `set_key()`:
```c
	/* Early partial transfer to GPU */
	if (index && !(index & (256*1024 - 1))) {
		HANDLE_CLERROR(clEnqueueWriteBuffer(queue[gpu_id], cl_saved_key, CL_FALSE, key_offset, key_idx - key_offset, saved_key + key_offset, 0, NULL, NULL), "Failed transferring keys");
		key_offset = key_idx;
		HANDLE_CLERROR(clEnqueueWriteBuffer(queue[gpu_id], cl_saved_idx, CL_FALSE, idx_offset, 4 * index - idx_offset, saved_idx + (idx_offset / 4), 0, NULL, NULL), "Failed transferring index");
		idx_offset = 4 * index;
		HANDLE_CLERROR(clFlush(queue[gpu_id]), "failed in clFlush");
		new_keys = 0;
	}
```

Then we have this in `crypt_all()`. Note the `if (!mask_gpu_is_static)` thing there:
```c
	if (new_keys || ocl_autotune_running) {
		/* Self-test kludge */
		if (idx_offset > 4 * (gws + 1))
			idx_offset = 0;

		BENCH_CLERROR(clEnqueueWriteBuffer(queue[gpu_id], cl_saved_key, CL_FALSE, key_offset, key_idx - key_offset, saved_key + key_offset, 0, NULL, multi_profilingEvent[0]), "Failed transferring keys");
		BENCH_CLERROR(clEnqueueWriteBuffer(queue[gpu_id], cl_saved_idx, CL_FALSE, idx_offset, 4 * (gws + 1) - idx_offset, saved_idx + (idx_offset / 4), 0, NULL, multi_profilingEvent[1]), "Failed transferring index");
		if (!mask_gpu_is_static)
			BENCH_CLERROR(clEnqueueWriteBuffer(queue[gpu_id], buffer_int_key_loc, CL_FALSE, 0, 4 * gws, saved_int_key_loc, 0, NULL, NULL), "failed transferring buffer_int_key_loc.");

		new_keys = 0;
	}
```
I think this means we might theoretically miss transfering `saved_int_key_loc` in rare cases: If last call to `set_key()` was with index 262144 (or a multiple of it), a partial transfer sets off and `new_keys` becomes zero. If that was actually the very last key, `crypt_all()` will not transfer `saved_int_key_loc`.

However, if we did call `set_key(262144)` it's not likely it was the last key in a batch (KPC would need to be 262145, or 0x40001, which is very odd). BUT it could happen in the very end of a session (such as wordlist end) or possibly when running single mode but that can't be used with internal mask so shouldn't be a problem. Anyway I think we need to fix this.

As far as I have found so far, only oldoffice-opencl, cryptoSafe-opencl (brand new) ~and the PR for revised zip-opencl~ are affected by this problem - but IIRC more formats have had this "early partial transfer" thing at least back in the days.

magnumripper:
oldoffice-opencl and cryptosafe-opencl were affected.

We could theoretically miss transferring saved_int_key_loc in cases where last key index was a multiple that triggered early transfer.  There was also a fence post error affecting key indeces in the same situation.

Closes #4816

### DIFF
--- a/src/opencl_cryptosafe_fmt_plug.c
+++ b/src/opencl_cryptosafe_fmt_plug.c
@@ -272,10 +272,15 @@ static void set_key(char *key, int index)
 	/* Early partial transfer to GPU */
 	if (index && !(index & (256*1024 - 1))) {
 		HANDLE_CLERROR(clEnqueueWriteBuffer(queue[gpu_id], cl_saved_key, CL_FALSE, key_offset, key_idx - key_offset, saved_key + key_offset, 0, NULL, NULL), "Failed transferring keys");
-		key_offset = key_idx;
-		HANDLE_CLERROR(clEnqueueWriteBuffer(queue[gpu_id], cl_saved_idx, CL_FALSE, idx_offset, 4 * index - idx_offset, saved_idx + (idx_offset / 4), 0, NULL, NULL), "Failed transferring index");
-		idx_offset = 4 * index;
+		HANDLE_CLERROR(clEnqueueWriteBuffer(queue[gpu_id], cl_saved_idx, CL_FALSE, idx_offset, 4 * (index + 2) - idx_offset, saved_idx + (idx_offset / 4), 0, NULL, NULL), "Failed transferring index");
+
+		if (!mask_gpu_is_static)
+			HANDLE_CLERROR(clEnqueueWriteBuffer(queue[gpu_id], buffer_int_key_loc, CL_FALSE, idx_offset, 4 * (index + 1) - idx_offset, saved_int_key_loc + (idx_offset / 4), 0, NULL, NULL), "failed transferring buffer_int_key_loc.");
+
 		HANDLE_CLERROR(clFlush(queue[gpu_id]), "failed in clFlush");
+
+		key_offset = key_idx;
+		idx_offset = 4 * (index + 1);
 		new_keys = 0;
 	}
 }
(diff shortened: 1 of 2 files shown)


magnumripper:
Now passes Test Suite while forcing the edge case (glad I did).

solardiz:
> Now passes Test Suite while forcing the edge case (glad I did).

Will this "forcing the edge case" be part of our standard regression testing? Perhaps you need to commit this change into the test suite?

magnumripper:
I hacked the formats so there were "an edge case" regardless of KPC, so it's hard to do in TS itself.

solardiz:
> I hacked the formats so there were "an edge case" regardless of KPC

Maybe keep this as an `#if 0`'ed piece inside the formats, with comments on why it exists and how to use it for testing?

magnumripper:
Nah, it's not worth #ifdef'ing. I just did this
```diff
diff --git a/src/opencl_oldoffice_fmt_plug.c b/src/opencl_oldoffice_fmt_plug.c
index 85ef1bd8f..94390f9b7 100644
--- a/src/opencl_oldoffice_fmt_plug.c
+++ b/src/opencl_oldoffice_fmt_plug.c
@@ -616,7 +616,7 @@ static void set_key(char *key, int index)
        new_keys = 1;
 
        /* Early partial transfer to GPU */
-       if (index && !(index & (256*1024 - 1))) {
+       if (1) {
                HANDLE_CLERROR(clEnqueueWriteBuffer(queue[gpu_id], cl_saved_key, CL_FALSE, key_offset, key_idx - key_offset, saved_key + key_offset, 0, NULL, NULL), "Failed transferring keys");
                HANDLE_CLERROR(clEnqueueWriteBuffer(queue[gpu_id], cl_saved_idx, CL_FALSE, idx_offset, 4 * (index + 2) - idx_offset, saved_idx + (idx_offset / 4), 0, NULL, NULL), "Failed transferring index");

```

PR #4841, opened 2021-10-08
PBKDF2-HMAC-SHA1-opencl: Fix big-endian build

skriesch:
We are receiving following error message for building for s390x:
```
error: 'i' undeclared (first use in this function)
   226 |         for (i = 0; i < len / sizeof(uint32_t); ++i) {
                          ^

```
Firstly - before the enablement for s390x - the syntax issue should have been fixed.
Therefore I have created the declaration of an integar of I inside of the for loop.

### DIFF
--- a/src/opencl_pbkdf2_hmac_sha1_fmt_plug.c
+++ b/src/opencl_pbkdf2_hmac_sha1_fmt_plug.c
@@ -227,6 +227,7 @@ static void *get_binary(char *ciphertext)
 #if !ARCH_LITTLE_ENDIAN
 	char *p = strrchr(ciphertext, '$') + 1;
 	int len = strlen(p) / 2;
+	int i;
 	for (i = 0; i < len / sizeof(uint32_t); ++i) {
 		((uint32_t*)out)[i] = JOHNSWAP(((uint32_t*)out)[i]);
 	}


solardiz:
@skriesch Thank you! I think the only big-endian builds we make ourselves are without OpenCL, which is why this wasn't detected sooner. In fact, is this really the only issue with such builds? Do you otherwise have working OpenCL in JtR on s390x? Also, if you don't mind telling, what OpenCL devices do you happen to have in a box like that?

As to this specific pull request, can you please change the commit message e.g. to "PBKDF2-HMAC-SHA1-opencl: Fix build on big-endian" (amend the one commit and force-push). I think this is more descriptive.

Incidentally, while declaring the loop variable inside `for` is a standard C99 thing, it isn't in this tree yet, with the only such cases existing in `pp.c` (per my grep for `for (int`). However, perhaps we should start using this throughout the tree, so I don't mind.

src/opencl_pbkdf2_hmac_sha1_fmt_plug.c
```diff
@@ -227,7 +227,7 @@ static void *get_binary(char *ciphertext)
 #if !ARCH_LITTLE_ENDIAN
 	char *p = strrchr(ciphertext, '$') + 1;
 	int len = strlen(p) / 2;
-	for (i = 0; i < len / sizeof(uint32_t); ++i) {
+  for (int i = 0; i < len / sizeof(uint32_t); ++i) {
 		((uint32_t*)out)[i] = JOHNSWAP(((uint32_t*)out)[i]);
 	}
```
solardiz, line 233: BTW, ideally instead of the typecasts that could break aliasing rules, this code should be using a `union` like `pbkdf2_hmac_sha1_binary` does. However, fixing this probably shouldn't be part of this same PR.

magnumripper:
> Incidentally, while declaring the loop variable inside `for` is a standard C99 thing, it isn't in this tree yet, with the only such cases existing in `pp.c` (per my grep for `for (int`). However, perhaps we should start using this throughout the tree, so I don't mind.

We do build pp.o (and only that one) specifically with `-std=c99`. I can't recall the exact reason, but it might be things like this. If we're to allow this in more files, we might need to use that option globally in order not to break build on some compilers?

magnumripper:
> @skriesch Thank you! I think the only big-endian builds we make ourselves are without OpenCL, which is why this wasn't detected sooner. In fact, is this really the only issue with such builds? Do you otherwise have working OpenCL in JtR on s390x? Also, if you don't mind telling, what OpenCL devices do you happen to have in a box like that?

Aye, this is interesting. Since we've never had a chance to test things like this I would expect loads of problems with a BE host running a LE device (if that combo even exist IRL) and probably many with a BE host and BE device as well.

The output of `./john --list=opencl-devices` would be valuable information, once it builds at all.

solardiz:
> We do build pp.o (and only that one) specifically with `-std=c99`. I can't recall the exact reason, but it might be things like this. If we're to allow this in more files, we might need to use that option globally in order not to break build on some compilers?

Oh, indeed. Modern gcc is C99 by default, but older versions were not (needed that option, once it was introduced). Since we don't yet enable this globally, @skriesch please avoid introducing the C99 dependency in this PR - that is, declare the variable just before the loop, not inside `for`.

As to OpenCL on s390x, my current best guess is it just happened to be auto-enabled by our `./configure` due to installed OpenCL frontend (headers, library?) on a certain openSUSE build box. There may or may not also be a working OpenCL backend on that system. Anyway, let's hear from @skriesch on this as well.

magnumripper:
Please also fix the current whitespace violation: We indent with tabs.

solardiz:
@skriesch I really would prefer that we go with "PBKDF2-HMAC-SHA1-opencl: Fix build on big-endian" for the commit message (describe primarily what the commit achieves, not its content), and please avoid changing the indentation of the `for` line. Thanks!

solardiz:
In case you're wondering, "PBKDF2-HMAC-SHA1-opencl" is the "format label" (it's defined earlier in the same file). We tend to prefix commit messages with "format label: ", so this would be consistent. Using the filename would also be OK.

solardiz:
Oh, also instead of adding commits, it'd be preferable if you force-push. I understand it might not be convenient for you to amend and force-push commits if you use the GitHub web UI, though, do you? In the latter case, we may "squash and merge" using GitHub UI as well... just let us know.

skriesch:
I have contributed this fix because of a syntax error, that i has to be declared.
I have tested it in our oppenSUSE build environment.

The new failure for PowerPC and s390x is the following:
```
[ 2131s] /usr/lib64/gcc/s390x-suse-linux/11/../../../../s390x-suse-linux/bin/ld: opencl_7z_fmt_plug.o: in function `sevenzip_decrypt':
[ 2131s] /home/abuild/rpmbuild/BUILD/john-1.9.0/src/opencl_7z_fmt_plug.c:683: undefined reference to `JOHNSWAP'
```

But that should not be part of this Pull Request, because this PR is because of general syntax issues for Big Endian.

solardiz:
@skriesch I'm really sorry that we're taking up so much of your time (and our time too) on this trivial fix, especially given the overall low quality of code in jumbo, but the `for` line in your latest commit is still misindented, and now the commit message is over-long (we prefer to keep the first/title/only line to 72 chars, subsequent lines wrapped at 75, as these work well with both GitHub and `git log` on an 80-wide terminal window).

> ```
> opencl_7z_fmt_plug.c:683: undefined reference to `JOHNSWAP'
> ```
> But that should not be part of this Pull Request, because this PR is because of general syntax issues for Big Endian.

Actually, I wouldn't mind all of these being in the same PR or even same commit (with commit message of e.g. "OpenCL formats big-endian build fixes", which captures what these achieve at the high level), however I am not sure whether these two are the only ones - perhaps you haven't yet identified all, or have you? Also, with us already spending too much time getting this trivial change in, I am concerned that working on many trivial fixes at once in this way would become unmanageable.

ghost:
Below reported message is from jumbo 1.

> `john-1.9.0/src/opencl_7z_fmt_plug.c:683`


This patch is a suitable fix to build on S390:
[0001-S390.txt](https://github.com/openwall/john/files/7331088/0001-S390.txt)

```
Target CPU ................................. s390x, 64-bit BE
Legacy arch header ......................... autoconf_arch.h

Optional libraries/features found:
Memory map (share/page large files) ........ yes
Fork support ............................... yes
OpenMP support ............................. yes (not for fast formats)
OpenCL support ............................. yes
Generic crypt(3) format .................... yes
OpenSSL (many additional formats) .......... yes
libgmp (PRINCE mode and faster SRP formats)  yes
128-bit integer (faster PRINCE mode) ....... yes
libz (pkzip and some other formats) ........ yes
libbz2 (gpg2john extra decompression logic)  yes
libpcap (vncpcap2john and SIPdump) ......... yes
Non-free unrar code (complete RAR support) . yes

Configure finished.  Now "make -s clean && make -sj4" to compile.
ar: creating poly1305-donna.a
ar: creating aes.a
ar: creating secp256k1.a
ar: creating ed25519-donna.a
opencl_rar_fmt_plug.c:441:2: warning: #warning ": target system requires aligned memory access, RAR OpenCL format disabled:" [-Wcpp]
 #warning ": target system requires aligned memory access, RAR OpenCL format disabled:"
  ^~~~~~~

Make process completed.
```

To spare resources, the build is upstream plus:
```
for file in $(ls *_fmt_plug.c); do
    if [[ $file != "opencl_"* && $file != *"common"* && $file != "sha256crypt_fmt_plug.c" && $file != "sha512crypt_fmt_plug.c" ]]; then
        rm "$file"
    fi
done
```

I'm going to run a `--test=0`, but I'm not sure if there is any ICD available and we believe it will fail.

solardiz:
Thanks, @claudioandre-br! So the fix in this PR is the only one needed to build our current OpenCL host code for big-endian.

@skriesch Please amend and force-push the one commit here one final time to avoid breaking the indentation of the `for` line and to have the commit message as I had suggested (or similar, fitting in 72 chars). Thank you!

For the other issue you mentioned, per Claudio's analysis it appears that this is something we had already fixed, so I guess you'll need either a patch specific to your package or you need to package our latest code from here rather than the older release.

ghost:
As expected:
```
Created directory: /root/.john
Created directory: /root/.john/opencl
Will run 4 OpenMP threads
Testing: descrypt, traditional crypt(3) [DES 64/64]... (4xOMP) PASS
Testing: bsdicrypt, BSDI crypt(3) ("_J9..", 725 iterations) [DES 64/64]... (4xOMP) PASS
Testing: md5crypt, crypt(3) $1$ (and variants) [MD5 32/64 X2]... (4xOMP) PASS
Testing: md5crypt-long, crypt(3) $1$ (and variants) [MD5 32/64]... (4xOMP) PASS
Testing: bcrypt ("$2a$05", 32 iterations) [Blowfish 32/64 X2]... (4xOMP) PASS
Testing: scrypt (16384, 8, 1) [Salsa20/8 32/64]... (4xOMP) PASS
Testing: LM [DES 64/64]... (4xOMP) PASS
Testing: AFS, Kerberos AFS [DES 48/64 4K]... PASS
Testing: tripcode [DES 64/64]... (4xOMP) PASS
Testing: sha256crypt, crypt(3) $5$ (rounds=5000) [SHA256 32/64]... (4xOMP) PASS
Testing: sha512crypt, crypt(3) $6$ (rounds=5000) [SHA512 64/64]... (4xOMP) PASS
No OpenCL devices found
```

Faster approach:
- an  admin should get my patch, edit the author and message, git am and push it to the bleeding-jumbo branch.

skriesch:
We are using the version 1.9.0. I will test the patch.

solardiz:
@skriesch For your package, why don't you simply `./configure --disable-opencl` when building for big-endian targets? While it can be made to build with only minor fixes, this combination is untested anyway, so probably not something you want to offer in a distro.

solardiz:
I've just cloned, amended, and force-pushed @skriesch's commit to her fork of the repo while authenticating to GitHub as me - apparently, the "allow maintainer edits" (or what it was called) GitHub checkbox allows for that (although the GitHub web UI would only allow adding commits, as far as I could find). This is to avoid breaking the indentation and having the over-long commit title while giving @skriesch due credit for the actual fix. I'll merge this now.

skriesch:
This additional patch has resolved the build problem for s390x:

```
--- src/opencl_7z_fmt_plug_copy.c	2021-10-17 18:24:14.112418452 +0200
+++ src/opencl_7z_fmt_plug.c	2021-10-17 18:25:35.984598919 +0200
@@ -40,6 +40,8 @@
 #include "config.h"
 #include "lzma/LzmaDec.h"
 #include "lzma/Lzma2Dec.h"
+#include "johnswap.h"
+
 
 #define FORMAT_LABEL		"7z-opencl"
 #define FORMAT_NAME		"7-Zip"
```

I will create an additional Pull Request for that.

solardiz:
@skriesch As I understand, no other fix is needed in this repo. According to @claudioandre-br's comments above, our current OpenCL host code builds fine for big-endian with only this PR's fix. That you needed the additional 7z fix is an artifact of you using an older version. The usage of `JOHNSWAP` that was in `opencl_7z_fmt_plug.c` got moved into `7z_common_plug.c` in 91bc30785942108257ded0e5bd51943f5d3d7420, and that other file does `#include "johnswap.h"`, so it's not surprising that issue is gone.

PR #4854, opened 2021-10-27
Replace all calls to DES_set_key() with DES_set_key_unchecked

### LINKED ISSUE #4831: 15 DES-using formats fail self-test with OpenSSL 3.0.0
:1st_place_medal::thumbsup:

Self test fails for opeenssh private key:
```console
➜  run git:(bleeding-jumbo) ✗ ssh-keygen                           
Generating public/private rsa key pair.
Enter file in which to save the key (/home/alexandre/.ssh/id_rsa): test
Enter passphrase (empty for no passphrase): 
Enter same passphrase again: 
Your identification has been saved in test
Your public key has been saved in test.pub
The key fingerprint is:
SHA256:eo9jvwao/+efJ7BIRvDwsWw4bJgyKhW/H+4H8lwl4s0 alexandre@DENGROUS
The key's randomart image is:
+---[RSA 3072]----+
|                 |
|  .   o .        |
|   o + B o       |
|  + + * O .      |
| o o + OSo       |
|o   o =.E .      |
|.    B.*.o o     |
|    . =.=o+ ...  |
|     oo+o*=ooo   |
+----[SHA256]-----+
➜  run git:(bleeding-jumbo) ✗ ssh-keygen -f test -l                
3072 SHA256:eo9jvwao/+efJ7BIRvDwsWw4bJgyKhW/H+4H8lwl4s0 alexandre@DENGROUS (RSA)
➜  run git:(bleeding-jumbo) ✗ ./ssh2john.py test > test.hash
➜  run git:(bleeding-jumbo) ✗ ./john test.hash --mask='?d?d?d'     
[ssh-opencl] cipher value of 6 is not yet supported with OpenCL!
Loaded 1 password hash (SSH, SSH private key [RSA/DSA/EC/OPENSSH 32/64])
Cost 1 (KDF/cipher [0=MD5/AES 1=MD5/3DES 2=Bcrypt/AES]) is 2 for all loaded hashes
Cost 2 (iteration count) is 16 for all loaded hashes
Will run 32 OpenMP threads
Self test failed (cmp_all(2))
```
The passphrase is `12345`

Output of `./john --list=build-info`:
```console
➜  run git:(bleeding-jumbo) ✗ ./john --list=build-info
Version: 1.9.0-jumbo-1+bleeding-5a19e3d66 2021-09-24 04:35:26 +0200
CPU tests: AVX2
$JOHN is ./
Format interface version: 14
Rec file version: REC4
Charset file version: CHR3
CHARSET_MIN: 1 (0x01)
CHARSET_MAX: 255 (0xff)
CHARSET_LENGTH: 24
SALT_HASH_SIZE: 1048576
SINGLE_IDX_MAX: 2147483648
SINGLE_BUF_MAX: 4294967295
Max. Markov mode level: 400
gcc version: 10.3.0
GNU libc version: 2.31 (loaded: 2.31)
OpenCL headers version: 1.2
Crypto library: OpenSSL
OpenSSL library version: 030000000
OpenSSL 3.0.0-dev xx XXX xxxx
GMP library version: 6.2.0
File locking: fcntl()
fseek(): fseek
ftell(): ftell
fopen(): fopen
memmem(): System's
HR timer: clock_gettime(), latency 40 ns
Total physical host memory: 32089 MiB
Available physical host memory: 27062 MiB
Terminal locale string: en_US.UTF-8
Parsed terminal locale: UTF-8
```
OpenSSH version:
```console
➜  run git:(bleeding-jumbo) ✗ ssh -V
OpenSSH_8.2p1 Ubuntu-4ubuntu0.3, OpenSSL 1.1.1f  31 Mar 2020
```
My CPU is a `AMD Ryzen 9 3950X`
My GPU is a `NVIDIA GeForce GTX 1080 Ti`

Files used: [files.zip](https://github.com/openwall/john/files/7227243/files.zip)

John is built from source on `Ubuntu 20.04` with kernel `5.4.0-86-generic`

Thanks

magnumripper:
This works around a regression in OpenSSL 3 where they changed the default behavior.  Closes #4831

Depending on how the OpenSSL team reacts to https://github.com/openssl/openssl/issues/16859 we might want to merge this - or maybe we want to so regardless?

### DIFF
--- a/src/KRB4_fmt_plug.c
+++ b/src/KRB4_fmt_plug.c
@@ -232,7 +232,7 @@ static int cmp_all(void *binary, int count)
 {
 	DES_cblock tmp;
 
-	DES_set_key(&saved_key.key, &saved_key.sched);
+	DES_set_key_unchecked(&saved_key.key, &saved_key.sched);
 
 	DES_pcbc_encrypt(saved_salt->tgt, (unsigned char *)&tmp,
 	                 sizeof(tmp), &saved_key.sched,
--- a/src/KRB5_std_plug.c
+++ b/src/KRB5_std_plug.c
@@ -184,9 +184,9 @@ inline static void derive_key(const void *constant, int len, krb5_key *krb5key) 
     // set the des schedule
     bk = (DES_cblock*) krb5key->key;
     s = (DES_key_schedule *) krb5key->schedule;
-    DES_set_key(&bk[0], &s[0]);
-    DES_set_key(&bk[1], &s[1]);
-    DES_set_key(&bk[2], &s[2]);
+    DES_set_key_unchecked(&bk[0], &s[0]);
+    DES_set_key_unchecked(&bk[1], &s[1]);
+    DES_set_key_unchecked(&bk[2], &s[2]);
 
     if (DES3_BLOCK_SIZE * 8 < DES3_KEY_BITS || len != DES3_BLOCK_SIZE) {
         nblocks = (DES3_KEY_BITS + DES3_BLOCK_SIZE * 8 - 1) / (DES3_BLOCK_SIZE * 8);
@@ -243,9 +243,9 @@ void des3_decrypt(krb5_key *key, char *cipher, char *plain, int len) {
     k = (DES_cblock *) key->key;
     s = (DES_key_schedule *) key->schedule;
 
-    DES_set_key(&k[0], &s[0]);
-    DES_set_key(&k[1], &s[1]);
-    DES_set_key(&k[2], &s[2]);
+    DES_set_key_unchecked(&k[0], &s[0]);
+    DES_set_key_unchecked(&k[1], &s[1]);
+    DES_set_key_unchecked(&k[2], &s[2]);
 
     DES_ede3_cbc_encrypt((const unsigned char*) cipher, (unsigned char*) plain, len, &s[0], &s[1], &s[2], &ivec, 0);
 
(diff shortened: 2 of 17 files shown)


ghost (approved):
IMO, what the OpenSSL team did makes sense.

solardiz:
How far back does `DES_set_key_unchecked` go? Wouldn't merging this break compatibility with old-yet-still-relevant OpenSSL? (I guess this dates back to older versions of OpenSSL than the oldest we support, probably even to SSLeay, but we need to check before merging.)

> IMO, what the OpenSSL team did makes sense.

"Makes [some] sense" doesn't make it right. I think what they did was wrong.

magnumripper:
> How far back does `DES_set_key_unchecked` go? Wouldn't merging this break compatibility with old-yet-still-relevant OpenSSL? (I guess this dates back to older versions of OpenSSL than the oldest we support, probably even to SSLeay, but we need to check before merging.)

It was added in 0.9.5 back in 1999: https://github.com/openssl/openssl/commit/cddfe788fbbc4726fcf9892963df3f3e823eb233

> > IMO, what the OpenSSL team did makes sense.
> 
> "Makes [some] sense" doesn't make it right. I think what they did was wrong.

Agreed, but OpenSSL 3 is released already as far as I understand.

So I guess we should just merge this. Awaiting @solardiz anyway.

solardiz:
> It was added in 0.9.5 back in 1999: [openssl/openssl@cddfe78](https://github.com/openssl/openssl/commit/cddfe788fbbc4726fcf9892963df3f3e823eb233)

Great. I think this change of ours will even result in slight speedup, as we're now skipping OpenSSL's global flag check.

PR #4855, opened 2021-10-29
Bug fix for --rules-skip-nop

magnumripper:
It would also skip any rules *containing* no-ops (including spaces or tabs), which was never intended.

### DIFF
--- a/src/rules.c
+++ b/src/rules.c
@@ -654,7 +654,7 @@ char *rules_reject(char *rule, int split, char *last, struct db_main *db)
 		return NULL;
 	}
 
-	if (!*rule && options.flags & FLG_RULE_SKIP_NOP)
+	if ((options.flags & FLG_RULE_SKIP_NOP) && !rule[strspn(rule, ": \t")])
 		return NULL;
 
 	while (RULE)
@@ -662,8 +662,6 @@ char *rules_reject(char *rule, int split, char *last, struct db_main *db)
 	case ':':
 	case ' ':
 	case '\t':
-		if (options.flags & FLG_RULE_SKIP_NOP)
-			return NULL;
 		break;
 
 	case '-':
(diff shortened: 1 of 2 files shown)


src/rules.c
```diff
@@ -654,16 +654,14 @@ char *rules_reject(char *rule, int split, char *last, struct db_main *db)
 		return NULL;
 	}
 
-	if (!*rule && options.flags & FLG_RULE_SKIP_NOP)
+	if ((!*rule || !strcmp(rule, ":")) && options.flags & FLG_RULE_SKIP_NOP)
 		return NULL;
```
solardiz, line 658: I'd add braces around `options.flags & FLG_RULE_SKIP_NOP`.

solardiz (approved):
I guess this won't catch trickier no-ops like `::` (two colons) or mixes of colons and whitespace. I don't know if trying to detect those (only no-op commands seen by the loop) is worth the complexity or not. If not, we could want to document the option as imperfect.

magnumripper:
> I guess this won't catch trickier no-ops like `::` (two colons) or mixes of colons and whitespace.

Thanks, it turned out that (while not super important) this was easily implemented.

> we could want to document the option as imperfect.

It also turned out this wasn't documented other than in NEWS so I added it in OPTIONS now.

src/rules.c
```diff
@@ -654,16 +654,14 @@ char *rules_reject(char *rule, int split, char *last, struct db_main *db)
...
 	}
 
-	if (!*rule && options.flags & FLG_RULE_SKIP_NOP)
+	if ((options.flags & FLG_RULE_SKIP_NOP) && (!*rule || strspn(rule, ": \t") == strlen(rule)))
```
magnumripper: Maybe I can drop the `!*rule ||` now...

solardiz (approved):
I think you could even do `!rule[strspn(rule, ": \t")]`, saving a `strlen()` call.

magnumripper:
Cool, I'll try that

solardiz:
I think this is ready to merge now, assuming you've tested it.

PR #4861, opened 2021-11-04
Tezos formats: use PBKDF2 result as secret key directly, avoid memcpy()

solardiz:
This is a minor optimization that makes very little performance difference, but should be made anyway.

### DIFF
--- a/src/opencl_tezos_fmt_plug.c
+++ b/src/opencl_tezos_fmt_plug.c
@@ -272,21 +272,20 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 #pragma omp parallel for
 #endif
 		for (index = 0; index < count; index++) {
-			uint64_t u[8];
-			unsigned char seed[64];
-			unsigned char buffer[20];
+			union {
+				uint64_t u64[4];
+				ed25519_secret_key sk;
+			} u;
 			ed25519_public_key pk;
-			ed25519_secret_key sk;
+			unsigned char buffer[20];
 			int j;
 
-			memcpy(u, host_crack[index].hash, 64);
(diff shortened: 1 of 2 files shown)


solardiz:
Re-tested both formats using our two test vectors and their corresponding passwords inside a larger shuffled wordlist. No issues.

solardiz:
One thing I don't understand is why we have to pass 20 as output size to `blake2b`. We seem to only use 8 bytes afterwards. However, trying to pass less than 20 results in self-test failing.

@kholia Maybe you recall?

solardiz:
> One thing I don't understand is why we have to pass 20 as output size to `blake2b`. We seem to only use 8 bytes afterwards. However, trying to pass less than 20 results in self-test failing.

Figured this out - BLAKE2 includes its parameters, including the requested output size, in the hash computation.

Another observation is that instead of the `memmem` for 8 bytes we could use `memcmp` at offset 2 of the full 20 bytes. However, with only two test vectors I don't dare to make this change just yet.

magnumripper:
> Another observation is that instead of the `memmem` for 8 bytes we could use `memcmp` at offset 2 of the full 20 bytes. However, with only two test vectors I don't dare to make this change just yet.

That's interesting but the good news is that offset 2 is very early. Have you benchmarked such change?

solardiz:
> the good news is that offset 2 is very early.

This doesn't help - it's 2 when we hit the correct password, but for performance we care about the case where the pattern is not found.

> Have you benchmarked such change?

No, not yet. I suspect that in the CPU format its effect would be negligible, and in the OpenCL format it would be hidden by changes in GPU clock rate (so would require a lengthy benchmark to see).

PR #4867, opened 2021-11-11
bitcoin-opencl: Avoid unnecessary limiting of LWS for main kernels

### LINKED ISSUE #4853: Bitcoin-opencl register pressure increase on Titan Kepler
Looks like we got a regression. A build from 2021/04/27 happily accepts explicit LWS=768 GWS=10752 for Bitcoin-opencl on Titan Kepler. A build from 2021/09/14 instead uses LWS=640 GWS=10240 when invoked with those old explicit values. I didn't try to bisect and troubleshoot this further yet.

Edit, by the way, the speeds at those adjusted values are actually A LOT worse than at the previously hand-tuned LWS=768 GWS=10752.

magnumripper:
Constraints of init_kernel would affect other kernels and thus overall performance, even though init_kernel itself isn't performance critical. This workaround removes that kernel from the equation and instead runs it without an explicit LWS if needed (letting OpenCL runtime decide).

Closes #4853

### DIFF
--- a/src/opencl_bitcoin_fmt_plug.c
+++ b/src/opencl_bitcoin_fmt_plug.c
@@ -128,6 +128,7 @@ static cl_int cl_error;
 static cl_mem mem_in, mem_salt, mem_state, mem_cracked;
 static struct fmt_main *self;
 static cl_kernel init_kernel, final_kernel;
+static size_t init_kernel_max_lws;
 
 static size_t in_size, salt_size, state_size, cracked_size;
 
@@ -152,8 +153,9 @@ static const char *warn[] = {
 /* ------- Helper functions ------- */
 static size_t get_task_max_work_group_size()
 {
-	size_t s = autotune_get_task_max_work_group_size(FALSE, 0, init_kernel);
-	s = MIN(s, autotune_get_task_max_work_group_size(FALSE, 0, crypt_kernel));
+	init_kernel_max_lws = autotune_get_task_max_work_group_size(FALSE, 0, init_kernel);
+
+	size_t s = autotune_get_task_max_work_group_size(FALSE, 0, crypt_kernel);
 	s = MIN(s, autotune_get_task_max_work_group_size(FALSE, 0, final_kernel));
 
 	return s;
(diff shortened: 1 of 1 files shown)


magnumripper:
Works fine on super's Kepler, we can merge this as soon as the build bots clears it.
```
$ make -sj8 && LWS=768 GWS=10752 ../run/john -test=0 -form:Bitcoin-opencl -dev=6 -v:5

Make process completed.
initUnicode(UNICODE, RAW/RAW)
RAW -> RAW -> RAW
Device 6: GeForce GTX TITAN
Testing: Bitcoin-opencl, Bitcoin Core [SHA512 AES OpenCL]... Loaded 13 hashes with 13 different salts to test db from test vectors
Options used: -I opencl -cl-mad-enable -DSM_MAJOR=3 -DSM_MINOR=5 -cl-nv-verbose -D__GPU__ -DDEVICE_INFO=131090 -D__SIZEOF_HOST_SIZE_T__=8 -DDEV_VER_MAJOR=418 -DDEV_VER_MINOR=39 -D_OPENCL_COMPILER -DSZ=128 -DPLAINTEXT_LENGTH=125 $JOHN/opencl/bitcoin_kernel.cl
LWS=768 GWS=10752 (14 blocks) PASS
```

src/opencl_bitcoin_fmt_plug.c
```diff
@@ -128,6 +128,7 @@ static cl_int cl_error;
 static cl_mem mem_in, mem_salt, mem_state, mem_cracked;
 static struct fmt_main *self;
 static cl_kernel init_kernel, final_kernel;
+static int init_kernel_max_lws;
```
solardiz: Nitpick: OpenCL spec and thus we use `size_t` for LWS elsewhere in this file, so maybe `init_kernel_max_lws` would be cleaner as a `size_t` rather than an `int`.

solardiz (approved):

src/opencl_bitcoin_fmt_plug.c
magnumripper: True. In this case it really doesn't matter but I'll change it for good measure.

Issue #4872, opened 2021-11-17
blockchain2john.py --base64 doesn't generate a 12700 hash from a blockchain-V0 wallet for hashcat


shoutmarble:
blockchain2john.py --base64 produces no output

Here are the steps that I tried

https://github.com/openwall/john/
./blockchain2john.py --base64 blockchain-v0.0-wallet.aes.json
Produces no hash output

I then wanted to plugin the hash from "blockchain2john.py --base64" into hashcat

I verified the wallet with BTCRecover and I wanted to verify cracking this wallet with HASHCAT

The V0 Blockchain wallet
https://github.com/gurnec/btcrecover
/btcrecover/btcrecover/test/test-wallets/blockchain-v0.0-wallet.aes.json

python btcrecover.py --wallet btcrecover/test/test-wallets/blockchain-v0.0-wallet.aes.json --typos-capslock --tokenlist ./docs/Usage_Examples/2020-05-08_Recovering_Blockchain_Wallet_Passwords/tokenListTest.txt

Password found: 'btcr-test-password'

I want to reproduce the crack from BTCRecover with HASHCAT below 

###HASHCAT 12700 HASH COMPATIBLITY CHECK
docker pull dizcza/docker-hashcat:cuda
docker run --gpus all --privileged -v $PWD/:/root/wallets/ -it dizcza/docker-hashcat:cuda /bin/bash
hashcat -w 3 -S --session sessiontestwalletdat -m 12700 -o decryptedpassphrase.txt -a 3 V0HASH.HASH INCR_MASK.hcmask

I want to be able to crack the 12700 V0 blockchain wallet with HASHCAT

magnumripper:
[blockchain-v0.0-wallet.aes.json](https://raw.githubusercontent.com/gurnec/btcrecover/master/btcrecover/test/test-wallets/blockchain-v0.0-wallet.aes.json)

> blockchain2john.py --base64 produces no output

Works for me:
```
$ ../run/blockchain2john.py --base64 ~/Desktop/blockchain-v0.0-wallet.aes.json
[...]

PR #4874, opened 2021-11-17
blockchain2john: Fix for python2 vs. python3

magnumripper:
This avoids a deprecation warning under python3 while still working under python2.

See #4872

### DIFF
--- a/run/blockchain2john.py
+++ b/run/blockchain2john.py
@@ -8,6 +8,8 @@ import argparse
 import json
 import traceback
 
+PY3 = sys.version_info[0] == 3
+
 if __name__ == '__main__':
 
     parser = argparse.ArgumentParser(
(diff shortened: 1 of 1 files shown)


solardiz:
@exploide Looks good to you?

magnumripper:
I nicked it almost verbatim from other formats

solardiz (approved):

exploide:
> @exploide Looks good to you?

I'm a bit late to the party, but yes, looks like it works.

Concerning the different function names for Python 2 and 3, there also is `base64.b64decode(...)` available for both versions. However, it returns bytes on py3 and string on py2. Depends on the further processing whether this is a problem. Most often an additional `.decode()` forces it to be a plain string. But since it is fed to `hexlify()` here that shouldn't be a problem I guess.

PR #4879, opened 2021-11-18
Add option --mask-internal-target=N

### LINKED ISSUE #4876: Consider adding a --mask-int-target=N option
See https://github.com/openwall/john/issues/4871#issuecomment-972959657

We could add an option for overriding any figure set by the format. The best-effort algorithms used for determining the size of the internal mask is far from perfect so in some cases we might want to set it manually for an actual cracking session. Apart from that it would be handy for testing as well.

magnumripper:
This overrides format's algorithm for selecting the target multiplier.

Closes #4876

### DIFF
--- a/doc/NEWS
+++ b/doc/NEWS
@@ -234,6 +234,10 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
 - Added apop2john.py: script to extract and format POP3 APOP challenge/responses
   [Mark Silinio; 2021]
 
+- Added option --mask-internal-target=N for overriding format's idea of how
+  large portion of a mask should be generated on device-side (or completely
+  disabling internal mask).  [magnum; 2021]
+
 
 Major changes from 1.8.0-jumbo-1 (December 2014) to 1.9.0-jumbo-1 (May 2019):
 
--- a/src/bench.c
+++ b/src/bench.c
@@ -51,6 +51,7 @@
 #include "gpu_common.h"
 #include "opencl_common.h"
 #include "mask.h"
+#include "mask_ext.h"
 #include "aligned.h"
 
 #ifndef BENCH_BUILD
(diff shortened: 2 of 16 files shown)


magnumripper:
Current commit adds a note in NEWS but not in doc/MASK because we haven't mentioned internal mask portions to that detail so far in any docs and I'm reluctant to start doing it now (because I'm lazy, if nothing else).

magnumripper:
Hang on, I'm pondering reverting most format source changes and move most logic to some shared place instead. If nothing else, that'd let me print/log things like "Using multiplier X, requested was Y overriding the default of Z" (obviously depending on verbosity and whatnot).

magnumripper:
> I'm reluctant to start doing it now (because I'm lazy, if nothing else).

Another reason is I actually do not know what logic is ultimately used for matching the target multiplier to whatever alternatives we have from the actual mask.

magnumripper:
>  I'm pondering reverting most format source changes and move most logic to some shared place instead

I'll postpone that and just merge this

magnumripper:
> > I'm pondering reverting most format source changes and move most logic to some shared place instead
> 
> I'll postpone that and just merge this

Redecided again, still considering shared code change

magnumripper:
Much smaller changes now, and logging when used

magnumripper:
I already noticed a problem similar to #4880 using this option: My AMD Radeon Pro Vega 20 (laptop, speed index of 947200) using nt-opencl with mask ?a?a will only get x95 and a speed of 3531M. If I add `--mask-internal-target=10000` I get x9025 and 6852M.

magnumripper:
The formats that don't use the shared autotune code wouldn't print the `x9025` figure after LWS/GWS figures, so I fixed that in a separate commit.

magnumripper:
Changed the code to allow `--mask-internal-target=0` for actually disabling internal mask. This revealed a bug in DEScrypt-opencl, I'll open that as a separate issue.

magnumripper:
> I'll open that as a separate issue.

I just fixed it (LM had the same problem) in a separate commit in this PR instead.

src/opencl_DES_bs_plug.c
```diff
@@ -1145,7 +1145,7 @@ size_t create_keys_kernel_set_args(int mask_mode)
 	cl_ulong const_cache_size;
 	int i;
 
-	if (mask_mode) {
+	if (mask_mode && mask_int_cand_target) {
```
magnumripper: self-reviewing this, it's not sufficient. I need to change the callers and ultimately whatever sets `mask_mode`.

magnumripper: I will remove the LM and DEScrypt bugfixes from this PR and open an issue.

Issue #4883, opened 2021-11-21
Make ccache2john.py Python 3 compatible


marksilinio:
ccache2john.py still uses `#!/usr/bin/env python2`
should be `#!/usr/bin/env python`

marksilinio:
https://github.com/openwall/john/pull/4884

solardiz:
> should be `#!/usr/bin/env python`

As discussed on the PR, we are not so sure yet. Let's continue to discuss on the PR.

PR #4884, opened 2021-11-21
env python2 -> python

marksilinio:
https://github.com/openwall/john/issues/4883

change scripts python path from `#!/usr/bin/env python2` to `#!/usr/bin/env python`

### DIFF
--- a/run/bitwarden2john.py
+++ b/run/bitwarden2john.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python
 
 """Utility to extract Bitwarden "hashes" from Google Chrome / Firefox / Android local data"""
 
--- a/run/ccache2john.py
+++ b/run/ccache2john.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python
 
 """
 This script extracts crackable hashes from krb5's credential cache files (e.g.


exploide:
Did you test whether `bitwarden2john.py` and `ccache2john.py` are really compatible with both, Python 2 and Python 3?

Unfortunately, i can't find [sample files](https://github.com/openwall/john-samples) to test with for these formats. If you indeed tested it, maybe you also want to contribute sample files there for future testing (given they don't contain sensitive real-world data).

exploide:
A colleague helped me out with his lab DC and created a ccache file containing a TGT.
[windomain-vagrant-vagrant.ccache.zip](https://github.com/openwall/john/files/7597174/windomain-vagrant-vagrant.ccache.zip)
User is `windomain\vagrant` and password is `vagrant`.

Unfortunately, this is not a suitable test file, because john and `ccache2john.py` don't support etype 18 yet.
However, it was enough to produce an error when running under python3, while python2 shows the expected unsupported message.

This means the tool is not yet compatible with python3, hence the shebang should stay as is. Once we have a proper etype 23 ccache file, we can start porting the tool to python3.

ghost:
Thanks @exploide.

@marksilinio, we'll probably close this as code changes are needed.

solardiz:
Thank you, @exploide!
This PR also updates `bitwarden2john.py`. What about that one?

> A colleague helped me out with his lab DC and created a ccache file containing a TGT.
> [windomain-vagrant-vagrant.ccache.zip](https://github.com/openwall/john/files/7597174/windomain-vagrant-vagrant.ccache.zip)
> User is `windomain\vagrant` and password is `vagrant`.

Even though we don't support this yet, I suggest you send a PR with it to `john-samples`. It's fine to add files to there proactively, and moreover as you point out the file was useful for some testing anyway. Perhaps add a note nearby about the lack of full support. Thank you!

exploide:
Ok, I will submit a PR with that file to john-samples soon.

Concerning bitwarden, I never used that software, so I don't have a file at hand. Over at hashcat, I read that there is a new format of bitwarden, so maybe it isn't even sufficient to download the latest version to produce a sample. Better someone who actually uses this software takes care about this.

solardiz:
@marksilinio Should we simply close this PR (I guess so), or can you "defend" its change for bitwarden and if so keep only that one (force-push)? Thanks!

exploide:
> Ok, I will submit a PR with that file to john-samples soon.

Ah, no. Actually I think it wasn't a good idea get a TGT. The encrypted ticket contained within should be a service ticket for the ticket granting service itself. And that is usually generated randomly and uncrackable anyway. But maybe I get a chance to obtain a service ticket for a SPN that can be cracked soon. If I get something I will contribute it.

marksilinio:
> @marksilinio Should we simply close this PR (I guess so), or can you "defend" its change for bitwarden and if so keep only that one (force-push)? Thanks!

I think it's ok to close this PR for now
I'll try to find out some ccache/bitwarden samples in the future

Issue #4882, opened 2021-11-21
APOP hash extraction script


marksilinio:
It would be useful to have a `pcap2apop.py` script to extract APOP salted hash (RFC1939):
```
APOP name digest

         Arguments:
             a string identifying a mailbox and a MD5 digest string
             (both required)

         Restrictions:
             may only be given in the AUTHORIZATION state after the POP3
             greeting or after an unsuccessful USER or PASS command

         Discussion:
             Normally, each POP3 session starts with a USER/PASS
             exchange.  This results in a server/user-id specific
             password being sent in the clear on the network.  For
             intermittent use of POP3, this may not introduce a sizable
             risk.  However, many POP3 client implementations connect to
             the POP3 server on a regular basis -- to check for new
             mail.  Further the interval of session initiation may be on
             the order of five minutes.  Hence, the risk of password
             capture is greatly enhanced.

             An alternate method of authentication is required which
             provides for both origin authentication and replay
             protection, but which does not involve sending a password
             in the clear over the network.  The APOP command provides
             this functionality.

             A POP3 server which implements the APOP command will
             include a timestamp in its banner greeting.  The syntax of
             the timestamp corresponds to the `msg-id' in [RFC822], and
             MUST be different each time the POP3 server issues a banner
             greeting.  For example, on a UNIX implementation in which a
             separate UNIX process is used for each instance of a POP3
             server, the syntax of the timestamp might be:

                <process-ID.clock@hostname>
```

solardiz:
We're done here, so I'll close the issue.

@marksilinio For further occasions, you can use the word "Closes" or "Fixes" in place of "Implements" - then GitHub would close the issue automatically. Also, I suggest putting this on a separate line (would be line 3+, after commit title and an empty line).

Thank you!

PR #4887, opened 2021-11-21
Add apop2john.py to extract POP3 APOP traffic. Implements #4882

marksilinio:
https://www.root-me.org/en/Challenges/Network/POP-APOP can be used for testing

#4882

### DIFF
--- a/run/apop2john.py
+++ b/run/apop2john.py
@@ -1,0 +1,66 @@
+#!/usr/bin/env python
+
+# This software is Copyright (c) 2021 Mark Silinio <mark.silinio-at-gmail.com>,
+# and it is hereby released to the general public under the following terms:
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted.
+#
+# Extract and format POP3 APOP challenge/responses for password cracking with JtR
+# Usage: ./apop2john.py <pcap files>
+#
+# This script depends on Scapy (https://scapy.net)
+# To install: pip install --user scapy
+
+try:
+    from scapy.all import *
+except ImportError:
+    print("scapy is missing, run 'pip install --user scapy' to install it!")
+    exit(1)
+
+from binascii import hexlify
+from sys import argv
+import os
+import re
+
+if len(argv) < 2:
+    print('Usage: ./apop2john.py <pcap files>')
+    exit(1)
+
+filenames = argv[1:]
+
+for filename in filenames:
+
+    capture_file = rdpcap(filename)
+
+    apop_salt = {}
+    apop_hash = {}
+    apop_user = {}
+
+    for packet in capture_file:
+
+        if not TCP in packet or (packet.sport != 110 and packet.dport != 110):
+           continue
+
+        pkt = bytes(packet[TCP].payload)
+
+        if packet.sport == 110 and re.search(b'\+OK\ .*\ \<.+\>', pkt):
+            src_ip = packet[IP].src
+            dst_ip = packet[IP].dst
+            res = re.search(b'\+OK\ .*\ (\<.+\>)', pkt)
+            apop_salt[(src_ip, dst_ip)] = res.group(1).strip()
+        elif packet.dport == 110 and re.search(b'APOP\ .+\ (.+)', pkt):
+            src_ip = packet[IP].dst
+            dst_ip = packet[IP].src
+            res = re.search(b'APOP\ (.+)\ (.+)', pkt)
+            apop_user[(src_ip, dst_ip)] = res.group(1).strip()
+            apop_hash[(src_ip, dst_ip)] = res.group(2).strip()
+
+    for ips_s, salt in apop_salt.items():
+        for ips_h, ahash in apop_hash.items():
+            for ips_u, user in apop_user.items():
+                if (ips_s == ips_h == ips_u): 
+                    print('{user}:$dynamic_1017${hash}$HEX${salt}'.format(
+                        user=user.decode('utf-8'),
+                        hash=ahash.decode('utf-8'),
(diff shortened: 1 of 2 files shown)


ghost (changes requested):
Please, create a PR and provide [sample files](https://github.com/openwall/john-samples) so we can test.

marksilinio:
> Please, create a PR and provide [sample files](https://github.com/openwall/john-samples) so we can test.

see https://github.com/openwall/john-samples/pull/8

ghost (approved):

run/apop2john.py
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env python
+
+# This software is Copyright (c) 2021 Mark Silinio <mark.silinio-at-gmail.com>,
+# and it is hereby released to the general public under the following terms:
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted.
+#
+# Utility to bruteforce APOP hash
```
solardiz: I think we need to change this comment and the similar one in `radius2john.py` to say something more reasonable like, "Extract and format POP3 APOP challenge/responses for password cracking with JtR" here and "Extract and format suitable RADIUS packets for shared secret cracking with JtR" there.

solardiz (approved):
What versions of Python has this been tested with? Is this known to be compatible or incompatible with Python 2, or not known either way?

I ask because we currently use explicit `python3` on the shebang line for scripts that are known to require Python 3. This one says simply `python`, so (given our current convention) claims to work with both Python 2 and 3. Ditto for `radius2john.py`, so if any fix is needed for both scripts then it can be in a separate PR.

exploide:
I'm wondering: Would it make sense to wrap this into a function and incorporate it in [pcap2john.py](https://github.com/openwall/john/blob/bleeding-jumbo/run/pcap2john.py) instead of having a new standalone `apop2john.py` script?

Most other hash extraction logic that operates on pcap files is bundled there. Only few other standalone pcap parsers exist.

Having it in `pcap2john.py` has the advantage that one does not need to know beforehand what types of hashes the pcap file contains. It just extracts everything it knows about.

solardiz:
> Would it make sense to wrap this into a function and incorporate it in [pcap2john.py](https://github.com/openwall/john/blob/bleeding-jumbo/run/pcap2john.py) instead of having a new standalone `apop2john.py` script?

I'd support this, but unfortunately I find `pcap2john.py` somewhat problematic: it has extra hard dependency on `dpkt` and no way to choose to extract traffic of just one protocol. (And for a multi protocol extractor, it's also somewhat inefficient in that it goes over the input multiple times, once per protocol. Since that's how it works anyway, it feels silly that there's no way to force it to just one protocol when desired.)

So for now this PR should stay as-is - adding a new script. We already have `radius2john.py` anyway.

BTW, @marksilinio this is a significant enough contributions that I think you should add a `doc/NEWS` entry for it (then we'll revise the entry if we merge the script into `pcap2john.py` before the next release).

solardiz (changes requested):
The reference to 4883 in the commit message is wrong, should be 4882 or nothing. Please correct.

solardiz:
I'd use "Add apop2john.py to extract POP3 APOP traffic" for the commit message.

exploide:
> I'd support this, but unfortunately I find pcap2john.py somewhat problematic: it has extra hard dependency on dpkt and no way to choose to extract traffic of just one protocol. (And for a multi protocol extractor, it's also somewhat inefficient in that it goes over the input multiple times, once per protocol. Since that's how it works anyway, it feels silly that there's no way to force it to just one protocol when desired.)

Fully agree. Having two network packet parsing libaries (dpkt and scapy) ist not very clean. Most of the extractors use dpkt, however I consider scapy being more powerful. It could also simplify portions of the code.

I also thought about the inefficiency, you are right. It could be improved in multiple ways. I can for example imagine that each extractor function is refactored in a class that holds the logic and also offers a filter function. In the main function, each packet would be checked against each filter function, that is supposed to be fast (only check protocol, port number or the like) and is passed to the extractor only if applicable. No need to re-read the whole pcap file in each and every extractor.

But this requires a lot of work. I think I don't find the time right now. But if anyone is interested, I'm willing to brain-storm further.

But this goes off-topic here.

marksilinio:
> I think we need to change this comment and the similar one in `radius2john.py` to say something more reasonable like, "Extract and format POP3 APOP challenge/responses for password cracking with JtR" here and "Extract and format suitable RADIUS packets for shared secret cracking with JtR" there.

Updated

> What versions of Python has this been tested with? Is this known to be compatible or incompatible with Python 2, or not known either way?

Updated - now it both python2/python3 compatible:
```
x@xxx:~/marksilinio/github/john$ python2.7 run/apop2john.py ../../tmp/apop-ch.pcapng 
billcl:$dynamic_1017$5aadb7d8a2639088a7036fe6d5cd5727$HEX$3c313733332e312e39663430333632352e4262574767704b7a5a4b4b4c386f7067576e307775413d3d407670732d61653961626337323e
x@xxx:~/marksilinio/github/john$ python3 run/apop2john.py ../../tmp/apop-ch.pcapng 
billcl:$dynamic_1017$5aadb7d8a2639088a7036fe6d5cd5727$HEX$3c313733332e312e39663430333632352e4262574767704b7a5a4b4b4c386f7067576e307775413d3d407670732d61653961626337323e
```

> The reference to 4883 in the commit message is wrong, should be 4882 or nothing. Please correct.

fixed

> I'd use "Add apop2john.py to extract POP3 APOP traffic" for the commit message.

fixed

solardiz (approved):
Looks good to me. Thank you!

PR #4891, opened 2021-11-22
Add knowledge of Apple M1 GPU's core count

magnumripper:
This changes our best-effort "speed index" for Apple M1 GPU's and, as a result of that, greatly increases speeds for most internal mask OpenCL formats (namely the ones that use said speed index to decide on a mask accleration target).

Info was taken from https://en.wikipedia.org/wiki/Apple_M1#GPU - @solardiz do you think it's correct to translate

*Each GPU core is split into 16 Execution Units, which each contain eight Arithmetic Logic Units (ALUs). In total, the M1 GPU contains up to 128 Execution units or 1024 ALUs,[11] which Apple says can execute up to 24,576 threads simultaneously and which have a maximum floating point (FP32) performance of 2.6 TFLOPs.[8][12]*

to

Stream processors = core_count * 16 *8 ?

Perhaps it should just be core_count * 16. Anyway with the former, speed for internal mask format was boosted significantly due to the fact that target mask multiplier is based on our "speed index":

Before
```
./john --list=opencl-devices
(...)
    Max clock (MHz):        1000
    Parallel compute cores: 8
    Speed index:            8000
(...)

Device 1: Apple M1
Benchmarking: LM-opencl [DES BS OpenCL/mask accel]... LWS=64 GWS=1048576
DONE
Raw:	140136K c/s real, 405260K c/s virtual

Benchmarking: NT-opencl [MD4 OpenCL/mask accel]... LWS=256 GWS=8192 (32 blocks) x26 DONE
Raw:	219455K c/s real, 279181K c/s virtual
```

After
```
./john --list=opencl-devices
(...)
    Max clock (MHz):        1000
    Parallel compute cores: 8
    Stream processors:      1024  (8 x 128)
    Speed index:            1024000
(...)

Device 1: Apple M1
Benchmarking: LM-opencl [DES BS OpenCL/mask accel]... LWS=32 GWS=16384
DONE
Raw:	283209K c/s real, 14868M c/s virtual

Benchmarking: NT-opencl [MD4 OpenCL/mask accel]... LWS=256 GWS=1048576 (4096 blocks) x3300 DONE
Raw:	2572M c/s real, 173015M c/s virtual
```

### DIFF
--- a/src/opencl_common.c
+++ b/src/opencl_common.c
@@ -2562,6 +2562,9 @@ cl_uint get_processors_count(int sequential_id)
 	} else if (gpu_intel(device_info[sequential_id])) {
 		// It seems all current models are x 8
 		core_count *= ocl_device_list[sequential_id].cores_per_MP = 8;
+	} else if (!strcmp(dname, "Apple M1")) {
+		// Each GPU core is split into 16 Execution Units, which each contain eight Arithmetic Logic Units (ALUs)
+		core_count *= ocl_device_list[sequential_id].cores_per_MP = 16 * 8;
 	} else if (gpu_amd(device_info[sequential_id])) {
 		// 16 thread proc * 5 SP
 		core_count *= (ocl_device_list[sequential_id].cores_per_MP = (16 *


magnumripper:
> Perhaps it should just be core_count * 16

I tried using just `16` instead of `16 * 8` but that resulted in lower benchmarks.
```
./john --list=opencl-devices
(...)
    Max clock (MHz):        1000
    Parallel compute cores: 8
    Stream processors:      128  (8 x 16)
    Speed index:            128000
(...)

Device 1: Apple M1
Benchmarking: LM-opencl [DES BS OpenCL/mask accel]... LWS=64 GWS=131072
DONE
Raw:	256953K c/s real, 2595M c/s virtual

Benchmarking: NT-opencl [MD4 OpenCL/mask accel]... LWS=256 GWS=262144 (1024 blocks) x330 DONE
Raw:	2108M c/s real, 10868M c/s virtual
```

solardiz:
> > Perhaps it should just be core_count * 16
> 
> I tried using just `16` instead of `16 * 8` but that resulted in lower benchmarks.

I think `16 * 8` is actually correct. By SPs, we refer (in the rest of the code) to an equivalent thing of CUDA cores (as of before Ampere, see #4892), not to anything larger. This also seems to match AMD's terminology.

solardiz (approved):

PR #4895, opened 2021-11-24
Minor fixes for NVIDIA Ampere support

### LINKED ISSUE #4892: "CUDA cores" reporting is wrong on Ampere
I think we should halve the number of "cores" in the computation in `opencl_speed_index()` when we're on NVIDIA Ampere and above, because half of their reported cores are FP-only, thus currently unusable for us.

Alternatively, we could halve our reported number of cores (fits our current code better, with GPU type checks all in one place), but then we need to adjust the wording from "CUDA cores" e.g. to "INT32 CUDA cores". Unfortunately, this might then become incorrect in some future NVIDIA architecture - not just incorrect speed index, but incorrect reporting of something that we claim to be so specific (INT32 cores). That's not ideal.

magnumripper:
Document (in code) Ampere's CC and try to recognize it even on macOS.
Change the wording in --list=opencl-devices to "CUDA INT32 cores" as there might be cores supporting only FP and so on.

Closes #4892

### DIFF
--- a/src/opencl_common.c
+++ b/src/opencl_common.c
@@ -2572,7 +2572,7 @@ cl_uint get_processors_count(int sequential_id)
 		               ((amd_gcn(device_info[sequential_id]) ||
 		                 amd_vliw4(device_info[sequential_id])) ? 4 : 5)));
 	} else {
-		// Nothing else known, we use half native vector width for long
+		// Nothing else known, we use the native vector width for long.
 		cl_uint v_width;
 
 		HANDLE_CLERROR(clGetDeviceInfo(devices[sequential_id],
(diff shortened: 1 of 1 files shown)


src/opencl_common.c
```diff
@@ -2541,8 +2541,8 @@ cl_uint get_processors_count(int sequential_id)
  * This will produce a *guessed* figure
  */
 
-		// Volta or Turing
-		else if (strstr(dname, "TITAN V") || strstr(dname, "RTX 2"))
+		// Volta, Turing, Ampere
+		else if (strstr(dname, "TITAN V") || strstr(dname, "RTX 2") || strstr(dname, "RTX 3"))
```
solardiz: We should also detect "RTX A", and in fact I think it'd be safer to make it just "RTX " instead of listing the 2, 3, A prefixes. Yes it would also detect future RTX series, but I think it's probably best to guess the same as previous known revision.

We can also do `dname[0] == 'A' && dname[1] >= '1' && dname[1] <= '9'` for the A100, etc.

solardiz (approved):

PR #4898, opened 2021-11-26
Add authenticator2john.py to extract authenticator app passwords. Closes #4893

### LINKED ISSUE #4893: Add format and extractor script for ```authenticator``` python app master password
It would be nice to have a script that allow to bruteforce master password of
https://github.com/JeNeSuisPasDave/authenticator

it uses the AES 256-bit symmetric key:
https://github.com/JeNeSuisPasDave/authenticator/blob/3caa2edebff01be3e6fd82c72ac3a0eb415cbe66/src/authenticator/data.py#L597-L606

to encrypt the second 16-byte data and compare it with the first 16-byte block (header):
https://github.com/JeNeSuisPasDave/authenticator/blob/3caa2edebff01be3e6fd82c72ac3a0eb415cbe66/src/authenticator/data.py#L802-L807

marksilinio:
see https://github.com/openwall/john/issues/4893 for details


marksilinio:
python2/python3 compatible:

```
$ python2.7 ./authenticator2john.py ~/.authenticator/authenticator.data 
authenticator.data:$authenticator$0$7a6a5a4a00000001000400007a6a5a4a$24b6b919b5685dc9c6ac4e6e4fba8026
$ python3.9 ./authenticator2john.py ~/.authenticator/authenticator.data 
authenticator.data:$authenticator$0$7a6a5a4a00000001000400007a6a5a4a$24b6b919b5685dc9c6ac4e6e4fba8026
```

magnumripper (approved):
Looks fine to me, if @solardiz is fine with the license blurb

marksilinio:
Are there any preferred licenses?

magnumripper:
Our most preferred license is the cut-down BSD license, but I believe GPLv2 is fine too.

solardiz:
@marksilinio magnum is correct, our preferred is the same one you had used in `apop2john.py` - can you do the same here? GPLv2 is also fine if you insist, just let us know if so. Thanks!

solardiz:
Do we want to merge this script now or only along with the corresponding JtR format?

We similarly had a PR with `bestcryptve2john.py` hanging around for months and only merged when we also had the format. The reason for not merging the script right away was not to cause confusion by something not really supported yet.

marksilinio:
> @marksilinio magnum is correct, our preferred is the same one you had used in `apop2john.py` - can you do the same here? GPLv2 is also fine if you insist, just let us know if so. Thanks!

Done

magnumripper:
> Do we want to merge this script now or only along with the corresponding JtR format?
> 
> We similarly had a PR with `bestcryptve2john.py` hanging around for months and only merged when we also had the format. The reason for not merging the script right away was not to cause confusion by something not really supported yet.

I'll write the formats soon so let's wait a little. I might even be able to push the formats commits to this same PR

solardiz:
> I'll write the formats soon so let's wait a little.

OK.

> I might even be able to push the formats commits to this same PR

I think it's better to have two PRs, but plan on merging them on the same day.

its0x08 (approved):

ghost:
This PR now has **Conflicting files** that must be resolved before merging. I think it is bad to wait so long for a merge.

We must merge it. Maybe move the 2john* file to the "unused folder " for a while.

solardiz:
> Merge branch 'bleeding-jumbo' into authenticator2john

Let's not be doing things like that. Let's be rebasing instead.

ghost:
> > Merge branch 'bleeding-jumbo' into authenticator2john
> 
> Let's not be doing things like that. Let's be rebasing instead.

I used the Github GUI to fix the source code conflict. 

* Or the author amend
* Or an administrator handles
* Or we use the squash option to merge

magnumripper:
Time flies, this PR is from 2021.

> I'll write the formats soon so let's wait a little. I might even be able to push the formats commits to this same PR

Evidently this didn't happen yet.

I think we should fix and merge this, perhaps just add some output to stderr that the format might not be supported by us yet. BTW the hashcat folks often use our tools and they might come up with a format (or maybe even did already?).

solardiz:
> I think we should fix and merge this, perhaps just add some output to stderr that the format might not be supported by us yet.

Please feel free to do that.

doc/NEWS
```diff
@@ -238,6 +238,10 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
   large portion of a mask should be generated on device-side (or completely
   disabling internal mask).  [magnum; 2021]
 
+- Added authenticator2john.py: script to extract and format
+  https://github.com/JeNeSuisPasDave/authenticator app passwords.
```
KG2304, line 242: Kaali

magnumripper:
Um, what is the current "correct" way of calling python? Shebangs like `#!/usr/bin/env python` doesn't work on Ubuntu (at least not my 24.04) but some of our tools (and this one) use that, while others use `#!/usr/bin/env python3` and a couple even has `#!/usr/bin/env python2`.
```
$ dpkg -S "/usr/bin/python*"
python3.12-minimal: /usr/bin/python3.12
python3-minimal: /usr/bin/python3
```

magnumripper:
Apparently Ubuntu has a package `python-is-python3` which creates a symlink to python3 and "python" is otherwise still expected to point to python2 (so nowadays neither exist). I assume python-is-python3 is not installed by default (My gear did not have it but I installed it now).

solardiz:
> Um, what is the current "correct" way of calling python? Shebangs like `#!/usr/bin/env python` doesn't work on Ubuntu (at least not my 24.04) but some of our tools (and this one) use that, while others use `#!/usr/bin/env python3` and a couple even has `#!/usr/bin/env python2`.

Our conversion was to use `#!/usr/bin/env python` for scripts that work with either Python 2 or 3, `#!/usr/bin/env python3` for scripts that are Python 3 only, and `#!/usr/bin/env python2` for scripts that are Python 2 only. Unfortunately, it looks like our convention for universal script is detached from distro reality.

If so, I suggest we convert them to use `#!/usr/bin/env python3` because a modern distro is unlikely to have anything else available by default, but add a comment just below saying that the script also works with Python 2.

This is especially important and unfortunate for `bitcoin2john.py`, which supports both versions of Python but is easier to get working with Python 2 (BDB bundled) than 3. So we may want to put a more elaborate comment into that one script.

Let's track this all on a new issue maybe?

PR #4900, opened 2021-11-26
DEScrypt & LM OpenCL bugfixes

### LINKED ISSUE #4897: Bug in LM and DEScrypt OpenCL formats for when target mask factor is 0
This happens when using `--rules-stack` because we need to disable internal mask then. It can (at least theoretically) also happen on very weak GPUs just from the formula that sets the target from `opencl_speed_index() / some_divisor`. The `--mask-internal-target` option added in #4879 can also be used to set the target to 0 and this is how I found this problem.

Both LM & DEScrypt formats segfaults, because they use code paths for internal mask where they shouldn't. All other formats seem to behave like they should.

My first fix:
```diff
diff --git a/src/opencl_DES_bs_plug.c b/src/opencl_DES_bs_plug.c
index 816f50735..87015409b 100644
--- a/src/opencl_DES_bs_plug.c
+++ b/src/opencl_DES_bs_plug.c
@@ -1145,7 +1145,7 @@ size_t create_keys_kernel_set_args(int mask_mode)
        cl_ulong const_cache_size;
        int i;
 
-       if (mask_mode) {
+       if (mask_mode && mask_int_cand_target) {
                fmt_opencl_DES.methods.set_key = set_key_mm;
                fmt_opencl_DES.methods.get_key = get_key_mm;
        }
diff --git a/src/opencl_lm_b_plug.c b/src/opencl_lm_b_plug.c
index a6646bbae..9af5ecae4 100644
--- a/src/opencl_lm_b_plug.c
+++ b/src/opencl_lm_b_plug.c
@@ -1175,7 +1175,7 @@ static void reset(struct db_main *db)
                release_buffer();
                create_buffer(num_loaded_hashes, offset_table, offset_table_size, hash_table_size, bitmaps, bitmap_size_bits);
 
-               if (options.flags & FLG_MASK_CHK) {
+               if (options.flags & FLG_MASK_CHK && mask_int_cand_target) {
                        mask_mode = 1;
                        fmt_opencl_lm.methods.set_key = opencl_lm_set_key_mm;
                        fmt_opencl_lm.methods.get_key = get_key_mm;
```
While I haven't *seen* any problems with this fix, it doesn't feel correct - because the `mask_mode` variable is still set and that affects other things.

So I tried changing the setting of the latter instead, but then the segfaults got back. So I dropped the above fix from #4879 and moved this problem to here.

magnumripper:
Fixes for multiple problems, WIP

Closes #4897

### DIFF
--- a/src/opencl_DES_bs_b_plug.c
+++ b/src/opencl_DES_bs_b_plug.c
@@ -597,7 +597,7 @@ static void reset(struct db_main *db)
 		release_clobj_kpc();
 		release_clobj();
 
-		if (options.flags & FLG_MASK_CHK)
+		if (options.flags & FLG_MASK_CHK && mask_int_cand.num_int_cand > 1)
 			mask_mode = 1;
 
 		create_clobj(db);
--- a/src/opencl_DES_bs_b_plug.c
+++ b/src/opencl_DES_bs_b_plug.c
@@ -684,7 +684,7 @@ static int des_crypt_25(int *pcount, struct db_salt *salt)
 
 	current_gws *= iter_count;
 	ret_code = clEnqueueNDRangeKernel(queue[gpu_id], kernels[gpu_id][0], 1, NULL, &current_gws, lws, 0, NULL, NULL);
-	HANDLE_CLERROR(ret_code, "Enque kernel DES_bs_25 failed.\n");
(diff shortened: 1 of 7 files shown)


magnumripper:
This is ripe now

PR #4905, opened 2021-11-30
Feature/rexgen 2.1.4

janstarke:
compiles with rexgen 2.1.4 (release feature branch and current master)

### DIFF
--- a/doc/INSTALL-UBUNTU
+++ b/doc/INSTALL-UBUNTU
@@ -79,16 +79,14 @@ build CPU fallback chains for any x86-64 CPU.
     as Open MPI (Message Passing Interface).  You do get OpenMP support
     compiled into JtR by default if your system supports it.
 
-==== Optional REXGEN support (experimental; additional cracking mode)
-
-    You almost certainly don't need this.
+==== Optional REXGEN support (experimental; additional "regex" cracking mode)
 
     NOTE this support is experimental and may not build at all until tweaked.
     The rexgen library is notorious for changing its API between versions,
     breaking stuff.  Leave this step out unless you really know that you want
-    it and are capable of fixing issues.  Last known good version is 2.0.8.
+    it and are capable of fixing issues.  Last known good version is 2.1.5.
 
-    sudo apt-get -y install cmake bison flex libicu-dev
+    sudo apt-get -y install cmake bison flex
     cd ~/src
     git clone --recursive https://github.com/teeshop/rexgen.git
     cd rexgen
--- a/doc/NEWS
+++ b/doc/NEWS
@@ -242,6 +242,9 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
   doesn't include bitslice formats.  It matches formats that will run SIMD
   using the default GPU, or per the  --device option.  [magnum; 2021]
 
+- Revised regex mode's support for librexgen, to version >= 2.1.5  [Jan Starke,
+  magnum; 2021]
+
 
 Major changes from 1.8.0-jumbo-1 (December 2014) to 1.9.0-jumbo-1 (May 2019):
 
--- a/doc/README.librexgen
+++ b/doc/README.librexgen
@@ -2,11 +2,10 @@ Rexgen.txt - This document describes how to use the rexgen library, to perform
 regex expression work within JtR.  Rexgen library is copywrite Jan Starke,
 jan.starke@outofbed.org   The regex.c code in JtR by JimF, Spring 2014.
 
-WARNING
-
-Since maintenance for the rexgen library appears to have stopped in
-April, 2019, support for --regex might have to be dropped in future releases!
-
+NOTE this support is experimental and may not build at all until tweaked.
+The rexgen library is notorious for changing its API between versions,
+breaking stuff.  Leave this step out unless you really know that you want
+it and are capable of fixing issues.  Last known good version is 2.1.5.
 
 First off, see the section at the bottom of this document about how to obtain
 build and install librexgen.
@@ -43,7 +42,6 @@ The current options we have are:
 
 
 Currently, rexgen can be used stand alone, OR with wordlist and rules.
-There are plans to also add this to single mode at some time.
 
 The command line switch for stand alone is --regex[=case]=expression
 The expression is a stand alone rexex expression.  If the optional
@@ -82,100 +80,8 @@ then all of the regular expressions will be performed on this word.
 ------------------------------------------------
 --- Obtaining, building, installing rexgen.  ---
 ------------------------------------------------
-First off, you must obtain and install rexgen library. It can be found here:
 
-https://github.com/teeshop/rexgen
+Clone https://github.com/teeshop/rexgen.git and build it according to
+its instructions.
 
-git clone --recursive https://github.com/teeshop/rexgen.git
-
-
-As of this writing there is no known good rexgen version.
-
-The latest commit as of this writing is
-
-    commit 585e86da6a07a63ec7a871320b307457661d557c
-    Author: Jan Starke <teeshop@users.noreply.github.com>
-    Date:   Thu Apr 11 08:38:41 2019 +0200
-
-        Update README.md
-
-The latest known good commit is
-
-commit 5b2f4b159ec948c1f9429eca4389ca2adc9c0b07
-Author: Jan Starke <jan.starke@t-systems.com>
-Date:   Wed Jan 23 14:53:50 2019 +0100
-
-    running with libFuzzer works
-
-
-Checkout this latest known good commit:
-
-cd rexgen
-git checkout 5b2f4b159ec948c1f9429eca4389ca2adc9c0b07
-
-
-Building rexgen requires cmake, flex, and bison.
-To build, see instructions on this page:
-
-https://github.com/teeshop/rexgen/blob/master/README.md
-
-
-
-Instead of just running ./install.sh, you can first run ./build.sh
-as a regular user and then just run ./install.sh as root.
-
-cd rexgen/src
-./build.sh
-sudo ./install.sh
-
-
-If there are no errors (especially from cmake), librexgen should be installed.
-Don't forget to run ldconfig afterwards.
-
-sudo ldconfig
-
-To test:
-rexgen '[0-2]password[A-C]'
-0passwordA
-1passwordA
-2passwordA
-0passwordB
-1passwordB
-2passwordB
-0passwordC
-1passwordC
-2passwordC
-
-
-Once rexgen is installed, JtR ./configure now needs
---enable-rexgen to detect and use the library.
-
-
-KNOWN ISSUES
-
-The rexgen manpage includes this example:
-
-Examples
-       rexgen index.php?id=[1-5]
-              Would create the results
-              index.php?id=1
-              index.php?id=2
-              index.php?id=3
-              index.php?id=4
-              index.php?id=5
-
-However, to produce the desired output, you have to run
-       rexgen 'index\.php\?id=[1-5]'
-
-(Note the single quotes and the need to escape both '.' and '?'.)
-
-
-
-
-Note, JtR help document is NOT here to troubleshoot installation of this
-library.  A simple how to do it (when thing work right), is what was listed.
-If there problems beyond what is addressed in this document, then the lib's
-author is probably the best way to go.  A quick 'help me' post on john-users
-may get the proper answer, AND if we get a quick resolve, we will likely
-update this document listing this issue.  But anything past the basics,
-are really beyond the scope of the JtR developers.
+Then add "--enable-rexgen" to ./configure when building john.
--- a/src/configure
+++ b/src/configure
@@ -13073,7 +13073,7 @@ See \`config.log' for more details" "$LINENO" 5; }
 else
   cat confdefs.h - <<_ACEOF >conftest.$ac_ext
 /* end confdefs.h.  */
-int main() { return ! (rexgen_version_int() >= 0x020009); }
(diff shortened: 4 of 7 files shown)


solardiz:
Thanks @janstarke.

I think you'll also need to update the documentation. Here's where rexgen is mentioned:

```
$ fgrep -rlwi rexgen ???
doc/INSTALL-UBUNTU
doc/README.librexgen
doc/CHANGES-jumbo
run/regex_alphabets.conf
run/john.bash_completion
run/rexgen2rules.pl
src/unused/bisect-script.sh
src/configure.ac
src/autoconfig.h.in
src/unicode.h
src/regex.h
src/regex.c
src/recovery.c
src/packaging/build.sh
src/configure
```

The files under `doc` refer to specific revisions, and there will need to be a `NEWS` entry about this change.

I hope someone who's actually used the rexgen mode will also comment.

magnumripper:
Thanks, I will test this

magnumripper:
I could install rexgen 2.1.4 on macOS with no problems, then configured john but it says `configure: rexgen check not enabled` because we're currently not looking for it unless `--enable-rexgen` is supplied. Maybe we should change that if things are stable again.

Reconfigured with `--enable-rexgen` and now it's found. It's currently listed in the end:
```
Development options (these may hurt performance when enabled):
AddressSanitizer ("ASan") .................. disabled
UndefinedBehaviorSanitizer ("UbSan") ....... disabled
Fuzzing test ............................... no
Experimental code .......................... no
 - librexgen (regex mode) .................. yes
```
Perhaps we should move the "Experimental code" heading on its own above "Development options". The regex mode is not a development option, and the phrase "these may hurt performance when enabled" is inappropriate (a better phrase for "Experimental code" heading could be "these options are less thoroughly tested and/or less actively maintained".

Anyway, I built john with no problems and (just a basic test of) the `-regex` mode seems to work fine.

magnumripper (approved):
I added a commit with documentation updates. I think we can merge this now. I've seen a bug already but it's in the lib so will be reported there.

solardiz:
This has the `configure` script regenerated with a different version of `autoconf`. I think we should stick to one version used by a JtR maintainer (perhaps @magnumripper) or by several of us, not have `autoconf` versions change arbitrarily with PRs like this. That creates lots of changes in `configure`, which creates noise in our git history, and which we can't reasonably review and shouldn't have to review for potential malware. ;-) Not that I expect anything like that this time, but in general.

So @magnumripper perhaps you should regenerate `configure` and force-push before merging this.

magnumripper:
Oh, I missed that! Fixed now.

solardiz:
The `configure` update is in the "rexgen version bump" commit. In fact, the changes to `configure.ac` are so trivial they can also be replicated in `configure` manually, although indeed it's more reliable to use the same version of `autoconf` as before.

magnumripper:
Perhaps we should also squash the three sequential commits to `regex.c` into one?
```
04eb06562cf82306f26faecce9d8cab38f80515d renamed c_regex_cb_mb2 back to c_regex_cb_mb
6b8c69500891e437211167be09909b9f9518479b compiles with rexgen 2.1.4
3cbb0ddc08102fc93007b96eddf45de3d78589f4 use specific types fot rexgen instead of NULL
```
New commit description could be something like "regex.c: Update for librexgen 2.1.4". Or maybe even squash them along with the `configure` commit... and why not squash the lot into a single commit 😆

solardiz:
Yes, I think this should be just one commit. I think it's wrong to have commits that make the code inconsistent, which I think the multiple commits here temporarily do. It's OK (even good) to have commits that make changes incrementally, but only as long as the tree stays consistent at any point.

magnumripper:
Fixed now. I amended my documentation commit to include "promotion" of librexgen and "experimental code", from the possibly muted "Developer options" to the normal configure summary.
```
(...)
Non-free unrar code (complete RAR support) . yes
librexgen (regex mode, default disabled) ... yes
Experimental code (default disabled) ....... no
```

solardiz:
On one hand, I agree with this "promotion". On the other, from experience many people ended up trying to build with rexgen and ran into issues for no reason (most of them didn't know nor intend to use rexgen). So maybe our documentation should continue to discourage building with rexgen. Something like "Only build with rexgen support if you know what this is and intend to use it. Most people don't need it."

It's great that @janstarke revived it now (thanks again!), but unfortunately (unless there's a specific reason not to) I do expect the API will break in some future rexgen update again (yet our releases are infrequent, and some people will end up building our latest release against latest GitHub rexgen).

magnumripper:
> On one hand, I agree with this "promotion". On the other, from experience many people ended up trying to build with rexgen and ran into issues for no reason (most of them didn't know nor intend to use rexgen). So maybe our documentation should continue to discourage building with rexgen. Something like "Only build with rexgen support if you know what this is and intend to use it. Most people don't need it."

We could (re)add some caveat warnings in INSTALL-UBUNTU and README.librexgen, but I think `configure` is fine as-is. They say "default disabled" just like OpenMPI support did for ages.

magnumripper:
> Most people don't need it

Well if it wasn't for the unstable history of it, it's really a fantastic cracking mode. With just a carefully thought out regexp you can produce candidates nearly impossible to produce with other modes such as mask and/or rules.

I might be able to tweak the configure summary to get a little more room for a short notice (such as "see doc/README.librexgen").

frank-dittrich:
I think this hasn't worked for a long time, but it used to work several years ago and really should work, no?
```
$ ./john --regex=alpha:leet --stdout --wordlist=password.lst 
syntax error, unexpected end of file
Error, invalid regex expression.  John exiting now  base_word=123456  Regex= 
```
This mode is defined in regex_alphabets.conf
```
# simple 1337 mode.  ONLY leet's lower case letters, and smallish alphabet. But VERY fast.
[List.Rexgen.Alpha:leet]
a=[a4@]
b=[b8]
e=[e3]
g=[g9]
i=[i!]
l=[l17]
o=[o0]
s=[s$5]
t=[t+7]
```

Trying something trivial instead:
```
$ ./john --regex="\0123" --stdout --wordlist=password.lst 
```

What am I doing wrong here?

Without a word list it works:
```
$ ./john --regex="[12][ab]" --stdout                         
1a
2a
1b
2b
4p 0:00:00:00 0.00% 133.3p/s 2b
```

This is on Fedora 35

```
$ ./john --list=build-info 
Version: 1.9.0-jumbo-1+bleeding-38b9536 2021-11-30 20:46:27 +0100
Build: linux-gnu 64-bit x86_64 AVX512BW AC OMP
SIMD: AVX512BW, interleaving: MD4:3 MD5:3 SHA1:1 SHA256:1 SHA512:1
CPU tests: AVX512BW
$JOHN is ./
Format interface version: 14
Rec file version: REC4
Charset file version: CHR3
CHARSET_MIN: 1 (0x01)
CHARSET_MAX: 255 (0xff)
CHARSET_LENGTH: 24
SALT_HASH_SIZE: 1048576
SINGLE_IDX_MAX: 2147483648
SINGLE_BUF_MAX: 4294967295
Max. Markov mode level: 400
gcc version: 11.2.1
GNU libc version: 2.34 (loaded: 2.34)
OpenCL headers version: 1.2
Crypto library: OpenSSL
OpenSSL library version: 0101010cf
OpenSSL 1.1.1l  FIPS 24 Aug 2021
GMP library version: 6.2.0
Regex library version: 2.1	(loaded: 2.1.4)
File locking: fcntl()
fseek(): fseek
ftell(): ftell
fopen(): fopen
memmem(): System's
HR timer: clock_gettime(), latency 130 ns
Total physical host memory: 15711 MiB
Available physical host memory: 12095 MiB
Terminal locale string: de_DE.UTF-8
Parsed terminal locale: UTF-8
```

frank-dittrich:
Additional segfault info:
```
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7b5a8b3 in rexgen::IteratorContainer::updateReferences(rexgen::IteratorState&) () from /usr/local/lib/librexgen.so.2.1
Missing separate debuginfos, use: dnf debuginfo-install gmp-6.2.0-7.fc35.x86_64 ocl-icd-2.3.0-2.fc35.x86_64 openssl-libs-1.1.1l-2.fc35.x86_64 zlib-1.2.11-30.fc35.x86_64
(gdb) bt
#0  0x00007ffff7b5a8b3 in rexgen::IteratorContainer::updateReferences(rexgen::IteratorState&) () from /usr/local/lib/librexgen.so.2.1
#1  0x00007ffff7b60a84 in rexgen::TopIterator::TopIterator(std::shared_ptr<rexgen::Regex>&) () from /usr/local/lib/librexgen.so.2.1
#2  0x00007ffff7b61b85 in c_regex_iterator () from /usr/local/lib/librexgen.so.2.1
#3  0x00000000006feaec in do_regex_hybrid_crack (db=db@entry=0x201f9e0 <database>, regex=0x214fff8 "\\01[23]", base_word=0x7fffffffaa80 "123456", regex_case=<optimized out>, regex_alpha=0x0) at regex.c:233
#4  0x00000000006e0155 in do_wordlist_crack (db=db@entry=0x201f9e0 <database>, name=<optimized out>, rules=0) at wordlist.c:1294
#5  0x00000000006b5979 in john_run () at john.c:1809
#6  main (argc=<optimized out>, argv=<optimized out>) at john.c:2082
(gdb) 
```

solardiz:
> This mode is defined in regex_alphabets.conf

BTW, does the `--regex` option accept either a regexp or a mode name without a way to request one of these explicitly? Ideally, we'd have different syntax to request one or the other.

frank-dittrich:
> > This mode is defined in regex_alphabets.conf
> 
> BTW, does the `--regex` option accept either a regexp or a mode name without a way to request one of these explicitly? Ideally, we'd have different syntax to request one or the other.

AFAIK, the regex mode can only be used with a word list (or with any other way to generate candidates, and nobody would use `\0` in a mode name, so IMHO there's no confusion whether the value has to be interpreted as a mode name or as a regexp. But I also wouldn't mind 2 different option names.

magnumripper:
The problem seems to be in the lib, not in our code. So does this:
```
$ ./john -stdout -w -regex='ab\d\0' 
Segmentation fault: 11
```

Same thing happens with `rexgen` utility on its own:
```
$ rexgen -f password.lst 'ab\d\0' 
Segmentation fault: 11
```
Backtrace from both cases are same.

magnumripper:
> > BTW, does the `--regex` option accept either a regexp or a mode name without a way to request one of these explicitly? Ideally, we'd have different syntax to request one or the other.
> 
> AFAIK, the regex mode can only be used with a word list (or with any other way to generate candidates, and nobody would use `\0` in a mode name, so IMHO there's no confusion whether the value has to be interpreted as a mode name or as a regexp. But I also wouldn't mind 2 different option names.

Not sure I understood you correct Frank, but the regex mode can be ran on its own without `\0`:
```
$ ../run/john -stdout -regex='deadcafe([a-f])\1'
deadcafeaa
deadcafebb
deadcafecc
deadcafedd
deadcafeee
deadcafeff
6p 0:00:00:00 0.00% 150.0p/s deadcafeff
```

magnumripper:
> BTW, does the `--regex` option accept either a regexp or a mode name without a way to request one of these explicitly?

For rules, we have the `=:` (or `::`) for telling them apart:
```
$ ./john -stdout -w -rules=:/?au -max-cand=5
Proceeding with wordlist:../run/password.lst, rules::/?au
PASSWORD
PASSWORD1
ABC123
COMPUTER
TIGGER
5p 0:00:00:00 0.34% (ETA: 17:57:03) 125.0p/s TIGGER
```
Perhaps we should change Jim's "regex modes" similarly. I think the "user base" of regex mode in JtR is small and whoever uses it are probably also power users with an ear here.

frank-dittrich:
> Not sure I understood you correct Frank, but the regex mode can be ran on its own without `\0`:
> 
> ```
> $ ../run/john -stdout -regex='deadcafe([a-f])\1'
> deadcafeaa
> deadcafebb
> deadcafecc
> deadcafedd
> deadcafeee
> deadcafeff
> 6p 0:00:00:00 0.00% 150.0p/s deadcafeff
> ```

I was just referring to Solar's distinction between a regexp and mode, so in this context, a mode is something which refers to a config file section.

Standalone --regex uses just regular expressions (without `\0`), combined with other attack modes --regex can be used with regular expressions (containing `\0`) or with modes defined in config files, their names should not contain  `\0`.

For --rules, that distinction isn't always possible. many rule definitions could also be used as a rules section name, that's why the additional `:` is required to specify rules directly on the command line.

magnumripper:
> The problem seems to be in the lib, not in our code. So does this:

This is fixed in librexgen 2.1.5

magnumripper:
> I think this hasn't worked for a long time, but it used to work several years ago and really should work, no?
> 
> ```
> $ ./john --regex=alpha:leet --stdout --wordlist=password.lst 
> Using default input encoding: UTF-8
> syntax error, unexpected end of file
> Error, invalid regex expression.  John exiting now  base_word=123456  Regex= 
> ```

This is indeed a bug in our code. It's hard to bisect as older versions of john need older versions of the lib.

frank-dittrich:
> This is indeed a bug in our code. It's hard to bisect as older versions of john need older versions of the lib.

You might even need an old build environment.
I remember trying to bisect a while ago and wasn't able to successfully build any working older version.

magnumripper:
I think we should open a separate issue for that problem, and merge this now.

solardiz:
> merge this now.

I agree, but this shows a conflict for `doc/NEWS` now and it will also clash with #4908. So perhaps merge #4908 first, then update this PR and merge it too?

PR #4908, opened 2021-12-03
configure: respect $STRINGS, fix bashisms

thesamesam:
- Respect $STRINGS from the user's environment (like we do for $AR, $CC, etc) but fall back to 'strings' like before
- Don't use Bashisms in configure (similar to #4822).


magnumripper (approved):
Looks fine to me

PR #4909, opened 2021-12-03
7z formats: Add support for Delta encoding

magnumripper:
Adds support for "Delta encoding".  This syncs us with upstream at:
https://github.com/philsmd/7z2hashcat/commit/346130d6


magnumripper:
Added a commit that implements Delta decoding in the formats as well. It's tested, but I don't have a short enough test vector to add it to the formats.

PR #4913, opened 2021-12-07
OpenCL: Improve handling of cached kernels

magnumripper:
If building from a cached binary fails, just build a new one instead of bailing out. This **does** affect DEScrypt's per-salt kernels as well.

See #3525

~I'm not 100% happy with the code (adding a struct and code in `john.c` just for this) but not sure how to make it safe & portable otherwise. At least it's simple code. I'm open for suggestions, as always.~

### DIFF
--- a/src/listconf.c
+++ b/src/listconf.c
@@ -143,7 +143,7 @@ static void listconf_list_build_info(void)
 	int gmp_major, gmp_minor, gmp_patchlevel;
 #endif
 	puts("Version: " JTR_GIT_VERSION);
-	puts("Build: " JOHN_BLD _MP_VERSION DEBUG_STRING ASAN_STRING UBSAN_STRING);
+	puts("Build: " JOHN_BLD _MP_VERSION OCL_STRING ZTEX_STRING DEBUG_STRING ASAN_STRING UBSAN_STRING);
 #ifdef SIMD_COEF_32
 	printf("SIMD: %s, interleaving: MD4:%d MD5:%d SHA1:%d SHA256:%d SHA512:%d\n",
 	       SIMD_TYPE,
--- a/src/listconf.h
+++ b/src/listconf.h
@@ -41,6 +41,18 @@
 #define UBSAN_STRING ""
 #endif
 
+#if HAVE_OPENCL
+#define OCL_STRING " OPENCL"
+#else
+#define OCL_STRING ""
+#endif
+
+#if HAVE_ZTEX
+#define ZTEX_STRING " ZTEX"
+#else
+#define ZTEX_STRING ""
+#endif
+
 #define _STR_VALUE(arg)			#arg
 #define STR_MACRO(n)			_STR_VALUE(n)
 
--- a/src/opencl_DES_bs_f_plug.c
+++ b/src/opencl_DES_bs_f_plug.c
@@ -226,7 +226,7 @@ static void modify_build_save_restore(WORD salt_val, int id_gpu, int save_binary
 		size_t program_size;
 		fclose(file);
 		program_size = opencl_read_source(kernel_bin_name, &kernel_source);
-		opencl_build_from_binary(id_gpu, program_ptr, kernel_source, program_size);
+		HANDLE_CLERROR(opencl_build_from_binary(id_gpu, program_ptr, kernel_source, program_size), "kernel build failed");
 
 		if (options.verbosity > VERB_DEFAULT)
 			fprintf(stderr, "Salt compiled from Binary:%d\n", salt_val);
--- a/src/opencl_DES_bs_h_plug.c
+++ b/src/opencl_DES_bs_h_plug.c
@@ -233,7 +233,7 @@ static void modify_build_save_restore(WORD salt_val, int id_gpu, int save_binary
 		size_t program_size;
 		fclose(file);
 		program_size = opencl_read_source(kernel_bin_name, &kernel_source);
-		opencl_build_from_binary(id_gpu, program_ptr, kernel_source, program_size);
+		HANDLE_CLERROR(opencl_build_from_binary(id_gpu, program_ptr, kernel_source, program_size), "kernel build failed");
 
 		if (options.verbosity > VERB_DEFAULT)
 			fprintf(stderr, "Salt compiled from Binary:%d\n", salt_val);
--- a/src/opencl_common.c
+++ b/src/opencl_common.c
@@ -51,7 +51,7 @@
 #include "recovery.h"
 #include "status.h"
 #include "john.h"
-#include "md5.h"
+#include "md4.h"
 #include "misc.h"
 #include "john_mpi.h"
 #include "timer.h"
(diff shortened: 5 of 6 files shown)


magnumripper:
Added a commit that cleans up `opencl_build_from_binary()` making it easier to follow. Also, I stopped disabling caching on nvidias since it does speed things up further - their cache is on another level.

Times for building raw-md5 kernel for nvidia 2080ti using version 465.19.01:
Existing valid cache | build time
-|-
none | 171.783 ms
only our own | 63.188 ms
only ComputeCache | 3.876 ms
both | 1.369 ms

While we only saved 2.5 ms with this simple kernel, it's a 65% boost - for complex kernels it might become significant.

solardiz:
> If cached kernel is older than john itself, ignore it and build a new one.

I'm not convinced this logic is what's desired. I think it wouldn't have helped in the cases I encountered in my work on #3525, or would it? For example, I was confused for a while when I returned to an old john tree to run benchmarks there and suddenly got the new code's speeds - indeed, newer cached kernels from `~/.nv` were used. Also, the changes in #3525 don't touch anything other than an OpenCL header file, so there wouldn't be a `john` rebuild after pulling that one commit - meaning we still need to ask users to explicitly clean the cache.

> While we only saved 2.5 ms with this simple kernel, it's a 65% boost - for complex kernels it might become significant.

That's great, and is desired.

BTW, in places where we need a good non-cryptographic hash/checksum, perhaps we should use MD4 rather than MD5, until possibly getting SipHash into the tree.

ghost (approved):
This is not a universal solution, but it is actually an improvement. IMO, should be merged (after a real review).

magnumripper:
> > If cached kernel is older than john itself, ignore it and build a new one.
> 
> I'm not convinced this logic is what's desired. I think it wouldn't have helped in the cases I encountered in my work on #3525, or would it? For example, I was confused for a while when I returned to an old john tree to run benchmarks there and suddenly got the new code's speeds - indeed, newer cached kernels from `~/.nv` were used. Also, the changes in #3525 don't touch anything other than an OpenCL header file, so there wouldn't be a `john` rebuild after pulling that one commit - meaning we still need to ask users to explicitly clean the cache.

You may be right - I had a couple of situations in mind but on second thought the majority of them will not be helped by this. A few will, though.

> > While we only saved 2.5 ms with this simple kernel, it's a 65% boost - for complex kernels it might become significant.
> 
> That's great, and is desired.

I'm not so sure anymore. More testing revealed it only shaves like 1.5 seconds off a full `--test` run 😢 

`time(1)` of a full `--test --form:opencl` run with all kernels already cached:
```
Current bleeding (all kernels in nvidia cache):
real    2m10.824s
user    2m1.867s
sys     0m9.287s

This PR (all kernels in our cache as well as in nvidia cache):
real    2m9.136s
user    2m0.317s
sys     0m9.172s
```
I'll try benchmarking 100 (or maybe 1000) DES salts next.

> BTW, in places where we need a good non-cryptographic hash/checksum, perhaps we should use MD4 rather than MD5, until possibly getting SipHash into the tree.

I don't mind doing that, but what would be the reason?

solardiz:
How are you able to `--test` all OpenCL formats, even cached, in just 2 minutes? I think our self-tests take longer than that.

Regarding MD4 vs. MD5, I was thinking using the simpler and faster function where the difference otherwise doesn't matter. I don't feel about this strongly, though.

magnumripper:
Oh, I meant `--test=0` - my objective was just to compare build times.

magnumripper:
I'm running 4,096 DES salts tests now. Seems it will take about one hour to build with nothing cached.

From what I can see so far there will be little gain 😢 
I first ran a test with 60 salts, and there were more like 35 ms regression.

OTOH looking at net time to build a single salt, it seems to be ~640 ms, down from ~900 ms so the speed-up could be near 30% or a total of 20 minutes for every possible salt. Depends on where the overhead is...

magnumripper:
I need to look further into this. Results for 4096 salts:
```
Current bleeding (all kernels in nvidia cache):
real    21m2.620s
user    20m58.047s
sys     0m4.539s

This PR (all kernels in our cache as well as in nvidia cache):
real    20m57.450s
user    20m52.771s
sys     0m4.615s
```
That's just five seconds improvement - not worth the added complexity and the added level of "stale cache" problems.

But again, the numbers I've seen for each kernel doesn't add up at all - I need to find out what more is taking a lot of time.

magnumripper:
@solardiz Odd fact: The number of files under `~/.nv/ComputeCache/` with all 4096 salt-kernels (plus the other needed DEScrypt kernels) cached is only 3153 (for bs_h) or 733 (for bs_f), reproducible.

Can some of the salts' kernels be identical after optimization? If not, I can't fathom this.

At first I thought nvidia was discarding LRU for reducing size of cache, but that's not it - then our added caching would show way better boost.

magnumripper:
For 60 salts, the figures were 134 after bs_h and 142 after bs_f. Now that's *more* files than kernels.

magnumripper (commented):
Here's a run after thrashing a cache file
```
../run/john -test=0 -form:raw-md5-opencl 
Device 2: NVIDIA GeForce RTX 2080 Ti
Testing: raw-MD5-opencl [MD5 OpenCL]... Binary build log: error   : Binary format for key='0', ident='' is not recognized

PASS
```
The log would be less dramatic
```
0:00:00:00 - Building kernel from cached binary
0:00:00:00 - Build from cached binary failed
0:00:00:00 - Building kernel from source and caching binary
```
I should suppress the build error log for cached files.

(EDIT: FIXED)

magnumripper:
> I need to look further into this. Results for 4096 salts:
> 
> ```
> Current bleeding (all kernels in nvidia cache):
> real    21m2.620s
> user    20m58.047s
> sys     0m4.539s
> 
> This PR (all kernels in our cache as well as in nvidia cache):
> real    20m57.450s
> user    20m52.771s
> sys     0m4.615s
> ```

> But again, the numbers I've seen for each kernel doesn't add up at all - I need to find out what more is taking a lot of time.

OK so here's some processing of logs with warm caches: Of all 4096 kernels, the average build time was 91 ms for a total of 374.38 seconds. The total net build time is just about 6 minutes. **Why is the total 20 minutes?!**

magnumripper:
> OK so here's some processing of logs with warm caches

Here's the same processing with cold caches:
The average build time was 311.7 ms, for a net total of 21 minutes. Why is the total just over one hour?!

Edit: These figures are for bs_h, not the (even slower) default of bs_f

solardiz:
Should we possibly have a command-line option to force kernel rebuild - ignore our cache? Can we also force ignoring of NVIDIA's cache?

ghost:
> Should we possibly have a command-line option to force kernel rebuild - ignore our cache? Can we also force ignoring of NVIDIA's cache?

Yes (it is 100% is JtR code, we can rebuild from source rather than build from previously cached binary version) and yes (https://developer.nvidia.com/blog/cuda-pro-tip-understand-fat-binaries-jit-caching/), setting CUDA_CACHE_DISABLE to 1.

magnumripper:
So we could have something like `--opencl-ignore-cache` that would disable our own caches and setenv `CUDA_CACHE_DISABLE`. Not a bad idea.

PR #4918, opened 2021-12-07
tezos-opencl: Compute ed25519_publickey() on device

solardiz:
Provides a 3x speedup on GTX 1080 when running just 1 CPU thread.

While at it, also set HASH_LOOPS to a divisor of ITERATIONS and simplify the logic around these.


solardiz:
> tezos-opencl: Set HASH_LOOPS to 512, which is a divisor of ITERATIONS

BTW, as I understand what happens with our PBKDF2 code under the hood is we get 513+512+512+511, which is good enough.

run/opencl/tezos_kernel.cl
```diff
@@ -171,3 +176,21 @@ __kernel void pbkdf2_sha512_tezos_kernel(__global const pass_t *inbuffer,
 		state[idx].W[i] = tmp_out[i];
 	}
 }
+
+#undef z /* was defined by opencl_sha2.h and used in SHA256_ZEROS, but conflicts with the below */
+#include "ed25519-donna/ed25519-donna.c"
```
magnumripper, line 181: Perhaps we should end `opencl_sha2.h` with `#undef z` and quite possibly more of such. That's a separate issue though.

magnumripper (approved):
Looks solid. I didn't really look at the ed25519-donna files.

run/opencl/tezos_kernel.cl
solardiz: > Perhaps we should end `opencl_sha2.h` with `#undef z`

This would break uses of the `SHA*_ZEROS` macros from outside of that header file, and we have many such uses. I think we should instead rename `z` and `Z` to something namespaced to that file, e.g. `SHA256_Z` and `SHA512_Z`. And yes, that change should be separate from this PR. I might do it along with my pending changes for #3525.

Issue #2349, opened 2016-11-29
"scenario" mode?


smarteam:
This is a feature request

Imagine a situation with a series of checks needed to be done on a single file with a big set of hashes or a series of files with similar checks. For example ranged by difficulty/result level to check compliance with password policy, check against previous findings (with modification rules) and also against predefined wordlists (like known public leaks). Like one big wordlist with different sets of hashes (with varying algorithms) or one big hashlist with lots of options to try (like different masks, rules etc.).

I guess it'll be handy to eliminate those re-loads and re-deduplication checks that will be made if launching in an ordinary way step-by-step with separate john process each time. By doing the resource allocation (loading/placing in memory) and after that getting the "scenario" with a list of commands/options (from a file for example) can save the time.

Something like the thing that is made during (1) (2) (3) steps of "no options" john run but more versatile and human-controlled.

john -w=verybigdict.txt -scenario=test1.list
with test1.list containing:
-format=hash1 system1.pass
-format=hash2 system2.pass
-format=hash3 system3.pass
...

john checkme.pass -scenario=test2.list
with test2.list contaning:
-mask=XXXXXXX
-mask=YYYYYYYYY
-w=previouslyfound.txt -rules=popular_modifications
-w=verybigleak.txt
...

solardiz:
Yes, it's been planned as an enhancement/replacement of the default hard-coded "batch mode" (those 3 steps), but I never got around to it.

magnumripper:
> john checkme.pass -scenario=test2.list
with test2.list contaning:
-mask=XXXXXXX
-mask=YYYYYYYYY
-w=previouslyfound.txt -rules=popular_modifications
-w=verybigleak.txt

The above can be handled with some extensions to existing batch mode. I like this idea and I have looked into it before. Should be fairly easy to implement.

> john -w=verybigdict.txt -scenario=test1.list
with test1.list containing:
-format=hash1 system1.pass
-format=hash2 system2.pass
-format=hash3 system3.pass

The above would be tricky. We'd need to re-load hashes inbetween, and we wouldn't gain anything compared to normal scripting or a `john this && john that` one-liner. I'm not sure we should support it.

smarteam:
>>john -w=verybigdict.txt -scenario=test1.list
>>with test1.list containing:
>>-format=hash1 system1.pass
>>-format=hash2 system2.pass
>>-format=hash3 system3.pass
>
[...]

PR #4921, opened 2021-12-08
Wordlist mode: Support --wordlist=<fifo>

magnumripper:
This trivially handles fifo's by using the existing code path for  --pipe.

### DIFF
--- a/src/wordlist.c
+++ b/src/wordlist.c
@@ -93,6 +93,8 @@ static struct rpp_context *rule_ctx;
 static char *word_file_str, **words;
 static int64_t nWordFileLines;
 
+static int file_is_fifo;
+
 static void save_state(FILE *file)
 {
 	fprintf(file, "%d\n%" PRId64 "\n%" PRId64 "\n",
@@ -167,7 +169,7 @@ static int restore_state(FILE *file)
 	if (restore_rule_number())
 		return 1;
 
-	if (word_file == stdin) {
+	if (word_file == stdin || file_is_fifo) {
 		restore_line_number();
 	} else
 	if (!nWordFileLines) {
@@ -221,7 +223,7 @@ static void fix_state(void)
 	rec_rule = rule_number;
 	rec_line = line_number;
 
-	if (word_file == stdin)
+	if (word_file == stdin || file_is_fifo)
 		rec_pos = line_number;
 	else
 	if (!mem_map && !nWordFileLines &&
@@ -240,7 +242,7 @@ void wordlist_hybrid_fix_state(void)
 	hybrid_rec_rule = rule_number;
 	hybrid_rec_line = line_number;
 
-	if (word_file == stdin)
+	if (word_file == stdin || file_is_fifo)
 		hybrid_rec_pos = line_number;
 	else
 	if (!mem_map && !nWordFileLines &&
@@ -266,7 +268,7 @@ static double get_progress(void)
 	if (progress)
 		return progress;
 
-	if (!word_file || word_file == stdin)
+	if (!word_file || word_file == stdin || file_is_fifo)
 		return -1;
 
 	if (nWordFileLines) {
@@ -550,6 +552,24 @@ void do_wordlist_crack(struct db_main *db, const char *name, int rules)
 	}
 
 	if (name) {
+		struct stat st;
+
+		if (!(word_file = jtr_fopen(path_expand(name), "rb")))
+			pexit(STR_MACRO(jtr_fopen)": %s", path_expand(name));
+
+		if (fstat(fileno(word_file), &st))
+			pexit("fstat");
+
+		file_is_fifo = ((st.st_mode & S_IFMT) == S_IFIFO);
+
+		if (john_main_process)
+			log_event("- %s %s: %.100s", loopBack ? "Loopback pot" : "Wordlist",
+			          file_is_fifo ? "FIFO" : "file",
+			          path_expand(name));
+	} else
+		file_is_fifo = 0;
+
+	if (name && !file_is_fifo) {
 		char *cp, csearch;
 		int64_t ourshare = 0;
 #ifdef HAVE_MMAP
@@ -560,13 +580,6 @@ void do_wordlist_crack(struct db_main *db, const char *name, int rules)
 		if (mmap_max == -1)
 			mmap_max = 1 << 20;
 #endif
-		if (!(word_file = jtr_fopen(path_expand(name), "rb")))
-			pexit(STR_MACRO(jtr_fopen)": %s", path_expand(name));
-		if (john_main_process)
-			log_event("- %s file: %.100s",
-			          loopBack ? "Loopback pot" : "Wordlist",
-			          path_expand(name));
-
 		jtr_fseek64(word_file, 0, SEEK_END);
 		if ((file_len = jtr_ftell64(word_file)) == -1)
 			pexit(STR_MACRO(jtr_ftell64));
@@ -854,7 +867,9 @@ skip:
  * in --stdin mode, we can NOT perform rules, due to we can not fseek stdin in
  * most OS's.
  */
-		word_file = stdin;
+		if (!file_is_fifo)
+			word_file = stdin;
+
 		if (options.flags & FLG_STDIN_CHK) {
 			log_event("- Reading candidate passwords from stdin");
 		} else {
@@ -1363,7 +1378,7 @@ next_rule:
 			}
 
 			line_number = 0;
-			if (!nWordFileLines && word_file != stdin) {
+			if (!nWordFileLines && word_file != stdin && !file_is_fifo) {
 				if (mem_map)
 					map_pos = mem_map;
 				else


magnumripper:
@solardiz I recall there may be problems with this, can't remember the details but I figure it can be a race between `fgetl()` and a growing buffer so lines could be truncated. OTOH I'm thinking the same problems should exist with --pipe, or am I wrong?

Anyway If there are caveats, we could simply add a warning. The pipe buffer is huge so problems are hopefully rare.

solardiz:
> can be a race between `fgetl()` and a growing buffer so lines could be truncated.

Unless we set the fd to non-blocking, it should block on an underlying `read()` syscall - not give false EOF.

src/wordlist.c
```diff
@@ -550,6 +552,24 @@ void do_wordlist_crack(struct db_main *db, const char *name, int rules)
...
+		struct stat st;
+
+		if (!(word_file = jtr_fopen(path_expand(name), "rb")))
+			pexit(STR_MACRO(jtr_fopen)": %s", path_expand(name));
+
+		if (fstat(fileno(word_file), &st))
+			pexit("fstat");
+
+		if ((st.st_mode & S_IFMT) == S_IFIFO)
+			file_is_fifo = 1;
```
solardiz: I think we should reset `file_is_fifo` to 0 on new calls to `do_wordlist_crack`. Right now it's just one such call per `john` invocation, but that can change.

solardiz (approved):
I'm OK with the idea.

solardiz:
@AlekseyCherepanov I recall you were using `john` with FIFOs a lot, so maybe you have input? If things worked for you as-is, (why) is this PR needed?

magnumripper:
Without this:
```
$ ./john -stdout -w:fifo 
ftell: Illegal seek
```

magnumripper:
I see some problems but they are inherent to the pipe mode code path, I recall they have annoyed me when using `--pipe`:
The buffer is huge and we wont process a word until we get a full buffer. That's great if our feed is fast (makes rules more efficient) but bad otherwise. I'll address that later.

src/wordlist.c
```diff
@@ -550,6 +552,23 @@ void do_wordlist_crack(struct db_main *db, const char *name, int rules)
...
 	if (name) {
+		struct stat st;
+
+		if (!(word_file = jtr_fopen(path_expand(name), "rb")))
+			pexit(STR_MACRO(jtr_fopen)": %s", path_expand(name));
+
+		if (fstat(fileno(word_file), &st))
+			pexit("fstat");
+
+		file_is_fifo = ((st.st_mode & S_IFMT) == S_IFIFO);
```
solardiz, line 563: That's still not it - we need `file_is_fifo` to be reset to 0 if `!name` as well. So just set it to 0 before the `if (name)`.

magnumripper: Oh, right

magnumripper: There'll likely be lots of other things going wrong in wordlist that day though 🤣

AlekseyCherepanov:
From usability standpoint, it should be great! Some users would prefer `--wordlist=/dev/stdin` over `--pipe`. Also input for keys is left free explicitly with `--wordlist=<(sometool)` instead of `--pipe < <(sometool)`.

I did not know about the race. I used `--pipe` in situations where I could not check results. I did not expect problems, but...

Yes, it seems problematic. I can get a truncated line:
```
$ perl -le 'print $a++ . "a" x 100 while 1' | ./run/john --pipe --stdout | perl -lne 'print $. if $. % 1000000 == 0; print "wrong $. >$_<" if !/^\d+a{100}$/'
[...]
39000000
wrong 39505289 >39505288aaaaaaaaaaaaaaa<
wrong 39505290 ><
40000000
[...]
59000000
wrong 59206966 ><
wrong 59206967 ><
60000000
[...]
```
(`$.` is current line number in Perl. I print it to see that input passes through. It is counted from 1.)

The numbers repeat on reruns, so it seems deterministic.

```
$ ./run/john
John the Ripper 1.9.0-jumbo-1+bleeding-5d450f267 2021-12-03 21:27:35 +0100 [linux-gnu 64-bit x86_64 SSE4.2 AC]
```

magnumripper:
> @AlekseyCherepanov I recall you were using `john` with FIFOs a lot, so maybe you have input? If things worked for you as-is, (why) is this PR needed?

I found https://www.openwall.com/lists/john-users/2012/06/29/5 which is "a bit dated", x-)

EDIT: mid-air collision with Aleksey's answer :)

magnumripper:
> Some users would prefer `--wordlist=/dev/stdin` over `--pipe`

What do you mean by this? Should `--wordlist=/dev/stdin` be something slightly different than `--stdin`?

magnumripper:
@AlekseyCherepanov please open an issue for your findings

AlekseyCherepanov:
> > Some users would prefer `--wordlist=/dev/stdin` over `--pipe`
>
> What do you mean by this? Should `--wordlist=/dev/stdin` be something slightly different than `--stdin`?

I thought it is going to be like `--pipe`, i.e. it will allow `--rules`, because `--wordlist` allows rules while `--stdin` does not allow rules.

magnumripper:
Well it doesn't work even after this PR.
```
$ ./john -stdout -w:/dev/stdin
Error, dictionary file is empty
```
As you probably understand, the reason `-stdin` doesn't work with rules is it's not seekable, so after processing first rule we can't seek to beginning of word list and run all words through next rule.

I don't see much point in making it work. Just use `--pipe`!

AlekseyCherepanov:
Hm, I did not think about `--wordlist=/dev/stdin` for interactive input. I thought about using it with a regular pipe.

A pipe works well:
```
$ echo asdf | ./run/john --stdout --wordlist=/dev/stdin
asdf
1p 0:00:00:00 100.00% (2021-12-09 16:33) 20.00p/s asdf
```

Somehow I get other message for input from tty:
```
$ ./run/john --stdout --wordlist=/dev/stdin
ftell: Illegal seek
```

```
$ ls -l /dev/stdin
lrwxrwxrwx 1 root root 15 Oct  7 12:02 /dev/stdin -> /proc/self/fd/0
$ ls -l /proc/self/fd/0
lrwx------ 1 a a 64 Dec  9 16:33 /proc/self/fd/0 -> /dev/pts/13
$ ls -l /dev/pts/13
crw--w---- 1 a tty 136, 13 Dec  9 16:33 /dev/pts/13
$ file /dev/pts/13
/dev/pts/13: character special (136/13)
```

```
$ ./run/john
John the Ripper 1.9.0-jumbo-1+bleeding-ff16ecb93 2021-12-08 20:03:10 +0100 [linux-gnu 64-bit x86_64 SSE4.2 AC]
```

Actually `cat |` can be used for interactive input:
```
$ cat | ./run/john --stdout --wordlist=/dev/stdin --rules=':Az"[12]"'
aaa   <- input
bbb   <- input, then I typed ^D
aaa1
bbb1
aaa2
bbb2
4p 0:00:00:03 100.00% (2021-12-09 16:45) 1.061p/s bbb2
```

For me tricks with `/dev/stdin` come from situations when tool's default use of stdin does not fit situation. For instance, `grep` reads content to filter from stdin, while it can read regexps. Actually `grep` supports `-` to read from stdin:
```
$ echo qwerasdfqwer > t
$ echo asdf | grep -nf /dev/stdin t
1:qwerasdfqwer
$ echo asdf | grep -nf - t
1:qwerasdfqwer
```

AlekseyCherepanov:
I updated john and tried `--wordlist=/dev/stdin` too like that:
```
$ perl -le 'print $a++ . "a" x 100 while 1' | ./run/john --stdout --wordlist=/dev/stdin --rules-stack=':' | perl -lne 'print $. if $. % 1000000 == 0; print "wrong $. >$_<" if !/^\d+a{100}$/'
```

`--stdout --wordlist=/dev/stdin` has the problem

`--stdout --wordlist=/dev/stdin --rules=':'` -- no problem (87M)

`--stdout --wordlist=/dev/stdin --rules=':l'` -- no problem (137M)

`--stdout --wordlist=/dev/stdin --rules-stack=':'` has the problem

`--stdout --wordlist=/dev/stdin --rules-stack=':l'` has the problem

AlekseyCherepanov:
I tried with a fifo too:
```
$ mkfifo fifo
$ perl -le 'print $a++ . "a" x 100 while 1' > fifo & ./run/john --stdout --wordlist=fifo | perl -lne 'print $. if $. % 1000000 == 0; print "wrong $. >$_<" if !/^\d+a{100}$/'
```

It has the problem too.

(After exit from john by ^C background perl exited with "Broken pipe". Bash showed message about it after next command. Convenient!)

AlekseyCherepanov:
Sorry, I posted onto wrong page again. The last 2 comments should go to #4923.

AlekseyCherepanov:
Cracking HIBP during training minicontest in 2017, I used fifos a lot to avoid reloading hashes.

The worker loaded the hashes and waiting for attacks:
```
$ john --stdin [...] < internal_fifo
```

On other terminal, "communicator" to merge attacks into one sequence (`date` shows end of attack):
```
$ while true; do cat < fifo; date 1>&2; done > internal_fifo
```

On another terminal, running an attack (one at a time):
```
$ john ... --stdout > fifo
```

The idea seemed cool at the time. Indeed it allowed me to run small attacks quickly saving time of loading. But it was not very convenient because it did not provide feedback on efficiency of attacks well.

Also this experience was behind my comment about turning john into a server:
https://github.com/openwall/john/issues/2349#issuecomment-543152059

PR #4933, opened 2021-12-11
tezos-opencl key transfer optimization

solardiz:
Only transfer keys when changed (not re-transfer when merely switching to next salt). Only transfer keys up to length 48, not 110. 48 was the maximum supported in the code anyway, but 110 was also used in some places including for transfer size.

Also, limit mnemonics to a maximum of 128 characters in valid() (was 512) and use that for the allocation size in the CPU format. The OpenCL format had it that way anyway (in the host code, and effectively at 112 in the device code - now increased), and could probably be made to crash due to the higher limit in valid().

### DIFF
--- a/run/opencl/tezos_kernel.cl
+++ b/run/opencl/tezos_kernel.cl
@@ -130,18 +130,20 @@ __kernel void pbkdf2_sha512_tezos_init(__global const pass_t *inbuffer,
 	ulong tmp_out[8];
 	uint i;
 	uint idx = get_global_id(0);
-	int passlen = gsalt->mnemonic_length;
 	uint rounds = gsalt->pbkdf2.rounds;
-	int saltlen;
+	uint saltlen;
 	union {
-		uchar bytes[8 + sizeof(gsalt->pbkdf2.salt) + 48 /* REAL_PLAINTEXT_LENGTH */];
-		ulong data[(8 + sizeof(gsalt->pbkdf2.salt) + 48 + 7) / 8];
+		uchar bytes[8 + sizeof(gsalt->pbkdf2.salt) + PLAINTEXT_LENGTH];
+		ulong data[(8 + sizeof(gsalt->pbkdf2.salt) + PLAINTEXT_LENGTH + 7) / 8];
 	} salt;
-
 	union {
-		uchar bytes[PLAINTEXT_LENGTH];
-		ulong data[(PLAINTEXT_LENGTH + 7) / 8];
+		uchar bytes[128];  // this large to also accommodate mnemonic
+		ulong data[128 / 8];
 	} pass;
+
+	uint passlen = inbuffer[idx].length;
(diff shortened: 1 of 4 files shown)


solardiz:
I also make this change (and similar for salt):

```diff
	memcpy_macro(pass.bytes, gsalt->mnemonic, passlen);  // actual password
-	for (int i = passlen; i < passlen + (8 - passlen % 8); i++)  // zeroize buffer correctly
-            pass.bytes[i] = 0;
+	if (passlen % 8)
+		for (int i = passlen; i < passlen + (8 - passlen % 8); i++)  // zeroize buffer correctly
+			pass.bytes[i] = 0;
```

This is because with the original code I'd have to increase the buffer size to 136 to accommodate a mnemonic of exactly 128 as allowed by `valid()`. I assume that the zeroization there is only needed for the last non-full 64-bit word, and is not needed if the last 64-bit word is full. @magnumripper I'd appreciate your review there.

solardiz:
> with the original code I'd have to increase the buffer size to 136 to accommodate a mnemonic of exactly 128 as allowed by `valid()`.

Writing this made me realize... While `tezos_valid()` disallowed only greater-than exact buffer sizes, `tezos_get_salt()` then used `strcpy()` on those strings into those buffers inside the struct. So we could overflow them, and then overwrite the NULs. If we fix that by changing the limit to 127 (with space for the NUL), then I could as well have avoided that change in the kernel.

solardiz:
Upon a closer look, `tezos_kernel.cl` makes assumptions about things fitting in a single SHA-512 block. This is where the mnemonic length limit of 128 comes from. That's actually problematic, because the mnemonics are 15 words, and even with the BIP-0039 English list the words can be up to 8 characters each (with some non-English lists, they can even be longer - especially in terms of bytes rather than characters). So for English the worst case is `15*8+14 = 134`, and that doesn't fit. With a limit of 128 or 127, we're betting that the average word length in a mnemonic isn't close to maximum. I think we may bet, but indeed we should refuse to process inputs that we don't actually support.

Further, there's also a limit on combined e-mail address and password length. We also don't check for it anywhere. We should, such as by refusing to load "hashes" with e-mail address length that combined with the maximum supported password length would not fit.

magnumripper:
> I assume that the zeroization there is only needed for the last non-full 64-bit word, and is not needed if the last 64-bit word is full. @magnumripper I'd appreciate your review there.

Correct. 64-bit memcpy was used so the last word will contain trailing garbage.

Many formats do something like that on host side, eg raw-md5-opencl:
```
	while (len > 4) {
		saved_plain[key_idx++] = *key++;
		len -= 4;
	}
	if (len)
		saved_plain[key_idx++] = *key & (0xffffffffU >> (32 - (len << 3)));
```
A 64-bit variant of that last line's bit twiddling could be used instead of the 8-bit for loop in tezos, but it wouldn't matter at all for performance there.

Issue #4934, opened 2021-12-12
Review PBKDF2_64_MAX_SALT_SIZE


solardiz:
We have this in `pbkdf2_hmac_common.h`:

```c
#define PBKDF2_32_MAX_SALT_SIZE            179 /* 3 limb md4/md5/sha1/sha256 max when 4 byte loop counter is appended */
#define PBKDF2_64_MAX_SALT_SIZE            107 /* 1 limb sha512 max when 4 byte loop counter is appended */
```

I understand the value for `PBKDF2_32_MAX_SALT_SIZE`, it's `3*64-8-4-1 = 179`, where 8 is for encoding the message length, 4 is for encoding PBKDF2 output block number (1), and 1 for the 0x80.

I don't understand the value for `PBKDF2_64_MAX_SALT_SIZE`. Applying the same formula, we'd have `128-8-4-1 = 115`. Indeed, that's the value I've just used in my `opencl/tezos_kernel.cl` update.

Is there any particular reason for the 107, or should we change it to 115? Ideally, we'd actually test values in that range to make sure.

solardiz:
Oh, I overlooked that for SHA-512 the message length is encoded as a 128-bit value, not 64-bit. That's ridiculous, but but that's how it is. So no issue on our part here, and I'll probably want to revise that 115 to 107 in `opencl/tezos_kernel.cl` (although this doesn't really matter while we don't propagate the error; either value is good enough to prevent buffer overflow).

solardiz:
On a related note, looks like these limits don't apply to our scalar implementations - we just pass `SL` (salt length) into `*_Update()`.

PR #4935, opened 2021-12-12
Tezos formats: Warn about over-long PBKDF2 salt lengths

solardiz:
These can result from over-long combination(s) of e-mail address and candidate password, so that's what we say in the warning messages.

Edit: see also #4934.

### DIFF
--- a/src/tezos_fmt_plug.c
+++ b/src/tezos_fmt_plug.c
@@ -36,6 +36,7 @@ john_register_one(&fmt_tezos);
 #include "blake2.h"
 #include "tezos_common.h"
 #define PBKDF2_HMAC_SHA512_VARYING_SALT 1
+#include "pbkdf2_hmac_common.h"
 #include "pbkdf2_hmac_sha512.h"
 
 #define FORMAT_NAME             "Tezos Key"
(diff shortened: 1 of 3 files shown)


src/tezos_fmt_plug.c
```diff
@@ -141,6 +146,17 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 			sin[i] = (unsigned char*)salt[i];
 			pout[i] = seed[i].seed;
 			slens[i] = strlen(salt[i]);
+			static int warned;
```
magnumripper: I'd personally put the declaration outside the parallel for (at, say, line 109) just to be sure.

magnumripper: I meant line 113 (new code)

solardiz: I think there's no difference, but OK - I moved that line now.

solardiz:
Had to add `&& !self_test_running` to the check in the CPU format - apparently, our self-test attempts very long candidates?

Didn't have to add it to the OpenCL format - apparently, because of its currently lower `PLAINTEXT_LENGTH`.

magnumripper:
> Had to add `&& !self_test_running` to the check in the CPU format - apparently, our self-test attempts very long candidates?

Sure. For example, max. length candidates are filled to all indeces and then read back to check for busted buffers.

PR #4937, opened 2021-12-12
Incremental mode: Correct log message for --max-length

solardiz:
This removes erroneous claim "for this hash type" when the actual limit was possibly `--max-length`.

Similar change might also be needed for Markov and PRINCE, and for uses of `--min-length`, but those look trickier to me, so I don't bother. Someone else might.


magnumripper (commented):
I have no time right now but there's `options.req_maxlength` and `options.force_maxlength` but I can't remember why, or the difference, right now.

Then there's `options.eff_maxlength` which should be whatever the effective max. length we actually ended up with, all things considered. But that's not what you'd use here.

solardiz:
I use `options.force_maxlength` because that's what the `if` just below my changes used. Essentially, it correctly suppressed the wrong message from being printed to `stderr`, but we missed the same suppression for the logging. This corrects that.

magnumripper:
Should be fine then

magnumripper (approved):

PR #4942, opened 2021-12-14
cisco2john: Ensure "password 7" obfuscation is properly decoded

magnumripper:
It turns out our nicked-from-the-net function for deobfuscating cisco "password 7" strings was incomplete (or not up to date).  We now hand it off to a perl lib that hopefully gets updated when needed.

### DIFF
--- a/run/cisco2john.pl
+++ b/run/cisco2john.pl
@@ -5,19 +5,19 @@
 #
 # Usage Examples:
 #
-#   $ ./cisco2john.pl cisco.config >cisco.in
-#   #!comment: Found type 7 passwords:
-#   companysecret
-#   test
+#	$ ./cisco2john.pl cisco.config >cisco.in
+#	#!comment: Found type 7 passwords:
+#	companysecret
+#	test
 #
 # (because of that output, we re-run it and save stderr to its own file)
-#   $ ./cisco2john.pl cisco.conf >cisco.in 2>cisco.seed
+#	$ ./cisco2john.pl cisco.conf >cisco.in 2>cisco.seed
 #
-#   $ cat cisco.in
-#   enable_secret_level_2:5e884898da28047151d0e56f8dc6292773603d0d6aabbdd62a11ef721d1542d8
-#   enable_secret:$1$4C5N$JCdhRhHmlH4kdmLz.vsyq0
+#	$ cat cisco.in
+#	enable_secret_level_2:5e884898da28047151d0e56f8dc6292773603d0d6aabbdd62a11ef721d1542d8
+#	enable_secret:$1$4C5N$JCdhRhHmlH4kdmLz.vsyq0
 #
-#   $ ./john -wo:cisco.seed -rules cisco.in
+#	$ ./john -wo:cisco.seed -rules cisco.in
 #
 # Credits:
 #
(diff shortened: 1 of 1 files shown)


solardiz:
I think it's preferable to fix the decoder embedded in here - shouldn't be hard, or is it? Otherwise we probably end up introducing dependency on having to retrieve a module instead of working with a bare Perl install out of the box. Not great, especially for a feature that is not always needed.

While we're at it, we should also update the example `john` command that this script prints - drop the `-format:md5` from there.

magnumripper:
New version. Function now nicked almost as-is from Crypt::Cisco

magnumripper:
Indentation was a real mess of tabs and spaces so I fixed that in a separate commit

run/cisco2john.pl
```diff
@@ -89,32 +89,42 @@ sub unique
...
-	    }
-	    if ($dp =~ m/[\x00-\x19]/) {
-		    $dp = unpack("H*", $dp);
-	    }
-	    return $dp;
+sub cisco_decrypt {
+    ##################################################
+    # Nicked late 2021 from CPAN, Crypt::Cisco
+    # Copyright (C) Michael Vincent 2010, 2017
+    # www.VinsWorld.com
```
solardiz: Perhaps we also need to mention the CPAN code's license, not only copyright.

magnumripper: OK... it says "This software is released under the same terms as Perl itself." 

https://dev.perl.org/licenses/

run/cisco2john.pl
```diff
@@ -44,188 +44,185 @@
...
+	my $estr;
+	if ( $Alphabets{$type} =~ /^(.{64})(.?)$/s ) { $str=$1; $estr=quotemeta($1); $pad=$2; } else { return undef; }
+	$data =~ s/[^$estr]//gs;
+
+	# Force Padding
+	$data .= $pad x (3-(((length($data)+3) % 4)));
+	$data=~s|(.)(.)(.?)(.?)|
+	  chr(((index($str,$1)<<2)&252)+((index($str,$2)>>4)&3)).	#{six bits from first with two bits from the second
+	  (($3 ne $pad)?chr(((index($str,$2)<<4)&240)+((index($str,$3)>>2)&15)):"").	#{last 4 bits from second with four bits from third unless third is pad
+	  (($4 ne $pad)?chr(((index($str,$3)<<6)&192)+((index($str,$4))&63)):"")	#{last 2 bits from third with six bits from the forth unless forth is pad
```
solardiz, line 80: Somehow this adds `{` characters to the beginning of the 3 comments. Doesn't look like an indentation-only fix to me.

run/cisco2john.pl
solardiz: So we should probably at least copy that statement.

run/cisco2john.pl
magnumripper: Wow, can't believe I missed that. So when I fix that, what should I do with that licensing thing? Can I just add what he did as in "This software is released under the same terms as Perl itself. If you don't know what that means visit http://perl.com/."

solardiz: We can do that. Please use https://dev.perl.org/licenses/ for the URL. We can also add from there:

```
This is free software; you can redistribute it and/or modify it under the terms of either:

a) the GNU General Public License as published by the Free Software Foundation; either version 1, or (at your option) any later version, or

b) the Perl "Artistic License".
```

PR #4943, opened 2021-12-14
OpenCL formats: Ensure we profile when autotuning

### LINKED ISSUE #4932: Review and make consistent the new_keys logic in OpenCL formats
In most OpenCL formats, we transfer the keys to device `if (ocl_autotune_running || new_keys)`. In some, the `ocl_autotune_running ||` part of the condition is missing - should we add it? If we do, perhaps we need to see the effect it has on auto-tuning and the resulting speeds for those formats.

```$ grep 'if (.*new_keys' opencl_*.c
opencl_7z_fmt_plug.c:   if (ocl_autotune_running || new_keys) {
opencl_androidbackup_fmt_plug.c:        if (ocl_autotune_running || new_keys) {
opencl_bitcoin_fmt_plug.c:      if (new_keys) {
opencl_blockchain_fmt_plug.c:   if (ocl_autotune_running || new_keys) {
opencl_cryptosafe_fmt_plug.c:   if (new_keys || ocl_autotune_running) {
opencl_dashlane_fmt_plug.c:     if (ocl_autotune_running || new_keys) {
opencl_dmg_fmt_plug.c:  if (ocl_autotune_running || new_keys) {
opencl_encfs_fmt_plug.c:        if (ocl_autotune_running || new_keys) {
opencl_enpass_fmt_plug.c:       if (ocl_autotune_running || new_keys) {
opencl_ethereum_presale_fmt_plug.c:     if (new_keys || ocl_autotune_running || global_work_size > keys_done) {
opencl_iwork_fmt_plug.c:        if (ocl_autotune_running || new_keys) {
opencl_keystore_fmt_plug.c:     if (new_keys || ocl_autotune_running) {
opencl_krb5_asrep_aes_fmt_plug.c:       if (ocl_autotune_running || new_keys) {
opencl_krb5pa-md5_fmt_plug.c:   if (set_new_keys || ocl_autotune_running) {
opencl_krb5pa-sha1_fmt_plug.c:  if (ocl_autotune_running || new_keys) {
opencl_lastpass_cli_fmt_plug.c: if (new_keys || ocl_autotune_running) {
opencl_lastpass_fmt_plug.c:     if (new_keys || ocl_autotune_running) {
opencl_md5crypt_fmt_plug.c:     if (new_keys)
opencl_mscash_fmt_plug.c:       if (set_new_keys || ocl_autotune_running) {
opencl_ntlmv2_fmt_plug.c:       if (set_new_keys || ocl_autotune_running) {
opencl_office_fmt_plug.c:       if (ocl_autotune_running || new_keys) {
opencl_oldoffice_fmt_plug.c:    if (new_keys || ocl_autotune_running) {
opencl_openbsdsoftraid_fmt_plug.c:      if (ocl_autotune_running || new_keys) {
opencl_pbkdf2_hmac_md4_fmt_plug.c:      if (ocl_autotune_running || new_keys) {
opencl_pbkdf2_hmac_md5_fmt_plug.c:      if (ocl_autotune_running || new_keys) {
opencl_pbkdf2_hmac_sha1_fmt_plug.c:     if (ocl_autotune_running || new_keys) {
opencl_pem_fmt_plug.c:  if (ocl_autotune_running || new_keys) {
opencl_phpass_fmt_plug.c:       if (new_keys) {
opencl_rawsha512_gpl_fmt_plug.c:        if (new_keys && key_idx > offset)
opencl_rawsha512_gpl_fmt_plug.c:        if (new_keys && mask_int_cand.num_int_cand > 1) {
opencl_sha1crypt_fmt_plug.c:    if (ocl_autotune_running || new_keys) {
opencl_sha256crypt_fmt_plug.c:  if (new_keys) {
opencl_sha512crypt_fmt_plug.c:  if (new_keys) {
opencl_solarwinds_fmt_plug.c:   if (ocl_autotune_running || new_keys) {
opencl_strip_fmt_plug.c:        if (new_keys || ocl_autotune_running) {
opencl_telegram_fmt_plug.c:     if (ocl_autotune_running || new_keys) {
opencl_vmx_fmt_plug.c:  if (ocl_autotune_running || new_keys) {
opencl_zip_fmt_plug.c:  if (new_keys || ocl_autotune_running) {
```

magnumripper:
If using 'new_keys' optimization, ensure we call that code path while
autotuning or the event may end up uninitialized and skew tuning.

Also *adds* new_keys optimization to a few formats (there seem to be more).

Also just fixes indentation for a couple of formats that did the right
thing otherwise (they don't use shared autotune though, so not quite in
scope of #4932).

Closes #4932

### DIFF
--- a/src/opencl_sha256crypt_fmt_plug.c
+++ b/src/opencl_sha256crypt_fmt_plug.c
@@ -573,7 +573,7 @@ static int crypt_all(int *pcount, struct db_salt *_salt)
 		output_hashes = calculated_hash;
 	}
 
-	if (new_keys) {
+	if (ocl_autotune_running || new_keys) {
 		// sort passwords by length
 		if (bitmap_of_lens & (bitmap_of_lens - 1)) {
 			if (count > indices_size) {
--- a/src/opencl_7z_fmt_plug.c
+++ b/src/opencl_7z_fmt_plug.c
@@ -323,10 +323,10 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 		any_cracked = 0;
 	}
 
+	global_work_size = GET_NEXT_MULTIPLE(count, local_work_size);
+
 	if (ocl_autotune_running || new_keys) {
 		int i;
-
-		global_work_size = GET_NEXT_MULTIPLE(count, local_work_size);
 
 		// Copy data to gpu
 		BENCH_CLERROR(clEnqueueWriteBuffer(queue[gpu_id], mem_in, CL_FALSE, 0,
@@ -353,9 +353,9 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 		BENCH_CLERROR(clEnqueueNDRangeKernel(queue[gpu_id], sevenzip_final, 1,
 			NULL, &global_work_size, lws, 0, NULL, multi_profilingEvent[3]),
 			"Run final loop kernel");
-	}
 
-	new_keys = 0;
+		new_keys = 0;
+	}
 
 	if (sevenzip_trust_padding || sevenzip_salt->type == 0x80) {
 		// Run AES kernel (only for truncated hashes)
--- a/src/opencl_bitcoin_fmt_plug.c
+++ b/src/opencl_bitcoin_fmt_plug.c
@@ -384,7 +384,7 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 	in_size = sizeof(password_t) * global_work_size;
 	cracked_size = sizeof(*cracked) * global_work_size;
 
-	if (new_keys) {
+	if (ocl_autotune_running || new_keys) {
 		// Copy data to gpu
 		BENCH_CLERROR(clEnqueueWriteBuffer(queue[gpu_id], mem_in, CL_FALSE, 0,
 			in_size, saved_key, 0, NULL, multi_profilingEvent[0]),
--- a/src/opencl_pbkdf2_hmac_sha256_fmt_plug.c
+++ b/src/opencl_pbkdf2_hmac_sha256_fmt_plug.c
@@ -49,6 +49,7 @@ static cl_int cl_error;
 static cl_mem mem_in, mem_out, mem_salt, mem_state;
 static cl_kernel split_kernel, final_kernel;
 static struct fmt_main *self;
+static int new_keys;
 
 #define STEP			0
 #define SEED			1024
@@ -228,9 +229,13 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 #endif
 
 	// Copy data to gpu
-	BENCH_CLERROR(clEnqueueWriteBuffer(queue[gpu_id], mem_in,
-		CL_FALSE, 0, global_work_size * sizeof(pass_t), host_pass, 0,
-		NULL, multi_profilingEvent[0]), "Copy data to gpu");
+	if (ocl_autotune_running || new_keys) {
+		BENCH_CLERROR(clEnqueueWriteBuffer(queue[gpu_id], mem_in,
+			CL_FALSE, 0, global_work_size * sizeof(pass_t), host_pass, 0,
+			NULL, multi_profilingEvent[0]), "Copy data to gpu");
+
+		new_keys = 0;
+	}
 
 	// Run kernel
 	BENCH_CLERROR(clEnqueueNDRangeKernel(queue[gpu_id], crypt_kernel,
@@ -283,6 +288,8 @@ static void set_key(char *key, int index)
 
 	memcpy(host_pass[index].v, key, saved_len);
 	host_pass[index].length = saved_len;
+
+	new_keys = 1;
 }
 
 static char *get_key(int index)
--- a/src/opencl_pbkdf2_hmac_sha512_fmt_plug.c
+++ b/src/opencl_pbkdf2_hmac_sha512_fmt_plug.c
@@ -75,6 +75,7 @@ static cl_mem mem_in, mem_out, mem_salt, mem_state;
 static cl_kernel split_kernel;
 static cl_int cl_error;
 static struct fmt_main *self;
+static int new_keys;
 
 #define STEP                     0
 #define SEED                     256
@@ -259,9 +260,13 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 #endif
 
 	// Copy data to gpu
-	BENCH_CLERROR(clEnqueueWriteBuffer(queue[gpu_id], mem_in, CL_FALSE, 0,
-		global_work_size * sizeof(pass_t), host_pass, 0, NULL,
-		multi_profilingEvent[0]), "Copy data to gpu");
+	if (ocl_autotune_running || new_keys) {
+		BENCH_CLERROR(clEnqueueWriteBuffer(queue[gpu_id], mem_in, CL_FALSE, 0,
+			global_work_size * sizeof(pass_t), host_pass, 0, NULL,
+			multi_profilingEvent[0]), "Copy data to gpu");
+
+		new_keys = 0;
+	}
 
 	// Run kernel
 	BENCH_CLERROR(clEnqueueNDRangeKernel(queue[gpu_id], crypt_kernel, 1,
@@ -306,6 +311,8 @@ static void set_key(char *key, int index)
 	// ^= the whole uint64 with the ipad/opad mask
 	strncpy((char*)host_pass[index].v, key, PLAINTEXT_LENGTH);
 	host_pass[index].length = saved_len;
+
+	new_keys = 1;
 }
 
 static char *get_key(int index)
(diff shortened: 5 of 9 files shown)


magnumripper:
This is untested as of yet

magnumripper:
I saw a number of formats completely lacking `new_keys` optimization. I guess most of them are so slow it doesn't matter much, but if unsure it never hurts to add it.

solardiz (approved):
Looks OK to me, but needs testing. Especially the formats that got `new_keys` introduced here should be re-tested with multiple salts loaded.

src/opencl_sha256crypt_fmt_plug.c
```diff
@@ -573,7 +573,7 @@ static int crypt_all(int *pcount, struct db_salt *_salt)
 		output_hashes = calculated_hash;
 	}
 
-	if (new_keys) {
+	if (ocl_autotune_running || new_keys) {
```
ghost, line 576: As per the PR description this is not necessary in 'my formats' (initialization always happens). Anyway, it doesn't harm.

ghost (approved):
I'm Ok with it.

magnumripper:
After testing this I'm starting to think the `ocl_autotune_running || ` isn't needed anymore. Absolutely nothing changed! All profiling data is there anyway. I'll review the functions in `opencl_common.c`. My current guess is it was needed for some earlier versions of them (I guess we re-send the keys each time now).

Adding `new_keys` logic is another thing, that's still sane to do. I didn't see any difference using `--test` though, likely because all of these formats use the `Raw` benchmark, right? Also, most are so slow the xfer times become insignificant.

magnumripper:
> Adding `new_keys` logic is another thing, that's still sane to do. I didn't see any difference using `--test` though, likely because all of these formats use the `Raw` benchmark, right? Also, most are so slow the xfer times become insignificant.

It's more or less useless even for pbkdf2-hmac-sha256 at a cost of 1000 (the fastest one of these):
At chosen GWS:
```
xfer: 315.680 us, init: 100.768 us, crypt: 13x10.446 ms, final: 18.144 us, res xfer: 178.336 us
gws:     34816   255186 c/s  3062232000 rounds/s  136.433 ms per crypt_all()+
```
The key transfer is 0.2% of the total.

solardiz:
I think we should add the `new_keys` logic to those 3 salted formats that missed it anyway. For the generic PBKDF2 ones, the iteration count might happen to be low. For WPA, it'd be just for consistency.

Also, you probably measure the 0.2% on PCIe x16, but some systems use x1 and there's also VCL and who knows what else in future/experimental systems/setups. (Incidentally, this is also why I think we should have some safety measures in our OpenCL kernels to allow for corrupted inputs and not let them affect more than just one corresponding output. I mean sanitizing host-provided lengths.)

magnumripper:
Will do

magnumripper:
If x1 means 5 ms xfer instead of 315 us, we're talking 3.7%, that's more rewarding (although I didn't account for 16x worse result xfer)..

BTW several formats that now do everything on GPU wastefully return a full GWS array of `int` which is actually used as bool for cracked or not. It would be trivial to change them (now that we've got it figured out, just pasting boilerplate code) into only returning a count, along with (if applicable) just the very indeces that was cracked. So a normal result transfer would be 4 bytes instead of eg. 128 KB at a GWS of 32K. If we again think PCIx1, that can become significant as well (nanoseconds instead of 2.85 ms). But it should probably be a separate PR.

That also goes for the formats that aren't fully GPU side but may do early rejection on GPU, such as 7z (where padding size is 1 or better). That would not only reduce xfer but also a probably significant amount of CPU / threading overhead. I've had this in my mind for a while.

solardiz:
> BTW several formats that now do everything on GPU wastefully return a full GWS array of `int` which is actually used as bool for cracked or not. [...] That would not only reduce xfer but also a probably significant amount of CPU / threading overhead. I've had this in my mind for a while.

Yes, I also thought/recalled that the other day. If/when I move tezos-opencl's BLAKE2b to device side, the next natural change would be to do comparison on device side as well and then the question arises of what to transfer to host. I sure would want to avoid that loop with `memcmp` in it on host, but replacing it with a loop that checks the many individual `int` flags feels silly. So I actually thought of moving the `any_cracked` logic to device side, and still transferring the whole thing when that flag is set, but indeed we can also replace that with a count.

What's our current reference format for "now that we've got it figured out, just pasting boilerplate code"?

solardiz:
> What's our current reference format for "now that we've got it figured out, just pasting boilerplate code"?

So we have `crack_count_ret` logic in `zip_kernel.cl` and `cryptosafe_kernel.cl`. These are different: looks like the former sends only the needed outputs (if any), whereas the latter sends all if any are needed. This difference makes sense: for `zip`, sending an output to host doesn't yet imply a successful crack (just that early-reject failed), whereas for `cryptosafe` it does.

solardiz:
> So we have `crack_count_ret` logic in `zip_kernel.cl` and `cryptosafe_kernel.cl`.

BTW, the host code blindly trusts the value of `crack_count_ret`, and may crash or worse if that value is out of bounds. Perhaps we should introduce some safety checks there. OpenCL devices and communication to them aren't necessarily as reliable as computation within one process on host, so it's wrong to unnecessarily make the latter less reliable.

magnumripper:
Right, sorry I missed your question there. Looking at them now, I guess I meant the ZIP format. The "tricky" part was matching output indeces against input indeces - but it's really trivial in hindsight. Also, while investigating whether we "can" use `atomic_inc()` at all I realized all GPU-side-mask formats used exactly that for years and years.

And yes, we might want to add some sanity checks there. Assuming we don't have internal mask, how about something like this?
```diff
diff --git a/src/opencl_zip_fmt_plug.c b/src/opencl_zip_fmt_plug.c
index 092e72c58..a9ca4aa2d 100644
--- a/src/opencl_zip_fmt_plug.c
+++ b/src/opencl_zip_fmt_plug.c
@@ -310,6 +310,8 @@ static int crypt_all(int *pcount, struct db_salt *salt)
        BENCH_CLERROR(clEnqueueReadBuffer(queue[gpu_id], cl_crack_count_ret, CL_TRUE, 0, sizeof(cl_uint), &crack_count_ret, 0, NULL, NULL), "failed reading results back");
 
        if (crack_count_ret) {
+               if (crack_count_ret > count)
+                       error_msg("Corrupt return: Got a claimed %u cracks out of %d\n", crack_count_ret, count);
                gws = GET_NEXT_MULTIPLE(crack_count_ret, local_work_size);
                BENCH_CLERROR(clEnqueueNDRangeKernel(queue[gpu_id], final_kernel, 1, NULL, &gws, lws, 0, NULL, multi_profilingEvent[2]), "Failed running crypt kernel");
                BENCH_CLERROR(clEnqueueReadBuffer(queue[gpu_id], cl_result, CL_TRUE, 0, sizeof(zip_hash) * crack_count_ret, outbuffer, 0, NULL, multi_profilingEvent[3]), "failed reading results back");
```
Could we do better? BTW in current code, `crack_count_ret` is unsigned int, `count` is signed (but AFAICU more or less guaranteed not to be negative).

solardiz:
> And yes, we might want to add some sanity checks there. Assuming we don't have internal mask, how about something like this?
> 
> ```diff
> diff --git a/src/opencl_zip_fmt_plug.c b/src/opencl_zip_fmt_plug.c
> index 092e72c58..a9ca4aa2d 100644
> --- a/src/opencl_zip_fmt_plug.c
> +++ b/src/opencl_zip_fmt_plug.c
> @@ -310,6 +310,8 @@ static int crypt_all(int *pcount, struct db_salt *salt)
>         BENCH_CLERROR(clEnqueueReadBuffer(queue[gpu_id], cl_crack_count_ret, CL_TRUE, 0, sizeof(cl_uint), &crack_count_ret, 0, NULL, NULL), "failed reading results back");
>  
>         if (crack_count_ret) {
> +               if (crack_count_ret > count)
> +                       error_msg("Corrupt return: Got a claimed %u cracks out of %d\n", crack_count_ret, count);
>                 gws = GET_NEXT_MULTIPLE(crack_count_ret, local_work_size);
>                 BENCH_CLERROR(clEnqueueNDRangeKernel(queue[gpu_id], final_kernel, 1, NULL, &gws, lws, 0, NULL, multi_profilingEvent[2]), "Failed running crypt kernel");
>                 BENCH_CLERROR(clEnqueueReadBuffer(queue[gpu_id], cl_result, CL_TRUE, 0, sizeof(zip_hash) * crack_count_ret, outbuffer, 0, NULL, multi_profilingEvent[3]), "failed reading results back");
> ```

Looks OK to me, but:

> Could we do better?

Maybe print a message starting with "Error: ", cap `crack_count_ret` to `count`, and continue. OTOH, would that mean we'd be likely reading uninitialized data from device (assuming the correct `crack_count_ret` was very different)? Anyway, even if it crashes there (unlikely?), that wouldn't be worse than your version.

> BTW in current code, `crack_count_ret` is unsigned int, `count` is signed (but AFAICU more or less guaranteed not to be negative).

Right.

magnumripper:
> and continue

Maybe, but that could mean we continue with things out of whack (eg. with false negatives) for ages and without later being able to resume from where it went berserk.

solardiz:
> without later being able to resume from where it went berserk.

Oh, that's a good point. OK, let's abort on such errors.

magnumripper:
I'll open an issue

PR #4944, opened 2021-12-14
tezos-opencl: Mostly avoid busy-waits on NVIDIA

solardiz:
This is a prototype for generic/shared code, to be first tested/used as part of this format. Works very well in my testing so far.

### DIFF
--- a/src/opencl_tezos_fmt_plug.c
+++ b/src/opencl_tezos_fmt_plug.c
@@ -250,6 +250,42 @@ static void set_salt(void *salt)
 	HANDLE_CLERROR(clFlush(queue[gpu_id]), "failed in clFlush");
 }
 
+#include "timer.h"
+#define WAIT_INIT(work_size) { \
+	static uint64_t wait_last_work_size; \
+	int wait_sleep = (work_size >= wait_last_work_size); \
+	wait_last_work_size = work_size; \
+	static uint64_t wait_times[20], wait_min; \
+	static unsigned int wait_index; \
+	uint64_t wait_start = 0;
+#define WAIT_SLEEP \
+	if (gpu_nvidia(device_info[gpu_id])) { \
+		wait_start = john_get_nano(); \
+		uint64_t us = wait_min >> 10; /* 2.4% less than min */ \
+		if (wait_sleep && us >= 1000) \
+			usleep(us); \
+	}
+#define WAIT_UPDATE \
+	if (gpu_nvidia(device_info[gpu_id])) { \
+		uint64_t wait_new = john_get_nano() - wait_start; \
+		if (wait_new < wait_min && wait_new < wait_min * 1000 / 1012) /* 1.2% less than min */ \
+			wait_new = wait_new * 7 / 8; /* we might have overslept and don't know by how much */ \
+		if (wait_times[wait_index] == wait_min) { /* about to replace former minimum */ \
+			unsigned int i; \
+			wait_times[wait_index] = wait_min = ~(uint64_t)0; \
+			for (i = 0; i < sizeof(wait_times) / sizeof(wait_times[0]); i++) \
+				if (wait_times[i] < wait_min) \
+					wait_min = wait_times[i]; \
+		} \
+		wait_times[wait_index++] = wait_new; \
+		if (wait_index >= sizeof(wait_times) / sizeof(wait_times[0])) \
+			wait_index = 0; \
+		if (wait_new < wait_min) \
+			wait_min = wait_new; \
+		wait_sleep = 1; \
+	}
+#define WAIT_DONE }
+
 static int crypt_all(int *pcount, struct db_salt *salt)
 {
 	const int count = *pcount;
@@ -282,14 +318,18 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 				NULL, &gws, lws, 0, NULL,
 				multi_profilingEvent[1]), "Run kernel");
 
+	WAIT_INIT(gws)
 	for (i = 0; i < (ocl_autotune_running ? 1 : loops); i++) {
 		BENCH_CLERROR(clEnqueueNDRangeKernel(queue[gpu_id],
 					split_kernel, 1, NULL,
 					&gws, lws, 0, NULL,
 					multi_profilingEvent[2]), "Run split kernel");
+		WAIT_SLEEP
 		BENCH_CLERROR(clFinish(queue[gpu_id]), "clFinish");
+		WAIT_UPDATE
 		opencl_process_event();
 	}
+	WAIT_DONE
 
 	BENCH_CLERROR(clEnqueueNDRangeKernel(queue[gpu_id],
 				final_kernel, 1, NULL,
@@ -297,9 +337,13 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 				multi_profilingEvent[3]), "Run final kernel");
 
 	// Read the result back
+	WAIT_INIT(gws)
+	WAIT_SLEEP
 	BENCH_CLERROR(clEnqueueReadBuffer(queue[gpu_id], mem_pk, CL_TRUE, 0,
 				gws * sizeof(tezos_pk_t), host_pk,
 				0, NULL, multi_profilingEvent[4]), "Copy result back");
+	WAIT_UPDATE
+	WAIT_DONE
 
 	if (!ocl_autotune_running) {
 #ifdef _OPENMP


solardiz:
BTW, besides saving energy and leaving the CPU available for other uses, this should actually speed things up for this very format on systems with fewer CPU cores than GPUs - e.g., former mining rigs. The expectation here is that after these changes a process will most likely be actively running on a CPU in the remainder of the busy-wait just before the condition of the busy-wait is met, whereas previously context switch to the process (from another busy-waiting process) would quite likely have occurred somewhat later than the condition is met.

src/opencl_tezos_fmt_plug.c
```diff
@@ -250,6 +250,42 @@ static void set_salt(void *salt)
...
+		} \
+		wait_times[wait_index++] = wait_new; \
+		if (wait_index >= sizeof(wait_times) / sizeof(wait_times[0])) \
+			wait_index = 0; \
+		if (wait_new < wait_min) \
+			wait_min = wait_new; \
+		wait_sleep = 1; \
+	}
+#define WAIT_DONE }
+
```
ghost, line 288: If you move this block of code to, for example, `opencl_common.h`, it will be easier to try it in other formats as well.

solardiz: That's the plan for later. Do you want to start testing this in other formats already this week? I can make this move now if so.

solardiz:
This format uses a split kernel with fixed iteration count, and the counts per loop are almost constant (513+512+512+511). In formats that use split kernel with variable iteration count, I think we should skip the SLEEP and UPDATE on the last invocation of the split kernel - the one for partial iteration count that might differ between salts. Otherwise we need to maintain the state added here per-salt.

solardiz:
Here's what's a bit illogical right now: the sleeping only starts when we've filled up the `wait_times` array (otherwise the minimum ends up 0), however upon a reduction of `work_size` the sleeping restarts as soon as we've added just _one_ more sample (as long as the array is otherwise full). That's controlled with the final `wait_sleep = 1;`. Simply removing that one line would result in the sleeping restarting upon next `crypt_all` with a similarly reduced `work_size` (e.g. for the same reduced batch of candidates but against next salt).

src/opencl_tezos_fmt_plug.c
ghost: > Do you want to start testing this in other formats already this week?

Right now, I probably won't work on it.

magnumripper:
My man page (macOS) for `usleep` says it "obsolescent" and suggests using `nanosleep` instead. But it also means it should work on macOS. I just realized my laptops can't build the new tezos kernel, will open an issue for that.

solardiz:
> My man page (macOS) for `usleep` says it "obsolescent" and suggests using `nanosleep` instead. But it also means it should work on macOS.

Before using `usleep` here, I saw it already used in our shared OpenCL code. So my changes shouldn't make portability worse.

magnumripper:
> In formats that use split kernel with variable iteration count, I think we should skip the SLEEP and UPDATE on the last invocation of the split kernel - the one for partial iteration count that might differ between salts. Otherwise we need to maintain the state added here per-salt.

Here's how I think I'll handle in in 7z-opencl, this should be correct right?
```diff
diff --git a/src/opencl_7z_fmt_plug.c b/src/opencl_7z_fmt_plug.c
index af8f85fb7..2e1f13c93 100644
--- a/src/opencl_7z_fmt_plug.c
+++ b/src/opencl_7z_fmt_plug.c
@@ -344,15 +344,17 @@ static int crypt_all(int *pcount, struct db_salt *salt)
                WAIT_INIT(global_work_size)
                for (i = 0; i < (ocl_autotune_running ? 1 : LOOP_COUNT); i++) {
                        BENCH_CLERROR(clEnqueueNDRangeKernel(queue[gpu_id],
                                crypt_kernel, 1, NULL, &global_work_size, lws, 0,
                                NULL, multi_profilingEvent[2]),
                                "Run loop kernel");
-                       WAIT_SLEEP
+                       if (i != LOOP_COUNT - 1)
+                               WAIT_SLEEP
                        BENCH_CLERROR(clFinish(queue[gpu_id]), "Error running loop kernel");
-                       WAIT_UPDATE
+                       if (i != LOOP_COUNT - 1)
+                               WAIT_UPDATE
                        opencl_process_event();
                }
                WAIT_DONE
```
Both macros seem safe to use this way, starts with `if () {` and ends with `}`.

Edit: That format doesn't seem to need it, but this tweak will be needed for some formats.

solardiz:
@magnumripper Your `opencl_7z_fmt_plug.c` patch above looks correct to me, but for formats that actually need that.

solardiz:
> ```diff
> -                       WAIT_SLEEP
> +                       if (i != LOOP_COUNT - 1)
> +                               WAIT_SLEEP
> ```

@magnumripper Looks like you haven't implemented this anywhere? Are you sure none of the formats you already got uses of `WAIT_*` into need this - that is, may have a final loop kernel invocation with a very different internal iteration count?

solardiz:
Will be the first example with skipping of the last loop kernel invocation's sleep:

```diff
+++ b/src/opencl_bitcoin_fmt_plug.c
@@ -398,11 +398,15 @@ static int crypt_all(int *pcount, struct db_salt *salt)
                &global_work_size, (local_work_size <= init_kernel_max_lws) ? lws : NULL, 0, NULL,
                multi_profilingEvent[1]), "Run init kernel");
 
+       // Better precision for WAIT_ macros
+       BENCH_CLERROR(clFinish(queue[gpu_id]), "clFinish");
+
        // Run loop kernel
        cl_uint left = cur_salt->cry_rounds - 1;
        cl_uint batch = HASH_LOOPS;
        HANDLE_CLERROR(clSetKernelArg(crypt_kernel, 1, sizeof(cl_uint),
                (void *)&batch), "Error setting kernel argument 1");
+       WAIT_INIT(global_work_size)
        do {
                if (batch > left) {
                        batch = left;
@@ -413,10 +417,15 @@ static int crypt_all(int *pcount, struct db_salt *salt)
                        crypt_kernel, 1, NULL,
                        &global_work_size, lws, 0, NULL,
                        multi_profilingEvent[2]), "Run loop kernel");
+               if (batch == HASH_LOOPS)
+                       WAIT_SLEEP
                BENCH_CLERROR(clFinish(queue[gpu_id]), "Error running loop kernel");
+               if (batch == HASH_LOOPS)
+                       WAIT_UPDATE
                opencl_process_event();
                left -= batch;
        } while (left && !ocl_autotune_running);
+       WAIT_DONE
 
        BENCH_CLERROR(clEnqueueNDRangeKernel(queue[gpu_id],
                final_kernel, 1, NULL,
```

magnumripper:
> @magnumripper Looks like you haven't implemented this anywhere? Are you sure none of the formats you already got uses of WAIT_* into need this - that is, may have a final loop kernel invocation with a very different internal iteration count?

I had it in mind but may well have missed it in some formats. For example, the sha2crypt formats can have a non-default iteration count, so they should have this.

solardiz:
> the sha2crypt formats can have a non-default iteration count, so they should have this.

That's what I thought earlier today, but they appear to only invoke the loop kernel for a fixed per-invocation iteration count and a separate final kernel afterwards.

magnumripper:
Just brainstorming a totally different solution, I wonder if `clSetEventCallback()` could be of use. Register a callback function in `signals.c` and then we could even just `sleep(1)` for a whole second. The callback function would raise some signal of choice, interrupting the sleep. This would mean we always oversleep, but only for extremely short time, if my idea holds at all? Maybe I totally misunderstand things.

It would probably not help #5048 though, as it doesn't look like we're oversleeping there.

solardiz:
> Just brainstorming a totally different solution, I wonder if `clSetEventCallback()` could be of use.

I have no idea - but this is worth trying. Will you?

> just `sleep(1)` for a whole second. The callback function would raise some signal of choice, interrupting the sleep.

Actually, `sleep()` would probably resume on its own, but you can use `select()` instead - that one will be interrupted.

```c
int sleep_select(int sec, int usec)
{
	struct timeval timeout;

	timeout.tv_sec = sec;
	timeout.tv_usec = usec;

	return select(0, NULL, NULL, NULL, &timeout);
}
```

magnumripper:
I tried using `sleep()` and it is indeed interrupted but only by SIGALRM, not SIGUSR2 which I wanted to repurpose. I'll try with `select()` instead.

We use #ifdef for SIGUSR1 and SIGUSR2 stuff in `signals.c`... Do we know of systems that don't have SIGUSR2? Such as Windows?

magnumripper:
I tried that out with `sleep_select` now. Technically it works like a champ - I used a 10 second sleep(!) but it was interrupted promptly whenever the queue was finished so the callback function raised a SIGUSR2 after 35-40 ms. However, once we were actually back out of the `sleep_select()` function, another 65 ms or so had past. So the idea is wonderful except the overhead is far to big 😢 .

Maybe I should profile it and see what the heck happens, 65 ms sounds like a very big overhead, doesn't it? The signal handler is just this:
```c
static void sig_handle_usr2(int signum)
{
	sig_install(sig_handle_usr2, SIGUSR2);
}
```

magnumripper:
No wait, I think I see the problem. Even `select` doesn't stop upon SIGUSR2 (so it stops at our timer's SIGALRM - those "65 ms" could be nearly a second in some runs). So we must use some other signal... which?

magnumripper:
Hmm or can we change the behavior? Or use `pselect`? This is all greek to me: sigmask, sigprocmask and whatnot.

solardiz:
@magnumripper Try `sigaction` without `SA_RESTART`.

magnumripper:
I tried this now but not sure if that's (enough of) what you meant. Didn't help.
```diff
@@ -145,7 +145,8 @@ static void sig_install(void *handler, int signum)
 
        memset(&sa, 0, sizeof(sa));
        sa.sa_handler = handler;
-       sa.sa_flags = SA_RESTART;
+       if (signum != SIGUSR2)
+               sa.sa_flags = SA_RESTART;
        sigaction(signum, &sa, NULL);
 #else
        signal(signum, handler);
```

Apart from that I'm at this point (after looking at https://man7.org/conf/lca2006/Linux_2.6_changes/pselect_4.html):
```c
static inline int sleep_select(int sec)
{
	struct timespec timeout;

	timeout.tv_sec = sec;
	timeout.tv_nsec = 0;

#if 0
	/* This all should be unneeded as we have a signal handler */
	sigset_t blockset;
	sigemptyset(&blockset);
	sigaddset(&blockset, SIGUSR2);
	sigprocmask(SIG_BLOCK, &blockset, NULL);
#endif
	sigset_t emptyset;
	sigemptyset(&emptyset);
	sigaddset(&emptyset, SIGALRM); // Ignore the timer

	return pselect(0, NULL, NULL, NULL, &timeout, &emptyset);
}
```
The "ignore the timer" does make it ignore the timer. However, it also ignores USR2.

magnumripper:
I have a very dirty branch at https://github.com/magnumripper/john/commits/busy-callback, only affecting NT-opencl for now.

solardiz:
I don't think these would help, but you can try:

When we use `sigaction` (rather than `signal`), the signal handler can be empty - it does not need to reinstall the handler (that is only a workaround for `signal` on some systems).

You can try another signal, like `SIGIO` or `SIGPIPE` or `SIGXCPU`.

For some of these (e.g., `SIGPIPE` or `SIGSEGV`), you can try raising them indirectly by triggering the corresponding event in the kernel.

This will likely help:

We can have a `select` return by having it wait on a file descriptor and us actually writing a byte into that file descriptor (perhaps a pipe) from the callback (no need for signals). Closing the file descriptor would probably also work. IIRC, OpenSSH uses this sort of communication between its master process and pre-auth privsep child for implementation of `MaxStartups` - you can take a look.

We can also try, and it's simpler, but risky (unless NVIDIA explicitly documents this is intended use maybe?) -

Jump out of the callback with `longjmp` or `siglongjmp`. (Also no need for signals.)

solardiz:
BTW, we don't currently use `SIGVTALRM`, so can try using it here. And we can have the kernel trigger it if `raise` somehow does not interrupt (but try `raise` first).

We can also use `SIGALRM`, and simply restart our sleep if it was interrupted by our regular 1-second alarm. We can distinguish by checking a global `volatile int` variable we'd add, which we'd set from the callback.

Also, I think the interruption by signal can occur even with `SIG_IGN` or `SIG_DFL` (for signals where the default is to ignore) - we shouldn't need a handler.

solardiz:
> Closing the file descriptor would probably also work. IIRC, OpenSSH uses this sort of communication between its master process and pre-auth privsep child for implementation of `MaxStartups` - you can take a look.

I just did - and yes it does. However, in its case this is so reasonable to do because it also needs to wait in `select` for incoming connections. So it's the same one `select` call for both incoming connections and privsep children progressing past authentication or dying. Anyway, it's an option we can use if `longjmp` isn't safe for us.

```c
                for (i = 0; i < num_listen_socks; i++)
                        FD_SET(listen_socks[i], fdset);
                for (i = 0; i < options.max_startups; i++)
                        if (startup_pipes[i] != -1)
                                FD_SET(startup_pipes[i], fdset);

                /* Wait in select until there is a connection. */
                ret = select(maxfd+1, fdset, NULL, NULL, NULL);
                if (ret < 0 && errno != EINTR)
                        error("select: %.100s", strerror(errno));
[...]
                if (ret < 0)
                        continue;

                for (i = 0; i < options.max_startups; i++)
                        if (startup_pipes[i] != -1 &&
                            FD_ISSET(startup_pipes[i], fdset)) {
                                /*
                                 * the read end of the pipe is ready
                                 * if the child has closed the pipe
                                 * after successful authentication
                                 * or if the child has died
                                 */
```

solardiz:
Here's a line of thought: how does NVIDIA's code know when to trigger the callback if it's just one process and it's sleeping? How does NVIDIA's userland code get control just at the right moment? Are we in context of our own thread (e.g., in a signal handler that NVIDIA set up - but there are not that many signals they could use) or perhaps in another thread (that NVIDIA spawned and is waiting e.g. on a `read` in there)? In the latter case (which feels more likely), perhaps the `longjmp` would be very wrong.

solardiz:
I think triggering `select` to return via a file descriptor event is the way to proceed here. If so, the choice is of the kind of event: either data (1 byte) or closing of the fd. Since we'll need many of these events, I think data will result in fewer syscalls per iteration: it's just `select`, `write`, and `read` (and we can keep the pipe fd's open). With closing, we'd have these per iteration: `pipe`, `select`, two `close` (for both ends of the pipe).

solardiz:
Also, before we proceed, doesn't usage of `clSetEventCallback` cause busy-wait, possibly in another thread? That depends on how NVIDIA implements it. We don't want to waste our time and end up with something that doesn't solve the original problem. So far you @magnumripper only mentioned the callback occurs at the right time - you did not mention what CPU usage was like during the wait time.

magnumripper:
Yeah that's a good question. I was hoping to get some Q'n'D code just good enough to see the resulting OpenCL performance and CPU load, but this final detail was harder than I thought.

No more time right now, will continue my little quest in a couple of days.

magnumripper:
I realized I could do a quick test: Change the 10s sleep to 30ms, never mind it not interrupted, and the result of that is 23% CPU load. That's worse than with current bleeding, but seems to indicate we don't just move the busy-wait.

solardiz:
Great. Now try the select/write/read approach (no signals)?

magnumripper:
Yep, but that'll be in a couple of days

solardiz:
I think here's why interrupting with a signal didn't work: the callback is probably in the context of another thread, and `raise` only sends the signal to the current thread. I think we need to record `getpid()` to a global variable before our call to `clSetEventCallback`, and then use `kill(pid, SIGUSR2)` in the callback.

magnumripper:
> the callback is probably in the context of another thread

Awesome, that was it. It's working now. Will do some performance tests.

magnumripper:
Early tests: 2% performance drop. CPU util decreased from 17.6% to 14.2%. Will make new macros and try other formats.

solardiz:
I just realized I overcomplicated things when I suggested using file descriptor based `select` events. We only need to wait for one event, so we can simply wait on `read`, and do a 1-byte `write` from the other end, without needing `select`. Then we would have no timeout, though - would get blocked on `read` forever if the callback is never invoked anymore.

solardiz:
> Early tests: 2% performance drop. CPU util decreased from 17.6% to 14.2%.

Is this 2% relative to no sleep or relative to our previous `WAIT` macros? I guess the CPU util decrease is relative to previous `WAIT`, since otherwise it'd be from 100%.

magnumripper:
Yes, it's relative to current bleeding with your macros.

Preliminary tests for tezos doesn't look good at all, utilization plummeted. Hopefully I just screwed up when I put the code in alternative macros, no more time today.

solardiz:
> Yes, it's relative to current bleeding with your macros.

In other words, this didn't help resolve the major regression for nt-opencl on the dual-socket rig, or have you not tried there yet?

magnumripper:
No I haven't. I will try both your macros with select sleep, and these alternate ones.

magnumripper:
With NT, I'm now below 1.5% performance drop compared to feature disabled, which I think is good enough given it can be disabled. Also, it's a single macro now, just replacing `clFinish()` so very easy to implement. But with slower formats (even non-looped, such as zip), it looks hopeless (eg. 29% GPU util instead of 98%) - and I have yet to figure out why? Somehow this method seem to work well in some situations but not others, but test results are inconsistent.

The kernel durations are very similar, so it's not that.

magnumripper:
OK I got zip performing fine as well now. The critera for this to work well seem to be that it ends with a transfer (tiny or not). With just a kernel, it adds significant overhead. That's strange, and very unfortunate for our looped kernels.

Just for the sake of it I tried adding a dummy read to a looped format, but that alone did not seem to help. I then modified it to read some actual result of the loop kernel (just 32 bits) and this helped - performance restored. Not sure what to make out of all this other than to ditch the whole idea.

magnumripper:
Scrap that, my code is flawed. Need to start over. Gosh, this is tricky. It has to be at least two macros, one for setting the event before enqueuing, and one for sleeping.

magnumripper:
Somehow my flawed code worked almost correctly. After fixes, I'm back at: This method "can" be used but it has a slight performance penalty, it's ugly and tricky. Dummy transfers need to be put in the split kernel loops. I'm now taking a break from this entirely until I get some new idea.

Issue #3124, opened 2018-01-24
Some dynamic formats, canonical hash not matching pot entry


magnumripper:
Here's dynamic_1034: md5($p.$u) username "jumbo" password "magnum"
```
$ echo 'jumbo:$dynamic_1034$16ccb184ee1f16dd9af509135f504442' > test.in
```
Here's what the format reports as the canonical hash (the second, `$$Ujumbo` one):
```
$ ../run/john --show=types test.in
jumbo:$dynamic_1034$16ccb184ee1f16dd9af509135f504442:::/:/:/:dynamic_1034:0:1:0:$dynamic_1034$16ccb184ee1f16dd9af509135f504442$$Ujumbo:0:
```

As expected, the same canonical hash is given for a bare hash and specifying format on command line instead:
```
$ echo 'jumbo:16ccb184ee1f16dd9af509135f504442' > test.in

$ ../run/john --show=types test.in -form:dynamic_1034
jumbo:16ccb184ee1f16dd9af509135f504442:::/:/:/:dynamic_1034:0:1:0:$dynamic_1034$16ccb184ee1f16dd9af509135f504442$$Ujumbo:0:
```

Fine so far. But here's what's actually in the pot file after cracking:
```
$ ../run/john test.in -mask:magnum
Using default input encoding: UTF-8
Loaded 1 password hash (dynamic_1034 [md5($p.$u) (PostgreSQL MD5) 256/256 AVX2 8x3])
Press 'q' or Ctrl-C to abort, almost any other key for status
magnum           (jumbo)
1g 0:00:00:00  10.00g/s 10.00p/s 10.00c/s 10.00C/s magnum
Use the "--show" option to display all of the cracked passwords reliably
Session completed

$ cat ../run/john.pot
$dynamic_1034$16ccb184ee1f16dd9af509135f504442$HEX$24556a756d626f:magnum
```
They don't match, the canonical format isn't canonical. How come? How can this even happen!? External tools should rely on the `--show=types` information and to the best of my knowledge they can, except for dynamic formats.

And it gets worse: Let's say I *did* hexify the input file, in order to make some external tool recognize the pot entries as matching the input file:
```
$ cat test.in
jumbo:$dynamic_1034$16ccb184ee1f16dd9af509135f504442$HEX$24556a756d626f

$ ../run/john --show=types test.in
[...]

Issue #2354, opened 2016-12-06
Support for $HEX[] wordlists (like in hashcat)


Zeokat:
Hashcat support $HEX[] wordlists to encode control chars. Will be nice have this option available into JTR, in that way we can use hashcat wordlists into JTR.

So, a password of "foobar1[CRLF]" becomes: $HEX[666f6f626172310d0a]

Is this format already supported in JTR? Do you plan to add it? Exist any way to handle control chars in wordlists?

Not sure if i should post this here or at john-users list, since it contains ideas and also questions 😕 .

Thanks in advance for your help.

jfoug:
This new enhancement request is just fine.  How about you also post to john-users (linking to this page), so that others can see what is requested.

jfoug:
NOTE, the way jtr loads files, it will almost certainly be best to not have too many of these $HEX[xxxx] words.  When they are needed, great, but taking a wordlist like rock-u (or worse yet, something 10's of times bigger), and converting every line into a $HEX[] string would very likely NOT be optimal for memory usage.  Things could be made to work, but at the expense (likely) of memory.   However, take for instance rock-u.  I think there are some binary items in there.  To make sure they are correctly handled, they could be properly converted into $HEX[] items, while leaving all of the 7 bit ascii data intact.

But I do like this idea.  It might be easy, but it also might not be 'super' easy.

I also think we should look at something like this for .pot storage.   

Also, a simple tool to convert into and out of $HEX[] format (enhancement to th base64_convert program should do the trick)

jfoug:
@magnumripper can you think of other areas where $HEX[] should be dealt with if/when we add this option?  If so, then edit the top post, adding a new task.

magnumripper:
I didn't know HC supported this format for wordlists (I knew about the pot files). What if a user sets his password to (verbatim) $HEX[deadbabe]? Will that make it impossible to crack by Hashcat?
I recently had a patch accepted for HC that stops outputting silly hex for perfectly fine UTF-8 strings - and nowadays that about gets rid of ALL the hex. Virtually everything uses Unicode now. Because of that and other things I see no need for this, but feel free to implement it.

@Zeokat please note that while HC can (I think) fully use input with CR, LF and even NULL characters, JtR can not - and this HEX format can't change that fact. We'd need to change every little part of JtR core for that to happen (using pascal strings starting with a length byte). It would probably be faster though, it's a pity Solar didn't make it that way.

[...]

PR #4947, opened 2021-12-16
Dynamic format: Don't un-hex ciphertexts too early.

magnumripper:
We don't want unprintable ciphertexts in the pot file.

NULLs within a salt were already handled so I just extended that to include anything that's not printable ASCII.

I believe I also fixed a bug in the NULL handling (would be double-hex-encoded in the end) but I didn't bother verifying that. What I did verify was that the problems seen went away and I really don't expect any regression but the dynamic formats' code is a real mess.

### DIFF
--- a/src/dynamic_fmt.c
+++ b/src/dynamic_fmt.c
@@ -362,7 +362,17 @@ private_subformat_data curdat;
  *********************************************************************************
  *********************************************************************************/
 
-char *RemoveHEX(char *output, char *input)
+#define KEEP_UNPRINTABLE 0
+#define KEEP_NULLS 1
+
+/*
+ * Decode HEX$deadcafe salt strings into bytes. If last parameter is KEEP_UNPRINTABLE and there's
+ * unprintable characters in the string, bail without decoding (still copy input to output).
+ * If last parameter is KEEP_NULLS, anything except embedded nulls will be decoded.
+ * Output buffer must at least as big as input buffer.  If hex is decoded, the resulting
+ * output will be shorter.
+ */
+static char *RemoveHEX(char *output, char *input, int only_null)
 {
 	char *cpi = input;
 	char *cpo = output;
@@ -380,12 +390,13 @@ char *RemoveHEX(char *output, char *input)
 	*cpo++ = *cpi;
 	cpi += 5;
 	while (*cpi) {
-		if (*cpi == '0' && cpi[1] == '0') {
-			strcpy(output, input);
-			return output;
-		}
 		if (atoi16[ARCH_INDEX(*cpi)] != 0x7f && atoi16[ARCH_INDEX(cpi[1])] != 0x7f) {
-			*cpo++ = atoi16[ARCH_INDEX(*cpi)]*16 + atoi16[ARCH_INDEX(cpi[1])];
+			unsigned char c = atoi16[ARCH_INDEX(*cpi)]*16 + atoi16[ARCH_INDEX(cpi[1])];
+			if ((only_null && c == 0) || (!only_null && (c < ' ' || c > 0x7e))) {
+				strcpy(output, input);
+				return output;
+			}
+			*cpo++ = c;
 			cpi += 2;
 		} else if (*cpi == '$') {
 			while (*cpi && strncmp(cpi, "$HEX$", 5)) {
@@ -402,6 +413,56 @@ char *RemoveHEX(char *output, char *input)
 	}
 	*cpo = 0;
 	return output;
+}
+
+/*
+ * Add $HEX$ encoding for salts where needed.  For canonicalized pot storage, we encode
+ * anything containing unprintables or field separators, but never anything else.
+ */
+static char *AddHEX(char *ciphertext) {
+	char *cp;
+
+	if (strncmp(ciphertext, "$dynamic_", 9) && strncmp(ciphertext, "@dynamic=", 9))
+		return ciphertext;  // not a dynamic format, so we can not 'fix' it.
+
+	if (!(cp = strchr(&ciphertext[1], ciphertext[0])) || !(cp = strchr(&cp[1], '$')))
+		return ciphertext;  // not a salted format.
+
+	if (!strncmp(cp, "$HEX$", 5))
+		return ciphertext;  // already HEX$
+
+	unsigned char *s = (unsigned char*)cp++;
+	int need_hex = 0;
+	if (*cp && (cp[strlen(cp) - 1] == ' ' || cp[strlen(cp) - 1] == '\t'))
+		need_hex = 1;
+	else
+	while (*++s) {
+		if (*s < ' ' || *s > 0x7e || *s == ':' || *s == '$') {
+			need_hex = 1;
+			break;
+		}
+	}
+
+	/* New length is length of ciphertext, the HEX$ and 2 bytes per char of salt string plus ending null. */
+	if (need_hex) {
+		static char *out;
+		static size_t out_size;
+		size_t size_needed = strlen(ciphertext) + 4 + strlen(cp) + 1;
+
+		if (size_needed > out_size) {
+			out_size = (size_needed + 1024) / 1024 * 1024;
+			out = mem_realloc(out, out_size);
+		}
+
+		char *cpx = out;
+		cpx += sprintf(out, "%.*sHEX$", (int)(cp - ciphertext), ciphertext);
+		while (*cp)
+			cpx += sprintf(cpx, "%02x", (unsigned char)(*cp++));
+
+		return out;
+	}
+
+	return ciphertext;
 }
 
 /*********************************************************************************
@@ -432,7 +493,7 @@ static int valid(char *ciphertext, struct fmt_main *pFmt)
 	// of leaving it in there. The ONLY problem we still have is NULL bytes.
 	if (strstr(ciphertext, "$HEX$")) {
 		if (strnlen(ciphertext, sizeof(fixed_ciphertext) + 1) < sizeof(fixed_ciphertext))
-			ciphertext = RemoveHEX(fixed_ciphertext, ciphertext);
+			ciphertext = RemoveHEX(fixed_ciphertext, ciphertext, KEEP_NULLS);
 	}
 
 	cp = &ciphertext[strlen(pPriv->dynamic_WHICH_TYPE_SIG)];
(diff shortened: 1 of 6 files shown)


magnumripper:
> I really don't expect any regression but the dynamic formats' code is a real mess.

Seems I was wrong, bots found a problem. Back to the drawing board.

magnumripper:
While it now passes self-test and bot checks, it fails TS for dynamic_15 (only that one).

The dynamic format is a proper mess.

magnumripper:
Here's how I *think* it oughta be (in hindsight, this code has sometimes grown too fast):

1. Hashes that don't have hexified salt but need it (contains unprintable ASCII, problematic whitespace or field separators) must be hexified in pot file.
2. Hashes that have, but doesn't really need, hexified salt (decodes to printable ASCII that are not field separators) should either be unhexified in pot file, or left as-is. Not sure what I like best, perhaps the latter - if it complies with §3:
3. Hexified input should obviously match non-hexfied pot entries and vice versa. We do this all the time in other formats, it's not rocket science. I'm rusty now but perhaps that means we either need to unhexify §2,  or **always** hexify all pot entries' salts. Hmmm no, I think as long as we treat them the same (eg. pass them through one same canonicalization function), it should be fine regardless.
4. `get_salt()` should (must) be able to parse `HEX$` (or not) salts and put it in "pascal strings" or whatever. Some comments mention that's exactly how it works, but I see no sign of any handling of NULL bytes or `HEX$` encoding there. Yet I do see test vectors with HEX encoded salts containing nulls. Not sure how that works.
5. The misplaced catches in `loader.c` and `logger.c` for dynamic format, callling `dynamic_FIX_SALT_TO_HEX()`, should not be needed if things are done right. No other format does that sort of thing because it's not needed if you do it right.

Current code is a nasty mix of all of the above. It sometimes unhexes things, sometimes not, sometimes it does and later fails because it did (that's even before my try here). For example, a hexified salt having `$$U` username tag will not be read correctly. The TS doesn't catch anywhere near all problems.

solardiz:
I'm afraid I won't review this for real because I'm not familiar with what we had in this respect before and have no time to dive into it now.

On a related note though, I'd like us to gain support for hashcat's `$HEX[]` notation for plaintexts (in wordlists, in pot files, perhaps even on status line). For example, there's plenty of that in hashes.org lists still available e.g. at https://github.com/rarecoil/hashes.org-list and it's a pity and wrong that it doesn't work with JtR directly.

magnumripper:
> I'm afraid I won't review this for real because I'm not familiar with what we had in this respect before and have no time to dive into it now.

Current version fixes the initial problem and (after a day and a half of fixes and re-testing) still passes `--test`, `--test-full` and the Test Suite in normal mode as well as `-internal`mode. Oh, and all of that is currently re-running with ASan as well and takes forever (looks like at least one more hour from now). I needed the initial problem fixed so once that passes too, I think I'll just merge it unless you find any caveats in the meantime.

> On a related note though, I'd like us to gain support for hashcat's `$HEX[]` notation for plaintexts (in wordlists, in pot files, perhaps even on status line). For example, there's plenty of that in hashes.org lists still available e.g. at https://github.com/rarecoil/hashes.org-list and it's a pity and wrong that it doesn't work with JtR directly.

That's up high on my to-do list as well but it's kind of a separate issue: I think what we'd do first is implement that transparently to modes and formats in `loader.c` (for pot entries) - and then implement it in any modes that read any kind of wordlist. After that, we could look into having the (now only two) places in `dynamic_fmt.c` that handles this support both versions. What I just did here will make it a whole lot easier.

magnumripper:
> After that, we could look into having the (now only two) places in `dynamic_fmt.c` that handles this support both versions.

Oh and at about the same time we'd look into incorporating it to more formats that need it, one at a time. I'm not sure I know any other format apart from the WPAPSK ones (when SSID's use problematic characters).

magnumripper:
Everything passed. I'll sleep on this, please do a shallow review if you like.

solardiz:
> implement that transparently to modes and formats in `loader.c` (for pot entries)

Hmm, is that only for `--make-charset` (for it to use the decoded bytes for statistics) or did you mean something else?

magnumripper:
> > implement that transparently to modes and formats in `loader.c` (for pot entries)
> 
> Hmm, is that only for `--make-charset` (for it to use the decoded bytes for statistics) or did you mean something else?

That, and I guess for `--show` (such as "if we're in UTF-8 mode and the password decodes well as printable UTF-8, decode it - if we're NOT in UTF-8 mode, perhaps decide on some other strategy), and for loopback mode...

src/dynamic_fmt.c
```diff
@@ -380,12 +383,13 @@ char *RemoveHEX(char *output, char *input)
 	*cpo++ = *cpi;
 	cpi += 5;
 	while (*cpi) {
-		if (*cpi == '0' && cpi[1] == '0') {
-			strcpy(output, input);
-			return output;
-		}
 		if (atoi16[ARCH_INDEX(*cpi)] != 0x7f && atoi16[ARCH_INDEX(cpi[1])] != 0x7f) {
-			*cpo++ = atoi16[ARCH_INDEX(*cpi)]*16 + atoi16[ARCH_INDEX(cpi[1])];
+			char c = atoi16[ARCH_INDEX(*cpi)]*16 + atoi16[ARCH_INDEX(cpi[1])];
```
solardiz: I'd use `unsigned char` here.

src/dynamic_fmt.c
```diff
@@ -380,12 +383,13 @@ char *RemoveHEX(char *output, char *input)
...
+			if ((only_null && c == 0) || (!only_null && (c < ' ' || c > 0x7e))) {
+				strcpy(output, input);
```
solardiz, line 396: How do we know(?) we won't overflow `output` here?

src/dynamic_fmt.c
```diff
@@ -380,12 +383,13 @@ char *RemoveHEX(char *output, char *input)
...
+				return output;
+			}
+			*cpo++ = c;
 			cpi += 2;
```
solardiz, line 400: And here. Maybe add a comment on the assumption this function makes on its input and output sizes.

src/dynamic_fmt.c
```diff
@@ -404,6 +408,46 @@ char *RemoveHEX(char *output, char *input)
 	return output;
 }
 
+// Add $HEX$ encoding for salts where needed.  For canonicalized pot storage, we encode anything
+// containing unprintables or field separators, but never anything else
+static char *AddHEX(char *ciphertext) {
+	static char out[1024];
```
solardiz: How do we know(?) 1024 is enough?

magnumripper:
> > > implement that transparently to modes and formats in `loader.c` (for pot entries)
> > 
> > 
> > Hmm, is that only for `--make-charset` (for it to use the decoded bytes for statistics) or did you mean something else?
> 
> That, and I guess for `--show` (such as "if we're in UTF-8 mode and the password decodes well as printable UTF-8, decode it - if we're NOT in UTF-8 mode, perhaps decide on some other strategy), and for loopback mode...

Oh and we'd probably make *storing* to pot file hexified, under conditions yet to decide. That should be transparent to formats as long as we still ignore the null byte problem.

I'm not sure how Hashcat currently does it, but they went from storing as-is to hexify any non-ASCII or anything that contains `:`, to (after a PR of mine) not hexify printable UTF-8.

src/dynamic_fmt.c
```diff
@@ -404,6 +408,46 @@ char *RemoveHEX(char *output, char *input)
 	return output;
 }
 
+// Add $HEX$ encoding for salts where needed.  For canonicalized pot storage, we encode anything
+// containing unprintables or field separators, but never anything else
```
solardiz: I think we normally use C++ style comments only for commenting things out and somehow also in OpenCL - I generally use the old C style comments in our main code in `src`. But I don't really mind, especially if that was already the style in nearby code?

src/dynamic_fmt.c
```diff
@@ -404,6 +408,46 @@ char *RemoveHEX(char *output, char *input)
...
+	// ok, see if this is salted:
+	if (!(cp = strchr(&ciphertext[1], ciphertext[0])) || !(cp = strchr(&cp[1], '$')))
+		return ciphertext;  // not a salted format.
+
+	if (!strncmp(cp, "$HEX$", 5))
+		return ciphertext;  // already HEX$
+
+	char *s = cp++;
+	int need_hex = 0;
+	if (cp[strlen(cp) - 1] == ' ' || cp[strlen(cp) - 1] == '\t' )
```
solardiz: I don't see whether/why `strlen(cp)` would be at least 1 here. Looks unsafe to me.

magnumripper:
> Oh and we'd probably make _storing_ to pot file hexified, under conditions yet to decide. That should be transparent to formats as long as we still ignore the null byte problem.

BTW perhaps even after that. The only solution I can picture (and it's conceptually trivial) is to return a pascal string instead of a C string. So hex/unhex could still be done transparent to the formats, in `loader.c` and `logger.c`.

src/dynamic_fmt.c
```diff
@@ -404,6 +408,46 @@ char *RemoveHEX(char *output, char *input)
...
+		return ciphertext;  // not a dynamic format, so we can not 'fix' it.
+
+	// ok, see if this is salted:
+	if (!(cp = strchr(&ciphertext[1], ciphertext[0])) || !(cp = strchr(&cp[1], '$')))
+		return ciphertext;  // not a salted format.
+
+	if (!strncmp(cp, "$HEX$", 5))
+		return ciphertext;  // already HEX$
+
+	char *s = cp++;
```
solardiz: Maybe `unsigned char *` and cast, so that we don't rely on signed `*s < ' '` also catching 8-bit (although it does).

src/dynamic_fmt.c
```diff
@@ -404,6 +408,46 @@ char *RemoveHEX(char *output, char *input)
...
+			break;
+		}
+	}
+
+	if (need_hex) {
+		char *cpx;
+		cpx = out;
+		cpx += sprintf(out, "%*.*sHEX$", (int)(cp-ciphertext), (int)(cp-ciphertext), ciphertext);
+		while (*cp)
+			cpx += sprintf(cpx, "%02x", (unsigned char)(*cp++));
```
solardiz, line 460: This makes the size of output larger than input's. Is `out` really large enough?

solardiz (commented):
Some nitpicks from my out-of-context "review". I also see the old code was problematic.

I also wonder whether `$HEX$` can occur in an ASCII salt by accident, and whether any scripts generating `dynamic_` hashes should be taught to handle that (encode the `$HEX$` in `$HEX$`).

magnumripper:
This also closes #2437

magnumripper:
> I also wonder whether `$HEX$` can occur in an ASCII salt by accident, and whether any scripts generating `dynamic_` hashes should be taught to handle that (encode the `$HEX$` in `$HEX$`).

That's a problem I'm quite sure is not addressed here, I *think* such encoding wouldn't even help if made manually. We did mention in in #2354 though.

magnumripper:
> > I also wonder whether `$HEX$` can occur in an ASCII salt by accident, and whether any scripts generating `dynamic_` hashes should be taught to handle that (encode the `$HEX$` in `$HEX$`).
> 
> That's a problem I'm quite sure is not addressed here, I _think_ such encoding wouldn't even help if made manually. We did mention in in #2354 though.

The chance of it appearing in random data is slim though, one in a trillion for 8-bit and one in 7.7 billion for printable ASCII.

src/dynamic_fmt.c
magnumripper: I understand it sometimes matter but does it matter here?

solardiz: No, it does not. `c < ' '` will catch 8-bit just fine. It's just a matter of style.

src/dynamic_fmt.c
magnumripper: Yeah that (already the style in nearby code) was the reason here. It's a mix really. This functions was mostly just moved from `dynamic_utils.c`

src/dynamic_fmt.c
magnumripper: I think that's pretty much guaranteed for the dynamics but I see now this actually wasn't in the original function but an assumption of mine. I'll replace it with a `mem_alloc_tiny`

src/dynamic_fmt.c
magnumripper: I'm not sure why you want this but I'm all ears

solardiz: A matter of style. If we do rely on the signed compare catching 8-bit elsewhere, then no reason to change the code here - OK to leave it as you have it.

src/dynamic_fmt.c
magnumripper: What I do (think I) know is that many tests in other parts of dynamics assume (unhexed) stuff is shorter total than 512, correct or not... But again, I should probably revert this to the `mem_alloc_tiny`.

src/dynamic_fmt.c
magnumripper: The caller is expected (from original code) to supply an output buffer at least as big as the input buffer and the result from "maybe removing hex" can never only ever be shorter. As far as I understand the callers do the right thing although I'm slightly wary about one that use a buffer of 512 instead of the ususal 1024. I just assumed Jim happened to know what he was smoking at that particular moment 😆

src/dynamic_fmt.c
magnumripper: Hmm... so as currently written, the `<` catches nearly all of the unprintables while the `>` catch only single char or so, right? So it works but not quite as I pictured it. I agree it might be a good habit to use unsigned then.

solardiz: > as currently written, the `<` catches nearly all of the unprintables while the `>` catch only single char or so, right?

Yes, that's how it works with signed `char`. BTW, the signedness of `char` is implementation specific, so as written the checks works differently under the hood on different platforms. Luckily, the outcome is the same.

src/dynamic_fmt.c
magnumripper: I opted for realloc instead, in steps of 1K. In case we load gazillions of some dynamic hashes it could be a problematic waste to mem_alloc_tiny each of them. This function is only called from `split()` which usually have a static buffer so there's no need to be thread safe or so.

magnumripper:
Given the state of the original code I'm more or less expecting this will end up in some very odd regression we'll find sooner or later (probably/hopefully later) that's not quite caused by this but surfaces due to it. We can't do much better than handle it once it hits us.

magnumripper:
Also closes #3124

PR #4950, opened 2021-12-20
Initial support for ENCDataVault

sylvainpelissier:
* Support for key derivation and decryption algorithms.
* Tested on Sony ENCDataVault lite 6.2.13, SanDisk SecureAccess 3.02 and ENC DataVault 7.1.1W
* Problem patched by Western Digital: https://www.westerndigital.com/en-ap/support/product-security/wdc-21014-sandisk-secureaccess-software-update

### DIFF
--- a/doc/NEWS
+++ b/doc/NEWS
@@ -245,6 +245,7 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
 - Revised regex mode's support for librexgen, to version >= 2.1.5  [Jan Starke,
   magnum; 2021]
 
+- Added support for ENCDataVault and encdatavault2john.py script [sylvainpelissier; 2021]
 
 Major changes from 1.8.0-jumbo-1 (December 2014) to 1.9.0-jumbo-1 (May 2019):
 
--- a/doc/README-ENCDataVault.md
+++ b/doc/README-ENCDataVault.md
@@ -1,0 +1,26 @@
+Cracking ENCSecurity DataVault
+==============================
+
+1. Run encdatavault2john.py on vault folder.
+
+```bash
+$ ../run/encdatavault2john.py path/to/vault > hash
+```
+
+2. Run john on the file.
+
+```bash
+$ ../run/john hash
+```
+
+The decryption test is done on a 32-bit magic value, meaning that occasional 
+false positives can occur. Cracked passwords for those false positives will not
+decrypt the vault correctly.
+
+The format has been tested on Sony ENCDataVault lite 6.2.13, SanDisk 
+SecureAccess 3.02 and ENC DataVault 7.1.1W.
+
+The key derivation problems were patched by Western Digital: 
+https://www.westerndigital.com/en-ap/support/product-security/wdc-21014-sandisk-secureaccess-software-update. 
+Their new solution is now called PrivateAccess: 
+https://kb.sandisk.com/app/answers/detail/a_id/21996/
--- a/run/encdatavault2john.py
+++ b/run/encdatavault2john.py
@@ -1,0 +1,67 @@
+#!/usr/bin/env python3
+
+# Helper script for cracking ENCSecurity DataVault.
+#
+# This software is Copyright (c) 2021, Sylvain Pelissier <sylvain.pelissier at kudelskisecurity.com>
+# and it is hereby released to the general public under the following terms:
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted.
+
+import sys
+import binascii
+import os  
+from pathlib import Path
+
+
+def process(vault):
+    salt = None
+
+    if not os.path.isdir(vault):  
+        sys.stderr.write(f"{vault} : should be a folder.\n")
+        sys.exit(1)
+
+    file_list = os.listdir(vault)
+    if "filesystem.dat" in file_list:
+        # Sandisk, Sony vault or ENCSecurity user password.
+        file_path = Path(vault) / "filesystem.dat"
+
+    elif "index.dat" in file_list and "keychain.dat" in file_list:
+        # ENCSecurity vault.
+        file_path = Path(vault) / "index.dat"
+    else:
+        sys.stderr.write(f"{vault} : Valid vault not found.\n")
+        return
+    
+    with open(file_path, "rb") as f:
+        header = f.read(4)                  # Read header
+        if header == b'\xd2\xc3\xb4\xa1':   # Test if we have a valid header
+            version = int.from_bytes(f.read(4),byteorder="little")
+            crypto = int.from_bytes(f.read(4),byteorder="little")
+            iv = binascii.hexlify(f.read(8))
+            header_enc = binascii.hexlify(f.read(4))
+        else:
+            sys.stderr.write(f"{file_list[0]} : Valid header not found.\n")
+            return
+
+        if len(header_enc) != 8:
+            sys.stderr.write(f"{file_list[0]} : Problem reading encrypted header.\n")
+            return
+
+    sys.stdout.write(f"{vault}:$encdv${version}${crypto}${iv.decode()}${header_enc.decode()}")
+
+    if version == 3:
+        with open(Path(vault) / "keychain.dat", "rb") as f:
+            f.seek(16)
+            keychain = binascii.hexlify(f.read(128))
+            sys.stdout.write(f"${keychain.decode()}\n")
+    else:
+        sys.stdout.write("\n")
+
+if __name__ == "__main__":
+    if len(sys.argv) < 2:
+        sys.stderr.write(f"Usage: {sys.argv[0]} <vault folder>\n")
+        sys.stderr.write(f"\nExample: {sys.argv[0]} vault/\n")
+        sys.exit(1)
+
+    process(sys.argv[1])
--- a/src/encdatavault_fmt_plug.c
+++ b/src/encdatavault_fmt_plug.c
@@ -1,0 +1,429 @@
+/*
+ * Cracker for ENCSecurity Data Vault.
+ *
+ * This software is Copyright (c) 2021 Sylvain Pelissier <sylvain.pelissier at kudelskisecurity.com>
+ * and it is hereby released to the general public under the following terms:
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted.
+ */
+
+#if FMT_EXTERNS_H
+extern struct fmt_main fmt_encdadatavault;
+#elif FMT_REGISTERS_H
+john_register_one(&fmt_encdadatavault);
+#else
+
+#include <string.h>
+
+#ifdef _OPENMP
+#include <omp.h>
+#endif
+
+#include "aes.h"
+#include "md5.h"
+#include "common.h"
+#include "formats.h"
+
+#define FORMAT_TAG              "$encdv$"
+#define FORMAT_TAG_LENGTH       (sizeof(FORMAT_TAG) - 1)
+#define FORMAT_LABEL            "ENCDataVault"
+#define FORMAT_NAME             ""
+#define ALGORITHM_NAME          "MD5"
+#define BENCHMARK_COMMENT       " (1000 iterations)"
+#define BENCHMARK_LENGTH        0x107
+#define PLAINTEXT_LENGTH        125
+#define BINARY_SIZE             0
+#define BINARY_ALIGN            1
+#define SALT_SIZE               sizeof(struct custom_salt)
+#define SALT_ALIGN              4
+#define MIN_KEYS_PER_CRYPT      1
+#define MAX_KEYS_PER_CRYPT      4
+
+#ifndef OMP_SCALE
+#define OMP_SCALE               16
+#endif
+
+#define ENC_DEFAULT_NUM_HASH_ITERATIONS 1000
+#define ENC_SALT_SIZE 16
+#define ENC_IV_SIZE 16
+#define ENC_BLOCK_SIZE 16
+#define ENC_KEY_SIZE 16
+#define ENC_NONCE_SIZE 8
+#define ENC_SIG_SIZE 4
+#define ENC_MAX_KEY_NUM 8
+#define ENC_KEYCHAIN_SIZE 128
+
+typedef union buffer_128_u {
+	uint8_t u8[8];
+	uint64_t u64[2];
+} buffer_128;
+
+static MAYBE_INLINE void enc_xor_block(uint64_t *dst, const uint64_t *src)
+{
+	dst[0] ^= src[0];
+	dst[1] ^= src[1];
+}
+
+// md5("f46fcf4f0d9d45b198eb240ff819aac0\x00\x00\x00" + index)
+static const char *const default_salts[] = {
+	"\x0f\xc9\xe7\xd0\x8b\xe4\x24\xf6\x56\x9d\x4e\x72\xed\xbc\x2c\x5c",
+	"\xdd\x79\x74\xf3\x3d\x83\x00\xc2\x9b\xd2\x93\xd5\x7f\x9d\x9b\x8c",
+	"\x60\x85\x0c\x47\x58\x46\xe2\x96\x2d\x99\x5d\x5e\xf1\xd0\x6a\x28",
+	"\xe2\x3f\x3d\x6b\x99\x61\x4b\xa9\xc4\xed\xc5\xdd\xd8\x25\x3c\xe1",
+	"\x2c\xa4\x59\x89\x1d\x78\x52\xdb\x30\x31\xd0\x9f\x9f\x34\x88\x35",
+	"\xdb\x1b\xb5\x27\xe8\x21\x4f\x79\xa0\xb2\xcb\x32\x42\xd9\xf2\x0a",
+	"\xae\xa8\xb6\x8e\xd0\x7b\x62\xa1\x40\x0e\x17\xc6\xad\x64\x20\xc8",
+	"\xea\xe3\xf4\x4e\xaf\x4a\x8f\x84\xf1\xfa\xb3\x08\x85\x69\xbe\xf8"
+};
+
+static struct fmt_tests encdatavault_tests[] = {
+	// Sandisk vaults
+	{ "$encdv$1$1$ae07a8354f6fe3ca$a6066363", "bbbb" },
+	{ "$encdv$1$1$7531d98593ea7b1d$d070650d", "openwall" },
+	// Sony vault
+	{ "$encdv$1$2$e8a5d78fa5511fa2$b96a4747", "bbbb" },
+	{ "$encdv$1$2$4e277e5547b88f9c$eb0ed650", "openwallopenwall!" },
+	// ENCDataVault 512 bits user password
+	{ "$encdv$1$3$d6209d17c0a87818$77608f04", "123456789ABCDEf" },
+	{ "$encdv$1$3$3563390c4d66944d$2b0470f5", "openwallopenwall" },
+	// ENCDataVault 128 bits vault
+	{ "$encdv$3$1$75c97f784cad5027$c58e34a9$6fa3c4085acadda7c94589fe3dd8209d59a79b0ea659c4af51861f659c2d6cc645dab7c821c2cc0e8da97ce66e9be779fcf8fc33c1250aee2cd46e08a3864763a5c7f790c9965376a36fbf3b1c8b944d096e3bbe586a952f9fab5ee2f1c1ca7d2cd06ff357bb397c3eb1da66c5562998411c0b2dff04860e6f6adf818c853941", "128vaultTest" },
+	{ "$encdv$3$1$af9479e81a896dbb$e643f5f6$8e8fd6ccf10cf651efcb561a167fa301849bef0a28718f67d31d53a1e7a8328b3aaa6c642d8a2483e8de5c3751fd0b81604ead7c6412e6197886735bcbd1dd50d23f293bf11856986bd07a7513275dea991968376703a643fc19c373c1d62a7779db88068e24752bb6761f42c8aec9cf7a80ef2b7f6384212ee6b670d8ba17a7", "OpenwallOpenwall" },
+	// ENCDataVault 256 bits vault
+	{ "$encdv$3$2$a44a9a62e131023e$d9fe5a04$0bc8c0c937cac8de5a226a9dc7ff5d2542bb8814973afdcc0e593fded8b337cac18acd60c1afea3550fdeabee339a36892eb99f0f502b9f74075e9cb26970e983189be1395ffbec8ebfb765a563db45c3e53d73040e41bcfe58dc211f03f1384c5080c298a5e4bdf7cf23b893b3d2dbe05c472180fdc3f7fd82d97ea0eec1e7e", "OpenwallOpenwall" },
+	{ "$encdv$3$2$85089cdde5cb099b$c3c98127$d96fd1e2709908e08994d1809e516985cf028cd73e2a829dee1ce7473c3c4d0ce35a2652b58ed7ad046641259b9a24a046c89d3251e1036e7117dd8221b07f5a45b6c215b3ad308e5f7b248ea8f1fabaa23da4840797c9052c3f8578187d514a92356b8138455db6424a41db6b0de2817d463f23d44f39537a6c28b2ba075a01", "Verylongbutstillfailpassword"},
+	// ENCDataVault 512 bits vault
+	{ "$encdv$3$3$bd837ac896f26983$15107b99$fb141e6db2cb586b2b445d6a6ca47f17d16947afa32298cc30c6fcdc4bfac7a874d53f2de4bcb196645e59e1c8e5883999875c9951d637e08f78d2bb16003c2abe8bfa1cb8b8d7b627828d61d775937308e7e119dea727da8af12490d50e8b8dc8f24daa7e101576112b52374f3ea4a73f7fa6bfb802bd6dfa845318a2884a9e", "123456789ABCDEf" },
+	{ "$encdv$3$3$ffd2a778b7f1304b$311893a8$5e48500065fc4138b134ce7858ac3b29d9f4b19c5f0d7ee1d07d8dd4dc3d5ae56b18e84d822087849573074a5776dee5309e5cb6bbd0d1470e0717463119bb080c96e24ebd563673060397803aebae5df7c59defcc8fa687b96c9a8245e540be699061c299a69830472e1a6a74ad72086dfac906a49e0ce84fc722da9715d675", "OpenwallOpenwall" },
+	// ENCDataVault 1024 bits vault
+	{ "$encdv$3$4$bc93a92cf625e360$fe120bfb$739f9d85964cef2f63b927ff77f3328cc5192014ae21c29954c322f4e808fe5c8abe64cc150dbcb08cb334f3b5c28357f10d8d5c6a103e2c899402136c14633aaf8a5347959b33b80ade2e3f5698864605940dc1423704999e5da859d6584491bbe940f00f162c75d2766b868ef2b4c6bf599a2e8ccea3f7cdfab193744a8d10", "123456789ABCDEf" },
+	{ "$encdv$3$4$e9786d82101f30ef$d690010a$5748af61f246c1569cc72cbd6c97e0be46b1612842fd3c6284637af5351508a602c731739acb056b845dfa2b5befd40f14136b4336e0c8e98144555a90befbc85abfc4069fbb71709fa39e29f8c6a98d28e14251e50ffdd43d75252de2b0b14c386e72927c62ae39ba18ff8b32a339d882fdd0b15284af246fa50b7c6992f783", "Verylongbutstillfailpassword" },
+	{ NULL }
+};
+
+static int *cracked;
+static int any_cracked;
+static size_t cracked_size;
+
+static char (*saved_key)[PLAINTEXT_LENGTH + 1];
+
+static struct custom_salt {
+	unsigned int version;
+	unsigned int algo_id;
+	unsigned char salt[ENC_SALT_SIZE];
+	unsigned char iv[ENC_IV_SIZE];
+	unsigned char encrypted_data[ENC_BLOCK_SIZE];
+	unsigned char keychain[ENC_KEYCHAIN_SIZE];
+} *cur_salt;
+
+void static enc_aes_ctr_iterated(const unsigned char *in, unsigned char *out, const unsigned char *key,
+                                 buffer_128 ivs[ENC_MAX_KEY_NUM], size_t len, size_t nb_keys, uint64_t counter)
+{
+	AES_KEY aes_key;
+	buffer_128 tmp_iv;
+	buffer_128 tmp_out;
+	int i, j;
+
+	AES_set_encrypt_key(key, ENC_KEY_SIZE * 8, &aes_key);
+	len >>= 4;
+	counter <<= 56;
+
+	for (i = 0; i < len; i++) {
+		tmp_iv.u64[0] = ivs[0].u64[0];
+		tmp_iv.u64[1] = counter;
+		AES_encrypt(tmp_iv.u8, tmp_iv.u8, &aes_key);
+		memcpy(tmp_out.u8, in, AES_BLOCK_SIZE);
+		enc_xor_block(tmp_out.u64, tmp_iv.u64);
+
+		for (j = 1; j < nb_keys; j++) {
+			tmp_iv.u64[0] = ivs[j].u64[0];
+			tmp_iv.u64[1] = counter;
+			AES_encrypt(tmp_iv.u8, tmp_iv.u8, &aes_key);
+			enc_xor_block(tmp_out.u64, tmp_iv.u64);
+		}
+		memcpy(out, tmp_out.u8, AES_BLOCK_SIZE);
+
+		// Increment counter, only 255 block encryption is supported.
+		counter += 0x0100000000000000;
+		out += AES_BLOCK_SIZE;
+		in += AES_BLOCK_SIZE;
+	}
+}
+
+static void init(struct fmt_main *self)
+{
+	omp_autotune(self, OMP_SCALE);
+	saved_key = mem_calloc(sizeof(*saved_key), self->params.max_keys_per_crypt);
+	cracked = mem_calloc(sizeof(*cracked), self->params.max_keys_per_crypt);
+	cracked_size = sizeof(*cracked) * self->params.max_keys_per_crypt;
+	any_cracked = 0;
+}
+
+static void done(void)
+{
+	MEM_FREE(saved_key);
+	MEM_FREE(cracked);
+}
+
+static int valid(char *ciphertext, struct fmt_main *self)
+{
+	char *p = ciphertext, *ctcopy, *keeptr;
+	int extra;
+
+	if (strncmp(ciphertext, FORMAT_TAG, FORMAT_TAG_LENGTH) != 0)
+		return 0;
+
+	ctcopy = strdup(ciphertext);
+	keeptr = ctcopy;
+	ctcopy += FORMAT_TAG_LENGTH;
+	if ((p = strtokm(ctcopy, "$")) == NULL) // version
+		goto err;
+	if (strcmp(p, "1") && strcmp(p, "3")) {
+		static int warned;
+		if (!warned++)
+			fprintf(stderr, "%s: Warning: version %d not supported, not loading such hashes!\n", self->params.label, atoi(p));
+		goto err;
+	}
+	char version = *p;
+	if ((p = strtokm(NULL, "$")) == NULL)   // algorithm id
+		goto err;
+	if (!isdec(p))
+		goto err;
+	if (atoi(p) > 4 || atoi(p) < 1) {
+		static int warned;
+		if (!warned++)
+			fprintf(stderr, "%s: Warning: algorithm id %d not supported!\n", self->params.label, atoi(p));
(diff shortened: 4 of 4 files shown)


src/encdatavault_fmt_plug.c
```diff
@@ -0,0 +1,407 @@
...
+	cs.iv[ENC_IV_SIZE - 1] = 1;
+
+	p = strtokm(NULL, "$");
+	for (i = 0; i < ENC_SIG_SIZE; i++)
+		cs.encrypted_data[i + 4] =
+		    (atoi16[ARCH_INDEX(p[2 * i])] << 4) | atoi16[ARCH_INDEX(p[2 * i +
+		            1])];
+
+	// Keychain for version 3
+	if (3 == cs.version) {
```
magnumripper: Our code style would put this as `if (cs.version == 3) {`. Same for line 249.

src/encdatavault_fmt_plug.c
```diff
@@ -0,0 +1,407 @@
...
+	if (!isdec(p))
+		goto err;
+	if ((p = strtokm(NULL, "$")) == NULL)   // algorithm id
+		goto err;
+	if (!isdec(p))
+		goto err;
+	if (atoi(p) > 4 || atoi(p) < 1) {
+		fprintf(stderr, "Warning: algorithm id %d not supported!\n", atoi(p));
+		goto err;
+	}
```
magnumripper: So we accept version 1, 2 and 3 but I only see tests for 1 or `else` in `crypt_all`. Is there a version 2 and does this code crack it too? If no to any of those, we should also reject version 2 here.

magnumripper (commented):
Look good to me in general. It's a pity SIMD isn't supported, would be really trivial and a good boost. Then again if vendor already fixed their weak KDF there may not be much point in spending time to implement it.

src/encdatavault_fmt_plug.c
sylvainpelissier: Corrected

src/encdatavault_fmt_plug.c
sylvainpelissier: You are right but the check needs to be done on the version number. I corrected it.

magnumripper: BTW we usually avoid printing warnings from `valid()` becuase it doesn't go well with auto-selection of format, but since this format has its own tag they won't hurt much. What we usually do in such cases though is something like:
```c
if (atoi(p) != 1 && atoi(p) != 3) {
	static int warning;
	if (!warning++)
 		fprintf(stderr, "%s: Warning: version %d not supported, not loading such hashes!\n", atoi(p), self->params.label);
	goto err;
}
```
another possibility is only showing warnings if verbosity was bumped:
```c
	if (options.verbosity >= VERB_DEBUG)
 		fprintf(stderr, "%s: Warning: version %d not supported, not loading such hashes!\n", atoi(p), self->params.label);
```
Perhaps @solardiz have some preference here?

run/encdatavault2john.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env python3
```
solardiz, line 1: Is this script known to require Python 3 or has it simply not been tested with Python 2? Either is OK - just asking.

run/encdatavault2john.py
```diff
@@ -0,0 +1,65 @@
...
+            keychain = binascii.hexlify(f.read(128))
+            sys.stdout.write("$%s\n" % keychain.decode())
+    else:
+        sys.stdout.write("\n")
+
+if __name__ == "__main__":
+    if len(sys.argv) < 2:
+        sys.stderr.write("Usage: %s <vault folder>\n" % sys.argv[0])
+        sys.stderr.write("\nExample: %s vault/\n" % sys.argv[0])
+        sys.exit(-1)
```
solardiz: Should use `1` for the exit code here, not `-1`. We fixed a bunch of those recently.

src/encdatavault_fmt_plug.c
```diff
@@ -0,0 +1,411 @@
...
+#include "arch.h"
+#include "md5.h"
+#include "misc.h"
+#include "common.h"
+#include "formats.h"
+#include "params.h"
+#include "options.h"
+#include "base64_convert.h"
+#include "johnswap.h"
+#include "simd-intrinsics.h"
```
solardiz: This format doesn't currently appear to use any SIMD, so perhaps the `#include "simd-intrinsics.h"` should be dropped. This suggests that other `#include` directives need to be reviewed as well. Perhaps they came from another format?

src/encdatavault_fmt_plug.c
```diff
@@ -0,0 +1,411 @@
...
+
+#define FORMAT_TAG              "$encdv$"
+#define FORMAT_TAG_LENGTH       (sizeof(FORMAT_TAG) - 1)
+#define FORMAT_LABEL            "ENCDataVault"
+#define FORMAT_NAME             ""
+#define ALGORITHM_NAME          "MD5 " MD5_ALGORITHM_NAME
+#define BENCHMARK_COMMENT       " (1000 iterations)"
+#define BENCHMARK_LENGTH        0x107
+#define PLAINTEXT_LENGTH        125
+#define HEX_BINARY_SIZE         (REAL_BINARY_SIZE*2)
```
solardiz: I see no `REAL_BINARY_SIZE` here, and perhaps `HEX_BINARY_SIZE` is unused. And this suggests `ecryptfs_fmt_plug.c` was probably the source format for this. ;-)

src/encdatavault_fmt_plug.c
```diff
@@ -0,0 +1,411 @@
...
+#define BINARY_SIZE             0
+#define BINARY_ALIGN            4
```
solardiz: Why align a zero-sized value? Use `1` for the alignment. BTW, we could want to enhance our self-test to detect this, if it doesn't already.

src/encdatavault_fmt_plug.c
```diff
@@ -0,0 +1,411 @@
...
+#define SALT_SIZE               sizeof(struct custom_salt)
+#define SALT_ALIGN              4
+#define MIN_KEYS_PER_CRYPT      1
+#define MAX_KEYS_PER_CRYPT      1
+
+#ifndef OMP_SCALE
+#define OMP_SCALE               2   // Tuned w/ MKPC for core i7
```
solardiz: Was this re-tuned for this format? It should be. And let's not proliferate those silly comments anyhow.

src/encdatavault_fmt_plug.c
```diff
@@ -0,0 +1,411 @@
...
+#define ENC_SALT_SIZE 16
+#define ENC_IV_SIZE 16
+#define ENC_BLOCK_SIZE 16
+#define ENC_KEY_SIZE 16
+#define ENC_IV_PREFIX_SIZE 8
+#define ENC_SIG_SIZE 8
+#define ENC_KEYCHAIN_SIZE 128
+
+// md5("f46fcf4f0d9d45b198eb240ff819aac0\x00\x00\x00" + index)
+static char *default_salts[] = {
```
solardiz: `static const char * const`.

src/encdatavault_fmt_plug.c
```diff
@@ -0,0 +1,411 @@
...
+	unsigned char iv[ENC_IV_SIZE];
+	unsigned char encrypted_data[ENC_BLOCK_SIZE];
+	unsigned char keychain[ENC_KEYCHAIN_SIZE];
+} *cur_salt;
+
+void enc_xor_block(uint64_t *dst, const uint64_t *src)
+{
+	dst[0] ^= src[0];
+	dst[1] ^= src[1];
+}
```
solardiz: Perhaps use `MAYBE_INLINE` on `enc_xor_block`, or make it a macro. Also, add `static` (if you keep this a function).

src/encdatavault_fmt_plug.c
```diff
@@ -0,0 +1,411 @@
...
+		out += AES_BLOCK_SIZE;
+		in += AES_BLOCK_SIZE;
+	}
+}
+
+static void init(struct fmt_main *self)
+{
+	omp_autotune(self, OMP_SCALE);
+	saved_key = mem_calloc_align(sizeof(*saved_key),
+	                             self->params.max_keys_per_crypt, MEM_ALIGN_WORD);
```
solardiz: I see no reason to use the `_align` function and `MEM_ALIGN_WORD` here.

src/encdatavault_fmt_plug.c
```diff
@@ -0,0 +1,411 @@
...
+	keeptr = ctcopy;
+	ctcopy += FORMAT_TAG_LENGTH;
+	if ((p = strtokm(ctcopy, "$")) == NULL) // version
+		goto err;
+	if (!isdec(p))
+		goto err;
+	if (atoi(p) != 1 && atoi(p) != 3) {
+		fprintf(stderr, "Warning: version %d not supported!\n", atoi(p));
+		goto err;
+	}
```
solardiz, line 187: You can simplify the `isdec` and value checks into one check of `strcmp(p, "1") && strcmp(p, "3")`.

src/encdatavault_fmt_plug.c
solardiz: My preference for the warnings style is like in magnum's first example here, except that we use `warned` for the variable name. Please implement that.

src/encdatavault_fmt_plug.c
```diff
@@ -0,0 +1,411 @@
...
+	return 1;
+
+err:
+	MEM_FREE(keeptr);
+	return 0;
+}
+
+static void *get_salt(char *ciphertext)
+{
+	static struct custom_salt cs = { 0 };
```
solardiz: Let's drop the initializer ` = { 0 }` given that we also use `memset`.

src/encdatavault_fmt_plug.c
```diff
@@ -0,0 +1,411 @@
...
+
+		for (j = 0; j < nb_keys; j++) {
+			memcpy(kdf_out[j], default_salts[j], ENC_SALT_SIZE);
+		}
+
+		for (j = 1; j < ENC_DEFAULT_NUM_HASH_ITERATIONS; j++) {
+			MD5_Init(&ctx);
+			MD5_Update(&ctx, hash, 16);
+			MD5_Final(hash, &ctx);
+			enc_xor_block((uint64_t *) result, (uint64_t *) hash);
```
solardiz: These casts violate C aliasing rules and will at least cause occasional failures on systems requiring alignment. Please use a `union` type instead. Ditto in other places.

src/encdatavault_fmt_plug.c
```diff
@@ -0,0 +1,411 @@
...
+			for (j = 1; j < nb_keys; j++) {
+				enc_xor_block((uint64_t *) ivs[j], (uint64_t *) kdf_out[j]);
+				enc_xor_block((uint64_t *) ivs[j], (uint64_t *) cur_salt->iv);
+			}
+			enc_aes_ctr_iterated(cur_salt->encrypted_data, result, kdf_out[0],
+			                     ivs, AES_BLOCK_SIZE, nb_keys, 1);
+			result_ptr = result + 4;
+			cracked[index] =
+			    (0 == memcmp(result_ptr, "\xd2\xc3\xb4\xa1\x00\x00\x00\x01",
+			                 ENC_SIG_SIZE));
```
solardiz: We'd just write `!memcmp`, and ditto in the other place. More importantly, I suggest you introduce the `any_cracked` logic (grep other formats for it), so that you don't need the loop in `cmp_all`.

src/encdatavault_fmt_plug.c
```diff
@@ -0,0 +1,411 @@
...
+	return count;
+}
+
+static int cmp_all(void *binary, int count)
+{
+	int index;
+
+	for (index = 0; index < count; index++)
+		if (cracked[index])
+			return 1;
```
solardiz: This loop would be gone.

src/encdatavault_fmt_plug.c
```diff
@@ -0,0 +1,411 @@
...
+{
+	return cracked[index];
+}
+
+static int cmp_exact(char *source, int index)
+{
+	return cracked[index];
+}
+
+static void encdv_set_key(char *key, int index)
```
solardiz: Don't prefix the function name with format name here. This function isn't any special compared to other format methods, where you already don't use any prefix.

solardiz (changes requested):
I made comments requesting some minor changes. Please don't add commits - just force-push one commit.

Please also add a `doc/NEWS` entry and maybe a `doc/README-ENCDataVault` file explaining what this is for, how to use it, and that the KDF is no longer the same on new devices (if I got it right).

In a separate commit, please feel free to also make the relevant cleanups to `ecryptfs_fmt_plug.c`.

Thank you!

solardiz:
> Tested on Sony ENCDataVault lite 6.2.13, SanDisk SecureAccess 3.02 and ENC DataVault 7.1.1W
> Problem patched by Western Digital: https://www.westerndigital.com/en-ap/support/product-security/wdc-21014-sandisk-secureaccess-software-update

This detail should go into the documentation to be added with this PR.

src/encdatavault_fmt_plug.c
```diff
@@ -0,0 +1,411 @@
...
+	counter = counter << 56;
+
+	for (i = 0; i < len; i++) {
+		tmp_iv[0] = ((uint64_t *) ivs[0])[0];
+		tmp_iv[1] = counter;
+		AES_encrypt((unsigned char *)tmp_iv, out, &aes_key);
+		enc_xor_block((uint64_t *) out, (uint64_t *) in);
+
+		for (j = 1; j < nb_keys; j++) {
+			tmp_iv[0] = ((uint64_t *)(ivs[j]))[0];
```
solardiz: The reads from `ivs` also violate aliasing rules and can end up misaligned.

src/encdatavault_fmt_plug.c
```diff
@@ -0,0 +1,411 @@
...
+	unsigned char keychain[ENC_KEYCHAIN_SIZE];
+} *cur_salt;
+
+void enc_xor_block(uint64_t *dst, const uint64_t *src)
+{
+	dst[0] ^= src[0];
+	dst[1] ^= src[1];
+}
+
+void enc_aes_ctr_iterated(const unsigned char *in, unsigned char *out,
```
solardiz: Missed `static` on `enc_aes_ctr_iterated`.

src/encdatavault_fmt_plug.c
```diff
@@ -0,0 +1,411 @@
...
+	cur_salt = (struct custom_salt *)salt;
+}
+
+static int crypt_all(int *pcount, struct db_salt *salt)
+{
+	const int count = *pcount;
+	int index;
+	int nb_keys;
+
+	memset(cracked, 0, sizeof(cracked[0]) * cracked_count);
```
solardiz: Just use `cracked_size` here, and drop `cracked_count` (don't need it anywhere).

src/encdatavault_fmt_plug.c
```diff
@@ -0,0 +1,411 @@
...
+	if (1 == cur_salt->version) {
+		nb_keys = 1 << (cur_salt->algo_id - 1);
+	} else {
+		nb_keys = 8;
+	}
+
+#ifdef _OPENMP
+	#pragma omp parallel for
+#endif
+	for (index = 0; index < count; index += MIN_KEYS_PER_CRYPT) {
```
solardiz: Please replace `index += MIN_KEYS_PER_CRYPT` with simply `index++`. The loop body doesn't currently support anything else anyway.

(We use the `index += MIN_KEYS_PER_CRYPT` in some SIMD-capable formats, where that corresponds to how many candidates are processed by one loop iteration when using SIMD.)

src/encdatavault_fmt_plug.c
frank-dittrich: May be the version should be reported as a tunable cost.

(Sorry, this souldn't have been a reply to Solar's comment.)

src/encdatavault_fmt_plug.c
```diff
@@ -0,0 +1,411 @@
...
+#include "options.h"
+#include "base64_convert.h"
+#include "johnswap.h"
+#include "simd-intrinsics.h"
+
+#define FORMAT_TAG              "$encdv$"
+#define FORMAT_TAG_LENGTH       (sizeof(FORMAT_TAG) - 1)
+#define FORMAT_LABEL            "ENCDataVault"
+#define FORMAT_NAME             ""
+#define ALGORITHM_NAME          "MD5 " MD5_ALGORITHM_NAME
```
magnumripper: The `MD5_ALGORITHM_NAME` resolvs to `AVX2 8x3` on my gear. That's not correct since there's no SIMD support.

src/encdatavault_fmt_plug.c
magnumripper: > May be the version should be reported as a tunable cost.

Commenting-out test vectors, it seems the shorter v1 hashes do about 65Kp/s while the longer v3 hash does 71K. I would have guessed the other way round.

src/encdatavault_fmt_plug.c
magnumripper: FWIW I tried retuning it (with best-effort `--tune=auto`) and speeds went up by 40%. MKPC was increased to 4 and OMP_SCALE to 16

src/encdatavault_fmt_plug.c
sylvainpelissier: I have not a big difference and the main time is spend in the 1000 MD5 iterations and it should not have big difference for AES decryption.

run/encdatavault2john.py
sylvainpelissier: Yes it requires Python3.

sylvainpelissier:
Thank you this was really helpful. I think I have integrated all the remarks. Let me know if something is missing.

doc/README-ENCDataVault
```diff
@@ -0,0 +1,17 @@
...
+```
+
+2. Run john on the file.
+```bash
+$ ../run/john hash
+```
+
+Tested on Sony ENCDataVault lite 6.2.13, SanDisk SecureAccess 3.02 and ENC DataVault 7.1.1W.
+
+The key derivation problems were patched by Western Digital: https://www.westerndigital.com/en-ap/support/product-security/wdc-21014-sandisk-secureaccess-software-update. Their new solution is now called PrivateAccess: https://kb.sandisk.com/app/answers/detail/a_id/21996/
```
magnumripper: This file lacks a final LF.

src/encdatavault_fmt_plug.c
magnumripper: This stands - line 36 can be removed

src/encdatavault_fmt_plug.c
```diff
@@ -0,0 +1,434 @@
...
+	ctcopy += FORMAT_TAG_LENGTH;
+	p = strtokm(ctcopy, "$");
+	cs.version = atoi(p);
+	p = strtokm(NULL, "$");
+	cs.algo_id = atoi(p);
+	p = strtokm(NULL, "$");
+	for (i = 0; i < ENC_NONCE_SIZE; i++)
+		cs.iv[i] =
+		    (atoi16[ARCH_INDEX(p[2 * i])] << 4) | atoi16[ARCH_INDEX(p[2 * i +
+		            1])];
```
magnumripper: BTW we've relaxed our code style from max. width of 80, to 120. Perhaps we should update `README.coding-style`.

src/encdatavault_fmt_plug.c
```diff
@@ -0,0 +1,434 @@
...
+} buffer_128;
+
+static MAYBE_INLINE void enc_xor_block(uint64_t *dst, const uint64_t *src)
+{
+	dst[0] ^= src[0];
+	dst[1] ^= src[1];
+}
+
+// md5("f46fcf4f0d9d45b198eb240ff819aac0\x00\x00\x00" + index)
+static const char *default_salts[] = {
```
solardiz: I really meant it with two `const` (`static const char * const default_salts[] = {`), so that _both_ the strings _and_ the pointers to them go in a read-only section.

Of course, there are plenty of other places in JtR where we don't get that right, but that's no reason not to start using this.

src/encdatavault_fmt_plug.c
```diff
@@ -0,0 +1,434 @@
...
+                                 size_t len, size_t nb_keys, uint64_t counter)
+{
+	AES_KEY aes_key;
+	buffer_128 tmp_iv = { 0 };
+	buffer_128 tmp_out = { 0 };
+	int i, j;
+
+	AES_set_encrypt_key(key, ENC_KEY_SIZE * 8, &aes_key);
+	len = len >> 4;
+	counter = counter << 56;
```
solardiz: Can use `>>=` and `<<=` here.

src/encdatavault_fmt_plug.c
```diff
@@ -0,0 +1,434 @@
...
+		/* AES iterated CTR */
+		if (cur_salt->version == 1) {
+			memcpy(ivs[0].u8, cur_salt->iv, ENC_NONCE_SIZE);
+			for (j = 1; j < nb_keys; j++) {
+				memcpy(ivs[j].u8, cur_salt->iv, ENC_NONCE_SIZE);
+				ivs[j].u64[0] ^= kdf_out[j].u64[0];
+			}
+			enc_aes_ctr_iterated(cur_salt->encrypted_data, result,
+			                     kdf_out[0].u8, ivs, AES_BLOCK_SIZE, nb_keys, 1);
+			result_ptr = result + 4;
```
solardiz: I see no need for the separate `result_ptr` variable here and in the other place (and thus at all). You can just pass `result + 4` into the `memcmp`.

src/encdatavault_fmt_plug.c
```diff
@@ -0,0 +1,434 @@
...
+#ifdef _OPENMP
+				#pragma omp atomic
+#endif
+				any_cracked |= 1;
+			}
+
+		}
+
+	}
+	return count;
```
solardiz, line 352: Weird placement of empty lines here.

doc/NEWS
```diff
@@ -245,6 +245,9 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
 - Revised regex mode's support for librexgen, to version >= 2.1.5  [Jan Starke,
   magnum; 2021]
 
+- Added support for ENCDataVault [sylvainpelissier; 2021]
+
+- Added encdatavault2john.py for cracking ENCDataVault. [sylvainpelissier; 2021]
```
solardiz: I'd combine these into one entry, but either approach works.

solardiz (approved):
Looks much better now, thanks! I've also made some new comments.

I'm also unsure whether the initializations to `= { 0 }` are needed. You can test by `memset` to some non-zero byte value instead.

src/encdatavault_fmt_plug.c
solardiz: Perhaps more importantly - `result` is kind of a two-dimensional array, but here (and in the other place) you always test the first "element", right? Is that right, and why?

doc/README-ENCDataVault
```diff
@@ -0,0 +1,17 @@
...
+1. Run encdatavault2john.py on vault folder.
+
+```bash
+$ ../run/encdatavault2john.py path/to/vault > hash
+```
+
+2. Run john on the file.
+```bash
+$ ../run/john hash
+```
```
solardiz: Either drop the backquotes or rename the file to `.md`.

run/encdatavault2john.py
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env python3
+
+# Helper script for cracking ENCSecurity DataVault.
+#
+# This software is Copyright (c) 2021, Kudelski Security <sylvain.pelissier at kudelskisecurity.com>
+# and it is hereby released to the general public under the following terms:
```
solardiz, line 6: Incidentally, the script is copyright by your company/employer and the C code by you personally - I guess that's deliberate? That's no problem, just pointing out in case it was in error.

src/encdatavault_fmt_plug.c
```diff
@@ -0,0 +1,434 @@
...
+
+static struct fmt_tests encdatavault_tests[] = {
+	// Sandisk vault
+	{ "$encdv$1$1$7531d98593ea7b1d$d070650d68d50e6e", "openwall" },
+	// Sony vault
+	{ "$encdv$1$2$4e277e5547b88f9c$eb0ed650ee5dffc6", "openwallopenwall!" },
+	// ENCDataVault 512 bits user password
+	{ "$encdv$1$3$d6209d17c0a87818$77608f044a4f113a", "123456789ABCDEf" },
+	// ENCDataVault 1024 bits vault
+	{ "$encdv$3$4$bc93a92cf625e360$fe120bfb658d27b6$739f9d85964cef2f63b927ff77f3328cc5192014ae21c29954c322f4e808fe5c8abe64cc150dbcb08cb334f3b5c28357f10d8d5c6a103e2c899402136c14633aaf8a5347959b33b80ade2e3f5698864605940dc1423704999e5da859d6584491bbe940f00f162c75d2766b868ef2b4c6bf599a2e8ccea3f7cdfab193744a8d10", "123456789ABCDEf" },
```
solardiz: Ideally, we'd have 2+ test vectors per type (so 8+ total). There are some kinds of errors that having only one test vector per type wouldn't catch but 2+ sometimes would.

src/encdatavault_fmt_plug.c
sylvainpelissier: I have used the commands in `CONTRIBUTING.md`:
```bash
$ indent -kr -i4 -ts4 -nlp -nbbo -ncs -l79 -lc79 -bad -il0
$ astyle --style=kr -t4 -U -H -p -xC79 -c -k3 -z2
```

run/encdatavault2john.py
sylvainpelissier: It was an error thank you.

src/encdatavault_fmt_plug.c
solardiz: @magnumripper has opened #4960 to fix that. You can change the 79 to 120 or 131.

src/encdatavault_fmt_plug.c
sylvainpelissier: Indeed I spotted a bug while adding tests, thank you !

src/encdatavault_fmt_plug.c
sylvainpelissier: `result` contains first the decrypted keys from the keychain (128 bytes) at line 322 and then I reuse the buffer to decrypt the data which is smaller.

solardiz: > result contains first the decrypted keys from the keychain (128 bytes) at line 322 and then I reuse the buffer to decrypt the data which is smaller.

You could want to add a source code comment on that.

src/encdatavault_fmt_plug.c
solardiz: FWIW, you still only have one test vector for each one prefix of `$1$3` and `$3$*`. Maybe adding another to each of those would uncover more bugs. ;-)

doc/README-ENCDataVault.md
```diff
@@ -0,0 +1,17 @@
...
+==============================
+
+1. Run encdatavault2john.py on vault folder.
+
+```bash
+$ ../run/encdatavault2john.py path/to/vault > hash
+```
+
+2. Run john on the file.
+```bash
```
solardiz, line 12: Missing empty line here.

solardiz (approved):
We're getting close. Can merge this as-is if you like.

src/encdatavault_fmt_plug.c
solardiz: BTW, there's also `--test-full=1` to run more elaborate tests than the default.

sylvainpelissier: I did not know `--test-full=1`, that's nice. I added the tests.

sylvainpelissier:
It should be fine now.

solardiz:
OK, we can merge this now.

I wonder, though: does matching on 32-bit value mean this will produce occasional false positives - cracked passwords sometimes not working for real? Or will those be genuine collisions - with all such passwords working?

In some formats with that (non-)issue, we set the `FMT_NOT_EXACT` flag, so that JtR keeps trying by default. OTOH, even when we don't, there's the `--keep-guessing` option.

Also, we probably misled you with comments on line wrapping length - the 120+ is only for source code. We still wrap at 79 max in the documentation, except for occasional lines that provide examples of lengthy command-lines, etc.

Oh, and separately, would you please submit sample inputs to the Python script to our https://github.com/openwall/john-samples repo? So that we can retest the Python script after possibly making changes to it later. Thank you!

solardiz:
BTW, we generally don't intend the `#pragma` lines (keep them to the left, like `#if` and such). I guess the automatic indentation didn't handle it that way.

solardiz:
> does matching on 32-bit value mean this will produce occasional false positives - cracked passwords sometimes not working for real? Or will those be genuine collisions - with all such passwords working?

The answer to the above would ideally be in the documentation added here.

solardiz:
Something's not quite right yet:

```
$ ../run/john -fuzz -form=ENCDataVault 
Fuzzing: ENCDataVault (1000 iterations) [MD5]... ENCDataVault: Warning: version 9 not supported, not loading such hashes!
ENCDataVault: Warning: algorithm id 9 not supported!
Segmentation fault
```

(in an `--enable-fuzz` build).

solardiz:
> Segmentation fault

OK, `valid()` is missing a check of version 3's extra field. I'm testing a fix now.

solardiz:
OK, this fixes it:

```diff
+++ b/src/encdatavault_fmt_plug.c
@@ -182,6 +182,7 @@ static int valid(char *ciphertext, struct fmt_main *self)
                goto err;
        if (!isdec(p))
                goto err;
+       int version = atoi(p);
        if (strcmp(p, "1") && strcmp(p, "3")) {
                static int warned;
 
@@ -208,6 +209,12 @@ static int valid(char *ciphertext, struct fmt_main *self)
                goto err;
        if (hexlenl(p, &extra) != ENC_SIG_SIZE * 2 || extra)
                goto err;
+       if (version == 3) {
+               if ((p = strtokm(NULL, "$")) == NULL)   // Keychain
+                       goto err;
+               if (hexlenl(p, &extra) != ENC_KEYCHAIN_SIZE * 2 || extra)
+                       goto err;
+       }
        MEM_FREE(keeptr);
        return 1;
 
```

BTW, not in this PR, but in general we should probably move away from that `strtokm()` approach - we fully check each field anyway, so can directly test the next character for `$` without any function call. We'd also be able to test that the string ends after the last expected field - something we can't(?) easily do with `strtokm()` now.

solardiz (changes requested):
Please make changes per my hopefully last set of comments - at least need to fix `valid()`.

solardiz:
With my `valid()` fix, also passes `-test-full=1` and `-fuzz` in an `--enable-asan` build.

sylvainpelissier:
> OK, we can merge this now.
> 
> I wonder, though: does matching on 32-bit value mean this will produce occasional false positives - cracked passwords sometimes not working for real? Or will those be genuine collisions - with all such passwords working?
> 
> In some formats with that (non-)issue, we set the `FMT_NOT_EXACT` flag, so that JtR keeps trying by default. OTOH, even when we don't, there's the `--keep-guessing` option.
> 
> Also, we probably misled you with comments on line wrapping length - the 120+ is only for source code. We still wrap at 79 max in the documentation, except for occasional lines that provide examples of lengthy command-lines, etc.
> 
> Oh, and separately, would you please submit sample inputs to the Python script to our https://github.com/openwall/john-samples repo? So that we can retest the Python script after possibly making changes to it later. Thank you!

Indeed false positives will make the solution to decrypt the files incorrectly and may crash it. I did not know the `FMT_NOT_EXACT` flag. It is really convenient. I have updated the code and the doc accordingly. Thank you.

doc/README-ENCDataVault.md
```diff
@@ -0,0 +1,20 @@
...
+
+2. Run john on the file.
+
+```bash
+$ ../run/john hash
+```
+
+The decryption test is done on a 32-bit magic value, meaning that occasional false positives can occur. Cracked passwords for those false positives will not decrypt the vault correctly.
+
+The plugin have beem tested on Sony ENCDataVault lite 6.2.13, SanDisk SecureAccess 3.02 and ENC DataVault 7.1.1W.
```
solardiz: Typo: `beem`. And we generally use the word "format" in the documentation.

Also, lines longer than 79, which we generally avoid in documentation (except for lengthy command-lines, URLs, etc.), although in an `.md` file it's arguably more acceptable.

src/encdatavault_fmt_plug.c
```diff
@@ -0,0 +1,431 @@
...
+
+	ctcopy = strdup(ciphertext);
+	keeptr = ctcopy;
+	ctcopy += FORMAT_TAG_LENGTH;
+	if ((p = strtokm(ctcopy, "$")) == NULL) // version
+		goto err;
+	if (!isdec(p))
+		goto err;
+	int version = atoi(p);
+	if (version < 1 || version > 3) {
```
solardiz: You allow version 2 now? Why?

solardiz (changes requested):
At least fix the typo and remove or explain the version 2 thing.

solardiz:
Suggested fix relative to yesterday's revision avoiding `isdec` and `atoi`:

```diff
+++ b/src/encdatavault_fmt_plug.c
@@ -180,8 +180,6 @@ static int valid(char *ciphertext, struct fmt_main *self)
        ctcopy += FORMAT_TAG_LENGTH;
        if ((p = strtokm(ctcopy, "$")) == NULL) // version
                goto err;
-       if (!isdec(p))
-               goto err;
        if (strcmp(p, "1") && strcmp(p, "3")) {
                static int warned;
 
@@ -189,6 +187,7 @@ static int valid(char *ciphertext, struct fmt_main *self)
                        fprintf(stderr, "%s: Warning: version %d not supported, not loading such hashes!\n", self->params.label, atoi(p));
                goto err;
        }
+       char version = *p;
        if ((p = strtokm(NULL, "$")) == NULL)   // algorithm id
                goto err;
        if (!isdec(p))
@@ -208,6 +207,12 @@ static int valid(char *ciphertext, struct fmt_main *self)
                goto err;
        if (hexlenl(p, &extra) != ENC_SIG_SIZE * 2 || extra)
                goto err;
+       if (version == '3') {
+               if ((p = strtokm(NULL, "$")) == NULL)   // Keychain
+                       goto err;
+               if (hexlenl(p, &extra) != ENC_KEYCHAIN_SIZE * 2 || extra)
+                       goto err;
+       }
        MEM_FREE(keeptr);
        return 1;
 
```

src/encdatavault_fmt_plug.c
sylvainpelissier: Sorry I was thinking of algo id again. I corrected it.

doc/README-ENCDataVault.md
```diff
@@ -0,0 +1,26 @@
...
+
+```bash
+$ ../run/john hash
+```
+
+The decryption test is done on a 32-bit magic value, meaning that occasional 
+false positives can occur. Cracked passwords for those false positives will not
+decrypt the vault correctly.
+
+The format have been tested on Sony ENCDataVault lite 6.2.13, SanDisk
```
solardiz: "has been"

sylvainpelissier:
> > OK, we can merge this now.
> > I wonder, though: does matching on 32-bit value mean this will produce occasional false positives - cracked passwords sometimes not working for real? Or will those be genuine collisions - with all such passwords working?
> > In some formats with that (non-)issue, we set the `FMT_NOT_EXACT` flag, so that JtR keeps trying by default. OTOH, even when we don't, there's the `--keep-guessing` option.
> > Also, we probably misled you with comments on line wrapping length - the 120+ is only for source code. We still wrap at 79 max in the documentation, except for occasional lines that provide examples of lengthy command-lines, etc.
> > Oh, and separately, would you please submit sample inputs to the Python script to our https://github.com/openwall/john-samples repo? So that we can retest the Python script after possibly making changes to it later. Thank you!
> 
> Indeed false positives will make the solution the file incorrectly and may crash it. I did not know the `FMT_NOT_EXACT` flag. It is really convenient. I have updated the code and the doc accordingly. Thank you.

I'll add the samples in the repository.

solardiz (approved):

solardiz:
Looks good to me and CI is happy again, so I'll merge.

> I'll add the samples in the repository.

Yes, please.

Thank you!

frank-dittrich:
I get a warning about version 0 not being supported when testing:
```
$ ./john --test --format=encdatavault 
Will run 8 OpenMP threads
Benchmarking: ENCDataVault (1000 iterations) [MD5]... (8xOMP) ENCDataVault: Warning: version 0 not supported, not loading such hashes!
DONE
Raw:	34560 c/s real, 4369 c/s virtual
```

magnumripper:
Oh, that's from the self-test's "mini fuzzer" for testing `valid()`. I'm creating a PR with this:
```diff
diff --git a/src/encdatavault_fmt_plug.c b/src/encdatavault_fmt_plug.c
index 440c5ef1b..2ca5cbbfb 100644
--- a/src/encdatavault_fmt_plug.c
+++ b/src/encdatavault_fmt_plug.c
@@ -181,7 +181,7 @@ static int valid(char *ciphertext, struct fmt_main *self)
                goto err;
        if (strcmp(p, "1") && strcmp(p, "3")) {
                static int warned;
-               if (!warned++)
+               if (!self_test_running && !warned++)
                        fprintf(stderr, "%s: Warning: version %d not supported, not loading such hashes!\n", self->params.label, atoi(p));
                goto err;
        }
```

PR #4955, opened 2021-12-21
tezos-opencl: Compute pkh with BLAKE2b on device, drop OpenMP

solardiz:
This has many commits with incremental improvements of tezos-opencl and the shared/shareable OpenCL files that it uses.

### DIFF
--- a/run/opencl/ed25519-donna/ed25519-hash.h
+++ b/run/opencl/ed25519-donna/ed25519-hash.h
@@ -1,3 +1,4 @@
+#if 0
 #include "../opencl_sha2_ctx.h"
 
 static void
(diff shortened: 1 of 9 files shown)


solardiz:
"All 91 formats passed self-tests!" on GTX 1080, so I hopefully haven't broken any shared code. CI is also happy. I'll merge.

magnumripper:
I just noticed tezos now fails on my 2080ti with 465.19.01
```
$ make -s kernel-cache-clean
$ ../run/john -test -form:tezos-opencl
Device 2: NVIDIA GeForce RTX 2080 Ti
Benchmarking: tezos-opencl, Tezos Key [PBKDF2-SHA512 OpenCL]... FAILED (cmp_all(1))
```
Reverting to f683f0f2e fixes it. I can try to see if I can workaround it when I get some time.

run/opencl/ed25519-donna/ed25519-hash.h
```diff
@@ -8,3 +9,44 @@ ed25519_hash(uint8_t *hash, const uint8_t *in, size_t inlen) {
...
+	A = SHA2_INIT_A;
+	B = SHA2_INIT_B;
+	C = SHA2_INIT_C;
+	D = SHA2_INIT_D;
+	E = SHA2_INIT_E;
+	F = SHA2_INIT_F;
+	G = SHA2_INIT_G;
+	H = SHA2_INIT_H;
+
+	/* Assume inlen is 32 */
```
magnumripper, line 29: Is it normally 32 in other (as in future) uses of this? If not, perhaps we should have them as two separate functions, with this being called something like `ed25519_hash32()` and not featuring an inlen argument. We could do so once it hits us but with some bad luck it might take some amount of head scratching before realizing this optimization is the culprit and `inlen` isn't honored at all.

magnumripper:
> I just noticed tezos now fails on my 2080ti with 465.19.01

All other formats pass BTW

solardiz:
> I just noticed tezos now fails on my 2080ti with 465.19.01

I guess it's a miscompile of the BLAKE2b code. Nasty. I wonder if we'll have to reintroduce the ability to compute it on host... I hope not.

run/opencl/ed25519-donna/ed25519-hash.h
solardiz: Yes, for this curve it will always be 32, so we won't have anything else in this directory. Makes sense to drop the `inlen` argument. I think no need for a separate function nor for a rename, we can just have it without `inlen` and that's it - enough to avoid surprise.

PR #4976, opened 2021-12-24
tezos-opencl: Compare pkh on device

solardiz:
This moves the pkh comparison from host to device, and transfers only one `uint` back most of the time.

As implemented, on a successful crack this re-transfers the entire zeroized `cracked` from host to device. Alternatively, we could re-transfer only one `uint` `*cracked`, but have the final kernel (re)set each other element of `cracked` each time it's run. This alternative approach would actually use more global memory bandwidth in the typical case of nothing getting cracked in a long time. However, it would also have the advantage of avoiding incorrect cracks in case something in `mem_final` on device side becomes non-zero by accident during a very long run. I have no clear preference here.

### DIFF
--- a/doc/NEWS
+++ b/doc/NEWS
@@ -247,6 +247,11 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
 
 - Added support for ENCDataVault and encdatavault2john.py script [sylvainpelissier; 2021]
 
+- Optimize tezos-opencl by computing ed25519_publickey() and BLAKE2b on-device,
+  removing the CPU bottleneck.  This work was funded by the Tezos Foundation.
+  [Solar; 2021]
+
+
 Major changes from 1.8.0-jumbo-1 (December 2014) to 1.9.0-jumbo-1 (May 2019):
 
 - Updated to 1.9.0 core, which brought the following relevant major changes:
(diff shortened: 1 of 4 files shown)


solardiz:
> avoiding incorrect cracks in case something in `mem_final` on device side becomes non-zero by accident during a very long run.

I've just implemented a workaround for that, will probably force-push soon. (Testing now.)

solardiz:
@magnumripper Would you test this on your 2080Ti, please? Before or after we merge? Thanks!

solardiz:
I thought `atomic_dec` was a good idea - set all/most bits rather than usually just one - however, it fails on one of two versions of Intel OpenCL I've tried. So I'll revert to `atomic_inc` like we use elsewhere.

solardiz:
I'm also testing this hack:

```diff
@@ -201,8 +201,9 @@ __kernel void pbkdf2_sha512_tezos_final(__global const crack_t *in, __constant t
        union {
                uchar uc[32];
                ulong u64[4];
+               ed25519_public_key pk;
        } sk;
-       ed25519_public_key pk;
+#define pk sk.pk
        uint idx = get_global_id(0);
        memcpy_macro(sk.u64, in[idx].hash, 4);
        for (int i = 0; i < 4; i++)
```

This could reduce register usage, I guess.

solardiz:
> This could reduce register usage, I guess.

No, in my test it didn't - still 168 on NVIDIA Pascal. So I won't be making this change now.

magnumripper:
> @magnumripper Would you test this on your 2080Ti, please? Before or after we merge? Thanks!

Will do right now

magnumripper:
Works fine, best speed of five runs now 506414 c/s, which IIRC is up a tad. EDIT: it's up from 501852 c/s so about 1% gain.

Also passes TS `-internal`.

PR #4977, opened 2021-12-24
Unicode formats with GPU-side encoding: Truncation matching

### LINKED ISSUE #4965: nt-opencl reports wrong password instead of implicitly cracked empty password
While working on #4964, I noticed that this command:

```
rm john.pot; LWS=64 GWS=2560 ./john -w=rockyou.txt passwords.txt -form=nt-opencl -dev=4 -mask='?w!' -v=5
```

correctly cracks Dark's password as `Password01!`, but also incorrectly cracks Administrator's empty password as:

```
$ tail -1 john.pot | od -tx1
0000000 24 4e 54 24 33 31 64 36 63 66 65 30 64 31 36 61
0000020 65 39 33 31 62 37 33 63 35 39 64 37 65 30 63 30
0000040 38 39 63 30 3a fa 74 76 61 72 70 21 0a
```

That is, the byte value 0xfa followed by the string `tvarp!`.

This is reproducible, including on a build from 20190619 that wasn't yet affected by #4964, as well as with our current code with that bug fixed.

The exact same issue (including the same wrong password) is seen when retesting on a different GPU and with auto-tuning enabled.

With the CPU format (`--format=nt`), this issue does not occur (Administrator's password is not cracked by the equivalent command).

My guess is that with the OpenCL format the password gets cracked for real, because it's empty and an under-full buffer just happened to have that, however in that case some stale candidate password (from a previous fill of the buffer?) gets reported. That wouldn't explain why the exact same wrong password is reported for different GWS figures, though. So that guess is at least incomplete, or wrong.

magnumripper:
The cracking modes can't afford exact rejecting based on multibyte candidate length, so simply allows eg. 81 bytes for a format that can process up to 27 Unicode characters.

The formats then truncate correctly (can also happen due to invalid UTF-8) and sometimes the truncated candidate ends up in "unintended" cracking.

When this happens, we must ensure that get_key() truncates the same way as the GPU code did, in order to report the correct candidate.

Closes #4965

### DIFF
--- a/src/opencl_DES_bs_plug.c
+++ b/src/opencl_DES_bs_plug.c
@@ -14,7 +14,7 @@
 #include "common.h"
 #include "opencl_DES_bs.h"
 #include "../run/opencl/opencl_DES_hst_dev_shared.h"
-#include "unicode.h"
+#include "options.h"
 #include "bt_interface.h"
 #include "mask_ext.h"
 #include "logger.h"
--- a/src/opencl_axcrypt2_fmt_plug.c
+++ b/src/opencl_axcrypt2_fmt_plug.c
@@ -26,7 +26,6 @@ john_register_one(&fmt_opencl_axcrypt2);
 #include "common.h"
 #include "formats.h"
 #include "options.h"
-#include "unicode.h"
 #include "opencl_common.h"
 #include "axcrypt_common.h"
 #define VERSION_2_SUPPORT 1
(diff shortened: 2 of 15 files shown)


magnumripper:
Still testing. I'm not yet sure the list of formats that got this treatment is complete or correct.

All pass self-test and TS

magnumripper:
I believe this is complete. I tested some of them confirming they had the same problem as NT in #4965 problem before this, and not after.

*If* the fix was added to some format that did not need it, it will do no harm anyway.

PR #4981, opened 2021-12-27
Zip & cryptoSafe OpenCL formats: Avoid a race condition

### LINKED ISSUE #4958: OpenCL: Sanity check return count
From https://github.com/openwall/john/pull/4943#issuecomment-998810528:

@solardiz:
> The host code blindly trusts the value of crack_count_ret, and may crash or worse if that value is out of bounds. Perhaps we should introduce some safety checks there. OpenCL devices and communication to them aren't necessarily as reliable as computation within one process on host, so it's wrong to unnecessarily make the latter less reliable.

Later discussion ended up with that we should bail with error as opposed to chug along with possibly b0rken environment.

### LINKED ISSUE #4975: OpenCL crack_count_ret race condition
The two OpenCL formats that use `crack_count_ret` are `ZIP-opencl` and `cryptosafe-opencl`. Both set `*crack_count_ret = 0;` when `get_global_id(0)` is 0. Both do it from the same OpenCL kernel that can also update that variable with `atomic_inc` or `atomic_max`.

We assume `get_global_id(0)` of 0 will be tried before all others, or at least no later than others. I think in practice it's typically part of the first batch along with many others, so when the first `atomic_*` update is possibly reached the `*crack_count_ret = 0;` would have already been issued.

However, there are issues with that:

1. I doubt the assumption above is guaranteed by OpenCL spec, or is it?

2. While the updates are atomic and are such that it doesn't matter which of them occurs first (if there's more than one), I don't know if the write of 0 is guaranteed to be atomic by the OpenCL spec - but maybe it is.

3. Even if the write is atomic and is issued before the first `atomic_*` is reached, that doesn't imply it completes before the `atomic_*`. Atomicity of operations isn't the same as guaranteeing their order of completion, even if they're issued in the desired order.

To get around these issues, maybe we should perform the zeroing on host or in a prior kernel invocation, not in the same kernel invocation that updates the variable.

Alternatively, maybe we don't need to zero this variable at all (except initially, to avoid using it uninitialized). We can accept it as ever-growing across multiple `crypt_all` calls, and on host use the difference between its current and previous value. Since it's unsigned, we can let it wrap and the difference would still be correct.

Edit: I think inserting a global memory write and read barrier(s) after the `*crack_count_ret = 0;` would address issues 2 and 3, but not issue 1. So if it is not guaranteed that `get_global_id(0)` of 0 is always part of the first batch, then it's not a sufficient fix.

magnumripper:
The crack return variable is atomically incremented on GPU. Reset it on CPU side to avoid a race condition.  Closes #4975.

Also adds sanity checking, closes #4958.

### DIFF
--- a/src/opencl_cryptosafe_fmt_plug.c
+++ b/src/opencl_cryptosafe_fmt_plug.c
@@ -103,6 +103,8 @@ static void create_clobj(size_t gws, struct fmt_main *self)
 
 	cl_crack_count_ret = clCreateBuffer(context[gpu_id], CL_MEM_READ_WRITE, sizeof(int), NULL, &ret_code);
 	HANDLE_CLERROR(ret_code, "Error creating device buffer");
+	crack_count_ret = 0;
+	HANDLE_CLERROR(clEnqueueWriteBuffer(queue[gpu_id], cl_crack_count_ret, CL_FALSE, 0, sizeof(cl_uint), &crack_count_ret, 0, NULL, NULL), "Failed resetting crack return");
 
 	pinned_int_key_loc = clCreateBuffer(context[gpu_id], CL_MEM_READ_ONLY | CL_MEM_ALLOC_HOST_PTR, sizeof(cl_uint) * gws, NULL, &ret_code);
 	HANDLE_CLERROR(ret_code, "Error creating page-locked memory pinned_int_key_loc.");
@@ -339,9 +341,7 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 
 	*pcount *= mask_int_cand.num_int_cand;
 
-	//printf("%s(%d) lws "Zu" gws "Zu" kidx %u k %d mult %u\n", __FUNCTION__, count, *lws, gws, key_idx, new_keys, mask_int_cand.num_int_cand);
-
-	if (new_keys || ocl_autotune_running) {
+	if (new_keys) {
 		/* Self-test kludge */
 		if (idx_offset > 4 * (gws + 1))
 			idx_offset = 0;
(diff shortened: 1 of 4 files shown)


src/opencl_cryptosafe_fmt_plug.c
```diff
@@ -337,11 +339,9 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 	size_t *lws = local_work_size ? &local_work_size : NULL;
 	size_t gws = GET_NEXT_MULTIPLE(count, local_work_size);
 
-	*pcount *= mask_int_cand.num_int_cand;
+	count *= mask_int_cand.num_int_cand;
```
solardiz: Don't we actually need to update `*pcount`? This change looks like a bug to me.

src/opencl_cryptosafe_fmt_plug.c
```diff
@@ -337,11 +339,9 @@ static int crypt_all(int *pcount, struct db_salt *salt)
...
 
-	//printf("%s(%d) lws "Zu" gws "Zu" kidx %u k %d mult %u\n", __FUNCTION__, count, *lws, gws, key_idx, new_keys, mask_int_cand.num_int_cand);
-
-	if (new_keys || ocl_autotune_running) {
+	if (new_keys) {
```
solardiz, line 344: I think these changes would better be in their own PR, although that would need testing and I understand it could be handy for you to make them for these two formats now because you retest them now anyway.

solardiz (commented):
I think you need to revert the `*pcount` update.

src/opencl_cryptosafe_fmt_plug.c
magnumripper: No, `*pcount` isn't used below this. At first, my sanity check (using `count` which didn't account for mask factor) triggered (caught only by TS) and when I looked into that I thought this change was cleaner.

magnumripper: Hmm wait, `*pcount` can be used even though we don't return it, right?

magnumripper: Yup, fixing that

solardiz: Updated `*pcount` is what tells the caller how many candidates we actually processed (including built-in mask). As to your check, you should either also write the updated value into `count` here or use `*pcount` in the check.

solardiz (approved):

PR #4982, opened 2021-12-27
OpenCL formats: New keys review

### LINKED ISSUE #4932: Review and make consistent the new_keys logic in OpenCL formats

magnumripper:
First commit drops the obsolete `ocl_autotune_running` check from `new_keys` logic.

Second commit adds `new_keys` logic to a bunch of formats that lacked it. Some of them may be so slow it's fairly pointless but it doesn't hurt anyway.

All tests out fine now. I reverted adding `new_keys` logic to pwsafe-opencl because it made it fail TS. I have yet to understand why.

Closes #4932

### DIFF
--- a/src/opencl_7z_fmt_plug.c
+++ b/src/opencl_7z_fmt_plug.c
@@ -316,17 +316,16 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 	int index;
 	size_t *lws = local_work_size ? &local_work_size : NULL;
 
-	//fprintf(stderr, "%s(%d) lws %zu gws %zu\n", __FUNCTION__, count, local_work_size, global_work_size);
-
 	if (any_cracked) {
 		memset(cracked, 0, cracked_size);
 		any_cracked = 0;
 	}
 
-	if (ocl_autotune_running || new_keys) {
-		int i;
+	global_work_size = GET_NEXT_MULTIPLE(count, local_work_size);
 
-		global_work_size = GET_NEXT_MULTIPLE(count, local_work_size);
+	/* Note: This format is effectively unsalted */
+	if (new_keys) {
+		int i;
 
 		// Copy data to gpu
 		BENCH_CLERROR(clEnqueueWriteBuffer(queue[gpu_id], mem_in, CL_FALSE, 0,
@@ -353,9 +352,9 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 		BENCH_CLERROR(clEnqueueNDRangeKernel(queue[gpu_id], sevenzip_final, 1,
 			NULL, &global_work_size, lws, 0, NULL, multi_profilingEvent[3]),
 			"Run final loop kernel");
-	}
 
-	new_keys = 0;
+		new_keys = 0;
+	}
 
 	if (sevenzip_trust_padding || sevenzip_salt->type == 0x80) {
 		// Run AES kernel (only for truncated hashes)
--- a/src/opencl_rawsha512_fmt_plug.c
+++ b/src/opencl_rawsha512_fmt_plug.c
@@ -71,7 +71,7 @@ typedef struct {
 
 static sha512_key *gkey;
 static sha512_hash *ghash;
-static uint8_t sha512_key_changed;
+static uint8_t new_keys;
 static uint8_t hash_copy_back;
 
 //OpenCL variables:
@@ -208,7 +208,7 @@ static void set_key(char *key, int index)
 		length = PLAINTEXT_LENGTH;
 	gkey[index].length = length;
 	memcpy(gkey[index].v, key, length);
-	sha512_key_changed = 1;
+	new_keys = 1;
 }
 
 static char *get_key(int index)
@@ -306,21 +306,21 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 
 	global_work_size = GET_NEXT_MULTIPLE(count, local_work_size);
 
-	///Copy data to GPU memory
-	if (sha512_key_changed || ocl_autotune_running) {
+	// Copy data to GPU memory
+	if (new_keys) {
 		BENCH_CLERROR(clEnqueueWriteBuffer
 		    (queue[gpu_id], mem_in, CL_FALSE, 0, insize, gkey, 0, NULL,
 			multi_profilingEvent[0]), "Copy memin");
+
+		new_keys = 0;
 	}
 
-	///Run kernel
+	// Run kernel
 	BENCH_CLERROR(clEnqueueNDRangeKernel
 	    (queue[gpu_id], crypt_kernel, 1, NULL, &global_work_size, lws,
 		0, NULL, multi_profilingEvent[1]), "Set ND range");
 
-	/// Reset key to unchanged and hashes uncopy to host
-	sha512_key_changed = 0;
-    hash_copy_back = 0;
+	hash_copy_back = 0;
 
 	return count;
 }
--- a/src/opencl_xsha512_fmt_plug.c
+++ b/src/opencl_xsha512_fmt_plug.c
@@ -81,7 +81,7 @@ typedef struct {
 static xsha512_key *gkey;
 static xsha512_hash *ghash;
 static xsha512_salt gsalt;
-static uint8_t xsha512_key_changed;
+static uint8_t new_keys;
 static uint8_t hash_copy_back;
 
 //OpenCL variables:
@@ -228,7 +228,7 @@ static void set_key(char *key, int index)
 		length = PLAINTEXT_LENGTH;
 	gkey[index].length = length;
 	memcpy(gkey[index].v, key, length);
-	xsha512_key_changed = 1;
+	new_keys = 1;
 }
 
 static char *get_key(int index)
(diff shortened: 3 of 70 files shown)


magnumripper:
> I reverted adding new_keys logic to pwsafe-opencl because it made it fail TS. I have yet to understand why.

I now found out why - the input buffer is modified by the init kernel, and holds state. I'll see if I can work around it - the format is a fairly fast one.

magnumripper:
> > I reverted adding new_keys logic to pwsafe-opencl because it made it fail TS. I have yet to understand why.
> 
> I now found out why - the input buffer is modified by the init kernel, and holds state. I'll see if I can work around it - the format is a fairly fast one.

Fixed now. I made it a separate commit.

src/opencl_rawsha512_fmt_plug.c
```diff
@@ -307,7 +307,7 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 	global_work_size = GET_NEXT_MULTIPLE(count, local_work_size);
 
 	///Copy data to GPU memory
-	if (sha512_key_changed || ocl_autotune_running) {
+	if (sha512_key_changed) {
```
solardiz: Why don't you rename `sha512_key_changed` in `opencl_rawsha512_fmt_plug.c` like you do in `opencl_xsha512_fmt_plug.c`?

src/opencl_7z_fmt_plug.c
```diff
@@ -316,17 +316,16 @@ static int crypt_all(int *pcount, struct db_salt *salt)
...
 		memset(cracked, 0, cracked_size);
 		any_cracked = 0;
 	}
 
-	if (ocl_autotune_running || new_keys) {
-		int i;
+	global_work_size = GET_NEXT_MULTIPLE(count, local_work_size);
 
-		global_work_size = GET_NEXT_MULTIPLE(count, local_work_size);
+	/* Note: This format is effectively unsalted */
```
solardiz, line 326: Why is 7z-opencl "effectively unsalted"?

src/opencl_rawsha512_fmt_plug.c
magnumripper: I simply overlooked that one

src/opencl_7z_fmt_plug.c
magnumripper: There is a salt field but the salt length is always 0. As long as the iteration count did not change, we don't have to recalculate the decryption keys for new new non-hash but can skip all the way to the AES checking kernel.

IIRC we recently added some safety checks, so the CPU format *will* handle a future salt correctly while the OpenCL format will reject and warn.

Also, `set_salt()` does set `new_keys` if the iteration count changed.

src/opencl_rawsha512_fmt_plug.c
```diff
@@ -306,21 +306,22 @@ static int crypt_all(int *pcount, struct db_salt *salt)
...
+	// Run kernel
 	BENCH_CLERROR(clEnqueueNDRangeKernel
 	    (queue[gpu_id], crypt_kernel, 1, NULL, &global_work_size, lws,
 		0, NULL, multi_profilingEvent[1]), "Set ND range");
 
-	/// Reset key to unchanged and hashes uncopy to host
-	sha512_key_changed = 0;
-    hash_copy_back = 0;
+	// Reset hashes uncopy to host [SIC]
+	hash_copy_back = 0;
```
solardiz, line 323: I don't understand what is meant by "// Reset hashes uncopy to host [SIC]", so this comment is unhelpful - it should either be dropped or edited so that it'd clarify things.

src/opencl_xsha512_fmt_plug.c
```diff
@@ -369,7 +369,7 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 		0, NULL, multi_profilingEvent[1]), "Set ND range");
 
 	/// Reset key to unchanged and hashes uncopy to host
-	xsha512_key_changed = 0;
+	new_keys = 0;
     hash_copy_back = 0;
```
solardiz: Perhaps you want to move the `new_keys = 0;` line like you did in `opencl_rawsha512_fmt_plug.c`. And there's the confusing comment about "uncopy" here, too. Please keep these in sync.

magnumripper: Fixed now

magnumripper:
I did only a single relbench run each for "before" and "after".

No format sticks out as being significantly slower (would indicate some mistake).

Presumably only "Many salts" speeds should reflect adding new_keys logic, and of them only GPG (14% boost) and RAKP (172% boost) does stick out as being significantly faster.

Or could even "Raw" speed reflect this? I'm not sure I understand the impact of the 0x100 and 0x500 flags... For example, ssh has 0x107 and a boost of 72%, o5logon also has 0x107 and a boost of 52% and there's a few more boosted by more than 5%. Then again, with only a single run of each version, there could be autotune differences (such as with raw-md5 which wasn't touched but showed an 88% boost).

magnumripper:
Perhaps we should introduce a way to override benchmark_length? Maybe just using an environment variable.

solardiz:
> Or could even "Raw" speed reflect this?

Yes, for salted formats that use 0x107 (rather than 0x507) the reported Raw speeds is for many salts.

> there could be autotune differences

We could want to introduce a way to have auto-tuned per-format LWS/GWS from one run saved to a file and then reused for another run, specifically for use with `relbench`. Only have these restored values take effect for the same format(s) and only if the device name is the same (e.g., "GeForce GTX 1080"). Then this would also be usable as an alternative to our current copy-paste, with some defense against misuse.

> Perhaps we should introduce a way to override benchmark_length? Maybe just using an environment variable.

For the actual length, we already can override it with `--mask`. For the flags, maybe, but wouldn't this be more confusing than useful?

PR #4995, opened 2022-01-11
Assorted quick fixes and workarounds

solardiz:
This fixes or works around 5 issues referenced in the 5 commits.

### DIFF
--- a/run/john.conf
+++ b/run/john.conf
@@ -141,10 +141,10 @@ PotFilePermissions = 0600
 # Default is N
 IgnoreChmodErrors = N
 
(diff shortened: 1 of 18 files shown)


solardiz:
These changes are obvious enough that they got to work, but I didn't test them much. Maybe we should test them more.

magnumripper:
I can't see a lot to test there, should be fine

solardiz:
> I can't see a lot to test there, should be fine

Well, for example the different order of OpenCL format tests in all-format `--test` could make a previously hidden issue show up... but even if so, we'd need to deal with that separately, not revert the change here.

PR #4998, opened 2022-01-13
Pipe mode bugfix

### LINKED ISSUE #4923: --pipe truncates lines sometimes
Working on #4921 @magnumripper mentioned "a race between fgetl() and a growing buffer so lines could be truncated" and possible affection of `--pipe`. The problem turned out true.

(`$.` is current line number in Perl. I print it to see that input passes through. It is counted from 1 while the first line is "0aa...".)
```
$ perl -le 'print $a++ . "a" x 100 while 1' | ./run/john --pipe --stdout | perl -lne 'print $. if $. % 1000000 == 0; print "wrong $. >$_<" if !/^\d+a{100}$/'
[...]
39000000
wrong 39505289 >39505288aaaaaaaaaaaaaaa<
wrong 39505290 ><
40000000
[...]
59000000
wrong 59206966 ><
wrong 59206967 ><
60000000
[...]
```
The numbers repeat on reruns, so it seems deterministic.

```
$ ./run/john
John the Ripper 1.9.0-jumbo-1+bleeding-5d450f267 2021-12-03 21:27:35 +0100 [linux-gnu 64-bit x86_64 SSE4.2 AC]
```

magnumripper:
The "last word" logic would thrash the memory buffer.

Closes #4923

### DIFF
--- a/src/wordlist.c
+++ b/src/wordlist.c
@@ -1055,6 +1055,7 @@ REDO_AFTER_LMLOOP:
 		prerule = rpp_next(&ctx);
 
 /* A string that can't be produced by fgetl(). */
+	last = aligned.buffer[1];
 	last[0] = '\n';
 	last[1] = 0;
 


solardiz:
Hmm, those writes into `last` come from core. Are they somehow only problematic with pipe mode (non-existent in core)?

BTW, there's a long-standing core bug I'm aware of: the check against `last` isn't doing the right thing when an external mode filter modifies the word. This appears non-trivial to fix while keeping the current `last` logic. Maybe something to avoid in a rewrite of that logic.

magnumripper:
> Hmm, those writes into `last` come from core. Are they somehow only problematic with pipe mode (non-existent in core)?

Sure. Pipe mode re-enters that write for "next batch" using goto.

For a while I thought wordlist.c should be rewritten from a pristine core file. Now I'm leaning towards rewriting it from scratch, with all features considered from start.

solardiz:
> For a while I thought wordlist.c should be rewritten from a pristine core file. Now I'm leaning towards rewriting it from scratch, with all features considered from start.

Right. I think our next release should be 2.0.0 based off jumbo, but no longer with jumbo in its name. We might not even bother to release an updated core - maybe 1.9.0 will remain the last released core.

Edit: Oh, but I didn't mean needing to include a rewritten `wordlist.c` specifically in 2.0.0 - that should probably wait until post-2.0 development. For 2.0.0, let's just continue to fix or workaround the most pressing issues like we've been doing lately.

Issue #4781, opened 2021-08-09
Nvidia busy-wait workaround


asaidac:
Hello,

Not sure if the JtR is still maintained - but I'll try anyway.
I have a headless 4 GPUs rig, with a 4 physical cores CPU

Using OpenCL "-dev=1,2,3,4" where -dev are the GPU's assigned numbers by JtR.
The GPU is working well - the problem is the CPU.

During the run all 4 cores are loaded to cca. 100%.
I know that part of the job is done by CPU - but not a that kind of load.

I am suspecting that 
cudaThreadSynchronize() is effectively a spin lock which polls the GPU at rather high frequency, waiting until the GPU kernel is finished. Because the CPU thread is just sitting in a polling loop, it actually isn’t doing much work - just using unnecessary power.

Any ideas how to fix this?

Thanks 

### Checklist ###
- [ ] :1st_place_medal: I've read and understood these instrucions;
  - This is not a support forum, it's a bug tracker. For questions and support, review postings on the [john-users mailing list](https://www.openwall.com/lists/john-users/).
- [ ] :thumbsup: I've tested using latest bleeding version from this repository.
  - Be clear about your environment and what you are doing. Share a sample hash or file that can be used to reproduce.
- [ ] :confused: I'm confused and I need guidance.
  - Please, read the instructions at https://www.openwall.com/john/#lists, then join the list before posting.

### IMPORTANT ###
We expect only reports of issues with the latest revision found in this GitHub repository. We do not expect in here, and have little use for, reports of issues only seen in a release or in a distro package.

Attach details about your OS and about john, including:
- The output of `./john --list=build-info`.
- The command line you are using.

ghost:
If you are using NVIDIA (it seems to be the case), this is a known issue (examples):
- https://forums.developer.nvidia.com/t/opencl-busy-wait-still-not-fixed/46441
- https://github.com/openmm/openmm/issues/1541

BTW: the workaround looks ugly.

ghost:
@solardiz this sounds bad:

> Not sure if the JtR is still maintained.

We are old and tired but not dead.
****
BTW: I think we will close the issue shortly, it is known and it is "by design" (?) from NVIDIA itself.

asaidac:
Hello
I know about Nvidia OpenCL problem.
My question - if JtR is still maintained - is why not moving to CUDA (since
currently OpenCL is deprecated.)
BTW there is a piece of code that can even fix the OpenCL problem if anyone
cares about it

[...]

PR #5006, opened 2022-01-19
Avoid busy wait

magnumripper:
Move the macros for avoiding busy-waits to shared header. Use it for 7z-opencl. I intend to add more formats to this PR before merging.

~Because of different code path/kernels used depending on salt, I added a couple of `clFinish()` and also changed a blocking `clEnqueueReadBuffer()` to a non-blocking one with an explicit `clFinish()` after. I think it's all sensible but would appreciate comments.~

Looking at virtual figures, they went from more or less 1:1 to eg. 26850 c/s real, 139264 c/s virtual (for a 3-second run) and I presume the latter should mean about 80% decrease of busy-waiting, right?

### DIFF
--- a/src/bench.c
+++ b/src/bench.c
@@ -345,7 +345,6 @@ char *benchmark_format(struct fmt_main *format, int salts,
 {
 	static void *binary = NULL;
 	static int binary_size = 0;
-	static char s_error[128];
 	static int wait_salts = 0;
 	struct fmt_tests *current;
 	int pass;
@@ -363,7 +362,7 @@ char *benchmark_format(struct fmt_main *format, int salts,
 #ifndef BENCH_BUILD
 	struct db_salt *two_salts_db[2];
 	unsigned int t_cost[2][FMT_TUNABLE_COSTS];
-	int ntests, pruned, i;
+	int i;
 #endif
 	int salts_done = 0;
 	int wait = 0;
@@ -377,43 +376,6 @@ char *benchmark_format(struct fmt_main *format, int salts,
 	benchmark_running = 1;
 #ifndef BENCH_BUILD
 	dyna_salt_init(format);
-
-	pruned = 0;
-	for (i = 0; i < FMT_TUNABLE_COSTS; i++)
-	if (options.loader.min_cost[i] > 0 ||
-	    options.loader.max_cost[i] < UINT_MAX) {
-		unsigned int cost;
-
-		if (format->methods.tunable_cost_value[i] == NULL) {
-			sprintf(s_error,
-			        "FAILED (cost %d not defined for format)\n", i);
-			return s_error;
-		}
-
-		ntests = 0;
-		current = format->params.tests;
-		while ((current++)->ciphertext)
-			ntests++;
-
-		current = format->params.tests;
-		for (index = 0; index < ntests; index++) {
-			cost = get_cost(format, index, i);
-			if (cost >= options.loader.min_cost[i] &&
-			    cost <= options.loader.max_cost[i])
-				memcpy(current++,
-				       &format->params.tests[index],
-				       sizeof(struct fmt_tests));
-			else
-				pruned++;
-		}
-		memset(current, 0, sizeof(struct fmt_tests));
-	}
-
-	if (pruned && !format->params.tests->ciphertext) {
-		sprintf(s_error, "FAILED (--cost pruned all %d test vectors)\n",
-		        pruned);
-		return s_error;
-	}
 
 	current = format->params.tests;
 #endif
@@ -714,10 +676,10 @@ void gather_results(struct bench_results *results)
 int benchmark_all(void)
 {
 	struct fmt_main *format;
+	int i;
 #if defined(HAVE_OPENCL)
 	char s_gpu[16 * MAX_GPU_DEVICES] = "";
 	char s_gpu1[16 * MAX_GPU_DEVICES] = "";
-	int i;
 #else
 	const char *s_gpu = "";
 	const char *s_gpu1 = "";
@@ -931,6 +893,45 @@ AGAIN:
 		if (options.flags & FLG_MASK_CHK)
 			mask_init(test_db, options.mask);
 
+		/* Prune test vectors for benchmark to honor --cost option */
+		struct fmt_tests *current;
+		int index, ntests, pruned = 0;
+
+		for (i = 0; i < FMT_TUNABLE_COSTS; i++) {
+			if (options.loader.min_cost[i] > 0 || options.loader.max_cost[i] < UINT_MAX) {
+				unsigned int cost;
+
+				if (format->methods.tunable_cost_value[i] == NULL) {
+					printf("FAILED (cost %d not defined for format)\n\n", i);
+					failed++;
+					format->methods.done();
+					goto next;
+				}
+
+				ntests = 0;
+				current = format->params.tests;
+				while ((current++)->ciphertext)
+					ntests++;
+
+				current = format->params.tests;
+				for (index = 0; index < ntests; index++) {
+					cost = get_cost(format, index, i);
+					if (cost >= options.loader.min_cost[i] && cost <= options.loader.max_cost[i])
+						memcpy(current++, &format->params.tests[index], sizeof(struct fmt_tests));
+					else
+						pruned++;
+				}
+				memset(current, 0, sizeof(struct fmt_tests));
+			}
+		}
+
+		if (pruned && !format->params.tests->ciphertext) {
+			printf("FAILED (--cost pruned all %d test vectors)\n\n", pruned);
+			failed++;
+			format->methods.done();
+			goto next;
+		}
+
 		/*
 		 * Re-init for benchmark.  While the self-tests were done with very
 		 * low work sizes, we now need a proper auto-tune for benchmark, with
@@ -938,10 +939,10 @@ AGAIN:
 		 */
 		format->methods.reset(test_db);
 #endif
-		if ((result = benchmark_format(format, salts,
-		                               &results_m, test_db))) {
+		if ((result = benchmark_format(format, salts, &results_m, test_db))) {
 			puts(result);
 			failed++;
+			format->methods.done();
 			goto next;
 		}
 #if HAVE_OPENCL
--- a/src/opencl_7z_fmt_plug.c
+++ b/src/opencl_7z_fmt_plug.c
@@ -337,16 +337,22 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 			NULL, &global_work_size, lws, 0, NULL, multi_profilingEvent[1]),
 			"Run init kernel");
 
+		// Better precision for WAIT_ macros
+		BENCH_CLERROR(clFinish(queue[gpu_id]), "clFinish");
+
 		// Run loop kernel
+		WAIT_INIT(global_work_size)
 		for (i = 0; i < (ocl_autotune_running ? 1 : LOOP_COUNT); i++) {
 			BENCH_CLERROR(clEnqueueNDRangeKernel(queue[gpu_id],
 				crypt_kernel, 1, NULL, &global_work_size, lws, 0,
 				NULL, multi_profilingEvent[2]),
 				"Run loop kernel");
-			BENCH_CLERROR(clFinish(queue[gpu_id]),
-				"Error running loop kernel");
+			WAIT_SLEEP
+			BENCH_CLERROR(clFinish(queue[gpu_id]), "Error running loop kernel");
+			WAIT_UPDATE
 			opencl_process_event();
 		}
+		WAIT_DONE
 
 		// Run final kernel
 		BENCH_CLERROR(clEnqueueNDRangeKernel(queue[gpu_id], sevenzip_final, 1,
--- a/src/opencl_common.h
+++ b/src/opencl_common.h
@@ -42,6 +42,7 @@
 #include "common.h"
 #include "formats.h"
 #include "path.h"
+#include "timer.h"
 #include "../run/opencl/opencl_device_info.h"
 
 #define MAX_PLATFORMS   8
@@ -329,6 +330,41 @@ void opencl_process_event(void);
 
 #define GET_EXACT_MULTIPLE(dividend, divisor)	  \
 	(divisor) ? ((dividend > divisor) ? ((dividend / divisor) * divisor) : divisor) : dividend
+
+#define WAIT_INIT(work_size) { \
+	static uint64_t wait_last_work_size; \
+	int wait_sleep = (work_size >= wait_last_work_size); \
+	wait_last_work_size = work_size; \
+	static uint64_t wait_times[20], wait_min; \
+	static unsigned int wait_index; \
+	uint64_t wait_start = 0;
+#define WAIT_SLEEP \
+	if (gpu_nvidia(device_info[gpu_id])) { \
+		wait_start = john_get_nano(); \
+		uint64_t us = wait_min >> 10; /* 2.4% less than min */ \
+		if (wait_sleep && us >= 1000) \
+			usleep(us); \
+	}
+#define WAIT_UPDATE \
+	if (gpu_nvidia(device_info[gpu_id])) { \
+		uint64_t wait_new = john_get_nano() - wait_start; \
+		if (wait_new < wait_min && wait_new < wait_min * 1000 / 1012) /* 1.2% less than min */ \
+			wait_new = wait_new * 7 / 8; /* we might have overslept and don't know by how much */ \
+		if (wait_times[wait_index] == wait_min) { /* about to replace former minimum */ \
+			unsigned int i; \
+			wait_times[wait_index] = wait_min = ~(uint64_t)0; \
+			for (i = 0; i < sizeof(wait_times) / sizeof(wait_times[0]); i++) \
+				if (wait_times[i] < wait_min) \
+					wait_min = wait_times[i]; \
+		} \
+		wait_times[wait_index++] = wait_new; \
+		if (wait_index >= sizeof(wait_times) / sizeof(wait_times[0])) \
+			wait_index = 0; \
+		if (wait_new < wait_min) \
+			wait_min = wait_new; \
+		wait_sleep = 1; \
+	}
+#define WAIT_DONE }
 
 /*
  * Shared function to find 'the best' local work group size.
--- a/src/opencl_dmg_fmt_plug.c
+++ b/src/opencl_dmg_fmt_plug.c
@@ -499,28 +499,54 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 
 	// Copy data to gpu
 	if (new_keys) {
+		WAIT_INIT(global_work_size)
+
 		BENCH_CLERROR(clEnqueueWriteBuffer(queue[gpu_id], mem_in, CL_FALSE, 0, key_buf_size, inbuffer, 0, NULL, multi_profilingEvent[0]), "Copy data to gpu");
+
+		BENCH_CLERROR(clFlush(queue[gpu_id]), "failed in clFlush");
+		WAIT_SLEEP
+		BENCH_CLERROR(clFinish(queue[gpu_id]), "failed in clFinish");
+		WAIT_UPDATE
+		WAIT_DONE
+
 		new_keys = 0;
 	}
 
 	// Run kernels
 	BENCH_CLERROR(clEnqueueNDRangeKernel(queue[gpu_id], pbkdf2_init, 1, NULL, &global_work_size, lws, 0, NULL, multi_profilingEvent[1]), "Run initial kernel");
 
+	BENCH_CLERROR(clFinish(queue[gpu_id]), "Error running init kernel");
+
+	WAIT_INIT(global_work_size)
 	for (j = 0; j < (ocl_autotune_running ? 1 : ((cur_salt->pbkdf2.outlen + 19) / 20)); j++) {
 		for (i = 0; i < (ocl_autotune_running ? 1 : LOOP_COUNT); i++) {
 			BENCH_CLERROR(clEnqueueNDRangeKernel(queue[gpu_id], pbkdf2_loop, 1, NULL, &global_work_size, lws, 0, NULL, multi_profilingEvent[2]), "Run loop kernel");
+
+			WAIT_SLEEP
 			BENCH_CLERROR(clFinish(queue[gpu_id]), "Error running loop kernel");
+			WAIT_UPDATE
+
 			opencl_process_event();
 		}
 
 		BENCH_CLERROR(clEnqueueNDRangeKernel(queue[gpu_id], pbkdf2_final, 1, NULL, &global_work_size, lws, 0, NULL, multi_profilingEvent[3]), "Run final pbkdf2 kernel");
 	}
+	BENCH_CLERROR(clFinish(queue[gpu_id]), "Error running final pbkdf2 kernel");
+	WAIT_DONE
+
+	WAIT_INIT(global_work_size)
 
 	// DMG post-processing
 	BENCH_CLERROR(clEnqueueNDRangeKernel(queue[gpu_id], dmg_final[cur_salt->headerver], 1, NULL, &scalar_gws, lws, 0, NULL, multi_profilingEvent[4]), "Run final dmg kernel");
 
 	// Read the result back
-	BENCH_CLERROR(clEnqueueReadBuffer(queue[gpu_id], mem_out, CL_TRUE, 0, sizeof(dmg_out) * scalar_gws, output, 0, NULL, multi_profilingEvent[5]), "Copy result back");
+	BENCH_CLERROR(clEnqueueReadBuffer(queue[gpu_id], mem_out, CL_FALSE, 0, sizeof(dmg_out) * scalar_gws, output, 0, NULL, multi_profilingEvent[5]), "Copy result back");
+
+	BENCH_CLERROR(clFlush(queue[gpu_id]), "failed in clFlush");
+	WAIT_SLEEP
+	BENCH_CLERROR(clFinish(queue[gpu_id]), "failed in clFinish");
+	WAIT_UPDATE
+	WAIT_DONE
 
 	return count;
 }
--- a/src/opencl_hash_check_128_plug.c
+++ b/src/opencl_hash_check_128_plug.c
@@ -429,9 +429,18 @@ int ocl_hc_128_extract_info(struct db_salt *salt, void (*set_kernel_args)(void),
 		set_kernel_args_kpc();
 	}
 
+	BENCH_CLERROR(clFinish(queue[gpu_id]), "clFinish");
+	WAIT_INIT(gws)
+
 	BENCH_CLERROR(clEnqueueNDRangeKernel(queue[gpu_id], crypt_kernel, 1, NULL, &gws, lws, 0, NULL, multi_profilingEvent[2]), "failed in clEnqueueNDRangeKernel");
 
-	BENCH_CLERROR(clEnqueueReadBuffer(queue[gpu_id], buffer_hash_ids, CL_TRUE, 0, sizeof(cl_uint), hash_ids, 0, NULL, multi_profilingEvent[3]), "failed in reading back num cracked hashes.");
+	BENCH_CLERROR(clEnqueueReadBuffer(queue[gpu_id], buffer_hash_ids, CL_FALSE, 0, sizeof(cl_uint), hash_ids, 0, NULL, multi_profilingEvent[3]), "failed in reading back num cracked hashes.");
+
+	BENCH_CLERROR(clFlush(queue[gpu_id]), "failed in clFlush");
+	WAIT_SLEEP
+	BENCH_CLERROR(clFinish(queue[gpu_id]), "failed in clFinish");
+	WAIT_UPDATE
+	WAIT_DONE
 
 	if (hash_ids[0] > num_loaded_hashes) {
 		fprintf(stderr, "Error, crypt_all kernel.\n");
--- a/src/opencl_keepass_fmt_plug.c
+++ b/src/opencl_keepass_fmt_plug.c
@@ -63,7 +63,7 @@ static size_t insize, outsize, saltsize;
 #define STEP			0
 #define SEED			256
 
-#define HASH_LOOPS		100
+#define HASH_LOOPS		1000
 
 #define LOOP_COUNT		((keepass_salt->key_transf_rounds + HASH_LOOPS - 1) / HASH_LOOPS)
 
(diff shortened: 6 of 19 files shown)


magnumripper:
BTW @solardiz I'd appreciate a brief description of the macros. I've tried to digest them but things are unclear to me: For example, why are we storing 20 values?

magnumripper:
BTW for the very first WAIT_SLEEP in each crypt_all call, the optimal sleep time should include key transfer (if `new_keys`) and the duration of the first kernel (be it very quick). Do you reckon that's insignificant?

Obviously it depends on format, and one could add calls to clFinish (plus a set of macros) if needed.

magnumripper:
Hmm, looking at these figures,
```
xfer: 293.856 us, init: 10.720 us, crypt: 32x38.264 ms, final: 22.528 us, aes: 127.680 us, xfer: 221.376 us
gws:     34816    28417 c/s 14898692096 rounds/s     1.225 s per crypt_all()+
```
Perhaps I should ditch complexity and only use the macro for the loop? I'll try that and see what happens.

Edit:
Yes, that seems to be the case. Here's with *only* the loop taken care of, and different test times:

One second
```
Raw:    28191 c/s real, 53154 c/s virtual, Dev#2 util: 100%
```
Three seconds
```
Raw:    26919 c/s real, 147109 c/s virtual, Dev#2 util: 100%
```
Ten seconds
```
Raw:    26730 c/s real, 250926 c/s virtual, Dev#2 util: 100%
```

I now force-pushed this simpler version.

magnumripper:
Another thing I'm not sure of are these alternatives (for eg. oldoffice, no loop):
```diff
diff --git a/src/opencl_oldoffice_fmt_plug.c b/src/opencl_oldoffice_fmt_plug.c
index 4ef9a45c6..b07bed3b4 100644
--- a/src/opencl_oldoffice_fmt_plug.c
+++ b/src/opencl_oldoffice_fmt_plug.c
@@ -542,9 +542,13 @@ static int crypt_all(int *pcount, struct db_salt *salt)
                new_keys = 0;
        }
 
+       WAIT_INIT(global_work_size)
        BENCH_CLERROR(clEnqueueNDRangeKernel(queue[gpu_id], crypt_kernel, 1, NULL, &gws, &lws, 0, NULL, multi_profilingEvent[2]), "Failed running crypt kernel");
 
+       WAIT_SLEEP
        BENCH_CLERROR(clEnqueueReadBuffer(queue[gpu_id], cl_salt, CL_TRUE, 0, sizeof(cs), cur_salt, 0, NULL, multi_profilingEvent[3]), "Failed transferring salt");
+       WAIT_UPDATE
+       WAIT_DONE
 
        if (cur_salt->cracked) {
                BENCH_CLERROR(clEnqueueReadBuffer(queue[gpu_id], cl_result, CL_TRUE, 0, sizeof(unsigned int) * *pcount, cracked, 0, NULL, NULL), "failed reading results back");
```

Versus
```diff
diff --git a/src/opencl_oldoffice_fmt_plug.c b/src/opencl_oldoffice_fmt_plug.c
index 4ef9a45c6..41b43fbce 100644
--- a/src/opencl_oldoffice_fmt_plug.c
+++ b/src/opencl_oldoffice_fmt_plug.c
@@ -528,6 +528,7 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 
        //printf("%s(%d) lws "Zu" gws "Zu" kidx %u k %d mult %u\n", __FUNCTION__, count, lws, gws, key_idx, new_keys, mask_int_cand.num_int_cand);
 
+       WAIT_INIT(global_work_size)
        if (new_keys) {
                /* Self-test kludge */
                if (idx_offset > 4 * (gws + 1))
@@ -544,7 +545,10 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 
        BENCH_CLERROR(clEnqueueNDRangeKernel(queue[gpu_id], crypt_kernel, 1, NULL, &gws, &lws, 0, NULL, multi_profilingEvent[2]), "Failed running crypt kernel");
 
+       WAIT_SLEEP
        BENCH_CLERROR(clEnqueueReadBuffer(queue[gpu_id], cl_salt, CL_TRUE, 0, sizeof(cs), cur_salt, 0, NULL, multi_profilingEvent[3]), "Failed transferring salt");
+       WAIT_UPDATE
+       WAIT_DONE
 
        if (cur_salt->cracked) {
                BENCH_CLERROR(clEnqueueReadBuffer(queue[gpu_id], cl_result, CL_TRUE, 0, sizeof(unsigned int) * *pcount, cracked, 0, NULL, NULL), "failed reading results back");
```
Dry running it I guess one should consider combinations of one salt or many, mask or not, and so on. Keys transfer time could be very significant or completely insignificant.

solardiz:
> Move the macros for avoiding busy-waits to shared header.

OK. I think (separately from a move-only commit) we'll soon need to add a knob to be able to manually disable the busy-wait avoidance. Maybe a `john.conf` setting?

I also got thinking that maybe some of the relative figures I use (like the 2.4%) should be changed to absolute (like 1ms), to better accommodate formats with fast kernels. And I'm not sure my approach of injecting a much lower figure on an over-sleep is the best one - this makes it stay for 20 more runs (currently) and then suddenly disappear. Maybe it's better to have a separate variable that would be slowly adjusted one way and the other, so that it'd stay closer to equilibrium. I almost have code I'd test on real hardware already "written" and "tested" in my mind, but just didn't get around to that yet.

> Use it for 7z-opencl. I intend to add more formats to this PR before merging.

OK. My plan was to add this to bitcoin-opencl next, but you're quicker.

> Because of different code path/kernels used depending on salt, I added a couple of `clFinish()` and also changed a blocking `clEnqueueReadBuffer()` to a non-blocking one with an explicit `clFinish()` after. I think it's all sensible but would appreciate comments.

Hmm, I didn't make a similar change in tezos-opencl. I use the blocking read as indicating kernel duration - the read is invoked while the kernel is running, and completes once the kernel and transfer have completed. So the wait is adjusted for that entire duration, and it's just one wait for both things. Now that I think of it again, actually we'd probably want the wait to cover only the kernel duration, which we could do by adding a `clFlush`, but then the wait macros' adjustments to wait for a slightly shorter period seemed to be just sufficient to cover the read time as well. Still, maybe something to improve in tezos-opencl.

> Looking at virtual figures, they went from more or less 1:1 to eg. 26850 c/s real, 139264 c/s virtual (for a 3-second run) and I presume the latter should mean about 80% decrease of busy-waiting, right?

Right, although I mostly tested with longer cracking runs.

> why are we storing 20 values?

The kernel duration varies even when it "shouldn't". Mostly because the GPU clock rate changes. We want to almost never oversleep, or when we occasionally do we want this to have minimal impact on overall performance. With minimum of 20 values, we almost never oversleep and (given our response to oversleeping) we sort of limit the cost to 5% of performance. I also experimented with storing 2 values (didn't have an array then) or 100 values, with reasonable results. With just 2 values, GPU throttling on overheating (where the clock rate drops to half for a moment) had unnecessarily worse effect on performance. With 100 values, increased CPU usage would stay for longer after even a tiny potential oversleep (due to the approach with injecting a lower value into the array, which we might want to reconsider).

Regarding (not) including key transfer time, I think if we do ever include it, then at least in salted formats (where key transfer doesn't occur every time) it should be a separate WAIT context from the first kernel invocation. Otherwise we'd have unnecessarily long 19 waits after first salt's, and unnecessarily short after we're at the first salt again after having tested 20+ other salts.

Also, I don't know whether the transfers to/from GPU even incur busy waits - do they? I know there's a busy wait when we're trying to read a kernel's results and the kernel is still running, but I don't know if the problem also exists when we're merely transferring already computed data.

solardiz:
> Now that I think of it again, actually we'd probably want the wait to cover only the kernel duration, which we could do by adding a `clFlush`

Maybe like this (mostly untested):

```diff
+++ b/src/opencl_tezos_fmt_plug.c
@@ -337,11 +337,13 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 
        WAIT_INIT(gws)
        WAIT_SLEEP
+       BENCH_CLERROR(clFinish(queue[gpu_id]), "clFinish");
+       WAIT_UPDATE
+       WAIT_DONE
+
        BENCH_CLERROR(clEnqueueReadBuffer(queue[gpu_id], mem_final, CL_TRUE, 0,
                                sizeof(*cracked), cracked,
                                0, NULL, multi_profilingEvent[4]), "Copy result back");
-       WAIT_UPDATE
-       WAIT_DONE
 
        if (*cracked) {
                BENCH_CLERROR(clEnqueueReadBuffer(queue[gpu_id], mem_final, CL_TRUE, 0,
```

What do you think, @magnumripper?

magnumripper:
> Maybe like this (mostly untested):

Not sure why that would be any better (or worse). The thing I did was because 7z has conditional kernels and transfers, so I needed separate sets of WAIT macros. And like I wrote later, I now dropped those changes altogether because those calls are not significant.

magnumripper:
> My plan was to add this to bitcoin-opencl next, but you're quicker.

My plan was mostly to get to know these macros, I'm not sure we'll merge this any time soon.

> we'll soon need to add a knob to be able to manually disable the busy-wait avoidance. Maybe a john.conf setting?

Right, perhaps that should be my next move.

solardiz:
> Not sure why that would be any better (or worse).

My thinking is that with the wait adjusted for kernel+read duration, we'd potentially oversleep. I think we only avoid oversleeping by the -2.4% adjustment. I think this patch corrects that. I'd appreciate your comments.

> I don't know if the problem also exists when we're merely transferring already computed data.

Also, I think our data transfers are generally below 1ms each, and that's the threshold I currently use for choosing to sleep at all. So a set of WAITs just for a data transfer is probably a no-op.

magnumripper:
> Also, I don't know whether the transfers to/from GPU even incur busy waits - do they? I know there's a busy wait when we're trying to read a kernel's results and the kernel is still running, but I don't know if the problem also exists when we're merely transferring already computed data.

Not sure I understand what you mean. After transferring to GPU, there's no busy wait until you also queued a kernel and either call clFinish or try to (blocking) read its results. However, theoretically the transfer times could be very significant for the total busy-wait.

magnumripper:
> > Not sure why that would be any better (or worse).
> 
> My thinking is that with the wait adjusted for kernel+read duration, we'd potentially oversleep. I think we only avoid oversleeping by the -2.4% adjustment. I think this patch corrects that. I'd appreciate your comments.

In my head, that change is merely a way to add margin and you could just as well adjust the 2.4 % figure. I see no difference so I think the latter is a cleaner choice.

> > I don't know if the problem also exists when we're merely transferring already computed data.
> 
> Also, I think our data transfers are generally below 1ms each, and that's the threshold I currently use for choosing to sleep at all. So a set of WAITs just for a data transfer is probably a no-op.

Oh right, so 1 ms is a figure I should remember when looking at autotune figures before experimenting with a format.

solardiz:
> After transferring to GPU, there's no busy wait until you also queued a kernel and either call clFinish or try to (blocking) read its results.

Right. So for unsalted formats, it does make sense to include the transfer time, but we happen to do that by simply measuring the duration of blocking on `clFlush` or read-back.

For salted formats, we seem to have the problem of inadvertently including the transfer time from first salt also in the next 19 salts, and similarly not including the transfer time in the first salt after we went through 20+ salts. (My testing was with fewer salts than that. Maybe I'd see a performance regression at 20+, although it'd be only for one salt, so would be hard to measure.) 

I think we need to flush the transferred keys before proceeding to the WAIT logic for the kernel invocation. Then we'd be tuning to the same duration (just the kernel's) regardless of whether it's the first salt or not.

solardiz:
> In my head, that change is merely a way to add margin and you could just as well adjust the 2.4 % figure. I see no difference so I think the latter is a cleaner choice.

I think the change is better than increasing the margin, because it makes the macros measure exactly what's needed, not measure a known-excessive duration and then apply an arbitrary higher margin to correct for that.

magnumripper:
> I think we need to flush the transferred keys before proceeding to the WAIT logic for the kernel invocation. Then we'd be tuning to the same duration (just the kernel's) regardless of whether it's the first salt or not.

Exactly, I guess we have a greater risk of oversleeping on second salt if we don't do this.

magnumripper:
> > In my head, that change is merely a way to add margin and you could just as well adjust the 2.4 % figure. I see no difference so I think the latter is a cleaner choice.
> 
> I think the change is better than increasing the margin, because it makes the macros measure exactly what's needed, not measure a known-excessive duration and then apply an arbitrary higher margin to correct for that.

I think you're wrong (but I could be wrong!). Given this:

1. key xfer, 700 us
2. kernel, 10 ms
3. result xfer, 600 us

The total queue duration will be 11.3 ms. Given the result transfer is blocking or we do a clFinish after it, we'd want to sleep just under 11.3 ms. In what way is anything excessive?

solardiz:
> In what way is anything excessive?

In your example, we want to _start_ the result xfer no later than 10.7 ms. So 11.3 ms is excessive.

Anyway, I realized/recalled why I kept things that way in tezos-opencl: its results transfer is just of one `uint`, so is on par with a `clFlush`. So my patch above isn't needed. However, similar changes are likely needed in other formats.

magnumripper:
> > In what way is anything excessive?
> 
> In your example, we want to _start_ the result xfer no later than 10.7 ms. So 11.3 ms is excessive.

Why would we want that?! It's a queue! We queue up three things, no busy-waits involved anywhere. Only when we do a blocking read or a clFinish (or wait for events etc), busy-waits are involved. At least that's how I always assumed it.

solardiz:
Oh, if we queue the 3 things up as non-blocking before the very first wait, then you're right! So the final read should be non-blocking, and then we do a `clFlush` surrounded by `WAIT*`.

magnumripper:
Exactly! So for an unsalted non-split kernel format, we'd probably want WAIT_INIT before the very first transfer and WAIT_UPDATE+DONE after the very last read (which is normally the only blocking one).

Edit: The WAIT_INIT can obviously come much later, I meant it's a single one context. It could be just this:
```c
	WAIT_INIT(gws)
	WAIT_SLEEP
	BENCH_CLERROR(clEnqueueReadBuffer(queue[gpu_id], mem_result, CL_TRUE, 0 (...));
	WAIT_UPDATE
	WAIT_DONE
```

magnumripper:
So to sum things up, I guess tezos only need this:
```diff
diff --git a/src/opencl_tezos_fmt_plug.c b/src/opencl_tezos_fmt_plug.c
index e78db0b7b..6c556028f 100644
--- a/src/opencl_tezos_fmt_plug.c
+++ b/src/opencl_tezos_fmt_plug.c
@@ -281,6 +281,7 @@ static int crypt_all(int *pcount, struct db_salt *salt)
                                NULL, &gws, lws, 0, NULL,
                                multi_profilingEvent[1]), "Run kernel");
 
+       BENCH_CLERROR(clFinish(queue[gpu_id]), "clFinish");
        WAIT_INIT(gws)
        for (i = 0; i < (ocl_autotune_running ? 1 : loops); i++) {
                BENCH_CLERROR(clEnqueueNDRangeKernel(queue[gpu_id],
```
Perhaps with a comment explaining why it's there.

magnumripper:
The current commit for 7z-opencl assumes the last call to the loop kernel isn't having a significantly lower number of iterations. I need to verify that. Perhaps we could add some "debug mode" to the WAIT_ macros where they'd somehow print or collect statistics on oversleeping? I have no idea how it would be done, just a thought.

magnumripper:
> The current commit for 7z-opencl assumes the last call to the loop kernel isn't having a significantly lower number of iterations. I need to verify that.

That very format seems to always have uniform number of iterations though.

solardiz:
@magnumripper Your tweak to tezos-opencl looks correct to me. Your suggested (in #4944) skipping of `WAIT_*` on last loop iteration is precisely what I intended to do for split kernels in formats where the last loop iteration may have a lower number of in-kernel iterations. In tezos-opencl, it's 513+512+512+511, which is uniform enough not to do this.

magnumripper:
(Edit: figures improved after adding macros to cmp_all)

Office format shows very good figures: Virtual speeds went up to 10-14x the real ones (in 3 sec `--test` runs) or 20x flat (in 10 sec runs). Using the same metric, tezos "only" see 10x for a 10 sec run. I saw some flaws with autotuning vs. `-cost` option that I fixed in the same commit (could make it a separate one but still in same PR if you really want to). The change to `bench.c` isn't dramatic at all if you look closer.

A curious thought: How much power/money/climate harm would this save on a typical GPU-rig with say 8 GPUs and 8 CPUs, running 24/7/365. Perhaps @asaidac had a point there in #4781...

Also, I'm wondering if this could boost post-processing because we might get more of that on "turbo clock", at least theoretically, yes?

magnumripper:
> Virtual speeds went up to 10-14x the real ones (in 3 sec `--test` runs) or 20x flat (in 10 sec runs)

Not sure what to use as a go-to benchmark for this. In a 30 second run, virtual is 28x. In a 60s run, it's 34x. For quick benchmarking, I think I'll just go with 10 seconds but the actual improvement is generally better than the figures received from it. 

Edit: A 300 second test run showed 35x, so I guess a 60s run is fairly close to reality.

magnumripper:
> I saw some flaws with autotuning vs. `-cost` option that I fixed in the same commit (could make it a separate one but still in same PR if you really want to). The change to `bench.c` isn't dramatic at all if you look closer.

Yet I see weirdness in md5crypt-opencl after that change to `bench.c`: Autotune runs twice. I can't see how that could happen but I obviously need to address it.

magnumripper:
OK it's because we now got an extra reset between "Many salts" and "Only one salt". I think I know how to handle it.

Edit: Fixed now. Made it a separate commit this time, as opposed to included in the office-opencl commit.

magnumripper:
md5crypt on 2080ti
```diff
diff --git a/src/opencl_md5crypt_fmt_plug.c b/src/opencl_md5crypt_fmt_plug.c
index a58a32147..b86891b5e 100644
--- a/src/opencl_md5crypt_fmt_plug.c
+++ b/src/opencl_md5crypt_fmt_plug.c
@@ -382,13 +382,16 @@ static int crypt_all(int *pcount, struct db_salt *salt)
                        0, insize, inbuffer, 0, NULL, multi_profilingEvent[0]),
                        "Copy memin");
 
+       // Better precision for WAIT_ macros
+       BENCH_CLERROR(clFinish(queue[gpu_id]), "clFinish error");
+
        ///Run kernel
        BENCH_CLERROR(clEnqueueNDRangeKernel(queue[gpu_id], crypt_kernel, 1,
                NULL, &global_work_size, lws, 0, NULL, multi_profilingEvent[1]),
@@ -396,8 +399,12 @@ static int crypt_all(int *pcount, struct db_salt *salt)
                0, outsize, outbuffer, 0, NULL, multi_profilingEvent[2]),
                "Copy data back");
 
        ///Await completion of all the above
+       WAIT_INIT(global_work_size)
+       WAIT_SLEEP
        BENCH_CLERROR(clFinish(queue[gpu_id]), "clFinish error");
+       WAIT_UPDATE
+       WAIT_DONE
 
        new_keys = 0;
        return count;
```
```
pass xfer: 111.744 us, crypt: 1.327 ms, result xfer: 88.448 us
gws:     34816   22792K c/s 22792977000 rounds/s    1.527 ms per crypt_all()!!
```
```
Many salts:     9570K c/s real, 174016K c/s virtual, Dev#2 util: 37%
Only one salt:  8397K c/s real, 30207K c/s virtual, Dev#2 util: 38%
```
The "one salt" improvement for a 10 second test is poor - virtual is only 3.6x better than real time.

Perhaps I should have the added clFinish conditional - only do it if running with more than one salt? Then we'd include the keys transfer time in the sleeps with no drawbacks.

magnumripper:
> Perhaps I should have the added clFinish conditional - only do it if running with more than one salt? Then we'd include the keys transfer time in the sleeps with no drawbacks.

I tried that (or rather simulated it) and it didn't help at all. Apparently the precision of the WAIT_* macros is too low for formats like this one.

magnumripper:
It's not just the short kernel duration (barely over a ms) - sometimes these are picked:
```
pass xfer: 14.185 ms, crypt: 180.641 ms, result xfer: 11.149 ms
gws:   4456448   21635K c/s 21635766000 rounds/s  205.975 ms per crypt_all()+
```
Yet, the virtual figures are just as (relatively) bad:
```
Many salts:     23035K c/s real, 55044K c/s virtual, Dev#2 util: 96%
Only one salt:  17157K c/s real, 56797K c/s virtual, Dev#2 util: 78%
```
Now obviously at these work sizes the password transfer time is also a candidate. I'll try that... Nope, didn't help.

On a sidenote we have an autotune problem for this format: It often picks much lower GWS ending up in half of these speeds.

magnumripper:
Just brainstorming freely now:

What if we'd have the array as `wait_times[2][20]` and the sleep as `WAIT_SLEEP(context)` where `context` would be treated as 0 or non-zero and give the first array dimension, such as using `wait_times[!!context][...]`?

For non-loop kernels, we wouldn't have to separate the new_keys case from the rest. We'd use `WAIT_SLEEP(new_keys)` (and obviously reset `new_keys` only after any such thing) and all of md5crypt would fit into a single set of macros, improving coverage of the sleeps.

For loop kernels we could similarly have a different context for the first iteration, using `WAIT_SLEEP(i)`.

Or perhaps somehow combine these two ideas.

magnumripper:
Note to self: This is great for testing with eg. `--test=10`
```diff
diff --git a/src/bench.c b/src/bench.c
index 69fb06020..ec364069b 100644
--- a/src/bench.c
+++ b/src/bench.c
@@ -1038,8 +1038,8 @@ AGAIN:
                if (john_main_process && benchmark_time) {
 #if !defined(__DJGPP__) && !defined(__BEOS__) && !defined(__MINGW32__) && !defined (_MSC_VER)
                        if (results_m.virtual)
-                               printf("%s:\t%s c/s real, %s c/s virtual%s\n",
-                                      msg_m, s_real, s_virtual, s_gpu);
+                               printf("%s:\t%s c/s real, %s c/s virtual (%.1fx)%s\n",
+                                      msg_m, s_real, s_virtual, (double)results_m.real / results_m.virtual, s_gpu);
                        else
 #endif
                                printf("%s:\t%s c/s%s\n",
```

magnumripper:
@solardiz I added spurious induction of oversleep (double the correct value) to test the code, and apparently it's handled very quickly (gone in a couple of iterations). That's great, but looking at the macros I can't understand how/where you handle it? Can you please describe that?

Below, I'm detecting oversleep by introducing another call to john_get_nano *after* the sleep. If less than 500 us past since that one (the `blocked` duration), I print it as an oversleep:
```
(...)
slept 33.604 ms, blocked 1.308 ms
slept 33.953 ms, blocked 1.158 ms
slept 33.604 ms, blocked 780.616 us
slept 34.325 ms, blocked 1.498 ms
slept 33.895 ms, blocked 1.067 ms
slept 33.416 ms, blocked 588.811 us
slept 33.820 ms, blocked 1.085 ms
Oversleep detected: slept 65.445 ms, blocked 22.938 us
Oversleep detected: slept 56.076 ms, blocked 22.968 us
Oversleep detected: slept 40.914 ms, blocked 25.180 us
Oversleep detected: slept 35.116 ms, blocked 23.193 us
slept 33.359 ms, blocked 3.220 ms
DONE
Raw:    470941 c/s real, 1572K c/s virtual (3.3x), Dev#2 util: 95%
```
```diff
diff --git a/src/opencl_common.h b/src/opencl_common.h
index de2dbb88e..d341169ad 100644
--- a/src/opencl_common.h
+++ b/src/opencl_common.h
@@ -331,23 +331,36 @@ void opencl_process_event(void);
 #define GET_EXACT_MULTIPLE(dividend, divisor)	  \
 	(divisor) ? ((dividend > divisor) ? ((dividend / divisor) * divisor) : divisor) : dividend
 
+#define n2s(ns)	human_prefix_small((ns) / 1E9)
+
 #define WAIT_INIT(work_size) { \
 	static uint64_t wait_last_work_size; \
 	int wait_sleep = (work_size >= wait_last_work_size); \
 	wait_last_work_size = work_size; \
 	static uint64_t wait_times[20], wait_min; \
 	static unsigned int wait_index; \
-	uint64_t wait_start = 0;
+	uint64_t wait_start = 0, wait_mid = 0;
 #define WAIT_SLEEP \
 	if (gpu_nvidia(device_info[gpu_id])) { \
-		wait_start = john_get_nano(); \
+		wait_start = wait_mid = john_get_nano(); \
 		uint64_t us = wait_min >> 10; /* 2.4% less than min */ \
-		if (wait_sleep && us >= 1000) \
+		if (wait_sleep && us >= 1000) { \
 			usleep(us); \
+			wait_mid = john_get_nano(); \
+		} \
 	}
 #define WAIT_UPDATE \
 	if (gpu_nvidia(device_info[gpu_id])) { \
-		uint64_t wait_new = john_get_nano() - wait_start; \
+		static int dice; /* Yeah very random lol */ \
+		uint64_t wait_end = john_get_nano(); \
+		uint64_t wait_new = wait_end - wait_start; \
+		if (wait_mid > wait_start) { \
+			uint64_t block_duration = wait_end - wait_mid; \
+			if (block_duration < 500000) \
+				fprintf(stderr, "Oversleep detected: slept %ss, blocked %ss\n", n2s(wait_new), n2s(block_duration)); \
+			else if (1) \
+				fprintf(stderr, "slept %ss, blocked %ss\n", n2s(wait_new), n2s(block_duration)); \
+		} \
 		if (wait_new < wait_min && wait_new < wait_min * 1000 / 1012) /* 1.2% less than min */ \
 			wait_new = wait_new * 7 / 8; /* we might have overslept and don't know by how much */ \
 		if (wait_times[wait_index] == wait_min) { /* about to replace former minimum */ \
@@ -362,6 +375,7 @@ void opencl_process_event(void);
 			wait_index = 0; \
 		if (wait_new < wait_min) \
 			wait_min = wait_new; \
+		if (1 && (++dice & 0x1f) == 0x1f) wait_min <<= 1; /* induce spurious oversleep */ \
 		wait_sleep = 1; \
 	}
 #define WAIT_DONE }
```
(Note this was run with only the kernel calls macro'd in tezos. 500 us is too short a threshold for the transfers, it should be slightly more complicated, depending on wait_min)

magnumripper:
> Oh, if we queue the 3 things up as non-blocking before the very first wait, then you're right! So the final read should be non-blocking, and then we do a `clFlush` surrounded by `WAIT*`.

Yes but `clFlush` is non-blocking so we need to do a `clFinish` surrounded by the macros.

Only now I realized we can't do this because the xfer isn't triggered until after the sleep:
```
	WAIT_INIT(gws)
	WAIT_SLEEP
	BENCH_CLERROR(clEnqueueReadBuffer(queue[gpu_id], mem_final, CL_TRUE, 0, (...));
	WAIT_UPDATE
	WAIT_DONE
```

We have to do this:
```
	BENCH_CLERROR(clEnqueueReadBuffer(queue[gpu_id], mem_final, CL_FALSE, (...));
	WAIT_INIT(gws)
	WAIT_SLEEP
	BENCH_CLERROR(clFinish(queue[gpu_id]), "clFinish");
	WAIT_UPDATE
	WAIT_DONE
```

And *probably* this as well (at least unless there's more is in the queue before it):
```
	BENCH_CLERROR(clEnqueueReadBuffer(queue[gpu_id], mem_final, CL_FALSE, (...));
	BENCH_CLERROR(clFlush(queue[gpu_id]), "failed in clFlush");
	WAIT_INIT(gws)
	WAIT_SLEEP
	BENCH_CLERROR(clFinish(queue[gpu_id]), "clFinish");
	WAIT_UPDATE
	WAIT_DONE
```

magnumripper:
> md5crypt on 2080ti

The poor results for md5crypt were completely fixed with the correct clFlush / clFinish stuff

solardiz:
> Also, I'm wondering if this could boost post-processing because we might get more of that on "turbo clock"

Yes, and also in many other scenarios, such as on a rig with more GPUs than CPU cores, and also without post-processing, but simply the candidate password generation on CPU. There's also effect on GPU clock rate via heat from the CPUs.

> spurious induction of oversleep (double the correct value) to test the code, and apparently it's handled very quickly (gone in a couple of iterations). That's great, but looking at the macros I can't understand how/where you handle it?

`wait_new = wait_new * 7 / 8;` injects a duration that is 7/8 of actual. With that in the array, the minimum and next sleep interval is correspondingly lower. It only takes a few of those to get to 1/2 lower, as in your testing. However, as I mentioned I have thoughts on reworking that approach to replace injections of adjusted values into the array with adjustments of the thresholds - replace the 2.4%, etc. with variables that we'd adjust on oversleep and then slowly adjust back on non-oversleep. My idea is to use an integer recording the number of 1/1024th. With the current code, it's implied 1000/1024, but that "1000" could be a variable we'd adjust.

> Only now I realized we can't do this because the xfer isn't triggered until after the sleep:

Hmm, but I am doing just that for the final kernel invocation in tezos-opencl. Is that part not working as intended? Have you tried repairing it there and see improvement? Thanks!

solardiz:
I think the final kernel in tezos-opencl may often be below 1 ms, depending on GWS.

solardiz:
> replace injections of adjusted values into the array with adjustments of the thresholds - replace the 2.4%, etc. with variables that we'd adjust on oversleep and then slowly adjust back on non-oversleep. My idea is to use an integer recording the number of 1/1024th. With the current code, it's implied 1000/1024, but that "1000" could be a variable we'd adjust.

I now wonder at what time it'd be best to make that change. It looks like you @magnumripper spend considerable time on testing the various formats with these macro additions to them. Maybe such testing would best proceed with the macros "finalized" first? Or maybe having more formats with the macros already in and already tested would help test the changes to the macros, to ensure no regressions.

magnumripper:
> Hmm, but I am doing just that for the final kernel invocation in tezos-opencl. Is that part not working as intended? Have you tried repairing it there and see improvement? Thanks!

Yes, it's fixed in this PR now

> I think the final kernel in tezos-opencl may often be below 1 ms, depending on GWS.

Yes, so I saw little change. I've seen several things that make me wonder if we should perhaps lower the 1 ms to 100 us.

> > replace injections of adjusted values into the array with adjustments of the thresholds - replace the 2.4%, etc. with variables that we'd adjust on oversleep and then slowly adjust back on non-oversleep. My idea is to use an integer recording the number of 1/1024th. With the current code, it's implied 1000/1024, but that "1000" could be a variable we'd adjust.
> 
> I now wonder at what time it'd be best to make that change. It looks like you @magnumripper spend considerable time on testing the various formats with these macro additions to them. Maybe such testing would best proceed with the macros "finalized" first? Or maybe having more formats with the macros already in and already tested would help test the changes to the macros, to ensure no regressions.

I'll likely take a break from this right now. I "had to" do a number of formats to get aquainted with the macros and get us discussing things but I'm fully prepared for the case I need to revisit/re-test every single file I have changed so far 🤣 

I think I've tried most different types of formats: unsalted, non-looped, internal mask or not. I have yet to see any really big caveat.

solardiz:
@magnumripper I have little time for this now and your changes look good enough, so I think we should merge this PR first. Maybe we need to wait to hear from @claudioandre-br, though, because this touches sha512crypt.

solardiz:
@magnumripper As I recall, you wrote in a comment here (that I can't find now) that md5crypt-opencl would autotune badly half of the time on your hardware. Is that issue gone? If not, you'd want to create a separate issue for it. Thanks!

ghost:
You can merge it right away. We'll test it as time goes on.

The change looks safe.

solardiz (approved):

PR #5008, opened 2022-01-20
New default wordlist

### LINKED ISSUE #5002: New default wordlist
I think we need to extend or replace the default `password.lst`. I've prepared a generated common password list based on HIBP v8 and RockYou - can describe the criteria and algorithm and simulated cracking with it vs. alternatives. The criteria are such that a password used by just one person many times is very unlikely to be included (I use the numbers of occurrences in HIBP and RockYou such that the password has to come from more than one set). The list is currently 1.8M lines, 15 MB. Sounds acceptable, given we have `.chr` files approaching this size (and larger than that combined)?

solardiz:
Fixes #5002. I tried to record some information on what came from where in our git history without wasting too much `.git` space.


magnumripper (commented):
I will check this out now and have a look at any encoding issues.

magnumripper (approved):
Looks good to me. Only 21 words contain non-ASCII and all of them are UTF-8

Issue #5004, opened 2022-01-18
Optimize default rule sets


solardiz:
As part of my work towards #5003, I came up with this command:

```
$ cut -c12- sim-single.log | uniq -c | fgrep -B1 '+ Cracked ' | perl -ne 'chomp; $r = $_ if (/- Rule #/); print "$_ $r\n" if (/\+ Cracked /);' | sort -rn | head -20
 188525 + Cracked ?       1 - Rule #1: ':' accepted as ''
   7670 + Cracked ?       1 - Rule #119: '<* l $1' accepted as '<*l$1'
   5742 + Cracked ?       1 - Rule #3: '-c (?a c Q' accepted as '(?acQ'
   3035 + Cracked ?       1 - Rule #59: 'r c' accepted as 'rc'
   2977 + Cracked ?       1 - Rule #129: '<* l $0' accepted as '<*l$0'
   2879 + Cracked ?       1 - Rule #120: '<* l $2' accepted as '<*l$2'
   2843 + Cracked ?       1 - Rule #121: '<* l $3' accepted as '<*l$3'
   2771 + Cracked ?       1 - Rule #56: '(?a )?d /?d a0 'p Xpz0' accepted as '(?a)?d/?da0'pXpz0'
   2763 + Cracked ?       1 - Rule #123: '<* l $5' accepted as '<*l$5'
   2680 + Cracked ?       1 - Rule #125: '<* l $7' accepted as '<*l$7'
   2671 + Cracked ?       1 - Rule #122: '<* l $4' accepted as '<*l$4'
   2406 + Cracked ?       1 - Rule #127: '<* l $9' accepted as '<*l$9'
   2386 + Cracked ?       1 - Rule #124: '<* l $6' accepted as '<*l$6'
   2371 + Cracked ?       1 - Rule #126: '<* l $8' accepted as '<*l$8'
   2174 + Cracked ?       1 - Rule #62: '-c u Q' accepted as 'uQ'
   1638 + Cracked ?       1 - Rule #7: '-<7 >7 '7 l' accepted as '>7'7l'
   1589 + Cracked ?       1 - Rule #130: '<* l $a' accepted as '<*l$a'
   1559 + Cracked ?       1 - Rule #6: '-<6 >6 '6' accepted as '>6'6'
   1368 + Cracked ?       1 - Rule #50: '/?d @?d >3 Az"123" <+ Q' accepted as '/?d@?d>3Az"123"<+Q'
[...]

Issue #5003, opened 2022-01-18
New default wordlist rule set


solardiz:
The existing wordlist rule set coming from core is such that it's easy to describe what it does in plain English and thus to use it as part of a password policy. However, it is not as effective at password cracking as other rule sets that we have are. I currently think we should simply replace the default wordlist rule set with our default `Single` rule set (just `.include` that section), and rename the current wordlist rule set e.g. to `Wordlist-old` or maybe `Policy` - suggestions welcome.

In terms of cracking more passwords early on, another good potential replacement would be the hashcat-derived `best64` rule set, but that one produces too many dupes. So it's not something we should do until we have #4728. In my testing it produces more cracks than `Single` does in the same number of candidates generated even _including_ dupes, but then `Single` proceeds to crack more passwords since it's a larger rule set. Doing `best64` followed by `Single` results in very few additional cracks when `Single` just starts (and produces its first `best64` worth of candidates by ~13% of its run time), and the total number of cracks is only slightly higher than `Single` alone (~13% more candidates, ~1% more cracks).

solardiz:
> `best64` rule set, but that one produces too many dupes.

Actually, when I compare against just the initial portion of a `Single` run, the difference in dupes is not so profound (as it was when comparing to the whole `Single`):

`-ru=best64`
```
Total lines read: 133529430, unique lines written: 74325781 (55%), no slow passes
```

`-ru=single -max-candidates=133529430`
```
Total lines read: 133529430, unique lines written: 78785273 (59%), no slow passes
```

So the bigger drawback is that after having run `best64`, running a larger rule set like `Single` produces a lot of cross-dupes early on (and thus few new cracks early on). With this, it looks like we need to make a decision: should our default optimize for maximum number of cracks early on, or should it allow to efficiently proceed with a longer run if needed.

solardiz:
Maybe a task here is to review which `best64` rules crack the most passwords that a corresponding initial portion of `Single` does not, and to incorporate into `Single` just those few(?) rules. That would avoid the dilemma.

PR #5010, opened 2022-01-22
Wordlist mode: Optionally log per-rule statistics

solardiz:
This adds the `PerRuleStats` setting, disabled by default, for logging of per-rule statistics and scores usable to generate an order-optimized rule set. The setting also aligns rules to whole buffers, by injecting just enough instances of a fake password to infer when the buffer has filled and been processed.

This is related to the ongoing work on #5003 and #5004.


solardiz:
The injected wrong password may be seen in the status line, which isn't pretty:

```
12779873g 0:00:03:48 DONE (2022-01-22 22:30) 55924g/s 6247Kp/s 6247Kc/s 278001GC/s 0171251900..injected wrong password
```

However, I think I'll leave that as-is for now.

solardiz:
Force-pushed with two changes:
1. Log statistics even for rejected rules (all-zero, but at least those rules are not lost and can be e.g. preserved at the end of the re-ordered list, just like it happens for rules that were accepted but generated no candidates).
2. (Re)set previous counts on entry to `do_wordlist_crack` (this is needed e.g. when in batch mode, where wordlist cracking starts after we've already done some cracking).

Issue #4971, opened 2021-12-23
Many ASan reports from fuzz_option.pl


solardiz:
We have that `fuzz_option.pl` script that we haven't run for years. Well, it does produce results when run as simply:

```
./fuzz_option.pl ./john cpu 2>&1 | tee fuzz.log
```

They're not well separated by format, though. Also the RAM usage became excessive - was over 100 GB when I interrupted. So I patched the script as below:

```diff
+++ b/run/fuzz_option.pl
@@ -22,7 +22,7 @@ use Errno;
 use File::Copy;
 
 # Processes per logical CPU
-$factor = 4;
+$factor = 2;
 
 
 $ENV{'OMP_NUM_THREADS'} = '1';
@@ -76,7 +76,7 @@ if ($#formats > $cpus) {
 for ($i = $from; $i <= $to; $i++) {
 
        #print "Fuzzing format:$formats[$i] \n";
-       open(JOHN, "| $john_path --fuzz=fuzz.dic --format=$formats[$i] ") || die;
+       open(JOHN, "| $john_path --fuzz=fuzz.dic --format=$formats[$i] > fuzz-$formats[$i]-out 2> fuzz-$formats[$i]-err ") || die;
        close(JOHN);
 
        die if ($? == 2 || $? == 15); # INT or TERM
@@ -90,7 +90,7 @@ for ($i = $from; $i <= $to; $i++) {
                print LOG "Fuzz formats=$formats[$i] WRONG.\n";
                close(LOG);
 
-               Reproduce($formats[$i]);
+#              Reproduce($formats[$i]);
        }
 }
 
```

and re-ran as:

```
(ulimit -t 110; time ./fuzz_option.pl ./john cpu)
```

A total of 417 formats were fuzzed. This completed in 11 minutes with 3 processes still running for a bit longer. 50 or so got killed on bumping into the 110 seconds CPU time limit. Dynamic formats were not included (are reported as one item `dynamic_n`, which of course is then unrecognized).

Processing of the results:

```
$ egrep -l 'AddressSanitizer|ASAN' fuzz-* 
fuzz-7z-err
fuzz-Fortigate-err
fuzz-Fortigate256-err
fuzz-LastPass-err
fuzz-OpenBSD-SoftRAID-err
fuzz-PHPS-err
fuzz-Padlock-err
fuzz-ZIP-err
fuzz-leet-err
fuzz-mscash-err
fuzz-mscash2-err
fuzz-mschapv2-naive-err
[...]

Issue #5023, opened 2022-01-26
--fuzz leaks memory in formats with `psalt->dsalt.salt_alloc_needs_free = 1`


AlekseyCherepanov:
ASan reports leaks after `--fuzz`ing format with dynamic salt allocation (#752).

I patched out most self tests in `pkzip_fmt_plug.c`, so only the last one is present. It is short and fuzzing finishes in 4 seconds. (Local paths are replaced by `...`.)
```
$ /usr/bin/time ./run/john --fuzz --format=pkzip
Fuzzing: PKZIP [32/64]...    Completed

All 1 formats passed fuzzing test!

=================================================================
==27080==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 386988 byte(s) in 271 object(s) allocated from:
    #0 0x7f8e3ac04518 in __interceptor_calloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cc:95
    #1 0x56312e113172 in mem_calloc .../john/src/memory.c:107
    #2 0x56312e010988 in get_salt .../john/src/pkzip_fmt_plug.c:675
    #3 0x56312e0f9520 in ldr_load_pw_line .../john/src/loader.c:1045
    #4 0x56312e1991f6 in fuzz_test .../john/src/fuzz.c:586
    #5 0x56312e199760 in fuzz .../john/src/fuzz.c:689
    #6 0x56312e0ee3ab in john_run .../john/src/john.c:1672
    #7 0x56312e0ee3ab in main .../john/src/john.c:2082
    #8 0x7f8e3a16009a in __libc_start_main ../csu/libc-start.c:308

SUMMARY: AddressSanitizer: 386988 byte(s) leaked in 271 allocation(s).
Command exited with non-zero status 1
3.51user 0.87system 0:04.44elapsed 98%CPU (0avgtext+0avgdata 612640maxresident)k
8inputs+1440outputs (0major+187626minor)pagefaults 0swaps
```

Same thing happens with 7z. It can finish in ~50 seconds without commenting out self-tests. But 7z requires fix for #4971. (I am about to send PR with it.)

Both formats do not leak in regular runs (both `--test` and real attacks). So I guess that's a problem with fuzzer.

Both formats have the following setting for salts:
```c
psalt->dsalt.salt_alloc_needs_free = 1;
```

I did not test formats with different variants of the setting.

solardiz:
[...]

Issue #4956, opened 2021-12-21
7z formats: Support stacked methods/filters


magnumripper:
See https://github.com/philsmd/7z2hashcat/issues/35

7z2hashcat now support this, so I need to merge that into 7z2john and implement support in the formats.

magnumripper:
There seems to be problems with upstream, need to sort that out first

magnumripper:
Everything's sorted so this is where I'm at right now. Alas, this is hard as h€ll to get right, so I'm now swamped until I get working code at all, and then swamped again until I straighten that code out as anywhere near sane.

PR #5025, opened 2022-01-26
7z formats: fix valid to sync with sevenzip_get_salt

AlekseyCherepanov:
Part of fixes for #4971.

### DIFF
--- a/src/7z_common_plug.c
+++ b/src/7z_common_plug.c
@@ -215,7 +215,7 @@ int sevenzip_valid(char *ciphertext, struct fmt_main *self)
 		goto err;
 	if (!ishexlc(p))
 		goto err;
-	if (c_type && c_type != 128) {
+	if (type && type != 128) {
 		if ((p = strtokm(NULL, "$")) == NULL) /* CRC len */
 			goto err;
 		if (!isdec(p))


AlekseyCherepanov:
There was crash in `sevenzip_get_salt` due to `p = strtokm(NULL, "$")` returning `NULL`. Such hash should be rejected by `valid`.

There was something fishy in `sevenzip_valid`.

Values:
```c
	type = atoi(p);
	c_type = type & 0xf;
	p_type = (type >> 4) & 0xf;
```

One use (looks ok):
```c
	if (c_type > 2
[...]
			    && type != 128) {
```

Other use:
```c
	if (c_type && c_type != 128) {
		if ((p = strtokm(NULL, "$")) == NULL) /* CRC len */
```

So `c_type` is not correct because 128 is not possible as `type & 0xf`. Fixing the second `c_type` here did not help. Fixing both helped.

Respective code in `get_salt` (crash is on the next line):
```c
	if (cs.type && cs.type != 128) {
		p = strtokm(NULL, "$"); /* CRC length */
```

So it looks consistent now. I cannot reason about true meaning though.

AlekseyCherepanov:
`--test` for `7z` and `7z-opencl` are ok.

`--fuzz` completes well. But ASan reports leak as in #5023.

magnumripper (approved):
From a relatively quick look, this looks correct - the old code would never hit that.

There are pending changes to this format soon (#4956) but we can do this first for sure.

Issue #5022, opened 2022-01-25
crash in opencl formats with gpu-side mask when in ASan build


AlekseyCherepanov:
john is fresh:
```
$ ../run/john
John the Ripper 1.9.0-jumbo-1+bleeding-b81ef254b 2022-01-24 23:51:52 +0100 ASan [linux-gnu 64-bit x86_64 AVX2 AC]
```

It was built with asan:
```
$ ./configure --enable-asan --disable-openmp
```

The crash:
```
$ ../run/john --test --format=nt-opencl
NOTE: This is a debug build, speed will be lower than normal
Device 1: Hawaii [AMD Radeon (TM) R9 390 Series]
Benchmarking: NT-opencl [MD4 OpenCL/mask accel]... Build log: /tmp/OCL137294T1.cl:332:18: warning: unknown attribute 'max_constant_size' ignored
                __attribute__((max_constant_size (NUM_INT_KEYS * 4)))
                               ^
1 warning generated.

=================================================================
==137294==ERROR: AddressSanitizer: global-buffer-overflow on address 0x55c837144355 at pc 0x55c8353d579f bp 0x7ffe09d61fb0 sp 0x7ffe09d61fa8
READ of size 1 at 0x55c837144355 thread T0
    #0 0x55c8353d579e in finalize_mask /home/user/john/src/mask.c:2459
    #1 0x55c8353d6eca in mask_init /home/user/john/src/mask.c:2312
    #2 0x55c835382048 in benchmark_all /home/user/john/src/bench.c:894
    #3 0x55c8353b45f0 in john_run /home/user/john/src/john.c:1661
    #4 0x55c8353b45f0 in main /home/user/john/src/john.c:2082
    #5 0x7f77e4bbdd09 in __libc_start_main ../csu/libc-start.c:308
    #6 0x55c834eb6469 in _start (/home/user/john/run/john+0x2d8469)

0x55c837144355 is located 11 bytes to the left of global variable 'cpu_mask_ctx' defined in 'mask.c:42:25' (0x55c837144360) of size 34288
0x55c837144355 is located 37 bytes to the right of global variable 'rec_ctx' defined in 'mask.c:42:39' (0x55c83713bd40) of size 34288
SUMMARY: AddressSanitizer: global-buffer-overflow /home/user/john/src/mask.c:2459 in finalize_mask
Shadow bytes around the buggy address:
  0x0ab986e20810: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
[...]

PR #5027, opened 2022-01-27
mscash / mscash2: fixes for lowercase

### LINKED ISSUE #5026: mscash/mscash2: lowercasing truncates hashes for some utf-8 chars
Chasing bug in #4971 I found lesser problem.

I found 3 chars that increase byte length in utf-8 being lowercased.

```
0130;LATIN CAPITAL LETTER I WITH DOT ABOVE;Lu;0;L;0049 0307;;;;N;LATIN CAPITAL LETTER I DOT;;;0069;
023A;LATIN CAPITAL LETTER A WITH STROKE;Lu;0;L;;;;;N;;;;2C65;
023E;LATIN CAPITAL LETTER T WITH DIAGONAL STROKE;Lu;0;L;;;;;N;;;;2C66;
```

Code to find:
```python3
with open('/usr/share/unicode/UnicodeData.txt') as f:
    for l in f:
        c = chr(int(l.split(";")[0], 16))
        try:
            cb1 = c.encode()
        except:
            continue
        cb2 = c.lower().encode()
        if len(cb1) < len(cb2):
            print(l)
```

(Python 2 has simple support for unicode's case folding. Python 3 seems good.)

All 3 chars increase their length from 2 to 3. Buffers are allocated well for this case. There are no overflows.

But `enc_strlwr()` used in `split` does not know length of buffer, so it assumes output length to be same and truncates output.

I added hex dump around `enc_strlwr`:
```c
	printf("before:");
	for (i = 0; out[i]; i++)
		printf(" %02x", out[i] & 0xff);
	printf("\n");
	enc_strlwr(&out[2]);
	printf("after: ");
	for (i = 0; out[i]; i++)
		printf(" %02x", out[i] & 0xff);
	printf("\n");
```

I added some more `printf`s to illustrate:
```
$ python -c 'print "M$" + "\xc8\xbe" * 19 + "#" + "b" * 32' > t.pw && ./run/john --format=mscash t.pw --skip-self-tests
prepare: [0]? [1]M$ȾȾȾȾȾȾȾȾȾȾȾȾȾȾȾȾȾȾȾ#bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
valid: M$ȾȾȾȾȾȾȾȾȾȾȾȾȾȾȾȾȾȾȾ#bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
split: M$ȾȾȾȾȾȾȾȾȾȾȾȾȾȾȾȾȾȾȾ#bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
before: 4d 24 c8 be c8 be c8 be c8 be c8 be c8 be c8 be c8 be c8 be c8 be c8 be c8 be c8 be c8 be c8 be c8 be c8 be c8 be c8 be 23 62 62 62 62 62 62 62 62 62 62 62 62 62 62 62 62 62 62 62 62 62 62 62 62 62 62 62 62 62 62 62 62
after:  4d 24 e2 b1 a6 e2 b1 a6 e2 b1 a6 e2 b1 a6 e2 b1 a6 e2 b1 a6 e2 b1 a6 e2 b1 a6 e2 b1 a6 e2 b1 a6 e2 b1 a6 e2 b1 a6 e2 b1 a6 e2 b1 a6 e2 b1 a6 e2 b1 a6 e2 b1 a6 e2 b1 a6 e2 b1 a6 23 62 62 62 62 62 62 62 62 62 62 62 62 62 62
get_binary: M$ⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦ#bbbbbbbbbbbbbb
```

So after `split`, hash is incomplete because salt becomes longer. But it is accepted already. `get_binary` finds beginning of hash value with `strrchr`, then it works like there are 32 chars. So there would be wrong binary and hash would become uncrackable. It can be fixed.

Let's get sample:
```
>>> from passlib.hash import msdcc
>>> msdcc.hash("test", user="ⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦ")
'e1f96892b35a78dfc84dfcd54012b392
>>> msdcc.hash("test", user="ȾȾȾȾȾȾȾȾȾȾȾȾȾȾȾȾȾȾȾ")
'e1f96892b35a78dfc84dfcd54012b392'
```

Salt in lowercase works well:
```
$ python -c 'print "M$" + "\xe2\xb1\xa6" * 19 + "#" + "e1f96892b35a78dfc84dfcd54012b392"' > t.pw
$ ./run/john --format=mscash t.pw --skip-self-tests --mask=test
[...]
test             (?)
[...]
```

Salt in uppercase does not match crack in pot file and cannot be cracked:
```
$ python -c 'print "M$" + "\xc8\xbe" * 19 + "#" + "e1f96892b35a78dfc84dfcd54012b392"' > t.pw
$ ./run/john --format=mscash t.pw --skip-self-tests --mask=test
[... my output for loading hashes ...]
get_binary: M$ⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦ#e1f96892b35a78
Loaded 1 password hash (mscash, MS Cache Hash (DCC) [MD4 32/64])
valid: M$ⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦ#e1f96892b35a78dfc84dfcd54012b392
[... my output for loading from pot ...]
get_binary: M$ⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦⱦ#e1f96892b35a78dfc84dfcd54012b392
0g 0:00:00:00  0g/s 9.090p/s 9.090c/s 9.090C/s test
Session completed. 
```

mscash2 should have the same problem. Except its salt can be 128 chars long. So increased length would pop hash out from ciphertext completely. OTOH there is second '#'.

I guess `enc_strlwr` should be replaced with direct call to `enc_lc` that accepts explicit sizes.

AlekseyCherepanov:
Related to #4971.

Currently incomplete. Being complete it will close #5026.

### DIFF
--- a/src/mscash_common_plug.c
+++ b/src/mscash_common_plug.c
@@ -69,11 +69,23 @@ int mscash1_common_valid(char *ciphertext, struct fmt_main *self)
 	char insalt[3*MSCASH1_MAX_SALT_LENGTH+1];
 	UTF16 realsalt[MSCASH1_MAX_SALT_LENGTH+2];
 	int saltlen;
+	/* Extra +4 over similar code in split() to catch truncation (#5027). */
+	char lc_buf[MSCASH1_MAX_CIPHERTEXT_LENGTH + 1 + 4];
 
 	if (strncmp(ciphertext, FORMAT_TAG, FORMAT_TAG_LEN))
 		return 0;
 
 	l = strlen(ciphertext);
+	if (l <= 32 || l > MSCASH1_MAX_CIPHERTEXT_LENGTH)
+		return 0;
+
+	/* lowercase can transform string in unexpected ways (#5026). */
+	memcpy(lc_buf, ciphertext, FORMAT_TAG_LEN);
+	l = enc_lc((UTF8*)&lc_buf[FORMAT_TAG_LEN], sizeof(lc_buf) - 1 - FORMAT_TAG_LEN,
+	           (UTF8*)&ciphertext[FORMAT_TAG_LEN], l - FORMAT_TAG_LEN);
+	l += FORMAT_TAG_LEN;
+	ciphertext = lc_buf;
+
 	if (l <= 32 || l > MSCASH1_MAX_CIPHERTEXT_LENGTH)
 		return 0;
 
@@ -112,13 +124,12 @@ int mscash1_common_valid(char *ciphertext, struct fmt_main *self)
 char *mscash1_common_split(char *ciphertext, int index, struct fmt_main *self)
 {
 	static char out[MSCASH1_MAX_CIPHERTEXT_LENGTH + 1];
-	int i;
 
-	for (i = 0; i < MSCASH1_MAX_CIPHERTEXT_LENGTH && ciphertext[i]; i++)
-		out[i] = ciphertext[i];
-	out[i] = 0;
+	memcpy(out, ciphertext, FORMAT_TAG_LEN);
 	// lowercase salt as well as hash, encoding-aware
-	enc_strlwr(&out[2]);
+	enc_lc((UTF8*)&out[FORMAT_TAG_LEN], sizeof(out) - 1 - FORMAT_TAG_LEN,
+	       (UTF8*)&ciphertext[FORMAT_TAG_LEN], strlen(ciphertext) - FORMAT_TAG_LEN);
+
 	return out;
 }
 
--- a/src/mscash_common_plug.c
+++ b/src/mscash_common_plug.c
@@ -70,7 +70,8 @@ int mscash1_common_valid(char *ciphertext, struct fmt_main *self)
 	UTF16 realsalt[MSCASH1_MAX_SALT_LENGTH+2];
 	int saltlen;
 	/* Extra +4 over similar code in split() to catch truncation (#5027). */
-	char lc_buf[MSCASH1_MAX_CIPHERTEXT_LENGTH + 1 + 4];
+	/* Extra +1 until #5029 is fixed. */
+	char lc_buf[MSCASH1_MAX_CIPHERTEXT_LENGTH + 1 + 4 + 1];
 
 	if (strncmp(ciphertext, FORMAT_TAG, FORMAT_TAG_LEN))
 		return 0;
@@ -123,7 +124,8 @@ int mscash1_common_valid(char *ciphertext, struct fmt_main *self)
 
 char *mscash1_common_split(char *ciphertext, int index, struct fmt_main *self)
 {
-	static char out[MSCASH1_MAX_CIPHERTEXT_LENGTH + 1];
+	/* Extra +1 until #5029 is fixed. */
+	static char out[MSCASH1_MAX_CIPHERTEXT_LENGTH + 1 + 1];
 
 	memcpy(out, ciphertext, FORMAT_TAG_LEN);
 	// lowercase salt as well as hash, encoding-aware
@@ -290,7 +292,8 @@ int mscash2_common_valid(char *ciphertext, int max_salt_length, struct fmt_main 
 	UTF16 realsalt[129];
 	int saltlen;
 	/* Extra +4 over similar code in split() to catch truncation (#5027). */
-	char lc_buf[MSCASH2_MAX_CIPHERTEXT_LENGTH + 1 + 4];
+	/* Extra +1 until #5029 is fixed. */
+	char lc_buf[MSCASH2_MAX_CIPHERTEXT_LENGTH + 1 + 4 + 1];
 
 	if (strncmp(ciphertext, FORMAT_TAG2, FORMAT_TAG2_LEN))
 		return 0;
@@ -357,7 +360,8 @@ int mscash2_common_valid(char *ciphertext, int max_salt_length, struct fmt_main 
 
 char *mscash2_common_split(char *ciphertext, int index, struct fmt_main *self)
 {
-	static char out[MSCASH2_MAX_CIPHERTEXT_LENGTH + 1];
+	/* Extra +1 until #5029 is fixed. */
+	static char out[MSCASH2_MAX_CIPHERTEXT_LENGTH + 1 + 1];
 
 	memcpy(out, ciphertext, FORMAT_TAG2_LEN);
 	// lowercase salt as well as hash, encoding-aware
--- a/src/mscash_common_plug.c
+++ b/src/mscash_common_plug.c
@@ -289,6 +289,8 @@ int mscash2_common_valid(char *ciphertext, int max_salt_length, struct fmt_main 
 	char insalt[3*MSCASH2_MAX_MAX_SALT_LEN+1];
 	UTF16 realsalt[129];
 	int saltlen;
+	/* Extra +4 over similar code in split() to catch truncation (#5027). */
+	char lc_buf[MSCASH2_MAX_CIPHERTEXT_LENGTH + 1 + 4];
 
 	if (strncmp(ciphertext, FORMAT_TAG2, FORMAT_TAG2_LEN))
 		return 0;
@@ -298,6 +300,16 @@ int mscash2_common_valid(char *ciphertext, int max_salt_length, struct fmt_main 
 		return 0;
 
 	l = strlen(ciphertext);
+	if (l <= 32 || l > MSCASH2_MAX_CIPHERTEXT_LENGTH)
+		return 0;
+
+	/* lowercase can transform string in unexpected ways (#5026). */
+	memcpy(lc_buf, ciphertext, FORMAT_TAG2_LEN);
+	l = enc_lc((UTF8*)&lc_buf[FORMAT_TAG2_LEN], sizeof(lc_buf) - 1 - FORMAT_TAG2_LEN,
+	           (UTF8*)&ciphertext[FORMAT_TAG2_LEN], l - FORMAT_TAG2_LEN);
+	l += FORMAT_TAG2_LEN;
+	ciphertext = lc_buf;
+
 	if (l <= 32 || l > MSCASH2_MAX_CIPHERTEXT_LENGTH)
 		return 0;
 
@@ -346,15 +358,11 @@ int mscash2_common_valid(char *ciphertext, int max_salt_length, struct fmt_main 
 char *mscash2_common_split(char *ciphertext, int index, struct fmt_main *self)
 {
 	static char out[MSCASH2_MAX_CIPHERTEXT_LENGTH + 1];
-	int i = 0;
 
-	for (; ciphertext[i] && i < MSCASH2_MAX_CIPHERTEXT_LENGTH; i++)
-		out[i] = ciphertext[i];
-
-	out[i] = 0;
-
+	memcpy(out, ciphertext, FORMAT_TAG2_LEN);
 	// lowercase salt as well as hash, encoding-aware
-	enc_strlwr(&out[6]);
+	enc_lc((UTF8*)&out[FORMAT_TAG2_LEN], sizeof(out) - 1 - FORMAT_TAG2_LEN,
+	       (UTF8*)&ciphertext[FORMAT_TAG2_LEN], strlen(ciphertext) - FORMAT_TAG2_LEN);
 
 	return out;
 }


AlekseyCherepanov:
I tested `mscash` only: `--test` is ok, `--fuzz` is ok.

My sample from #5026 works. Should I add the sample into self-tests?

In `valid`, I use `+1` more comparing to `split`: if salt expands then it could cut garbage from end of hash. Example (rejected now):
```
python -c 'print "M$" + "\xc8\xbe" * 19 + "#" + "e1f96892b35a78dfc84dfcd54012b392AAAAAAA"'
```

Also I use value returned from `enc_lc` instead of `l = strlen(ciphertext)`.

Should I fix `mscash2` the same way?

solardiz:
> Should I fix `mscash2` the same way?

I guess so, and also the OpenCL formats need to be tested. Thanks!

AlekseyCherepanov:
Extra `+1` is not enough to catch truncation by `enc_lc` in all cases: if trailing char is not ascii, then it may not fit, so `enc_lc` would not place it into buffer. Example:
```
python -c 'print "M$" + "\xc8\xbe" * 4 + "\xe2\xb1\xa6" * 15 + "#" + "E1F96892B35A78DFC84DFCD54012B392\xf02aa"' > t.pw && ./run/john --format=mscash t.pw
```

`\xf02aa` would be cut by `enc_lc` because it does not fit into 1 additional byte. Then hash would be accepted.

So either check for 32 hex should be duplicated, or there should be bigger room for extra chars.

AlekseyCherepanov:
`\x0130` becomes two chars during lowercasing. If there is room for the first char, then it is written. So a room for one unicode char is needed.

Code of `utf16_to_utf8_r` suggests that chars longer than 4 bytes would be replaced by `UNI_REPLACEMENT_CHAR` (`0x0000FFFD`).

So 4 extra bytes should be enough.

AlekseyCherepanov:
I rewrote commit to include `+4`. Also I replaced out length for `enc_lc` as `sizeof(lc_buf) - 1 - FORMAT_TAG_LEN`, so `+4` happens in one place only. Similarly I changed in `split`.

AlekseyCherepanov:
Hm, I guess I could do `enc_lc` in `prepare` instead of `valid` and `split`. Then `split` would be empty.

Unspecific to `mscash`, I dislike slightly that `prepare` returns `split_fields[1]` for both seemingly good and obviously bad cases. I would return `NULL` for bad cases. So `prepare` could do part of `valid`'s work (and `valid` would not repeat same checks). OTOH it might be error prone. Also `prepare` cannot mark ciphertext as perfectly valid, so `valid` would repeat work anyway for this case.

magnumripper:
> Unspecific to `mscash`, I dislike slightly that `prepare` returns `split_fields[1]` for both seemingly good and obviously bad cases. I would return `NULL` for bad cases.

Agreed. I think we only later ensured that prepare could return NULL with no ill effects.

magnumripper (approved):
The idea sounds correct to me but I have no time really reviewing it.

AlekseyCherepanov:
I realized one more slight change in behaviour: hex part of ciphertext is pure ascii, lowercasing before check of hex allows non-ascii chars that become ascii during lowercasing (I don't know if any exist).

AlekseyCherepanov:
I tried to move `enc_lc` into `prepare`. It works.

It did not speed up fuzzing though (it does not seem that a few `strlen` calls affect it noticeably; maybe fuzzing is just a bad way to benchmark parsing of hashes).

There is flag `FMT_SPLIT_UNIFIES_CASE`. Should it be kept? `prepare` unifies case instead of `split`.

I tried to return `NULL` to reject in `prepare`. There is code path that does not support it: to trigger, it is needed to load a hash good for one format and a hash bad for all formats yielding `NULL` from `prepare`.

`loader.c`:
```c
			prepared = alt->methods.prepare(fields, alt);
			if (alt->methods.valid(prepared, alt)) {
				alt->params.flags |= FMT_WARNED;
				if (john_main_process)
				fprintf(stderr,
				    "Warning: only loading hashes of type "
				    "\"%s\", but also saw type \"%s\"\n"
				    "Use the \"--format=%s\" option to force "
				    "loading hashes of that type instead\n",
```

Somehow `printf` in `valid` tells that `valid` gets called with `*` as `ciphertext` during self-tests. Is it ok?

Without it, I could remove `strncmp` for tag from valid. And at this point, it might be less confusing to move everything from `valid` into `prepare` leaving only `return 1` in `valid`.

There is check `if (!split_fields[0])` in `prepare`. It does not seem needed.

Changes turned out bigger than I expected. Also I drifted from simple fix into more thoughtful changes that does not seem good now.

Should I replace commit with new changes? Or maybe share it other way? Should I postpone this change for later?

AlekseyCherepanov:
My `memcpy(out, ciphertext, FORMAT_TAG_LEN);` in `mscash1_common_split` could be replaced with `memcpy(out, FORMAT_TAG, FORMAT_TAG_LEN);`.

AlekseyCherepanov:
I just realized I got confused by `enc_strlwr`. I thought that `enc_lc` writes 1 byte more than size to put null byte. I tried separately, `enc_lc` does not exceed specified size. Then my `- 1` in arguments should be changed.

JFYI `enc_lc` with `dst_bufsize` 0 causes crash with ASan irrespective of real size of buffer.

JFYI in `unt16_lc`, `dst_len` 0 would write at index 0.
```c
int utf16_lc(UTF16 *dst, unsigned dst_len, const UTF16 *src, unsigned src_len) {
	int i, j = 0;
	UTF16 cur_src;
	for (i = 0; i < src_len; ++i, ++j, ++src, ++dst) {
		if (j == dst_len) {
			*dst = 0;
			return -j;
```

AlekseyCherepanov:
No, `enc_lc` does write 1 more byte in utf-8 mode. I was testing with code in `split` during self-tests (`--test=0`). I had to add `--encoding=utf-8` to get overflow.

Hm, so it behaves differently based on encoding. I could allocate 1 more byte to work in both cases.

Test code:
```c
	static unsigned char test[8];
	int i;
	test[7] = 'A';
	const unsigned char *tin = "asdfasdf";
	i = enc_lc(test, 8, tin, strlen(tin));
	printf("test: %s [%d]\n", test, i);
```

Backtrace from ASan:
```
    #0 0x5581b3d06cb0 in utf16_to_utf8_r /home/user/john/src/unicode.c:673
    #1 0x5581b3d0738d in utf16_to_enc_r /home/user/john/src/unicode.c:793
    #2 0x5581b3d0cca1 in enc_lc /home/user/john/src/unicode.c:1773
```

In the end, `enc_lc` calls `utf16_to_utf8_r` when in utf-8 mode. The problem is there.
```c
UTF8 *utf16_to_utf8_r(UTF8 *dst, int dst_len, const UTF16 *source)
{
	UTF8 *tpt = dst;
	UTF8 *targetEnd = tpt + dst_len;
[...]
		tpt += bytesToWrite;
		if (tpt > targetEnd) {
			tpt -= bytesToWrite;
			break;
[...]
		case 1: *--tpt =  (UTF8)(ch | firstByteMark[bytesToWrite]);
[...]
	*tpt = 0;   // <--- crash is here
	return dst;
}
```
I think there should be `>=` instead of '>'. When `tpt` is equal to `targetEnd` before addition, it would cause end of cycle and `tpt` would be reverted to `targetEnd` that is too much already.

BTW I heard that standard of C does not allow pointers past end of objects further than +1. So it is not good to add arbitrary offset and compare then. I guess `bytesToWrite >= targetEnd - tpt` would be better (@ch3root ?). I don't know if UBSan can catch it.

Should I fix `>` in `unicode.c`?

solardiz:
I also have no time for this. I asked Aleksey to help with these bugs so that I save time, and now his analysis is so deep that even to understand it I'd have to spend more time than I'd have spent on a quick fix preventing the crash. So I won't. This isn't almost clean code needing just this issue fixed, it won't be anywhere close to clean even if we fix this issue. So no point in over-spending our resources on it. Aleksey, please suggest how we fix this (just merge this PR as-is and that's it?) and move to other bugs. Thank you!

solardiz:
> > Should I fix `mscash2` the same way?
> 
> I guess so, and also the OpenCL formats need to be tested. Thanks!

What's the status on this?

> There is flag `FMT_SPLIT_UNIFIES_CASE`. Should it be kept? `prepare` unifies case instead of `split`.

Please keep it as-is, although it should not matter after 8f95245c4c10cc5d0936f30891133aeb7f945210.

> Somehow `printf` in `valid` tells that `valid` gets called with `*` as `ciphertext` during self-tests. Is it ok?

Yes, this is as intended.

> Should I replace commit with new changes? Or maybe share it other way? Should I postpone this change for later?

I did not spend time to fully understand what you're proposing, so I'll reply in general: if the fix currently in this PR is not sufficient, then definitely replace it. I think you also need to extend it for mscash2 anyway. However, if the fix is sufficient as-is, then over-complicating it might not be needed. You're the one to decide because no one else has time for this (and I think you actually don't either - you have more fuzz bugs to fix).

> standard of C does not allow pointers past end of objects further than +1.

That's correct. If you identify such cases, it's best if you fix them - if in separate parts of code from your current work's, then do it with separate commits.

> Should I fix `>` in `unicode.c`?

If it's a bug and that's a correct fix, yes. I don't know whether either or both of these are true.

As an option, you can open a separate issue for the `unicode.c` issues, and keep the fixes here limited to `mscash*` and correct even with unpatched `unicode.c` (unless that is impractical).

solardiz:
> My sample from #5026 works. Should I add the sample into self-tests?

I guess so, and similar for mscash2 too, if that would serve as regression testing for this bug.

AlekseyCherepanov:
Oops, I lost check for length.

AlekseyCherepanov:
I made a separate commit for fix for `mscash2`. It is done the same way as `mscash`.

I'll add 1 more byte to buffer to be sure in a separate commit. I guess it should not affect current code (non-utf-8 mode would not fill the buffer anyway; but I am not sure). But it will protect code from breaking with changes for #5029.

Then I'll test OpenCL.

AlekseyCherepanov:
`--test` fails for `mscash-opencl` due to #5022. `--test=0` is ok.

`--test` is ok for `mscash,mscash2,mscash2-opencl`.

I tested with and without `--encoding=utf-8`.

`mscash2-opencl` failed with `--fuzz`. I'll investigate.

AlekseyCherepanov:
Offending ciphertext:
```
python -c 'print "$DCC2$10240#" + "\x80" * 100 + "test1#607bbe89611e37446e736f7856515bf8"' > t.pw
```

Crash is in `get_salt` in `src/opencl_mscash2_fmt_plug.c:165`.

AlekseyCherepanov:
The crash happens because buffer is too small. Other `mscash*` formats use `* 3` max salt length. I just enlarged the buffer.

I found that I did not move check for `#` in `valid`. I moved it past lowercasing because it can corrupt `#`.

So +2 commits. Should I join some commits together?

AlekseyCherepanov:
`--test` is ok (`--test=0` for `mscash-opencl`). `--fuzz` is ok.

Crashes are fixed now.

`mscash-opencl` cannot crack my sample. Lowercased outside of john variant cannot be cracked too. It does not work without my changes too.

I would like to finish with this pull request and postpone remaining problems.

solardiz:
> Should I join some commits together?

This is not necessary. The first two are for different formats, and the third looks like something we might need to revert later.

I'll merge.

solardiz:
> Should I join some commits together?

This is not necessary. The first two are for different formats, and the third looks like something we might need to revert later.

I'll merge.

AlekseyCherepanov:
Thanks!

> the third looks like something we might need to revert later.

#5029 ending with `enc_lc` accepting `sizeof`-1 would lead to plain reverting. `enc_lc` accepting `sizeof` would lead to reverting and removing `- 1` in `enc_lc` call.

PR #5030, opened 2022-01-29
Fix valid() bugs found by fuzzing in several formats

AlekseyCherepanov:
Related to #4971.

`--test` is ok. `--fuzz` is ok.

### DIFF
--- a/src/FG2_fmt_plug.c
+++ b/src/FG2_fmt_plug.c
@@ -102,6 +102,8 @@ static int valid(char *ciphertext, struct fmt_main *self)
 		return 0;
 	if (strlen(ciphertext) != HASH_LENGTH)
 		return 0;
+	if (ciphertext[HASH_LENGTH - 1] != '=')
+		return 0;
 
 	return 1;
 }
--- a/src/FGT_fmt_plug.c
+++ b/src/FGT_fmt_plug.c
@@ -109,6 +109,8 @@ static int valid(char *ciphertext, struct fmt_main *self)
 		return 0;
 	if (strlen(ciphertext) != HASH_LENGTH)
 		return 0;
+	if (ciphertext[HASH_LENGTH - 1] != '=')
+		return 0;
 
 	return 1;
 }
--- a/src/lastpass_sniffed_fmt_plug.c
+++ b/src/lastpass_sniffed_fmt_plug.c
@@ -160,12 +160,12 @@ static void *get_salt(char *ciphertext)
 static void *get_binary(char *ciphertext)
 {
 	static unsigned int out[4];
-	char Tmp[48];
+	char Tmp[sizeof(out)];
 	char *p;
 	ciphertext += FORMAT_TAG_LEN;
 	p = strchr(ciphertext, '$')+1;
 	p = strchr(p, '$')+1;
-	base64_convert(p, e_b64_mime, strlen(p), Tmp, e_b64_raw, sizeof(Tmp), 0, 0);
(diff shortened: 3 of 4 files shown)


AlekseyCherepanov:
```c
	{"AK1wTiFOMv7mZOTvQNmKQBAY98hZZjSRLxAY8vZp8NlDWU=", "fortigate"},
[...]
static void * get_binary(char *ciphertext)
[...]
	char buf[SALT_SIZE+BINARY_SIZE+1];
[...]
	base64_convert(ciphertext+3, e_b64_mime, CIPHERTEXT_LENGTH, buf, e_b64_raw, sizeof(buf), flg_Base64_NO_FLAGS, 0);
```

`AK1` is tag. `CIPHERTEXT_LENGTH` does not include the tag.

After tag, there is 32 bytes encoded in base64. So base64 encoded string ends with `=`. Replacing `=` with a letter gives 33 bytes during decoding and it overflows the buffer.

`valid` could be stricter to reject everything that does not look like base64.

AlekseyCherepanov:
Oh, there is fortigate256 too. So I'll add it too here.

AlekseyCherepanov:
The same story: 44 bytes, single `=` on the end in base64.

`--test` and `--fuzz` are ok.

AlekseyCherepanov:
`lastpass` is unrelated to `fortigate`. But it has a similar problem with `base64_convert`. Hash in base64 does not have a fixed length, so I fix invocation of `base64_convert`. Also I reduced output buffer size to portion used later.

`--test` and `--fuzz` are ok.

Offending sample:
```
$lastpass$hackme@mailinator.com$5$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$00$i+hJCwPOj5eQN4tvHcMguoejx4VEmiqzOXOdWIsZKlk=
```

Currently it is loaded instead of rejection.

AlekseyCherepanov:
Now the sample is rejected.

I ensure that there are 3 fields exactly: I call `strtokm(NULL, "$")` one more time and it should return `NULL` for good ciphertexts.

`--test` and `--fuzz` are ok.

AlekseyCherepanov:
`OpenBSD-SoftRAID` has similar fix to ensure number of fields. Also I replaced `strdup`+`strrchr` with simpler code.

`--test` for cpu is ok.

Crafted sample:
```
python -c 'print "$openbsd-softraid$8192$" + "a" * 256 + "$" + "b" * 4096 + "$" + "c" * 40 + "$1$2"' > t.pw
```

Search of fields with 2 `strrchr` is sensitive to extra fields. So wrong field was treated as hash. It caused reading out of bounds.

Somehow `--fuzz` completed for me well without fixes.

Also there is opencl format I need to test.

AlekseyCherepanov:
`openbsd-softraid` and `openbsd-softraid-opencl`: `--test` is ok, `--fuzz` is ok (but it was ok before changes too), my crafted crasher is ok (rejected).

solardiz (approved):

AlekseyCherepanov:
> `openbsd-softraid` and `openbsd-softraid-opencl`: `--test` is ok, `--fuzz` is ok (but it was ok before changes too)

I got why `--fuzz` worked without changes: I used regular `--fuzz`, `--fuzz=fuzz.dic` was used in #4971. There is `fuzz.dic` in `run/`. It causes crash on a version before changes. It does not crash on a fixed version.

PR #5031, opened 2022-01-30
PHPS: made valid() stricter to avoid overflows

AlekseyCherepanov:
Related to #4971.

### DIFF
--- a/src/PHPS_fmt_plug.c
+++ b/src/PHPS_fmt_plug.c
@@ -74,7 +74,8 @@ static struct fmt_tests phps_tests[] = {
 	{"$PHPS$73616c$aba22b2ceb7c841473c03962b145feb3", "password"},
 	{"$PHPS$247824$ad14afbbf0e16d4ad8c8985263a3d051","test"},  // salt is $x$ (I want to test that a $ works)
 	{"$dynamic_6$ad14afbbf0e16d4ad8c8985263a3d051$HEX$247824","test"},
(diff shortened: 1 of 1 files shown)


AlekseyCherepanov:
One self-test was commented out with such comment: `fails $HEX$ IS required by dyna`. I don't know if it was true then, but now it is not true. After my fix it can be loaded well. So I uncommented the self-test. Also I added one.

Somehow dynamic's valid is quite permissive. There is sample that can be loaded:
```
$dynamic_6$612f95cf86c4e196f6b627bab3f01b62$HEX$$$$$$$
```

The sample does not cause a crash. So I am going to open a separate issue.

AlekseyCherepanov:
`--test` and `--fuzz` are ok.

src/PHPS_fmt_plug.c
```diff
@@ -162,10 +163,12 @@ static int phps_valid(char *ciphertext, struct fmt_main *self)
 
 		if (!val)
 			return 0;
-		cp = strrchr(ciphertext, '$');
-		if (strstr(ciphertext, "$HEX$"))
-			wanted_len += 3; // salt is in hex.
-		return cp && strlen(cp) == wanted_len;
+		cp = &ciphertext[11 + MD5_HEX_SIZE];
```
solardiz, line 166: How do we know(?) we're not(?) over-jumping beyond `ciphertext` length here?

AlekseyCherepanov: I hoped that `pDynamic_6->methods.valid` would catch shorter ciphertexts. I tried: it seems to reject them as expected.

solardiz (approved):

PR #5033, opened 2022-01-30
mscash2: #s past lowercasing, enlarge buffer (lost commits from #5027)

AlekseyCherepanov:


AlekseyCherepanov:
#5027 lost two last commits. Part of conversation happened after the commits:
> I found that I did not move check for # in valid. I moved it past lowercasing because it can corrupt #.

> So +2 commits.

The commits were pushed into the branch before the merge (commits: 22 hours ago, merging: 21 hours ago):
https://github.com/AlekseyCherepanov/JohnTheRipper/commits/fix4

I overlooked that they did not appear in the pull request. Fortunately github did not say that I can delete the branch safely. So I had to think what's wrong.

AlekseyCherepanov:
I just tested: github shows time of commit, not the time of actual push. I don't have any evidence that I pushed before merging actually. But I pushed it at some time. Well, I'll try to be more attentive. Sorry!

AlekseyCherepanov:
It made me very uneasy. My bash history tells such story: I did `git push ...` after the last commit, then I consulted `dict`. I remember that I needed `dict` to formulate question about joining commits. So it should happen before merging.

I retested: `--fuzz`, `--test`, `--test --encoding=utf-8` are ok for `mscash`, `mscash2`, `mscash2-opencl` and `mscash-opencl` (with `--test=0` for this one).

solardiz (approved):

PR #5034, opened 2022-01-30
Fix more valid() bugs found by fuzzing in several formats

AlekseyCherepanov:
Related to #4971.

### DIFF
--- a/src/leet_cc_fmt_plug.c
+++ b/src/leet_cc_fmt_plug.c
@@ -164,7 +164,9 @@ static char *prepare(char *split_fields[10], struct fmt_main *self)
 	cp = mem_alloc(strlen(split_fields[0]) + strlen(split_fields[1]) + 2);
 	sprintf(cp, "%s$%s", split_fields[0], split_fields[1]);
 	if (valid(cp, self)) {
-		return cp;
+		char *cipher = str_alloc_copy(cp);
+		MEM_FREE(cp);
+		return cipher;
 	}
 	MEM_FREE(cp);
 	return split_fields[1];
--- a/src/leet_cc_fmt_plug.c
+++ b/src/leet_cc_fmt_plug.c
@@ -131,7 +131,7 @@ static int valid(char *ciphertext, struct fmt_main *self)
 	char *p, *q;
 
 	p = ciphertext;
-	q = strchr(p, '$'); // end of salt
+	q = strrchr(p, '$'); // end of salt
 	if (!q)
 		return 0;
 
--- a/src/padlock_fmt_plug.c
+++ b/src/padlock_fmt_plug.c
@@ -141,13 +141,13 @@ static int valid(char *ciphertext, struct fmt_main *self)
 		goto bail;
 	if ((p = strtokm(NULL, "$")) == NULL) // salt
 		goto bail;
-	if (hexlenl(p, &extra) > res * 2 || extra)
+	if (hexlenl(p, &extra) != res * 2 || extra)
 		goto bail;
 	if (!ishexlc(p))
 		goto bail;
 	if ((p = strtokm(NULL, "$")) == NULL) // iv
 		goto bail;
-	if (hexlenl(p, &extra) > 16 * 2 || extra)
+	if (hexlenl(p, &extra) != 16 * 2 || extra)
 		goto bail;
 	if (!ishexlc(p))
 		goto bail;
@@ -160,7 +160,7 @@ static int valid(char *ciphertext, struct fmt_main *self)
 		goto bail;
 	if ((p = strtokm(NULL, "$")) == NULL) // add
 		goto bail;
-	if (hexlenl(p, &extra) > res * 2 || extra)
+	if (hexlenl(p, &extra) != res * 2 || extra)
 		goto bail;
 	if (!ishexlc(p))
 		goto bail;
@@ -173,7 +173,7 @@ static int valid(char *ciphertext, struct fmt_main *self)
 		goto bail;
 	if ((p = strtokm(NULL, "$")) == NULL) // ct
 		goto bail;
-	if (hexlenl(p, &extra) > res * 2 || extra)
+	if (hexlenl(p, &extra) != res * 2 || extra)
 		goto bail;
 	if (!ishexlc(p))
 		goto bail;
--- a/src/pgpwde_common_plug.c
+++ b/src/pgpwde_common_plug.c
@@ -153,11 +153,11 @@ int pgpwde_valid(char *ciphertext, struct fmt_main *self)
 		goto bail;
 	if ((p = strtokm(NULL, "*")) == NULL) // salt
 		goto bail;
-	if (hexlenl(p, &extra) > 16 * 2 || extra)
+	if (hexlenl(p, &extra) != 16 * 2 || extra)
 		goto bail;
 	if ((p = strtokm(NULL, "*")) == NULL) // ESK
 		goto bail;
-	if (hexlenl(p, &extra) > 128 * 2 || extra)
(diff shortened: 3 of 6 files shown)


AlekseyCherepanov:
padlock: `--test` and `--fuzz` are ok.

I replaced `>` with `!=` in `hexlenl(p, &extra) > 16 * 2` and similar checks. Some fields contain lengths of other fields. With `>`, relying on values from these fields causes `AddressSanitizer: heap-buffer-overflow` with reading because there may be less data than specified by field with length.

I am not 100% sure but syntax seems using exact lengths.

I'll add fixes for more formats.

AlekseyCherepanov:
`pgpwde` and `pgpwde-opencl`: `--test` and `--fuzz` are ok.

Again `hexlenl` and `>` vs `!=`.

AlekseyCherepanov:
`radius`: `--test` and `--fuzz` are ok.

I rewrote commit to exclude change from `>` to `!=`. The change might be wrong. It is not needed to fix crash.

Crash is in unhexing. It outputs 16 bytes and reads 32. `MIN_CIPHERTEXT_LENGTH` is 16. So I added `* 2` in `valid` to reject shorter hashes.

AlekseyCherepanov:
`sappse`: `--test` and `--fuzz` are ok.

`sappse-opencl`: `--test` does not work for me with and without changes (`cmp_all(1)`). `--fuzz` is ok.

I modified `valid` to check values from lengths fields against actual lengths of data fields (like in `padlock`).

AlekseyCherepanov:
`leet`: `--test` and `--fuzz` are ok.

Offending sample:
```
s$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$alt$f86036a85e3ff84e73bf10769011ecdbccbf5aaed9df0240310776b42f5bb8776e612ab15a78bbfc39e867448a08337d97427e182e72922bbaa903ee75b2bfd4
```

Simple syntax: `salt$hash`. Simple code... `get_salt` looks for end of salt with `strrchr`, `valid` uses `strchr` for the same purposes. Salt with `$` bypasses checks of length.

There is a leak in `prepare`.

AlekseyCherepanov:
Fixed. The leak was not seen with `--test`. I saw it with `--fuzz`. It happens for invalid hashes. I added a few missing lines to make code look very similar to code in `mscash`s `prepare`. `str_alloc_copy` uses `mem_alloc_tiny` that gives memory that will be freed then.

Maybe all that code should be replaced with buffer allocated once and remembered in a static variable like in most of other formats.

AlekseyCherepanov:
`zip`: `--test` and `--fuzz` are ok (`--fuzz` leaks memory as in #5023).

```
==11354==ERROR: AddressSanitizer: attempting free on address which was not malloc()-ed: 0x607000076b51 in thread T0
    #0 0x7f39911d75d7 in __interceptor_free (/usr/lib64/libasan.so.1+0x545d7)
    #1 0x6118cb in winzip_common_split /home/solar/j/syl/john/src/pkzip.c:179
```

Original code:
```c
	static uint64_t len;
	static char *buf = NULL;
[...]
	if (!buf || len < strlen(ciphertext)+1) {
		MEM_FREE(buf);
		len = strlen(ciphertext)+1;
		buf = mem_alloc_tiny(len, 1);
	}
```

So `mem_alloc_tiny` (when in debug build) returned pointer `+ align`. So it cannot be freed.

I completed `--fuzz` twice: without `+ align` and without `MEM_FREE`. Memory usage: ~744 Mb and ~756 Mb. So I guess the difference is not critical.

With `MEM_FREE` there, I got `double-free` in the end because memory from `mem_alloc_tiny` will be freed then.

To fix the problem, I just removed `MEM_FREE`. I don't know other way. Is it ok?

AlekseyCherepanov:
Well, I added `printf` to get sizes of allocations there. Total allocation is 4669 bytes in 6 allocations during fuzzing. Max allocation in one call is 1142 bytes. Max length of ciphertext was 11141, but not all ciphertexts need this path in code. So it seems ok.

I don't understand ~12 Mb difference in memory usage.

AlekseyCherepanov:
`zip,zip-opencl,pkzip`: `--fuzz` is ok, `--test` is ok with exception for `zip-opencl`. `--test=0` is ok for `zip-opencl`.

Regular `--test --format=zip-opencl` fails for me without changes too:
```
Benchmarking: ZIP-opencl, WinZip [PBKDF2-SHA1 OpenCL]... Corrupt return: Got a claimed 1102416593 cracks out of 1048576
```

solardiz:
> With `MEM_FREE` there, I got `double-free` in the end because memory from `mem_alloc_tiny` will be freed then.
> 
> To fix the problem, I just removed `MEM_FREE`. I don't know other way. Is it ok?

Yes. Of course, pointers/memory returned by `mem_alloc_tiny` cannot be freed. Another fix, and one I thing would fit this use case better, would be to switch to `mem_alloc` instead.

src/leet_cc_fmt_plug.c
```diff
@@ -164,7 +164,9 @@ static char *prepare(char *split_fields[10], struct fmt_main *self)
 	cp = mem_alloc(strlen(split_fields[0]) + strlen(split_fields[1]) + 2);
 	sprintf(cp, "%s$%s", split_fields[0], split_fields[1]);
 	if (valid(cp, self)) {
-		return cp;
+		char *cipher = str_alloc_copy(cp);
+		MEM_FREE(cp);
+		return cipher;
```
solardiz, line 169: Was there anything wrong with just returning `cp` here? Even if so, I don't see the extra allocation and copying improving anything, well, except for maybe hiding the memory leak if we "whitelist" leaks caused by `mem_alloc_tiny` (IIRC, Jim introduced debugging code to optionally forcibly free all such memory at program exit, so that we can look into the rest of the leaks without being bothered by these).

solardiz:
@AlekseyCherepanov Do you intend to add more to this PR, or is it ready to merge? Thanks!

AlekseyCherepanov:
You may merge if I don't have to fix anything according to your comments.

solardiz:
@AlekseyCherepanov You don't "have to" fix anything per my comments, but you may - your call. Also, have you taken care of all the issues in #4971 by now, or are there more to fix?

AlekseyCherepanov:
There are 3 unfixed problems remaining in #4971. So I guess memory improvements should be postponed.

solardiz:
So do you want this merged now (and you create a new PR for the 3 remaining problems) or do you want to be adding to here?

AlekseyCherepanov:
Merge it, please. Thanks!

src/leet_cc_fmt_plug.c
AlekseyCherepanov: Oh, I thought `mem_alloc_tiny` is smarter. I thought this memory would be freed once hashes are loaded.

The change silences ASan. OTOH there are other formats with similar silenced code, so I guess it might be fixed altogether later.

AlekseyCherepanov:
I tried `mem_alloc` + `MEM_FREE` in `pkzip`: it works ok. But I don't understand: it should leak memory, but ASan does not report it.

BTW sizes of allocations during fuzzing:
```
67
68
1125
1126
1141
1142
```

Allocating slightly more would reduce number of reallocations.

AlekseyCherepanov:
> Allocating slightly more would reduce number of reallocations.

I mean add +32 bytes on each allocation. Then we would have only 2 allocations and leak only 67+32 bytes (besides from final allocation).

PR #5036, opened 2022-01-31
Fix more valid() bugs found by fuzzing in several formats, part 2

AlekseyCherepanov:
Related to #4971.

### DIFF
--- a/src/MSCHAPv2_bs_fmt_plug.c
+++ b/src/MSCHAPv2_bs_fmt_plug.c
@@ -272,7 +272,7 @@ static char *prepare(char *split_fields[10], struct fmt_main *pFmt)
 		if (cp2 && cp2-cp1 == CHALLENGE_LENGTH/4) {
 			++cp2;
 			cp3 = strchr(cp2, '$');
-			if (cp3 && cp3-cp2 == CIPHERTEXT_LENGTH && (strlen(cp3) > 2 || cp3[2] != '$')) {
+			if (cp3 && cp3-cp2 == CIPHERTEXT_LENGTH && (strlen(cp3) > 2 || cp3[1] != '$')) {
 				ret = out;
 				memcpy(ret, split_fields[1], cp3-split_fields[1] + 1);
 				ret[(cp3-split_fields[1])+1] = '$';
--- a/src/MSCHAPv2_bs_fmt_plug.c
+++ b/src/MSCHAPv2_bs_fmt_plug.c
@@ -262,6 +262,9 @@ static char *prepare(char *split_fields[10], struct fmt_main *pFmt)
 	if (!strncmp(split_fields[1], FORMAT_TAG, FORMAT_TAG_LEN)) {
 		// check for a short format that has any extra trash fields, and if so remove them.
 		char *cp1, *cp2, *cp3;
+		static char *out;
+		if (!out)
+			out = mem_alloc_tiny(FORMAT_TAG_LEN + CHALLENGE_LENGTH/4 + 1 + CIPHERTEXT_LENGTH + 3, MEM_ALIGN_NONE);
 		cp1 = split_fields[1];
 		cp1 += FORMAT_TAG_LEN;
 		cp2 = strchr(cp1, '$');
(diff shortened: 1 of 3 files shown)


AlekseyCherepanov:
This one is cool: there was underflow in `salt_hash` and it was not easy to understand.

Original code:
```c
	int identity_length = ((char *)salt)[0];
	unsigned int hash;
	char *chal = ((char*)salt)+1+identity_length+1+2+8;

	hash = chal[0] + (chal[1] << 8) + (chal[2] << 16) + (((unsigned int)chal[3]) << 24);
```

The main suspect is `identity_length`. The first byte of `salt` is length in bytes of something. I checked `valid` and things seem good.

But then ASan's report gave me answer: `located 106 bytes to the left of 741-byte region` allocated in `get_salt`. It is underrun due to big negative offset. In `get_salt`, buffer has type `unsigned char *`. The length in bytes can be greater than 127. So cast to `char *` is incorrect.

JFYI there is `enc_strupper` that would truncate things incorrectly like `enc_strlwr`. But it is called in `prepare`, so it does not cause overflows. Yet some `utf-8` hashes might be loadable and uncrackable because the last field does not have precise length.

JFYI `2 *` length is passed to `enc_to_utf16`. It seems incorrect: `enc_to_utf16` treats argument as length in `UTF16`, so length should not be doubled.

AlekseyCherepanov:
`netntlmv2`: `--test` is ok, `--fuzz` was/is ok, `--fuzz=run/fuzz.dic` is ok.

AlekseyCherepanov:
`mschapv2-naive`: `--test` and `--fuzz` are ok.

Original code (de-indented):
```c
// check for a short format that has any extra trash fields, and if so remove them.
[...]
cp3 = strchr(cp2, '$');
if (cp3 && cp3-cp2 == CIPHERTEXT_LENGTH && (strlen(cp3) > 2 || cp3[2] != '$')) {
	ret = str_alloc_copy(split_fields[1]);
	ret[(cp3-split_fields[1])+1] = '$';
	ret[(cp3-split_fields[1])+2] = 0;
```

Crasher from fuzzer:
```
$MSCHAPv2$8710da60ebfc4cab$c4e3bb55904c966927ee68e5f1472e1f5d8ec165713b5360$
```

There is similar self-test:
```
{"$MSCHAPv2$8710da60ebfc4cab$c4e3bb55904c966927ee68e5f1472e1f5d8ec165713b5360$$", "bar4" },
```

So the code should drop third field and further and put `$`, so there would `$$` on the end. But empty last field means that we would append `$`, while `str_alloc_copy` gives us buffer without room for that.

Skipping `prepare`, hashes with single `$` on the end are loaded well and can be cracked. So I chose thorough approach and replaced `str_alloc_copy` with pre-allocated buffer with exact needed size. So hashes are saved into pot file with `$$` like others.

AlekseyCherepanov:
`mschapv2-naive`: `--test` and `--fuzz` are ok.

`cp3[2] != '$'` should be `cp3[1] != '$'`. For `strlen` 1, it reads too far, but it does not trigger crash in ASan.

There are other fields past end of ciphertext. So it is possible to trigger incorrect behaviour with such input file:
```
:$MSCHAPv2$8710da60ebfc4cab$c4e3bb55904c966927ee68e5f1472e1f5d8ec165713b5360$:$
```
Hash loaded from this line would not get second `$` before the fix.

AlekseyCherepanov:
Idea: copy fields into separate buffers in ASan builds, so overflows would be caught. Also read-only buffers for fields might be interesting.

AlekseyCherepanov:
`telegram,telegram-opencl`: `--test` and `--fuzz` are ok.

Fuzzer found such sample for me:
```
$telegram$1*4000*e693c27ff92fe83a5a247cce198a8d6a0f3a89ffedc6bcddbc39586bb1bcb50b**[... 575 hex chars ...]
```

ASan caught negative length for `memcpy`:
```
==29678==ERROR: AddressSanitizer: negative-size-param: (size=-16)
    #0 0x7fe140ba838a in __interceptor_memcpy ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:737
    #1 0x5594679ba191 in telegram_check_password /home/user/john/src/telegram_common_plug.c:40
```

Respective code:
```c
	int encrypted_data_length = cs->encrypted_blob_length - 16;
[...]
	memcpy(encrypted_data, cs->encrypted_blob + 16, encrypted_data_length);
```

There are no fields with lengths. So `**` is an empty data field that produces length 0. I added check for minimal acceptable length.

Also I added check for additional fields.

Original code:
```c
	if ((p = strtokm(NULL, "*")) == NULL)  // encrypted_blob
		goto err;
	if (hexlenl(p, &extra) > ENCRYPTED_BLOB_LEN * 2 || extra)
		goto err;
```

Looking at this code, I had to think twice to understand that `extra` is 0 because `hexlenl` goes to end of field cut by `strtokm`. So `extra` tells nothing about garbage on the end in a form of additional fields.

AlekseyCherepanov:
This PR is ready! I worked over all ASan reports in #4971. But I did not use `fuzz_option.pl` or `--fuzz=run/fuzz.dic` (in most cases).

PR #5039, opened 2022-02-01
Unicode tweaks to recently added rules

magnumripper:
It's unclear how these got into the rule set (well it IS clear that we're dealing with double-conversions, but not exactly how), and they were obviously not of any value as written.  This commit fixes it for most situations where they'd be applicable at all.

### DIFF
--- a/run/rules/OneRuleToRuleThemAll.rule
+++ b/run/rules/OneRuleToRuleThemAll.rule
@@ -8207,9 +8207,11 @@ $o$p
 ^o^o^z
 ^o^o^w
 ^o^o^t
-!! hashcat logic OFF
-^o^\x85^\xc4^\x82^\xc4^o^t
-!! hashcat logic ON
+# This was originally ^o^ą^Ă^o^t, which obviously must have been thrashed by
+# multiple codepage conversions along the way in unknown history.  Assuming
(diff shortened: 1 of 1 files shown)


magnumripper:
I also see this:
```diff
diff --git a/run/rules/OneRuleToRuleThemAll.rule b/run/rules/OneRuleToRuleThemAll.rule
index 78da96396..694295446 100644
--- a/run/rules/OneRuleToRuleThemAll.rule
+++ b/run/rules/OneRuleToRuleThemAll.rule
@@ -28,7 +28,7 @@
 ,4,5
 ,3,4
 ,2,6
-,2,5 
+,2,5
 ,2,4
 ,2,3
 ,2,1*02
```
There's a whitespace after `,2,5`. I'm investigating the meaning of this, if any.

magnumripper:
> There's a whitespace after `,2,5`. I'm investigating the meaning of this, if any.

OK so that space should be a no-op - it says "replace character in position 2 with the one preceeding it, and the same for character in position 5"

Anyway, I could include a fix for it in this PR.

solardiz:
> Anyway, I could include a fix for it in this PR.

Up to you. Thanks!

solardiz (approved):

PR #5044, opened 2022-02-02
bitcoin-opencl: Reduce nvidia busy-waits

solardiz:
Only tested with `--test` for now.


solardiz:
Briefly tested with several different test wallets' hashes with different iteration counts, loaded at once, on an overheating and throttling GPU - this code handled those difficulties fine, albeit eventually bringing CPU usage up to ~30% of one core.

solardiz:
> on an overheating and throttling GPU - this code handled those difficulties fine, albeit eventually bringing CPU usage up to ~30% of one core.

Using this on more reasonable (non-overheating) GPUs now, the CPU usage is typically 5% to 9%.

PR #5040, opened 2022-02-01
mask: added check of number of placeholders and ranges, fixes overflow

AlekseyCherepanov:
Closes #4115.

ASan showed where is the problem. For `?l?l...`:
```
==1869==ERROR: AddressSanitizer: global-buffer-overflow on address 0x56213f22b448 at pc 0x56213d5eafd1 bp 0x7ffccd2e5af0 sp 0x7ffccd2e5ae8
WRITE of size 4 at 0x56213f22b448 thread T0
    #0 0x56213d5eafd0 in parse_qtn /home/user/john/src/mask.c:1160
    #1 0x56213d5f73fc in finalize_mask /home/user/john/src/mask.c:2356
    #2 0x56213d5f9895 in mask_init /home/user/john/src/mask.c:2312
    #3 0x56213d5d7676 in john_run /home/user/john/src/john.c:1777
[...]

0x56213f22b448 is located 56 bytes to the left of global variable 'mask_gpu_is_static' defined in 'mask_ext.c:22:5' (0x56213f22b480) of size 4
0x56213f22b448 is located 0 bytes to the right of global variable 'parsed_mask' defined in 'mask.c:41:24' (0x56213f22ae60) of size 1512
```

For `[ab][ab]...`:
```
==3462==ERROR: AddressSanitizer: global-buffer-overflow on address 0x55dc83b1c448 at pc 0x55dc81ee1ce6 bp 0x7fff2b185ee0 sp 0x7fff2b185ed8
WRITE of size 4 at 0x55dc83b1c448 thread T0
    #0 0x55dc81ee1ce5 in parse_braces /home/user/john/src/mask.c:1112
    #1 0x55dc81ee83c9 in finalize_mask /home/user/john/src/mask.c:2349
    #2 0x55dc81eea895 in mask_init /home/user/john/src/mask.c:2312
    #3 0x55dc81ec8676 in john_run /home/user/john/src/john.c:1777
[...]

0x55dc83b1c448 is located 56 bytes to the left of global variable 'mask_gpu_is_static' defined in 'mask_ext.c:22:5' (0x55dc83b1c480) of size 4
0x55dc83b1c448 is located 0 bytes to the right of global variable 'parsed_mask' defined in 'mask.c:41:24' (0x55dc83b1be60) of size 1512
```

New behaviour:
```
$ run/john --mask="`python -c 'print "?l" * 126'`" --stdout --verbosity=1
Warning: Verbosity decreased to minimum, candidates will not be printed!
[...]

Issue #4115, opened 2019-10-11
mask with a lot (>260) of ranges/placeholders crashes with sigsegv/segfault


AlekseyCherepanov:
Actually @rofl0r has found it.

The crash:
```
$ gdb ./JohnTheRipper/run/john
[...]
(gdb) run  --mask="`python -c 'print "[ab]" * 1000'`" --stdout
[...]

Program received signal SIGSEGV, Segmentation fault.
_IO_vfscanf_internal (s=0x6bc000006b8, format=0x555555b7a917 "%lu\n", 
    argptr=argptr@entry=0x7fffffffcb98, errp=errp@entry=0x0) at vfscanf.c:347
347	vfscanf.c: Нет такого файла или каталога.
(gdb) bt
#0  _IO_vfscanf_internal (s=0x6bc000006b8, format=0x555555b7a917 "%lu\n", 
    argptr=argptr@entry=0x7fffffffcb98, errp=errp@entry=0x0) at vfscanf.c:347
#1  0x00007ffff5f78db7 in ___vfscanf (s=<optimized out>, 
    format=<optimized out>, argptr=argptr@entry=0x7fffffffcb98)
    at vfscanf.c:3066
#2  0x00007ffff5f801c7 in __fscanf (stream=stream@entry=0x6bc000006b8, 
    format=format@entry=0x555555b7a917 "%lu\n") at fscanf.c:31
#3  0x00005555558ac390 in mask_restore_state (file=0x6bc000006b8)
    at mask.c:1913
#4  0x00005555558bfb1c in rec_restore_mode (
    restore_mode=restore_mode@entry=0x5555558ac360 <mask_restore_state>)
    at recovery.c:653
#5  0x00005555558b42a6 in mask_crk_init (db=0x5555571d22a0 <database>)
    at mask.c:2446
#6  0x00005555558a4ef3 in john_run () at john.c:1941
#7  main (argc=<optimized out>, argv=<optimized out>) at john.c:2221
```

For me, the problem starts with 261 or more ranges. Ranges of single letter (e.g. `[a]`) works well. `?l` x 261 crashes too.
```
$ ./JohnTheRipper/run/john --mask="`python -c 'print "?l" * 261'`" --stdout --verbosity=1
Warning: Verbosity decreased to minimum, candidates will not be printed!
Segmentation fault
```

It crashes on `super` too.

john is quite fresh (has opencl; openmp is disabled):
```
[...]

PR #5045, opened 2022-02-02
mask: fix crash on `"?l[aa]" * 125`

AlekseyCherepanov:
#5040 fixed problem with too many placeholders (#4115) and similar problem with too many ranges. But `"?l[aa]" * 125` causes crash in other place. `* 124` does not crash.

```
$ run/john --mask="`python -c 'print "?l[aa]" * 125'`" --stdout
==21478==ERROR: AddressSanitizer: global-buffer-overflow on address 0x55cf2ab055c8 at pc 0x55cf28ec2b32 bp 0x7ffc67081b70 sp 0x7ffc67081b68
READ of size 4 at 0x55cf2ab055c8 thread T0
    #0 0x55cf28ec2b31 in search_stack /home/user/john/src/mask.c:1182
    #1 0x55cf28ed069c in finalize_mask /home/user/john/src/mask.c:2372
    #2 0x55cf28ed2a05 in mask_init /home/user/john/src/mask.c:2322
[...]
0x55cf2ab055c8 is located 56 bytes to the left of global variable 'mask_gpu_is_static' defined in 'mask_ext.c:22:5' (0x55cf2ab05600) of size 4
0x55cf2ab055c8 is located 0 bytes to the right of global variable 'parsed_mask' defined in 'mask.c:41:24' (0x55cf2ab04fe0) of size 1512
SUMMARY: AddressSanitizer: global-buffer-overflow /home/user/john/src/mask.c:1182 in search_stack
```

I just added checks to end loops in the crashing function before reading past max number of elements.

Unless there would be comments requiring a changes, this PR is ready.

### DIFF
--- a/src/mask.c
+++ b/src/mask.c
@@ -1082,7 +1082,8 @@ static void parse_braces(char *mask, mask_parsed_ctx *parsed_mask)
 	int i, j ,k;
 	int cl_br_enc;
 
-	for (i = 0; i < MAX_NUM_MASK_PLHDR; i++) {
+	/* The last element is worst-case boundary for search_stack(). */
+	for (i = 0; i <= MAX_NUM_MASK_PLHDR; i++) {
 		store_cl(i, -1);
 		store_op(i, -1);
 	}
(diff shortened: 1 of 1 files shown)


AlekseyCherepanov:
Other way to fix: initialize the additional element actually. Other elements are initialized with `-1`, so it would be changes from `<` to `<=` in the loops with initialization.

It might explain the original intention behind the additional element.

The code looks like it was written with optimizations in mind. OTOH the whole parsing stage could be done in one pass instead of multiple transformations.

AlekseyCherepanov:
I replaced commit with `<` to '<=' change plus a comment. Also I changed commit message.

solardiz:
@AlekseyCherepanov Is this ready to merge? Thanks!

AlekseyCherepanov:
@solardiz yes, this PR is ready! thanks!

Issue #4728, opened 2021-05-12
Opportunistic dupe suppression for wordlist+rules


solardiz:
We currently have full duplicate candidate suppression in the separate `unique` tool, and we also reuse code from there to suppress duplicate input wordlist entries and wordlist rules. However, when we're running through wordlist+rules (maybe after suppression of duplicates within the wordlist and within the rules separately), we only suppress consecutive duplicates (compare against the immediately preceding candidate password).

I think we could (optionally) do better (for slow hashes or/and many salts) by maintaining and comparing against a configurable-size buffer of previous candidates.

In terms of implementation, I think this could be a ring buffer and a hash table. In `unique`, we use a non-ring buffer and a hash table. We reset them each time the buffer fills up. I think for this new feature we would instead need to be progressively removing old entries from the hash table as the ring buffer's contents are about to be overwritten with new entries. Further, with variable-size entries (like `unique` already uses for different string lengths), we need to support a case when adding just one new entry requires removal of several previous entries from the hash table first.

We can possibly improve the effectiveness of such dupe suppression (percentage suppressed per buffer size) by maintaining (recent) hit counts on entries, so that frequently or/and recently hit entries would be retained for longer even if they're near the overwrite pointer (were entered into the ring buffer long ago). We can either keep them intact (jump over them in our overwrites) or re-add them right away (if needed to avoid gaps that would result from changes in nearby string lengths).

We can also reuse this logic to implement a hashcat brain alike later. Per my reading of hashcat brain documentation, it cannot maintain a maximum size and thus could eventually run out of storage resources and fail. The alternative I suggest here can instead maintain a maximum size, at the expense of imperfect dupe suppression. So it could make sense as an alternative brain implementation, including for use with hashcat if we implement the same protocol.

[...]

PR #5018, opened 2022-01-24
Rules update

solardiz:
This is my work-in-progress on #4851, #5003, #5004.

PR #5046, opened 2022-02-02
Dupe suppression

solardiz:
Initial working implementation of #4728

### DIFF
--- a/doc/EXTERNAL
+++ b/doc/EXTERNAL
@@ -57,7 +57,8 @@ current cracking session to be aborted or the status line to be
 displayed (just like on a keypress), respectively.  These actions are
 taken after having tested at least all of the candidate passwords that
 were in external mode's "word" so far.  In other words, the actions may
-be delayed in order to process any buffered candidate passwords.
+be delayed in order to process any buffered candidate passwords.  Setting
+"status" to 2 results in printing of detailed multi-line status.
 
 "cipher_limit" contains the maximum password length in bytes, either from
 the format definition or from --stdout[=LENGTH].  This variable should
--- a/doc/OPTIONS
+++ b/doc/OPTIONS
@@ -76,18 +76,38 @@ Set the terminal up for reading status/quit keystrokes even if we're not
 the foreground process.  The downside is we might get race conditions so
 you may end up with strange problems hard to reproduce.
 
---dupe-suppression		suppress all duplicates from wordlist
+--dupe-suppression[=SIZE]	suppress some duplicates from wordlist+rules
 
-Normally, consecutive duplicates are ignored when reading a wordlist file.
-This switch enables full dupe suppression, using some memory and a little
-extra start-up time.  This option implies preload regardless of file size,
-see the --mem-file-size option.
+Opportunistic duplicate candidate password suppression for wordlist+rules.
+This creates a buffer in memory, which is then used to detect and suppress
+some (hopefully, most) duplicates resulting from application of wordlist
+rules (or otherwise, but rules are the most common source of duplicates).
+
+When you use wordlist rules (explicitly or as part of batch mode), this is
+enabled by default, but it also gets automatically disabled if efficiency
+isn't high enough (ratio of candidate passwords suppressed does not look
+high enough to justify the overhead given the current attack's p/s rate).
+
+When in batch mode, the suppression can continue into its incremental mode
+pass - to suppress incremental mode's passwords that happen to match those
+previously generated in wordlist mode.
+
+When enabled along with --rules-stack, the suppression is applied before the
+stacked rules (but after the main --rules, which would be the reason to have
+it enabled).
+
+You can use this option to override the default behavior, such as to enable
+suppression for a wordlist without rules, keep it enabled regardless of
+current efficiency (e.g., in case you expect more duplicates later in the
+run), disable suppression despite of use of rules (--dupe-suppression=0
+does that), or request a specific allocation size (in MiB).  See also
+[Options:Suppressor] in john.conf.
 
 --loopback[=FILE]		use (a) pot file as a wordlist
 
-This mode implies --dupe-suppression.  This is a special variant of wordlist
-mode.  Parse the file as a .pot file, so only the actual password part of
-each line will be used.  If FILE is not specified, the active .pot file will
+This is a special variant of wordlist mode.  Parse the file as a .pot file,
+so only the actual password part of each line will be used, and duplicate
+passwords suppressed.  If FILE is not specified, the active .pot file will
 be used.  This is very effective for applying rules to further permute words
 that has been successful, or without rules after a run using --salts in case
 the cracks are successful on other salts than the ones that has already been
--- a/run/john.conf
+++ b/run/john.conf
@@ -368,6 +368,11 @@ SAPhalfHashes = N
 # The default is 256 MiB, the maximum is just below 256 GiB (on 64-bit).
 Size = 256
 
+# Whether to lock the oldest half of entries as write-once (and only ever
+# update the other half).  Empirically, this often provides best results,
+# but it can also backfire.  The default is yes.
+LockHalf = Y
+
 [Options:CPUtune]
 # If preset is given, use it and skip autotune (NOTE: non-intel archs will
 # currently ignore this option and always autotune)
--- a/run/john.conf
+++ b/run/john.conf
@@ -362,6 +362,12 @@ AbortGraceTime = 30
 # This may produce some false positives if enabled, at least for SAP-B.
 SAPhalfHashes = N
 
+# Opportunistic duplicate candidate password suppressor for wordlist+rules
+[Options:Suppressor]
+# Memory allocation size in MiB per process, 0 to disable unless forced.
+# The default is 256 MiB, the maximum is just below 256 GiB (on 64-bit).
+Size = 256
+
 [Options:CPUtune]
 # If preset is given, use it and skip autotune (NOTE: non-intel archs will
 # currently ignore this option and always autotune)
--- a/src/Makefile.in
+++ b/src/Makefile.in
@@ -103,7 +103,7 @@ JOHN_OBJS = \
 	batch.o bench.o charset.o common.o compiler.o config.o cracker.o crc32.o external.o \
 	formats.o getopt.o idle.o inc.o john.o list.o loader.o logger.o mask.o mask_ext.o \
 	memory.o misc.o options.o params.o path.o recovery.o rpp.o rules.o signals.o single.o status.o \
-	tty.o  wordlist.o \
+	suppressor.o tty.o wordlist.o \
 	mkv.o mkvlib.o \
 	subsets.o unicode_range.o \
 	listconf.o \
--- a/src/Makefile.legacy
+++ b/src/Makefile.legacy
@@ -148,7 +148,7 @@ JOHN_OBJS = \
 	crc32.o external.o formats.o getopt.o idle.o inc.o john.o list.o \
 	loader.o logger.o mask.o mask_ext.o memory.o misc.o options.o \
 	params.o path.o recovery.o rpp.o rules.o signals.o single.o status.o \
-	tty.o wordlist.o \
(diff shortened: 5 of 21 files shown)


solardiz:
This works for wordlist(+rules) and a subsequent incremental mode (if in batch mode).

However, as currently implemented it "does not work" for `--rules-stack` (suppresses dupes only prior to the stacked rules). Thinking of this, it'd actually make sense to have two instances of the suppressor in that case - before and after the stacked rules - but this is tricky, and I currently won't bother.

Also, the knobs are not completed - this is just undocumented extension of `--dupe-suppression`. We'll need to document it, add a way to control the memory usage, and perhaps drop the old functionality of that option.

solardiz:
Full RockYou with `best64` rules, with `--dupe-suppression` as currently committed here, so 256 MiB memory usage:

```
729954712p 0:00:03:46 100.00% (2022-02-02 23:18) 3219Kp/s
Total lines read: 729954712, unique lines written: 685780342 (93%), no slow passes
```

without `--dupe-suppression`:

```
957872315p 0:00:03:43 100.00% (2022-02-02 23:46) 4285Kp/s
Total lines read: 957872315, unique lines written: 685780342 (71%), no slow passes
```

The file output by `unique` is ~7 GB. I think it's pretty good we avoid most dupes with a cache that's ~1/30 the size of the (remaining) candidates list.

solardiz:
Also testing with our new default wordlist+rules, which are not supposed to produce many dupes as-is and which produce too many candidates for efficient use of the default 256 MiB cache:

```
$ ./john -form=nt -verb=1 -encoding=cp1252 -w -ru -progress-every=120 -dupe-suppression pp8-sample-100m
Loaded 53848563 password hashes with no different salts (NT [MD4 128/128 AVX 4x3])
Proceeding with wordlist:./password.lst, rules:Wordlist
11980332g 0:00:02:00 8.19% (ETA: 12:30:16) 99794g/s 3267Kp/s 3267Kc/s 146095GC/s amahacke..uminaz34
14283927g 0:00:04:00 19.41% (ETA: 12:26:27) 59508g/s 3711Kp/s 3711Kc/s 157254GC/s dinamo09201..urmama12201
15360605g 0:00:06:00 31.08% (ETA: 12:25:09) 42667g/s 3940Kp/s 3940Kc/s 161968GC/s yamyrelle..yaaicul
16041937g 0:00:07:59 42.62% (ETA: 12:24:35) 33421g/s 4057Kp/s 4057Kc/s 163835GC/s _kittens95_.._dynamic5_
16562762g 0:00:09:59 54.16% (ETA: 12:24:18) 27605g/s 4172Kp/s 4172Kc/s 165843GC/s 282282282su..carpet8su
16960567g 0:00:11:59 64.78% (ETA: 12:24:21) 23557g/s 4218Kp/s 4218Kc/s 165700GC/s dunnie1560..bobo80560
17309897g 0:00:13:59 75.54% (ETA: 12:24:22) 20608g/s 4260Kp/s 4260Kc/s 165658GC/s pass1919726..jen911726
17635546g 0:00:15:59 87.13% (ETA: 12:24:12) 18371g/s 4341Kp/s 4341Kc/s 167265GC/s txlee1979..txbrutal69
17897569g 0:00:17:59 98.81% (ETA: 12:24:04) 16573g/s 4390Kp/s 4390Kc/s 167742GC/s 545422bl..0angelbl
17906157g 0:00:18:04 DONE (2022-02-03 12:23) 16505g/s 4389Kp/s 4389Kc/s 167648GC/s Gunot0138..Sambarock38
```

This is to be compared against https://github.com/openwall/john/pull/5018#issuecomment-1026286263:

```
$ ./john -form=nt -verb=1 -encoding=cp1252 -w -ru -progress-every=120 pp8-sample-100m
Loaded 53848563 password hashes with no different salts (NT [MD4 128/128 AVX 4x3])
Proceeding with wordlist:./password.lst, rules:Wordlist
13694837g 0:00:02:01 15.32% (ETA: 23:11:37) 113105g/s 6380Kp/s 6380Kc/s 274317GC/s 020f00..042f
15664147g 0:00:04:01 35.78% (ETA: 23:09:41) 64977g/s 7291Kp/s 7291Kc/s 297171GC/s xandgee..1983g1958
16624474g 0:00:06:01 55.71% (ETA: 23:09:15) 46044g/s 7594Kp/s 7594Kc/s 301096GC/s digestive565..robbie20565
17341434g 0:00:08:01 76.60% (ETA: 23:08:55) 36050g/s 7936Kp/s 7936Kc/s 308370GC/s nghia123443..10121968443
17862308g 0:00:10:01 97.02% (ETA: 23:08:47) 29720g/s 8087Kp/s 8087Kc/s 309690GC/s coolcat4..gore4
17906157g 0:00:10:14 DONE (2022-01-31 23:08) 29144g/s 8097Kp/s 8097Kc/s 309632GC/s Chokolate538..Sambarock38
```

Same guess count, much longer run time (dupe suppression costs way more than it helps when the hash is fast and there are few dupes).

Unfortunately, we don't report the candidate password count, but going from p/s and time, after dupe suppression we have:

`4389*(18*60+4)/10^6  = 4.76 billion`

Before we had:

`8097*(10*60+14)/10^6 = 4.97 billion`

So there _is_ an improvement by 4.3%, which would be significant for slow/salted hashes. So it _does_ make sense to have dupe suppression enabled in batch mode, but we need to somehow get it auto-disabled when running on fast hashes with no or few salts - maybe by actual p/s rate.

solardiz:
I am getting a couple of extra commits in here as I need the detailed status to properly tune this:

- Print detailed status on 's' keypress
- Print a help message on 'h' keypress

solardiz:
Added support for arbitrary (not only powers of two) configurable allocation sizes (up to almost 256 GiB). Had to revise the hash function to accommodate that - luckily, now it not only produces more uniform values, but is also simpler.

solardiz:
This is now a "release candidate" - I am likely to merge this PR tomorrow, and it will close the related issue, even though not all of the ideas have been implemented yet (some might never be, or maybe much later).

solardiz:
> * Print detailed status on 's' keypress

For the detailed status, I now print the usual status line followed by multi-line detail that mostly does not overlap with what's on the first line. I think we should later revise this to only print a multi-line status, not print the usual status line, but to include all of its info into the multi-line status in a consistent form.

magnumripper:
Note that we also have `--prince-loopback` option that also sets `FLG_DUPESUPP` but has its own code path for it in `pp.c`. Perhaps it's unaffected by these changes? Trying it, your new dupe suppression seem to remain disabled.

solardiz:
> Note that we also have `--prince-loopback` option that also sets `FLG_DUPESUPP` but has its own code path for it in `pp.c`. Perhaps it's unaffected by these changes? Trying it, your new dupe suppression seem to remain disabled.

Oh, I missed that, but looks like we got lucky - I assume it continuing to work like before is the intended behavior, and we don't need the new suppressor there? Or maybe we do need it for PRINCE when `--dupe-suppression` is explicitly requested? Anyway, the latter is for another PR - I have no time to test more things now.

Checking for other uses of `FLG_DUPESUPP`, I now see it's also in `--mem-file-size` - perhaps we used to disallow the combination, but should allow it now? That would be something to fix in _this_ PR.

solardiz:
> Checking for other uses of `FLG_DUPESUPP`, I now see it's also in `--mem-file-size` - perhaps we used to disallow the combination

...because `FLG_DUPESUPP` implied equivalent of `--mem-file-size`, making the option redundant - but that's not the case anymore.

magnumripper:
> Checking for other uses of `FLG_DUPESUPP`, I now see it's also in `--mem-file-size` - perhaps we used to disallow the combination, but should allow it now? That would be something to fix in _this_ PR.

Yes, I think we should allow that now.

magnumripper:
PRINCE mode should also auto-enable this when rules are used. I tried the below but it doesn't seem enough
```diff
diff --git a/src/pp.c b/src/pp.c
index 8a2c58458..50d4c3e38 100644
--- a/src/pp.c
+++ b/src/pp.c
@@ -114,6 +114,7 @@
 #include "recovery.h"
 #include "options.h"
 #include "external.h"
+#include "suppressor.h"
 #include "cracker.h"
 #include "john.h"
 #include "unicode.h"
@@ -1330,6 +1331,11 @@ void do_prince_crack(struct db_main *db, const char *wordlist, int rules)
     struct rpp_context ctx, *rule_ctx;
     int active_rules = 0, rule_number = 0;
 
+    if (dupe_check || rules) {
+      int force = (dupe_check || (options.flags & FLG_STDOUT)) && options.suppressor_size;
+      suppressor_init(SUPPRESSOR_UPDATE | (force ? SUPPRESSOR_FORCE : 0));
+    }
+
     if (options.activewordlistrules)
       log_event("- Rules: %.100s", options.activewordlistrules);
 
```
I do get `Enabling duplicate candidate password suppressor` but the statistics are all zero.

solardiz:
> I tried the below but it doesn't seem enough

I think you need to move this to below `crk_init()`, which is further down the source file.

magnumripper:
That worked! Here's a complete patch
```diff
diff --git a/src/pp.c b/src/pp.c
index 8a2c58458..1a87ffc1b 100644
--- a/src/pp.c
+++ b/src/pp.c
@@ -114,6 +114,7 @@
 #include "recovery.h"
 #include "options.h"
 #include "external.h"
+#include "suppressor.h"
 #include "cracker.h"
 #include "john.h"
 #include "unicode.h"
@@ -1974,6 +1975,11 @@ void do_prince_crack(struct db_main *db, const char *wordlist, int rules)
   mpf_mul_ui(count, count, rule_count);
 
   crk_init(db, fix_state, NULL);
+
+  if (dupe_check || rules) {
+    int force = (dupe_check || (options.flags & FLG_STDOUT)) && options.suppressor_size;
+    suppressor_init(SUPPRESSOR_UPDATE | (force ? SUPPRESSOR_FORCE : 0));
+  }
 #endif
 
   /**
```

solardiz:
> That worked! Here's a complete patch

I also just built with similar - testing...

solardiz:
However, can't specify explicit `-dup` along with `-prince` - I think we should allow that, but we can't(?) without reworking the related option flags.

magnumripper:
On another note I noticed delayed status never(?) prints in stdout mode. Not a big deal, but maybe you want to fix that.

solardiz:
> On another note I noticed delayed status never(?) prints in stdout mode. Not a big deal, but maybe you want to fix that.

If it's not a new bug with this PR, then I won't bother right now.

magnumripper:
> > On another note I noticed delayed status never(?) prints in stdout mode. Not a big deal, but maybe you want to fix that.
> 
> If it's not a new bug with this PR, then I won't bother right now.

Oh right, that was older code

solardiz:
> explicit `-dup` along with `-prince` - I think we should allow that, but we can't(?) without reworking the related option flags.

I think I'll simply allow for any cracking mode now, even though most won't call `init`. We can restrict this later.

solardiz:
Oh wait, we have `FLG_RULES_ALLOW`. Maybe that's it.

solardiz:
> Oh wait, we have `FLG_RULES_ALLOW`. Maybe that's it.

That worked! Pushed another commit with the PRINCE stuff. I really would like to merge this PR today, and move on.

magnumripper:
Isn't there any other mode we'd want this in? Such as external modes? IIRC, the external:subsets mode produce lots of dupes (or was it some other one).

solardiz:
> IIRC, the external:subsets mode produce lots of dupes (or was it some other one).

Oh, right. However, my checking `FLG_RULES_ALLOW` wouldn't work for that. So I think that's for a later rework of option flags, or we need to allow for any `FLG_CRACKING_CHK` and accept that only some modes will actually enable the suppressor.

solardiz:
Also, we'd need to edit the option description _again_ if it's no longer for rules. I am tired now, and have other work to do.

magnumripper:
Things seem to work fine now, let's merge and build on it later

Issue #5032, opened 2022-01-30
dynamic's valid is too permissive


AlekseyCherepanov:
Working on #5031 for PHPS that relies on `pDynamic_6->methods.valid`, I found that dynamic's valid is quite permissive.

It does not reject garbage after `$HEX$`:
```
$dynamic_6$612f95cf86c4e196f6b627bab3f01b62$HEX$$$$$$$
```

Is it ok?

I tried to make samples expecting that salt is `HEX$$$$$$$` or salt is empty. I could not crack them.

It affects PHPS format allowing it to load such hashes while it should not (its salt is 3 chars exactly).

AlekseyCherepanov:
Also fuzzing `phps` yielded 2 samples that crash `dynamic_6` when in ASan build:
```
$dynamic_6$ad14afbbf0e16d4ad8c8985263a3d051$HEX$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$247824
$dynamic_6$ad14afbbf0e16d4ad8c8985263a3d051$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$HEX$247824
```

They produce similar outputs from ASan:
```
$ ./run/john --format=dynamic_6 t.pw
[...]
==26330==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x63300001ac00 at pc 0x556af7e5b60a bp 0x7fff7fbed1c0 sp 0x7fff7fbed1b8
WRITE of size 4 at 0x63300001ac00 thread T0
    #0 0x556af7e5b609 in __SSE_append_string_to_input /home/user/john/src/dynamic_fmt.c:3327
    #1 0x556af7e62c2a in __append_string /home/user/john/src/dynamic_fmt.c:3375
    #2 0x556af7e63134 in DynamicFunc__append_salt /home/user/john/src/dynamic_fmt.c:4940
    #3 0x556af7e7f02c in crypt_all /home/user/john/src/dynamic_fmt.c:1898
    #4 0x556af822283c in crk_password_loop /home/user/john/src/cracker.c:894
[...]

PR #5047, opened 2022-02-03
dynamic: tweaks for valid

AlekseyCherepanov:
Related to #5032.

### DIFF
--- a/src/dynamic_fmt.c
+++ b/src/dynamic_fmt.c
@@ -2142,7 +2142,7 @@ static int get_hash_6(int index)
 
 /************************************************************************
  * We now fully handle all hashing of salts, here in the format. We
- * return a pointer ot an allocated salt record. Thus, we search all
+ * return a pointer to an allocated salt record. Thus, we search all
  * of the salt records, looking for the same salt.  If we find it, we
  * want to return THAT pointer, and not allocate a new pointer.
  * This works great, but forces us to do salt comparision here.
--- a/src/dynamic_fmt.c
+++ b/src/dynamic_fmt.c
@@ -587,6 +587,43 @@ static int valid(char *ciphertext, struct fmt_main *pFmt)
 	}
 	else if (!regen_salts_options && pPriv->dynamic_FIXED_SALT_SIZE < -1 && strlen(&ciphertext[pPriv->dynamic_SALT_OFFSET]) > -(pPriv->dynamic_FIXED_SALT_SIZE)) {
 		char *cpX;
+// NOTE if looking at this in the future, that was added by Aleksey for #5032
+		if (!pPriv->b2Salts && !pPriv->nUserName && !pPriv->FldMask) {
+			/* salt-only, strict mode: salt without $HEX$ or $HEX$hex till the end */
+			if (!memcmp(&ciphertext[pPriv->dynamic_SALT_OFFSET], "HEX$", 4)) {
+				for (i = pPriv->dynamic_SALT_OFFSET + 4; ciphertext[i]; i++)
+					if (atoi16[ARCH_INDEX(ciphertext[i])] == 0x7f)
+						return 0; // not hex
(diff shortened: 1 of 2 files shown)


AlekseyCherepanov:
I added extra strict code. It duplicates existing checks partially. Conveniently it is localized to formats with salts with variable length.

While my code rejects more than I would like, it accepts all self-tests without adjustments. `--format=dynamic --test=0` and `--format=disabled --test=0` are ok.

The following comment made me think @jfoug might want fixes to be emphasized and localized:
```c
// NOTE if looking at this in the future, this was not my fix.
```

I did not find any existing format using fields (`$$F`). There are formats with salt-2 (`$$2`) and usernames (`$$U`), but they seem to use flat buffers and do not suffer from overflows.

Maybe I should investigate, how flat buffers handle overly long intermediate input, and fix `__SSE_append_string_to_input` to behave similarly. OTOH it might affect performance.

Besides from salt-only formats, the following formats have the problem and `SaltLen` correct for both salt and username. So I use its value to check both lengths:
```
dynamic_1015 [md5(md5($p.$u).$s) (PostgreSQL 'pass the hash') 128/128 SSE4.1 4x3]
dynamic_1034 [md5($p.$u) (PostgreSQL MD5) 128/128 SSE4.1 4x3]
dynamic_1506 [md5($u.:XDB:.$p) (Oracle 12c "H" hash) 128/128 SSE4.1 4x3]
dynamic_1550 [md5($u.:mongo:.$p) (MONGODB-CR system hash) 128/128 SSE4.1 4x3]
```

This format has fixed length of salt and no info about allowed length of username:
```
dynamic_1551 [md5($s.$u.(md5($u.:mongo:.$p)) (MONGODB-CR network hash) 128/128 SSE4.1 4x3]
```

I guess `dynamic_1551` should be disabled (temporarily?) in favor of slower `dynamic_1552` without overflows. But I don't know how.

Should I disable `dynamic_1551` commenting it out completely? Or should I mark it as disabled so it could be enabled with `--format=`? The second variant seem weak: it is easy to enable a disabled format inadvertently (e.g. with `--format=-asdf`).

Also I might add commits with fixes of typos in comments and documentation into this PR.

solardiz:
@AlekseyCherepanov I have no time to dive into this. If you need someone's input before we merge this, then let's see if anyone else hopefully comments. If this is ready to merge as-is (and your questions here are to be addressed or not separately), let me know. Thanks!

solardiz:
@AlekseyCherepanov What's the current status of this PR?

AlekseyCherepanov:
Now `dynamic_1551` is the only predefined dynamic crashing.

The only question: should I fix it?

I could disable it or add very specific code into `valid` just for this particular format until there would be a better solution.

AlekseyCherepanov:
I added checks of length in the crashing appending functions. So `dynamic_1551` does not crash, instead it works incorrectly for too long usernames.

`--format=dynamic,disabled --test=0` is ok. I did not test speeds though.

The fix seems good enough to fix crash. But I dislike it: it should be rejected in `valid`. Also I would say that skipping appending could produce incorrect crack, but it seems that structure of `dynamic_1551` does not allow it.

AlekseyCherepanov:
Fixed typo in commit message.

solardiz:
OK, is this ready to merge (once CI is happy again)?

AlekseyCherepanov:
Added insufficient fixes to `dynamic_fmt.c` and `dynamic_preloads.c`.

This PR is ready.

AlekseyCherepanov:
> (16 is max for safe `$u` in the first concatenation)

That's incorrect. Max is 25. Also there is a comment in `dynamic.conf`:
```
MaxInputLen=23
MaxInputLenX86=110
# note, saltlen + length(:mongo:) + length(plain) must stay <= 55 for SIMD
#   so 25+7+23 == 55
```

PR #5051, opened 2022-02-05
7z formats: fixed overflow in warning in valid()

AlekseyCherepanov:
Related to #4971.

### DIFF
--- a/src/7z_common_plug.c
+++ b/src/7z_common_plug.c
@@ -112,7 +112,7 @@ int sevenzip_valid(char *ciphertext, struct fmt_main *self)
 {
 	char *ctcopy, *keeptr, *p;
 	int type, c_type, p_type, len, NumCyclesPower;
-	static char warned[128];
+	static char warned[256];
 
 	if (strncmp(ciphertext, FORMAT_TAG, TAG_LENGTH) != 0)
 		return 0;
@@ -127,7 +127,7 @@ int sevenzip_valid(char *ciphertext, struct fmt_main *self)
 	type = atoi(p);
 	c_type = type & 0xf;
 	p_type = (type >> 4) & 0xf;
-	if (strlen(p) == 0 || type < 0 || !precomp_type[p_type] || !comp_type[c_type]) /* Codec(s) needed for CRC check */
+	if (strlen(p) == 0 || type < 0 || type >= 256 || !precomp_type[p_type] || !comp_type[c_type]) /* Codec(s) needed for CRC check */
 		goto err;
 	if (c_type > 2
 #if HAVE_LIBBZ2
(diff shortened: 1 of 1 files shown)


AlekseyCherepanov:
`7z,7z-opencl`: `--test`, `--fuzz`, `--fuzz=run/fuzz.dic` are ok.

I rechecked formats I fixed for #4971 with `--fuzz=run/fuzz.dic`. It found another crash in 7z.

`c_type` and `p_type` use 4 bit from `type`. So I assume that `type` should not be greater than 256.

Also I replaced `++` for warnings with assignments on separate lines. The type is `char`, so it can overflow during realistic usage.

AlekseyCherepanov:
> So I assume that `type` should not be greater than 256.

I mean greater or equal. It should not be 256 too.

AlekseyCherepanov:
There is similar `++` for type `int` in the same file:
```c
	if (len > 0 && strstr(self->params.label, "-opencl")) {
		static int warned;

		if (!warned++)
```

Such trick seems quite common:
```
$ grep -lF '!warned++' src/*.c
src/7z_common_plug.c
src/encdatavault_fmt_plug.c
src/gpu_common.c
src/krb5pa-sha1_fmt_plug.c
src/loader.c
src/logger.c
src/mask.c
src/mscash_common_plug.c
src/oldoffice_common_plug.c
src/opencl_diskcryptor_aes_fmt_plug.c
src/opencl_diskcryptor_fmt_plug.c
src/opencl_krb5pa-sha1_fmt_plug.c
src/opencl_lm_b_plug.c
src/opencl_ntlmv2_fmt_plug.c
src/opencl_oldoffice_fmt_plug.c
src/pp.c
src/rar_common.c
src/rpp.c
src/subsets.c
src/telegram_common_plug.c
src/wordlist.c
```

In theory it can overflow. It would be undefined behaviour. Is it a problem?

AlekseyCherepanov:
This PR is ready. Thanks!

solardiz:
> $ grep -lF '!warned++' src/*.c

> In theory it can overflow. It would be undefined behaviour. Is it a problem?

Yes. Please fix those like you did here - to explicitly assign `1`. Otherwise we need to upgrade the type to `uint64_t`.

solardiz:
> explicitly assign `1`.

BTW, that's what I do in the few places in core, so without a real reason to do differently it'd be most consistent for us to continue with the same approach in jumbo.

```
$ fgrep -r warned .
./formats.c:	int binary_align_warned = 0, salt_align_warned = 0;
./formats.c:		    !binary_align_warned) {
./formats.c:			binary_align_warned = 1;
./formats.c:		    !salt_align_warned) {
./formats.c:			salt_align_warned = 1;
./formats.h:/* We've already warned the user about hashes of this type being present */
./c3_fmt.c:	static int warned = 0;
./c3_fmt.c:#pragma omp parallel for default(none) private(index) shared(warned, count, crypt_out, saved_key, saved_salt, crypt_data, stderr)
./c3_fmt.c:			if (!warned) {
./c3_fmt.c:				warned = 1;
./c3_fmt.c:#pragma omp parallel for default(none) private(index) shared(warned, count, crypt_out, saved_key, saved_salt, stderr)
./c3_fmt.c:			if (!warned) {
./c3_fmt.c:				warned = 1;
```

AlekseyCherepanov:
Hm, maybe I should give a warning about type `>= 256` like it is done for unsupported types.

Issue #5035, opened 2022-01-31
Uses of malloc()'ed pointers without NULL check


solardiz:
As reported in #5028 and https://huntr.dev/bounties/315d2d54-6ae4-47a1-a6ff-84fea6cb7d01/ by @michaellrowley with assistance of @JamieSlome and @huntr-helper (how many engineers does it take to get a one-liner fix in, and how ridiculous can a bug disclosure process become):

```
NULL Pointer Dereference in openwall/john
Unverified

Reported on

Jan 28th 2022
Description

John The Ripper makes an allocation using malloc() that isn't validated, meaning that - in instances where malloc() returns a null pointer; the null pointer will be dereferenced, typically causing a crash or other undefined behaviour (a system-specific response that is usually something along the lines of a SIGSEGV)
Impact

This vulnerability is capable of allowing attackers to trigger a denial-of-service (or other undefined behaviour) in the event of John The Ripper being unable to access memory (or any other condition where malloc() might fail.)
Occurences

racf_kdfaes_fmt_plug.c L334

This is the original allocation:

unsigned char *t1f = malloc(HASH_OUTPUT_SIZE * cur_salt->mfact);

racf_kdfaes_fmt_plug.c L378

memcpy(t1f + (n*HASH_OUTPUT_SIZE), h_out, HASH_OUTPUT_SIZE);

racf_kdfaes_fmt_plug.c L375

memcpy(m, t1f + (n_key * HASH_OUTPUT_SIZE), HASH_OUTPUT_SIZE);

racf_kdfaes_fmt_plug.c L382-L384

memcpy(t1f + (HASH_OUTPUT_SIZE * (cur_salt->mfact-1)), "\x00\x00\x00\x01", 4);
m1 = (HASH_OUTPUT_SIZE * (cur_salt->mfact-1))+4;
JTR_hmac_sha256(key_p, HASH_OUTPUT_SIZE, t1f, ml, h_out, HASH_OUTPUT_SIZE);

racf_kdfaes_fmt_plug.c L365

memcpy(t1f+(n*HASH_OUTPUT_SIZE), t1, HASH_OUTPUT_SIZE);
```

Of course, this is not a vulnerability, but nevertheless a bug for us to fix. I think proper fix is to replace `malloc` with our `mem_alloc`.

[...]

PR #5054, opened 2022-02-07
Fix uses of malloc()'ed pointers without NULL check (issue #5035)

Shashank-Rangu:
**Issue #5035 Uses of malloc()'ed pointers without NULL check** 

Changes: 
Replaced malloc() with mem_alloc() wrapper, calloc() with mem_calloc() wrapper, and a new wrapper xstrdup() is created for strdup() and is used in the following:

- In files which are common or which are compiled to john
- In *2john files, if they already include "memory.h" and are already linking to memory.o
- Remaining *2john files are ignored for now

Further file by file comments are provided in the comments to the individual commits


src/vnc_fmt_plug.c
```diff
@@ -123,7 +123,7 @@ static struct fmt_tests vnc_tests[] = {
 	{"$vnc$*0805B790B58E967F2A350A0C99DE3881*AECB26FAEAAA62D79636A5934BAC1078", "Password"},
 	{"$vnc$*ADDC021F444F999B8E27144C0DCE7389*AFAF1BB57588784333962A124668A2C6", "openwall"},
 	{"$vnc$*1D03C57F2DFFCC72A5AE3AD559C9C3DB*547B7A6F36A154DB03A2575C6F2A4EC5", "openwall"},
-	{"$vnc$*84076F040550EEA9341967633B5F3855*807575689582379F7D807F736DE9E434", "pass\xc2\xA3"}, // high bit set is 'fun' for DES. c2a3 is � in utf8 bytes.
+	{"$vnc$*84076F040550EEA9341967633B5F3855*807575689582379F7D807F736DE9E434", "pass\xc2\xA3"}, // high bit set is 'fun' for DES. c2a3 is � in utf8 bytes.
```
magnumripper: This change is not only unrelated, but also incorrect.

magnumripper (changes requested):
I find it a bit weird that you put the function in `pp.c` and declare it in both `jumbo.h` and `memory.h`. I think you should put in in `memory.c` and declare it only in `memory.h` (sourcing that header if/where needed).

Also, the build bots currently fail to build some targets, not sure if you can see that or if only we can?

src/vnc_fmt_plug.c
Shashank-Rangu: > I find it a bit weird that you put the function in `pp.c` and declare it in both `jumbo.h` and `memory.h`. I think you should put in in `memory.c` and declare it only in `memory.h` (sourcing that header if/where needed).
> 
> Also, the build bots currently fail to build some targets, not sure if you can see that or if only we can?

I will change the location of the function. I thought that the function needs to be put in will other memory wrapper functions like mem_alloc() and malloc_tiny(). 

As per the build targets, I had to create the pull request in order to see them. Is there any other way I can run them without creating a pull request? 

The change in vnc_fmt_plug.c:126 was unintentional, my editor was not able to identify the symbol and that got copied. I will revert it.

magnumripper: > I will change the location of the function. I thought that the function needs to be put in will other memory wrapper functions like mem_alloc() and malloc_tiny().

Oh, I see. They are actually in `memory.c` - you were mislead by the fact some of them are duplicated in `pp.c` for  use outside of this project. Sorry about that confusion.

> As per the build targets, I had to create the pull request in order to see them. Is there any other way I can run them without creating a pull request?

As you push (or force push) to this branch, they will re-trigger.

src/memory.c
```diff
@@ -129,6 +129,20 @@ void *mem_realloc(void *old_ptr, size_t size)
...
+char *xstrdup (const char* str)
+{
+  char *res = strdup(str);
+
+  if (res == NULL)
+  {
+    fprintf (stderr, "strdup: %s\n", strerror (ENOMEM));
+
+    exit (-1);
+  }
```
magnumripper: OK so now I think we only have some coding style nitpicks left to deal with.

First, the bracket att line 137 should be in the end of line 136 (with a space before it of course).
Second, there should be no space between `fprintf` and the left paren after it.

And finally, please squash your fixes so there's only a single commit for "Added xstrdup() wrapper and replaced accordingly" and force-push it here.

solardiz: I think @Shashank-Rangu was following the coding style of `pp.c`, which isn't our project's. So `xstrdup` should be rewritten using the style of a nearby function in `memory.c` now - this includes not only indentation, but also e.g. use of `error();` in place of `exit (-1);`.

> please squash your fixes so there's only a single commit for "Added xstrdup() wrapper and replaced accordingly" and force-push it here.

Right. There should be 3 commits here (like this PR originally had), not 4.

Shashank-Rangu: > OK so now I think we only have some coding style nitpicks left to deal with.
> 
> First, the bracket att line 137 should be in the end of line 136 (with a space before it of course). Second, there should be no space between `fprintf` and the left paren after it.
> 
> And finally, please squash your fixes so there's only a single commit for "Added xstrdup() wrapper and replaced accordingly" and force-push it here.

Done. Please check if everything is up to the mark.

solardiz: You tried to edit `xstrdup` still based on what it was in `pp.c` - I suggest you drop it and rewrite it based on a nearby function in `memory.c`. There are many minor style differences.

solardiz (approved):
You still didn't go for that `xstrdup` rewrite and thus still have minor style differences from its surrounding code, but we can live with that. This is pretty dirty code anyway, so little point in polishing one function.

src/memory.h
```diff
@@ -126,6 +133,7 @@ extern char *strdup_MSVC(const char *str);
 #define free(a) __mingw_aligned_free(a)
 #define strdup(a) strdup_MSVC(a)
 extern char *strdup_MSVC(const char *str);
+extern char *xstrdup(const char *str);
```
solardiz: Unnecessary duplicate prototype.

solardiz (changes requested):
Reviewing the individual commits, I see that the vnc file's character corruption first gets introduced with one commit, then fixed with another. We can't accept that - it should never be introduced in the first place.

solardiz:
> I see that the vnc file's character corruption first gets introduced with one commit, then fixed with another.

Oh, that's not the case. I didn't expect to see this change at all, but I still do see a change like it - yet in just one commit.

What we have in the original source file is one byte value A3, which isn't valid utf-8. The commit include a fix of that into valid utf-8. Maybe that's even a correct fix. However, it has nothing to do with that specific commit, and shouldn't be in there.

Shashank-Rangu:
> > I see that the vnc file's character corruption first gets introduced with one commit, then fixed with another.
> 
> Oh, that's not the case. I didn't expect to see this change at all, but I still do see a change like it - yet in just one commit.
> 
> What we have in the original source file is one byte value A3, which isn't valid utf-8. The commit include a fix of that into valid utf-8. Maybe that's even a correct fix. However, it has nothing to do with that specific commit, and shouldn't be in there.

Removed that portion of the commit and force pushed it again

magnumripper (approved):
Looks good (enough) to me now

solardiz (approved):

magnumripper:
Thank you @Shashank-Rangu for your contribution! We're currently short on contributors.

PR #5059, opened 2022-02-09
Fix more uses of malloc()'ed pointers without NULL check (issue #5035)

### LINKED ISSUE #5035: Uses of malloc()'ed pointers without NULL check

solardiz:
This is untested yet. It should fix #5035.

I think the only remaining place would be in MSVC-specific `vc_fixed_snprintf` - I'll just let that one crash if it's ever compiled in.

### DIFF
--- a/src/bks_fmt_plug.c
+++ b/src/bks_fmt_plug.c
@@ -116,7 +116,7 @@ static int valid(char *ciphertext, struct fmt_main *self)
 
 	if (strncmp(ciphertext, FORMAT_TAG, FORMAT_TAG_LENGTH) != 0)
 		return 0;
-	ctcopy = strdup(ciphertext);
+	ctcopy = xstrdup(ciphertext);
 	keeptr = ctcopy;
 	ctcopy += FORMAT_TAG_LENGTH;
 	if ((p = strtokm(ctcopy, "$")) == NULL) // format
@@ -189,7 +189,7 @@ static void *get_salt(char *ciphertext)
 	char *p = ciphertext, *ctcopy, *keeptr;
 	memset(&cs, 0, sizeof(cs));
 
-	ctcopy = strdup(ciphertext);
+	ctcopy = xstrdup(ciphertext);
 	keeptr = ctcopy;
 	ctcopy += FORMAT_TAG_LENGTH;
 	p = strtokm(ctcopy, "$");
--- a/src/diskcryptor_common_plug.c
+++ b/src/diskcryptor_common_plug.c
@@ -6,10 +6,11 @@
  */
 
 #include "arch.h"
-#include "diskcryptor_common.h"
+#include "memory.h"
 #include "jumbo.h"
 #include "johnswap.h"
 #include "xts.h"
+#include "diskcryptor_common.h"
 
 // #define DEBUG 1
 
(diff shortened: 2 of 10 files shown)


solardiz:
`tgtsnarf.c` is weird in that it includes `memory.h` and uses its `MEM_FREE` macro, yet does not link against `memory.o`.

magnumripper (approved):
Not that you are ever a suspect for careful review, but I looked it through

solardiz:
With my changes here, `mem_calloc_align` can not longer be called with `size` of zero - it'd divide by zero. I don't know if we want to fix that (perhaps by returning `NULL` for such zero-sized allocations, like we already do in some other functions there), but I thought I'd mention.

magnumripper:
I can't see that being a problem, and if it is we'll probably get aware of more problems along with it

Issue #5048, opened 2022-02-04
NT-opencl performance regression on dual socket rig after the "avoid nvidia busy-wait"


magnumripper:
I had a report of significant regression, from ~25 Gp/s to more like 17 Gp/s per GPU, running incremental plus `-mask:?w?a?a` under MPI on a GPU rig with 8x 2080ti (and more CPU cores than GPU cards). CPU utilization went from ~86% to ~61%.

I was granted access to the rig enough to reproduce, and to confirm that only reverting 7f6aed9de fixes the problem completely.

Not sure when/how much I can experiment more with that very rig. My single-GPU tests show nothing of the sort so I'm puzzled as to why it would happen with more processes and GPUs. Perhaps 3x fork would show it, so some tests on "super" (with the nvidias not simultaneously used for more things) is needed.

Meanwhile perhaps we should consider reverting that commit?

solardiz:
@magnumripper Can you test on that rig whether the regression is in fact specific to this format, or would also be seen with another format we treated similarly? Reverting for just one format isn't such a good idea if the problem isn't format-specific.

magnumripper:
I believe it's only NT (and that format is kinda special with its high speed and short durations). I'll see if I can get more data.

solardiz:
Maybe we need to actually increase the 1ms threshold? Also, the 2.4% and 1.2% figures become rather low in absolute terms when we're near that threshold. While we could make code changes to make sure these are no lower than some numbers of microseconds, simply increasing the sleep threshold e.g. from 1ms to 10ms would avoid using ridiculously low figures there as well (they would still be computed, just never used).

What's the kernel duration for `nt-opencl` on 2080Ti with that mask and I guess certain autotuned LWS/GWS?

solardiz:
Maybe this is both a good approach and a trivial enough change:

```diff
+++ b/src/opencl_common.h
@@ -342,8 +342,8 @@ void opencl_process_event(void);
        if (gpu_nvidia(device_info[gpu_id])) { \
                wait_start = john_get_nano(); \
                uint64_t us = wait_min >> 10; /* 2.4% less than min */ \
-               if (wait_sleep && us >= 1000) \
-                       usleep(us); \
+               if (wait_sleep && us >= 2000) \
+                       usleep(us - 1000); \
        }
 #define WAIT_UPDATE \
        if (gpu_nvidia(device_info[gpu_id])) { \
```

[...]

PR #5061, opened 2022-02-10
Add a config setting for disabling the busy-wait workaround for nvidia

magnumripper:
See #5048

### DIFF
--- a/run/john.conf
+++ b/run/john.conf
@@ -418,6 +418,10 @@ AbortTemperature = 95
 # interpret as intent to keep the GPU temperature around the limit.
 SleepOnTemperature = 1
 
+# Enable a workaround for busy-waits, introducing calls to usleep(3).  This
+# currently only applies to nvidia.
+AvoidBusyWait = Y
+
 [Options:OpenCL]
 # Set default OpenCL device(s). Command line option will override this.
 # If not set, we will search for a GPU or fall-back to the most
--- a/src/opencl_common.c
+++ b/src/opencl_common.c
@@ -80,6 +80,7 @@ size_t ocl_max_lws;
 static char opencl_log[LOG_SIZE];
 static int opencl_initialized;
 int opencl_unavailable;
+int opencl_avoid_busy_wait[MAX_GPU_DEVICES];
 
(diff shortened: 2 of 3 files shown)


solardiz (approved):
Looks good to me, but I think we should drop "Nvidia" from the setting name, so that it's reusable without a rename later in case that is ever needed. Keep this in the comment.

Also, can we reasonably move the ` && gpu_nvidia(device_info[gpu_id])` checks to be done just once at initialization time? We could do `opencl_avoid_busy_wait &&= ...` if we're running post-fork (so that it's a separate variable per GPU). OTOH, that would further tie us to the current model of a process per GPU.

magnumripper:
I pushed a new version - it needed to be put in `opencl_prepare_dev()` instead because some formats use that directly. I also added logging.

I tried it using two devices (GPU and CPU) with `--fork` and it ends up correct. With mscash2-opencl, the logging looks OK but it obviously ends up with the macros enabled even for the CPU device. Not sure how to remedy that other than to put the checks back in the macros.

magnumripper:
I could make it `opencl_avoid_busy_wait[gpu_id]` of course. I'll look into that.

magnumripper:
Done. This should be fine now.

magnumripper:
>  We could do `opencl_avoid_busy_wait &&= ...` if we're running post-fork (so that it's a separate variable per GPU).

I'm not quite sure what you meant there so I'd need a review again.

solardiz (approved):

solardiz:
This mostly looks good to me, except:

> With mscash2-opencl, the logging looks OK but it obviously ends up with the macros enabled even for the CPU device.

We don't even use the macros in that format, and I think we should not until we rewrite it.

I think this actually points out another issue: you do the logging regardless of whether the format uses the macros or not. This can be confusing. Maybe omit the logging until we start using the macros (almost) universally (maybe with very few exceptions that we'd check for)?

magnumripper:
Changed it to only mention "disabled" for now.

Before we merge, I'll investigate one more thing: IIRC the macros aren't needed on macOS. I just need to find my ten year old nvidia macbook...

PR #5066, opened 2022-02-12
Busy-wait with callback

magnumripper:
Currently only implemented in nt-opencl, will try tezos next.

### DIFF
--- a/src/opencl_common.h
+++ b/src/opencl_common.h
@@ -131,6 +131,8 @@ extern struct db_main *ocl_autotune_db;
 extern int autotune_real_db;
 extern int opencl_unavailable;
 extern int opencl_avoid_busy_wait[MAX_GPU_DEVICES];
+extern int opencl_sleep_pid;
+extern cl_event opencl_sleep_event;
 
 #define ocl_any_test_running	(bench_or_test_running || ocl_autotune_running)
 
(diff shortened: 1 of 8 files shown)


solardiz:
If this is successful, are we going to drop our previous usage of `SIGUSR2` and reuse it like this? Or use another signal? Or move away from signals for this (to other `select` events)?

solardiz:
Also, was `clSetEventCallback` introduced long enough ago and universally enough (across OpenCL implementations) that we can use it unconditionally (a compile-time dependency at least)?

magnumripper:
It's OpenCL 1.2 so should be fine. I'm not sure about repurposing USR2 or find some other signal, first let's see if this is usable at all.

src/opencl_common.h
```diff
@@ -337,6 +337,22 @@ void opencl_process_event(void);
...
+	if (opencl_avoid_busy_wait[gpu_id] && !sleep_event) { \
+		opencl_sleep_pid = getpid(); \
+		sleep_event = &opencl_sleep_event; \
+	}
+
+#define WAIT2_SLEEP(sleep_event)	  \
+	if (sleep_event == &opencl_sleep_event) { \
+		BENCH_CLERROR(clSetEventCallback(opencl_sleep_event, CL_COMPLETE, &opencl_event_callback, NULL), "set callback"); \
+		opencl_sleep(); \
+	}
```
magnumripper: Note to self: Need to drop BENCH_CLERROR here, and instead only sleep if `clSetEventCallback` succeeds.

magnumripper:
I'm giving up on this for now.

Issue #5068, opened 2022-02-12
TrueCrypt formats: ciphertexts contain filenames


AlekseyCherepanov:
TrueCrypt formats (`TrueCrypt-opencl`, `tc_aes_xts`, `tc_ripemd160`, `tc_ripemd160boot`, `tc_sha512-out`, `tc_whirlpool`) support keyfiles. They are referenced by filenames from ciphertexts.

I see that zip or pkzip format was changed to avoid files and to embed files into hash. I guess similar change might be needed here.

In practice, I have problem that `--fuzz` exits because `get_salt` cannot open file with synthetic name. Sample:
```
$ python -c 'print "truecrypt_SHA_512$" + "aa" * 512 + "$1$2$3"' > t.pw
$ ./run/john t.pw
Warning: detected hash type "tc_aes_xts", but the string is also recognized as "tc_sha512"
Use the "--format=tc_sha512" option to force loading these as that type instead
fopen 2$3: No such file or directory
```

`strace` suggests that actual name is `2`. So name is reported wrongly.

Respective code:
```c
		fp = fopen(tpath, "rb");
		if (!fp)
			pexit("fopen %s", p);
```

This part of code is same in `truecrypt_fmt_plug.c` and in `opencl_tc_fmt_plug.c`.

Also I guess that `jtr_fopen` macro should be used instead of regular `fopen`.

(Backreference to #5067, where I mentioned it.)

AlekseyCherepanov:
`1$2$3` suffix is the number of filenames + 2 filenames, but the second filename is not used due to provided number lower than actual number of filenames. So `valid` could check that.

Changing syntax of ciphertext to embed files number of keyfiles might be removed.

solardiz:
Aleksey, I think we should fix actual bugs (and you're working on just that - thanks!), but keep the files external (not embed them in the "hashes").

In fact, I think we should (later, perhaps after our next release) rework the rest of `*2john` and formats to never embed large files in "hashes", but reference them externally. It's just ridiculous and confusing and inconvenient and wasteful to have huge "hashes". We have an open issue to have the tools tell the users this is OK - to hopefully reduce the confusion - but a better fix would be for us to stop being ridiculous in this respect.

AlekseyCherepanov:
Theoretical advantage of embedding is that it would be possible to open all input files (and .pot files), prohibit most syscalls through seccomp and then parse input.

solardiz:
[...]

PR #5069, opened 2022-02-12
TrueCrypt formats: fixed overflows in handling of keyfiles

AlekseyCherepanov:
- fixed the minor problem about wrong name from #5068
- added check of file sizes in cpu and opencl formats in `get_salt`
- opencl format allocated memory in wrong function, so keyfiles did not work; fixed
- opencl format used strings without null bytes for candidates, so it crashed in `crypt_all` after a few seconds; so I changed 2 calls of `apply_keyfiles` providing length

`--format=TrueCrypt-opencl,tc_aes_xts,tc_ripemd160,tc_ripemd160boot,tc_sha512,tc_whirlpool --test` is ok.

`--fuzz` does not work due to filenames (#5068).

PR is ready.

### DIFF
--- a/src/opencl_tc_fmt_plug.c
+++ b/src/opencl_tc_fmt_plug.c
@@ -153,9 +153,6 @@ static void create_clobj(size_t gws, struct fmt_main *self)
 		&mem_out), "Error while setting mem_out kernel argument");
 	HANDLE_CLERROR(clSetKernelArg(crypt_kernel, 2, sizeof(mem_setting),
 		&mem_setting), "Error while setting mem_salt kernel argument");
-
-	keyfiles_data = mem_calloc(MAX_KEYFILES, sizeof(*keyfiles_data));
-	keyfiles_length = mem_calloc(MAX_KEYFILES, sizeof(int));
 }
 
 static void release_clobj(void)
@@ -167,13 +164,13 @@ static void release_clobj(void)
 
 		MEM_FREE(inbuffer);
 		MEM_FREE(outbuffer);
-		MEM_FREE(keyfiles_data);
-		MEM_FREE(keyfiles_length);
 	}
 }
 
 static void done(void)
 {
+	MEM_FREE(keyfiles_data);
+	MEM_FREE(keyfiles_length);
 	if (program[gpu_id]) {
 		release_clobj();
 
@@ -188,6 +185,9 @@ static void init(struct fmt_main *_self)
 {
 	self = _self;
 	opencl_prepare_dev(gpu_id);
+
+	keyfiles_data = mem_calloc(MAX_KEYFILES, sizeof(*keyfiles_data));
+	keyfiles_length = mem_calloc(MAX_KEYFILES, sizeof(int));
 }
 
 static void reset(struct db_main *db)
@@ -311,12 +311,18 @@ static void* get_salt(char *ciphertext)
 		/* read this into keyfiles_data[idx] */
 		fp = fopen(tpath, "rb");
 		if (!fp)
-			pexit("fopen %s", p);
+			pexit("fopen %s", tpath);
 
 		if (fseek(fp, 0L, SEEK_END) == -1)
 			pexit("fseek");
 
 		sz = ftell(fp);
+
+		if (sz > MAX_KFILE_SZ) {
+			if (john_main_process)
+				fprintf(stderr, "Error: keyfile '%s' is bigger than maximum size (MAX_KFILE_SZ is %d).\n", tpath, MAX_KFILE_SZ);
+			error();
+		}
 
 		if (fseek(fp, 0L, SEEK_SET) == -1)
 			pexit("fseek");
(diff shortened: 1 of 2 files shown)


AlekseyCherepanov:
Sample of crasher:
```
$ python -c 'print "truecrypt_SHA_512$" + "aa" * 512 + "$1$2$3"' > t.pw
$ python -c 'print "a" * (1048576 * 256)' > 2
```

Buffer's size is 256 Mb (256 files, 1 Mb each), so additional `\n` added by python's `print` causes crash.
```
==25638==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x7fd707018800 at pc 0x7fd70b2af800 bp 0x7ffd9d2cd5c0 sp 0x7ffd9d2ccd70
WRITE of size 268435457 at 0x7fd707018800 thread T0
    #0 0x7fd70b2af7ff in __interceptor_fread ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:936
    #1 0x55e7ec7e8313 in get_salt /home/user/john/src/truecrypt_fmt_plug.c:340
```

solardiz:
I think there may be more similarly bad issues to fix here. In this snippet:

```c
		if (!q) { // last file
			memset(tpath, 0, sizeof(tpath) - 1);
			strncpy(tpath, p, sizeof(tpath));
		} else {
			memset(tpath, 0, sizeof(tpath) - 1);
			strncpy(tpath, p, q-p);
		}
```

Why do we need the `memset` and why of 1 byte less? Maybe the `- 1` was meant to be on the first `strncpy`? Even then, truncating a filename isn't appropriate - does `valid` somehow make this impossible (I don't see that). Similarly, is `q-p` somehow guaranteed to fit (and leave room for NUL, which this code forgets to add?) Why does the code even use the NUL-padding `strncpy` rather than something else?

Also, `valid` calls `atoi` without a preceding `isdec` check.

AlekseyCherepanov:
JFYI using `strnlen` in `apply_keyfiles`, it is possible to have same code for `apply_keyfiles` in both files. Then `set_key` in opencl format would need saving of null byte when it fits into buffer. But it would look more like a hack.

AlekseyCherepanov:
@solardiz I'll check these problems tomorrow. So PR is not ready.

AlekseyCherepanov:
I did not switch from `fopen` to `jtr_fopen` because I was not sure about subtle consequences.

AlekseyCherepanov:
Yes, `strncpy` can drop null byte, then `fopen` would read the name too far:
```
$ python -c 'print "truecrypt_RIPEMD_160$" + "aa" * 512 + "$1$" + "b" * 4096' > t.pw
$ ./run/john t.pw
[...]
==15627==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7fff84a68cf0 at pc 0x7fc58229a014 bp 0x7fff84a67c90 sp 0x7fff84a67440
READ of size 4097 at 0x7fff84a68cf0 thread T0
    #0 0x7fc58229a013 in __interceptor_fopen ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:5377
    #1 0x560bd9198365 in get_salt /home/user/john/src/truecrypt_fmt_plug.c:329
    #2 0x560bd920f3e8 in ldr_load_pw_line /home/user/john/src/loader.c:1045
[...]
```

`PATH_BUFFER_SIZE` seems to be 4096 as `PATH_MAX` in my build.

BTW I heard that `atoi` with input that overflows integer is undefined behaviour, i.e. `atoi` cannot be used on arbitrary input.

AlekseyCherepanov:
> BTW I heard that atoi with input that overflows integer is undefined behaviour, i.e. atoi cannot be used on arbitrary input.

`isdec` seems to protect from that.

AlekseyCherepanov:
I added `isdec`, checks that given number matches actual list of filenames, checks for lengths of filenames. PR is ready. Thanks!

solardiz:
@AlekseyCherepanov I think the `fprintf` would make more sense in `valid` than in an impossible case in `get_salt`. I understand you put them in the latter because `valid` allows you to reject a hash, but in `get_salt` it's too late for that. However, I think it's more appropriate to use `assert` for impossible cases like this - also saves you from having to invent a user-suitable message for something that, if it ever happens, would be in part because of our bug.

AlekseyCherepanov:
I mimicked other `fprintf` from `truecrypt_fmt_plug.c`:
```c
	} else {
		// should never get here!  valid() should catch all lines that do not have the tags.
		fprintf(stderr, "Error, unknown type in truecrypt::get_salt(), [%s]\n", ciphertext);
		error();
	}
```

I had idea that `valid` would be a better place for warning. But on systems with `PATH_MAX` it seems unlikely that we could ever see the warning for original ciphertexts. So I considered that silent rejection would be good enough.

OTOH there are systems without such limits. GNU Hurd is well-known for that and has [notes about elimination of the limit](https://www.gnu.org/software/hurd//community/gsoc/project_ideas/maxpath.html).

`params.h`:
```c
/*
 * Buffer size for path names.
 */
#ifdef PATH_MAX
#define PATH_BUFFER_SIZE		PATH_MAX
#else
#define PATH_BUFFER_SIZE		0x400
#endif
```

`/usr/include/linux/limits.h`:
```c
#define PATH_MAX        4096	/* # chars in a path name including nul */
```

Should I add the warning? Should I switch onto `assert`?

solardiz:
> Should I add the warning? Should I switch onto `assert`?

Oh, I guess let's not bother now. I'll merge as-is. Thanks.

PR #5071, opened 2022-02-14
Update to support new key derivation of PrivateAccess and ENCDatavault

sylvainpelissier:
Support of PBKDF2-SHA256 key derivation function. Test have been commented because they use 100000 iterations. Fake hashes have been generated.

### DIFF
--- a/doc/README-ENCDataVault.md
+++ b/doc/README-ENCDataVault.md
@@ -7,6 +7,9 @@ Cracking ENCSecurity DataVault
 $ ../run/encdatavault2john.py path/to/vault > hash
 ```
 
+For Sandisk and Sony the script must be run in the "Settings" folder, for example, 
+"SanDiskSecureAccess Settings". 
+
 2. Run john on the file.
 
 ```bash
@@ -17,10 +20,12 @@ The decryption test is done on a 32-bit magic value, meaning that occasional
 false positives can occur. Cracked passwords for those false positives will not
 decrypt the vault correctly.
 
-The format has been tested on Sony ENCDataVault lite 6.2.13, SanDisk 
-SecureAccess 3.02 and ENC DataVault 7.1.1W.
+The format using the key derivation algorithm based on MD5 has been tested on 
+Sony ENCDataVault lite 6.2.13, SanDisk SecureAccess 3.02 and ENC DataVault 
+7.1.1W. The format using PBKDF-SHA256 has been tested on PrivateAccess 6.3.6W 
+and ENCDataVault 7.2.1.
 
-The key derivation problems were patched by Western Digital: 
+The first key derivation based on MD5 was replaced by Western Digital: 
 https://www.westerndigital.com/en-ap/support/product-security/wdc-21014-sandisk-secureaccess-software-update. 
 Their new solution is now called PrivateAccess: 
 https://kb.sandisk.com/app/answers/detail/a_id/21996/
--- a/run/encdatavault2john.py
+++ b/run/encdatavault2john.py
@@ -2,7 +2,7 @@
 
 # Helper script for cracking ENCSecurity DataVault.
 #
-# This software is Copyright (c) 2021, Sylvain Pelissier <sylvain.pelissier at kudelskisecurity.com>
+# This software is Copyright (c) 2021-2022, Sylvain Pelissier <sylvain.pelissier at kudelskisecurity.com>
 # and it is hereby released to the general public under the following terms:
 #
 # Redistribution and use in source and binary forms, with or without
@@ -25,18 +25,23 @@ def process(vault):
     if "filesystem.dat" in file_list:
         # Sandisk, Sony vault or ENCSecurity user password.
         file_path = Path(vault) / "filesystem.dat"
-
     elif "index.dat" in file_list and "keychain.dat" in file_list:
-        # ENCSecurity vault.
+        # ENCSecurity vault prior to 7.2.1.
         file_path = Path(vault) / "index.dat"
     else:
         sys.stderr.write(f"{vault} : Valid vault not found.\n")
         return
     
+    if "enckey.dat" in file_list:
+        # Sandisk PrivateAccess or ENCSecurity vault 7.2.1 and later.
+        version = 0x10
+    else:
+        version = 0x00
+
     with open(file_path, "rb") as f:
         header = f.read(4)                  # Read header
         if header == b'\xd2\xc3\xb4\xa1':   # Test if we have a valid header
-            version = int.from_bytes(f.read(4),byteorder="little")
+            version += int.from_bytes(f.read(4),byteorder="little")
             crypto = int.from_bytes(f.read(4),byteorder="little")
             iv = binascii.hexlify(f.read(8))
             header_enc = binascii.hexlify(f.read(4))
@@ -50,7 +55,22 @@ def process(vault):
 
     sys.stdout.write(f"{vault}:$encdv${version}${crypto}${iv.decode()}${header_enc.decode()}")
 
-    if version == 3:
+    if version > 0x10:
+        file_size = os.path.getsize(Path(vault) / "enckey.dat")
+        if file_size <= 8:
+            sys.stderr.write("enckey.dat : Problem file too small.\n")
+            return
+        with open(Path(vault) / "enckey.dat", "rb") as f:
+            f.seek(4)
+            length = int.from_bytes(f.read(4),byteorder="big")
+            if length >= file_size - 8:
+                sys.stderr.write("enckey.dat : Problem reading length.\n")
+                return
+            salt = binascii.hexlify(f.read(length))
+            iterations = int.from_bytes(f.read(4),byteorder="big")
+            sys.stdout.write(f"${length}${salt.decode()}${iterations}")
+
+    if version & 0x0F == 3:
         with open(Path(vault) / "keychain.dat", "rb") as f:
             f.seek(16)
             keychain = binascii.hexlify(f.read(128))
--- a/src/encdatavault_fmt_plug.c
+++ b/src/encdatavault_fmt_plug.c
@@ -1,7 +1,7 @@
 /*
  * Cracker for ENCSecurity Data Vault.
  *
- * This software is Copyright (c) 2021 Sylvain Pelissier <sylvain.pelissier at kudelskisecurity.com>
+ * This software is Copyright (c) 2021-2022 Sylvain Pelissier <sylvain.pelissier at kudelskisecurity.com>
  * and it is hereby released to the general public under the following terms:
  *
  * Redistribution and use in source and binary forms, with or without
@@ -24,27 +24,33 @@ john_register_one(&fmt_encdadatavault);
 #include "md5.h"
 #include "common.h"
 #include "formats.h"
+#include "pbkdf2_hmac_sha256.h"
 
 #define FORMAT_TAG              "$encdv$"
 #define FORMAT_TAG_LENGTH       (sizeof(FORMAT_TAG) - 1)
 #define FORMAT_LABEL            "ENCDataVault"
 #define FORMAT_NAME             ""
-#define ALGORITHM_NAME          "MD5"
-#define BENCHMARK_COMMENT       " (1000 iterations)"
+#define ALGORITHM_NAME          "MD5 / PBKDF2-SHA256, AES"
+#define BENCHMARK_COMMENT       ""
 #define BENCHMARK_LENGTH        0x107
 #define PLAINTEXT_LENGTH        125
 #define BINARY_SIZE             0
 #define BINARY_ALIGN            1
-#define SALT_SIZE               sizeof(struct custom_salt)
+#define SALT_SIZE               sizeof(struct salt)
 #define SALT_ALIGN              4
+#ifdef SIMD_COEF_32
+#define MIN_KEYS_PER_CRYPT      SSE_GROUP_SZ_SHA256
+#define MAX_KEYS_PER_CRYPT      SSE_GROUP_SZ_SHA256
+#else
 #define MIN_KEYS_PER_CRYPT      1
 #define MAX_KEYS_PER_CRYPT      4
+#endif
 
 #ifndef OMP_SCALE
 #define OMP_SCALE               16
 #endif
 
-#define ENC_DEFAULT_NUM_HASH_ITERATIONS 1000
+#define ENC_DEFAULT_MD5_ITERATIONS 1000
 #define ENC_SALT_SIZE 16
 #define ENC_IV_SIZE 16
 #define ENC_BLOCK_SIZE 16
@@ -53,6 +59,8 @@ john_register_one(&fmt_encdadatavault);
 #define ENC_SIG_SIZE 4
 #define ENC_MAX_KEY_NUM 8
 #define ENC_KEYCHAIN_SIZE 128
+
+#define PBKDF2_32_MAX_SALT_SIZE 128
 
 typedef union buffer_128_u {
 	uint8_t u8[8];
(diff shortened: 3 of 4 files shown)


doc/README-ENCDataVault.md
```diff
@@ -17,10 +20,12 @@ The decryption test is done on a 32-bit magic value, meaning that occasional
 false positives can occur. Cracked passwords for those false positives will not
 decrypt the vault correctly.
 
-The format has been tested on Sony ENCDataVault lite 6.2.13, SanDisk 
-SecureAccess 3.02 and ENC DataVault 7.1.1W.
+The format using the key derivation algorithm based on MD5, has been tested on 
+Sony ENCDataVault lite 6.2.13, SanDisk SecureAccess 3.02 and ENC DataVault 
+7.1.1W. The format using PBKDF-SHA256, has been tests on PrivateAccess 6.3.6W 
+and ENCDataVault 7.2.1.
```
solardiz, line 26: I think the commas before "has" should be removed, and in the second instance "tests" replaced with "tested".

run/encdatavault2john.py
```diff
@@ -2,7 +2,7 @@
 
 # Helper script for cracking ENCSecurity DataVault.
 #
-# This software is Copyright (c) 2021, Sylvain Pelissier <sylvain.pelissier at kudelskisecurity.com>
+# This software is Copyright (c) 2022, Sylvain Pelissier <sylvain.pelissier at kudelskisecurity.com>
```
solardiz: Should be "2021-2022".

src/encdatavault_fmt_plug.c
```diff
@@ -1,7 +1,7 @@
 /*
  * Cracker for ENCSecurity Data Vault.
  *
- * This software is Copyright (c) 2021 Sylvain Pelissier <sylvain.pelissier at kudelskisecurity.com>
+ * This software is Copyright (c) 2022 Sylvain Pelissier <sylvain.pelissier at kudelskisecurity.com>
```
solardiz: Should be "2021-2022".

src/encdatavault_fmt_plug.c
```diff
@@ -24,27 +24,32 @@ john_register_one(&fmt_encdadatavault);
...
+#include "pbkdf2_hmac_sha256.h"
 
 #define FORMAT_TAG              "$encdv$"
 #define FORMAT_TAG_LENGTH       (sizeof(FORMAT_TAG) - 1)
 #define FORMAT_LABEL            "ENCDataVault"
 #define FORMAT_NAME             ""
-#define ALGORITHM_NAME          "MD5"
-#define BENCHMARK_COMMENT       " (1000 iterations)"
+#define ALGORITHM_NAME          "MD5 / PBKDF2-SHA256, AES"
+#define BENCHMARK_COMMENT       ""
```
solardiz, line 34: While we do have other formats that mix very different algorithms for different input "hashes", this is problematic at least for benchmarks. At the very least, you should expose the different algorithms as a "tunable cost", so that the algorithm(s) in use is/are reported and the `--cost` option can be used. As another option, you could define two separate format structs, even if sharing this same source file and several functions between the formats - that would include both benchmarks in the default `--test` run. Fixing this either way is not essential to this one PR, but you should plan on implementing one of these fixes shortly thereafter - we'll need to open an issue and keep it opened until fixed.

src/encdatavault_fmt_plug.c
```diff
@@ -24,27 +24,32 @@ john_register_one(&fmt_encdadatavault);
...
 #define BINARY_ALIGN            1
-#define SALT_SIZE               sizeof(struct custom_salt)
+#define SALT_SIZE               sizeof(struct enc_salt)
 #define SALT_ALIGN              4
+#ifdef SIMD_COEF_32
+#define MIN_KEYS_PER_CRYPT      SSE_GROUP_SZ_SHA256
+#define MAX_KEYS_PER_CRYPT      SSE_GROUP_SZ_SHA256
+#else
 #define MIN_KEYS_PER_CRYPT      1
 #define MAX_KEYS_PER_CRYPT      4
```
solardiz, line 46: We'd normally include what kind of implementation is used (including bitness and SIMD) at the end of `ALGORITHM_NAME`. In this case, it looks like it's always scalar for the MD5-based algorithm, but is usually SIMD for the SHA-256 based one. That's tricky, and unless you make MD5 use SIMD in the same cases (should be quite easy for you now that you've learned the API?)  is yet another reason to split into two format structs.

Also, with split formats it'd probably be optimal to use `MAX_KEYS_PER_CRYPT` of 1 (yes, max) for the very slow SHA-256 based algorithm when built without SIMD (not the typical case, but nevertheless).

src/encdatavault_fmt_plug.c
```diff
@@ -24,27 +24,32 @@ john_register_one(&fmt_encdadatavault);
...
-
+#endif
```
solardiz, line 47: The empty line changes in this PR are weird - all(?) look like they hurt rather than improve clarity. I guess some might have been because of editor settings, but probably not the one above. In general, please undo them.

src/encdatavault_fmt_plug.c
```diff
@@ -84,21 +91,37 @@ static struct fmt_tests encdatavault_tests[] = {
...
+	// Fake hash to test pbkdf2
+	{
+		"$encdv$17$1$eb012112b3561e6e$85389b0f$32$9b1c17f7a467cf4d13b73dfd3ce12b3dd51cb980cc1f0ccfad4e47e9c9f3d774$100",
+		"123456Aa!"
+	},
+	{
+		"$encdv$17$1$eb012112b3561e6e$014f10cd$32$fa79c83e85a41973799522f525e1316ca07ab663def47b816da76205dc1e3e80$100",
+		"openwallopenwall"
+	},
+	{ "$encdv$19$4$4f1d3889c629968c$e8b68bb8$32$fa79c83e85a41973799522f525e1316ca07ab663def47b816da76205dc1e3e80$50$090a1bb3fc129b1549823a0ed70ffa1eff9033994766f25fcf3645b9a48bcd2175ab187ad1bf9a5f63d4a11aaea3c3b4d8cba240e439c484c7ed0e1e74828ca166fb5de6ea3f56eaf7c97ab217c3ecb5d4231976501fb5a35041bed0aa439cc75bcd8313e00bd24b84a2ed89d5c1246d36c394ba5c774925b48999df75e65b3a", "123456789ABCDEf" },
```
solardiz, line 119: This usage of a fake hash is problematic for benchmarks, where we could otherwise use `--test --cost=...` to get actual performance. I suppose the user can manually scale by 1000, but only if the lower and typical iteration counts are reported as a (second) tunable cost. OTOH, if the actual iteration count is fixed, it feels possibly wrong to support accepting arbitrary ones and report that tunable cost.

BTW, I don't see 100k as necessarily a problematic iteration count for self-test and benchmarks - we do have similar e.g. in `bitcoin*` formats. It is arguably silly to have that startup delay when the actual loaded "hash" is 1000x MD5, though - yet another reason to have separate format structs.

src/encdatavault_fmt_plug.c
```diff
@@ -108,14 +131,16 @@ static size_t cracked_size;
...
 	unsigned int algo_id;
-	unsigned char salt[ENC_SALT_SIZE];
 	unsigned char iv[ENC_IV_SIZE];
+	unsigned int salt_length;
+	unsigned char salt[PBKDF2_32_MAX_SALT_SIZE];
+	unsigned int iterations;
 	unsigned char encrypted_data[ENC_BLOCK_SIZE];
 	unsigned char keychain[ENC_KEYCHAIN_SIZE];
-} *cur_salt;
+} *enc_cur_salt;
```
solardiz: I don't see a reason to rename this struct and pointer. IIRC, we don't do that in other formats, except where this gets moved into an `.h` file for sharing between CPU and OpenCL formats.

src/encdatavault_fmt_plug.c
```diff
@@ -185,19 +210,25 @@ static int valid(char *ciphertext, struct fmt_main *self)
...
+		goto err;
+	int version = atoi(p);
+
+	if (version != 3 && version != 1 && version != 0x13 && version != 0x11) {
 		static int warned;
+
 		if (!self_test_running && !warned++)
-			fprintf(stderr, "%s: Warning: version %d not supported, not loading such hashes!\n", self->params.label, atoi(p));
+			fprintf(stderr, "%s: Warning: version %d not supported, not loading such hashes!\n", self->params.label,
+			        atoi(p));
```
solardiz: Drop the empty lines (I think they hurt clarity), unwrap the line (OK to go to 131 chars), replace the second `atoi(p)` with `version`.

src/encdatavault_fmt_plug.c
```diff
@@ -210,12 +241,34 @@ static int valid(char *ciphertext, struct fmt_main *self)
...
+		if ((p = strtokm(NULL, "$")) == NULL)   // Salt length
+			goto err;
+		if (!isdec(p))
+			goto err;
+		saltlen = atoi(p);
+		if (saltlen > PBKDF2_32_MAX_SALT_SIZE) {
+			static int warned;
+
+			if (!warned++)
+				fprintf(stderr, "%s: Warning: salt length %d too big!\n", self->params.label, atoi(p));
```
solardiz: Drop the empty line, replace the second `atoi(p)` with `saltlen`.

src/encdatavault_fmt_plug.c
```diff
@@ -276,82 +339,118 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 		any_cracked = 0;
 	}
 
-	nb_keys = 1 << (cur_salt->algo_id - 1);
+	nb_keys = 1 << (enc_cur_salt->algo_id - 1);
 
 #ifdef _OPENMP
-#pragma omp parallel for
+	#pragma omp parallel for
```
solardiz, line 336: We normally don't indent `#pragma`, nor contents of an `#ifdef` block.

src/encdatavault_fmt_plug.c
```diff
@@ -276,82 +339,118 @@ static int crypt_all(int *pcount, struct db_salt *salt)
...
 
-		// Key derivation
-		tmp.u64[0] = 0;
-		tmp.u64[1] = 0;
-		MD5_Init(&ctx);
-		MD5_Update(&ctx, saved_key[index], strlen(saved_key[index]));
-		MD5_Final(hash.u8, &ctx);
+		// Key derivation based on MD5
+		if ((enc_cur_salt->version >> 4) == 0) {
+			for (i = 0; i < MIN_KEYS_PER_CRYPT; ++i) {
```
solardiz, line 349: Hmm, this reminds us - you get `MIN_KEYS_PER_CRYPT` artificially raised when you have both algorithms in one format, yet don't support SIMD for MD5. That's yet another reason to either add SIMD for MD5 or split the formats.

src/encdatavault_fmt_plug.c
```diff
@@ -276,82 +339,118 @@ static int crypt_all(int *pcount, struct db_salt *salt)
...
+					memcpy(tmp.u8, result + j * 16, ENC_NONCE_SIZE);
+					ivs[j].u64[0] ^= tmp.u64[0];
+				}
+				// result buffer is reused here to hold the decrypted data.
+				enc_aes_ctr_iterated(enc_cur_salt->encrypted_data, result, result, ivs, AES_BLOCK_SIZE, nb_keys, 1);
+				if (!memcmp(result + 4, "\xd2\xc3\xb4\xa1", ENC_SIG_SIZE)) {
+					cracked[index + i] = 1;
 #ifdef _OPENMP
-#pragma omp atomic
+					#pragma omp atomic
```
solardiz, line 441: We normally don't indent `#pragma`, nor contents of an `#ifdef` block.

solardiz (changes requested):
I made some comments - some are nitpicks and easy to fix (please do), some are important but hard to address (please feel free to leave the tunable costs or splitting into two formats for a separate PR). You'll probably also want to update your `NEWS` entry at least adding the 2022 year.

solardiz:
Also, the title (first line) of the commit message should be 72 chars max (later ones, if any and after a blank line, 75 max), and it should start with subsystem name (in this case format name), so e.g.:

```
ENCDataVault: Support PBKDF2-HMAC-SHA256 key derivation

Tested on PrivateAccess 6.3.6W and ENCDataVault 7.2.1.
```

src/encdatavault_fmt_plug.c
sylvainpelissier: I think I would use separate structs. Could you open an issue please ?

src/encdatavault_fmt_plug.c
sylvainpelissier: I tried to make MD5 using SIMD but I was not able, I always got segfaults. I would need help to make it SIMD. I'll update the NAME in the two formats accordingly.

src/encdatavault_fmt_plug.c
sylvainpelissier: The actual iteration count is fixed by the software thus it is wrong to support arbitrary ones but the iterations count is stored in a file in clear maybe for future releases. I'll uncomment the tests and remove the fake when I'll separate the formats.

src/encdatavault_fmt_plug.c
sylvainpelissier: The empty line was done by using the format code commands in `CONTRIBUTING.md`.

src/encdatavault_fmt_plug.c
solardiz: > the iterations count is stored in a file in clear maybe for future releases.

Oh, then it makes sense to support arbitrary iteration counts - and we should report iteration count as a tunable cost (after you separate the formats).

src/encdatavault_fmt_plug.c
```diff
@@ -279,79 +333,115 @@ static int crypt_all(int *pcount, struct db_salt *salt)
...
+					ivs[j].u64[0] ^= kdf_out[i][j].u64[0];
+				}
+				// result buffer is used here to hold the decrypted data.
+				enc_aes_ctr_iterated(cur_salt->encrypted_data, result, kdf_out[i][0].u8, ivs, AES_BLOCK_SIZE,
+				                     nb_keys, 1);
+				if (!memcmp(result + 4, "\xd2\xc3\xb4\xa1", ENC_SIG_SIZE)) {
+					cracked[index + i] = 1;
 #ifdef _OPENMP
-#pragma omp atomic
+					#pragma omp atomic
```
solardiz, line 415: `#pragma` shouldn't be indented.

src/encdatavault_fmt_plug.c
solardiz: `#pragma` shouldn't be indented - please fix.

src/encdatavault_fmt_plug.c
```diff
@@ -108,14 +126,16 @@ static size_t cracked_size;
...
 	unsigned int algo_id;
-	unsigned char salt[ENC_SALT_SIZE];
 	unsigned char iv[ENC_IV_SIZE];
+	unsigned int salt_length;
+	unsigned char salt[PBKDF2_32_MAX_SALT_SIZE];
+	unsigned int iterations;
 	unsigned char encrypted_data[ENC_BLOCK_SIZE];
 	unsigned char keychain[ENC_KEYCHAIN_SIZE];
-} *cur_salt;
+} * cur_salt;
```
solardiz: We don't normally put a space after `*` here.

doc/README-ENCDataVault.md
```diff
@@ -17,10 +20,12 @@ The decryption test is done on a 32-bit magic value, meaning that occasional
 false positives can occur. Cracked passwords for those false positives will not
 decrypt the vault correctly.
 
-The format has been tested on Sony ENCDataVault lite 6.2.13, SanDisk 
-SecureAccess 3.02 and ENC DataVault 7.1.1W.
+The format using the key derivation algorithm based on MD5 has been tested on 
+Sony ENCDataVault lite 6.2.13, SanDisk SecureAccess 3.02 and ENC DataVault 
+7.1.1W. The format using PBKDF-SHA256, has been tested on PrivateAccess 6.3.6W 
+and ENCDataVault 7.2.1.
```
solardiz, line 26: This still has a weird comma before "has".

doc/README-ENCDataVault.md
```diff
@@ -7,6 +7,9 @@ Cracking ENCSecurity DataVault
 $ ../run/encdatavault2john.py path/to/vault > hash
 ```
 
+For Sandisk and Sony the script must be run in the "Settings" folder for example 
+"SanDiskSecureAccess Settings".
```
solardiz, line 11: Add punctuation like:

```
For SanDisk and Sony the script must be run in the "Settings" folder - for example,
"SanDiskSecureAccess Settings". 
```

solardiz (approved):

src/encdatavault_fmt_plug.c
solardiz: > The empty line was done by using the format code commands in `CONTRIBUTING.md`.

Which one of the two commands `indent` or `astyle`? Maybe we need to clarify that only one should be run, and which one is preferred (if either one works better for us than the other - I don't know if that's the case).

sylvainpelissier: `indent`

src/encdatavault_fmt_plug.c
solardiz: OK, I'll give up on this one. ;-)

sylvainpelissier: Sorry I was testing indent, it redo it.

PR #5073, opened 2022-02-14
LUKS: made valid() much stricter to avoid overflows

### LINKED ISSUE #5070: LUKS: crashes with synthetic inputs with ASan
While fuzzer gets killed by OOM (#4979) it is possible to crash manually:

`keybytes` 0 and `stripes` 0 taken from binary header:
```
$ python -c 'print "$luks$1$592$" + "4c554b53babe00" + "0" * 1170 + "$0$A$" + "b" * 40' > t.pw
$ ./run/john --format=luks t.pw
[...]
==21413==ERROR: AddressSanitizer: FPE on unknown address 0x55b1a44ea869 (pc 0x55b1a44ea869 bp 0x7ffcbd237d60 sp 0x7ffcbd237d20 T0)
    #0 0x55b1a44ea868 in AF_merge /home/user/john/src/luks_fmt_plug.c:184
    #1 0x55b1a44ec3fe in crypt_all /home/user/john/src/luks_fmt_plug.c:582
```

Insufficient data:
```
$ python -c 'print "$luks$1$592$" + "4c554b53babe000161657300000000000000000000000000000000000000000000000000000000006362632d65737369763a73686132353600000000000000000000000000000000736861310000000000000000000000000000000000000000000000000000000000000408000000104f386b50df3fcd9132589a934851faaff16709ff628ed0b628b0d7151b3600c0b3f95d8404a8b35fdf5dd6b6ff10f4c352fde11900010b1762663664393836622d633836352d343261622d616534662d6165313336633938383735360000000000ac71f3000430f5d9c39e349b48d7cf1771d9c152840b389a4353ff186436ec75cc397529ed40260000000800000fa00000dead0000000000000000000000000000000000000000000000000000000000000000000000000000008800000fa00000dead0000000000000000000000000000000000000000000000000000000000000000000000000000010800000fa00000dead0000000000000000000000000000000000000000000000000000000000000000000000000000018800000fa00000dead0000000000000000000000000000000000000000000000000000000000000000000000000000020800000fa00000dead0000000000000000000000000000000000000000000000000000000000000000000000000000028800000fa00000dead0000000000000000000000000000000000000000000000000000000000000000000000000000030800000fa00000dead0000000000000000000000000000000000000000000000000000000000000000000000000000038800000fa0" + "$64000$A$" + "b" * 40' > t.pw
$ ./run/john --format=luks t.pw
[...]
==21937==ERROR: AddressSanitizer: SEGV on unknown address 0x625000010000 (pc 0x7f696d7f2c3d bp 0x000000000000 sp 0x7ffe9b2daab8 T0)
==21937==The signal is caused by a READ memory access.
    #0 0x7f696d7f2c3c in AES_decrypt (/lib/x86_64-linux-gnu/libcrypto.so.1.1+0x8fc3c)
    #1 0x7f696d8f3815 in CRYPTO_cbc128_decrypt (/lib/x86_64-linux-gnu/libcrypto.so.1.1+0x190815)
    #2 0x55b25fab01d5 in decrypt_aes_cbc_essiv /home/user/john/src/luks_fmt_plug.c:241
```

AlekseyCherepanov:
Closes #5070

I did not touch dmg. But everything else is fixed. `--test` is ok. Fuzzing does not complete in reasonable time. But it does not get killed quickly by OOM anymore. OTOH `valid` checks redundancies, so dumb fuzzer is unlikely to bypass it.

### DIFF
--- a/src/johnswap.h
+++ b/src/johnswap.h
@@ -87,4 +87,16 @@
 	}
 #endif
 
+#if ARCH_LITTLE_ENDIAN
+#define john_htonl(x) JOHNSWAP((x))
+#define john_ntohl(x) JOHNSWAP((x))
+#define john_htonll(x) JOHNSWAP64((x))
+#define john_ntohll(x) JOHNSWAP64((x))
+#else
+#define john_htonl(x) (x)
+#define john_ntohl(x) (x)
+#define john_htonll(x) (x)
+#define john_ntohll(x) (x)
+#endif
+
 #endif // __JOHN_SWAP_H__
--- a/src/luks_fmt_plug.c
+++ b/src/luks_fmt_plug.c
@@ -96,17 +96,6 @@ john_register_one(&fmt_luks);
 #define MAX_KEYS_PER_CRYPT  1
 #endif
 
-#if ARCH_LITTLE_ENDIAN
-#define john_htonl(x) ((((x)>>24) & 0xffL) | (((x)>>8) & 0xff00L) | \
-		(((x)<<8) & 0xff0000L) | (((x)<<24) & 0xff000000L))
-
-#define john_ntohl(x) ((((x)>>24) & 0xffL) | (((x)>>8) & 0xff00L) | \
-		(((x)<<8) & 0xff0000L) | (((x)<<24) & 0xff000000L))
-#else
-#define john_htonl(x) (x)
-#define john_ntohl(x) (x)
-#endif
-
 #include "luks_insane_tests.h"
 
 /* taken from LUKS on disk format specification */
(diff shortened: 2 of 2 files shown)


solardiz (approved):
Looks good to me, although I can't claim to have fully reviewed it. I worry a bit that "check base64 and precise number of trailing" probably didn't have all of its code paths tested with our test vector, or did it? Also, this kind of checking should probably be moved to a shared file and reused later. Anyway, this is probably OK to merge now.

AlekseyCherepanov:
Self-tests cover `=` with `afsize` 128000 and `==` with `afsize` 64000. I tested `case 0` and wrong char for `case 1`. `default` means that there are more `=` than 2. I did not test it and wrong chars for `case 2`.

`base64_valid_length` can return length with `=`, but it does not help checking. While excessive `=` can be a problem sometimes. Allowing only `=` and `==` could be moved into `base64_valid_length` (with a separate flag?).

Part of my checks are possible because there is length of binary data in ciphertext. I guess it is not common.

PR #5077, opened 2022-02-15
dmg2john: replaced inlined htonl() with john_htonl(), fixes big-endian

### LINKED ISSUE #5075: dmg2john broken on big-endian
As reported by @AlekseyCherepanov in #5070:
`ntohll` in `dmg2john.c` is defined outside of endianness check, so I guess `dmg2john.c` should work incorrectly in BE environment.

AlekseyCherepanov:
Closes #5075

I tested this way:
```
$ ./run/dmg2john ../john-samples/Apple_DMG/* > dmg_orig.pw
[...]

[... changed code ...]
$ ./run/dmg2john ../john-samples/Apple_DMG/* > dmg_new.pw
[...]
$ cmp dmg_orig.pw dmg_new.pw
```

I removed `#ifndef ntohl`. `ntohl` is not defined for me. Actually all calls are located in `*_fix` functions at the beginning of file with linear code. So `sed` would work too.

Somehow `dmg2john` is not mentioned in `Makefile.legacy`.

`htonll` is not used. Together with names of functions, it makes me think that `htonl` should be `ntohl` really.

Changing functions, it might be more elegant to use specific macro to assign in place, like that:
```c
	inplace_fix(hdr->kdf_iteration_count);
// instead of
	hdr->kdf_iteration_count = htonl(hdr->kdf_iteration_count);
```

(And to make it more convenient (and implicit), it is possible to use `_Generic` to distinguish 32-bit and 64-bit fields. It would bump C to C11 though. Expanding idea further, I would list all fields with macro and sum `sizeof` of fields to `assert` that none of fields are skipped (in run-time unfortunately).)

If these possibilities should not be implemented, then PR is ready.

### DIFF
--- a/src/Makefile.in
+++ b/src/Makefile.in
@@ -304,7 +304,7 @@ DES_std.o:	DES_std.c arch.h common.h memory.h DES_std.h os.h os-autoconf.h autoc
 
 detect.o:	detect.c
 
-dmg2john.o:	dmg2john.c autoconfig.h arch.h filevault.h misc.h jumbo.h memory.h os.h os-autoconf.h
+dmg2john.o:	dmg2john.c autoconfig.h arch.h filevault.h misc.h jumbo.h memory.h johnswap.h os.h os-autoconf.h
 
 dummy.o:	dummy.c common.h arch.h memory.h formats.h params.h misc.h jumbo.h autoconfig.h options.h list.h loader.h getopt.h os.h os-autoconf.h
 
--- a/src/dmg2john.c
+++ b/src/dmg2john.c
@@ -239,7 +239,7 @@ static void hash_plugin_parse_hash(char *in_filepath)
 			fprintf(stderr, "Unable to seek in %s\n", filename);
 			goto bailout;
 		}
-		if (read(fd, &header, sizeof(cencrypted_v1_header)) < 1) {
+		if (read(fd, &header, sizeof(cencrypted_v1_header)) != sizeof(cencrypted_v1_header)) {
 			fprintf(stderr, "%s is not a DMG file!\n", filename);
 			goto bailout;
 		}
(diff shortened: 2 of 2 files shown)


solardiz:
> If these possibilities should not be implemented, then PR is ready.

I think it's good to implement all of these up to and including `inplace_fix`, but excluding what you wrote in braces (C11 stuff, some of sizes). Also, I think it's good to add that cleanup as a separate commit, keeping your initial minimal commit intact.

solardiz:
> up to and including `inplace_fix`, but excluding what you wrote in braces

This means that for the 64-bit fields you'll add a separate macro, or just handle them without a macro (like they are now) - I think it's only two fields.

AlekseyCherepanov:
Command over selection to fix `header_byteorder_fix`, check lines, change lines:
```
perl -pe 's/htonl/ntohl/'
grep -v -P '^\s+(\S+) = ntohl\(\1\);$'
perl -pe 's/^\t.* = /\tinplace_/'
```

Then I reverted change for 64-bit fields manually. I considered that `inplace_ntohll` looks very similar to `inplace_ntohl`. Regular assignments make these 2 lines to stand out. I only added `john_` prefix in them.

(`inplace_ntohl` macro could do `static_assert` to check that argument's `sizeof` is 4 to avoid misuse. `static_assert` is C11 again, but it can be implemented through array with negative size.)

AlekseyCherepanov:
I added checks that full structures are read. Most calls use variable `count` for that.

Before the first changed call, there is `lseek` with the same size stepped from the end. So `lseek` would fail if there is not enough data. The second call reads from the beginning, but size is smaller than in the first call, so there should be enough data to fill structure even in the worst case. So my change is not important during regular usage. There can be difference if file is truncated between `lseek` and `read` or between two `read` calls.

AlekseyCherepanov:
Actually I did not add checks, I fixed existing. So I changed commit message.

solardiz (approved):

Issue #5029, opened 2022-01-29
off-by-one write in utf16_to_utf8_r or misuse in enc_lc/enc_uc


AlekseyCherepanov:
Working on #5027 I found that `enc_lc` can write one more byte than specified by `dst_bufsize` argument. Name suggests that it is ok to pass `sizeof` of a buffer as this argument, but it causes overflow. It happens only with `--encoding=utf-8` because write happens in `utf16_to_utf8_r`.

I am not sure if it is a bug in `utf16_to_utf8_r` because respective argument there is called `dst_len`. It might be misuse of `utf16_to_utf8_r` in `enc_lc`/`enc_uc`. Does `dst_len` include terminating null byte?

(I am not sure of meaning of arguments. I saw already that `base64_convert` can put null byte at `to_len` position. BTW is it ok?)

The problem is kind of real: I can trigger similar crash through `enc_uc` in a dynamic format.

`dynamic_fmt.c`:
```c
static char *HandleCase(char *cp, int caseType)
{
	static UTF8 dest[256];
[...]
			enc_uc(dest, sizeof(dest), (unsigned char*)cp, strlen(cp));
[...]
			enc_lc(dest, sizeof(dest), (unsigned char*)cp, strlen(cp));
```

```
$ python -c 'print "b" * 32' > t.pw
$ python -c 'print "\xce\x90" * 42 + "aaaaa"' > t.txt
$ ./run/john --format='dynamic=md5(uc($p))' t.pw --wordlist=t.txt
[...]
==23927==ERROR: AddressSanitizer: global-buffer-overflow on address 0x557a9c5e4c00 at pc 0x557a9c083cb1 bp 0x7fffff6ffcd0 sp 0x7fffff6ffcc8
WRITE of size 1 at 0x557a9c5e4c00 thread T0
    #0 0x557a9c083cb0 in utf16_to_utf8_r /home/user/john/src/unicode.c:673
    #1 0x557a9c08438d in utf16_to_enc_r /home/user/john/src/unicode.c:793
    #2 0x557a9c08a387 in enc_uc /home/user/john/src/unicode.c:1834
    #3 0x557a9bc397b9 in HandleCase /home/user/john/src/dynamic_fmt.c:8389
    #4 0x557a9bc39991 in set_key /home/user/john/src/dynamic_fmt.c:1315
    #5 0x557a9bfeb09c in process_key /home/user/john/src/cracker.c:1166
[...]

PR #5082, opened 2022-02-18
small changes, including disabling init of ztex boards during fuzzing

AlekseyCherepanov:
`--fuzz` did not work for `ztex` formats without boards connected with `No ZTEX devices found.`, so I added a check.

Meanwhile `--fuzz-dump` worked.

I completed `--format=ztex --fuzz=run/fuzz.dic` with the fix without real boards connected.

`max_keys_per_crypt` is set in `device_format_reset` based on number of boards (and mask in use). But fuzzing does not call `reset`.

I checked my last statement with `gdb` (both `nt` and `raw-md5` use `fmt_default_reset`):
```
$ printf 'break fmt_default_reset \n run --format=raw-md5,nt --fuzz\n bt' | gdb ./run/john
[...]
[Inferior 1 (process 31908) exited with code 01]
(gdb) No stack.
(gdb) quit
```

Other changes are minor.

Maybe I should join commits for `doc` into single.

If I should not, then this PR is ready.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,10 +25,10 @@ John the Ripper is released under GNU GPL v2 "or later", with portions also avai
 Solar Designer's current preference is that new code contributions be licensed under very liberal terms:
 ```
 /*
- * This software is Copyright (c) YEAR YOUR NAME <your at e-mail.address>,<br>
- * and it is hereby released to the general public under the following terms:<br>
- * Redistribution and use in source and binary forms, with or without<br>
- * modification, are permitted.<br>
+ * This software is Copyright (c) YEAR YOUR NAME <your at e-mail.address>,
+ * and it is hereby released to the general public under the following terms:
+ * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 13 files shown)


AlekseyCherepanov:
The fixed integer overflow is in my code added in #5073.

Besides from undefined behaviour, it might lead to buffer overflow if `INT_MAX` would be possible in `afsize` at this point: `+2` overflows into negative numbers, `/3*4` overflows back into positive but gives lower value than expected, so `valid` would accept shorter data than it should. But `INT_MAX` and close values are not possible there due to other check. Other overflowing values would give negative `int` producing huge `size_t` that would not pass further checks.

AlekseyCherepanov:
I enlarged buffer of uppercase/lowercase. I know that length can grow 3 times. So `110 * 3` is required for candidates. But they might not fit later in buffer (safely). So I am not sure how much is needed really. Maybe it is not needed at all.

OTOH usernames can have unlimited lengths. Hence temporary fix for #5029.

`--format=dynamic,disabled --test=0` is ok.

PR #5083, opened 2022-02-20
ENCDataVault: Split into two formats

### LINKED ISSUE #5076: Split ENCDataVault into two formats
As discussed with @sylvainpelissier in #5071, the plan is to split the ENCDataVault format in two. Keep them in the same source file, and share much of the code, but have separate format structs. Also, expose the PBKDF2-HMAC-SHA256 iteration count as a tunable cost.

sylvainpelissier:
- Splits the ENCDataVault format in two format structs.
- Exposes the PBKDF2-HMAC-SHA256 iteration count as a tunable cost.
- Updates the extraction script with the new tag name.
- Closes #5076


src/encdatavault_fmt_plug.c
```diff
@@ -26,24 +28,32 @@ john_register_one(&fmt_encdadatavault);
...
+#define ALGORITHM_NAME_MD5        "MD5, AES"
 #ifdef SIMD_COEF_32
-#define MIN_KEYS_PER_CRYPT      SSE_GROUP_SZ_SHA256
-#define MAX_KEYS_PER_CRYPT      SSE_GROUP_SZ_SHA256
+#define ALGORITHM_NAME_PBKDF2     "PBKDF2-SHA256 HMAC-256 " SHA256_ALGORITHM_NAME
 #else
-#define MIN_KEYS_PER_CRYPT      1
-#define MAX_KEYS_PER_CRYPT      4
+#define ALGORITHM_NAME_PBKDF2     "PBKDF2-SHA256 HMAC-256 32/" ARCH_BITS_STR
+#endif
```
solardiz: Please either drop "HMAC-256 " entirely (there's no such thing) or move it into "PBKDF2-HMAC-SHA256" (which is the full name for this construction), in both places.

src/encdatavault_fmt_plug.c
```diff
@@ -192,24 +203,30 @@ static void done(void)
 	MEM_FREE(cracked);
 }
 
-static int valid(char *ciphertext, struct fmt_main *self)
-{
+static int valid_common(char *ciphertext, struct fmt_main *self, int is_pbkdf2) {
```
solardiz: For functions (only), we put the opening curly brace on its own line.

src/encdatavault_fmt_plug.c
```diff
@@ -246,15 +263,15 @@ static int valid(char *ciphertext, struct fmt_main *self)
 			goto err;
 		}
 		if ((p = strtokm(NULL, "$")) == NULL)   // Salt
-			goto err;
+		goto err;
```
solardiz: What's happened with the indentation here?

src/encdatavault_fmt_plug.c
```diff
@@ -266,51 +283,78 @@ static int valid(char *ciphertext, struct fmt_main *self)
...
+
+}
+
+static int valid_md5(char *ciphertext, struct fmt_main *self) {
+	return valid_common(ciphertext, self, 0);
+}
+
+static int valid_pbkdf2(char *ciphertext, struct fmt_main *self) {
+	return valid_common(ciphertext, self, 1);
 }
```
solardiz: Curly braces onto their own lines, please.

src/encdatavault_fmt_plug.c
```diff
@@ -266,51 +283,78 @@ static int valid(char *ciphertext, struct fmt_main *self)
...
 
-static void *get_salt(char *ciphertext)
+static void get_salt_common(char *ciphertext, struct salt *cs, int is_pbkdf2)
 {
-	static struct salt cs;
 	int i;
 	char *p = ciphertext, *ctcopy, *keeptr;
 
-	memset(&cs, 0, sizeof(cs));
```
solardiz, line 277: Please keep `cs` and its `memset` inside the function, and return a pointer to the `struct` like you did before. No need to use a separate return buffer for the two `get_salt` functions. Also, even if you do use separate return buffers, you can keep the `memset` shared in the common function.

src/encdatavault_fmt_plug.c
```diff
@@ -266,51 +283,78 @@ static int valid(char *ciphertext, struct fmt_main *self)
...
+	return (void *)&cs;
+}
+
+static void *get_salt_pbkdf2(char *ciphertext) {
+	static struct salt cs;
+
+	memset(&cs, 0, sizeof(cs));
+	get_salt_common(ciphertext, &cs, 1);
 	return (void *)&cs;
 }
```
solardiz: Please move at least the `memset` back into the common function, and curly braces onto separate lines.

src/encdatavault_fmt_plug.c
```diff
@@ -26,24 +28,32 @@ john_register_one(&fmt_encdadatavault);
...
+#define BINARY_ALIGN              1
+#define SALT_SIZE                 sizeof(struct salt)
+#define SALT_ALIGN                4
+#ifdef SIMD_COEF_32
+#define MIN_KEYS_PER_CRYPT        SSE_GROUP_SZ_SHA256
+#define MAX_KEYS_PER_CRYPT        SSE_GROUP_SZ_SHA256
+#else
+#define MIN_KEYS_PER_CRYPT        1
+#define MAX_KEYS_PER_CRYPT        1
 #endif
```
solardiz: Part of the reason to split into two formats is not to force `*_KEYS_PER_CRYPT` to `SSE_GROUP_SZ_SHA256` in the MD5-based format (and once we implement SIMD for it later, to use the constants proper for the SIMD implementation of MD5, not SHA-256). So you need a separate set of MIN/MAX per format.

src/encdatavault_fmt_plug.c
```diff
@@ -345,59 +389,129 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 		unsigned char result[ENC_KEY_SIZE * ENC_MAX_KEY_NUM] = { 0 };
 
 		// Key derivation based on MD5
-		if ((cur_salt->version >> 4) == 0) {
-			for (i = 0; i < MIN_KEYS_PER_CRYPT; ++i) {
-				tmp.u64[0] = 0;
-				tmp.u64[1] = 0;
+		for (i = 0; i < MIN_KEYS_PER_CRYPT; ++i) {
```
solardiz: Can simplify to assume `MIN_KEYS_PER_CRYPT` is 1 for MD5 for now (once you split the MIN/MAX between formats), or just revert to using old code (prior to the introduction of PBKDF2), although arguably written this way the code is slightly closer to what it'd be for SIMD support.

solardiz (commented):
Please also unindent _all_ of the `#pragma` lines.

src/encdatavault_fmt_plug.c
sylvainpelissier: If you don't mind I would like to keep it that way. I may try to support SIMD later.

solardiz: OK, keep it.

src/encdatavault_fmt_plug.c
```diff
@@ -26,41 +28,50 @@ john_register_one(&fmt_encdadatavault);
...
+#define SALT_SIZE                 	sizeof(struct salt)
+#define SALT_ALIGN                	4
+#ifdef SIMD_COEF_32
+#define MIN_KEYS_PER_PBKDF2_CRYPT	SSE_GROUP_SZ_SHA256
+#define MAX_KEYS_PER_PBKDF2_CRYPT   SSE_GROUP_SZ_SHA256
+#else
+#define MIN_KEYS_PER_PBKDF2_CRYPT   1
+#define MAN_KEYS_PER_PBKDF2_CRYPT   1
+#endif
+#define KEYS_PER_MD5_CRYPT   1
```
solardiz: Please use two separate macros for `*_KEYS_PER_MD5_CRYPT`. Otherwise it's quite ridiculous to have the `+=` in the `for` loop, which would skip the full range if `MAX` is increased from 1.

src/encdatavault_fmt_plug.c
```diff
@@ -26,41 +28,50 @@ john_register_one(&fmt_encdadatavault);
...
+#define BINARY_SIZE               	0
+#define BINARY_ALIGN              	1
+#define SALT_SIZE                 	sizeof(struct salt)
+#define SALT_ALIGN                	4
+#ifdef SIMD_COEF_32
+#define MIN_KEYS_PER_PBKDF2_CRYPT	SSE_GROUP_SZ_SHA256
+#define MAX_KEYS_PER_PBKDF2_CRYPT   SSE_GROUP_SZ_SHA256
+#else
+#define MIN_KEYS_PER_PBKDF2_CRYPT   1
+#define MAN_KEYS_PER_PBKDF2_CRYPT   1
```
solardiz: Fix the `MAN` typo.

src/encdatavault_fmt_plug.c
```diff
@@ -26,41 +28,50 @@ john_register_one(&fmt_encdadatavault);
...
+#endif
+#define KEYS_PER_MD5_CRYPT   1
 
 #ifndef OMP_SCALE
-#define OMP_SCALE               16
+#define OMP_SCALE               	16
 #endif
```
solardiz: Ideally, we'd have higher `OMP_SCALE` for MD5 (like 16 here) and a lower one (perhaps 1 is optimal? tune it) for the PBKDF2.

solardiz (changes requested):
The `#pragma` lines still look indented.

solardiz:
Oh, the differences between the formats became extensive enough that maybe it'd have been cleaner to have a common file and then two per-format files - then there would be no need to rename so many things.

solardiz:
Oh, wow. I did suggest that splitting this into 3 files like that would help clean it up, but that is not fully done yet. I'd have _many_ new comments now.
@sylvainpelissier How about we merge this as-is, and I clean it up with a follow-up PR of my own? Trying to do that collaboratively is too inefficient.

sylvainpelissier:
> Oh, wow. I did suggest that splitting this into 3 files like that would help clean it up, but that is not fully done yet. I'd have _many_ new comments now. @sylvainpelissier How about we merge this as-is, and I clean it up with a follow-up PR of my own? Trying to do that collaboratively is too inefficient.

Yes sure.

solardiz:
I'll merge this now, even though I'm not available to proceed with the cleanups right away.

A question not directly related to this PR - shouldn't we have some error checking (exception handling?) on the `f.seek` operations in the Python script?

sylvainpelissier:
> I'll merge this now, even though I'm not available to proceed with the cleanups right away.
> 
> A question not directly related to this PR - shouldn't we have some error checking (exception handling?) on the `f.seek` operations in the Python script?

You are right I'll correct that.

PR #5091, opened 2022-03-13
multibit2john.py: broke out protobuf code for .wallet file parsing

### LINKED ISSUE #5089: Problem with multibit2john.py: TypeError: Descriptors should not be created directly, but only retrieved from their parent.
I'm having a problem using the latest version of multibit2john.py.  When I try to execute the script with a Multibit Classic .key file, I get the following error:

~/Desktop$ python3 multibit2john.py multibit.key 
  File "multibit2john.py", line 56, in <module>
    _descriptor.EnumValueDescriptor(
  File "/home/crypto/.local/lib/python3.8/site-packages/google/protobuf/descriptor.py", line 731, in __new__
    _message.Message._CheckCalledFromGeneratedFile()
TypeError: Descriptors should not be created directly, but only retrieved from their parent.

I have verified that I'm using the latest version of the multibit2john.py script, found here: https://github.com/openwall/john/blob/bleeding-jumbo/run/multibit2john.py

I'm using python 3.8.10 on Ubuntu:
~/Desktop$ python -V
Python 3.8.10

I have also verified that I have the latest version of the google protobuf module:
~/Desktop$ sudo pip install --user protobuf
Requirement already satisfied: protobuf in /usr/lib/python3/dist-packages (3.6.1)

The only workaround that I've found to date is to remove all of the protobuf code, lines 20 - 1295.  

If I remove those lines of code, multibit2john.py appears to execute correctly, and it generates a reasonable looking hash (according to https://hashcat.net/wiki/doku.php?id=example_hashes).  

But, I'm concerned that, perhaps, the protobuf code was included for a reason -- and I have routed around that reason.

Any suggestions for fixing the error?

Thanks, Chris

exploide:
Importing compiled protobuf definitions seems to only work when the file ends in _pb2.
Hence, the protobuf code is moved to a separate file that can be imported.

fixes #5089

Using this approach, generating hashes for the multibit `.key` file from john-samples works again.

Concerning the `.wallet` format, that this change is about, there is no sample file. So I could only check that the imports still work.


solardiz:
> Concerning the `.wallet` format, that this change is about, there is no sample file. So I could only check that the imports still work.

@kholia Maybe you can add a sample?

run/multibit2john.py
```diff
@@ -17,1282 +17,10 @@
...
-_sym_db.RegisterMessage(ExchangeRate)
-
-
-DESCRIPTOR.has_options = True
-DESCRIPTOR._options = _descriptor._ParseOptions(descriptor_pb2.FileOptions(), _b('\n\023org.bitcoinj.walletB\006Protos'))
-# @@protoc_insertion_point(module_scope)
-
-# End of automatically generated code.
+# The precompiled file is shipped in protobuf/wallet_pb2.py with John.
+from protobuf.wallet_pb2 import *
```
solardiz, line 23: I'm not familiar with how Python resolves `from protobuf.wallet_pb2 import *` into script file to import. If the main script is run as e.g. `/path/to/john/run/multibit2john.py`, what directories would be checked for the imported script? Would it be only `/path/to/john/run/protobuf` or perhaps also some others - perhaps at least some system directories? Would this include the current directory (a security concern)?

exploide: It prefers the location of the script file, i.e. John's `run/` directory. So it should be safe.

solardiz:
@exploide diff'ing the new `protobuf/wallet_pb2.py` against the old `multibit2john.py`, I see this change:

```diff
@@ -32,7 +10,7 @@
     from google.protobuf import symbol_database as _symbol_database
     from google.protobuf import descriptor_pb2
 except ImportError:
-    sys.stderr.write("Install the missing protobuf package, use 'pip install --user protobuf' command to do so.\n")
+    sys.stderr.write("Install the missing protobuf package, use 'pip install protobuf' command to do so.\n")
     sys.exit(1)
 # @@protoc_insertion_point(imports)
 
```

Intentional? Why?

exploide:
> Intentional? Why?

pip defaults to `--user` install if run without root permissions, so it's not necessary. Besides, there are reasons to not use `--user`, e.g. when installing within a virtualenv.

I believe it is sufficient to guide the user towards the right direction. How to install the dependencies in the specific situation is up to the user.

Issue #5090, opened 2022-03-13
cloudkeychain ASan crash


solardiz:
As reported by @sylvainpelissier in https://huntr.dev/bounties/d2f44226-775f-4871-95ab-a33b14755355

```
Description

For 1Password Cloud Keychain plugin, the length of inputs are not properly checked. Then inputs are copied to fixed length buffers. For example, creating a salt with a larger length allow a buffer overflow.
Proof of Concept

Using the [cloudkeychain.hash](https://www.swisstransfer.com/d/73891c11-55a1-487e-ac2d-7a4777824fab) file:

$ ./run/john cloudkeychain.hash 
=================================================================
==103166==ERROR: AddressSanitizer: global-buffer-overflow on address 0x55f917d99688 at pc 0x55f9172ebbb6 bp 0x7fff28faa160 sp 0x7fff28faa150
WRITE of size 1 at 0x55f917d99688 thread T0
    #0 0x55f9172ebbb5 in get_salt /home/sylvain/software/john/src/cloudkeychain_fmt_plug.c:124
    #1 0x55f91751bd43 in ldr_load_pw_line /home/sylvain/software/john/src/loader.c:1045
    #2 0x55f91751883f in read_file /home/sylvain/software/john/src/loader.c:255
    #3 0x55f91751e8e0 in ldr_load_pw_file /home/sylvain/software/john/src/loader.c:1198
    #4 0x55f91750cf42 in john_load /home/sylvain/software/john/src/john.c:1134
    #5 0x55f91750cf42 in john_init /home/sylvain/software/john/src/john.c:1578
    #6 0x55f91750cf42 in main /home/sylvain/software/john/src/john.c:2065
    #7 0x7f5306c760b2 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x270b2)
    #8 0x55f916fe52cd in _start (/home/sylvain/software/john/run/john+0x1ee2cd)

0x55f917d99688 is located 56 bytes to the left of global variable 'cur_salt' defined in 'cloudkeychain_fmt_plug.c:77:28' (0x55f917d996c0) of size 8
0x55f917d99688 is located 0 bytes to the right of global variable 'cs' defined in 'cloudkeychain_fmt_plug.c:116:28' (0x55f917d97e00) of size 6280
SUMMARY: AddressSanitizer: global-buffer-overflow /home/sylvain/software/john/src/cloudkeychain_fmt_plug.c:124 in get_salt
[...]

PR #5092, opened 2022-03-13
cloudkeychain: check input salt and masterkey lengths

sylvainpelissier:
Check lengths to avoid buffer overflow. Solves #5090 .


solardiz (approved):
Looks reasonable to me, thanks!

Issue #4736, opened 2021-05-26
Consider working around Git on Windows' use of CRLF vs. bash scripts


bytetheripper:
When I try to run ./configure on the /john/src file in Cygwin on my Windows 10 device, I get the following error:

$ ./configure
./configure: line 16: $'\r': command not found
./configure: line 31: syntax error near unexpected token `newline'
'/configure: line 31: `     ;;

When I looked at the /src file I found this disclaimer starting on line 17211:

# On cygwin, bash can eat \r inside `` if the user requested igncr.
# But we know of no other shell where ac_cr would be empty at this
# point, so we can use a bashism as a fallback.

Is there a workaround I can use?

ghost:
I ran a build myself on May 11th [1] (on Windows 2016 using Cygwin 3.2.0) and it worked fine. Although this issue is worded as a bug report, not as a question, I'm afraid you're mistaken.

* Could you make your point more clear?
* if you need support as an end-user, ask for help on the mailing list john-users at lists.openwall.com (but first see the archives at http://www.openwall.com/lists/john-users/ and make sure you set a descriptive message Subject when you post).

[1]
https://dev.azure.com/claudioandre-br/JohnTheRipper/_build/results?buildId=320&view=logs&s=52e5c710-3f09-5ad6-0c38-decbe96b0346

magnumripper:
https://www.openwall.com/lists/john-users/2021/05/26/1

solardiz:
Searching for mentions of related problems seen by other projects finds e.g.:

https://github.com/reconquest/import.bash/issues/2
https://community.perforce.com/s/article/2973

Reading and summarizing these and the like, it appears that Git on Windows defaults to using CRLF for text files ("On windows systems core.autocrlf is normally set to "true" globally.")

Some workarounds were mentioned in the thread on john-users. However, maybe as a project we should "add .gitattributes file [...] and force git to handle shell scripts eol=lf"? Re-opening for us to consider this.

This leaves me wondering why we didn't get this issue reported to us many times before, and why it didn't occur in @claudioandre-br's builds. Any ideas? Or maybe "is normally set to" doesn't mean it's the "default", but is something a Git user does, and most of those advanced users also worked around the consequences without needing our help? Anyway, we could want to provide the above workaround in our project.

ghost:
> add .gitattributes file [...] and force git to handle shell scripts eol=lf"

[...]

PR #5099, opened 2022-03-29
Use "jtr" prefixed fseek & ftell functions to allow larger files

visuve:
- Optimal way would be to use fstat, but will do for now
  - More info: https://wiki.sei.cmu.edu/confluence/display/c/FIO19-C.+Do+not+use+fseek%28%29+and+ftell%28%29+to+compute+the+size+of+a+regular+file

### DIFF
--- a/src/bitlocker2john.c
+++ b/src/bitlocker2john.c
@@ -70,7 +70,7 @@ unsigned char value_type[2] = { 0x00, 0x05 };
 unsigned char padding[16] = {0};
 
 int userPasswordFound = 0, recoveryPasswordFound = 0, found_ccm = 0;
-long int fp_before_aes = 0, fp_before_salt = 0;
+int64_t fp_before_aes = 0, fp_before_salt = 0;
 FILE *fp_eimg;
 int salt_pos[2] = {12, 32};
 int aes_pos[1] = {147};
@@ -97,22 +97,22 @@ static int rp_search_salt_aes(void)
 	int ret = 0, x, y;
 
 	for (x = 0; x < 2; x++) {
-		ret = fseek(fp_eimg, salt_pos[x], SEEK_CUR);
+		ret = jtr_fseek64(fp_eimg, salt_pos[x], SEEK_CUR);
 		FRET_CHECK(ret)
 
 		fillBuffer(fp_eimg, r_salt, SALT_SIZE);
 		printf("Salt: ");
 		print_hex(r_salt, SALT_SIZE, stdout);
 		printf("\n");
-		fp_before_aes = ftell(fp_eimg);
+		fp_before_aes = jtr_ftell64(fp_eimg);
 		FRET_CHECK(fp_before_aes)
-		fprintf(stderr, "Searching AES-CCM from 0x%lx\n", fp_before_aes);
+		fprintf(stderr, "Searching AES-CCM from 0x%llx\n", (unsigned long long)fp_before_aes);
 
 		for (y = 0; y < 1; y++) {
-			ret=fseek(fp_eimg, aes_pos[y], SEEK_CUR);
+			ret = jtr_fseek64(fp_eimg, aes_pos[y], SEEK_CUR);
 			FRET_CHECK(ret)
 
-			fprintf(stderr, "Trying offset 0x%lx....\n", ftell(fp_eimg));
+			fprintf(stderr, "Trying offset 0x%llx....\n", (unsigned long long)jtr_ftell64(fp_eimg));
 			a=(uint8_t)fgetc(fp_eimg);
 			b=(uint8_t)fgetc(fp_eimg);
 			if ((a != value_type[0]) || (b != value_type[1])) {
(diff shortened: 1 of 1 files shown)


solardiz (changes requested):
Thanks. I think you also need to upgrade these to an explicitly 64-bit type:

```
long int fp_before_aes = 0, fp_before_salt = 0;
```

`long` is commonly 32-bit on 32-bit systems and on Windows (including 64-bit Windows).

Also, please prefix the commit message title line with `bitlocker2john: ` (yet keep it within 72 chars, perhaps by re-wording the rest of it).

solardiz:
@visuve Have you tested these changes? Perhaps on input that wasn't correctly processed before, but with these changes is?

solardiz (changes requested):
There are still a few `fprintf` that use the `%lx` format on what's now the return value from `jtr_ftell64`. This is incorrect on some systems. One way to fix that for all would be to cast the value passed to `fprintf` to `(unsigned long long)` and to use `%llx`. BTW, I don't understand why we choose to print those offsets in hex - so maybe we should use `%llu` instead (edit: and drop the `0x` prefix).

visuve:
> @visuve Have you tested these changes? Perhaps on input that wasn't correctly processed before, but with these changes is?

@solardiz you could consider this as a first step towards migrating to fstat which should allow passing devices directly e.g. 
``bitlocker2john.exe -i \\.\PHYSICALDRIVE9``



> There are still a few `fprintf` that use the `%lx` format on what's now the return value from `jtr_ftell64`. This is incorrect on some systems. One way to fix that for all would be to cast the value passed to `fprintf` to `(unsigned long long)` and to use `%llx`. BTW, I don't understand why we choose to print those offsets in hex - so maybe we should use `%llu` instead (edit: and drop the `0x` prefix).

@solardiz ``%lld`` to be exact ;-) Personally I think the hex prefix makes sense in offsets that might be huge (e.g. memory or drive), because hex strings are usually easier to remember than long integers. Hence I would not change it. However I will fix them to int64.

solardiz (approved):
Those `fprintf` are still not exactly right because `jtr_ftell64` won't always return the same type as a `long long`. However, yes, I am OK treating these changes as temporary, and won't insist further.

solardiz:
Actually, no, can't ignore the wrongness of those `fprintf` formats even temporarily - our CI is unhappy:

```
bitlocker2john.c: In function 'rp_search_salt_aes':
bitlocker2john.c:109:19: error: format '%llx' expects argument of type 'long long unsigned int', but argument 3 has type 'int64_t {aka long int}' [-Werror=format=]
   fprintf(stderr, "Searching AES-CCM from 0x%llx\n", fp_before_aes);
                   ^
bitlocker2john.c:115:20: error: format '%llx' expects argument of type 'long long unsigned int', but argument 3 has type 'long int' [-Werror=format=]
    fprintf(stderr, "Trying offset 0x%llx....\n", jtr_ftell64(fp_eimg));
                    ^
bitlocker2john.c: In function 'process_encrypted_image':
bitlocker2john.c:177:20: error: format '%llx' expects argument of type 'long long unsigned int', but argument 3 has type 'long int' [-Werror=format=]
    fprintf(stderr, "\nSignature found at 0x%llx\n", (jtr_ftell64(fp_eimg) - i - 1));
                    ^
bitlocker2john.c:202:20: error: format '%llx' expects argument of type 'long long unsigned int', but argument 3 has type 'long int' [-Werror=format=]
    fprintf(stderr, "\nVMK entry found at 0x%llx\n", (jtr_ftell64(fp_eimg) - i));
                    ^
bitlocker2john.c:212:21: error: format '%llx' expects argument of type 'long long unsigned int', but argument 3 has type 'int64_t {aka long int}' [-Werror=format=]
     fprintf(stderr, "VMK not encrypted.. stored clear! (0x%llx)\n", fp_before_salt);
                     ^
bitlocker2john.c:214:21: error: format '%llx' expects argument of type 'long long unsigned int', but argument 3 has type 'int64_t {aka long int}' [-Werror=format=]
     fprintf(stderr, "VMK encrypted with TPM...not supported! (0x%llx)\n", fp_before_salt);
                     ^
bitlocker2john.c:216:21: error: format '%llx' expects argument of type 'long long unsigned int', but argument 3 has type 'int64_t {aka long int}' [-Werror=format=]
     fprintf(stderr, "VMK encrypted with Startup Key...not supported! (0x%llx)\n", fp_before_salt);
                     ^
bitlocker2john.c:218:21: error: format '%llx' expects argument of type 'long long unsigned int', but argument 3 has type 'int64_t {aka long int}' [-Werror=format=]
     fprintf(stderr, "\nVMK encrypted with Recovery Password found at 0x%llx\n", fp_before_salt);
                     ^
cc1: all warnings being treated as errors
```

This shows why I had suggested explicit casts to `(unsigned long long)`.

solardiz (changes requested):
Please implement explicit casts to `(unsigned long long)` on `fprintf`.

visuve:
@solardiz is there some readme how to build this locally? The src/configure halts on syntax errors when run with sh (I'm using WSL).

solardiz:
@visuve Can you show those syntax errors? It should just work.

visuve:
@solardiz ``configure: 32: Syntax error: newline unexpected (expecting ")")``

solardiz:
@visuve That `configure` error looks like an instance of #4736 - please look in there for possible workarounds, and we'd appreciate it if you help us fix that issue.

Does your bumping into that issue imply that the changes you're proposing in this PR are untested by you? Or if not, how did you test those? Thanks!

solardiz (approved):
The changes look good to me now, although I'd prefer to also hear of whether/how they've been tested.

visuve:
@solardiz I found the issue with the src/configure file. I set .gitconfig ``[core] autocrlf = false`` and now it works... :D

solardiz:
@visuve So, can you answer my question on whether/how you tested the changes in this PR? By the way, we already had to apply a further fix to them in #5104.

> I found the issue with the src/configure file. I set .gitconfig `[core] autocrlf = false` and now it works... :D

Great. As I recall, that's one of the workarounds mentioned in #4736. Maybe we need to have something like this by default - please feel free to send us a PR fixing #4736.

Issue #5084, opened 2022-02-20
incremental mask misbehaves for DEScrypt & LM opencl formats


AlekseyCherepanov:
I noticed a problem with `--mask='[a-z]?a' --min-length=1 --max-length=5` with `descrypt-opencl`. Just `?a` works as expected. `descrypt` seems ok.

```
$ echo AAAAAAAAAAAAA > t.pw
$ ./john/run/john --format=descrypt-opencl t.pw --mask='[a-z]?a' --min-length=1 --max-length=5 --session=s/s2
Device 1: Hawaii [AMD Radeon (TM) R9 390 Series]
Using default input encoding: UTF-8
Loaded 1 password hash (descrypt-opencl, traditional crypt(3) [DES OpenCL])
Note: descrypt-opencl format can't currently increment length from 1, using 2 instead
Build log: /tmp/OCL194688T2.cl:208:23: warning: unknown attribute 'max_constant_size' ignored
                                   __attribute__((max_constant_size((ACTIVE_PLACEHOLDER * 32 * ITER_COUNT))))
                                                  ^
1 warning generated.

LWS=128 GWS=262144 x95
Press 'q' or Ctrl-C to abort, 'h' for help, almost any other key for status
0g 0:00:00:00  (2) 0g/s 0p/s 0c/s 0C/s

0g 0:00:00:03 31.75% (2) (ETA: 23:40:51) 0g/s 210783Kp/s 210783Kc/s 210783KC/s maH9vU..maH9vU
0g 0:00:00:09 49.99% (5) (ETA: 23:41:00) 0g/s 216443Kp/s 216443Kc/s 216443KC/s gac>|U..gac>|U
0g 0:00:00:19 DONE (5) (2022-02-20 23:41) 0g/s 218021Kp/s 218021Kc/s 218021KC/s map~}..map~}
Warning: incremental mask started at length 2 - try the CPU format for shorter lengths.
Session completed. 
```

Notice `31.75% (2) [...] maH9vU..maH9vU`. It has `(2)` but candidates shown as current have length 6 (while `--max-length=5`). It is a problem.

There is `Note: [...]` about length. `--min-length=2` removes it, but does not solve the problem.

The problem is not cosmetic. It cannot crack 2 chars long password:
```
$ echo AAQg8DP4VZOHs > t.pw
$ ./john/run/john --format=descrypt-opencl t.pw --mask='[a-z]?a' --min-length=2 --max-length=5
[...]
[...]

PR #5103, opened 2022-04-06
Mask fixes

magnumripper:
Assorted bug fixes for internal mask. See https://github.com/openwall/john/issues/5084#issuecomment-1090346607

### DIFF
--- a/src/mask.c
+++ b/src/mask.c
@@ -2150,6 +2150,15 @@ void mask_init(struct db_main *db, char *unprocessed_mask)
 	mask_fmt = db->format;
 	mask_bench_index = 0;
 
+#if defined(HAVE_OPENCL) || defined(HAVE_ZTEX)
+	/* Disable internal mask */
+	if (options.req_int_cand_target == 0) {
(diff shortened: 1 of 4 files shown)


src/mask.c
```diff
@@ -2330,6 +2330,12 @@ void mask_init(struct db_main *db, char *unprocessed_mask)
 	if (format_cannot_reset && mask_increments_len && mask_skip_ranges[0] != -1) {
 		int inc_min =
 			mask_int_cand.int_cpu_mask_ctx->ranges[mask_max_skip_loc].pos + 1;
+		if (inc_min > options.eff_maxlength) {
+			if (john_main_process)
+				fprintf(stderr, "Error: %s cannot use internal mask under these premises,\n"
+				        "try using --mask-internal-target=0 option.\n", mask_fmt->params.label);
+			error();
+		}
```
magnumripper, line 2347: The reason we don't simply turn off internal mask here (with a warning), is that it's too late - thus the message and bailout.

PR #5117, opened 2022-04-24
minor changes: pfx's valid() and bug in fallback in enc_uc()

AlekseyCherepanov:
pfx: I removed `ishexlc` calls that repeat checks by `hexlenl`. I replaced `<` with `!=` for `saltlen` because `saltlen` is read from ciphertext, so there are no reasons for data and length to mismatch. Also I replaced `strrchr` check of the last field with plain check that the field is the last.

`enc_uc`: ascii fallback uses `| 0x20` to uppercase char, so it does not work. I guess it was copy-pasted from `enc_lc`. I replaced it with `^ 0x20`. Same change is possible for `enc_lc` to keep code similar. But I did not change it.

Locally I added test code for `enc_uc` in `valid()` of `raw-md5`:
```c
        char test[1024];
        enc_uc(test, 5, "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa", 600);
        printf("test: %s\n", test);
        exit(1);
```

```
$ ../run/john --test --format=raw-md5 --encoding=utf-8
Benchmarking: Raw-MD5 [MD5 128/128 SSE4.1 4x3, UTF-8]... test: aaaa
```

`--encoding=utf-8` is important here, otherwise fallback is not triggered. After fix, the example prints `AAAA`.

### DIFF
--- a/src/pfx_common_plug.c
+++ b/src/pfx_common_plug.c
@@ -11,7 +11,7 @@
 
 int pfx_valid(char *ciphertext, struct fmt_main *self)
 {
-	char *p = ciphertext, *ctcopy, *keeptr, *p2;
+	char *p = ciphertext, *ctcopy, *keeptr;
 	int mac_algo, saltlen, hashhex, extra;
 
 	if (strncasecmp(ciphertext, FORMAT_TAG, FORMAT_TAG_LENGTH))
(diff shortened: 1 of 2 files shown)


magnumripper (approved):
Looks sensible to me

AlekseyCherepanov:
This PR is ready. Thanks!

PR #5119, opened 2022-04-25
PDF format: Allow P as signed or unsigned int - 32-bit build bugfix

### LINKED ISSUE #5114: PDF format is failing at 32 bit
```
Testing: PDF [MD5 SHA2 RC4/AES 32/32]... (4xOMP) FAILED (cmp_all(32))
[...]
1 out of 401 tests have FAILED

 FAILED: -test-full=0 --format=cpu
```
I still need to reproduce this somehow to tag it as confirmed.

:1st_place_medal: :thumbsup:

```
Target CPU ......................................... armv7l NEON, 32-bit LE
Target OS .......................................... linux-gnueabihf
Legacy arch header ................................. arm32le.h
```

magnumripper:
Value can be -2147483648 .. 4294967295.  Closes #5114

### DIFF
--- a/src/pdf_fmt_plug.c
+++ b/src/pdf_fmt_plug.c
@@ -318,7 +318,7 @@ static void *get_salt(char *ciphertext)
 	p = strtokm(NULL, "*");
 	cs.length = atoi(p);
 	p = strtokm(NULL, "*");
-	cs.P = atoi(p);
+	cs.P = (int)strtoll(p, (char **)NULL, 10);
 	p = strtokm(NULL, "*");
 	cs.encrypt_metadata = atoi(p);
 	p = strtokm(NULL, "*");


src/pdf_fmt_plug.c
```diff
@@ -318,7 +318,7 @@ static void *get_salt(char *ciphertext)
 	p = strtokm(NULL, "*");
 	cs.length = atoi(p);
 	p = strtokm(NULL, "*");
-	cs.P = atoi(p);
+	cs.P = (int)strtoll(p, (char **)NULL, 10);
```
magnumripper, line 321: `atoi(p)` is equivalent to `(int)strtol(p, (char **)NULL, 10)` which is why this fix wasn't needed on systems where long is 64-bit.

Issue #5072, opened 2022-02-14
TrueCrypt formats: incorrect handling of keyfiles, works only for single ciphertext


AlekseyCherepanov:
Working on #5069 I did not realize real problem here:
```c
static unsigned char (*keyfiles_data)[MAX_KFILE_SZ];
[...]
static void init(struct fmt_main *self)
[...]
	keyfiles_data = mem_calloc(MAX_KEYFILES,
			sizeof(*keyfiles_data));
[...]
static void* get_salt(char *ciphertext)
[...]
	for (idx = 0; idx < s->nkeyfiles; idx++) {
[...]
		/* read this into keyfiles_data[idx] */
		fp = fopen(tpath, "rb");
```

So keyfiles for all ciphertexts are read into the same place in memory. So it would work for single ciphertext and would not work for two ciphertexts with different keyfiles (one ciphertext would become uncrackable). Data from keyfiles should be a part of salt.

(It is purely theoretical. I don't have samples.)

Fix would require switching onto dynamic salt. Or it is possible to exit with error when two ciphertexts with keyfiles are loaded.

Keyfiles in `john` do not seem popular. Their support in OpenCL format did not work due to allocation happening after loading (fixed in #5069).

AlekseyCherepanov:
There might be some related notes in #5068 in the future.

solardiz:
I don't fully understand this issue, but perhaps at least this must be implemented before our next release:

> it is possible to exit with error when two ciphertexts with keyfiles are loaded.

@AlekseyCherepanov Can you take care of this, please?

AlekseyCherepanov:
I can. Maybe I could do proper fix too.

AlekseyCherepanov:
`john-samples` has TrueCrypt volumes with keyfiles.

Also `truecrypt2john.py` has the following bug: single keyfile is not included into ciphertext. That's a matter of `>` vs `>=`. So I can produce in practice the original problem I described.

`TrueCrypt/TC-Volume` has 2 volumes (normal and hidden) that use different keyfiles. So I get 12 ciphertexts in 2 invocations of `truecrypt2john.py` (2 volumes x 3 algos x 2 sets of keyfiles).

Depending on the order of ciphertexts I can crack either one volume or the other.
```
$ rm run/john.pot
[...]

PR #5120, opened 2022-04-25
TrueCrypt fixes related to keyfiles

AlekseyCherepanov:
`truecrypt2john.py` got simple fix. Without the fix, the tool omits keyfiles when only one keyfile is specified on command line.

Formats got fix for #5072. Each ciphertext with keyfiles gets its own allocated memory that the keyfiles are read into. So I can crack both volumes in the example with `TrueCrypt/TC-Volume`.

I used `mem_alloc_tiny`, so allocated memory is freed at the end of john's work. The proper way is to use `dyna_salt`, but I did not check this way yet. I had idea to collect allocated pointers into a list and free them in `done()`, but it seems similar to `mem_alloc_tiny`.

I repeated the fix for opencl format. But I did not test it. It turned out that keyfiles do not work at all (on pocl).

Cpu format is a bit tested.

So at the moment, this PR is a preview of changes.

Is it normal to include pointers into the salt struct? Comments mention salt dupe-removal. So I know that I touch ground unknown to me.

### DIFF
--- a/run/truecrypt2john.py
+++ b/run/truecrypt2john.py
@@ -88,7 +88,7 @@ if __name__ == "__main__":
     options, remainder = parser.parse_args()
 
     keyfiles = []
-    if len(remainder) > 2:
+    if len(remainder) >= 2:
         keyfiles = remainder[1:]
 
     process_file(remainder[0], keyfiles, options)
--- a/src/truecrypt_fmt_plug.c
+++ b/src/truecrypt_fmt_plug.c
@@ -101,9 +101,6 @@ static unsigned char (*first_block_dec)[16];
 #define MAX_KFILE_SZ            1048576 /* 1 MB */
 #define MAX_KEYFILES            256
 
-// keyfile(s) data
-static unsigned char (*keyfiles_data)[MAX_KFILE_SZ];
-static int (*keyfiles_length);
 static int *cracked;
 
 static struct cust_salt {
@@ -125,6 +122,7 @@ static struct cust_salt {
 	int num_iterations;
 	int hash_type;
 	int nkeyfiles;
+	unsigned char kpool[KPOOL_SZ];
 } *psalt;
 
 static struct fmt_tests tests_ripemd160[] = {
@@ -171,10 +169,6 @@ static void init(struct fmt_main *self)
 			sizeof(*key_buffer));
 	first_block_dec = mem_calloc(self->params.max_keys_per_crypt,
 			sizeof(*first_block_dec));
-	keyfiles_data = mem_calloc(MAX_KEYFILES,
-			sizeof(*keyfiles_data));
-	keyfiles_length = mem_calloc(MAX_KEYFILES,
-			sizeof(int));
 	cracked = mem_calloc(sizeof(*cracked),
 			self->params.max_keys_per_crypt);
 	Twofish_initialise();
@@ -184,8 +178,6 @@ static void done(void)
 {
 	MEM_FREE(first_block_dec);
 	MEM_FREE(key_buffer);
-	MEM_FREE(keyfiles_data);
-	MEM_FREE(keyfiles_length);
 	MEM_FREE(cracked);
 }
 
@@ -295,12 +287,13 @@ static void* get_salt(char *ciphertext)
 {
 	static char buf[sizeof(struct cust_salt)+4];
 	struct cust_salt *s = (struct cust_salt *)mem_align(buf, 4);
-	unsigned int i;
 	char tpath[PATH_BUFFER_SIZE];
 	char *p, *q;
-	int idx;
+	int i, idx, kpool_idx;
 	FILE *fp;
 	size_t sz, len;
+	uint32_t crc;
+	unsigned char *keyfile_data;
 
 	memset(s, 0, sizeof(struct cust_salt));
 
@@ -361,7 +354,7 @@ static void* get_salt(char *ciphertext)
 		}
 		memcpy(tpath, p, len);
 		tpath[len] = '\0';
-		/* read this into keyfiles_data[idx] */
+		/* read this into keyfile_data */
 		fp = fopen(tpath, "rb");
 		if (!fp)
 			pexit("fopen %s", tpath);
@@ -370,6 +363,11 @@ static void* get_salt(char *ciphertext)
 			pexit("fseek");
 
 		sz = ftell(fp);
+
+		if (sz == 0) {
+			fclose(fp);
+			continue;
+		}
 
 		if (sz > MAX_KFILE_SZ) {
 			if (john_main_process)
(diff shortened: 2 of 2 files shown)


AlekseyCherepanov:
> Is it normal to include pointers into the salt struct? Comments mention salt dupe-removal. So I know that I touch ground unknown to me.

No, it is not normal. Example: the same ciphertext repeated twice gets loaded as `2 password hashes with 2 different salts`. (I guess pointers might be ok if they are same for same data.)

AlekseyCherepanov:
I implemented `dyna_salt` approach. It uses old global variable as temporary buffer, so I rewrote commit. Also I removed changes to opencl format.

Interestingly repeated ciphertext emitted message `Loaded 2 password hashes with no different salts`. After moving onto `dyna_salt`, I get this message: `Loaded 1 password hash`. So I changed path to a keyfile to include additional `../john-samples/` and I got again `Loaded 2 password hashes with no different salts`.

I did not test it fully yet.

`get_salt` with `dyna_salt` returns pointer to static buffer with pointer to actual structure. So I chose to make static variable with pointer to structure and return pointer to the variable without explicit buffer.

Code in `pkzip.c` for the same purpose:
```c
void *winzip_common_get_salt(char *ciphertext)
{
	uint64_t i;
	winzip_salt salt, *psalt;
	static unsigned char *ptr;
[...]
	if (!ptr)
		ptr = mem_alloc_tiny(sizeof(winzip_salt*),sizeof(winzip_salt*));
[...]
	memcpy(ptr, &psalt, sizeof(winzip_salt*));
	return (void*)ptr;
}
```

My code:
```c
static void* get_salt(char *ciphertext)
{
	static struct cust_salt salt, *s;
[...]
	s = &salt; /* assign every time */
[...]
	return &s;
}
```

Code in `pkzip.c` is more verbose. I dislike this verbosity. OTOH it allowed me to understand intention with buffer. Thinking in such terms, my code requires more steps in head to realize what is going on. So I am not sure that my code is better.

`assign every time` is an important point. I assign pointer to heap buffer to `s` later. With `s = &salt` in initialization of the static variable, the code was wrong.

I mix static and dynamic salts and everything works fine. At least ASan does not report leaks or attempts to `free()` static memory.

I think order of keyfiles does not affect result, so after more experiments I am going to add sorting to allow dupe-removal for such ciphertexts.

AlekseyCherepanov:
`pkzip.h`:
```c
typedef struct winzip_salt_t {
[...]
	unsigned char datablob[1];
} winzip_salt;
```

I use explicit flexible array member for `keyfiles_length`, i.e. I don't specify size. This feature is a part of C99. Hm, [Wikipedia article](https://en.wikipedia.org/wiki/Flexible_array_member) suggests that my style is not the best. Hm, I guess I need to review my use of this bit too: `s->dsalt.salt_cmp_offset = sizeof(s->dsalt);`

BTW I guess `pkzip.c` allocates 1 byte more than needed for salt.
```c
	psalt = (winzip_salt*)mem_calloc(1, sizeof(winzip_salt) + salt.comp_len);
```

But I need more than just one array: I have array `keyfiles_length` with variable length and then there is data from keyfiles that I store after the structure. With fixed length for `keyfiles_length`, the data could be part of structure. It would mean addition of 256 `int`s per salt.

AlekseyCherepanov:
`apply_keyfiles()` computes `kpool` from keyfiles and then mixes it into candidate. `kpool` can be computed loading ciphertexts. So `kpool` is 64 bytes big and it can replace all keyfiles.

`kpool` could be precomputed by `truecrypt2john.py` and embedded into ciphertext instead of filenames.

magnumripper:
I am unable to context-switch into this now but I did review your current PR and saw nothing that struck me as incorrect.

AlekseyCherepanov:
JFYI I found possible problem in my current code.

```c
static struct cust_salt {
[...]
	// Tightly packed blob with keyfiles' data is put right after the end.
	int keyfiles_length[];
} *psalt;
```

I expect such layout:
```
[main struct][keyfiles_length][blob with keyfiles]
             ^                ^ sizeof() + actual size of keyfiles_length[]
              \ sizeof(struct cust_salt)
```

The [Wikipedia article](https://en.wikipedia.org/wiki/Flexible_array_member):
> It is common to allocate sizeof(struct) + array_len*sizeof(array element) bytes.
> 
> This is not wrong, but it may allocate a few more bytes than necessary: the compiler may be re-purposing some of the padding that is included in sizeof(struct). Should this be a concern, macros are available[3] to compute the minimum size while ensuring that the compiler's padding is not disrupted.
> 
> As the array may start in the padding before the end of the structure, its content should always be accessed via indexing (arr[i]) or offsetof, not sizeof.

So I should expect such layout:
```
[main struct][keyfiles_length][blob with keyfiles]
             ^ offsetof(struct cust_salt, keyfiles_length)
```

I use `keyfiles_length[]` through field. So it may start earlier than `sizeof()` if `offsetof` and `sizeof` are different. Then using max offset for the blob leaves uninitialized memory between `keyfiles_length[]` and the blob. But this memory is hashed by `fmt_default_dyna_salt_hash`. So salts valid for dupe-removal might or might not merge depending on contents of memory.

I should use `mem_calloc` instead of `mem_alloc` to allocate the salt.

But I implemented `kpool` approach already and I like it.

AlekseyCherepanov:
I mean my actual use would look the following way in case `offsetof != sizeof`:
```
               v sizeof         vv not initialized
(main struct[  )keyfiles_length]  [blob with keyfiles]
            ^ offsetof            ^ sizeof + size of keyfiles_length[]
```

AlekseyCherepanov:
I rewrote commit reverting back from `dyna_salt`. I replaced keyfiles' data with precomputed `kpool`. So now there is additional 64 bytes in the salt structure that are not used without keyfiles.

> I think order of keyfiles does not affect result, so after more experiments I am going to add sorting to allow dupe-removal for such ciphertexts.

Precomputed `kpool` in salt means that ciphertexts with the same keyfiles in different order would be matched as identical.

Also I set `nkeyfiles` to 1 in the end. `nkeyfiles` is used as boolean value in other places. There can be sets of keyfiles that produce identical `kpool` contents. For instance, empty keyfile does not change `kpool` contents, so it can be added to any set. (Though such example should be unlikely in real life.)

I moved `keyfiles -> kpool` conversion from `apply_keyfiles` to `get_salt`. So it should speed up cracking.

New code can crack `TC-Normal-Volume` from `john-samples`. Normal and hidden volumes can be cracked in one session. So it solves #5072 for cpu formats.

I could not crack `TC-STD-Volume` sample. But older version without my changes could not crack it too. I'll solve it later.

This PR is ready. Thanks!

solardiz (approved):
I cannot properly review this without looking into the underlying issues myself first, but nothing in here looks necessarily wrong and I guess @AlekseyCherepanov knows what he's doing here.

@magnumripper Do we just merge this PR, or do you want to review first?

magnumripper:
> I cannot properly review this without looking into the underlying issues myself first, but nothing in here looks necessarily wrong and I guess @AlekseyCherepanov knows what he's doing here.
> 
> @magnumripper Do we just merge this PR, or do you want to review first?

Let's trust Aleksey

PR #5129, opened 2022-05-08
ztex: fix crash on failed open, improve warning, add -g to Makefile.in

### LINKED ISSUE #5125: ztex formats: segfault after failed opening of /dev/bus/...
I re-installed my system, so I did not have settings for usb devices to use ztex boards. Attempt to use john resulted into segfault.

```
$ ../run/john --format=descrypt-ztex --test
Segmentation fault
```

Output from `strace`:
```
openat(AT_FDCWD, "/dev/bus/usb/007/002", O_RDWR|O_CLOEXEC) = -1 EACCES (Permission denied)
--- SIGSEGV {si_signo=SIGSEGV, si_code=SEGV_MAPERR, si_addr=0x46} ---
```

I would prefer to see an informative message from john about problem with permissions and maybe a link to `doc/README-ZTEX`.

AlekseyCherepanov:
Closes #5125.

Now it shows such warning:
```
$ ../run/john --test --format=descrypt-ztex
Warning: unable to open 2 board(s): insufficient permissions (see doc/README-ZTEX) or another instance of john running.
No ZTEX devices found.
```

The warning without `(see doc/README-ZTEX) ` was in place already. It was not reached though.

### DIFF
--- a/src/ztex/Makefile.in
+++ b/src/ztex/Makefile.in
@@ -5,7 +5,7 @@
 # modification, are permitted.
 
 CC = gcc
-CFLAGS = -c -Wall -O2
+CFLAGS = -c -Wall -O2 -g
 CFLAGS_TEST = -O
 #LD = ld
 RM = /bin/rm -f
--- a/src/ztex/pkt_comm/Makefile.in
+++ b/src/ztex/pkt_comm/Makefile.in
@@ -5,7 +5,7 @@
 # modification, are permitted.
 
 CC = gcc
-CFLAGS = -c -Wall -O2
+CFLAGS = -c -Wall -O2 -g
 
 OBJS = pkt_comm.o word_gen.o word_list.o cmp_config.o inpkt.o init_data.o
 
--- a/src/ztex/ztex.c
+++ b/src/ztex/ztex.c
@@ -494,6 +494,8 @@ int ztex_scan_new_devices(struct ztex_dev_list *new_dev_list,
 				break;
 			}
 		}
+		if (!ztex_dev)
+			continue;
 
 		// found new device
 		if (ZTEX_DEBUG) printf("found: SN %s productId: %d.%d\n",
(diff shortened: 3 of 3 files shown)


solardiz:
> ztex: add -g to Makefile.in

I'm not sure this is right. In our main `Makefile.in`, we have `-g` in `LDFLAGS`, but under `ztex` you're instead adding it to `CFLAGS`. Let's be consistent?

Edit: also, `LDFLAGS` isn't relevant under `ztex`, right?
Aleksey, perhaps you should describe the debugging issue you're fixing by this change.

solardiz:
@AlekseyCherepanov Should we wait for you to address my comment above?

AlekseyCherepanov:
I described the problem here:
https://github.com/openwall/john/issues/5125#issuecomment-1115060338

Lack of debugging info in ztex files only is not convenient.

`src/Makefile.in`:
```Makefile
# Default build is always using -g (stripped when 'make install') but this
# target turns off optimizations for better debugging, and defines DEBUG
# so code can adopt.
debug:
	$(MAKE) default CFLAGS="-DDEBUG $(CFLAGS) @O_DEBUG@" CFLAGS_MAIN="-DDEBUG $(CFLAGS_MAIN) @O_DEBUG@"
```

So I thought `-g` everywhere is ok. Is not it?

> Aleksey, perhaps you should describe the debugging issue you're fixing by this change.

Do you mean I should change the commit message?

AlekseyCherepanov:
> @AlekseyCherepanov Should we wait for you to address my comment above?

I intended to reply. Thanks for waiting!

solardiz:
OK, I took a closer look - while we don't have `-g` in `CLFAGS` in top-level `Makefile.in` explicitly, it appears to be getting added by autoconf. Anyway, I think the real issue is `ztex` subdirectory is inconsistent in that it overrides `CFLAGS` - unlike `aes` and others.

For now, I'm fine accepting this PR as-is, but longer term we might want to have `ztex` use consistent flags with the rest of the tree, not just add `-g` to there.

solardiz (approved):

Issue #5000, opened 2022-01-13
--wordlist=fifo misses cracks with --fork


AlekseyCherepanov:
#4921 introduced ability to use a fifo as argument for `--wordlist=`. `--stdin`/`--pipe` did not allow use of `--fork`. `--wordlist=` allows `--fork=`. But with a fifo, fork works badly.

Setup:
```
main$ mkfifo fifo

main$ seq 100 | sed -e 's/^/$0$/' > t.pw
```

Test without `--fork`: there should be 100 cracks.
```
main$ ./run/john --wordlist=fifo t.pw --pot=t.pot

other$ seq 100 > fifo
```

Test with `--fork`:
```
main$ rm t.pot

main$ ./run/john --wordlist=fifo t.pw --pot=t.pot --fork=2

other$ seq 100 > fifo
```

There are 50 cracks only instead of 100. All cracks are either odd or even numbers. So I guess one process consumes the whole input and skips half of candidates expecting other process to try them.

magnumripper:
I wonder if this is a problem inherent of fifos rather than much anything else. Perhaps we should just disable it for forked runs

solardiz:
Of course, we should disallow this combination. Perhaps right after setting `file_is_fifo`, check if we're running with `--fork` and print an error and `error();` out. @AlekseyCherepanov You want to send us a PR fixing this? Thanks!

AlekseyCherepanov:
`--loopback` can use fifo as argument. But it opens the fifo twice.
```
$ ./run/john --loopback=fifo --log-stderr --rules=':' --stdout
Using default input encoding: UTF-8
[... waits; entered in other shell: echo :asdf > fifo ]
0:00:00:00 Command line: ./run/john --loopback=fifo --log-stderr --rules=: --stdout 
0:00:00:00 - UTF-8 input encoding enabled
0:00:00:00 - Passwords will be stored UTF-8 encoded in .pot file
0:00:00:00 Proceeding with loopback mode
Permutation rules: :
0:00:00:00 - Rules: :
[... waits; entered in other shell: echo :qwer > fifo ]
0:00:00:00 - Loopback pot FIFO: fifo
0:00:00:00 - No stacked rules
0:00:00:00 - 1 preprocessed word mangling rules
0:00:00:00 Enabling duplicate candidate password suppressor
Enabling duplicate candidate password suppressor
0:00:00:00 - Rule #1: '' accepted
qwer
[...]

PR #5131, opened 2022-05-10
prohibit FIFOs for wordlist+fork and as pot files

AlekseyCherepanov:
- The first commit addresses #5000 as @solardiz suggested. Opening of file is done before `fstat`, so john waits other end of fifo to be opened and only then exits with error. I guess it is ok because user would be going to provide input, if we protect from intentional use.
- The second commit prohibits use of fifos as .pot files. Fifo as argument to `--loopback=` would be opened twice. So it would hang even with intentional use.
- Also I noticed that several warnings print 'name' without `path_expand()`. But 2 cases had `path_expand()` called earlier. The 2 fixed cases are not such.

I cannot say that I tested it well. I think there might be some consequences I could not foresee.

### DIFF
--- a/src/loader.c
+++ b/src/loader.c
@@ -225,7 +225,8 @@ static void read_file(struct db_main *db, char *name, int flags,
 	}
 
 	if (!(file = fopen(path_expand(name), "r"))) {
-		if ((flags & RF_ALLOW_MISSING) && errno == ENOENT) return;
+		if ((flags & RF_ALLOW_MISSING) && errno == ENOENT)
+			return;
 		pexit("fopen: %s", path_expand(name));
 	}
 
--- a/src/loader.c
+++ b/src/loader.c
@@ -247,15 +247,13 @@ static void read_file(struct db_main *db, char *name, int flags,
 			     options.input_enc == UTF_8)) {
 				if (!valid_utf8((UTF8*)u8check)) {
 					warn_enc = 0;
-					fprintf(stderr, "Warning: invalid UTF-8"
-					        " seen reading %s\n", name);
+					fprintf(stderr, "Warning: invalid UTF-8 seen reading %s\n", path_expand(name));
 				}
 			} else if (options.input_enc != UTF_8 &&
 			           (line != line_buf ||
 			            valid_utf8((UTF8*)u8check) > 1)) {
 				warn_enc = 0;
-				fprintf(stderr, "Warning: UTF-8 seen reading "
-				        "%s\n", name);
+				fprintf(stderr, "Warning: UTF-8 seen reading %s\n", path_expand(name));
 			}
 		}
 		process_line(db, line);
--- a/src/loader.c
+++ b/src/loader.c
@@ -209,13 +209,19 @@ static void read_file(struct db_main *db, char *name, int flags,
 	warn_enc = (john_main_process && (options.target_enc != ENC_RAW) &&
 	            cfg_get_bool(SECTION_OPTIONS, NULL, "WarnEncoding", 0));
 
-	if (flags & RF_ALLOW_DIR) {
-		if (stat(name, &file_stat)) {
-			if (flags & RF_ALLOW_MISSING)
-				if (errno == ENOENT) return;
-			pexit("stat: %s", path_expand(name));
-		} else
-			if (S_ISDIR(file_stat.st_mode)) return;
+	if (stat(name, &file_stat)) {
+		if ((flags & RF_ALLOW_MISSING) && errno == ENOENT)
+			return;
+		pexit("stat: %s", path_expand(name));
+	}
+
+	if ((flags & RF_ALLOW_DIR) && S_ISDIR(file_stat.st_mode))
+		return;
+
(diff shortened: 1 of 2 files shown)


src/loader.c
```diff
@@ -209,13 +209,20 @@ static void read_file(struct db_main *db, char *name, int flags,
...
-	if (flags & RF_ALLOW_DIR) {
-		if (stat(name, &file_stat)) {
-			if (flags & RF_ALLOW_MISSING)
-				if (errno == ENOENT) return;
-			pexit("stat: %s", path_expand(name));
-		} else
-			if (S_ISDIR(file_stat.st_mode)) return;
+	if (stat(name, &file_stat)) {
+		if (flags & RF_ALLOW_MISSING)
+			if (errno == ENOENT) return;
```
magnumripper: Our current code style says we should put `return;` on next line. Same below.

src/loader.c
```diff
@@ -209,13 +209,20 @@ static void read_file(struct db_main *db, char *name, int flags,
...
+		pexit("stat: %s", path_expand(name));
+	}
+
+	if (flags & RF_ALLOW_DIR)
+		if (S_ISDIR(file_stat.st_mode)) return;
+
+	if (ldr_in_pot && S_ISFIFO(file_stat.st_mode)) {
+		if (john_main_process)
+			fprintf(stderr, "Error, cannot use FIFO as pot"
+				" file (%s)\n", path_expand(name));
```
magnumripper: Our current code style recommends to avoid breaking output strings even if it exceeds normal column width. Same for several other changes in this ~commit~ PR.

magnumripper (approved):

magnumripper:
> I cannot say that I tested it well. I think there might be some consequences I could not foresee.

The logic looks good to me

AlekseyCherepanov:
I fixed the style problems near my changes.

I used parentheses in `((flags & RF_ALLOW_DIR) && S_ISDIR(file_stat.st_mode))` because there was code with parentheses below already. But they are not necessary because `&&` has lower precedence than `&`.

Both styles are present in john:
```
$ grep '([a-zA-Z_0-9]\+ & [a-zA-Z_0-9]\+) &&' src/*.c
src/bench.c:		if ((flags & 0x200) && pass >= 2)
src/external.c:	if ((ext_flags & EXT_REQ_GENERATE) && !f_generate) {
[...]

$ grep '([a-zA-Z_0-9]\+ & [a-zA-Z_0-9]\+ &&' src/*.c
src/simd-intrinsics.c:	if (SSEi_flags & SSEi_REVERSE_STEPS && !(SSEi_flags & SSEi_CRYPT_SHA224))
src/simd-intrinsics.c:	if (SSEi_flags & SSEi_REVERSE_STEPS && !(SSEi_flags & SSEi_CRYPT_SHA384))
src/wpapcap2john.c:			if (l & 1 && l < 10)
```

src/wordlist.c
```diff
@@ -574,6 +575,15 @@ void do_wordlist_crack(struct db_main *db, const char *name, int rules)
 
 		file_is_fifo = ((st.st_mode & S_IFMT) == S_IFIFO);
 
+#if OS_FORK
+		if (options.fork && file_is_fifo) {
+			if (john_main_process)
+				fprintf(stderr, "Error, cannot use --fork "
+					"with FIFO as dictionary file.\n");
```
solardiz: We normally say "wordlist", not "dictionary", and no need to wrap source code lines shorter than 132.

solardiz (approved):
I've just made a nitpick comment.

Also, I wouldn't introduce the coding style changes as a separate commit - when those changes are limited to places touched by the previous commits, they can be right in those commits. However, this is also a nitpick - not strictly required to be "fixed", we can also merge as-is.

src/wordlist.c
AlekseyCherepanov: Oops, I forgot that I wrapped lines in `wordlist.c`.

I replicated 'dictionary' from other warning. Now both places are fixed.

AlekseyCherepanov:
I rewrote commits to include changes of style for my changes and to near places unrelated to the changes.

Also I did not split lines shorter than 132. Example, new vs previous variants:
```c
				fprintf(stderr, "Warning: UTF-8 seen reading %s\n", path_expand(name));
// older
				fprintf(stderr, "Warning: UTF-8 seen reading %s\n",
					path_expand(name));
// oldest
				fprintf(stderr, "Warning: UTF-8 seen reading "
				        "%s\n", name);
```
(I noticed in "older" that I put tab to align arguments. It was not intentional.)

AlekseyCherepanov:
> AlekseyCherepanov force-pushed the fix26 branch from e6661da to f988634 2 hours ago
> Compare

Pretty convenient: after `git push -f` github shows a line with event. `force-pushed` and `Compare` button are links to comparisons between states, so it is possible to check cumulative change in addition to or instead of reviewing separate patches.

AlekseyCherepanov:
I changed ` (%s)` to `: %s` in error message. I rewrote commits to include changes to code style only in my changes. Change for code style and `dictionary`->`wordlist` substitution in other places were extracted into a separate commit.

PR #5132, opened 2022-05-11
Update kirbi2john from nidem/kerberoast

daniruiz:
This python script was originally based on a previous version of the kirbi2john.py script from https://github.com/nidem/kerberoast but is now longer working. This changes are just an adaptation of the latest script version from the original repo

### DIFF
--- a/run/kirbi2john.py
+++ b/run/kirbi2john.py
@@ -19,7 +19,6 @@
 # limitations under the License
 
 from pyasn1.codec.ber import decoder
-import glob
 
 
 def extract_ticket_from_kirbi(filename):
@@ -45,17 +44,16 @@ if __name__ == '__main__':
     parser.add_argument('-o', dest='crack_file', metavar='crack_file', type=argparse.FileType('w'), default=sys.stdout, nargs='?',
                     help='File to save crackable output to (default is stdout')
     parser.add_argument('files', nargs='+', metavar='file.kirbi', type=str,
-                    help='File name to crack. Use asterisk \'*\' for many files.\n Files are exported with mimikatz or from extracttgsrepfrompcap.py')
+                    help='File name to crack.\n Files are exported with mimikatz or from extracttgsrepfrompcap.py')
 
     args = parser.parse_args()
 
     enctickets = []
 
-    for path in args.files:
-        for filename in glob.glob(path):
-            et = extract_ticket_from_kirbi(filename)
-            if et:
-                enctickets.append((et,filename))
+    for filename in args.files:
+        et = extract_ticket_from_kirbi(filename)
+        if et:
+            enctickets.append((et,filename))
 
     #out=open("crack_file","wb")
     for et in enctickets:
--- a/run/kirbi2john.py
+++ b/run/kirbi2john.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 # Based on the Kerberoast script from Tim Medin to extract the Kerberos tickets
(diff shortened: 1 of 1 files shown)


solardiz:
Thanks @daniruiz. What made the script no longer work? Also, is it compatible with Python 2 or 3 or both? Before and after your changes?

daniruiz:
The old script was python2 only and the new one is python3 only

The update also fixes an issue with the hash output not being recognized by john https://github.com/nidem/kerberoast/issues/19

Also, here's the original kirbi2john.py script from the kerberoast project. I just included the functions that were called from kerberos.py in the same script
https://github.com/nidem/kerberoast/blob/master/kirbi2john.py
https://github.com/nidem/kerberoast/blob/master/kerberos.py#L182

solardiz:
> The old script was python2 only and the new one is python3 only

If so, please update the `#!/usr/bin/env python` line to use `python3`.

Also, please wrap the commit message lines at 75 chars. And fix the typo in "now longer" (should be "no longer").

> The update also fixes an issue with the hash output not being recognized by john [nidem/kerberoast#19](https://github.com/nidem/kerberoast/issues/19)
> 
> Also, here's the original kirbi2john.py script from the kerberoast project. I just included the functions that were called from kerberos.py in the same script https://github.com/nidem/kerberoast/blob/master/kirbi2john.py https://github.com/nidem/kerberoast/blob/master/kerberos.py#L182

This makes me wonder why not take the script from there almost literally, just replacing the `import kerberos` line with two(?) tiny functions we use from there. Would make further updates easier?

However, I'm fine keeping your changes the way they are now, assuming you've tested this.

@exploide @kholia Do you happen to have any comments before we merge this? Thanks!

exploide:
I second the comments from @solardiz.

But I think at first it must be decided whether to include the script from nidem/kerberoast as close as possible, or if we want to deviate form it. Looking at the history there, the script seems to originate from a copy of john in 2015. So it's going back and forth now. (That's the problem of not having a single source of truth...)

If we want to deviate I propose a change to the description of the `files` parameter, which relates to the lines:

```python
for path in args.files:
    for filename in glob.glob(path):
```

Instead of relying to shell globbing (what we should) this accept `'*'` as an argument and resolves it on its own. I propose removing the inner loop, leaving just `for filename in args.files`. And change the parameter's description and remove that part about `'*'`. We neither need to explain to the user how shell globbing works, nor do we need to implement it by accepting the quoted asterisk. It already works when giving an unquoted asterisk because the shell does the globbing for us.

(Side node, I have not run the tool on any test files.)

solardiz:
Oh, sure we need to get rid of `glob`. That's an unexpected misfeature, inconsistent with the rest of our scripts. It can be especially nasty if a filename happens to include a literal shell globbing character in it.

daniruiz:
Agreed with removing glob, that should be handled by the shell :+1: 
I've made the script now look more similar to the original one, adding the 2 extra functions on top. Also removed the glob function. So, now the script is easier to maintain if the original project is updated but does include some changes too

solardiz:
With the glob removal commit, please also remove this from the usage help message: `Use asterisk \'*\' for many files.`

And please prefix that commit message with `kirbi2john.py: `. In fact, the first commit message can also have `kirbi2john.py: Update from nidem/kerberoast` as its first line, but that change is optional since the message does mention the script.

daniruiz:
Nice catch @solardiz. I didn't read that part.
Fixed and pushed

exploide (approved):

PR #5137, opened 2022-05-15
Tweaked mask

e2002e:
Hi, this is a conclusion of all my attempts to include a special technique into john the ripper.
mask.c has two main changes:
 - Alternate on multiple lengths
 - Deep mangling

I thought there is no harm in doing this appart from a potential loss of speed (which i'm unsure of).
These modifications allow the program to find passwords more quickly, because of all the lengths treated in the same time and because the state evolution. I am sorry for all those trailing commits. These are from my attempts at a generator I told solar about, but couldn't complete. To get back to the matter, mask is the fastest, but lengths increase one by one, and from the time a length is reached where a lot of time is needed to change a span of the word, this module will have much more effectiveness even with a loss of speed. I'm quite sure that it only slows a little bit down if slowing down at all.

Since this is just a tweak to -mask, it'll work with opencl as expected.
I cracked a handfull of passwords belonging to a list of 'unfound hashes' with it. It is a good way to work on left-asides.


solardiz:
@e2002e I'm happy that you're having fun with this, but we wouldn't even consider merging a hack/tweak that breaks existing functionality (mask mode's steady length increases should remain an option), has lots of irrelevant commits/changes, and isn't illustrated by an example showing its usefulness (vague wording like "I cracked a handfull of passwords belonging to a list of 'unfound hashes'" isn't enough - you need to show this worked better than other modes would have in the same conditions).

Separately, I think we also want to keep our cracking modes mostly as-is until after our next release. I was thinking of and experimented with improving incremental mode (upgrading it from 3-grams to 4-grams), but got mixed results - better on some tests, worse on some others - and even if it's better on average, I decided to postpone it for now, hoping to arrive at something more consistently better later. There would have to be a really compelling reason to change anything in incremental or mask now, and I just don't see that reason.

PR #5140, opened 2022-05-18
Fixes for generic build

### LINKED ISSUE #5135: "make generic" broken
f2a53fa3cbae8341f02a2f76caf209ed1bf453be broke `make -f Makefile.legacy generic`. This fixes it:

```diff
+++ b/src/bench.c
@@ -382,6 +382,7 @@ char *benchmark_format(struct fmt_main *format, int salts,
 #ifdef BENCH_BUILD
        const char *where;
        if ((where = fmt_self_test(format, test_db))) {
+               static char s_error[128];
                snprintf(s_error, sizeof(s_error), "FAILED (%s)\n", where);
                return s_error;
        }
```

Found while trying to build for an arch not yet recognized by our `configure`... to fix which we need to make our `configure.ac` friendly to newer autoconf.

magnumripper:
Closes #5135

This works on 'super' but I'm not sure about the canonicality of the feature test macros in `wordlist.c`.

### DIFF
--- a/src/bench.c
+++ b/src/bench.c
@@ -382,6 +382,7 @@ char *benchmark_format(struct fmt_main *format, int salts,
 #ifdef BENCH_BUILD
 	const char *where;
 	if ((where = fmt_self_test(format, test_db))) {
+		static char s_error[128];
 		snprintf(s_error, sizeof(s_error), "FAILED (%s)\n", where);
 		return s_error;
 	}
@@ -676,7 +677,6 @@ void gather_results(struct bench_results *results)
 int benchmark_all(void)
 {
 	struct fmt_main *format;
-	int i;
 #if defined(HAVE_OPENCL)
 	char s_gpu[16 * MAX_GPU_DEVICES] = "";
 	char s_gpu1[16 * MAX_GPU_DEVICES] = "";
@@ -685,6 +685,7 @@ int benchmark_all(void)
 	const char *s_gpu1 = "";
 #endif
 #ifndef BENCH_BUILD
+	int i;
 	unsigned int loop_fail = 0, loop_total = 1;
 	int nvidia_mem = 0;
 #endif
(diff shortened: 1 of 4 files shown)


src/bench.c
```diff
@@ -382,6 +382,8 @@ char *benchmark_format(struct fmt_main *format, int salts,
 #ifdef BENCH_BUILD
 	const char *where;
 	if ((where = fmt_self_test(format, test_db))) {
+		static char s_error[128];
+
```
solardiz: Nitpick: I wouldn't be introducing an empty line here.

src/wordlist.c
```diff
@@ -14,13 +14,14 @@
 #include "autoconfig.h"
 #else
 #ifndef sparc
```
solardiz: The `#ifndef sparc` is probably a bug - this should have nothing to do with the arch, and even if it had something to do with an OS typical for that arch then our change from `_POSIX_SOURCE` to `_XOPEN_SOURCE` may affect that. Should we review its history, and try dropping it (the `#ifndef sparc` and its `#endif`)?

solardiz (approved):
Looks OK to me, ideally we'd also test on non-Linux. Also, instead of `#undef _XOPEN_SOURCE` we could use `#ifndef _XOPEN_SOURCE`, so that we don't override a possible tree-wide setting of this macro to some value (e.g., 500).

src/wordlist.c
magnumripper: It comes from Jim's 98c144450d in 2014
```diff
+/* NOTE, _POSIX_SOURCE fuk's up solaris 32 bit 64 bit processing!!! */
+#ifndef sparc
 #define _POSIX_SOURCE /* for fileno(3) */
+#endif
```

magnumripper: I changed `#ifndef sparc` to `#ifndef _XOPEN_SOURCE`

PR #5142, opened 2022-05-19
pem2john.py: allow hashing only for sha1 encoding

pradkrish:
partial fix for #4834

### DIFF
--- a/run/pem2john.py
+++ b/run/pem2john.py
@@ -66,7 +66,7 @@ def unwrap_pkcs8_data(blob):
         if data["encryption_algorithm"]["algorithm"] != "pbes2":
             sys.stderr.write("[%s] encryption_algorithm <%s> is not supported currently!\n" %
                              (sys.argv[0], data["encryption_algorithm"]["algorithm"]))
-            return False
+            return True
 
         # encryption data
         encrypted_data = data["encrypted_data"]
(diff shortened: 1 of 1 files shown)


solardiz (changes requested):
I think the newly added `if` block for `prf` should be consistent with the existing one for `kdf`. As this PR currently is, the added block uses different style in the `if` line, in the error message text (I suggest `"[%s] prf algorithm <%s> is not supported currently!\n"`), and the return value (`False` vs. `True`).

```python
        if kdf["algorithm"] != "pbkdf2":
            sys.stderr.write("[%s] kdf algorithm <%s> is not supported currently!\n" %
                             (sys.argv[0], kdf["algorithm"]))
            return False
        kdf_params = kdf["parameters"]
        if kdf_params['prf']['algorithm'] is not 'sha1':
            sys.stderr.write("[%s] is generated using %s and it is currently not supported\n" %
                             (sys.argv[1], kdf_params['prf']['algorithm']))
            return True
```

Also, the commit message is weird. I suggest:

```
pem2john.py: Check "prf algorithm" and reject anything other than "sha1"
```

pradkrish:
Thanks for the review. What should the return value be? Returning `False` may emit another error message from the calling function `[%s] is this really a private key in PKCS #8 format?\n`, which is why I went for a return value of `True`

solardiz:
> Returning `False` may emit another error message from the calling function `[%s] is this really a private key in PKCS #8 format?\n`, which is why I went for a return value of `True`

Oh, good point. You're right. Then the previous two instances of returning `False` (for other unsupported algorithms in correctly formatted input) should be changed to `True`. Please do that. Thanks!

solardiz:
Looks almost ready. Nitpick: there are still minor discrepancies on the below two lines compared to similar lines before:

```python
            sys.stderr.write("[%s] prf algorithm <%s> is not supported currently\n" %
                             (sys.argv[0], kdf_params['prf']['algorithm']))
```

The two similar blocks before end their messages with an exclamation point, and would use double quotes in `kdf_params["prf"]["algorithm"]`. Good that you spotted the wrong `argv` index in your previous revision - something I had missed.

PR #5149, opened 2022-05-28
Wordlist mode: Avoid PerRuleStats infinite loop with dupe suppressor

### LINKED ISSUE #5148: `PerRuleStats = Y` hangs due to dupe suppression
I tried `PerRuleStats = Y` a bit. By default, it hangs. Disabling dupe suppressor helps.

Preparations:
```
$ printf '.include <john.conf>\n[Options]\nPerRuleStats = Y\n' > stats.conf
$ echo '$NT$aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa' > nt.pw
$ echo test > test.txt
```

Short attack with `PerRuleStats = Y` hangs after the following output:
```
$ ../run/john nt.pw --wordlist=test.txt --rules=':' --config=stats.conf
Warning: detected hash type "NT", but the string is also recognized as "NT-long"
Use the "--format=NT-long" option to force loading these as that type instead
Warning: detected hash type "NT", but the string is also recognized as "NT-opencl"
Use the "--format=NT-opencl" option to force loading these as that type instead
Loaded 1 password hash (NT [MD4 256/256 AVX2 8x3])
Enabling duplicate candidate password suppressor
```

It works well with `--dupe-suppression=0`.

John is fresh:
```
John the Ripper 1.9.0-jumbo-1+bleeding-78e8d466c 2022-05-24 12:09:11 +0200 [linux-gnu 64-bit x86_64 AVX2 AC]
```

solardiz:
Also ignore (force-disable) PerRuleStats with mask or when not logging.

Fixes #5148

### DIFF
--- a/src/cracker.c
+++ b/src/cracker.c
@@ -101,7 +101,6 @@ int64_t crk_pot_pos;
 void (*crk_fix_state)(void);
 int (*crk_process_key)(char *key);
 
-static int process_key(char *key);
 static int process_key_stack_rules(char *key);
 
 /* Expose max_keys_per_crypt to the world (needed in recovery.c) */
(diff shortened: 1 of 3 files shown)


AlekseyCherepanov:
I tested. It seems ok.

Issue #5157, opened 2022-06-05
crashes under ASan from fuzzing: PBKDF2-HMAC-MD5, SSH, hdaa, scrypt, and others


AlekseyCherepanov:
`ssh`:
```
$ perl -le 'print q{$sshng$1$16$570F498F6FF732775EE38648130F600D$10000$} . "00" x 10000' > t.pw
$ ./john/run/john t.pw
[...]
==395326==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x625000002188 at pc 0x562401e4a8c2 bp 0x7ffd00791a40 sp 0x7ffd00791a38
READ of size 1 at 0x625000002188 thread T0
    #0 0x562401e4a8c1 in ssh_get_salt /home/user/john/src/ssh_common_plug.c:121
    #1 0x562401ee374b in ldr_load_pw_line /home/user/john/src/loader.c:1050
    #2 0x562401ede227 in read_file /home/user/john/src/loader.c:260
    #3 0x562401ee61eb in ldr_load_pw_file /home/user/john/src/loader.c:1203
    #4 0x562401ed5583 in john_load /home/user/john/src/john.c:1134
    #5 0x562401ed5583 in john_init /home/user/john/src/john.c:1578
    #6 0x562401ed5583 in main /home/user/john/src/john.c:2065
    #7 0x7f98f0943d09 in __libc_start_main ../csu/libc-start.c:308
    #8 0x5624019d6489 in _start (/home/user/john/run/john+0x2dd489)

0x625000002188 is located 0 bytes to the right of 8328-byte region [0x625000000100,0x625000002188)
allocated by thread T0 here:
    #0 0x7f98f105e817 in __interceptor_strdup ../../../../src/libsanitizer/asan/asan_interceptors.cpp:452
    #1 0x562401efd7c5 in xstrdup /home/user/john/src/memory.c:136
    #2 0x562401e4a6a2 in ssh_get_salt /home/user/john/src/ssh_common_plug.c:99
    #3 0x562401ee374b in ldr_load_pw_line /home/user/john/src/loader.c:1050
```

`hdaa`:
```
$ perl -le 'print q{$response$8663faf2337dbcb2c52882807592ec2c$} . "9" x 200 . q{user$myrealm$GET$/$8c12bd8f728afe56d45a0ce846b70e5a$}' > t.pw
$ ./john/run/john t.pw
AddressSanitizer:DEADLYSIGNAL
=================================================================
[...]

PR #5158, opened 2022-06-06
improvements for fuzzer

### LINKED ISSUE #5147: simple fuzzer is weak against len+data fields
#5086 fixed overflow in PEM format. Built-in simple fuzzer could not find this problem.

```
$ python -c 'l = 8000; print "$PEM$1$1$0201000282010100$2048$308204bd02010030$" + str(l) + "$" + "a" * (l * 2)' > t.pw

$ ../run/john t.pw
[...]
==24239==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x62100002693f at pc 0x560d44b28c24 bp 0x7ffd702644c0 sp 0x7ffd702644b8
WRITE of size 1 at 0x62100002693f thread T0
    #0 0x560d44b28c23 in pem_get_salt /home/user/john/src/pem_common_plug.c:132
    #1 0x55eb63c45546 in ldr_load_pw_line /home/user/john/src/loader.c:1050
[...]

$ python -c 'l = 4097; print "$PEM$1$1$0201000282010100$2048$308204bd02010030$" + str(l) + "$" + "a" * (l * 2)' > t.pw

$ ../run/john t.pw
[...]
==24734==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7ffe1d33c6c0 at pc 0x55ed2c8de033 bp 0x7ffe1d33b1e0 sp 0x7ffe1d33b1d8
READ of size 1 at 0x7ffe1d33c6c0 thread T0
    #0 0x55ed2c8de032 in check_pkcs_pad /home/user/john/src/jumbo.c:432
    #1 0x55ed2cc43f48 in pem_decrypt /home/user/john/src/pem_common_plug.c:203
    #2 0x55ed2cc45981 in crypt_all /home/user/john/src/pem_fmt_plug.c:124
[...]
```

Big overflow gets caught by ASan early. But small overflow hits later because overflow happens inside structure.

`pem_common.h`:
```c
struct custom_salt {
	int salt_length;
	unsigned char salt[SALTLEN];
	int iv_length;
	unsigned char iv[16];
	int iterations;
	int ciphertext_length;
	unsigned char ciphertext[CTLEN];
	int cid;  // cipher id
	int key_length;
};
```

`pem_common_plug.c`:
```c
// $PEM$type$cipher$$salt$iterations$iv$blob_length$blob  // type, and cipher should be enough for all possible combinations
```

The overflow happens in the `blob` field. But it requires respective `blog_length` to happen.

Overflow happens decoding hex in `pem_get_salt`:
```c
	cur_salt->ciphertext_length = atoi(p);
	p = strtokm(NULL, "$");
	for (i = 0; i < cur_salt->ciphertext_length; i++)
		cur_salt->ciphertext[i] = atoi16[ARCH_INDEX(p[i * 2])] * 16
			+ atoi16[ARCH_INDEX(p[i * 2 + 1])];
```

I tried to set `ciphertext_length` to `strlen(p) / 2` right before the cycle. With such change, simple fuzzer triggers overflow.

There are a lot of formats with redundant length fields. We might want to remove these fields. Saving compatibility, we might just ignore them completely instead of rejection when they are not in sync with respective data fields.

I guess AFL / libFuzzer can trigger the overflow.

I guess wordlist of simple fuzzer could be improved to trigger the overflow. I'll investigate it.

AlekseyCherepanov:
It closes #5147. Additions should be enough to cause all crashes in #5157.

New line in `fuzz.dic` is needed for the crash in `PBKDF2-HMAC-MD5`.

`update_length_data` implements the fast approach from #5147 and does not affect overall fuzzing time much.

`insert_zeros` generates many more candidates and increases overall fuzzing time quite much. I did not measure precisely, but I guess it is not too bad.

Maybe I should not make my own macros and use standard `isalpha` and `isalnum`. They would be affected by locale. OTOH `strtoll` is affected by locale too. I am not sure. I am not going to change it without feedback.

Also I fixed some memory leaks.


solardiz:
I didn't actually read (only skimmed) your rationale for these changes. I also can't claim to have properly reviewed the code, but this is just an optional fuzzer, so what can go wrong. ;-)

Anyway, one issue I did spot is you don't check for possible `strtoll` errors, yet assume it sets the `after` pointer. I don't know what standards say about this, but it feels unsafe/non-portable. For example, in `popa3d` I have this:

```c
/* SUSv2 says:
 * "Because 0, LONG_MIN and LONG_MAX are returned on error and are also
 * valid returns on success, an application wishing to check for error
 * situations should set errno to 0, then call strtol(), then check errno." */
                errno = 0;
                value = strtol(param, &error, 10);
                if (errno || !*param || *error ||
                    value < 0 || (long)(int)value != value)
                        return -1;
```

So you could want to similarly play with `errno` here, and only rely on `after` being valid if `errno` stayed at zero.

Checking `john-core` now, looks like we have the same issue in `config.c`, but do set/check `errno` in `compiler.c`. There are probably more uses in jumbo.

AlekseyCherepanov:
[SUSv2](https://pubs.opengroup.org/onlinepubs/007908799/xsh/strtol.html):
> Because 0, LONG_MIN and LONG_MAX are returned on error and are also valid returns on success, an application wishing to check for error situations should set errno to 0, then call strtol(), then check errno.

But this phrase is not actual on glibc: empty string returns 0 and does not set `errno`.

```c
#include <stdio.h>
#include <stdlib.h>
#include <errno.h>
int main(int argc, char *argv[])
{
    char *end;
    long long r;
    if (argc != 2)
        return 1;
    errno = 0;
    r = strtoll(argv[1], &end, 16);
    printf("r = %lld, errno = %d, delta = %zd\n", r, errno, end - argv[1]);
    return 0;
}
```

```
$ ./test_strtoll '1'
r = 1, errno = 0, delta = 1
$ ./test_strtoll '01'
r = 1, errno = 0, delta = 2
$ ./test_strtoll '01abc'
r = 6844, errno = 0, delta = 5
$ ./test_strtoll '01xyz'
r = 1, errno = 0, delta = 2
$ ./test_strtoll '  -$'
r = 0, errno = 0, delta = 0
$ ./test_strtoll '  -1$'
r = -1, errno = 0, delta = 4
$ ./test_strtoll '7fffffffffffffff'
r = 9223372036854775807, errno = 0, delta = 16
$ ./test_strtoll '8000000000000000'
r = 9223372036854775807, errno = 34, delta = 16
$ ./test_strtoll 'ffffffffffffffff$abc'
r = 9223372036854775807, errno = 34, delta = 16
$ ./test_strtoll '0x'
r = 0, errno = 0, delta = 1

$ /lib/x86_64-linux-gnu/libc.so.6
GNU C Library (Debian GLIBC 2.28-10+deb10u1) stable release version 2.28.
Copyright (C) 2018 Free Software Foundation, Inc.
```

Later there is word "may":
> If no conversion could be performed, 0 is returned and errno may be set to [EINVAL].

Also there are the following phrases about `endptr`:
> A pointer to the final string is stored in the object pointed to by endptr, provided that endptr is not a null pointer.

> If the subject sequence is empty or does not have the expected form, no conversion is performed; the value of str is stored in the object pointed to by endptr, provided that endptr is not a null pointer.

[C11 standard](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf): chapter 7.22.1.4 has the same phrases about `endptr`. But the chapter does not mention possible `EINVAL` value for `errno` at all.

So my guess is that `endptr` gets the pointer irrespective to errors. While the phrase "Because ..." looks like a mistake in the standard.

I don't care about errors. My code can truncate a long hex field replacing it with `7fffffffffffffff` + my increment (it found a crash in `wowrsp`). Also my code would replace parts of `  -1` (two spaces and -1) 4 times, it was not intended actually. But excessive replacements are ok because overall number of ciphertexts is very low (maybe I should add more values into `times[]`...).

AlekseyCherepanov:
> (maybe I should add more values into `times[]`...)

I tried. New crash: `bitlocker`. So I'll play with it a bit more.

solardiz:
> So my guess is that `endptr` gets the pointer irrespective to errors.

Hopefully, yes, but I'm not sure whether this is guaranteed or is merely implementation courtesy.

> While the phrase "Because ..." looks like a mistake in the standard.

I think it refers to `ERANGE` and inability to reliably detect it other than by the described algorithm.

> I don't care about errors.

If `endptr` is not set, the fuzzer will likely crash - not what we want.

AlekseyCherepanov:
> > While the phrase "Because ..." looks like a mistake in the standard.
> 
> I think it refers to `ERANGE` and inability to reliably detect it other than by the described algorithm.

The phrase mentions 0, so it could not be about `ERANGE` only.

> > So my guess is that `endptr` gets the pointer irrespective to errors.
> 
> Hopefully, yes, but I'm not sure whether this is guaranteed or is merely implementation courtesy.

Crash would be bad. But I don't see anything in the standards to suggest that it is possible. There are 2 phrases about `*endptr` being set: one for parseable data and one for completely incorrect.

`SUSv2`:
> If the correct value is outside the range of representable values, LONG_MAX or LONG_MIN is returned (according to the sign of the value), and errno is set to [ERANGE].

So overflows seem to be part of the parseable case.

For me, the behaviour seems similar to [saturation arithmetic](https://en.wikipedia.org/wiki/Saturation_arithmetic). Without a pointer in `*endptr`, it would not be possible to parse overflowing numbers conveniently as regular numbers, while the behaviour suggests it should be possible to handle them.

I rely on this logic. Avoiding it would mean search for end of number myself. I could implement it and check digits myself too to avoid effects of locale.

solardiz (approved):

AlekseyCherepanov:
So I think it is guaranteed that `*endptr` is set having non-null `endptr`. To be honest, I did not think about this aspect writing my code. Thank you for your concern, @solardiz!

PR #5166, opened 2022-07-01
fixes for crashes found with fuzzing

### LINKED ISSUE #5157: crashes under ASan from fuzzing: PBKDF2-HMAC-MD5, SSH, hdaa, scrypt, and others

AlekseyCherepanov:
It closes #5157. It affects the following formats:
```
rsvp,gpg,bitlocker,net-md5,o10glogon,o3logon,SSHA512,ike,dynamic_15,vtp,nsec3,wowsrp,SSH,hdaa,PBKDF2-HMAC-MD5,krb5asrep,scrypt,dynamic_1552,bitlocker-opencl,gpg-opencl,SSH-opencl,PBKDF2-HMAC-MD5-opencl
```

`--test=0` and `--fuzz=../run/fuzz.dic` are ok.

I prepared fixes quickly, so some of them might be suboptimal.

JFYI I was caught by the following problem fixing `wowsrp`: I increased buffer size right in `get_salt` instead of change to `SALT_SIZE`. So salt would be truncated and there would be a crash during cracking or uncrackable hash. Self-tests do not catch such mistake: all tests worked well with smaller buffer. Fuzzer does not catch it too.

My fixes for memory leaks in #5158 introduced use-after-free for dynamic salts because `ldr_free_db` is called in fuzzer and in john.
```
==4006==ERROR: AddressSanitizer: heap-use-after-free on address 0x60d00001bbe8 at pc 0x5604db9ddef1 bp 0x7ffc3e4223f0 sp 0x7ffc3e4223e8
READ of size 1 at 0x60d00001bbe8 thread T0
    #0 0x5604db9ddef0 in dyna_salt_remove_fp /home/user/john/src/dyna_salt.c:56
    #1 0x5604dba2dd33 in ldr_free_db /home/user/john/src/loader.c:1358
    #2 0x5604dba1ede3 in john_done /home/user/john/src/john.c:1953
    #3 0x5604dba1ede3 in main /home/user/john/src/john.c:2083
```

The problem affects a subset of formats using `dyna_salt` (approximately `gpg,7z,pkzip,zip` + their opencl variants).

The last commit fixes the use-after-free. To allow excessive calls to `ldr_free_db`, I just set `db->salts` to `NULL` in `ldr_free_db`. I am not sure if it is ok. But ASan does not report leaks. I would like this bit to be reviewed better.

If everything is fine, this PR is ready. Thanks!

### DIFF
--- a/src/HDAA_fmt_plug.c
+++ b/src/HDAA_fmt_plug.c
@@ -68,6 +68,9 @@ john_register_one(&fmt_HDAA);
 // This is 8 x 64 bytes, so in MMX/SSE2 we support up to 9 limbs of MD5
 #define HTMP                    512
 
+// That's arbitrary because uri part is not limited by anything natural.
+#define MAX_CIPHERTEXT_LEN      8192
+
 typedef struct
 {
 	size_t h1tmplen;
@@ -174,6 +177,8 @@ static int valid(char *ciphertext, struct fmt_main *self)
 
 	if (strncmp(ciphertext, FORMAT_TAG, TAG_LENGTH) != 0)
 		return 0;
+	if (strlen(ciphertext) > MAX_CIPHERTEXT_LEN)
+		return 0;
 	ctcopy = xstrdup(ciphertext);
 	keeptr = ctcopy;
 	ctcopy += TAG_LENGTH;
@@ -216,6 +221,8 @@ static char *split(char *ciphertext, int index, struct fmt_main *self)
 	char *cp;
 	if (strncmp(ciphertext, FORMAT_TAG, TAG_LENGTH))
 		return ciphertext;
+	if (strlen(ciphertext) > MAX_CIPHERTEXT_LEN)
+		return ciphertext;
 	cp = ciphertext + TAG_LENGTH;
 	cp = strchr(cp, '$'); if (!cp) return ciphertext;
 	cp = strchr(cp+1, '$'); if (!cp) return ciphertext;
@@ -224,8 +231,10 @@ static char *split(char *ciphertext, int index, struct fmt_main *self)
 	cp = strchr(cp+1, '$'); if (!cp) return ciphertext;
 	// now if we have $binary_hash$ then we remove the last '$' char
 	if (strlen(cp) == 1 + BINARY_SIZE*2 + 1) {
-		static char out[256];
-		strnzcpy(out, ciphertext, sizeof(out));
+		static char *out;
+		if (!out)
+			out = mem_alloc_tiny(MAX_CIPHERTEXT_LEN + 1, MEM_ALIGN_NONE);
+		strnzcpy(out, ciphertext, MAX_CIPHERTEXT_LEN + 1);
 		out[strlen(out)-1] = 0;
 		return out;
 	}
--- a/src/HDAA_fmt_plug.c
+++ b/src/HDAA_fmt_plug.c
@@ -174,6 +174,7 @@ static void done(void)
 static int valid(char *ciphertext, struct fmt_main *self)
 {
 	char *ctcopy, *keeptr, *p;
+	size_t user_len, realm_len, nonce_len, noncecount_len, clientnonce_len, qop_len;
 
 	if (strncmp(ciphertext, FORMAT_TAG, TAG_LENGTH) != 0)
 		return 0;
@@ -189,7 +190,12 @@ static int valid(char *ciphertext, struct fmt_main *self)
 		goto err;
 	if ((p = strtokm(NULL, "$")) == NULL) /* user */
 		goto err;
+	user_len = strlen(p);
 	if ((p = strtokm(NULL, "$")) == NULL) /* realm */
+		goto err;
+	realm_len = strlen(p);
+	/* snprintf() later would truncate data making hash uncrackable. */
+	if (user_len + realm_len + 2 > HTMP - PLAINTEXT_LENGTH - 1)
 		goto err;
 	if ((p = strtokm(NULL, "$")) == NULL) /* method */
 		goto err;
@@ -197,17 +203,25 @@ static int valid(char *ciphertext, struct fmt_main *self)
 		goto err;
 	if ((p = strtokm(NULL, "$")) == NULL) /* nonce */
 		goto err;
+	nonce_len = strlen(p);
 	if ((p = strtokm(NULL, "$")) == NULL) /* End of legacy HDAA or noncecount */
 		goto end_hdaa_legacy;
+	noncecount_len = strlen(p);
 	if ((p = strtokm(NULL, "$")) == NULL) /* clientnonce */
 		goto err;
+	clientnonce_len = strlen(p);
 	if ((p = strtokm(NULL, "$")) == NULL) /* qop */
 		goto err;
+	qop_len = strlen(p);
 	if ((p = strtokm(NULL, "$")) != NULL)
+		goto err;
+	if (nonce_len + noncecount_len + clientnonce_len + qop_len + 32 + 5 > HTMP - CIPHERTEXT_LENGTH - 1)
 		goto err;
 
 end_hdaa_legacy:
 	MEM_FREE(keeptr);
+	if (nonce_len + 32 + 2 > HTMP - CIPHERTEXT_LENGTH - 1)
+		return 0;
 	return 1;
 
 err:
--- a/src/bitlocker_common.h
+++ b/src/bitlocker_common.h
@@ -4,6 +4,7 @@
 
 #include "formats.h"
 
+#define MAX_DATALEN             256
 #define MACLEN                  16
 #define SALTLEN                 16
 #define IVLEN                   12  // nonce length
@@ -24,7 +25,7 @@ typedef struct {
 	int data_size;
 	unsigned char salt[SALTLEN];
 	unsigned int iterations;
-	unsigned char data[256];
+	unsigned char data[MAX_DATALEN];
 	unsigned char iv[IVLEN]; // nonce
 	unsigned char mac[MACLEN];
 } bitlocker_custom_salt;
--- a/src/bitlocker_variable_code.h
+++ b/src/bitlocker_variable_code.h
@@ -152,6 +152,8 @@ static int bitlocker_common_valid(char *ciphertext, struct fmt_main *self)
 	value = atoi(p);
 	if ((p = strtokm(NULL, "$")) == NULL)   // data encrypted by aes_ccm key, contains encrypted volume master key (vmk)
 		goto err;
+	if (value > MAX_DATALEN)
+		goto err;
 	if (hexlenl(p, &extra) != value * 2 || extra)
 		goto err;
 
--- a/src/dynamic_fmt.c
+++ b/src/dynamic_fmt.c
@@ -566,7 +566,8 @@ static int valid(char *ciphertext, struct fmt_main *pFmt)
 	if (cp[cipherTextLen] && cp[cipherTextLen] != '$')
 		return 0;
 // NOTE if looking at this in the future, this was not my fix.
-	if (strlen(&cp[cipherTextLen]) > SALT_SIZE)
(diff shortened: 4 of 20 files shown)


src/HDAA_fmt_plug.c
```diff
@@ -68,6 +68,9 @@ john_register_one(&fmt_HDAA);
 // This is 8 x 64 bytes, so in MMX/SSE2 we support up to 9 limbs of MD5
 #define HTMP                    512
 
+// That's arbitrary because uri part is not limited by anything natural.
+#define MAX_CIPHERTEXT_LEN      8096
```
solardiz: 8096 is a weird choice for an arbitrary limit - perhaps you meant 4096 or 8192?

src/HDAA_fmt_plug.c
```diff
@@ -224,8 +245,10 @@ static char *split(char *ciphertext, int index, struct fmt_main *self)
 	cp = strchr(cp+1, '$'); if (!cp) return ciphertext;
 	// now if we have $binary_hash$ then we remove the last '$' char
 	if (strlen(cp) == 1 + BINARY_SIZE*2 + 1) {
-		static char out[256];
-		strnzcpy(out, ciphertext, sizeof(out));
+		static char *out;
+		if (!out)
+			out = mem_alloc_tiny(MAX_CIPHERTEXT_LEN, MEM_ALIGN_NONE);
+		strnzcpy(out, ciphertext, MAX_CIPHERTEXT_LEN);
 		out[strlen(out)-1] = 0;
```
solardiz, line 252: I think you need to allocate `MAX_CIPHERTEXT_LEN + 1` and pass that to `strnzcpy`, so that `MAX_CIPHERTEXT_LEN` is treated literally as the maximum length, which is also how you treat ti in the check in `valid`.

Also, drop the unneeded `out[strlen(out)-1] = 0;`

solardiz (approved):
I only looked at what's in patch context. I made some nitpicks, but overall this looks fine. Thank you, Aleksey!

solardiz:
> JFYI I was caught by the following problem fixing wowsrp: I increased buffer size right in get_salt instead of change to SALT_SIZE. So salt would be truncated and there would be a crash during cracking or uncrackable hash. Self-tests do not catch such mistake: all tests worked well with smaller buffer. Fuzzer does not catch it too.

I guess this means we need to enhance the self-tests to truncate a copy of the returned salt to the format's `salt_size`.

However, we seem to have no way to _guarantee_ the data following the truncated salt won't just happen to match what's in the full returned salt (unless we over-read the returned salt just to ensure we produce different e.g. first byte there, but such over-reading is unsafe, so let's not do it). So this will be a best-effort test, which is fine. We can over-allocate, say, 1 byte to put some varying value in that extra byte.

src/HDAA_fmt_plug.c
AlekseyCherepanov: `out[strlen(out)-1] = 0;` is needed: it removes the last char (`$`).

But I need `+1` for the length. Otherwise with input of max length, `strnzcpy` would remove the last char itself and then the next line would remove one more char.

src/HDAA_fmt_plug.c
AlekseyCherepanov: Ouch, I meant 8192 indeed.

AlekseyCherepanov:
I fixed `hdaa: harden valid()...` commit and rebased further commits on it. I used `magit-commit-instant-fixup` in emacs, that should be something like `git commit --fixup=... ...` + `git rebase --autosquash ...`. So all further commits are re-written. With such number of commits, button `Compare` seems invaluable.

solardiz:
> With such number of commits, button `Compare` seems invaluable.

Yes, but it only shows cumulative changes - it would not help detect e.g. the desired changes inadvertently being in a wrong commit, or some commit introducing extra changes that another commit then removes. I wish GitHub provided a tool to easily detect those.

AlekseyCherepanov:
> I guess this means we need to enhance the self-tests to truncate a copy of the returned salt to the format's `salt_size`.

I guess copying into test db truncates at `salt_size`. The problem should be that self-tests do not include cases for maximum acceptable length.

Fuzzer could find overflow for too small buffers. Static analysis could find that buffer size is not in sync with `salt_size` in the format structure (also formats where buffer size might vary depending on data). Actually debugging info should contain sizes and ASan tracks such info for heap allocations, so it should be possible to check at run-time during fuzzing.

Talking about checks of additional constraints during fuzzing, I would check that result from `split` passes `valid`.

solardiz:
> self-tests do not include cases for maximum acceptable length.

I don't understand this.

> I would check that result from `split` passes `valid`.

Please feel free to implement that. Thanks!

AlekseyCherepanov:
> it only shows cumulative changes
> I wish GitHub provided a tool to easily detect those.

Force-pushing to keep nice "source history" seems a popular variant of workflow and GitHub is [informed](https://github.com/github-community/community/discussions/3478) about it. OTOH I don't see discussions about highlighting commits with new changes to review breakdown of cumulative changes.

AlekseyCherepanov:
>> self-tests do not include cases for maximum acceptable length.

Some formats has salt of variable length. Usually it is stored with padding and treated as fixed size by code outside of format. (There is dyna_salt feature allowing variable lengths without padding, but it is used for a small number of formats where really big salts can happen.) So some of formats with variable lengths of salts do not include test cases for maximum allowed length of salt. For instance, `wowsrp` has self-tests that use 22 bytes of salt at maximum, while old declared `salt_size` is 67 and new `salt_size` is 101.

PR #5169, opened 2022-07-06
Newer Streebog code (a.k.a. GOST R 34.11-2012) that doesn't require SIMD

### LINKED ISSUE #4974: Stribog format issues (Streebog, GOST R 34.11-2012)
@vt-alt pointed out to me that the SSE4.1 implementation of Streebog that we use is actually slower than SSE2 and even than scalar reference. We could want to replace it, which would also enable these formats to work on systems without SSE4.1.

This prompted me to take a look, and other issues I notice are:

Why do we even call these two formats `Stribog*` when the actual English name for the hashes is Streebog? Should we rename?

Also, we define a `FORMAT_LABEL` in `stribog_fmt_plug.c`, but it's never used and the actual two format labels are different. We should probably at least drop the `FORMAT_LABEL` line, or replace it with two for labels that we actually use and then use those two new macros.

The "problem" is no one really cares about these formats after they've been used in a contest once (which is when and why they were added).

magnumripper:
Also minor tweaks to raw "stribog" formats.

Closes #4974


magnumripper:
The code is from https://github.com/adegtyarev/streebog and seem to be same we always used, but newer. There's still a regression for AVX CPU's if using the SSE 4.1 code, so we don't. I presume a pure SSE 4.1 CPU will benefit from it.

solardiz:
> There's still a regression for AVX CPU's if using the SSE 4.1 code, so we don't. I presume a pure SSE 4.1 CPU will benefit from it.

@vt-alt Do you happen to know if the SSE4.1 code is beneficial (over SSE2, right?) on any CPU at all?

vt-alt:
I think that reference Streebog implementation is unmaintained long ago, and in particular the SSE4.1 implementation is useless (actually it's slower than any other implementation, even than `ref` (from gost-engine) if compared on modern CPU, and surely it's slower than it's own SSE2/SSE3 impl).
Slightly (but not much) better maintained is `gosthash2012*` files in https://github.com/gost-engine/engine At least t's used for something practical and updated sometimes.

solardiz:
@magnumripper Let's at least drop the SSE4.1 implementation from our tree, then?

magnumripper:
> @magnumripper Let's at least drop the SSE4.1 implementation from our tree, then?

Will do. I'll drop the unused MMX files as well.

magnumripper:
I verified that it not only builds and runs with `--disable-simd` but also on Apple M1 (with good speed).

magnumripper:
Let's not merge this until I try out that version from https://github.com/gost-engine/engine

vt-alt:
It improves on `ref` implementation, mainly on `#define XLPS` https://github.com/gost-engine/engine/blob/master/gosthash2012_ref.h#L33

magnumripper:
I see no difference at all in performance between this PR and the code in https://github.com/gost-engine/engine - neither for AVX or the vanilla code. I also tried the alternate XLPS macro on its own with this PR's tree - again no measurable difference. Also, the license in that repo is "under the same license as OpenSSL" while the one for this PR is (I think) [better](https://github.com/openwall/john/blob/bleeding-jumbo/src/LICENSE.gost).

BTW I didn't touch that license in this PR - perhaps I should. it's misplaced as `src/LICENSE.gost` ever since ebbbfc1a. Perhaps we should move that file to `doc/LICENSE.gost3411-2012` before merging?

magnumripper:
Oh, and according to https://github.com/adegtyarev/streebog/blob/master/README.md "*This work is dual-licensed under BSD 2-Clause "Simplified" License and GNU General Public License v2.0 (or later). You can choose between one of them if you use this software.*".

vt-alt:
Also, there (at gost-engine) is `addcarry` optimization. About XLPS, gost-engine's benchmarks shown XLPS optimization resulting in 20% speed up on old Xeon E5630. (And together 40% speed up on E5-2640.) Old discussion there https://github.com/gost-engine/engine/pull/211

Maybe this is only relevant for older processors and older compilers? But att it was significant,

vt-alt:
> Also, the license in that repo is "under the same license as OpenSSL"

Which is Apache-2.0 now.

(AFAIK, Historically, the Streebog code was the same in both implementations. I also att requested Degtyarev to add that GPL-2.0 second license to his impl to better fit it into Linux kernel (so the same non-SIMD code is also in Linux; including XLPS opt). Even though GPL it's not strictly required, since BSD is compatible with it.)
And I think both BSD or Apache should be ok for john[?]

solardiz:
> misplaced as `src/LICENSE.gost` ever since [ebbbfc1](https://github.com/openwall/john/commit/ebbbfc1a8dbcddca16540622a17e21705890c619). Perhaps we should move that file to `doc/LICENSE.gost3411-2012` before merging?

I'm not so sure. We also have `aes/aesni/LICENSE` and we have license statements inside various source files.

magnumripper:
> Also, there (at gost-engine) is `addcarry` optimization. About XLPS, gost-engine's benchmarks shown XLPS optimization resulting in 20% speed up on old Xeon E5630. (And together 40% speed up on E5-2640.) Old discussion there [gost-engine/engine#211](https://github.com/gost-engine/engine/pull/211)

I tried that as well, on its own and I couldn't see *any* change in performance. Perhaps modern compilers can recognize opportunity for that instruction without the intrinsic?

> > Perhaps we should move that file to `doc/LICENSE.gost3411-2012` before merging?
> 
> I'm not so sure. We also have `aes/aesni/LICENSE` and we have license statements inside various source files.

Then we can merge as-is, or perhaps I could move the contents of `LICENSE.gost` plus that statement about dual licensing into comments at top of `gost3411-2012-core_plug.c`?

solardiz:
> Perhaps modern compilers can recognize opportunity for that instruction without the intrinsic?

Perhaps test with default `gcc` on `super`, then? It's pretty old.

> move the contents of `LICENSE.gost` plus that statement about dual licensing into comments at top of `gost3411-2012-core_plug.c`?

This makes sense to me.

magnumripper:
Tried again - I can see `adcq` instructions being emitted when using that intrinsic (and only then), but the resulting performance difference is smaller than normal fluctuations. For using it, we'd need a new test in `configure` for adding `-madx` and I think the incentive is simply too weak.

I'm merging as-is now.

magnumripper:
Just for the record, here's my test code
```diff
diff --git a/src/gost3411-2012-core_plug.c b/src/gost3411-2012-core_plug.c
index 324a93236..4e6b835b9 100644
--- a/src/gost3411-2012-core_plug.c
+++ b/src/gost3411-2012-core_plug.c
@@ -32,6 +32,10 @@
 
 #include "gost3411-2012-core.h"
 
+#if __ADX__
+#include <x86intrin.h>
+#endif
+
 #define BSWAP64(x) \
     (((x & 0xFF00000000000000ULL) >> 56) | \
      ((x & 0x00FF000000000000ULL) >> 40) | \
@@ -81,10 +85,14 @@ static inline void
 add512(const uint512_u *x, const uint512_u *y, uint512_u *r)
 {
 #ifndef __GOST3411_BIG_ENDIAN__
-    unsigned int CF;
+    unsigned int CF = 0;
     unsigned int i;
 
-    CF = 0;
+#if __ADX__
+#warning Using _addcarry_u64
+	for (i = 0; i < 8; i++)
+		CF = _addcarry_u64(CF, x->QWORD[i] , y->QWORD[i], &(r->QWORD[i]));
+#else
     for (i = 0; i < 8; i++)
     {
         const unsigned long long left = x->QWORD[i];
@@ -95,6 +103,7 @@ add512(const uint512_u *x, const uint512_u *y, uint512_u *r)
             CF = (sum < left);
         r->QWORD[i] = sum;
     }
+#endif
 #else
     const unsigned char *xp, *yp;
     unsigned char *rp;
```
The above along with adding `-madx` to CFLAGS will use the intrinsic.

magnumripper:
> Perhaps test with default `gcc` on `super`, then? It's pretty old.

It doesn't even recognize the `-madx` option.

solardiz:
I didn't realize this was about ADX. Anyway, you can test that on `well` - it's about when ADX first appeared. I was also confused by @vt-alt mentioning old CPUs that don't even have ADX - but I guess that was about the other optimization.

> I can see `adcq` instructions being emitted when using that intrinsic

I guess you mean `adcxq`? Since `adcq` doesn't require any instruction set extensions.

> For using it, we'd need a new test in `configure` for adding `-madx`

We could possibly be testing/adding it along with AVX2. Intel CPUs introduced these at the same time. I don't know if there are any CPUs that have AVX2 yet not ADX - but maybe...

> and I think the incentive is simply too weak.

That's true. No one really cares about Streebog in JtR anyway, and if we wanted to optimize it for real we'd be computing multiple concurrent instances in SIMD, not trying to optimize one instance.

vt-alt:
I didn't mean `_addcarry_u64` is using `ADCX` it's only `ADC` on [gcc](https://github.com/gost-engine/engine/pull/206#issuecomment-580326754)/clang. But then gcc wasn't able to optimize that loop better w/o it. So `ADCX` of course don't give much performance in compare to `ADC`, but any of them gives improvement over "unoptimized" (or naively optimized by gcc) loop.

Also I don't know how you benchmark, all our benchmarks are over gost-engine (with `-O2`) + `openssl speed`.

magnumripper:
> > I can see `adcq` instructions being emitted when using that intrinsic
> 
> I guess you mean `adcxq`? Since `adcq` doesn't require any instruction set extensions.

Apparently gcc emits ADC rather than ADCX because [it's faster](https://github.com/gost-engine/engine/pull/206#issuecomment-580326754) - yet it doesn't emit ADC without the ADCX intrinsic. What we'd really want is an intrinsic for ADC 🙃 

> > For using it, we'd need a new test in `configure` for adding `-madx`
> 
> We could possibly be testing/adding it along with AVX2. Intel CPUs introduced these at the same time. I don't know if there are any CPUs that have AVX2 yet not ADX - but maybe...

I was hoping `-mavx2` would infer `-madx` but it doesn't.

vt-alt:
ps. If `_addcarry_u64 ` does not provide adc optimization perhaps we should remove it from gost-engine too, but I don't understand how code should be organized that gcc optimize it properly (it should understand that there is just a carry that should be passed into next iteration to add).

magnumripper:
> Also I don't know how you benchmark, all our benchmarks are over gost-engine (with `-O2`) + `openssl speed`.

That's definitely a difference: While OpenSSL probably hashes few big chunks of data, JtR formats would typically hash many small chunks. I do not think you should drop it from gost-engine.

vt-alt:
> That's true. No one really cares about Streebog in JtR anyway, and if we wanted to optimize it for real we'd be computing multiple concurrent instances in SIMD, not trying to optimize one instance.

BTW I tried to experiment with this approach using `ispc` but not able to achieve much (or rather anything). In theory it's possible to calculate two compression functions in parallel and there's enough registers.

solardiz:
I meant starting with scalar implementation and creating a new multi-instance SIMD implementation from it. Perhaps we'd compute 8 instances on AVX-512. If there are enough registers and not enough parallelism within one instance, then we could also interleave a couple of these, for 16 concurrent instances. Similar to how we already do it for MD*/SHA*, or how an OpenCL backend would likely do it automatically if we wrote a non-vectorized kernel using `ulong`.

However, if no one really needs these formats in JtR, it's a waste of time.

solardiz:
> starting with scalar implementation and creating a new multi-instance SIMD implementation from it.

This would require gather loads, though, which are rather slow on CPUs currently (GPUs handle gather loads better, perhaps because not doing so would hurt them more). Instructions exist on AVX2+, can be combined from two half-loads on SSE2+/AVX, but are slow anyhow. So overall effect is uncertain.

PR #5173, opened 2022-07-07
Internal: Allow --rules-skip-nop with single mode

magnumripper:
Using --single did not set FLG_RULES_IN_USE. It should!


solardiz (approved):
Looks good to me. I assume you're removing `FLG_SINGLE_CHK` from a couple of places simply because it became redundant there.

While reviewing this, found that you had introduced `FLG_RULES_IN_USE` in 8d9ff4ac3b6beecf4f3c71f5243bf1295bbfb891 to have a flag set by both `--rules` and `--rules-stack`.

solardiz:
BTW, I don't get why this PR/commit's title is prefixed with "Internal". This is a change to what option combinations are available to a user, why is it called "internal"? And we don't have a component called "internal", do we? Weird. Anyway, I already merged this.

magnumripper:
Oh, that was a mistake. I have a branch called "internal" for WIP stuff not ready for PR or of no value for others. I see now same happened with b3ec292c38 four years ago.

PR #5180, opened 2022-07-19
Change meaning of %N rules where N=0

### LINKED ISSUE #5179: Rule for finding last occurrence of character/class
A trivial example is "bump last digit": With current code, this can be done with eg. `r /?d +p r` but that is a pretty wasteful operation. 

The %NX (or %N?C) command can currently be used with N=0 ~but it behaves exactly like N=1~ (EDIT not true, see below). This experimental patch changes N=0 to find the *last* occurence of a character or class:
```diff
diff --git a/src/rules.c b/src/rules.c
index a80c99e03..2f30e4af3 100644
--- a/src/rules.c
+++ b/src/rules.c
@@ -1177,12 +1177,12 @@ char *rules_apply(char *word_in, char *rule, int split, char *last)
 
 		case '%':
 			{
-				int count = 0, required, pos;
+				int count = 0, last = -1, required, pos;
 				POSITION(required)
 				CLASS_export_pos(0,
-				    if (++count >= required) break, {})
-				if (count < required) REJECT
-				rules_vars['p'] = pos;
+				    last = pos; if (++count >= required && required) break, {})
+				if (!count || count < required) REJECT
+				rules_vars['p'] = (last >= 0 ? last : pos);
 			}
 			break;
 
```
It will still reject if no occurence of the character/class was found. While the above patch doesn't seem to slow down normal use of `%N`, it appears to be 7-8% faster than that reverse-search-reverse rule, using (also simpler to write) `%0?d +p` rule.

Can this be achieved with current functionality using some other, better trick? Or is this patch something we'd consider to merge (along with documentation)?

magnumripper:
Until now it was just the same as N=1.  This changes its meaning to find the *last* occurrence of character/class (useful because p will be set to the position of it).

Example use: Permute last digit using `%0?d op[0-9]`

Closes #5179

### DIFF
--- a/doc/RULES
+++ b/doc/RULES
@@ -279,9 +279,11 @@ s?CY	replace all characters of class C in the word with Y
 U	reject the word unless it is valid UTF-8 (use with -u rule reject)
 eC	title case, where C is the separation character
 e?C	title case, with chars from class C as separation characters.
-	NOTE, see hashcat specific 'E' rule which is generic title case for space
-	only characters.  The 'e' rule extendeded more powerful version.
+	NOTE, see hashcat specific 'E' rule which is generic title case for
+	space only characters.  The 'e' rule is its extended version.
 
+For %NX and %N?C, using N=0 will make it find the last occurrence of X or
+class C (and as always, set p to the position of it) or reject if none found.
 Note that U will accept plain ASCII. It will only reject words that contain
 8-bit characters but can't be parsed as UTF-8. It can be used to reject
 invalid input words, or to reject invalid output words after applying other
--- a/src/rules.c
+++ b/src/rules.c
@@ -1177,12 +1177,12 @@ char *rules_apply(char *word_in, char *rule, int split, char *last)
 
 		case '%':
 			{
-				int count = 0, required, pos;
+				int count = 0, last = -1, required, pos;
 				POSITION(required)
 				CLASS_export_pos(0,
-				    if (++count >= required) break, {})
-				if (count < required) REJECT
-				rules_vars['p'] = pos;
+				    last = pos; if (++count >= required && required) break, {})
+				if (!count || count < required) REJECT
+				rules_vars['p'] = (last >= 0 ? last : pos);
 			}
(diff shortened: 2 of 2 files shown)


doc/RULES
```diff
@@ -279,9 +279,11 @@ s?CY	replace all characters of class C in the word with Y
 U	reject the word unless it is valid UTF-8 (use with -u rule reject)
 eC	title case, where C is the separation character
 e?C	title case, with chars from class C as separation characters.
-	NOTE, see hashcat specific 'E' rule which is generic title case for space
-	only characters.  The 'e' rule extendeded more powerful version.
+	NOTE, see hashcat specific 'E' rule which is generic title case for
+	space only characters.  The 'e' rule extendeded more powerful version.
```
solardiz: "extendeded"? I'd rewrite the sentence as "The 'e' rule is its extended version."

solardiz (approved):
I don't mind, although I'm not sure this is worth it, and maybe Frank's suggestion of a `%-` command would have been better (implementation wise, I think we'd branch to a separate piece of code upon seeing the `-`, so this would more obviously not slow down the existing command).

An interesting aspect is what happens to this command's original behavior - rejection on too few matches. Previously, 0 would mean to never reject, right? Now it rejects on fewer than 1 match, right? This is probably OK if no one was using that behavior, but still. The old behavior could possibly be used though a preprocessor macro or a variable letter (with that variable's value happening to be 0) - but again, I doubt any such uses exist, so we're probably OK.

The variable aspect is interesting in that it raises the question of what we'd do with negative variable values if we were to introduce the `-` command. Would we branch to the other implementation upon seeing a negative value? Or rather, should the branching to negative be not upon seeing the `-` sign, but upon seeing negative value after its full parsing (including variable to value)?

This gets complicated... so I guess let's just merge and be done with it if there's no real demand for anything more generic.

solardiz:
Also, I assume you've tested this against our existing rule sets and made sure the output stays unchanged?

PR #5183, opened 2022-08-15
mask: change error to warning if ?w missing in hybrid mask

rofl0r:
during contest, it's often convenient to change command line from
a previous hybrid mask to a mask not containing ?w, expecting the
wordlist argument to be just ignored. but the error forces one to
remove the wordlist argument (often via ssh, with high lag for
every keypress) and start again, just to add the wordlist later
again if the attempt didn't produce the expected result.
changing it to a warning instead improves usability.

### DIFF
--- a/src/mask.c
+++ b/src/mask.c
@@ -2405,8 +2405,9 @@ static void finalize_mask(int len)
 			    mask_add_len >= (unsigned int)len &&
 			    mask_num_qw == 1) {
 				if (john_main_process)
-				fprintf(stderr, "Error: Hybrid mask must contain ?w/?W after truncation for max. length\n");
-				error();
+				fprintf(stderr, "Warning: Hybrid mask must contain ?w/?W after truncation for max. length, ignoring wordlist\n");
+				options.flags &= ~(FLG_MASK_STACKED|FLG_WORDLIST_CHK);
+				options.flags |= FLG_MASK_CHK;
 			}
 		} else {
 			i++;
@@ -2420,8 +2421,9 @@ static void finalize_mask(int len)
 
 		if (mask_num_qw == 0) {
 			if (john_main_process)
-				fprintf(stderr, "Error: Hybrid mask must contain ?w or ?W\n");
-			error();
+				fprintf(stderr, "Warning: Hybrid mask must contain ?w or ?W, ignoring wordlist\n");
+			options.flags &= ~(FLG_MASK_STACKED|FLG_WORDLIST_CHK);
+			options.flags |= FLG_MASK_CHK;
 		}
 	} else {
 		if (mask_num_qw && john_main_process)


magnumripper (commented):
The changes don't seem to match the description. I can't see how they would allow a stray `?w` for non-hybrid mode. But they will allow running hybrid mode without `?w`, which is kinda weird and might need more testing (in case you didn't already): For example what if the wordlist has three lines and you give `?d`. Would it produce just the ten digits once, or three times with dupes?

magnumripper:
> The changes don't seem to match the description.

Re-reading the description, I take that back. If you tested this and verified the wordlist is effectively ignored, I think we could merge this but I'd wait for Solar's opinion.

AlekseyCherepanov:
Your changes are not sufficient to achieve your described behaviour:
```
$ ../run/john --mask='[12]' --wordlist=<(echo a; echo b) --stdout
[...]
1
2
1
2
4p 0:00:00:00 100.00% (2022-08-16 14:29) 44.44p/s 2
```
In the above example mask is repeated for each word in the wordlist because the mode is still hybrid internally.

Personally I would prefer error over warning because it is very easy to overlook a warning. But it might be interesting to experiment with this behaviour, so an option in config file might be desirable.

Unrelated to john, it is possible to improve usability for ssh connections with high lag using something for line editing on your side. I mean something like readline wrapper around the whole remote shell. Personally I use shell-mode in emacs, so I edit input in emacs, hit return and emacs sends the whole line into shell. It has drawbacks:
- shell-mode does not work with tui apps, so for instance I cannot run remote nano (that's special to my setup of emacs, there are ways around that I did not enable).
- History of commands is shared across nested shells, sometimes it is convenient, but sometimes it backfires because it is easier to leak some commands not intended for remote side. Particularly in CTFs I quite often leak command to run exploit and get shell into this opened shell.
- As of input is committed altogether, remote side may stop reading before end of input leaving a part to upper shell (wrapping only ssh you would need to invoke ssh from ssh to let it happen).

Example for partial reading mentioned above:
```
$ python -c 'import os; os.read(0, 5)'
12345echo hithere
$ hithere
$ 
```
where `python -c 'import os; os.read(0, 5)'` and `12345echo hithere` are input and everything else is output. shell-mode sends `12345echo hithere` altogether to its shell. `python` reads 5 bytes and exits, so remaining output gets into upper shell and is interpreted as command. `$ ` is the prompt, `hithere` is printed because `echo hithere` was read by shell instead of python.

rofl0r:
@AlekseyCherepanov thanks for your testcase. force-pushed branch so it now works as expected.
as for various forms of having lag-free ssh connection, i'm aware that some solutions exist, but they all seem to have serious drawbacks.

magnumripper (commented):
I think you need to unset `FLG_WORDLIST_CHK` instead of `FLG_WORDLIST_SET`, or strange side effects *may* occur.

frank-dittrich:
> during contest, it's often convenient to change command line from a previous hybrid mask to a mask not containing ?w, expecting the wordlist argument to be just ignored. but the error forces one to remove the wordlist argument (often via ssh, with high lag for every keypress) and start again, just to add the wordlist later again if the attempt didn't produce the expected result. 

I can't imagine that I ever would use this "feature". Can you provide examples of masks you would use with and without a wordlist?
I'd rather get the error when specifying an incorrect command. 
Otherwise I'm afraid I might waste more time running commands I didn't intend to run.

rofl0r:
> Can you provide examples of masks you would use with and without a wordlist?

contest is a dynamic process where you try stuff, check whether it gives cracks, and change strategy often.

imagine you've been using long command line like 
`john --session=foo --device=1 --wordlist=rockyou.txt  --format=foo-opencl --mask=?w?d?d?d?d foo.pw`while studying cracked passwords others have uploaded and see this password: 06-11-91, so you want to quickly change mask to ?d?d-?d?d-?d?d, then continue with another variation of rockyou mask cracking. currently john forces you to edit the command line and remove the --wordlist argument for this attempt, just to add it back later. it's been annoying me ever since i started using mask mode in contests.

frank-dittrich:
Pressing `<ctrl>-<cursor_left>` and `<ctrl>-<w>` a few times doesn't look like much of a problem for me.
I would be more worried that I might later re-try some of the commands on the other formats, not noticing that warning, and spending processing time on useless double work.

But others don't seem to mind changing the current behaviour, so may be my scepticism isn't justified.

ghost:
I'm also skeptical about the patch. Anyway, if it's useful for real users, well, I can't say nothing against it.

rofl0r:
> Pressing <ctrl>-<cursor_left> and <ctrl>-<w> a few times doesn't look like much of a problem for me.

thanks for pointing out `<ctrl>-<cursor_left>`. i wasn't aware of this shell feature. but even so, in the above command line example this stops at every non-ascii character, so at 3 seconds lag for every keypress, it still takes you like 30 seconds to remove the gratuitous wordlist argument from the command line.

solardiz:
My primary reason to consider merging this would be not to discourage @rofl0r, who I hope would continue to contribute. Other than that, I find these changes inconsistent with how we typically handle incompatible command-line options.

I find it weird that one would want to use almost the same mask with and then without a wordlist. I'd expect the mask itself to be edited in a major way between these runs, and then removing the wordlist option doesn't feel like an excessive step for our code to expect.

Also, our `?w` in mask isn't only for wordlists - it's also for other modes such as `--incremental`, which can then have a mask added on top of them. How will(?) this PR affect that? How should(?) it?

I recall the pain of working with a lag like @rofl0r mentions, however I feel that we're addressing too much of a rare special case here, and are introducing more inconsistency.

magnumripper:
> My primary reason to consider merging this would be not to discourage @rofl0r, who I hope would continue to contribute. 

My sentiments exactly!

solardiz:
Overall, I think let's not merge this. We end up wasting everyone's time, including @rofl0r's, by discussing detail such as (my question on) what happens in non-wordlist runs, which could even lead to a further revision of this hack.

I think @rofl0r can instead contribute to addressing #5184. We'd appreciate that. For #5184, it should be no change in what's accepted vs. rejected - just reject earlier. Thanks!

rofl0r:
right, i was about to close this myself after reading @solardiz's reply - overall it seems that nobody else is interested in this so having this merged just as a personal favor isn't worth it, especially considering that there might be unintended sideeffects by this change.

PR #5185, opened 2022-08-21
NSEC3: correctly handle records without a salt

anonion0:
Prior to this, NSEC3 hashes lacking a salt were not properly recognized.

I also added new test vectors for a few corner cases, including some w/o a salt.


solardiz (approved):
The changes here look OK to me, thanks!

(The code being changed looks worse, but that's a separate issue.)

anonion0:
Yeah tbh I haven't looked at this code in years. I imagine it's due for an overhaul.

I've also been meaning to add IDNA (Punycode) support, should I ever find the time for it.

PR #5187, opened 2022-08-21
Add support for Cardano legacy's encrypted secret keys. 

### LINKED ISSUE #5186: Add support for Cardano legacy's encrypted secret keys.
### Add support for Cardano legacy's encrypted secret keys.

In 2017 the old wallets used encrypted master keys for keeping the users funds safe.
Unfortunately, lot of users lost their mnemonic and forgot their spending/encryption password, that caused locking their funds forever.

The PR I created tackles this issues by trying to decrypt the encrypted keys.
Unfortunately, it requires a slight modification in the ed25519-donna package as it does support extended (64-byte length) secret keys. This minor modification adds the possibility of generating a ed25519 public key from an extended secret key.

The Cardano legacy's extended secret key is 128-byte length that contains: a
1. 64-byte length encrypted secret key (from entropy), a
2. 32-byte length ed25519 public key (not-encrypted) and a
3. 32-byte length chaincode from entropy (not-encrypted)

To decrypt an encrypted (master or any derived) key needs the following steps:
- `blake2b_256` hash of the password candidate,
- a 40-byte long key from `pbkdf2_hmac_sha512`, using a static salt, and the pasword hash,
- decrypt the encrypted secret using `ChaCha20` stream encryption alg with the derived key (as IV and Key) and
- generate a public from the decrypted candidate secret key and 
- compare it with the store public key. If the both match then the password is found.

Cardano had a second method to validate whether the secret key was encrypted or not, by using `Scrypt` with some random salt.
Though, using the `scrypt`, to find the encryption password, is much slower than the above and also it does not guarantee (due to some bug in the early wallet implementations) that password generated the scrypt based password hash can decrypt the encrypted key at all.

ilap:
Resolves #5186

### DIFF
--- a/doc/NEWS
+++ b/doc/NEWS
@@ -254,6 +254,8 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
 - Added suport for PBKDF2-SHA256 key derivation algorithm of PrivateAccess and 
   ENCDataVault. [sylvainpelissier; 2022]
 
+- Added support for Cardano legacy's encrypted secret keys. [ilap; 2022]
+
 Major changes from 1.8.0-jumbo-1 (December 2014) to 1.9.0-jumbo-1 (May 2019):
 
 - Updated to 1.9.0 core, which brought the following relevant major changes:
--- a/src/cardano_common.h
+++ b/src/cardano_common.h
@@ -1,0 +1,42 @@
+/*
+ * Common code for cracking Cardano 128-byte length legacy secret Keys (a.k.a XPrv).
+ *
+ * This software is Copyright (c) 2022, Pal Dorogi <pal dot dorogi at gmail.com>
+ * and it is hereby released to the general public under the following terms:
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted.
+ *
+ * This file takes replicated but common code, shared between the CPU
+ * and the GPU formats, and places it into one common location.
+ *
+ * This file is common for its future use, and currently there isn't an OpenCL format
+ * implementing this feature.
+ *
+ */
+#include "formats.h"
+
+#define FORMAT_TAG      "$cardano$"
+#define FORMAT_TAG_LEN  (sizeof(FORMAT_TAG) - 1)
+#define SK_LEN          64
+#define PK_LEN          32
+#define CC_LEN          32
+#define ESK_LEN         (SK_LEN + PK_LEN + CC_LEN)
+
+
+// Chacha20 parameters
+#define KEY_SIZE    32
+#define IV_SIZE     8
+#define BUF_SIZE    (KEY_SIZE + IV_SIZE)
+
+// Blake2B parameters
+#define PWD_HASH_LEN    32
+
+struct custom_salt {
+	// The 128-byte length legacy encryted key.
+	unsigned char esk[ESK_LEN];
+};
+
+extern struct fmt_tests cardano_tests[];
+int cardano_valid(char *ciphertext, struct fmt_main *self);
+void *cardano_get_salt(char *ciphertext);
--- a/src/cardano_common_plug.c
+++ b/src/cardano_common_plug.c
@@ -1,0 +1,84 @@
+/*
+ * Common code for cracking Cardano 128-byte length legacy secret Keys (a.k.a XPrv).
+ *
+ * This software is Copyright (c) 2022, Pal Dorogi <pal dot dorogi at gmail.com>
+ * and it is hereby released to the general public under the following terms:
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted.
+ *
+ * This file takes replicated but common code, shared between the CPU
+ * and the GPU formats, and places it into one common location.
+ *
+ * This file is common for its future use, and currently there isn't an OpenCL format
+ * implementing this feature.
+ *
+ */
+
+#include "common.h"
+#include "cardano_common.h"
+
+struct fmt_tests cardano_tests[] = {
+	//
+	// https://github.com/CardanoSolutions/ByronWalletRecovery/blob/63ac66ee3532b2a1c72a88c0aa726d0f1ce3ed7a/tools/inspect-keystore/index.js#L34
+	// XPrv is BIP32-Ed25519 key, derived from an extended Ed25519 key (64-byte sk + 32-byte pk)
+	// by appending a 32-byte length chaincode to it, and clear and extra bit ( esk[31] &= 0x20).
+	//      +---------------------------------+-----------------------+-----------------------+
+	//      | Extended Private Key (64 bytes) | Public Key (32 bytes) | Chain Code (32 bytes) |
+	//      +---------------------------------+-----------------------+-----------------------+
+	//      <------------ ENCRYPTED ---------->
+	// Initial version only accepts v1 scheme identified as "$1$", which means the payload is a
+	// single 128-byte length legacy encrypted master secret key.
+	{ "$cardano$1$b57361ebe335fa171a260fea7d3277579c212dc74fc2a408d6cbd8a6e7a847cab3c44c5fb190705ddd2698f2d5390798893349b4321e7474b1ce06c9d410b3d6055b42d4a95f19cb34b516a160a306c0eaef398e70ea91da450ccb2a7819e95b0102030405060708091011121314151617181920212223242526272829303132", "Secret1234" },
+	{ NULL }
+};
+
+int cardano_valid(char *ciphertext, struct fmt_main *self)
+{
+	char *ctcopy, *keeptr, *p;
+	int len;
+
+	if (strncmp(ciphertext, FORMAT_TAG, FORMAT_TAG_LEN) != 0)
+		return 0;
+
+	ctcopy = xstrdup(ciphertext);
+	keeptr = ctcopy;
+
+	ctcopy += FORMAT_TAG_LEN;
+
+	if ((p = strtokm(ctcopy, "$")) == NULL) // scheme
+		goto err;
+	if (strcmp(p, "1")) // only "1" is supported at the moment.
+		goto err;
+	if ((p = strtokm(NULL, "$")) == NULL) // 128-byte long esk
+		goto err;
+
+	len = strlen(p) ;
+	if (len != 2 * ESK_LEN)
+		goto err;
+	MEM_FREE(keeptr);
+	return 1;
+
+err:
+	MEM_FREE(keeptr);
+	return 0;
+}
+
+void *cardano_get_salt(char *ciphertext)
+{
+	static struct custom_salt cs;
+	char *ctcopy = xstrdup(ciphertext);
+	char *keeptr = ctcopy;
+	char *p;
+	int i;
+
+	memset(&cs, 0, sizeof(cs));
+	ctcopy += FORMAT_TAG_LEN; // skip over "$cardano$"
+	p = strtokm(ctcopy, "$"); // skip over "$1$"
+	p = strtokm(NULL, "$");
+	for (i = 0; i < ESK_LEN; i++)
+		cs.esk[i] = atoi16[ARCH_INDEX(p[i * 2])] * 16 + atoi16[ARCH_INDEX(p[i * 2 + 1])];
+
+	MEM_FREE(keeptr);
+	return (void *)&cs;
+}
--- a/src/cardano_fmt_plug.c
+++ b/src/cardano_fmt_plug.c
@@ -1,0 +1,237 @@
+/*
+ * JtR format to crack password protected Cardano 128-byte length legacy secret Keys (a.k.a XPrv).
+ *
+ * This software is Copyright (c) 2022, Pal Dorogi <pal dot dorogi at gmail.com>
+ * and it is hereby released to the general public under the following terms:
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted.
+ *
+ */
+
+#if FMT_EXTERNS_H
+extern struct fmt_main fmt_cardano;
+#elif FMT_REGISTERS_H
+john_register_one(&fmt_cardano);
+#else
+
+#include <string.h>
+
+#ifdef _OPENMP
+#include <omp.h>
+#endif
+
+#define OMP_SCALE 1
+
+#include "cardano_common.h"
+#include "blake2.h"
+#include "pbkdf2_hmac_sha512.h"
+#include "chacha.h"
+#include "ed25519.h"
+
+#define FORMAT_NAME         "Cardano Encrypted 128-byte Secret Key (a.k.a XPrv)"
+#define FORMAT_LABEL        "cardano"
+#ifdef SIMD_COEF_64
+#define ALGORITHM_NAME      "PBKDF2-SHA512/BLAKE2b/ChaCha20 " SHA512_ALGORITHM_NAME
+#else
+#if ARCH_BITS >= 64
+#define ALGORITHM_NAME      "PBKDF2-SHA512/BLAKE2b/ChaCha20  64/" ARCH_BITS_STR
+#else
+#define ALGORITHM_NAME      "PBKDF2-SHA512/BLAKE2b/ChaCha20  32/" ARCH_BITS_STR
+#endif
+#endif
+#define BENCHMARK_COMMENT   ""
+#define BENCHMARK_LENGTH    0x107
+#define BINARY_SIZE         0
+#define BINARY_ALIGN        1
+#define SALT_SIZE           sizeof(struct custom_salt)
+#define SALT_ALIGN          sizeof(uint32_t)
+#ifdef SIMD_COEF_64
+#define MIN_KEYS_PER_CRYPT  SSE_GROUP_SZ_SHA512
+#define MAX_KEYS_PER_CRYPT  SSE_GROUP_SZ_SHA512
+#else
+#define MIN_KEYS_PER_CRYPT  1
+#define MAX_KEYS_PER_CRYPT  1
+#endif
+#if !defined(PLAINTEXT_LENGTH) || PLAINTEXT_LENGTH > 64
+#undef PLAINTEXT_LENGTH
+#define PLAINTEXT_LENGTH    64
+#endif
+
+static char (*saved_key)[PLAINTEXT_LENGTH + 1];
+static int any_cracked, *cracked;
+static size_t cracked_size;
+
+static struct custom_salt *cur_salt;
+
+static void init(struct fmt_main *self)
+{
+	omp_autotune(self, OMP_SCALE);
+
+	saved_key = mem_calloc(sizeof(*saved_key), self->params.max_keys_per_crypt);
+	cracked_size = sizeof(*cracked) * self->params.max_keys_per_crypt;
+	any_cracked = 0;
+	cracked = mem_calloc(cracked_size, 1);
+}
+
+static void done(void)
+{
+	MEM_FREE(saved_key);
+	MEM_FREE(cracked);
+}
+
+static void set_salt(void *salt)
+{
+	cur_salt = (struct custom_salt *)salt;
+}
+
+static void cardano_set_key(char *key, int index)
+{
+	strnzcpy(saved_key[index], key, sizeof(*saved_key));
+}
+
+static char *get_key(int index)
+{
+	return saved_key[index];
+}
+
+static int crypt_all(int *pcount, struct db_salt *salt)
+{
+	const int count = *pcount;
+	int index;
+
+	if (any_cracked) {
+		memset(cracked, 0, cracked_size);
+		any_cracked = 0;
+	}
+	// KDF2 params
+	static const int kdf_rounds = 15000;
+	static const char kdf_salt[] = "encrypted wallet salt";
+	static int kdf_salt_len = sizeof(kdf_salt);
+
+	// ChaCha20 params
+	static const int chacha_rounds = 20;
+
+#ifdef _OPENMP
+#pragma omp parallel for
+#endif
+	for (index = 0; index < count; index += MIN_KEYS_PER_CRYPT) {
+		int i;
+
+		unsigned char kdf_out[MIN_KEYS_PER_CRYPT][BUF_SIZE];
+		unsigned char blake_out[MIN_KEYS_PER_CRYPT][PWD_HASH_LEN];
+
+#if SIMD_COEF_64
+		int lens[MIN_KEYS_PER_CRYPT];
+		unsigned char *pin[MIN_KEYS_PER_CRYPT]; // blake2b_256 hashed password
+		unsigned char *pout[MIN_KEYS_PER_CRYPT]; // 40-byte length KDF
+#endif
+
+		for (i = 0; i < MIN_KEYS_PER_CRYPT; ++i) {
+			// blake2b_256 hash of the password.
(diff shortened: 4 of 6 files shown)


solardiz (changes requested):
Thank you for the contribution, @ilap! Let's polish it a bit and then likely get it merged.

To ease review, what format's source files did you base yours on?

There are a few issues here right now:

In my testing, this fails self-test when run with more than one OpenMP thread. The failures are inconsistent - they're `FAILED (cmp_all(14))`, but with indices changing (`14` in this example, but it varies). This suggests there's a race condition or/and usage of uninitialized memory.

(The good news, though, is the `tezos` format still works fine for me after your changes to the shared code.)

There's only one test vector. Please add at least one more, for 2+ total. Some realistic issues will go undetected when there's only one test vector.

You do split out a few things into `common` files, but you don't actually have an OpenCL format yet - do you intend to contribute one with your next PR? If not, then at least comments need to be edited to say these files are `common` for their future use, and currently are not, and there isn't currently an OpenCL format. BTW, we do have all of the required primitives in OpenCL, so it shouldn't be too hard to add an OpenCL format implementing this.

Thank you for including our preferred copyright+license statement in the main `cardano_fmt_plug.c` file. Please also add that to other source files you're adding here.

The `ALGORITHM_NAME` is wrong. Please change it perhaps to:

```
#ifdef SIMD_COEF_64
#define ALGORITHM_NAME      "PBKDF2-SHA512/BLAKE2b/ChaCha20 " SHA512_ALGORITHM_NAME
#else
#if ARCH_BITS >= 64
#define ALGORITHM_NAME      "PBKDF2-SHA512/BLAKE2b/ChaCha20 64/" ARCH_BITS_STR
#else
#define ALGORITHM_NAME      "PBKDF2-SHA512/BLAKE2b/ChaCha20 32/" ARCH_BITS_STR
#endif
```

(The most time-consuming component goes first, and the SIMD'ness and bitness at the end corresponds to its implementation's use of CPU.)

I and/or others will likely have more feedback later, perhaps after you address the above.

Thanks again!

solardiz:
Somehow Circle CI doesn't run on this PR - given that it ran just fine on a nearby PR by someone else, perhaps their abuse detection somehow got (wrongly) triggered for @ilap? This happened for some of our contributors IIRC a year ago, but didn't happen since...

solardiz:
> this fails self-test when run with more than one OpenMP thread.

The below fixes the obvious issues, and makes tests pass for me (but we need more test vectors to be more certain, we're lucky the issue was detected at all with only one test vector):

```diff
+++ b/src/cardano_fmt_plug.c
@@ -125,8 +125,8 @@ static int crypt_all(int *pcount, struct db_salt *salt)
        for (index = 0; index < count; index += MIN_KEYS_PER_CRYPT) {
                int i;
 
-               static unsigned char kdf_out[MIN_KEYS_PER_CRYPT][BUF_SIZE];
-               static unsigned char blake_out[MIN_KEYS_PER_CRYPT][PWD_HASH_LEN];
+               unsigned char kdf_out[MIN_KEYS_PER_CRYPT][BUF_SIZE];
+               unsigned char blake_out[MIN_KEYS_PER_CRYPT][PWD_HASH_LEN];
 
 #if SIMD_COEF_64
                int lens[MIN_KEYS_PER_CRYPT];
@@ -151,12 +151,12 @@ static int crypt_all(int *pcount, struct db_salt *salt)
                }
 #endif
 
-               static ed25519ext_secret_key dsks[MIN_KEYS_PER_CRYPT];
-               static ed25519_public_key pk;
+               ed25519ext_secret_key dsks[MIN_KEYS_PER_CRYPT];
+               ed25519_public_key pk;
 
                for (i = 0; i < MIN_KEYS_PER_CRYPT; ++i) {
 
-                       static struct chacha_ctx ckey;
+                       struct chacha_ctx ckey;
 
                        // Decrypt the encrypted esk with the retrieved KDFs.
                        chacha_keysetup(&ckey, kdf_out[i], KEY_SIZE * 8);
```

solardiz:
Oh, also please avoid over-long commit message lines. Please split lines:

```
Add support for Cardano legacy's encrypted secret keys

Resolves #5186
```

solardiz:
> Somehow Circle CI doesn't run on this PR

Logged in to Circle CI, and I don't see this PR's jobs there at all. There are for other recent activities. Also, we appear to be very far from bumping into any resource limit this month.

Maybe just a glitch. Maybe they will pick this up on the next update to this PR.

ilap:
Sorry, different timezone.

> To ease review, what format's source files did you base yours on?

It's a mix and match, just had a look at some source codes and got hints from them to implement.

> You do split out a few things into `common` files, but you don't actually have an OpenCL format yet - do you intend to contribute one with your next PR? If not, then at least comments need to be edited to say these files are `common` for their future use, and

I think it will be `common` for a little while, but I might will create some `cl` code and some python tools for easier usability.

> (The good news, though, is the `tezos` format still works fine for me after your changes to the shared code.)

Yeah, it just adds an extra functionality, no breaking API changes.


> There's only one test vector. Please add at least one more, for 2+ total. Some realistic issues will go undetected when there's only one test vector.

Unfortunately, I could get only one proper testvector (encrypted and has its en/decryption key) by retrieving it from some old keystore file (`secret.key`) that was/is used for testing some old functionality of the officiall wallet.

> Oh, also please avoid over-long commit message lines. Please split lines:

Will do, thank you.

> Thank you for including our preferred copyright+license statement in the main `cardano_fmt_plug.c` file. Please also add that to other source files you're adding here.

Done

> (The most time-consuming component goes first, and the SIMD'ness and bitness at the end corresponds to its implementation's use of CPU.)

Done 

> The below fixes the obvious issues, and makes tests pass for me (but we need more test vectors to be more certain, we're lucky the issue was detected at all with only one test vector):

Thank you, I have removed the `static` keywords and it makes sense why it's failing.

I will commit the changes soon, thx again.

ghost:
Since no CicleCI job is running again, I'm running the upstream CI at https://app.circleci.com/pipelines/github/claudioandre-br/JohnTheRipper/478/workflows/1292c785-5593-44c9-9fb6-e6a6c2b1f04e

The patch adds trailing spaces. 

```
warning: 2 lines add whitespace errors.
```

ilap:
> Since no CicleCI job is running again, I'm running the upstream CI at https://app.circleci.com/pipelines/github/claudioandre-br/JohnTheRipper/478/workflows/1292c785-5593-44c9-9fb6-e6a6c2b1f04e
> 
> The patch adds trailing spaces.
> 
> ```
> warning: 2 lines add whitespace errors.
> ```

What tools do you use to check  whitespace errors and which parameters?
As my one does not recognised trailing spaces in comments.

ilap:
> Since no CicleCI job is running again, I'm running the upstream CI at https://app.circleci.com/pipelines/github/claudioandre-br/JohnTheRipper/478/workflows/1292c785-5593-44c9-9fb6-e6a6c2b1f04e
> 
> The patch adds trailing spaces.
> 
> ```
> warning: 2 lines add whitespace errors.
> ```

Fixed

solardiz:
> some python tools for easier usability.

Yes, and/or please add some documentation on how to use this. (Not necessarily in this same PR.)

> Unfortunately, I could get only one proper testvector (encrypted and has its en/decryption key) by retrieving it from some old keystore file (`secret.key`) that was/is used for testing some old functionality of the officiall wallet.

OK, so we'll have to accept having only one for now.

Can you please add the `secret.key` file and its corresponding password via a pull request to https://github.com/openwall/john-samples?

Please squash all commits here into one, and if you make any further changes please amend and force-push the one commit instead of adding fixup commits. (Alternatively, we can use GitHub's squash-and-merge, but that makes GitHub the committer per git history and it's a feature we normally keep disabled for this repo, to avoid inadvertent use.)

> Since no CicleCI job is running again, I'm running the upstream CI at https://app.circleci.com/pipelines/github/claudioandre-br/JohnTheRipper/478/workflows/1292c785-5593-44c9-9fb6-e6a6c2b1f04e

Thank you, @claudioandre-br! As I can see, this passed all of our tests now. Great.

ilap:
> Can you please add the `secret.key` file and its corresponding password via a pull request to https://github.com/openwall/john-samples?

Uploaded and a PR is created.

> Please squash all commits here into one, and if you make any further changes please amend and force-push the one commit instead of adding fixup commits. (Alternatively, we can use GitHub's squash-and-merge, but that makes GitHub the committer per git history and it's a feature we normally keep disabled for this repo, to avoid inadvertent use.)

Done

src/cardano_common.h
```diff
@@ -0,0 +1,50 @@
...
+ * and the GPU formats, and places it into one common location.
+ *
+ * This files is common for its future use, and currently there isn't an OpenCL format
+ * implementing this feature.
+ *
+ */
+
+#include "common.h"
+#include "formats.h"
+#include "jumbo.h"
```
solardiz: Out of these 3, probably only `#include "formats.h"` is needed here.

src/cardano_common.h
```diff
@@ -0,0 +1,50 @@
...
+ * This software is Copyright (c) 2022, Pal Dorogi <pal dot dorogi at gmail.com>
+ * and it is hereby released to the general public under the following terms:
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted.
+ *
+ * This file takes replicated but common code, shared between the CPU
+ * and the GPU formats, and places it into one common location.
+ *
+ * This files is common for its future use, and currently there isn't an OpenCL format
```
solardiz: Should be "This file" (singular), and ditto in the other file with that comment.

src/cardano_common.h
```diff
@@ -0,0 +1,50 @@
...
+	// master secret key a.k.a: XPrv.
+	uint32_t scheme;
+	// The 128-byte length legacy encryted key.
+	unsigned char esk[ESK_LEN];
+};
+
+extern struct fmt_tests cardano_tests[];
+int cardano_valid(char *ciphertext, struct fmt_main *self);
+void *cardano_get_salt(char *ciphertext);
+unsigned int cardano_iteration_count(void *salt);
```
solardiz: There isn't actually a `cardano_iteration_count()` - so drop the prototype.

src/cardano_common_plug.c
```diff
@@ -0,0 +1,85 @@
...
+		goto err;
+
+	if ((p = strtokm(NULL, "$")) == NULL)   // esk
+		goto err;
+
+	len = (strlen(p)) / 2;
+	if (len != ESK_LEN)
+		goto err;
+	MEM_FREE(keeptr);
+	return 1;
```
solardiz, line 60: Doesn't this allow any trailing garbage after the hash? Perhaps it should insist on no trailing garbage?

src/cardano_common_plug.c
```diff
@@ -0,0 +1,85 @@
...
+		return 0;
+
+	ctcopy = xstrdup(ciphertext);
+	keeptr = ctcopy;
+
+	ctcopy += FORMAT_TAG_LEN;
+	if ((p = strtokm(ctcopy, "$")) == NULL) // scheme
+		goto err;
+	if (!isdec(p))
+		goto err;
```
solardiz, line 52: If we currently only support `1` there, we should insist on that - so e.g. instead of `!isdec(p)`, we can `strcmp(p, "1")`.

Also, we don't have to decode that field into the salt structure until we use it, nor to have a field in there. We can just have a comment on its possible future use, here in `valid`.

src/cardano_fmt_plug.c
```diff
@@ -0,0 +1,244 @@
...
+#include "common.h"
+#include "formats.h"
+#include "params.h"
+#include "options.h"
+#include "jumbo.h"
+#include "cardano_common.h"
+#include "blake2.h"
+#include "pbkdf2_hmac_sha512.h"
+#include "chacha.h"
+#include "ed25519.h"
```
solardiz, line 30: A few of these `#include` are probably irrelevant, I guess at least `#include "options.h"`.

src/cardano_fmt_plug.c
```diff
@@ -0,0 +1,244 @@
...
+#define BINARY_ALIGN        1
+#define SALT_SIZE           sizeof(struct custom_salt)
+#define SALT_ALIGN          sizeof(uint32_t)
+#ifdef SIMD_COEF_64
+#define MIN_KEYS_PER_CRYPT  SSE_GROUP_SZ_SHA512
+#define MAX_KEYS_PER_CRYPT  (SSE_GROUP_SZ_SHA512 * 4)
+#else
+#define MIN_KEYS_PER_CRYPT  1
+#define MAX_KEYS_PER_CRYPT  4
+#endif
```
solardiz, line 55: I think 15000 iterations is so high that the `* 4` and a non-SIMD max of 4 are unneeded - can set them to 1 for greater program interactivity at no significant effect on performance.

src/cardano_fmt_plug.c
```diff
@@ -0,0 +1,244 @@
...
+#if !defined(PLAINTEXT_LENGTH) || PLAINTEXT_LENGTH > 15
+#undef PLAINTEXT_LENGTH
+#define PLAINTEXT_LENGTH    15
+#endif
```
solardiz, line 59: Why limit `PLAINTEXT_LENGTH` to 15 here? Looks wrong to me, perhaps copy-paste from elsewhere. What's the actual maximum this format can handle?

src/cardano_fmt_plug.c
```diff
@@ -0,0 +1,244 @@
...
+	static const int kdf_rounds = 15000;
+	static const char kdf_salt[] = "encrypted wallet salt";
+	static int kdf_salt_len = sizeof(kdf_salt);
+
+	// ChaCha20 params
+	static const int chacha_rounds = 20;
+
+#ifdef _OPENMP
+	#pragma omp parallel for
+#endif
```
solardiz, line 117: We normally don't indent `#pragma` directives - just keep them on the leftmost column. Ditto for the other one below.

src/cardano_fmt_plug.c
```diff
@@ -0,0 +1,244 @@
...
+		}
+
+		pbkdf2_sha512_sse((const unsigned char **)pin, lens, (unsigned char *)kdf_salt, kdf_salt_len, kdf_rounds, pout, BUF_SIZE, 0);
+#else
+			pbkdf2_sha512(blake_out[i], PWD_HASH_LEN,
+			              (unsigned char *)kdf_salt, kdf_salt_len, kdf_rounds, kdf_out[i], BUF_SIZE, 0);
+		}
+#endif
+
+		ed25519ext_secret_key dsks[MIN_KEYS_PER_CRYPT];
```
solardiz: Looks like the `dsks` array can be replaced with one `dsk` - only one element is used at a time. It's the same as `pk` and `ckey` in that respect - so for consistency don't use arrays for any.

`kdf_out` is different and actually needs to stay an array, because it's produced by a separate loop that can use SIMD, thus computing multiple KDF values concurrently.

src/ed25519-donna/ed25519-donna.c
```diff
@@ -40,19 +40,24 @@ ed25519_hram(hash_512bits hram, const ed25519_signature RS, const ed25519_public
...
+        ge25519_pack(pk, &A);
 }
 
+void ED25519_FN(ed25519_publickey)(const ed25519_secret_key sk, ed25519_public_key pk)
+{
+        ed25519ext_secret_key extsk;
+
+        ed25519_extsk(extsk, sk);
+        ed25519ext_publickey(extsk, pk);
+}
```
solardiz, line 60: You're changing tabs to spaces here - please don't, please keep the indentation consistent with the original.

solardiz (approved):
This is potentially merge-able now, but I'm suggesting some more changes for you to make before we merge.

You could also want to add an entry to `doc/NEWS`.

src/cardano_common.h
ilap: I have removed all unnecessary includes.

src/cardano_common.h
ilap: Fixed

src/cardano_common_plug.c
ilap: Fixed

src/cardano_common.h
ilap: Fixed

src/cardano_common_plug.c
ilap: Fixed

src/cardano_fmt_plug.c
ilap: Fixed

src/cardano_fmt_plug.c
ilap: Fixed

src/cardano_fmt_plug.c
ilap: Yeah, you're right, fixed.

src/ed25519-donna/ed25519-donna.c
ilap: Fixed

src/cardano_fmt_plug.c
ilap: Set to 64, as it's unrealistic to have longer user's password.

src/cardano_fmt_plug.c
ilap: I have removed all unnecessary ones.

doc/NEWS
```diff
@@ -1,5 +1,7 @@
 Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
 
+- Added support for Cardano legacy's encrypted secret keys. [ilap; 2022]
+
```
solardiz: This should be at the end of the list, around line 256.

src/cardano_common.h
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * This file takes replicated but common code, shared between the CPU
+ * and the GPU formats, and places it into one common location.
+ *
+ * This file is common for its future use, and currently there isn't an OpenCL format
+ * implementing this feature.
+ *
+ */
+
+#include "common.h"
```
solardiz: I don't see you use anything from `#include "common.h"` in this file.

src/cardano_common_plug.c
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * This file takes replicated but common code, shared between the CPU
+ * and the GPU formats, and places it into one common location.
+ *
+ * This file is common for its future use, and currently there isn't an OpenCL format
+ * implementing this feature.
+ *
+ */
+
+#include "cardano_common.h"
```
solardiz, line 19: I didn't mean you'd rely on includes by `cardano_common.h` to also include what's needed by files including it.

src/cardano_common_plug.c
```diff
@@ -0,0 +1,75 @@
...
+struct fmt_tests cardano_tests[] = {
+	//
+	// https://github.com/CardanoSolutions/ByronWalletRecovery/blob/63ac66ee3532b2a1c72a88c0aa726d0f1ce3ed7a/tools/inspect-keystore/index.js#L34
+	// XPrv is BIP32-Ed25519 key, derived from an extended Ed25519 key (64-byte sk + 32-byte pk)
+	// by appending a 32-byte length chaincode to it, and clear and extra bit ( esk[31] &= 0x20).
+	//      +---------------------------------+-----------------------+-----------------------+
+	//      | Extended Private Key (64 bytes) | Public Key (32 bytes) | Chain Code (32 bytes) |
+	//      +---------------------------------+-----------------------+-----------------------+
+	//      <------------ ENCRYPTED ---------->
+	{ "$cardano$b57361ebe335fa171a260fea7d3277579c212dc74fc2a408d6cbd8a6e7a847cab3c44c5fb190705ddd2698f2d5390798893349b4321e7474b1ce06c9d410b3d6055b42d4a95f19cb34b516a160a306c0eaef398e70ea91da450ccb2a7819e95b0102030405060708091011121314151617181920212223242526272829303132", "Secret1234" },
```
solardiz: I didn't mean you'd drop the `$1` component - I meant you'd only insist on the value being 1 in `valid` and then not parse/store it (and don't have that struct field, which you now correctly removed) until later. But OK, let's stay without the field as well.

src/cardano_fmt_plug.c
solardiz: I think we differ in what we consider "unnecessary". I meant you'd remove includes of stuff that you don't use. However, you seem to have removed everything without which the program won't build. That approach results in it building "by luck" - due to includes of other needed headers from some of the headers that you still include her. I think that's not ideal.

src/cardano_fmt_plug.c
```diff
@@ -0,0 +1,237 @@
...
+#else
+#define ALGORITHM_NAME      "PBKDF2-SHA512/BLAKE2b/ChaCha20  32/" ARCH_BITS_STR
+#endif
+#endif
+#define BENCHMARK_COMMENT   ""
+#define BENCHMARK_LENGTH    0x107
+#define BINARY_SIZE         0
+#define BINARY_ALIGN        1
+#define SALT_SIZE           sizeof(struct custom_salt)
+#define SALT_ALIGN          sizeof(uint32_t)
```
solardiz, line 48: `SALT_ALIGN` can now be 1 now that you removed the `uint32_t` field from there.

src/cardano_fmt_plug.c
solardiz: That's now how we do it in JtR. We determine the actual maximum (the lower of what our code supports and what the target software's code supports) and use that here unless there's performance advantage to use a smaller limit (shouldn't be the case here). Also, that `#if` and `#undef` logic should be unneeded here.

solardiz (approved):
I made some comments again, but for most of them I think we got to a point where it's quicker to merge this as-is and then for me or someone else to polish it a bit maybe as otherwise we'd keep running into misunderstandings on minor detail.

src/cardano_fmt_plug.c
solardiz: Given that you're passing passwords through `blake2b()`, which doesn't have a low limit, our global limit of 125 should apply. So just set `PLAINTEXT_LENGTH` to 125.

doc/NEWS
ilap: Thx, on it.

src/cardano_common_plug.c
ilap: I am reverting it  to $1 as it seems to me much clearer.

src/cardano_fmt_plug.c
ilap: We would need it as I reverted the 1.

src/cardano_common.h
ilap: fixing it

src/cardano_common_plug.c
```diff
@@ -0,0 +1,84 @@
...
+		return 0;
+
+	ctcopy = xstrdup(ciphertext);
+	keeptr = ctcopy;
+
+	ctcopy += FORMAT_TAG_LEN;
+
+	if ((p = strtokm(ctcopy, "$")) == NULL) // scheme
+		goto err;
+	if (strcmp(p, "1")) // only "1" is supported at the moment.
```
ilap, line 51: Ooops, fixing this.

solardiz:
I'll merge this now, and will possibly make further adjustments in a separate commit of my own. Thank you!

src/cardano_fmt_plug.c
solardiz: You didn't reintroduce the struct field, so we can have `SALT_ALIGN` at 1.

PR #5188, opened 2022-08-23
NSEC3: performance improvements, OpenMP and more

anonion0:
This should significantly speed up NSEC3 cracking, especially at lower iteration counts.

OpenMP support isn't tuned that well yet, but I doubt it's ever going to be great anyway given most zones these days use very low (<= 10) iteration counts (hence I marked it with `FMT_OMP_BAD`).

I wasn't entirely sure about removing `FMT_HUGE_INPUT`, but `formats.h` says it's for ciphertexts longer than 896 bytes, which does not apply here. Please let me know if I missed something.

Let me know if you need me to split these changes into separate PRs, or if you want me to squash the commits.

### DIFF
--- a/src/nsec3_fmt_plug.c
+++ b/src/nsec3_fmt_plug.c
@@ -60,10 +60,10 @@ john_register_one(&fmt_nsec3);
 
 struct salt_t {
 	size_t salt_length;
-	size_t zone_length;
+	size_t zone_wf_length;
 	uint16_t iterations;
 	unsigned char salt[NSEC3_MAX_SALT_SIZE];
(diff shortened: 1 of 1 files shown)


solardiz (approved):
Thanks. This looks about right to me.

I can't easily review the code invoked from `valid` for never accessing anything out-of-bounds and for performing sufficient validation that further code wouldn't do anything unsafe either. We'll need to rerun the fuzzer on this format, @AlekseyCherepanov. (And also fuzz the new `cardano` format.)

As it relates to coding style, we generally don't align to opening brace - so those changes made in here are inconsistent with other code we have. But then we do have plenty of coding style inconsistencies across the tree... so I don't mind.

PR #5191, opened 2022-08-27
Added cardano2john.py: exports secret.key to john format.

ilap:


run/cardano2john.py
```diff
@@ -0,0 +1,89 @@
...
+#
+# Export Cardano's legacy secret.key to John The Ripper format
+#
+# This software is Copyright (c) 2022, Pal Dorogi (ilap) <pal dot dorogi at gmail.com>
+# and it is hereby released to the general public under the following terms:
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted.
+#
+# Requireements: pip3 at OS level
```
solardiz: Typo.

Also, does this not work with pip3 at user level? Why not?

run/cardano2john.py
```diff
@@ -0,0 +1,89 @@
...
+            # Go through all user's keys a.k.a _usKeys and the the Wallet set.
+            keys = cbor[2] + cbor[3]
+            keys_len = len(keys)
+
+            if keys_len == 0:
+                print("No keys found in the keystore file.")
+            else:
+                out_file = "cardano.esks"
+                try:
+                    with open(out_file, "w") as fp:
```
solardiz: This is inconsistent with other 2john scripts, which output to stdout (and expect it to be redirected). I think we should be consistent.

run/cardano2john.py
```diff
@@ -0,0 +1,89 @@
...
+                        print(
+                            "Exporting",
+                            keys_len,
+                            "keys to the %s file%s" % (out_file, os.linesep),
```
solardiz: We don't use `os.linesep` in other scripts... but maybe we should?

run/cardano2john.py
```diff
@@ -0,0 +1,89 @@
...
+    except:
+        sys.stderr.write("Wrong secret key file or path: %s%s" % (filename, os.linesep))
+        sys.exit(1)
+
+
+if __name__ == "__main__":
+
+    if len(sys.argv) < 2:
+        sys.stderr.write(
+            "Usage: %s -f <Cardano's secret.key file (./secret.key)>%s"
```
solardiz: Looks unnecessarily complicated and inconsistent with other 2john scripts (although I think in some we do have that?)

Instead, we should just expect the filename (or multiple) on the command line, without requiring any option. So no need for `argparse` in this script yet. We can add that later if needed for something other than an input filename.

run/cardano2john.py
```diff
@@ -0,0 +1,89 @@
...
+    try:
+        with open(filename, "rb") as fp:
+            # cbor[0]: _usVssMaybe, not interesting
+            # cbor[1]: _usPrimKey,  not interesting
+            # cbor[2]: _usKeys, interested only the esk part, atm, [[esk, passwordHash],...,]
+            # cbor[3]: _usWalletSet, interested if there is any, but only in its [0]
+            # i.e.: _wusRootKey which is an esk
+            cbor = load(fp)
+
+            # Go through all user's keys a.k.a _usKeys and the the Wallet set.
```
solardiz: Typo: "the the"

solardiz (changes requested):
Thanks, and I'm sorry for the delayed review.

@exploide You might want to take a look as well. Thank you!

run/cardano2john.py
ilap: Fixed, sorry it would run on user level too.

run/cardano2john.py
ilap: Yeah, I refactored the code completely.

run/cardano2john.py
ilap: Not for print as it appends an OS dependent new line anyway.

run/cardano2john.py
ilap: Refactored

run/cardano2john.py
ilap: Fixed.

exploide (approved):
LGTM

solardiz (approved):

solardiz:
Thank you, @ilap and @exploide!

Issue #5182, opened 2022-08-05
rar2john should reject non-RAR files


siltal:
### Checklist ###
- [x] :1st_place_medal: I've read and understood these instructions;
  - This is not a support forum, it's a bug tracker. For questions and support, review postings on the [john-users mailing list](https://www.openwall.com/lists/john-users/).
- [ ] :thumbsup: I've tested using latest bleeding version from this repository.
  - Be clear about your environment and what you are doing. Share a sample hash or file that can be used to reproduce.
- [x] :confused: I'm confused and I need guidance.
  - Please, read the instructions at https://www.openwall.com/john/#lists, then join the list before posting.

### IMPORTANT ###
We expect only reports of issues with the latest revision found in this GitHub repository. We do not expect in here, and have little use for, reports of issues only seen in a release or in a distro package.

Attach details about your OS and about john, including:
- OS `Windows 10 Pro WorkStation 19044.1826`
- The output of `./john --list=build-info`.
```
Version: 1.9.0-jumbo-1
Build: cygwin 64-bit x86_64 AVX2 AC OMP
SIMD: AVX2, interleaving: MD4:3 MD5:3 SHA1:1 SHA256:1 SHA512:1
CPU tests: AVX2
CPU fallback binary: john-xop
OMP fallback binary: john-avx2-non-omp
$JOHN is /run/
Format interface version: 14
Max. number of reported tunable costs: 4
Rec file version: REC4
Charset file version: CHR3
CHARSET_MIN: 1 (0x01)
CHARSET_MAX: 255 (0xff)
CHARSET_LENGTH: 24
SALT_HASH_SIZE: 1048576
SINGLE_IDX_MAX: 2147483648
SINGLE_BUF_MAX: 4294967295
Effective limit: Number of salts vs. SingleMaxBufferSize
Max. Markov mode level: 400
Max. Markov mode password length: 30
gcc version: 7.4.0
OpenCL headers version: 2.2
Crypto library: OpenSSL
OpenSSL library version: 01010102f
OpenSSL 1.1.1b  26 Feb 2019
GMP library version: 6.1.2
File locking: fcntl()
fseek(): fseek
ftell(): ftell
fopen(): fopen
memmem(): System's
```
### Issue
- The command line I am using.
```
rar2john file.rar
```
 - its output
```
bad rar crypt version byte
```
> when I using `bandizip` to open rar file, it can list inner file in rar but encryption.
file size 616MB, too large to uploading.

solardiz:
[...]

PR #5192, opened 2022-08-30
rar2john: Fix logic flaws for non-RAR input files

magnumripper:
#5182.

Tested like this:
- All rar files in our sample repo still emits hashes.
- I have yet to find a false positive (tried with all files in `run`) and due to corrected logic the risk of it should be really slim.

### DIFF
--- a/src/rar2john.c
+++ b/src/rar2john.c
@@ -72,6 +72,14 @@
 /* Max file (path) name length, in characters */
 #define PATH_BUF_SIZE 256
 
+/* File magics */
+#define RAR_OLD_MAGIC      "\x52\x45\x7e\x5e"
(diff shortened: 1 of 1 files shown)


src/rar2john.c
```diff
@@ -256,7 +267,6 @@ static void process_file(const char *archive_name)
 			MEM_FREE(gecos);
 			if (process_file5(archive_name))
 				return;
-			fprintf(stderr, "! %s: Not a RAR file\n", archive_name);
```
magnumripper, line 259: That same message would now be printed from `process_file5()` before we got here.

solardiz (approved):

src/rar2john.c
magnumripper: I see now that probably goes for line 258 in new code as well. Will test more and amend this.

PR #5199, opened 2022-10-13
Zip-opencl: Autotune tweak for large data size

magnumripper:
Main (last) commit fixes autotune for very large data sizes.

The other commit adds fallbacks in case pinned (page locked) allocations fail. I wrote that one first - while it turned out those allocs weren't the problem it's a Good Idea™ to keep it.


magnumripper:
I just realized I need to amend `release_clobj()` as well, will do so soon.

magnumripper:
> I just realized I need to amend `release_clobj()` as well, will do so soon.

Fixed and tested (simulating failed pinned alloc).

ghost (approved):
Seems reasonable to me.

PR #5202, opened 2022-10-16
ENCDataVault: Increase signature length to reduce fault positive

sylvainpelissier:
As discussed in https://github.com/hashcat/hashcat/issues/3467 the number of fault positive can be reduced by looking at the following 4 bytes which represent the number of vaults which have been created on the machine. Tests have been updated.

### DIFF
--- a/src/encdatavault_common_plug.c
+++ b/src/encdatavault_common_plug.c
@@ -104,7 +104,9 @@ int valid_common(char *ciphertext, struct fmt_main *self, int is_pbkdf2)
 		goto err;
 	if ((p = strtokm(NULL, "$")) == NULL)   // Encrypted header
 		goto err;
-	if (hexlenl(p, &extra) != ENC_SIG_SIZE * 2 || extra)
+	if (hexlenl(p, &extra) > ENC_SIG_SIZE * 2 || extra)
+		goto err;
+	if (hexlenl(p, &extra) < (ENC_SIG_SIZE - 4) * 2 || extra)
 		goto err;
 	if (is_pbkdf2) {
 		if ((p = strtokm(NULL, "$")) == NULL)   // Salt length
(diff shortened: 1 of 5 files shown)


solardiz (changes requested):
Thanks. Some minor comments:
1. Please fix the typo in the commit message - should be "false" in place of "fault".
2. You're completely dropping support for "hashes" produced by the old script version. I thought the C code would support both old and new (detect the field length). I don't insist, though - this is up to you.
3. In the hashcat issue, you mention you actually exceeded 256 vaults in your experiments. I wonder if a user realistically could, too. So we might prefer to check only 6 bytes, not 7. Especially with PBKDF2, where it's unlikely to trigger a 6-byte collision.

sylvainpelissier:
I like the idea to be flexible in the signature length, I added that and the comparison on 6 bytes maximum as well.

src/encdatavault_common_plug.c
```diff
@@ -161,14 +164,15 @@ void *get_salt_common(char *ciphertext, int is_pbkdf2)
 	p = strtokm(NULL, "$");
 	cs.algo_id = atoi(p);
 	p = strtokm(NULL, "$");
-	for (i = 0; i < ENC_NONCE_SIZE; i++)
+	for (i = 0; i < ENC_IV_SIZE; i++)
 		cs.iv[i] = (atoi16[ARCH_INDEX(p[2 * i])] << 4) | atoi16[ARCH_INDEX(p[2 * i + 1])];
```
solardiz, line 168: Looks like only half of the IV was decoded before, yet somehow this did not matter. Why did it not matter?

solardiz (approved):

src/encdatavault_common_plug.c
sylvainpelissier: the second half is not used in the hashing algorithm.

PR #5210, opened 2022-11-04
PEM format: Logic fixes

magnumripper:
Some corrections after #5146

The 'if (!warned)' logic must only affect screen output, not actions (such as rejecting the ciphertext).
Different types of messages (kdf vs. prf here) now have separate flags for being warned for or not.
Finally, some warnings triggered by self-tests are now muted.


solardiz (approved):
Ouch, skipping part of the validation after first warning was indeed wrong.

PR #5211, opened 2022-11-04
Add "Compiler version" to --list=opencl-devices

magnumripper:
Example:
```
    (...)
    Device version:         OpenCL 3.0 CUDA
    Compiler version:       OpenCL C 1.2
    Driver version:         470.141.03 [recommended]
    (...)
```
"Device version" corresponds to CL_DEVICE_VERSION while "Compiler version" corresponds to CL_DEVICE_OPENCL_C_VERSION.  The example output is a bit weird with the latter lower than the former - it should mean we can't use OpenCL 2.0 on this device.  Perhaps they are not 100% compliant yet?

### DIFF
--- a/src/opencl_common.c
+++ b/src/opencl_common.c
@@ -3019,7 +3019,10 @@ void opencl_list_devices(void)
 			if (strstr(dname, "OpenCL 1.0")) {
 				printf(" <the minimum REQUIRED is OpenCL 1.1>");
 			}
-			printf("\n    Driver version:         %s\n",
+			clGetDeviceInfo(devices[sequence_nr], CL_DEVICE_OPENCL_C_VERSION,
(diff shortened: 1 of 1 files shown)


magnumripper:
https://stackoverflow.com/questions/67371938/nvidia-opencl-device-version

solardiz:
I think calling this "Compiler version" is wrong. It's more like "OpenCL version" or "Language spec".

magnumripper:
A correct description is "the highest OpenCL C version supported by the compiler for this device" but we're limited in length here. This would fit: `OpenCL version support:`. Is that OK?

solardiz:
@magnumripper Looks good to me, but please also update the example in the commit message.

magnumripper:
Good call, done.

solardiz (approved):

Issue #4964, opened 2021-12-23
Bug? [Nvidia 1070] format=NT-opencl fails to crack password within wordlist but format=NT does.


Promethus029:
Hi, this is my first time posting, so I apologize if something is incorrect. Bleeding-jumbo john was build about two hours ago with Regex enabled. Attempting to run '/opt/john/run/john --pot=john.pot --wordlist=rockyou-modified.txt --format=NT-opencl passwords.txt' does not yield Dark's password but finds the empty admin password.
![NT-opencl](https://user-images.githubusercontent.com/23091547/147169064-ca39fb6b-45b7-47ba-bc83-54c61a3268bb.png)
Re-running the command only adds the password at the edge of the GPU temp.
![NT-result-edit](https://user-images.githubusercontent.com/23091547/147169199-7e73a9c9-1827-4dfd-805a-ec1761771990.png)
However, running the command without opencl gives Dark's password.
![NT](https://user-images.githubusercontent.com/23091547/147167677-8569716e-7e80-48b9-8012-cf88028546e5.png)

Looking at console I see that the CPU is using MD4 128/128 XOP 4x2 while GPU is using MD4 opencl. Which it should?

I've been on the fence about posting because i don't know if this is an actual bug or I'm just using john incorrectly and this report shouldn't be here. Thank you for your time.

(john -test=0 reports all passed except for that tezos-opencl, which you know about.)

'john --show --format=NT-opencl' ->[checking-passwords.txt](https://github.com/openwall/john/files/7765793/checking-passwords.txt)
[john_build_info.txt](https://github.com/openwall/john/files/7765794/john_build_info.txt)
[lscpu-info.txt](https://github.com/openwall/john/files/7765795/lscpu-info.txt)
Nvidia Card -[opencl-devices-info.txt](https://github.com/openwall/john/files/7765796/opencl-devices-info.txt)
Hashes - [passwords.txt](https://github.com/openwall/john/files/7765797/passwords.txt)
file to crack hash - [rockyou-modified.txt](https://github.com/openwall/john/files/7765798/rockyou-modified.txt)

solardiz:
@Promethus029 Thank you for reporting. This does look like a bug - apparently, the correct password was detected by on-device comparison (which is how it'd get exactly to the beginning of the returned "range"), but not fully seen that way by host.

> except for that tezos-opencl, which you know about

[...]

Issue #4012, opened 2019-05-28
Device-side mask mode: Revisit the fmt_reset() logic


magnumripper:
This isn't quite perfected yet. It's a two-step rocket:

1. If mask positions or ranges changed at all and `FMT_MASK`, we have to call `crk_process_buffer()` followed by format's `reset()` for updating the GPU-side. Obviously, we currently do.
2. Most or all formats currently always (re-)auto-tune whenever the above happens. This is **not** needed unless the "mask multiplier" (the actual one, not the requested) changed.

So going from eg. `?a?a?d` to `?a?d?a` using a device-side mask format, we always need to call `reset()` but within that function (so unfortunately in every such format, unless I find some short-cut) we should have logic for detecting whether a new auto-tune is needed or not. It's only needed in case the new actual "mask multiplier" is significantly different from the previously used one. Simple example: Going from `?u?l?l` to `?l?l?l` would not change the multiplier so would not need re-tuning.

magnumripper:
This would ~~probably~~ close #3516 as well

solardiz:
Isn't this too invasive and risky to be "Definitely 1.9.0-jumbo-2 material"?

magnumripper:
I think not. It's not complicated really, and it's easily tested.

magnumripper:
Waiting is good. I'm starting to see the light. I'm pretty sure we can do ALL special logic in `opencl_autotune.h` and that's a terrific fact - all formats' `reset()` will be trivial (though a bit less so for internal mask formats currently but again I think I see the light). Still a couple of hours from posting first PR and I'm about to take a break now.

PR #5219, opened 2022-11-30
Assorted fixes

magnumripper:
I'll probably add more (unrelated) commits to this PR.

### DIFF
--- a/run/7z2john.pl
+++ b/run/7z2john.pl
@@ -11,13 +11,13 @@ use File::Basename;
 # magnum (added proper handling of BCJ et. al. and adapt to JtR use)
 
 # version:
-# 1.6
+# 1.9
 
 # date released:
 # April 2015
 
 # date last updated:
-# 3rd Dec 2021
+# July 13 2022
 
 # dependencies:
 # Compress::Raw::Lzma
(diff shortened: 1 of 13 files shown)


solardiz:
I'd easily approve the first 2 commits here, but the 3rd one is less obvious to me:

> [OpenCL NTLMv2 and MSCASH formats: Bug fix for reset()](https://github.com/openwall/john/pull/5219/commits/ae44ceb8eb2954778de188928a5acb207d04567b)
> 
> When running without mask, format would not reinitialize hash tables
> between self-tests and the real run, causing a hang at best and false
> negatives at worst.
> 
> Problem introduced with https://github.com/magnumripper/john/commit/bd7b8267764bfda5bce9df40cf1a6dfcf89e0329 (https://github.com/openwall/john/issues/4012).

The issue sounds similar to #4964, which I fixed in ddfcc27d7b2097afa34c3c300ca170290a680bbf and 2bf0b31278070fa5e072b9f986221f822f73944b, but your fix here is different. Maybe we should use the same approach here, and make the kernel + hash tables rebuild unconditional?

solardiz:
> Maybe we should use the same approach here, and make the kernel + hash tables rebuild unconditional?

Actually, seeing how I had done it in two commits back then, I don't mind the commit here introducing the extra condition first, and then dropping all of those conditions with a separate commit.

magnumripper:
Wouldn't using `num_loaded_hashes != db->salts->count` fail when the number of loaded hashes happens to be the same as the number of test vectors? Good thing that's not in there anymore.

We could make it unconditional (and be sure to avoid bugs like this one), but some systems might be very slow at building and/or tuning even trivial kernels so I'd rather not rebuild unless actually needed. When mask mode iterates over length it calls `reset()` between each length.

Instead of adding that static `*last_db` we could test it as `|| (db->real == db)` though.

loader.h:
```c
/*
 * Pointer to real db. NULL if there is none. If this db *is* the real db
 * this points back to ourself (db->real == db).
 */
       struct db_main *real;
```

magnumripper:
> Instead of adding that static `*last_db` we could test it as `|| (db->real == db)` though.

No wait, that would defeat the purpose - it would be more or less the same as unconditional. The current code in this PR is better.

solardiz:
> Wouldn't using `num_loaded_hashes != db->salts->count` fail when the number of loaded hashes happens to be the same as the number of test vectors?

Yes, and I recall concluding that was fine. However, the reasoning was tricky and I did prefer not to have the condition.

solardiz:
> I recall concluding that was fine.

Maybe I was wrong. I now see the logic was skipping not only kernel rebuild, but also a `prepare_tables`. Also, there are many more potentially affected formats listed in my comments in #4964 - @magnumripper, please take a look. If we're only patching two kernels here now, we need to have a comment (I mean here on GitHub) explaining why we don't need to patch the rest of these.

magnumripper:
> there are many more potentially affected formats listed in my comments in #4964 - @magnumripper, please take a look.

The bug hit me with NTLMv2-opencl (crashed unless using mask) and I did look for it in more formats - that's how I found it in mscash-opencl. The rest did look fine to me but it wasn't a super thorough review. IIRC I reviewed all formats with `FMT_MASK` and `prepare_*`.

magnumripper:
> We could make it unconditional (and be sure to avoid bugs like this one), but some systems might be very slow at building and/or tuning even trivial kernels so I'd rather not rebuild unless actually needed. When mask mode iterates over length it calls `reset()` between each length.

I had a look at this now - it seems mask mode is smart enough not to call `reset()` unless needed. So perhaps we should make it unconditional!

magnumripper:
I reviewed all `FMT_MASK` formats again, and dropped obsolete code from a bunch of them. They were creating a database out of test vectors but we're doing that in core nowadays.

Now a good deal of testing ahead...

magnumripper:
> Now a good deal of testing ahead...

They all seem fine. I tried `-test=0`, `-test-full=0`, `-test` with or without `-no-mask` and all available tests in the TS, plus some random manual testing.

DES-opencl also has obsolete handling of test vectors, but I'm not sure I dare touching it.

solardiz:
@magnumripper Perhaps we should get this PR merged already, at the same time with today's contributed PR? Thanks!

ghost (approved):
Since the PR has been tested by magnum and Solar has given his LGMT to merge, I will merge and create a new dev package.

Issue #5078, opened 2022-02-16
blockchain2john.py: Add support for v4


cryptoprogramer:
blockchain2john.py problem with blockchain wallet v4

today i tried to extract the hash of blockchain wallet v4 using blockchain2john.
initially I couldn't extract the hash and gave this error
My Wallet Version 2 or 3 seems to be used, adding --json option is required!

later I removed the following information from the beginning of data inside json file:

``` {"pbkdf2_iterations":5000,"version":4,"payload":"``` 

and used --base64 command to get the hash

 and it gave me a strange hash which is 17500 characters long. 
and hashcat recognized it as -m 12700 mode but couldn't crack it with a known password.

is this problem because of hashcat ? or due to strange hash or faulty hash from blockchain2john??
the hash  starts with: $blockchain$v2$5000$8704$

**Comment by solardiz** [comment]
@isieditors Can that hash you obtained be cracked with `john`?

**Comment by cryptoprogramer** [comment]
john didn't crack it although the correct password was inserted in the password file
I used the following command:

```C:\john\run>john --wordlist=password.txt blockchaintestv4.txt```

and got the following output
```
Warning: detected hash type "Blockchain", but the string is also recognized as "blockchain-opencl"
Use the "--format=blockchain-opencl" option to force loading these as that type instead
Using default input encoding: UTF-8
Loaded 1 password hash (Blockchain, My Wallet [PBKDF2-SHA1 AES 256/256 AVX2 8x])
Will run 8 OpenMP threads
Press 'q' or Ctrl-C to abort, almost any other key for status
Warning: Only 43 candidates left, minimum 64 needed for performance.
0g 0:00:00:00 DONE (2022-02-16 17:18) 0g/s 1433p/s 1433c/s 1433C/s dddf@@ee55
Session completed
```

another thing to add is that btcrecover cracks the hash obtained by:

```extract-blockchain-main-data.py```

and gives the following output:

```Wallet Type: btcrpass.WalletBlockchain
Wallet difficulty: 5,000 PBKDF2-SHA1 iterations
2022-02-16 08:33:32 : Using 8 worker threads
Notice: Loading File:  .\passwords\password.txt
Notice: Finished File:  .\passwords\password.txt
| 36  elapsed: 0:00:01  rate:  28.28  P/s

Password found: 'gggg@@GGGG1'
```

solardiz:
[...]

PR #5226, opened 2023-01-17
Blockchain2john.py updated. Now you can extract wallet v4 hash hassle free

TCRetriever:


magnumripper:
Thanks but there are numerous issues with this PR, including but not limited to:
1. You seem to have violated line endings in `blockchain2john.py` - every single line was changed?
2. The file `blockchain2john-main.zip` was inadvertently added and should be dropped.
3. We/you should add the change Solar made to `blockchain_common_plug.c` in this same PR.
4. You need to rebase upon latest bleeding, dropping any local merges. Or we can "squash and merge".
5. Update doc/NEWS.
6. Include a reference to #5078 in the commit description - For example ending it with `Closing #5078`

solardiz:
@magnumripper While @witee-star's help is appreciated, I think we should just create our own clean PR fixing these issues.

TCRetriever:
@solardiz that's a great idea. It's best you should create your own PR.

PR #5227, opened 2023-01-17
Add support for Blockchain v4

magnumripper:
Add support for Blockchain v4. See #5078


solardiz (approved):
Looks good to me for now.

The NEWS update is incomplete - we made more newsworthy changes since the previous release than are in NEWS even with this update - but it's good to have as a draft that we'll hopefully edit further before the next release. Thanks!

PR #5232, opened 2023-02-17
user level john-local.conf

xambroz:
The current .include '$JOHN/john-local.conf' is actually loading the system-wide file from the john package/installation typically from the /usr/share/john/john-local.conf To allow users to have possibility to load their own local additions to the global john.conf config,  it would be nice to attempt to load one from their home directory. 

```
┌──(kali㉿kali)-[~/exercises/john_the_ripper] 2023-02-17 17:48:59 +0100 
└─$ strace john --incremental=digits --stdout 2>&1  |grep -e 'open.*\.conf' 
openat(AT_FDCWD, "/home/kali/.john/john.conf", O_RDONLY) = 3 
openat(AT_FDCWD, "/usr/share/john/dynamic_disabled.conf", O_RDONLY) = 4
openat(AT_FDCWD, "/usr/share/john/unisubst.conf", O_RDONLY) = 4
openat(AT_FDCWD, "/usr/share/john/korelogic.conf", O_RDONLY) = 4
openat(AT_FDCWD, "/usr/share/john/hybrid.conf", O_RDONLY) = 4
openat(AT_FDCWD, "/usr/share/john/dumb16.conf", O_RDONLY) = 4
openat(AT_FDCWD, "/usr/share/john/dumb32.conf", O_RDONLY) = 4
openat(AT_FDCWD, "/usr/share/john/repeats16.conf", O_RDONLY) = 4
openat(AT_FDCWD, "/usr/share/john/repeats32.conf", O_RDONLY) = 4
openat(AT_FDCWD, "/usr/share/john/dynamic.conf", O_RDONLY) = 4
openat(AT_FDCWD, "/usr/share/john/dynamic_flat_sse_formats.conf", O_RDONLY) = 5
openat(AT_FDCWD, "/usr/share/john/regex_alphabets.conf", O_RDONLY) = 4
openat(AT_FDCWD, "/usr/share/john/john.local.conf", O_RDONLY) = -1 ENOENT (No such file or directory) 
openat(AT_FDCWD, "/usr/share/john/john-local.conf", O_RDONLY) = -1 ENOENT (No such file or directory) 
openat(AT_FDCWD, "/home/kali/.john/john-local.conf", O_RDONLY) = 4
```

### DIFF
--- a/run/john.conf
+++ b/run/john.conf
@@ -4195,8 +4195,12 @@ void init()
 # the new john-local.conf if you so choose.
 .include '$JOHN/john.local.conf'
 
-# include john-local.conf (This file can be created by user, to override defaults in this john.conf file)
+# include john-local.conf (typically /usr/share/john/john-local.conf)
+# This file can be created by administrator, to override defaults in this john.conf file for whole machine
 .include '$JOHN/john-local.conf'
+
+# include john-local.conf (This file can be created by user, to override defaults in this john.conf file)
+.include '~/.john/john-local.conf'
 
 # include john-local.conf in local dir, it can override john.conf, john-local.conf (or any other conf file loaded)
 # This is disabled by default since it's a security risk in case JtR is ever run with untrusted current directory


solardiz:
We shouldn't accept this PR as-is because the tilde expansion is only supported in system-wide builds, whereas this adds the `.include` line unconditionally. So in a non-system-wide build we'd end up having:

```
open("~/.john/john-local.conf", O_RDONLY) = -1 ENOENT (No such file or directory)
```

which is not only weird and wasteful, but is also a security concern in case the current directory is untrusted.

Separately, looking at the implementation in `path.c`, I see that the tilde and `$JOHN` expansion will silently not be done (resulting in the same behavior as above) if the user home directory path could not be determined or is too long. We should probably revise the code so that it instead skips those file accesses (doesn't fit in the current API, where `path_expand*()` can't fail, so would need changes in the many callers) or terminates the program (treat those errors in `path_init()` as fatal, calling `error();` on them instead of the current silent `return;`). However, these issues are separate and don't have to be fixed in the same PR (and definitely should not be in the same commit).

Finally, let's wrap commit description text at 75 chars per line. (And title line is 72 max, which is already the case here.)

solardiz:
We should probably make the tilde expansion available in non-system-wide builds as well, but even then I'm unsure if we want this `~/.john/john-local.conf` lookup performed by default in such builds - do we? In those builds, `$JOHN` is the `run` directory, so we do not process `~/.john/john.conf` in them, and it'd be inconsistent to process `~/.john/john-local.conf`. So maybe we need a way to skip a `.include` (or any portion of the config) in a (non-)system-wide build. This gets tricky.

xambroz:
Hello,
I believe having this option only in the system wide is not problem for the run. 
In local installation the file just wont be found, but the processing will continue.
It will be looking for the characters ~/ literaly and wont't find them, but that should not be a problem.

In my system I also do not have the 
openat(AT_FDCWD, "/usr/share/john/john.local.conf", O_RDONLY) = -1 ENOENT (No such file or directory) 
openat(AT_FDCWD, "/usr/share/john/john-local.conf", O_RDONLY) = -1 ENOENT (No such file or directory) 

But yes ... you are right its a security risk when executed in untrusted local directory where folder "~" might exist.

xambroz:
> I'm unsure if we want this `~/.john/john-local.conf` lookup performed by default in such builds - do we?

yes we do ... same way as we are looking for the local full configuration in ~/.john/john.conf
The logic is same. Directory ~/.john is controlled by user. 
$JOHN/ is controlled only by the administrator.
For user to define only the ~/.john/john-local.conf has definitely the same benefits as $JOHN/john-local.conf has for the local compilation = user can have his specific overrides modularly in his own file configuration and he doesn't have to sync and compare his whole configuration to the whole configuration of the stock package.

solardiz:
> It will be looking for the characters ~/ literaly and wont't find them, but that should not be a problem.

Like I wrote, this is "a security concern in case the current directory is untrusted." Like you note, the code will be looking for literal directory named `~` in the current directory. If an attacker were to create that, they'd control John.

> > I'm unsure if we want this ~/.john/john-local.conf lookup performed by default in such builds - do we?

> yes we do ... same way as we are looking for the local full configuration in ~/.john/john.conf
> The logic is same. Directory ~/.john is controlled by user.

Yes, it's the same logic, and `~/.john/john.conf` is _not_ checked in non-system-wide builds, so it would be _inconsistent_ to be checking `~/.john/john-local.conf` in those.

xambroz:
Then we would probably have to move this logic from the config to the code right?
Like the ~/.john/john.conf is also checked from the code as first thing to do.

So should we check the ~/.john/john-local.conf and $JOHN/john-local.conf as last things from code as well instead of having that in config?

But yes it wont be the same - now the all-in-one config has got possibility to opt-out from loading any additonal files by simply not including them right?

solardiz:
> Then we would probably have to move this logic from the config to the code right?

Yes, that's an option.

> now the all-in-one config has got possibility to opt-out from loading any additonal files by simply not including them right?

Right, and this is why in jumbo we introduced `john-local.conf` with different semantics (being parsed on top of previous configs, not preventing their parsing). We can replicate the same intentionally different semantics in code.

xambroz:
Even if "~/.john/john-local.conf" would not make it to the upstream default config
 ... parsing the ~/ to users home might be still beneficial even in the local compilation directory.
I can imagine someone would like to use his favorite ruleset no matter whether he is currently working with system wide install or his local git snapshot.

solardiz:
> parsing the ~/ to users home might be still beneficial even in the local compilation directory.

I agree. However, if `~/.john/john-local.conf` would then be parsed, we'd immediately need to decide on whether we also want `~/.john/john.conf` parsed in such builds, which would be a change from the current behavior that could even break things on systems that use both a system-wide and a user-local build (where the user previously relied on `~/.john/john.conf` only overriding settings for the system-wide build).

So I'd rather somehow skip the `.include` of `~/.john/john-local.conf` in non-system-wide builds.

solardiz:
Given that this PR is not suitable for merging as-is and no clear decision on a different approach has been made and no progress is being made, I think let's just close it. Sounds right?

magnumripper:
Agreed. @xambroz if you need the change, open an issue for it and we'll discuss until consensus.

PR #5235, opened 2023-02-21
Cracker: Keep status.*_count in sync with crk_db

solardiz:
Turns out the code I added for detailed status a year ago updated the copies of password and salt count in `status` too early - before the cracker would update the original values being copied - which resulted in a lag for the reported values. In fact, with no further guesses the reported values would remain stale for the rest of the session. This PR/commit fixes it.


solardiz:
The change of how `status.guess_count` is updated is just for consistency, and should be a no-op.

PR #5242, opened 2023-03-07
Add OpenCL support for krb5tgs

magnumripper:


magnumripper:
I'm stumped by the circleci problem:
```
Testing: krb5pa-md5-opencl, Kerberos 5 AS-REQ Pre-Auth etype 23 [MD4 HMAC-MD5 RC4 OpenCL, UTF-8]... Build log: Compilation started
Compilation done
Linking started
Linking done
Device build started
Device build done
Kernel <krb5pa_md5> was not vectorized
Done.
FAILED (get_key(0))
```
On Linux I can't reproduce at all (the circleci test *is* linux) but on my macbook, I actually can - sometimes. It's intermittent. I didn't touch **any** code that would affect this, yet I can't reproduce it in bleeding-jumbo.
```
../run/john -stress-test=0 -form:krb5pa-md5-opencl -enc:utf8
```
I did confirm that the minor bugfix in `formats.c` didn't affect this - problem remains with it reverted.

ghost:
Maybe some detail below will make your life easier:

The environment is:
- OMP_NUM_THREADS: 2
- it is a --test-full=0

To test john using the same CI environment you can:
```
docker run -it --rm -v $(pwd):/host claudioandre/john:opencl
# then 
cd /host
gcc --version
# and so on

```

Did you get any useful information using a `printf` inside `get_key()`? Is it a buffer overflow or something really crazy?

magnumripper:
I'm debugging it. It seems the test is actually not only about `set/get_key()` as I thought, `crypt_all()` is indeed called in between (also changing the key index due to internal compare). So it can actually be some code I did touch.

ghost:
Something like this:
```
export OMP_NUM_THREADS=2
../run/john --test-full=0 -form:opencl -enc:utf8
```

magnumripper:
Nailed it
```diff
diff --git a/run/opencl/krb5pa-md5_kernel.cl b/run/opencl/krb5pa-md5_kernel.cl
index d952360a3..d9f6b7fa2 100644
--- a/run/opencl/krb5pa-md5_kernel.cl
+++ b/run/opencl/krb5pa-md5_kernel.cl
@@ -206,6 +206,9 @@ void krb5pa_md5_final(const uint *K,
        /* First decrypt just one block for early rejection. */
        rc4(rc4_ctx, salts, plain, 16);
 
+       /* Ensure old data isn't messing with early rejection */
+       K2[0] = 0;
+
        /* Known-plain ASCII timestamp as in "20121123161534Z" for 2012-11-23 16:15:34 UTC */
        if (cleartext[14] != '2' || cleartext[15] != '0')
                return;
```
I did remove a similar reset, thinking it couldn't possibly be needed 😝

src/formats.c
```diff
@@ -607,9 +607,9 @@ static char* is_key_right(struct fmt_main *format, int index,
 
 	if (i == -1) {
 		if (options.verbosity > VERB_LEGACY)
-			snprintf(err_buf, sizeof(err_buf), "cmp_one(%d) %s", match, ciphertext);
+			snprintf(err_buf, sizeof(err_buf), "crypt_all(%d) %s", index + 1, ciphertext);
 		else
-			sprintf(err_buf, "cmp_one(%d)", match);
+			sprintf(err_buf, "crypt_all(%d)", index + 1);
 		return err_buf;
 	}
```
magnumripper: @solardiz If you don't have time to review the rest, perhaps you can just agree or disagree on this change of wording. I always got confused by `FAILED (cmp_one(0))` actually not pointing to a problem with `cmp_one` but with `crypt_all` returning zero after failing internal compare or early rejection.

solardiz: @magnumripper I disagree, or at least am not yet convinced. Sure this could have been confusing, but before your change we consistently report the function name as of when we determined that something went wrong, not function name where things _presumably_ actually went wrong. With your change, we become inconsistent. Also, with your change the output would no longer tell us that `match` was 0, so it'd probably be less useful for our analysis of the failure (assuming that whoever looks into the issue is not confused).

solardiz: While we're at it, maybe when `match == count && count == index + 1` we can just test `cmp_one` for `index` once, omitting the `for` loop? Or do we have any formats that can reorder the outputs while keeping their count? Not material for this PR, though - just bringing up the idea here.

magnumripper: When match was 0, `i == -1` here and we never even called `cmp_one` yet we state `FAILED (cmp_one(0))`. What good is that?

We could add some text that `crypt_all` returned zero, but there's no other way `crypt_all()` can fail so it's kind of given.

magnumripper: Oh, wait the loop goes backwards. This means we can reach `-1` in other cases. My change is not good but my point is still valid.

magnumripper: How about reverting it and then insert a new test after `cmp_all` right before the for loop, that catches `match == 0` and says something like `FAILED (crypt_all(n) zero return)`?

magnumripper: I committed that now.

solardiz: I'm OK with the separate check for `match == 0` like you have implemented now. Thanks.

PR #5246, opened 2023-03-09
NT-long OpenCL: Support lengths up to 125

### LINKED ISSUE #5245: NT-long-opencl
@solardiz, I have a trivial patch to NT-opencl that bumps its max length to 59 (or more). Due to the extra overhead and perhaps more because we can no longer reverse MD4 steps, it makes the format about 9% slower even at single-block lengths. I'm pondering two alternatives:

a) The changes are `#if PLAINTEXT_LENGTH > 27`. I could commit the changes except the actual change of `PLAINTEXT_LENGTH`. This is a no-op but makes it fairly easy to enable the support for longer passwords and rebuild.

b) I could create a separate NT-long-opencl (using same files but two format structs).

I'm not quite sure which way to go - what do you think?

BTW I also played with the idea to actually store two sets of binaries, with and without reversed steps, and really try to minimize overhead. But I think it's overkill - passwords this long are definitely used but they are not very common so they can be attacked separately anyway.

magnumripper:
1 block of MD4 is up to 27 characters.
2 blocks is 59, 3 is 91, 4 is 123 and 5 is 125 (due to core max).

As long as we bump it over 27 we don't seem to gain any speed by limiting it to less than 125 characters.

Closes #5245


solardiz:
> As long as we bump it over 27 we don't seem to gain any speed by limiting it to less than 125 characters.

I guess you mean with mask mode only. Transfers from host probably do become slower, but then those non-mask speeds are such that the attack better be run using the CPU format. However, then there's the case of small mask (e.g., one or two mask characters added to host-provided base words), where usage of OpenCL does help, yet transfers from host also affect the overall speed.

magnumripper:
> > As long as we bump it over 27 we don't seem to gain any speed by limiting it to less than 125 characters.
> 
> I guess you mean with mask mode only. Transfers from host probably do become slower, but then those non-mask speeds are such that the attack better be run using the CPU format. However, then there's the case of small mask (e.g., one or two mask characters added to host-provided base words), where usage of OpenCL does help, yet transfers from host also affect the overall speed.

Right. We do have "compressed" buffer transfer though, so transfers only become slower when there are (many) actually long words in there. BTW I see this format doesn't utilize early partial transfer of the key buffer, I should try adding it as well.

magnumripper:
At 13777559 we're back to a single format, and single-block speed seem to have (best of ~10 runs each on 2080ti) less than 1% penalty. I will investigate the "binary size 8" idea further but I don't really expect it to really fly so we might want to merge at this point (maybe we want to anyway).

Oh, and I'll also experient with partial key buffer transfer, but that too is somewhat bug prone so would be separate anyway and may come later.

ghost:
> we might want to merge at this point (maybe we want to anyway).

The patch is clean, small, and well organized. Merge it as is (to test and mature new ideas) is a very good option.

solardiz (approved):
Assuming this passes tests, it looks good to me and can be merged.

solardiz:
I think we have unnecessarily too much code divergence here:

```c
#if PLAINTEXT_LENGTH > 27
		if (md4_size > 27)
			nt_crypt_long(hash, nt_buffer, md4_size);
		else
#endif
			nt_crypt(hash, nt_buffer, md4_size);
```

If `md4_size > 27` varies within a 64 candidate password block, the code above will probably result in a 3x+ slowdown (compared to all passwords being `<= 27`), whereas it should likely be possible to have 2x+ (below 3x) by having the threads execute the same code at the same time for one of the limbs of the longer passwords (assuming they fit in 2 limbs).

I arrive at the 3x+ figure as follows: the `md4_size > 27` condition will result in an execution mask being set (instead of a conditional branch, which is only possible when all threads follow the same path), and so the two "calls" into `nt_crypt_long` and `nt_crypt` are made sequentially with the mask inverted. It's 2 limbs + 1 limb = processing cost equal to that of 3 limbs.

In other words, the high-level separation into `nt_crypt_long` vs. `nt_crypt`, while very efficient when all threads follow the same code path, unnecessarily duplicates and masks some code that could have been shared by threads even when they don't follow fully the same path (but do for one limb).

_If_ addressing this, please keep the commits so far as-is, but add a commit addressing this concern. I didn't look into how exactly it can be addressed, I just suspect that it can be by implementing the check inside `nt_crypt`.

OTOH, with usage of on-device mask such divergence should be very rare, and when on-device mask is not used the bottleneck is in the communication with host and in the host code. So the overall difference between these approaches is probably tiny (much smaller than 3 vs. 2).

magnumripper:
Unfortunately it seems to get a significant performance hit on gfx900 [Radeon RX Vega] (super's AMD-APP 2766.4). It went from 20467M to only 13930M. Limiting from 125 to 59 didn't help much (14163M).

I can't really see why it happens but it's a blocker, right? Perhaps we'll have to go back to a separate nt-long format. It would be nice to know performance with current AMD drivers though.

Maybe we could detect devices before host code decides on max. length, but that would be very confusing for users.

magnumripper:
> I think we have unnecessarily too much code divergence here

That sounds plausible. I'm not quite sure how to address it though. Need to think.

magnumripper:
OK I think I see how to do it. Let's always use the original code for the first block. Right before we stop and skip steps, we put the branch. So we either continue with one or more extra blocks (using normal MD4 macros) or we do the single-block special magic in the end.

We'll lose a tiny bit of performance for single-block crypts as we can no longer assume last word is zero. But adding branches for that is likely worse.

solardiz:
> I can't really see why it happens but it's a blocker, right?

Yes, it's a pretty bad performance regression. In case it's caused by code size increase, my suggestion to share more code between the paths could help there. Having less code total could also help the compiler reduce register pressure.

> Let's always use the original code for the first block. [...] we can no longer assume last block is zero.

What about using the original code for the last block instead?

magnumripper:
> What about using the original code for the last block instead?

That will instead switch from using INIT_A..INIT_D in the first steps (and even optimized at that but I suspect the compiler would do it for us anyway) to reading them from the buffer in the first steps.

magnumripper:
> > What about using the original code for the last block instead?
> 
> That will instead switch from using INIT_A..INIT_D in the first steps (and even optimized at that but I suspect the compiler would do it for us anyway) to reading them from the buffer in the first steps.

Perhaps we can do a stranger thing: Start with the first half of the original MD4, then a branch for possibly more blocks "in the middle of it". But it would be hard to follow and perhaps bug prone.

solardiz:
Is there any way to hint the compiler to keep code specific to longer passwords separately, like `if (unlikely(md4_size > 27))`? Does OpenCL have anything like that?

magnumripper:
> Is there any way to hint the compiler to keep code specific to longer passwords separately, like `if (unlikely(md4_size > 27))`? Does OpenCL have anything like that?

I never heard of that. I suspect some compilers might have it but others will fail because it's not in the standard AFAIK.

solardiz:
Can we just try [__builtin_expect](https://llvm.org/docs/BranchWeightMetadata.html#builtin-expect) and see whether/where it works and then possibly use it conditionally?

magnumripper:
I implemented it like you said, with original code calculating the last or only block. Right now I need three branches in there.

> Can we just try [__builtin_expect](https://llvm.org/docs/BranchWeightMetadata.html#builtin-expect) and see whether/where it works and then possibly use it conditionally?

Interestingly enough it works on Mac (all devices), and on Linux both with AMD and nvidia. I'm not seeing a lot of difference though, but a little: AMD now does 14786M with full length support of 125.

Edit: POCL also supports it. Given that virtually all drivers use llvm under the hood AFAIK, there's probably not many platforms that don't!

magnumripper:
> I implemented it like you said, with original code calculating the last or only block. Right now I need three branches in there.

Strangely enough, that very change ruined performance on Apple with AMD Radeon Pro Vega 20, from 6G c/s to 2G c/s. But that platform is so flakey we should care much about it - Linux/AMD is more important.

solardiz:
> I implemented it like you said, with original code calculating the last or only block. Right now I need three branches in there.

It's a pity we lose optimizations in the first few steps, such as the 0x77777777 constant. I think we can accept a little longer divergence, keeping the first few steps specialized (based on whether a/b/c/d are the constants or not) in the if/else above. We should as our first priority optimize for the <=27 case, and reduce divergence as a second priority.

magnumripper:
Yes I already have versions that do exactly that. Still too much hit on AMD. I'm currently trying to reverse a tad more of the MD4 in order to compensate but I fear I'll need to go the 64-bit-binary route (which, once it's all finished and bug free is a very good thing anyway of course) to get it really good.

magnumripper:
> I'm currently trying to reverse a tad more of the MD4

No, that's not possible.

> I fear I'll need to go the 64-bit-binary route

This would make it possible though.

magnumripper:
I dropped the other branches and put the best one here.

Currently we have two separate formats, but nt-long-opencl is faster than the old nt-opencl. It's just that the new nt-opencl is even faster, so we might just want to keep them separate. Anyway that's a later decision and easy to do then.

Best speed on super's "gfx900 [Radeon RX Vega]" (up from 20744M or so) is now 25059M (nt-opencl) and 19845M (nt-long-opencl). On a real crack against single hash it does 25526M (nt-opencl) and 20421M (nt-long-opencl). Those speeds did not drop when attacking two hashes.

Speeds on my 2080ti went from 32458M or so, to 35026M (nt-opencl) and 31215M (nt-long-opencl) but even faster speeds are seen in actual cracking (even over 51G for a single hash, over 40G for two hashes, using the "short" format).

magnumripper:
This might look like invasive changes but I'm confident with it. Bear in mind that all changes to `bt*[.ch]`, the many "unrelated" format files (well they are related to the perfect hash tables) and most of `opencl_hash_check*.[ch]` are *pure refactoring* of dangerous global names such as "hash_table_size".

Also, once I dug into it this wasn't particularly hard. There was no guesswork involved as I could look at LM-opencl for how to build and use the 64-bit tables. I ran it with canary code for an hour or so, everything seems to be great.

magnumripper:
I couldn't stop, so I added a commit that moves the bitmap lookup to as soon as at all possible for the non-long format. Tested on AMD (no regression, maybe slight boost) and nvidia (another 1G boost on 2080ti, to 36019M c/s benchmark). Edit: I've seen 37280M.

I also threw in a commit implementing early partial transfer of key buffer, every 2 MB of data.

magnumripper:
We're definitely keeping them two separate formats (for now at least). The difference in real cracking (with no candidates actually longer than 27) is just too high.

magnumripper:
I now saw a case of 64-bit collision (once in a LOT of tests). The collision is obviously only there when hash is reversed. It was during tests against 5M loaded hashes (max. size bitmaps). This is what we print:
```
False positive or 64-bit collision detected in cmp_exact(39)
  expected : 666ab666 93b79e79 92ad3ff4 cd46d5d9 
calculated : 666ab666 93b79e79 8dc1d349 61ec2523 
unreversed : 678dfbcd 71b15860 908afa8c 439b07ea 
Plaintext  : sensing?kc
```
We should probably just drop that printout, it’s not really interesting now that I confirmed we don’t have a bugful lot of them.

**Edit**: Unreversed it's are a mere 32-bit collision, I added code and reproduced:
```
False positive or 64-bit collision detected in cmp_exact(39)
  expected : 666ab666 93b79e79 92ad3ff4 cd46d5d9 
calculated : 666ab666 93b79e79 8dc1d349 61ec2523 
unreversed   expected : 678dfbcd 71b15860 908afa8c 439b07ea 
unreversed calculated : 678dfbcd e3918143 8b9e8ee2 d7405833 
Plaintext  : sensing?kc
```
I will put that printout code in `#if DEBUG` and push here.

magnumripper:
No small tests show any regression. Here's result from 5M loaded hashes. The difference get smaller but new code always wins (even nt-long over bleeding):

I used pwned-passwords-ntlm-v4 5M last hashes (worst count) and some meaningless wordlists repeated to get a suitable size for test (like a minute or two).

#### wordlist (389M lines), no mask acceleration
bleeding: 16787Kp/s
new code: 18768Kp/s
nt-long: 18046Kp/s (and cracked more)

#### mask
bleeding: 5432Mp/s
new code: 5674Mp/s
nt-long: 5432Mp/s

#### wordlist (77M lines) +mask
bleeding: 4720Mp/s
new code: 5248Mp/s
nt-long: 5023Mp/s (and cracked more)

I expected the smaller bitmap to do more good, but at some point the birthday paradox should become significant - perhaps here. After all, we have 5M hashes in the 64-bit space, and calculate billions of hashes per second comparing against them. Even when not producing a full 64-bit collision as mentioned above, it has to get deeper into the compare code on GPU before rejecting.

magnumripper:
I found this problem with current code:
```
$ ../run/john -test -form:nt-opencl -lws=1 -gws=1
Device 3: AMD Radeon Pro Vega 20 Compute Engine
Benchmarking: NT-opencl [MD4 OpenCL/mask accel]... FAILED (cmp_one(1))
```
Only happens with `-gws=1`, not any higher. Doesn't happen in bleeding.

magnumripper:
> I found this problem with current code:

Fixed. Also rebased upon current bleeding

magnumripper:
I need to re-tune the selection of bitmap size, compare steps, and whether it's local or not. I see a bad regression at certain numbers of loaded hashes and a good boost at others.

magnumripper:
> I need to re-tune the selection of bitmap size, compare steps, and whether it's local or not. I see a bad regression at certain numbers of loaded hashes and a good boost at others.

Sayantan's code has it in steps. For nvidia, the bitmap mask is 0x7fff up to 5100 loaded hashes, then 0xffff up to 10100 loaded hashes, then (for some reason) lowered to 0x7fff again(?) up to 20100 hashes. Above that we never use local memory but the hardcoded steps go on for 250100, 1500100 and 2700100 hashes and above that it's a formula. I suspect he spent most time optimizing for AMD back then.

0x7fff works great for a few hashes, but has 1 false positive per 11 candidates with 5100 hashes loaded. Not so great. I'm thinking we shold always use a formula, aiming for less than 1 false positive (in the bitmap) out of 128 candidates (perhaps better at low loaded counts), or something like that. I'm experimenting with that now.

magnumripper:
> I need to re-tune the selection of bitmap size, compare steps, and whether it's local or not. I see a bad regression at certain numbers of loaded hashes and a good boost at others.

Done. I now see **no** regression in any situation for nt-opencl, sometimes (on high counts of loaded hashes) it's a lot faster. And in many situations, nt-long-opencl is totally up to par - at least on nvidia.

> Sayantan's code has it in steps. For nvidia, the bitmap mask is 0x7fff up to 5100 loaded hashes, then 0xffff up to 10100 loaded hashes, then (for some reason) lowered to 0x7fff again(?) up to 20100 hashes.

It turned out the 20100 case used more bitmap "steps" so had an equivalent size. I had to add code for a 4-step bitmap (peeking at similar existing code in the 8-step bitmap for 128-bit hashes), then just slight tuning of the existing selections. 

BTW my testing has involved hash counts just below and just above the mentioned thresholds, such as 10000 and 11000.

> Above that we never use local memory but the hardcoded steps go on for 250100, 1500100 and 2700100 hashes and above that it's a formula.

I'm planning on replacing all that threshold code with a formula for any hash count. In fact I already have the code but I don't want to include it in this PR, it needs much more testing. It's fairly straight-forward, based on `log2(loaded_hash_count)` and then some logic for picking number of "steps" and sometimes preferring local memory use over bitmap size.

magnumripper:
@solardiz while I consider this code stable now, the 64-bit hash table didn't do as much for speed as I hoped it would. The bitmap lookups are just equal of course, but we do calculate a couple less full MD4 steps. Despite that, the boost is there but not often significant IRL. For huge numbers of loaded hashes, we obviously use half the memory for the hash tables but that's not a large part of the total used, and in GPU context it's very small figures anyway (eg. 8.8 MB gain for 11 M hashes loaded).

But I think we can merge this now. I'll revisit optimizations and tweaks later - I can't see get why we didn't get a better boost.

magnumripper:
I now did extensive regression testing on AMD as well. For both AMD and nvidia, this code is same speed or faster at any number of loaded hashes I've tested (many sizes from 1 .. 5,500,000). @solardiz I intend to merge this soon unless you want to review it.

I have very promising new code for bitmap parameters selection (it picks optimal values for "any" number of loaded hashes based on math, whereas the old code has several "gaps" in the threshold checks, so may pick far too small bitmaps rendering excessive false positives with resulting expensive hash table lookups) but it'll be separate and later.

solardiz:
Hi @magnumripper. I'm sorry I didn't have time to review this yet. I was going to take at least a cursory look, but kept postponing. I still might kind of review, but please feel free to merge without waiting.

Regarding testing and benchmarking, I suggest that you try with very high hash counts as well. IIRC, we previously tested 320M (or was it 306M?), which took something like 11G of GPU RAM - perhaps repeat something similar for old vs. new code. What's the maximum that fits in your 2080Ti?

run/opencl/opencl_misc.h
```diff
@@ -38,6 +38,15 @@ typedef uint32_t host_size_t;
 #define NULL ((void*)0)
 #endif
 
+/* Change the below to '#if 0' if your platform can't handle __builtin_expect */
+#if 1
```
solardiz: It's not great to require code edit on non-LLVM. Maybe there's an `#ifdef` we can check to detect LLVM-based OpenCL?

run/opencl/opencl_misc.h
```diff
@@ -38,6 +38,15 @@ typedef uint32_t host_size_t;
...
+#define likely(x)       __builtin_expect((x), 1)
```
solardiz: We'll need to remember to only pass a Boolean expression to `likely()`, and not say a pointer (for implied check for non-NULL). We could use `!!` in the macro to avoid this, but you'd need to retest for any performance regressions.

solardiz (approved):
I just skimmed. This is too much to review for real, and I'd need to duplicate your research and testing to review this properly.

solardiz:
GTX 570 with driver 319.23:

```
$ ../run/john -te -form=nt-opencl,nt-long-opencl
Device 1: GeForce GTX 570
Benchmarking: NT-opencl [MD4 OpenCL/mask accel]... LWS=256 GWS=61440 (240 blocks) x2600 DONE
Raw:	2045M c/s real, 2035M c/s virtual

Benchmarking: NT-long-opencl [MD4 OpenCL/mask accel]... LWS=256 GWS=61440 (240 blocks) x2600 DONE
Raw:	1636M c/s real, 1628M c/s virtual
```

Prior to this PR's changes:

```
Device 1: GeForce GTX 570
Benchmarking: NT-opencl [MD4 OpenCL/mask accel]... LWS=256 GWS=61440 (240 blocks) x2600 DONE
Raw:	1888M c/s real, 1888M c/s virtual
```

solardiz:
Tahiti 1.0 GHz:

New:

```
$ ./john -test -form=nt-opencl,nt-long-opencl -dev=4
Device 4: Tahiti [AMD Radeon HD 7900 Series]
Benchmarking: NT-opencl [MD4 OpenCL/mask accel]... Options used: -I opencl -cl-mad-enable -D__GPU__ -DDEVICE_INFO=138 -D__SIZEOF_HOST_SIZE_T__=8 -DDEV_VER_MAJOR=1642 -DDEV_VER_MINOR=5 -D_OPENCL_COMPILER -D OFFSET_TABLE_SIZE=26 -D HASH_TABLE_SIZE=45 -DUCS_2 -D NUM_LOADED_HASHES=44 -D NUM_INT_KEYS=1 -D SELECT_CMP_STEPS=2 -D BITMAP_MASK=0x7ffffU -D USE_LOCAL_BITMAPS=0 -D IS_STATIC_GPU_MASK=0 -D CONST_CACHE_SIZE=65536 -DENC_RAW -DENC_RAW -DPLAINTEXT_LENGTH=27 -D LOC_0=-1 -D LOC_1=-1 -D LOC_2=-1 -D LOC_3=-1 ./opencl/nt_kernel.cl
Build log: Error: Undeclared function index 1024

Error building kernel ./opencl/nt_kernel.cl. DEVICE_INFO=138
0: OpenCL CL_BUILD_PROGRAM_FAILURE (-11) error in opencl_common.c:1297 - clBuildProgram
```

That's a weird error. @magnumripper would you bisect maybe?

Old:

```
$ ./john -test -form=nt-opencl -dev=4
Device 4: Tahiti [AMD Radeon HD 7900 Series]
Benchmarking: NT-opencl [MD4 OpenCL/mask accel]... LWS=64 GWS=32768 (512 blocks) x6760 DONE
Raw:    6171M c/s real, 155058M c/s virtual
```

Two  reruns:

```
Benchmarking: NT-opencl [MD4 OpenCL/mask accel]... LWS=64 GWS=65536 (1024 blocks) x6760 DONE
Raw:    6206M c/s real, 256953M c/s virtual

Benchmarking: NT-opencl [MD4 OpenCL/mask accel]... LWS=64 GWS=32768 (512 blocks) x6760 DONE
Raw:    6171M c/s real, 155058M c/s virtual
```

solardiz:
> Build log: Error: Undeclared function index 1024

Apparently, that referred to `__builtin_expect`. Changing to `#if 0`, I get:

```
[solar@well run]$ ./john -test -form=nt-opencl,nt-long-opencl -dev=4
Device 4: Tahiti [AMD Radeon HD 7900 Series]
Benchmarking: NT-opencl [MD4 OpenCL/mask accel]... LWS=64 GWS=131072 (2048 blocks) x6760 DONE
Raw:	8293M c/s real, 420872M c/s virtual

Benchmarking: NT-long-opencl [MD4 OpenCL/mask accel]... LWS=64 GWS=131072 (2048 blocks) x6760 DONE
Raw:	6202M c/s real, 310116M c/s virtual

2 formats benchmarked.
[solar@well run]$ ./john -test -form=nt-opencl,nt-long-opencl -dev=4
Device 4: Tahiti [AMD Radeon HD 7900 Series]
Benchmarking: NT-opencl [MD4 OpenCL/mask accel]... LWS=64 GWS=131072 (2048 blocks) x6760 DONE
Raw:	8334M c/s real, 280581M c/s virtual

Benchmarking: NT-long-opencl [MD4 OpenCL/mask accel]... LWS=64 GWS=131072 (2048 blocks) x6760 DONE
Raw:	6202M c/s real, 310116M c/s virtual

2 formats benchmarked.
[solar@well run]$ ./john -test -form=nt-opencl,nt-long-opencl -dev=4
Device 4: Tahiti [AMD Radeon HD 7900 Series]
Benchmarking: NT-opencl [MD4 OpenCL/mask accel]... LWS=64 GWS=131072 (2048 blocks) x6760 DONE
Raw:	8334M c/s real, 280581M c/s virtual

Benchmarking: NT-long-opencl [MD4 OpenCL/mask accel]... LWS=64 GWS=131072 (2048 blocks) x6760 DONE
Raw:	6171M c/s real, 310116M c/s virtual

2 formats benchmarked.
```

run/opencl/opencl_misc.h
solardiz: FWIW, here's how we have these in `passwdqc`:

```c
#ifdef __GNUC__
#define force_inline    __attribute__ ((always_inline)) inline
#define likely(x)       __builtin_expect(!!(x), 1)
#define unlikely(x)     __builtin_expect(!!(x), 0)
#else
#define force_inline    inline
#define likely(x)       (x)
#define unlikely(x)     (x)
#endif
```

We could use similar in JtR's host code as well.

For use in OpenCL, I _guess_ we'll need:

```c
#ifdef __llvm__
#define likely(x)       __builtin_expect(!!(x), 1)
#define unlikely(x)     __builtin_expect(!!(x), 0)
#else
#define likely(x)       (x)
#define unlikely(x)     (x)
#endif
```

but this is completely untested.

run/opencl/opencl_misc.h
magnumripper: I'll investigate. Perhaps even the likely/unlikely macros already exist where applicable (alas I doubt that), so we could just check for them. Also, I'm not sure that thing is very important with current code, I should test that as well.

run/opencl/opencl_misc.h
magnumripper: According to https://clang.llvm.org/docs/LanguageExtensions.html#feature-checking-macros we should try this:
```c
#ifndef __has_builtin
#define __has_builtin(x) 0  // Compatibility with non-clang compilers.
#endif

#if __has_builtin(__builtin_expect)
#define likely(x)       __builtin_expect(!!(x), 1)
#define unlikely(x)     __builtin_expect(!!(x), 0)
#else
#define likely(x)       (x)
#define unlikely(x)     (x)
#endif
```
This can't fail other than possibly not detecting some platform that *has* it. But it should work for "Clang 10" and later, whatever that means in terms of llvm.

solardiz: If we determine that `#ifdef __llvm__` also works, I'd prefer that. Clang 10+ sounds unnecessarily too recent, and with OpenCL this isn't exactly clang.

magnumripper: The above works fine (as in positively detects it) on all my MacOS devices, as well as Linux/nvidia. I'm going with that.

solardiz:
Still same speeds on ancient NVIDIA:

```
Device 1: GeForce GTX 570
Benchmarking: NT-opencl [MD4 OpenCL/mask accel]... LWS=256 GWS=61440 (240 blocks) x2600 DONE
Raw:	2035M c/s real, 2035M c/s virtual

Benchmarking: NT-long-opencl [MD4 OpenCL/mask accel]... LWS=256 GWS=61440 (240 blocks) x2600 DONE
Raw:	1628M c/s real, 1636M c/s virtual

2 formats benchmarked.
```

and now passes test without any local edit on Tahiti (obviously, with same speeds as before).

magnumripper:
> Regarding testing and benchmarking, I suggest that you try with very high hash counts as well. IIRC, we previously tested 320M (or was it 306M?), which took something like 11G of GPU RAM - perhaps repeat something similar for old vs. new code. What's the maximum that fits in your 2080Ti?

The bitmap size(s) selection logic will handle it just the same (maxes out at 512M bitmap, i.e. 64 MB). I'm thinking we'll actually disable bitmaps *completely* above whatever threshold where we get too many false positives from it - it only hurts performance (and suck memory) from that point.

Theoretically I thought this would actually happen already at eg. 32M loaded hashes: At that point, with the max. of 512M bitmap, we have an "effective mask" of 5 bits unless I calculate it wrongly? But it seems that even just a couple (literally) of "bits" does help - I need to investigate more.

run/opencl/opencl_misc.h
magnumripper: > If we determine that `#ifdef __llvm__` also works

Please try that on your GTX 570 to start with. BTW I put things like `#warning "Notice: using __builtin_expect"` in there to see what's going on - but some drivers just quietly disregards that (and even `#error (...)`) while otherwise doing what they should. The only surefire way to see if we do get the built-ins is to write some actual error code in the else clause.

magnumripper: `#if __llvm__` works fine on MacOS. The nvidia is busy loading hundreds of millions of hashes 🤣 but anyway I suggest we simply do it like this:
```c
#ifndef __has_builtin
#define __has_builtin(x) 0
#endif

#if __llvm__ || __has_builtin(__builtin_expect)
(...)
```

magnumripper:
> Theoretically I thought this would actually happen already at eg. 32M loaded hashes: At that point, with the max. of 512M bitmap, we have an "effective mask" of 5 bits unless I calculate it wrongly? But it seems that even just a couple (literally) of "bits" does help - I need to investigate more.

OK, 32M hashes doesn't even get the max, it gets "only" a 256M bitmap - but it's probably too small. According to my debug code's calculations it's effectively 3 bits (0x7). The empirical results confirm that this figure is in the correct ballpark, with an actual 1/9 false positives in a run (the debug kernel is actually counting them):
```
Loaded 33554432 password hashes with no different salts (NT-opencl [MD4 OpenCL])
offset table size 47437676, hash table size 268730008, loaded hashes size 268435456
33554432 hashes: bitmap size 1x268435456 bits, 33554432 bytes, mask 0xfffffff, effectively 3 bits (0x7)
LWS=256 GWS=69632 (272 blocks) x9025
0g 0:00:01:31 DONE (2023-03-28 20:54) 0g/s 2210Mp/s 2210Mc/s 74182TC/s Dev#2:70°C util:100% laac>|..laac>|
Session completed.
201183043750 crypts, fp from bitmap: 1/9 (11.75%), 23641282113 hash table lookups
```
I'll sort out the exact figures and limits later (doesn't really matter for this PR) but anyway I'm pretty sure that at some point we're much better off not using the bitmap at all - and *definitely* so at 320M loaded hashes...

**EDIT** sorry, the above was the experimental code with new algorithms. Here's bleeding-jumbo - just with the debug code added:
```
Loaded 33554432 password hashes with no different salts (NT-opencl [MD4 OpenCL])
offset table size 47437644, hash table size 268730008, loaded hashes size 536870912
33554432 hashes: bitmap size 1x268435456 bits, 33554432 bytes, mask 0xfffffff, effectively 3 bits (0x7)
LWS=256 GWS=69632 (272 blocks) x9025 
0g 0:00:01:31 DONE (2023-03-28 21:31) 0g/s 2210Mp/s 2210Mc/s 74182TC/s Dev#2:69°C util:100% laac>|..laac>|
Session completed. 
201183043750 crypts, fp from bitmap: 1/9 (11.75%), 23640549466 hash table lookups
```
The selections were the same though, and the speed - but bleeding uses more GPU memory for full binaries.

magnumripper:
> The bitmap size(s) selection logic will handle it just the same (maxes out at 512M bitmap, i.e. 64 MB)

Actually the code (even the current in bleeding) has it wrong - if bitmap size really goes to the hardcoded limit of 512M, the kernel will fail with `#error BITMAP_MASK too large` in a macro assertion, because we're hitting 2**32 and current code use 32-bit integers.

The actual number of hashes you and/or Sayantan tested at some point appears to be 320294464, it's even in there:
```
$ git grep -FB1 'This many number of hashes'
bt.c-   if (num_ld_hashes > 320294464)
bt.c:           fprintf(stderr, "This many number of hashes have never been tested before and might not succeed!!\n");
```
...but if that worked, the limit must have been different at the time: We can only go to 256M. I think I will leave this as-is for now - like I said I'm planning more changes soon (in much smaller portions than this PR) and they will get it right - plus possibly disable bitmaps at some point.

**EDIT** the hardcoded limit is **512MB bitmap in bytes** which means 4 G bits (hence overflowing int32). But what I wrote above is basically correct - 320294464 would end up oversized and current kernel would bug out.
```
Build log: <program source>:42:2: error: BITMAP_SIZE_BITS_LESS_ONE too large
#error BITMAP_SIZE_BITS_LESS_ONE too large
 ^
```
So the correct kernel limit is 256 MB bitmap, which is 2 Gbits and a mask of 0x7fffffff. Besides, I doubt a bitmap that large will be more effective than just looking up the two integers, but I will test that. After all, such lookup includes modulus operations that might be expensive.

solardiz:
Maybe add a comment on `log2_128` and `pow128` that "faster algorithms exist, but these are optimized for size and simplicity". BTW, we do implement a faster algorithm in `powi` in `charset.c`.

run/opencl/opencl_misc.h
solardiz: Retested on the legacy systems, still works.

solardiz:
> lookup includes modulus operations that might be expensive.

A way to speed them up would be to precompute (at runtime, but out of loop - perhaps on host) a multiplicative reciprocal, then compute the remainder via division-optimized-into-multiplication, multiplication, and subtraction. Here's an example by @ch3root:

```c
unsigned mod3(unsigned x) { return x - (x * 2863311531ull >> 33) * 3; }
```

magnumripper:
> Maybe add a comment on `log2_128` and `pow128` that "faster algorithms exist, but these are optimized for size and simplicity". BTW, we do implement a faster algorithm in `powi` in `charset.c`.

Yeah they don't need to be fast at all. That `powi` is simple enough though - I'd just use uint128_t instead of double. Perhaps I'll throw it in while I amend something else.

> > lookup includes modulus operations that might be expensive.
> 
> A way to speed them up would be to precompute (at runtime, but out of loop - perhaps on host) a multiplicative reciprocal, then compute the remainder via division-optimized-into-multiplication, multiplication, and subtraction. Here's an example by @ch3root:
> 
> ```c
> unsigned mod3(unsigned x) { return x - (x * 2863311531ull >> 33) * 3; }
> ```

That is interesting, provided I can wrap my head around it. On the other hand, my tests with massive numbers of loaded hashes seem to indicate we actually do pretty fine with the bitmaps, contrary to what I thought. Especially with some tweaks I have in mind. More on that later. I seem to be limited by host RAM more than anything else right now (and we already mentioned ways to tackle that).

solardiz:
> provided I can wrap my head around it.

Multiplication by the large constant turns the integer number into fixed-point and at the same time multiplies it by the desired fraction. The constant is the fraction of 1.0 in that fixed-point notation. The right shift converts the fixed-point number back to integer (drops the fractional part).

solardiz:
> ```c
> unsigned mod3(unsigned x) { return x - (x * 2863311531ull >> 33) * 3; }
> ```

This can also be done without subtraction:

```c
unsigned mod3(unsigned x) { return (((x * 2863311531ull) & ((1ull << 33) - 1)) * 3) >> 33; }
```

Instead of taking the integer part of the fixed-point quotient, this takes its fractional part, brings it to the target range for the remainder, and then takes the integer part of that. It's probably almost the same performance - still two multiplies, one shift, and one other cheap operation (subtraction or bitwise-AND).

If we replace 33 with 32 and change the constant accordingly, then maybe the mask can be optimized out (just use the low-half register as input to the second multiplication) and the right shift sometimes too (just use the high-half register). This also works in my testing, but only for `x` up to `2**31-1`:

```c
unsigned mod3(unsigned x) { return (((x * 0x55555556u) & ((1ull << 32) - 1)) * 3) >> 32; }
```

Edit: here's an updated version that works for the full 32-bit input range:

```c
unsigned mod3(unsigned x) { return ((((x + 1) * 0x55555555u) & ((1ull << 32) - 1)) * 3) >> 32; }
```

or the same written in a way exposing MAD:

```c
unsigned mod3(unsigned x) { return (((x * 0x55555555u + 0x55555555u) & ((1ull << 32) - 1)) * 3) >> 32; }
```

On an arch with 32-bit registers, like we effectively have on GPUs, this is probably just two 32x32 to 64 widening multiplies in a row, and no other operations - just using the right registers.

Edit 2: unfortunately, generalizing this approach still fails for me for high inputs:

```c
#include <stdio.h>
#include <stdint.h>

int main(void)
{
	uint32_t j = 1;
	do {
		printf("Testing %u\n", j);
		uint32_t rec = 0xffffffffU / j;
		//uint32_t rec = (1ull << 32) / j;
		//uint32_t rec = ((1ull << 32) + j - 1) / j;
		uint32_t i = 0;
		do {
			uint32_t res = ((i + 1) * rec * (uint64_t)j) >> 32;
			//uint32_t res = (i * rec * (uint64_t)j) >> 32;
			if (res != i % j) {
				printf("%u %% %u = %u got %u\n", i, j, i % j, res);
				break;
			}
		} while (++i);
	} while (++j);

	return 0;
}
```

```
Testing 1
Testing 2
2147483648 % 2 = 0 got 1
Testing 3
Testing 4
1073741824 % 4 = 0 got 3
Testing 5
Testing 6
1073741824 % 6 = 4 got 3
Testing 7
1073741824 % 7 = 1 got 0
Testing 8
536870912 % 8 = 0 got 7
Testing 9
1073741824 % 9 = 1 got 0
Testing 10
715827882 % 10 = 2 got 1
Testing 11
1073741824 % 11 = 1 got 0
Testing 12
1073741824 % 12 = 4 got 3
Testing 13
477218588 % 13 = 2 got 1
Testing 14
1073741824 % 14 = 8 got 7
Testing 15
Testing 16
268435456 % 16 = 0 got 15
Testing 17
Testing 18
1073741824 % 18 = 10 got 9
Testing 19
715827882 % 19 = 13 got 12
Testing 20
268435456 % 20 = 16 got 15
```

The failures with powers of 2 are not a big deal - it's easy to detect powers of 2 and use bitmask - but others are a problem.

solardiz:
> A way to speed them up would be to precompute (at runtime, but out of loop - perhaps on host) a multiplicative reciprocal

See also code and discussion at https://github.com/P-H-C/phc-winner-argon2/pull/306

solardiz:
@magnumripper Just to be clear, I think you should merge this PR as-is now. Then create issue(s) to keep track of other ideas we discussed. Thanks!

magnumripper:
> @magnumripper Just to be clear, I think you should merge this PR as-is now. Then create issue(s) to keep track of other ideas we discussed. Thanks!

Yeah I'll probably merge tomorrow. I just need to analyze a bunch of tests I made (including with hundreds of millions of hashes). No real problems seen but perhaps one or two bitmap size/level decisions need a little tweaking: The "4-step bitmaps" end up with different requirements and a smaller effective mask now, compared to the 128-bits version, as we only have 64 bits to pick from. It's faster than before anyway but it can be even better.

That, and I'll refactor one variable name in the new file `opencl_hash_check_64.c`. It ended up different from `opencl_hash_check_128.c` due to code copying from LM/DES files (`bitmap_size` vs. `bitmap_size_bits`), and that makes it unnecessary hard to compare them side by side. Better fix that before the new file enters the repo.

magnumripper:
> the hardcoded limit is **512MB bitmap in bytes** which means 4 G bits (hence overflowing int32). But what I wrote above is basically correct - 320294464 would end up oversized and current kernel would bug out.
> (...)
> So the correct kernel limit is 256 MB bitmap, which is 2 Gbits and a mask of 0x7fffffff. Besides, I doubt a bitmap that large will be more effective than just looking up the two integers, but I will test that. After all, such lookup includes modulus operations that might be expensive.

I have since found out that 320M hashes will not hit that 31-bit limit. Exactly 691225600 hashes would, if someone would try to fit them. I figured out very trivial changes to bump that limit and allow all 32 bits and 4G (1 GB) bitmaps (basicly passing `0xffffffffU` to the kernel instead of `0x100000000` which would need `UL` or overflow). I will include those fixes for NT-opencl in this PR, but the fixes to other 128-bit formats (which I have ready) will go in a separate PR.

claudioandre-br:
Just to keep documented, performance regression can happen on some hardware:

From (this is the rolling Windows package):
```
run\john --test=5 --format=nt-opencl
Device 1: gfx902 [AMD Radeon(TM) Vega 8 Graphics]
Benchmarking: NT-opencl [MD4 OpenCL/mask accel]... Build log: .\OCL8388T1.cl:332:18: warning: unknown attribute 'max_constant_size' ignored
                __attribute__((max_constant_size (NUM_INT_KEYS * 4)))
                               ^
1 warning generated.

Build log: .\OCL8388T2.cl:332:18: warning: unknown attribute 'max_constant_size' ignored
                __attribute__((max_constant_size (NUM_INT_KEYS * 4)))
                               ^
1 warning generated.

LWS=64 GWS=512 (8 blocks) x2470 DONE
Raw:    330109K c/s real, 2332M c/s virtual
```

To:
```
C:\Temp\Devel>run\john --test=5 --format=nt-opencl
Device 1: gfx902 [AMD Radeon(TM) Vega 8 Graphics]
Benchmarking: NT-opencl [MD4 OpenCL/mask accel]... LWS=64 GWS=512 (8 blocks) x2470 DONE
Raw:    287571K c/s real, 2857M c/s virtual


C:\Temp\Devel>run\john --test=5 --format=nt-long-opencl
Device 1: gfx902 [AMD Radeon(TM) Vega 8 Graphics]
Benchmarking: NT-long-opencl [MD4 OpenCL/mask accel]... LWS=64 GWS=512 (8 blocks) x2470 DONE
Raw:    279167K c/s real, 2366M c/s virtual
```
****
```
C:\Temp\Devel>run\john --list=opencl-devices
Platform #0 name: AMD Accelerated Parallel Processing, version: OpenCL 2.1 AMD-APP (3075.13)
    Device #0 (1) name:     gfx902
    Board name:             AMD Radeon(TM) Vega 8 Graphics
    Device vendor:          Advanced Micro Devices, Inc.
    Device version:         OpenCL 2.0 AMD-APP (3075.13)
    OpenCL version support: OpenCL C 2.0
    Driver version:         3075.13 (PAL,HSAIL) - AMDGPU-Pro
    Native vector widths:   char 4, short 2, int 1, long 1
    Preferred vector width: char 4, short 2, int 1, long 1
    Global Memory:          5672 MiB
    Global Memory Cache:    16 KiB
    Local Memory:           32 KiB (Local)
    Constant Buffer size:   3081 MiB
    Max memory alloc. size: 3081 MiB
    Max clock (MHz):        1200
    Profiling timer res.:   1 ns
    Max Work Group Size:    256
    Parallel compute cores: 8
    Stream processors:      512  (8 x 64)
    Speed index:            614400
    SIMD width:             16
    Wavefront width:        64
    ADL:                    Overdrive0, device id -1
    PCI device topology:    05:00.0

```

solardiz:
Thank you, @claudioandre-br! I'd say a bigger problem here is all of those speeds are so very low. There's something very wrong going on, both before and after this PR's changes. Per the device info, I'd expect speeds of around 1/4 of Tahiti, so around 2G for the new code, but you were and still are only getting around 300M. Was `GWS=512` autodetected? Can you try larger? Can you show `-v=5` of the autodetection? Optimal might be `GWS=32768`.

claudioandre-br:
```
c:\Temp\Devel>run\john --test=5 --format=nt-opencl --verbosity=6 --gws=32768
initUnicode(UNICODE, RAW/RAW)
RAW -> RAW -> RAW
Device 1: gfx902 [AMD Radeon(TM) Vega 8 Graphics]
Benchmarking: NT-opencl [MD4 OpenCL/mask accel]... Loaded 44 hashes with 1 different salts to test db from test vectors
TestDB LWS=8 GWS=32760 (4095 blocks) PASS,
Test mask: ?a?a?l?u?d?d?s
Expecting "no" false positives
44 hashes: bitmap 4x32768 bits, mask 0x7fff, effectively 38 bits (0x3fffffffff), 16 KiB (local)
Offset tbl 104 B, Hash tbl 360 B, Results 532 B, Dupe bmp 8 B, TOTAL on GPU: 17388 B
Calculating best LWS for GWS=32768
Tuning for password length 7
Note: Profiling timers seem buggy
Testing LWS=64 GWS=32768 ... 1266874354.810 s
Testing LWS=128 GWS=32768 ... 1266874354.779 s
Testing LWS=256 GWS=32768 ... 1266874354.780 s
TestDB LWS=64 GWS=32768 (512 blocks) x2470 DONE
Raw:    1327M c/s real, 110526M c/s virtual
```

For reference (I don't remember the best value I've seen)
```
c:\Temp\Devel>run\john --test=5 --format=sha512crypt-opencl --verbosity=6
initUnicode(UNICODE, RAW/RAW)
RAW -> RAW -> RAW
Device 1: gfx902 [AMD Radeon(TM) Vega 8 Graphics]
Benchmarking: sha512crypt-opencl, crypt(3) $6$ (rounds=5000) [SHA512 OpenCL]... Loaded 6 hashes with 4 different salts to test db from test vectors
TestDB LWS=7 GWS=49 (7 blocks) PASS,
Test mask: ?a?a?l?u?d?d?s
Calculating best GWS for LWS=64; max. 100 ms single kernel invocation.
Tuning for iteration count of 5000 and password length 7
xfer: 6.720 us, crypt: 59x4.212 ms, xfer back: 8.960 us, prep: 3.155 ms, pp: 38.240 us, final: 1.969 ms, var: 4.932 ms/4.221 ms
gws:      1024     3787 c/s    18935000 rounds/s  270.378 ms per crypt_all()!
xfer: 9.440 us, crypt: 59x5.167 ms, xfer back: 13.600 us, prep: 3.258 ms, pp: 51.680 us, final: 1.870 ms, var: 5.455 ms/6.107 ms
gws:      2048     6074 c/s    30370000 rounds/s  337.165 ms per crypt_all()+
xfer: 17.120 us, crypt: 59x8.762 ms, xfer back: 48.160 us, prep: 6.273 ms, pp: 103.040 us, final: 4.029 ms, var: 8.865 ms/9.095 ms
gws:      4096     7575 c/s    37875000 rounds/s  540.706 ms per crypt_all()+
xfer: 32.480 us, crypt: 59x18.085 ms, xfer back: 316.480 us, prep: 13.124 ms, pp: 356 us, final: 7.784 ms, var: 18.071 ms/18.555 ms
gws:      8192     7398 c/s    36990000 rounds/s     1.107 s per crypt_all()
xfer: 54.880 us, crypt: 59x36.137 ms, xfer back: 473.280 us, prep: 26.633 ms, pp: 497.599 us, final: 14.964 ms, var: 36.096 ms/35.526 ms
gws:     16384     7513 c/s    37565000 rounds/s     2.180 s per crypt_all()
xfer: 101.440 us, crypt: 59x70.359 ms, xfer back: 1.223 ms, prep: 53.713 ms, pp: 1.047 ms, final: 22.346 ms, var: 68.174 ms/66.627 ms
gws:     32768     7897 c/s    39485000 rounds/s     4.149 s per crypt_all()+
xfer: 680.840 us, crypt: 59x117.271 ms (exceeds 100 ms)
xfer: 54.400 us, crypt: 59x33.795 ms, xfer back: 618.400 us, prep: 23.863 ms, pp: 486.720 us, final: 9.588 ms, var: 30.231 ms/29.353 ms
gws:     16384     8680 c/s    43400000 rounds/s     1.887 s per crypt_all()!!
xfer: 34.560 us, crypt: 59x15.076 ms, xfer back: 413.120 us, prep: 8.576 ms, pp: 225.120 us, final: 6.660 ms, var: 16.675 ms/16.653 ms
gws:      8192     8390 c/s    41950000 rounds/s  976.344 ms per crypt_all()-
Calculating best LWS for GWS=16384
Testing LWS=64 GWS=16384 ... 223.928 ms+
Testing LWS=128 GWS=16384 ... 216.670 ms+
Testing LWS=256 GWS=16384 ... 230.644 ms
Calculating best GWS for LWS=128; max. 200 ms single kernel invocation.
xfer: 4.960 us, crypt: 59x4.289 ms, xfer back: 5.600 us, prep: 3.225 ms, pp: 37.920 us, final: 1.991 ms, var: 4.214 ms/4.470 ms
gws:       512     1949 c/s     9745000 rounds/s  262.675 ms per crypt_all()!
xfer: 17.760 us, crypt: 59x4.048 ms, xfer back: 8.160 us, prep: 3.522 ms, pp: 37.600 us, final: 1.899 ms, var: 4.359 ms/4.050 ms
gws:      1024     4052 c/s    20260000 rounds/s  252.665 ms per crypt_all()!
xfer: 24.320 us, crypt: 59x6.498 ms, xfer back: 13.920 us, prep: 3.339 ms, pp: 49.120 us, final: 1.859 ms, var: 7.025 ms/7.572 ms
gws:      2048     4831 c/s    24155000 rounds/s  423.854 ms per crypt_all()+
xfer: 17.120 us, crypt: 59x8.847 ms, xfer back: 44 us, prep: 6.418 ms, pp: 92.800 us, final: 3.642 ms, var: 8.864 ms/8.562 ms
gws:      4096     7705 c/s    38525000 rounds/s  531.547 ms per crypt_all()+
xfer: 32.640 us, crypt: 59x15.169 ms, xfer back: 372.960 us, prep: 12.105 ms, pp: 299.840 us, final: 5.253 ms, var: 16.936 ms/16.573 ms
gws:      8192     8325 c/s    41625000 rounds/s  983.937 ms per crypt_all()+
xfer: 58.560 us, crypt: 59x34.249 ms, xfer back: 592.800 us, prep: 25.343 ms, pp: 490.720 us, final: 10.405 ms, var: 30.064 ms/30.023 ms
gws:     16384     8583 c/s    42915000 rounds/s     1.908 s per crypt_all()+
xfer: 100.160 us, crypt: 59x62.373 ms, xfer back: 1.018 ms, prep: 49.318 ms, pp: 1.012 ms, final: 23.231 ms, var: 61.919 ms/61.852 ms
gws:     32768     8696 c/s    43480000 rounds/s     3.768 s per crypt_all()+
xfer: 681.760 us, crypt: 59x122.810 ms, xfer back: 1.985 ms, prep: 83.582 ms, pp: 2.095 ms, final: 52.364 ms, var: 127.398 ms/125.858 ms
gws:     65536     8620 c/s    43100000 rounds/s     7.602 s per crypt_all()
xfer: 1.916 ms, crypt: 59x268.318 ms (exceeds 200 ms)
xfer: 54.560 us, crypt: 59x32.399 ms, xfer back: 594.080 us, prep: 23.453 ms, pp: 485.760 us, final: 10.876 ms, var: 31.157 ms/30.710 ms
gws:     16384     8596 c/s    42980000 rounds/s     1.905 s per crypt_all()-
TestDB LWS=128 GWS=32768 (256 blocks) DONE
Raw:    7597 c/s real, 484256 c/s virtual
```

ghost:
I removed my previous comments. Correct values are:

From:
```
C:\Temp\rolling> run\john --test=5 --format=nt-opencl --gws=32768
Device 1: gfx902 [AMD Radeon(TM) Vega 8 Graphics]
Benchmarking: NT-opencl [MD4 OpenCL/mask accel]... LWS=64 GWS=32768 (512 blocks) x2470 DONE
Raw:    1433M c/s real, 34609M c/s virtual
```

To:
```
C:\Temp\debug> run\john --test=5 --format=nt-opencl --verb=6
initUnicode(UNICODE, RAW/RAW)
RAW -> RAW -> RAW
Device 1: gfx902 [AMD Radeon(TM) Vega 8 Graphics]
Benchmarking: NT-opencl [MD4 OpenCL/mask accel]... Loaded 44 hashes with 1 different salts to test db from test vectors
TestDB LWS=8 GWS=48 (6 blocks) PASS,
Test mask: ?a?a?l?u?d?d?s
Expecting "no" false positives
44 hashes: bitmap 4x32768 bits, mask 0x7fff, effectively 38 bits (0x3fffffffff), 16 KiB (local)
Offset tbl 112 B, Hash tbl 360 B, Results 532 B, Dupe bmp 8 B, TOTAL on GPU: 17396 B
Internal mask, multiplier: 2470 (target: 2048)
Calculating best GWS for LWS=64; max. 200 ms single kernel invocation.
Tuning for password length 7
key xfer: 3.680 us, idx xfer: 1.919 us, crypt: 4.018 ms, res xfer: 3.360 us
gws:      1024  627949K c/s   627949471 rounds/s    4.027 ms per crypt_all()!
key xfer: 4.640 us, idx xfer: 2.880 us, crypt: 4.018 ms, res xfer: 3.520 us
gws:      2048    1255M c/s  1255400254 rounds/s    4.029 ms per crypt_all()+
key xfer: 5.440 us, idx xfer: 3.200 us, crypt: 12.256 ms, res xfer: 3.360 us
gws:      4096  824611K c/s   824611050 rounds/s   12.268 ms per crypt_all()
key xfer: 7.200 us, idx xfer: 4.800 us, crypt: 22.505 ms, res xfer: 3.040 us
gws:      8192  898468K c/s   898468970 rounds/s   22.520 ms per crypt_all()
key xfer: 12 us, idx xfer: 6.560 us, crypt: 38.292 ms, res xfer: 3.360 us
gws:     16384    1056M c/s  1056234392 rounds/s   38.313 ms per crypt_all()
key xfer: 20.480 us, idx xfer: 10.720 us, crypt: 65.417 ms, res xfer: 9.440 us
gws:     32768    1236M c/s  1236470119 rounds/s   65.458 ms per crypt_all()
key xfer: 35.040 us, idx xfer: 19.200 us, crypt: 111.924 ms, res xfer: 3.200 us
gws:     65536    1445M c/s  1445536207 rounds/s  111.981 ms per crypt_all()+
key xfer: 67.840 us, idx xfer: 33.920 us, crypt: 252.095 ms (exceeds 200 ms)
key xfer: 21.120 us, idx xfer: 10.880 us, crypt: 61.302 ms, res xfer: 3.040 us
gws:     32768    1319M c/s  1319529112 rounds/s   61.337 ms per crypt_all()-
Calculating best LWS for GWS=65536
Note: Profiling timers seem buggy
Testing LWS=64 GWS=65536 ... 1266874552.150 s
Testing LWS=128 GWS=65536 ... 1266874552.118 s
Testing LWS=256 GWS=65536 ... 1266874552.101 s
Calculating best GWS for LWS=64; max. 200 ms single kernel invocation.
key xfer: 3.520 us, idx xfer: 1.919 us, crypt: 3.967 ms, res xfer: 3.200 us
gws:       512  318055K c/s   318055611 rounds/s    3.976 ms per crypt_all()!
key xfer: 4 us, idx xfer: 1.760 us, crypt: 4.019 ms, res xfer: 3.040 us
gws:      1024  627949K c/s   627949471 rounds/s    4.027 ms per crypt_all()+
key xfer: 4.320 us, idx xfer: 2.720 us, crypt: 4.019 ms, res xfer: 3.200 us
gws:      2048    1255M c/s  1255450105 rounds/s    4.029 ms per crypt_all()+
key xfer: 5.760 us, idx xfer: 3.360 us, crypt: 10.686 ms, res xfer: 4.160 us
gws:      4096  945525K c/s   945525233 rounds/s   10.700 ms per crypt_all()
key xfer: 22.560 us, idx xfer: 14.080 us, crypt: 23.233 ms, res xfer: 3.360 us
gws:      8192  869395K c/s   869395443 rounds/s   23.273 ms per crypt_all()
key xfer: 12.480 us, idx xfer: 7.200 us, crypt: 36.503 ms, res xfer: 4.320 us
gws:     16384    1107M c/s  1107904719 rounds/s   36.527 ms per crypt_all()
key xfer: 60.480 us, idx xfer: 33.120 us, crypt: 70.694 ms, res xfer: 3.040 us
gws:     32768    1143M c/s  1143324676 rounds/s   70.790 ms per crypt_all()
key xfer: 35.200 us, idx xfer: 18.560 us, crypt: 123.492 ms, res xfer: 3.200 us
gws:     65536    1310M c/s  1310197194 rounds/s  123.549 ms per crypt_all()+
key xfer: 208.160 us, idx xfer: 106.560 us, crypt: 252.464 ms (exceeds 200 ms)
key xfer: 20.160 us, idx xfer: 10.720 us, crypt: 61.423 ms, res xfer: 3.200 us
gws:     32768    1316M c/s  1316952651 rounds/s   61.457 ms per crypt_all()!!
key xfer: 35.840 us, idx xfer: 20.800 us, crypt: 34.542 ms, res xfer: 3.360 us
gws:     16384    1169M c/s  1169511901 rounds/s   34.602 ms per crypt_all()-
TestDB LWS=64 GWS=32768 (512 blocks) x2470 DONE
Raw:    1479M c/s real, 50635M c/s virtual
```

solardiz:
> precompute (at runtime, but out of loop - perhaps on host) a multiplicative reciprocal

I realized that since we pass the used-to-be-runtime-variable `HASH_TABLE_SIZE` from host to OpenCL kernel as its build-time constant, the OpenCL compiler has opportunity to perform the same optimization on its own - and maybe it already does!

PR #5250, opened 2023-03-12
Alternative solution nt kernel for long plaintexts

magnumripper:
See #5246 and #5245

This version has only two branches (if configured for length 59) and doesn't lose any optimizations for single-block crypts.

Despite this, it's too slow on AMD w/ old driver.

On nvidia, it's about as fast as the version in #5246.

### DIFF
--- a/run/opencl/nt_kernel.cl
+++ b/run/opencl/nt_kernel.cl
@@ -1,4 +1,5 @@
-/* NTLM kernel (OpenCL 1.2 conformant)
+/*
+ * NTLM kernel (OpenCL 1.2 conformant)
  *
  * Written by Alain Espinosa <alainesp at gmail.com> in 2010 and modified by
  * Samuele Giovanni Tonon in 2011. No copyright is claimed, and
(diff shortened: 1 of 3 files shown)


magnumripper:
This version has two branches but no `else` clauses. That makes `unlikely()` less likely to make a difference, right?

solardiz:
> it's too slow on AMD w/ old driver.

Are you adding a/b/c/d variables on top of `hash` array (which is normally also in registers)? If so, this increases register pressure, and perhaps you can avoid that.

> This version has two branches but no `else` clauses. That makes `unlikely()` less likely to make a difference, right?

I'd expect it to make a difference anyway. The whole block can be moved out - yes, with an extra branch at the end.

magnumripper:
> This version has only two branches (if configured for length 59)

Weird fact: At lengths over 59 we add code for another full MD4 in a `while` loop. But the longer support I use, the faster nvidia gets. It's at best at 125 and significantly slower at 59.

magnumripper:
> Are you adding a/b/c/d variables on top of `hash` array (which is normally also in registers)? If so, this increases register pressure, and perhaps you can avoid that.

Well, I need them! For single block, the very last thing that happens is adding `INIT_A..INIT_D` to the values (unless skipping last steps). But for 2nd block on we need to add the original content that was in `hash[0]..hash[3]` so they're stored in `a..d`.

PR #5251, opened 2023-03-13
Singlebranch

magnumripper:
Third try.

Now there's just a single `if unlikely()` in the end of the function, ~but AMD *still* acts up with poor performance~ 😢


magnumripper:
> Now there's just a single `if unlikely()` in the end of the function, but AMD _still_ acts up with poor performance 😢

That was a commit mistake. Performance is up a good deal now, 19144M seen. Lowering from 125 to 59 doesn't make any difference.

PR #5269, opened 2023-04-02
Update packaging instructions

ghost:
I don't run this script myself, but as it was copied from the flatpak build process it should be ok.


magnumripper (approved):
Just skimmed through the changes, looks reasonable to me

PR #5270, opened 2023-04-03
Enable fallback if binary name given

ghost:
I'm thinking about proposing a patch like this: if you specify a binary name, fallback will be enabled. It will make the building process cleaner and easier to read.

I run all the basic tests, but it deserves more testing. Anyway, let's focus and discuss the main idea now.

From:
```
./configure --something   CPPFLAGS="-DOMP_FALLBACK -DOMP_FALLBACK_BINARY=\"\\\"john-avx2\\\"\" -DCPU_FALLBACK -DCPU_FALLBACK_BINARY=\"\\\"john-avx-omp\\\"\""
```

To:
```
./configure --something   CPPFLAGS="-DOMP_FALLBACK_BINARY=\"\\\"john-avx2\\\"\" -DCPU_FALLBACK_BINARY=\"\\\"john-avx-omp\\\"\""
```

### DIFF
--- a/src/john.c
+++ b/src/john.c
@@ -354,7 +354,7 @@ static void john_omp_init(void)
 		john_omp_threads_orig = john_omp_threads_new;
 }
 
-#if OMP_FALLBACK
+#if OMP_FALLBACK || defined(OMP_FALLBACK_BINARY)
 #if defined(__DJGPP__)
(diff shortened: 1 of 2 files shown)


magnumripper (approved):
This looks perfectly reasonable to me

PR #5273, opened 2023-04-11
Warn if a pre RAR 3 file is detected

### LINKED ISSUE #5271: RAR3-p hash with *35 ending won't find password in wordlist
I have a relatively old RAR file which I know the password for, and placed it in a wordlist, which was passed to john, but john couldn't recover the password for the RAR file. 

My hash looks like this:
`$RAR3$*1*0000000000000000*12b2c880*22304*61431*1*f9...999*35`

Used commands:
`.\rar2john.exe .\t.rar > thash`
`.\john.exe --wordlist=pass.lst thash`

I have tried to use john both on Windows and MacOS.
I think I did everything correct, so I don't know why john can't find the correct password in the list.

build-info:

Version: 1.9.0-jumbo-1
CPU tests: AVX2
CPU fallback binary: john-xop
$JOHN is /run/
Format interface version: 14
Rec file version: REC4
Charset file version: CHR3
CHARSET_MIN: 1 (0x01)
CHARSET_MAX: 255 (0xff)
CHARSET_LENGTH: 24
SALT_HASH_SIZE: 1048576
SINGLE_IDX_MAX: 2147483648
SINGLE_BUF_MAX: 4294967295
Max. Markov mode level: 400
gcc version: 7.4.0
OpenCL headers version: 2.2
Crypto library: OpenSSL
OpenSSL 1.1.1b  26 Feb 2019
GMP library version: 6.1.2
File locking: fcntl()
fseek(): fseek
ftell(): ftell
fopen(): fopen
memmem(): System's

ghost:
Closes #5271 

Closes the bug, we can keep the issue open as an "enhancement" to request to implement the "new format".

### DIFF
--- a/src/rar2john.c
+++ b/src/rar2john.c
@@ -509,6 +509,13 @@ next_file_header:
 		if (gecos_len + strlen((char*)file_name) < LINE_BUFFER_SIZE)
 			gecos_len += snprintf(&gecos[gecos_len], LINE_BUFFER_SIZE - gecos_len - 1, "%s ", (char*)file_name);
 
+		/* try to detect if the file is >= RAR 3
+		   - best try: LHD_SALT is not set and UnpVer <= 20
+		*/
+		if (file_hdr_block[24] <= 20 && !(file_hdr_head_flags & 0x400)) {
+			fprintf(stderr, "Probably this file was NOT created using RAR 3 or higher, bailing out.\n");
+			goto BailOut;
+		}
 		/* salt processing */
 		if (file_hdr_head_flags & 0x400) {
 			ext_time_size -= 8;


magnumripper (commented):
I came up with a slightly different patch before I saw this PR:
```diff
diff --git a/src/rar2john.c b/src/rar2john.c
index c7a4ae78b..290ba18d0 100644
--- a/src/rar2john.c
+++ b/src/rar2john.c
@@ -562,6 +562,10 @@ next_file_header:
 			fprintf(stderr, "! not encrypted, skipping\n");
 			jtr_fseek64(fp, file_hdr_pack_size, SEEK_CUR);
 			goto next_file_header;
+		} else if (file_hdr_block[24] < 29) {
+			fprintf(stderr, "! %s: Too old RAR file version (v%0.1f encryption), not supported.\n",
+			        archive_name, (float)file_hdr_block[24] / 10.);
+			goto err;
 		}
 
 		method = file_hdr_block[25];
```
The reason I chose v2.9 is our unpack code is called `rar_unpack29()`. I tested the above patch against our samples and it seems fine - but I'm not sure mine is better or worse.

ghost:
I was confused whether to go to `err` or `BailOut`. anyway, Please feel free to choose

magnumripper:
I'll do some more testing real quick.

magnumripper:
OK so yours and mine version handle all samples the same. I'm not quite sure but I think we should go with my version - I'm pretty sure the version (archive format version, not RAR version) where it changed was v2.9.

I'll make a PR with that, then let's sleep on it.

magnumripper:
> I was confused whether to go to `err` or `BailOut`. anyway, Please feel free to choose

I amended my patch to do neither - it now skips to next file in the archive. In the (remote) case there are a mix of format versions in the archive, we might be able to handle some later file.

PR #5275, opened 2023-04-11
RAR formats: Reject old archives

### LINKED ISSUE #5271: RAR3-p hash with *35 ending won't find password in wordlist

magnumripper:
These have "unknown" KDF and encryption, and no salt. We could work it out by examining public unrar source code, but such archives are so rare I'm not sure we'll ever bother.

Closes #5271 - see also #5273

### DIFF
--- a/src/rar2john.c
+++ b/src/rar2john.c
@@ -562,6 +562,11 @@ next_file_header:
 			fprintf(stderr, "! not encrypted, skipping\n");
 			jtr_fseek64(fp, file_hdr_pack_size, SEEK_CUR);
 			goto next_file_header;
+		} else if (file_hdr_block[24] < 29) {
+			fprintf(stderr, "! %s: Too old RAR file version (v%u.%u encryption), not supported.\n",
+			        archive_name, file_hdr_block[24] / 10, file_hdr_block[24] % 10);
+			jtr_fseek64(fp, file_hdr_pack_size, SEEK_CUR);
+			goto next_file_header;
 		}
 
 		method = file_hdr_block[25];


solardiz:
Should we ideally also do such rejection in the format? For "hashes" generated by `rar2john` from prior to this PR.

magnumripper:
> Should we ideally also do such rejection in the format? For "hashes" generated by `rar2john` from prior to this PR.

I added that at first but rejected it. The only thing the formats can detect is an all-zero salt and I'm not sure that could not happen (and be valid) in some other remote cases. The formats could/should emit a warning about it though.

PR #5277, opened 2023-04-12
OpenCL formats fixes

### LINKED ISSUE #5264: A no-SIMD build fails
It is not a real build (since it is OpenCL without SIMD), but if a new arch supports OpenCL that could be a problem.

```
opencl_nt_fmt_plug.c: In function 'cmp_exact':
opencl_nt_fmt_plug.c:686:9: error: implicit declaration of function 'md4_reverse' [-Werror=implicit-function-declaration]
  686 |         md4_reverse(hash);
      |         ^~~~~~~~~~~
cc1: all warnings being treated as errors
make[1]: *** [Makefile:1572: opencl_nt_fmt_plug.o] Error 1

Exited with code exit status 2

```

magnumripper:
NT-opencl: Minor bugfix for logging of bitmap parameters.
Move reversing/unreversing hash functions from SIMD-code to common code. Closes #5264.
Support largest bitmaps (0xffffffff) without failing kernel build (formats using perfect hash tables).
Fix a bug where some workgroup sizes would fail to properly copy bitmap to local memory (formats using perfect hash tables).
While at it, for DES/LM OpenCL formats fix a similar copy to local to avoid modulo arithmetic.

### DIFF
--- a/run/opencl/DES_bs_kernel.cl
+++ b/run/opencl/DES_bs_kernel.cl
@@ -1,5 +1,6 @@
 /*
- * This software is Copyright (c) 2012-2015 Sayantan Datta <std2048 at gmail dot com>
+ * This software is Copyright (c) 2012-2015 Sayantan Datta <std2048 at gmail dot com>,
+ * Copyright (c) 2023 magnum,
  * and it is hereby released to the general public under the following terms:
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted.
@@ -150,15 +151,15 @@ __kernel void DES_bs_25_b(constant uint *key_map
 	__local DES_bs_vector s_des_bs_key[56 * WORK_GROUP_SIZE];
 	int s_key_offset = lid * 56;
 	for (i = 0; i < 56; i++)
-		s_des_bs_key[lid * 56 + i] = des_bs_key[section + i * gws];
+		s_des_bs_key[s_key_offset + i] = des_bs_key[section + i * gws];
 #endif
 
 #if USE_LOCAL_MEM
 	__local ushort s_key_map[768];
 	int lws = get_local_size(0);
 
-	for (i = 0; i < 768; i += lws)
-		s_key_map[(lid + i) % 768] = key_map[(lid + i) % 768];
+	for (i = lid; i < 768; i += lws)
+		s_key_map[i] = key_map[i];
 #endif
 
 #if USE_LOCAL_MEM || WORK_GROUP_SIZE > 0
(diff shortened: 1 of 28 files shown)


solardiz:
> While at it, for DES/LM OpenCL formats fix a similar copy to local to avoid modulo arithmetic.

I don't fully understand what the code was supposed to do, but these changes are not no-ops: they no longer write into elements before `lid`. Can you explain why they don't need to do those writes, yet do(?) need to write to elements after `lid`? This looks like a bug to me - it's either outright wrong or non-optimal - but maybe there's some subtle reason why it's right and optimal?

magnumripper:
The thread(s) with lower lids will write them. This is shared memory! The code before this change was VERY weird.

solardiz:
> The thread(s) with lower lids will write them. This is shared memory! The code before this change was VERY weird.

To me, it looks weird after the change as well. If we rely on lower-numbered threads writing the lower-numbered elements, then why don't we do the same for higher-numbered? In other words, why have those loops at all (instead of single assignments by each thread)?

magnumripper:
```c
for (i = lid; i < 768; i += lws)
	s_key_map[i] = key_map[i];
```
If lws is 4,
lid 0 will write 0, 4, 8 and so on
lid 1 will write 1, 5, 9 and so on
lid 2 will, well you get the idea.

After the copy we have a barrier, before anyone tries to do anything with the complete array.

Now, the _previous_ code did some really weird shit - where several threads could copy some of the elements - I'm not even sure it was bullet proof safe.

solardiz:
OK, makes sense. Thanks.

magnumripper:
Now that we're on that page, I'm starting to think we should change this nearby code too (In this PR I merely replaced the repetition of `56 * lid` that the compiler surely did for us anyway):
```c
		__local DES_bs_vector s_des_bs_key[56 * WORK_GROUP_SIZE];
		int lid = get_local_id(0);
		int s_key_offset = 56 * lid;

		for (i = 0; i < 56; i++)
			s_des_bs_key[s_key_offset + i] = des_bs_key[section + i * gws];
		barrier(CLK_LOCAL_MEM_FENCE);
```
As written, every single thread will read and try to write the exact same memory simultaneously, causing hickups in performance at best and possibly worse problems.

I think it should be like this (same boiler plate code as we discussed)
```c
		for (i = lid; i < 56; i += lws)
			s_des_bs_key[s_key_offset + i] = des_bs_key[section + i * gws];
		barrier(CLK_LOCAL_MEM_FENCE);
```

magnumripper:
No wait, there's obviously the `56 * lid` already (that I just hid to myself)... Never mind, the code is OK (I think). LOL.

Issue #5267, opened 2023-04-02
Make BitLocker OpenCL use shared SHA-256 code


solardiz:
Right now, `opencl_bitlocker.h` and `bitlocker_kernel.cl` have their own implementation of SHA-256. We could both reduce source code size/duplication and likely speed things up by switching to our shared and maintained SHA-256 code.

In particular, I notice that BitLocker's current code does not use `bitselect` - it can only use LUT3 and classic (two-operand) Boolean operations. This likely makes it less optimal on AMD GPUs. It also does not use `rotate`. So a temporary fix (before the refactoring to use shared code) may be to add such usage (in the same cases where we do it in our shared SHA-256 code).

magnumripper:
We also have some of `#if 0 && SHA256_LUT3` in the shared code - where other alternatives are faster. We currently use LUT3 for Ch and Maj but not for the Sigma/sigma functions. That's subject to change though, it's not tested with recent drivers.

magnumripper:
nvidia 2080ti, speeds at -lws=64 -gws=8704 (not optimal size, the format is so incredibly slow I just picked a number for rough evaluation):

Original kernel binary size: 885314 bytes and 2371 c/s.
Using rotate(): 884633 bytes and 2378 c/s.
Also using our shared lut3() function instead of the existing, slightly different, one: 882102 bytes and 2397 c/s.

The speeds might be just the usual flux, but the sizes are interesting.

I tried using plain code instead of lut3 for all three functions (one at a time) but only one of them made for smaller code size (the xor) while not really gaining any speed.

More interesting would be testing it on AMD of course. I'm not aware of bitselect alternatives for the "xor" or the "xorand" though. Current code is in #5278.

The difference of lut3 implementation is interesting. Here's our shared one:
```c
inline uint lut3(uint a, uint b, uint c, uint imm)
{
	uint r;
	asm("lop3.b32 %0, %1, %2, %3, %4;"
	    : "=r" (r)
	    : "r" (a), "r" (b), "r" (c), "i" (imm));
	return r;
}
```

Here's the old one in bitlocker:
```c
inline unsigned int LOP3LUT_XOR(unsigned int a, unsigned int b, unsigned int c)
{
      unsigned int d;

      asm("lop3.b32 %0, %1, %2, %3, 0x96;": "=r"(d):"r"(a), "r"(b), "r"(c));
      return d;
}
```
Is the difference that the latter somehow isn't using "immediate" for the 0x96??

[...]

PR #5278, opened 2023-04-12
Bitlocker-opencl format tweaks

magnumripper:
Use the rotate() OpenCL instruction instead of 11 silly macros.
Use bitselect() for Maj() if available.
Use our shared lut3() inline asm function for nvidia instead of reimplementing it (they differed and this made for a 2 KB smaller binary).

See #5267


solardiz (approved):
I might not spot a typo in a rotate count, but I assume this has been tested and these are changes in the right direction.

magnumripper:
> I might not spot a typo in a rotate count

Luckily they were all replaced by sed(1) and not manually, so there should be no typos. And yes I tested it.

PR #5281, opened 2023-04-14
Fix fallback for a build using Makefile.legacy

ghost:
Commit 0bbec45bbe26da288ad9b8c1c8b2c65700500a65 doesn't handle Makefile.legacy usage properly.

### DIFF
--- a/src/john.c
+++ b/src/john.c
@@ -354,7 +354,7 @@ static void john_omp_init(void)
 		john_omp_threads_orig = john_omp_threads_new;
 }
 
-#if OMP_FALLBACK || defined(OMP_FALLBACK_BINARY)
+#if OMP_FALLBACK || (defined(OMP_FALLBACK_BINARY) && defined(AC_BUILT))
 #if defined(__DJGPP__)
 #error OMP_FALLBACK is incompatible with the current DOS code
 #endif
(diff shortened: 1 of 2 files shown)


ghost:
Lets hear from CI. Anyway, I cannot reproduce any bug.

solardiz:
No, let's not apply these changes. I can reproduce the bug, and I have a different fix.

solardiz:
> Commit [0bbec45](https://github.com/openwall/john/commit/0bbec45bbe26da288ad9b8c1c8b2c65700500a65) doesn't handle Makefile.legacy usage properly.

The bug isn't limited to usage of `Makefile.legacy` - it would also affect `configure` if run on a 32-bit system:

```
$ fgrep ARCH_LINK= configure | tail -2
  ARCH_LINK=x86-mmx.h
  ARCH_LINK=x86-sse.h
```

ghost:
> > Commit [0bbec45](https://github.com/openwall/john/commit/0bbec45bbe26da288ad9b8c1c8b2c65700500a65) doesn't handle Makefile.legacy usage properly.
> 
> The bug isn't limited to usage of `Makefile.legacy` - it would also affect `configure` if run on a 32-bit system:
> 
> ```
> $ fgrep ARCH_LINK= configure | tail -2
>   ARCH_LINK=x86-mmx.h
>   ARCH_LINK=x86-sse.h
> ```

It doesn't affect CI Wine build. But I was able to reproduce it using LXD, 32 bits, and ubuntu 18.

Suggestion: when creating the next `john cloud edition`, let's tag the same commit in this repository as `jumbo 2` and deprecate all these things:
- Legacy Makefile;
- DOS;
- 32 bit X86;
- Old Mac;
- ...
****
**[edited]**

A sort of soft release that packagers can follow to avoid a the big gap between full releases. Deprecation and removal is also a good reason to increase `john`'s release number.

solardiz:
> Suggestion: when creating the next `john cloud edition`, let's tag the same commit in this repository as `jumbo 2` and deprecate all these things

I don't mind tagging and using the `-jumbo-N` numbering for "unreleased" versions, but as to deprecation I don't want to remove anything before 2.0 and without a real reason to remove. Those things you listed take very little time to maintain now. This can change if we do some related major refactoring - and that would also be a reason and the time to drop the legacy functionality - perhaps in 2.1+.

Issue #5240, opened 2023-03-05
Fail to compile with Clang because of -m64 linker flag


lucic71:
I'm trying to compile john with Clang 15.0.7 and I get the following error:
```
/usr/bin/ld.lld mkvcalcproba.o -g -m64  -L/usr/local/lib -L/usr/lib64 -L/lib64   -lm  -o ../run/mkvcalcproba
ld.lld: error: unknown emulation: 64
```
I ran configure with the following env variables:
```
LD=/usr/bin/ld.lld CC=/usr/bin/clang CXX=/usr/bin/clang++ ./configure
```
The configure output shows the following line:
```
checking if /usr/bin/clang  supports -m64 w/ linking... yes
```
I greped in src/configure and found this:
https://github.com/openwall/john/blob/2f4b2dfee51e1477c5147d5b02ee71dc04b3ce0d/src/configure#L7034

If I understand this right then this line runs the linker to see if -m64 flag is supported. However the linker is called through clang in this case. If you call ld.lld as a standalone program with the -m64 it will complain that 64 is an unknown emulation. The same case happens for GNU gold if called as standalone.

The linkers (both ld.ldd and GNU gold) indeed support some emulation flags, but they take different values (elf_x86_64, etc).

So why set the LDFLAGS to -m64 in the following line?
https://github.com/openwall/john/blob/2f4b2dfee51e1477c5147d5b02ee71dc04b3ce0d/src/configure#L7078

Another thing that confused me is this:
https://github.com/openwall/john/blob/2f4b2dfee51e1477c5147d5b02ee71dc04b3ce0d/src/configure#L7025

Why set 64 bit emulation flags for 64 bit platforms?

Thanks

magnumripper:
The autoconf stuff initially mostly tried to mimic the original Makefile (now `Makefile.legacy`).

Not sure what you mean with emulation flags? As for adding it to LDFLAGS, I think what we do is correct as long as `LD = $(CC)` which is what we use. Then -m64 generates 64-bit code, not emulation. 

We're open to improvements to the config stuff though.

lucic71:
Sorry for the confusion. I was looking at the -m flag for the linker which is used for the emulation layer.

As for LDFLAGS, even if you call LD through CC you don't need -m64 as it does not affect in any way the linking step. It only affects the compilation step.

magnumripper:
[...]

PR #5282, opened 2023-04-14
Fix "Enable fallback if binary name given" bugs

### LINKED ISSUE #5280: "Enable fallback if binary name given" bugs
While the idea behind #5270 is reasonable, the implementation we merged here needs more work. As it is, I think it introduced bugs.

In `x86-sse.h`, we have:

```c
#ifndef CPU_FALLBACK_BINARY_DEFAULT
#define CPU_FALLBACK_BINARY_DEFAULT     "john-non-sse"
#endif

#ifndef CPU_FALLBACK
#define CPU_FALLBACK                    0
#endif
#if !defined(CPU_FALLBACK_BINARY) && defined(CPU_FALLBACK_BINARY_DEFAULT)
#define CPU_FALLBACK_BINARY             CPU_FALLBACK_BINARY_DEFAULT
#endif
```

In `x86-mmx.h`, we have:

```c
#ifndef CPU_FALLBACK
#define CPU_FALLBACK                    0
#endif
#ifndef CPU_FALLBACK_BINARY
#define CPU_FALLBACK_BINARY             "john-non-mmx"
#endif
```

So building with these now, we'll end up with fallback enabled where that wasn't intended. We probably need to revise these.

Also, while the old behavior is documented in a comment in `params.h` and in some files under `doc/`, the new behavior is undocumented. Maybe that's fine as documenting both the old and the new behavior would be cryptic, and the old behavior is also still relevant when default filenames are used and not overridden.

solardiz:
Fixes #5280


src/x86-mmx.h
```diff
@@ -46,7 +46,7 @@
 #ifndef CPU_FALLBACK
 #define CPU_FALLBACK			0
 #endif
-#ifndef CPU_FALLBACK_BINARY
+#if CPU_FALLBACK && !defined(CPU_FALLBACK_BINARY)
```
ghost, line 49: LGTM.

src/Makefile.legacy
```diff
@@ -573,7 +573,7 @@ linux-x86-64-32-any:
 		LDFLAGS="$(LDFLAGS) -m32 -lcrypt -ldl" \
 		AESNI_ARCH=86 YASM_FORMAT="elf32"
 	@echo "Failing after this point just means some helper tools did not build:"
-	$(MAKE_ORIG) $(PROJ_PCAP) CFLAGS="$(CFLAGS) -m32"
+	$(MAKE_ORIG) $(PROJ_PCAP) CFLAGS="$(CFLAGS) -m32" LDFLAGS="$(LDFLAGS) -m32"
```
ghost, line 576: Isn't this a hint that #5240 '-m64' is needed, at least in the meantime?

ghost (approved):
I think I cannot test this *-32 bits builds on recent Ubuntu. Anyway, looks like good stuff.

src/Makefile.legacy
solardiz: No, needing to use `-m32` to build/link 32-bit code on a typical 64-bit system is a hint that such system actually _doesn't_ need explicit `-m64` to build/link 64-bit code.

PR #5285, opened 2023-04-15
Fix for darwin not using SIMD flag

### LINKED ISSUE #5284: Mac OS build on Cloud SIMD detection
It looks like autoconf gets confused. I built using: "--disable-native-tests --enable-simd=avx"

```
Target CPU ......................................... x86_64 SSE4.1, 64-bit LE
[...]
Build: darwin21.6.0 64-bit x86_64 SSE4.1 AC OPENCL
SIMD: AVX, interleaving: MD4:4 MD5:5 SHA1:2 SHA256:1 SHA512:1
[...]
Testing: descrypt, traditional crypt(3) [DES 128/128 AVX]... PASS
```

ghost:
Closes #5284

My version of `autoconf` will add too much changes to `configure`.

### DIFF
--- a/src/m4/jtr_x86_logic.m4
+++ b/src/m4/jtr_x86_logic.m4
@@ -42,8 +42,6 @@ dnl ======================================================================
 dnl               Intel Active CPU probe test of build-host.
 dnl ======================================================================
 
-CFLAGS="$CFLAGS $SIMD_FLAGS -O0"
-
   AS_CASE([$host_os], [darwin*],
     [CC="$CC_BACKUP"
     CFLAGS="-mavx"
@@ -79,6 +77,7 @@ CFLAGS="$CFLAGS $SIMD_FLAGS -O0"
     [CC="$CC_BACKUP"]
     [CFLAGS="$CFLAGS_BACKUP"]]
   )
+CFLAGS="$CFLAGS $SIMD_FLAGS -O0"
 
 if test "x$simd" != xno; then
  if test "x$enable_native_tests" != xno && test "x$simd" = xyes; then


magnumripper:
I'll test this on my macbook soon-ish. But your whole issue is weird: Why is it testing and setting stuff when you told it not to?

ghost:
> Why is it testing and setting stuff when you told it not to?

Yes, I cannot see why the testing (the whole `else` part) is needed. 

But, it's been here for a decade and it works on Linux (we've got dozens of Linux packages CI logs to prove it).
****
**[edited]**

**Note**: <ins>Thinking out loud, this patch should be merged even if we remove all the tool-chain test: there is no reason to MacOS to behave differently than Linux.</ins>

This is an recent example (Cloud Ubuntu 22, AVX):
```
:: configure: Trying to force AVX using default method (--enable-simd=avx).
:: checking if gcc supports -mavx w/ linking... yes
:: checking for extra ASFLAGS... None needed
:: checking for X32 ABI... no
:: checking special compiler flags... Intel x86
:: configure: Testing tool-chain's CPU support with given options    <== is it useful? Anyway, it doesn't hurt.
:: checking for MMX... yes
:: checking for SSE2... yes
:: checking for SSSE3... yes
:: checking for SSE4.1... yes
:: checking for SSE4.2... yes
:: checking for AVX... yes
:: checking for XOP... no
:: checking for AVX2... no
```

MacOS without this patch (cloud, AVX):
```
configure: Trying to force AVX using default method (--enable-simd=avx).
checking if clang supports -mavx w/ linking... yes
checking for extra ASFLAGS...  -DUNDERSCORES -DBSD -DALIGN_LOG
checking for X32 ABI... no
checking whether OS X 'as' needs -q option... no                   <== oops
configure: Testing tool-chain's CPU support with given options     <== 
checking for MMX... yes
checking for SSE2... yes
checking for SSSE3... yes
checking for SSE4.1... yes
checking for SSE4.2... no
checking for AVX... no

```

solardiz (approved):

PR #5286, opened 2023-04-21
Discard -m64 from LDFLAGS

lucic71:


lucic71:
Closes #5240

ghost:
Since it passed CI, that's fine with me, but probably magnum wants to say something.

BTW: 
* we believe you have tested it and it solves your problem;
* the commit message should be more like this:

autoconf: discard -m64 from LDFLAGS

magnumripper (approved):
It's probably perfectly fine but I have no time for regression testing. Is this the only place we needed to remove it though? I would have guessed more places but it was many many years ago (and the m4 files are a mess).

ghost:
@lucic71 could you please amend the commit message? I cannot Squash and Merge using GitHub UI.

You have troubles:
```
git commit -a --amend # just edit the text
git push origin lucic71:bleeding-jumbo -f
```
BTW: could you add a note to Changes-jumbo:

Something simple as:
```
Fix configure. Remove '-m64' from configure, do not add -m64 in LDFLAGS.
```
Thanks.
****
> It's probably perfectly fine but I have no time for regression testing. Is this the only place we needed to remove it though? I would have guessed more places but it was many many years ago (and the m4 files are a mess).

I haven't found any other places except Makefile.legacy. And I'd rather leave it alone.

solardiz:
> could you add a note to Changes-jumbo:

We only have `CHANGES-jumbo` (which we haven't updated since our release in 2019 and _might_ update before next release again using git history or drop it) and `NEWS` (into which noteworthy changes go, and this one is not). So let's not add any note about the minor change here.

As to the commit message, yes, prefixing it with `autoconf: ` or `configure*: ` would be appropriate, but I wouldn't insist.

solardiz:
And yes, let's leave `Makefile.legacy` alone. It does use `-m64` in some make targets, but that's because it was needed in those. Incidentally, this gives us a check list for which platforms to test this PR's change on.

solardiz:
We auto-detect Solaris 10 and 11 sparc64 systems as 32-bit and produce a working 32-bit build, but `gmake -f Makefile.legacy solaris-sparc64-gcc` produces a working 64-bit build, which runs faster - and it needs the `-m64` on linking there. Omitting `-m64` from `LDFLAGS` results in:

```
ld: fatal: file jumbo.o: wrong ELF class: ELFCLASS64
ld: fatal: file john_mpi.o: wrong ELF class: ELFCLASS64
ld: fatal: file DES_fmt.o: wrong ELF class: ELFCLASS64
```

and so on for all other object files built with `-m64`.

Given this knowledge, instead of merging this PR should we perhaps fix auto-detection of 64-bit for systems like this?

ghost:
> it needs the `-m64` on linking there.

Like magnum predicted. 

Well, at least for now, we should avoid to merge.
***

It seems it is possible to CI using Solaris: I am going to try.

solardiz (approved):
Looks good to me. Thanks!

solardiz:
Now the commit message isn't exactly right because with this change we don't add (not discard) `-m64` to all of `CFLAGS`, `ASFLAGS`, and `LDFLAGS` (not only `LDFLAGS`). But I'll go ahead and merge this anyway - it's taking us too much time already.

ghost:
>
> Now the commit message isn't exactly right because ...
>

Something to think about: re-enable the squash option so that the approver
can edit the commit message if needed.

magnumripper:
It's disabled? Where, how?

ghost:
> It's disabled? Where, how?

There are no more PRs available here. Anyway, you can understand by looking at the picture below (merge commit is disabled):

image

Edit at (probably at https://github.com/openwall/john/settings): 

image

PR #5291, opened 2023-04-22
Solaris sparc64 portability fixes

solardiz:
Tested on Solaris 10, 11, 32-bit OpenMP configure, 64-bit non-OpenMP legacy (four combinations). Hopefully, these improve portability elsewhere more than they break anything. Let's see what CI says.

### DIFF
--- a/src/DES_bs_b.c
+++ b/src/DES_bs_b.c
@@ -94,7 +94,7 @@ typedef uint32x2_t vtype;
 	(dst) = vshr_n_u32((src), (shift))
 
 #elif defined(__ALTIVEC__) && DES_BS_DEPTH == 128
-#ifdef __linux__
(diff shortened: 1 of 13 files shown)


solardiz:
CI is happy (except for `more-tests`, which is pending as usual). I also tested `make -f Makefile.legacy linux-x86-64-avx -j32`.

solardiz:
I was using `export MAKE=$_cv_gnu_make_command`, but the `export` is unnecessary and requires that the shell allows for both assignment and export on the same line (something the rest of `configure` script does not assume, using separate lines for such things). I dropped the `export` now. Re-tested on Solaris 10, still OK.

ghost (approved):
This was tested on X86_64 Solaris 11.4

PR #5294, opened 2023-04-23
Use floating-point for speed reporting in status.c

solardiz:
This lets us use ticks all the time, instead of sometimes having to switch to seconds (lower precision) to avoid integer overflows like we did before.

### DIFF
--- a/src/bench.c
+++ b/src/bench.c
@@ -620,21 +620,19 @@ void benchmark_cps(uint64_t crypts, clock_t time, char *buffer)
 		return;
 	}
 
-	unsigned int cps = crypts * clk_tck / time;
-	uint64_t cpsl = crypts * clk_tck / time;
+	uint64_t cps = crypts * clk_tck / time;
 
-	if (cpsl >= 1000000000000ULL) {
-		sprintf(buffer, "%uG", (uint32_t)(cpsl / 1000000000ULL));
-	} else if (cpsl >= 1000000000) {
-		sprintf(buffer, "%uM", (uint32_t)(cpsl / 1000000));
-	} else
-	if (cps >= 1000000) {
-		sprintf(buffer, "%uK", cps / 1000);
+	if (cps >= 1000000000000ULL) {
+		sprintf(buffer, "%uG", (unsigned int)(cps / 1000000000ULL));
+	} else if (cps >= 1000000000) {
+		sprintf(buffer, "%uM", (unsigned int)(cps / 1000000));
+	} else if (cps >= 1000000) {
+		sprintf(buffer, "%uK", (unsigned int)cps / 1000);
 	} else if (cps >= 100) {
-		sprintf(buffer, "%u", cps);
+		sprintf(buffer, "%u", (unsigned int)cps);
 	} else {
 		unsigned int frac = crypts * clk_tck * 10 / time % 10;
-		sprintf(buffer, "%u.%u", cps, frac);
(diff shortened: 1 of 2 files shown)


solardiz:
@magnumripper I did very little testing of this, you'll probably want to do some more before we merge. Thanks!

solardiz:
We have an issue with x86 MMX here, at least in theory. In `x86-mmx.S`, we skip the EMMS instruction thinking that we probably wouldn't need it. Then in the many implementations of `get_progress()` we invoke it. This worked well so far.

Now we kind of need to add EMMS to the places where we add usage of floating-point here. However, actually doing so in all of such places hurts (in a test I just ran), breaking the ETA reporting. Apparently, using EMMS when there's already required state in FP registers corrupts such state, and we cannot easily isolate its usage only to required places yet be confident we do it after all transitions from MMX. With two functions - caller and callee - both using floating-point, should EMMS be in the caller or the callee? (We do have such cases here. Even simply passing 0 as a `double` argument to a called function is in itself already potential usage of floating-point.) If in the callee only, it can be too late. If in the caller only, it leaves the callee unsuitable for arbitrary calls from elsewhere. If in both, it can corrupt the FP state that might need to stay across the call.

Maybe that's a reason to do EMMS in `x86-mmx.S` just in case, at slight performance impact for those ancient systems.

Anyway, in this PR as it currently is the problem is not seen for me in a test MMX build I made. (It is seen when I try to add more EMMS.)

Of course, such legacy stuff is not really worth the trouble... but there's certain hack value in supporting such a wide range of systems, including legacy.

magnumripper:
> @magnumripper I did very little testing of this, you'll probably want to do some more before we merge. Thanks!

I tried rebasing my bitmaptest branch onto this and it seems to work well. I'll continue using this tomorrow

solardiz:
> In `x86-mmx.S`, we skip the EMMS instruction thinking that we probably wouldn't need it.

We also have MMX DES code in `x86.S`, which may be used by bsdicrypt set_key() even when building with SSE2+ support (but 32-bit).

We also have some MMX intrinsics in a number of source files, but we don't appear to ever enable `__MMX__` without also enabling `__SSE2__`, so those are never compiled in. If we did, we'd fail here in `pseudo_intrinsics.h`:

```c
#elif __MMX__
#include <mmintrin.h>

typedef __m64i vtype;

#error MMX intrinsics not implemented (contributions are welcome!)
```

So far, I am only able to build for MMX (and with assembly files only, no intrinsics) by using one of the MMX targets in `Makefile.legacy`. Trying to do the same via `configure` I get:

```
configure: Trying to force MMX using default method (--enable-simd=mmx).
checking if gcc supports -mmmx w/ linking... yes
checking for X32 ABI... no
checking for MMX... yes
checking for SSE2... yes
checking for SSSE3... no
checking for arch.h alternative... x86-64.h
```

and thus SSE2. Maybe when run on a CPU that has MMX but not SSE2, it'd actually work here and fail at build time because of the `pseudo_intrinsics.h` `#error`.

magnumripper:
For some time (well, almost exactly nine years ago) we actually planned to support MMX in autoconf, but time was already running out for that technology. Nowadays even 32-bit builds are kinda odd so I can't see us spending any time on MMX. It would be a curious experiment though... maybe I'll do it just for fun once I'm retired and have all time in the world. Provided said world lasts that long in the first place o.O

solardiz:
Reverted `status_get_time` to be integer-only, added one `emms();` - should be good enough for MMX now.

PR #5305, opened 2023-05-01
autotools: standardize binary's commit id size

ghost:
Even Solaris has a version of git that supports the optional length parameter.

I almost used 8, but git from Ubuntu 23.04 uses 9 already.

### DIFF
--- a/src/Makefile.in
+++ b/src/Makefile.in
@@ -55,6 +55,10 @@ JTR_GIT_HEAD_TS = \" $(shell git show -s --format=%ci HEAD 2>/dev/null)\"
 JTR_GIT_VERSION = JUMBO_VERSION "$(JTR_GIT_COMMIT) $(JTR_GIT_HEAD_TS)"
 endif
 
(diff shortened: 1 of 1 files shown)


solardiz (changes requested):
The first commit looks good to me.

The second, not so much:

I think the commit message title is wrong - it says "compare" where it should set "allow to compare" or better yet "optionally exclude git commits from version strings" (and then explain the rationale in the long description, like you already do).

Where does the name `_STABLE_BINARY` come from? Is it our own arbitrary choice? If so, let's make it e.g. `JTR_STABLE_BINARY` to be consistent with other nearby make variables we use.

I think the suffix `-static` is confusing. Let's make it e.g. `-stabilized`, although I find this also non-ideal - any better suggestions?

Nitpick: drop the double space in:
```
 JTR_GIT_VERSION =  JUMBO_VERSION \"-static\"
```

solardiz (approved):

PR #5309, opened 2023-05-08
Switch to python3

bmwiedemann:
Switch to python3
because python2 is long EOL
and gets dropped from distributions (e.g. openSUSE Leap 15.5)


solardiz (changes requested):
This is not acceptable, at least not as-is. Our current convention is to use explicit `python2` or `python3` for scripts that don't work (or are expected not to work) with the other version, and to use generic `python` for scripts that work (or are expected to work) with either version. As it is, this PR even changes two instances of `python2` to plain `python`, which likely makes those two scripts non-working on modern systems (as they are, they could work if Python 2 is installed) - or did you test that they actually work with Python 3?

It also changes many scripts to use `python3` instead of `python`. That's fine for modern systems, but it drops useful information on which scripts are also usable with Python 2, which is handy on older systems that are still in use. Is this change really necessary on openSUSE? Why doesn't it provide a `python` e.g. as a symlink to `python3` if that's all it's going to have?

Can you suggest a better change that would fix things without breaking other things? Thanks!

bmwiedemann:
At least they compile with python3
```
> python3 run/ccache2john.py
Usage: run/ccache2john.py <input credential cache file>

Example: run/ccache2john.py /tmp/krb5cc_1000

> python3 run/bitwarden2john.py
[WARNING] Please install the plyvel module for full functionality!
```

There could still be UTF-8 string handling issues somewhere, but at least it is not awfully broken.
I asked our python package maintainer, if we can get a symlink.

The other way out would be to officially deprecate python2 support in john, because nobody should be using it anymore.

solardiz:
We need to search past discussions in here on why we left `ccache2john.py` and `bitwarden2john.py` with `python2`.

> deprecate python2 support in john, because nobody should be using it anymore.

I disagree, especially with the latter part. For example, for `bitcoin2john.py` I continue to recommend running it under Python 2 over Python 3, even though we've updated it to support both, because Python 2 comes with a Berkeley DB module that just works for Bitcoin and many altcoin wallets whereas for Python 3 the module is to be installed manually and is more likely to end up with a wrong BDB version.

solardiz:
> We need to search past discussions in here on why we left `ccache2john.py` and `bitwarden2john.py` with `python2`.

I just did. We previously discussed this in #4884. The conclusion was that `ccache2john.py` fails under Python 3, but no conclusion was made for `bitwarden2john.py`. We appear to lack proper inputs to retest these scripts with after a possible switch/conversion of them to be presumably Python 3 compatible. We need such sample inputs.

Bitwarden support was introduced in #3086 / #3102. Unfortunately, no proper input is referenced in there. We can try e-mailing @kholia, who currently does not watch our GitHub notifications but is otherwise still around.

solardiz:
Is there a portable command like `env` also usable from the shebang line but capable of trying more than one program name? We could probably do something with `awk` or `perl` in the shebang line. Alternatively, we could probably turn those Python scripts into shell scripts that embed the Python code and choose and run a Python interpreter on that. All of these options feel way over-engineered and non-elegant: a Python script depending on support for another scripting language.

Without a portable, simple, and elegant `env`-like option, I think we should simply close this PR for now.

If a distro knows it will only ever have `python3` and not provide Python 2 nor anything called `python`, it's up to that distro to patch its package of `john` (e.g., with `sed -i`) to use `python3` for all these scripts (and accept the likelihood that some will end up broken - but perhaps fewer than without that change on that distro). As a reasonable option, it can choose to do so only for scripts where we currently use `python` rather than explicit `python2` - all of those scripts are expected to work with Python 3 (but may in some cases require installation of additional modules that are readily available with Python 2).

PR #5310, opened 2023-05-09
Mute warning of possible truncation

ghost:
This is a solution for "simple" text. 
* If one needs to bail out on a truncation, another macro must be created.
***

IMPORTANT:

Note that the patch was dealing with **an error**, not with a "regular truncation" (the caller should take care of it).

### DIFF
--- a/src/jumbo.h
+++ b/src/jumbo.h
@@ -440,4 +440,9 @@ extern int check_pkcs_pad(const unsigned char* data, size_t len, int blocksize);
  */
 extern char *replace(char *string, char c, char n);
 
+#define SNPRINTF(buffer, sizeof, format, ...)	  \
+	do { \
+		snprintf(buffer, sizeof, format, __VA_ARGS__) < 0 ? buffer[sizeof - 1] = '\0': (void)0; \
+	} while (0)
+
 #endif /* _JTR_JUMBO_H */
--- a/src/opencl_common.c
+++ b/src/opencl_common.c
@@ -350,12 +350,13 @@ static char *opencl_driver_info(int sequential_id)
 	if (gpu_amd(device_info[sequential_id]) &&
 	    get_platform_vendor_id(get_platform_id(sequential_id)) == DEV_AMD) {
 
+		//The buffer may not be NULL terminated, avoid this using the macro.
 		if (major < 1912)
-			snprintf(buf, sizeof(buf), "%s - Catalyst %s", dname, name);
+			SNPRINTF(buf, sizeof(buf), "%s - Catalyst %s", dname, name);
 		else if (major < 2500)
-			snprintf(buf, sizeof(buf), "%s - Crimson %s", dname, name);
+			SNPRINTF(buf, sizeof(buf), "%s - Crimson %s", dname, name);
 		else
-			snprintf(buf, sizeof(buf), "%s - AMDGPU-Pro %s", dname, name);
+			SNPRINTF(buf, sizeof(buf), "%s - AMDGPU-Pro %s", dname, name);
 		snprintf(tmp, sizeof(tmp), "%s", buf);
 	} else
 		snprintf(tmp, sizeof(tmp), "%s", dname);


src/jumbo.h
```diff
@@ -440,4 +440,9 @@ extern int check_pkcs_pad(const unsigned char* data, size_t len, int blocksize);
  */
 extern char *replace(char *string, char c, char n);
 
+#define SNPRINTF(buffer, sizeof, format, ...)	  \
+	do { \
+		snprintf(buffer, sizeof, format, __VA_ARGS__) < 0 ? buffer[sizeof - 1] = '\0': (void)0; \
+	} while (0)
+
```
magnumripper, line 447: According to my man page:
*The snprintf() function will write at most size-1 of the characters printed into the output string (the size'th character then gets the terminating ‘\0’); if the return value is greater than or equal to the size argument, the string was too short and some of the printed characters were discarded.  The output is always null-terminated, unless size is 0.*

Am I missing something here?

ghost: The url I read about this doesn't say anything about the output always being terminated with null (if a truncation happens). E.g.,now I am at ibm.com and they do not guarantee it.

Anyway, people complained to gcc about the warning and they said:
* Unhandled output truncation is typically a bug in the program [...]

To mute it, we need to handle the function return somehow.

solardiz: I suppose the (non-)issue here was that the compiler warned us of possible truncation, not of the impossible lack of NUL termination. Truncation is considered an issue on its own since the resulting string is different from what a programmer might have intended - e.g., when concatenating parts of a pathname, it can end up accessing a wrong file.

We could provide input and output buffers of such sizes that the compiler would know there's no truncation, or we could "use" the return value so that the compiler assumes we're taking care of or at least acknowledging possible truncation.

Maybe simply adding `(void)` before `snprintf()` will do?

magnumripper: You wont get a negative return from `snprintf` at truncation; only for I/O errors and other weird situations (see man page for `write()` as well). So the macro doesn't do what you thought.

Just to make the compiler happy, we could do something like
```c
#define SNPRINTF(buffer, sizeof, format, ...)	  \
	do { \
		snprintf(buffer, sizeof, format, __VA_ARGS__) >= sizeof ? log_event("- String truncated at %s:%u", __FILE__, __LINE__) : (void)0; \
	} while (0)
```

ghost: > Maybe simply adding (void) before snprintf() will do?

Not here. 

Well, if you only want to make gcc shut up, there are the `pragma GCC` clauses. IMO, much worse.

solardiz: > Well, if you only want to make gcc shut up, there are the `pragma GCC` clauses. IMO, much worse.

What is much worse than what? Your changes are really not fixing anything - they're just an elaborate way to shut up the warning.

We can in fact use the `pragma`, within `#ifdef __GNUC__`. Good idea.

ghost: > Just to make the compiler happy, we could do something like
> 
> ```
> #define SNPRINTF(buffer, sizeof, format, ...)	  \
> 	do { \
> 		snprintf(buffer, sizeof, format, __VA_ARGS__) >= sizeof ? log_event("- String truncated at %s:%u", __FILE__, __LINE__) : (void)0; \
> 	} while (0)
> ```

My first version has a `exit()` in it and I'm almost sure MS did something bad in this matter in the past.

Anyway, log "on error" seems nice to me.

solardiz: Can we try:

```c
			snprintf(buf, sizeof(buf), "%s - Catalyst %.32s", dname, name);
```

That is, make possible truncation of the second field explicit. The way `buf` and `dname` are defined, there's 64 more bytes in `buf`, so the compiler can figure out there's no truncation, but we need to ensure the constant string and the `name` fit in 64.

magnumripper: There's also `-Wno-format-truncation` for disabling that warning globally, if we like. Preferably we'd have that mutually exclusive with `-DDEBUG`.

magnumripper: > There's also `-Wno-format-truncation` for disabling that warning globally

Hmm, actually we already do that - except for ASan builds.
```
   dnl False positives from gcc 8
   if (test "x$asan" != xyes); then
      JTR_NOWARN_CHECK([stringop-truncation], 1)
      JTR_NOWARN_CHECK([format-overflow], 1)
   fi

   dnl False positives from gcc 9
   if (test "x$asan" != xyes); then
      JTR_NOWARN_CHECK([format-truncation], 1)
   fi

   dnl Really silly clang warnings stopping us from doing sanity checks
   JTR_NOWARN_CHECK([tautological-constant-out-of-range-compare], 1)
```

solardiz: @claudioandre-br Did you perhaps only see these warnings in a build with ASan?

ghost: > @claudioandre-br Did you perhaps only see these warnings in a build with ASan?

I remember to see this a lot.

Anyway, if the function works fine even when the function fails (nothing is undefined), I see no reason for `gcc` to complain and I can't do anything to fix it.

Closing.

PR #5321, opened 2023-06-16
aix2john.py: Refactor password hash extraction script

tsvenbla:
The password hash extraction script has been refactored to improve readability, maintainability, and compatibility. 

1. The file handling mechanism has been updated to use a `with` block to open files, ensuring that files are closed properly even if an error occurs.

2. The script now uses the `logging` module for error messages instead of writing directly to stderr. This provides more flexibility for handling and redirecting error messages. 

3. The conditional logic for identifying the password type and printing the formatted output has been refactored into separate functions (`print_password` and `parse_line`), making the code more modular and easier to understand. 

4. A regular expression is now used to parse and validate the username. The `re` module is better suited for pattern matching than the previous `split` method. 

5. The script works with both Python 2 and Python 3, thanks to the import of `print_function` from `__future__`. However, it would be best to migrate towards Python 3 entirely to avoid technical debt. 

6. The argparse setup has been updated for better usability. The filename argument is now required, which means the script will automatically display a usage message if the filename is not provided. This is a more user-friendly approach than the previous method of manually checking and exiting.

### DIFF
--- a/run/aix2john.py
+++ b/run/aix2john.py
@@ -1,40 +1,55 @@
 #!/usr/bin/env python
 
 """
-This script is designed to extract and display password hash information from an AIX /etc/security/passwd file. The
-output is sent to stdout and can be used for auditing or analysis. This script supports different password
-hash types (des, md5, smd5, sha256, and sha512), recognizing them automatically.
+This Python script is designed to parse and extract user credentials data from
+an IBM AIX /etc/security/passwd file. The credentials consist of the username
+and the corresponding password hashes.
+
+The parser handles different password hash types including DES, MD5, SMD5,
+SHA256, and SHA512, with automatic recognition based on the hash length. In
+addition to the normal operation, the script provides an option to list all
+users, even those with locked or unset passwords.
+
+Features:
+- Automatic password hash type identification.
+- Option to display all users, regardless of account status.
+- Error handling for file opening exceptions.
 
 Usage:
-	python aix2john.py
+The script can be used from the command line with the following options:
 
-Arguments:
-	-f : Specify a file path. The default is /etc/security/passwd.
- 	-a : Show all users, including those with locked or unset passwords.
+    python aix2john.py -f [filepath] : To specify a file path for processing.
+    The default is /etc/security/passwd.
+    
+    python aix2john.py -a : To include all users, including those with locked
+    or unset passwords.
 
-Example /etc/security/passwd file with different hashes:
-	guest:
+The script prints the username, the password hash, and the password hash type
+for each user in a formatted way.
+
+Example /etc/security/passwd file:
+
+        guest:
         password = *
-	
- 	paul:
+
+        paul:
         password = YFf0/OmVZz6tQ
         lastupdate = 1026394230
-        flags = 
+        flags =
 
-	jonas:
+        jonas:
         password = 5e884898da28047151d0e56f8dc6292773603d0d6aabbdd62a11ef721d1542d8
         lastupdate = 1026394230
-        flags = 
+        flags =
 
-The script parses the file line by line, identifies the username and corresponding password hashes, and then prints
-them to stdout in a formatted way.
+Compatibility:
+The script is compatible with both Python 2 and Python 3 environments.
 
-Note: Some of the functionality offered by this script may overlap with the unshadow program.
-
-Compatible with both Python 2 and Python 3.
-
-Please remember that the usage of this script should comply with all relevant laws and regulations, and it should not
-be used for unauthorized access to password data.
+Security and Compliance:
+This script handles sensitive data and should be used with appropriate caution.
+Ensure your usage of this script complies with all relevant laws and
+regulations, and refrain from using it for unauthorized access to password
+data.
 """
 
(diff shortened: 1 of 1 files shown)


ghost:
>> 5. The script works with both Python 2 and Python 3, thanks to the import of print_function from __future__. However, it would be best to migrate towards Python 3 entirely to avoid technical debt.

Recently a discussion took place on #5309, specially [#5309 (comment).](https://github.com/openwall/john/pull/5309#issuecomment-1541055930).

Thank you.

BTW: 
- your text editor replaced tabs with spaces in the `perl` PR.
- `john` uses a code style (recommends a line length limit of 120). Any reason to use very long lines?

tsvenbla:
The line length exceeded the recommended limit of 120 characters primarily due to the configuration of my development environment rather than a conscious decision to deviate from `john`'s coding style guidelines. I realize the importance of adhering to established code conventions and have updated the commit. I will make the necessary adjustments to align with the recommended line length for future contributions.

ghost (approved):
It is a 2john tool. We should allow more people to use and test it. We trust your testing. 

I'll try to test it (if we have samples in john-samples). If you have goods samples, please create a PR at https://github.com/openwall/john-samples.

solardiz:
> `john` uses a code style (recommends a line length limit of 120).

We might want to revisit this recommendation. A style I commonly use these days, including in JtR, is letting code lines occasionally go up to 131 (since 132 is a standard terminal width, which I use these days), but keep documentation and `git log` and even code comments within 79 (so that no horizontal scrolling is needed to read it even on an 80-wide terminal). For `git log`, this means wrapping git message lines at 75 (`git log` prefixes them with 4 spaces). The latter is also how Linux kernel git messages are typically wrapped, and that's where git originates from.

run/aix2john.py
```diff
@@ -1,87 +1,172 @@
...
+import io
 import re
 
+# Configure logging
+logging.basicConfig(level=logging.ERROR, format="%(levelname)s: %(message)s")
+
+# Define password types and associated lengths
+PASSWORD_TYPES = {
+    "des": 13,
+    "md5": 32,
```
solardiz, line 68: Why 32 for MD5? Does AIX support raw MD5 now? Looking at this script prior to the changes here, it supported smd5, which is 37 chars, sometimes (by command-line option) turning them into standard md5crypt, which is 34 chars.

run/aix2john.py
```diff
@@ -1,87 +1,172 @@
...
+    "smd5": 46,
```
solardiz, line 69: Why 46? The script previously checked for 37, and the smd5 samples we have are 37.

run/aix2john.py
```diff
@@ -1,87 +1,172 @@
...
-    for line in fd.readlines():
-        if re.match('^\s*\S+\s*:\s*$',line):
-            username = line.split(':')[0]
-
-        if "password = " in line and "smd5" in line:
-            h = line.split("=")[1].strip()
-            if len(h) != 37:
-                continue
-            if is_standard:
-                sys.stdout.write("%s:$1$%s\n" % (username, h[6:]))
```
solardiz, line 44: It's a weird feature we had here where by a command-line option we'd force rewrite from smd5 to standard md5crypt prefix. Perhaps this was needed? As the old script's usage message said, it'd be needed when `lpa_options = std_hash=true`.

run/aix2john.py
```diff
@@ -1,87 +1,172 @@
...
-                sys.stdout.write("%s:%s\n" % (username,
-                    h))
-
-        elif "password = " in line and "ssha" in line:
-            h = line.split("=")[1].strip()
-
-            tc, salt, h = h.split('$')
-
-            sys.stdout.write("%s:%s$%s$%s\n" % (username,
-                    tc, salt, h))
```
solardiz, line 55: This is also a weird rewrite we had where we'd split a hash encoding by dollar signs and then recombine it. That looks like a no-op to me, so I don't know why we did that - could be for no reason. However, it suggests that strings with `ssha` in them were of this kind of format, and wouldn't match any of your detected lengths here. So you seem to be dropping support for them.

solardiz (changes requested):
I'm not convinced that this almost-replaced script does what `john` needs. We had some specialized logic here, which these changes appear to remove in favor of just outputting a guessed hash type as a third field - but we have no corresponding logic in `john` to process that further, and even if we did it looks like you're not supporting all of the types that the previous script revision did.

Thank you for the corresponding pull request to john-samples, but unfortunately the input files there don't include full coverage of the hash types. Can we obtain more complete samples?

Finally, this misses a copyright and license statement.

In GitHub terms, I am "requesting changes" as that's the only option to decline changes other than closing the pull request. However, I am actually not convinced these changes are even in the right direction. Please feel free to try and convince me. Somehow your description of this PR only talks about cleanups, not of major functional changes. Sure the script may be cleaner, but that's of no help if its functionality doesn't match the needs.

solardiz:
@kholia @exploide @stufus Please help review this rewrite of `aix2john.py`. Also, it'd help if you have more sample files. I notice there's one in #1910. Thank you!

Issue #5324, opened 2023-06-21
Unable to build on Windows after cygwin update


ghost:
Recently `cygwin` has been updated [1] to:
```
Cygwin 3.4.7
gcc (GCC) 11.4.0
```

Since then we cannot build on Windows:
``` 
configure: error: cannot compute sizeof (size_t)
``` 

The root cause is related to #1093's workaround (default being C89). Maybe also #5297.

Anyway, there is no proposed fix for now (other than removing the C89 flag):
* JtR will have to deprecate some things sooner or later. Probably sooner rather than later.

*******
[1] From:
```
Cygwin 3.4.6
gcc (GCC) 11.3.0
```

ghost:
I reverted https://github.com/openwall/john/commit/f310326a79dc869b0db1329675b1cd072798e436 and it works (Ubuntu LTS):

As seen at https://github.com/openwall/john/commit/f310326a79dc869b0db1329675b1cd072798e436#commitcomment-12118858, why do we keep '-std=gnu89`?
****

Well, other workaround is:
```diff
diff --git a/src/configure b/src/configure
index f159349b5..88198503e 100755
--- a/src/configure
+++ b/src/configure
@@ -6901,8 +6901,8 @@ ac_compiler_gnu=$ac_cv_c_compiler_gnu
 
 fi
       if test 1 -gt 0; then :
-  { $as_echo "$as_me:${as_lineno-$LINENO}: checking if $CC supports -std=gnu89" >&5
-$as_echo_n "checking if $CC supports -std=gnu89... " >&6; }
+  { $as_echo "$as_me:${as_lineno-$LINENO}: checking if $CC supports -std=gnu99" >&5
+$as_echo_n "checking if $CC supports -std=gnu99... " >&6; }
 fi
   ac_ext=c
 ac_cpp='$CPP $CPPFLAGS'

[...]
```

```sh
[...]
checking if gcc supports -std=gnu99... yes
[...]
Version: 1.9.0-jumbo-1+bleeding-c57beb2733 2023-06-22 08:12:50 -0300
Build: cygwin 64-bit x86_64 AVX2 AC OMP
SIMD: AVX2, interleaving: MD4:3 MD5:3 SHA1:1 SHA256:1 SHA512:1
CPU tests: AVX2
$JOHN is /run/
Format interface version: 14
Max. number of reported tunable costs: 4
Rec file version: REC4
Charset file version: CHR3
CHARSET_MIN: 1 (0x01)
[...]

PR #5325, opened 2023-06-22
autotools: revert '-std=gnu89' (commit f310326)

ghost:
Apparently commit e2e868d fixed the remaining issues with gcc 5.x. See also #1093, https://github.com/openwall/john/commit/e2e868db3e153b3f959e119a51703d4afb99c624, and https://github.com/openwall/john/commit/f310326a79dc869b0db1329675b1cd072798e436.

Reconfigured using a:
```
docker run --rm -v $(pwd):/host -w /host ubuntu:18.04 bash -c \
  "apt update && apt install -y autoconf sudo; autoconf --version && cd src && useradd -U -m JtR; sudo -u JtR autoreconf -i"

```

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -6901,55 +6901,6 @@ ac_compiler_gnu=$ac_cv_c_compiler_gnu
 
 fi
       if test 1 -gt 0; then :
-  { $as_echo "$as_me:${as_lineno-$LINENO}: checking if $CC supports -std=gnu89" >&5
-$as_echo_n "checking if $CC supports -std=gnu89... " >&6; }
-fi
-  ac_ext=c
-ac_cpp='$CPP $CPPFLAGS'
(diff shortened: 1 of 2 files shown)


ghost:
Windows CI is Ok. This CI is also happy (it has gcc  5.x and, e.g., gcc 10). 

Closes #5324.

solardiz:
Why is the move of the `CFLAGS="$CFLAGS $SIMD_FLAGS -O0"` line seen in the diff here - haven't we done it before? Have we not regenerated the `configure` file then? If so, I think this should be a separate commit now (first commit in this PR, then make the actual new change with a second commit).

> Apparently commit e2e868db3e153b3f959e119a51703d4afb99c624 fixed the remaining issues with gcc 5.x.

That commit looks really weird/wrong. If we still have those changes/hacks, we'll need to revisit them - but can do so separately from this PR.

solardiz (approved):

Issue #5326, opened 2023-06-24
Compiler not emiting inline code


ghost:
See #1093 and #5324.

As the original fix could be improved, some discussion might be needed. As stated in #5325:

> if we still have those changes/hacks, we'll need to revisit them.
****

This happens in ASAN + make check build only. Everything else is ok.
```
/usr/bin/ld: bt_hash_type_128.o: in function `calc_ht_idx_128':
/home/JtR/project/src/bt_hash_type_128.c:59: undefined reference to `add128'
/usr/bin/ld: bt_hash_type_192.o: in function `calc_ht_idx_192':
/home/JtR/project/src/bt_hash_type_192.c:61: undefined reference to `add192'
collect2: error: ld returned 1 exit status
make[1]: *** [Makefile:1442: ../run/john] Error 1
make: *** [Makefile:194: default] Error 2
```

ghost:
This patch fixes the problem:

```diff
From 4dbc6d107432f7dc9d3fdea2af4800fd31b54344 Mon Sep 17 00:00:00 2001
From: =?UTF-8?q?Claudio=20Andr=C3=A9?= <claudioandre.br@gmail.com>
Date: Sat, 24 Jun 2023 18:24:40 -0300
Subject: [PATCH] maint: force gcc to emit code for 'inline'

---
 src/bt_hash_type_128.c | 2 +-
 src/bt_hash_type_192.c | 2 +-
 2 files changed, 2 insertions(+), 2 deletions(-)

diff --git a/src/bt_hash_type_128.c b/src/bt_hash_type_128.c
index 1f92d352c..c1d5cb2c6 100644
--- a/src/bt_hash_type_128.c
+++ b/src/bt_hash_type_128.c
@@ -23,7 +23,7 @@ inline unsigned int modulo128_31b(bt_uint128_t a, unsigned int N, uint64_t shift
 	return (unsigned int)p;
 }
 
-inline bt_uint128_t add128(bt_uint128_t a, unsigned int b)
+extern inline bt_uint128_t add128(bt_uint128_t a, unsigned int b)
 {
 	bt_uint128_t result;
 	result.LO64 = a.LO64 + b;
diff --git a/src/bt_hash_type_192.c b/src/bt_hash_type_192.c
index b00ce6e60..f96b08e5d 100644
--- a/src/bt_hash_type_192.c
+++ b/src/bt_hash_type_192.c
[...]

PR #5327, opened 2023-06-25
maint: force gcc to emit code for 'inline'

ghost:
It looks like valid C before the change, but the change has some interesting aspects.


solardiz:
I think the title of this PR and commit message are wrong. "Force to emit code" is not what `static inline` does. Rather, `static` correctly limits visibility of those functions to where they're needed. BTW, I think many more functions in these files should be `static`.

solardiz:
Fixes #5326

solardiz (approved):

PR #5328, opened 2023-06-30
loader.c bugfix: Be more picky about detecting pwdump format.

magnumripper:
Any uid or gid field of length 32 in an input file would result in --show=types outputting incorrect information because the line was then erroneously parsed as pwdump.

Not sure if there were other problems than just --show=types. The canonical hash field would still be correct, so perhaps not.

### DIFF
--- a/src/loader.c
+++ b/src/loader.c
@@ -603,7 +603,8 @@ static int ldr_split_line(char **login, char **ciphertext,
 	if (SPLFLEN(1) > MAX_CIPHERTEXT_SIZE) {
 		huge_line = 1;
 	}
-	else if (SPLFLEN(2) == 32 || SPLFLEN(3) == 32) {
+	else if ((SPLFLEN(3) == 32 && strspn(fields[3], HEXCHARS_all) == 32) ||
+	         (SPLFLEN(2) == 32 && strspn(fields[2], HEXCHARS_all) == 32)) {
 		/* PWDUMP */
 		/* user:uid:LMhash:NThash:comment:homedir: */
 		*uid = fields[1];


ghost (approved):
LGTM. I'll merge.

solardiz:
I'm concerned this change could slow down processing of real PWDUMP format files a lot. I think `strspn` can be rather slow, depending on what case it's optimized for under the hood (if at all).

We might want to use our own `ishex` or `hexlen` instead.

@magnumripper Have you benchmarked this?

solardiz:
I benchmarked loading of a 1 million line PWDUMP format file, and there's no significant slowdown. That's on Linux with glibc. Yet we may want to make the below change not to depend on what `strspn` in libc is optimized for.

```diff
+++ loader.c    2023-07-11 20:57:53.590144768 +0200
@@ -26,6 +26,7 @@
 #include <ctype.h>
 
 #include "arch.h"
+#include "common.h"
 #include "misc.h"
 #include "params.h"
 #include "path.h"
@@ -603,8 +604,8 @@
        if (SPLFLEN(1) > MAX_CIPHERTEXT_SIZE) {
                huge_line = 1;
        }
-       else if ((SPLFLEN(3) == 32 && strspn(fields[3], HEXCHARS_all) == 32) ||
-                (SPLFLEN(2) == 32 && strspn(fields[2], HEXCHARS_all) == 32)) {
+       else if ((SPLFLEN(3) == 32 && ishex(fields[3])) ||
+                (SPLFLEN(2) == 32 && ishex(fields[2]))) {
                /* PWDUMP */
                /* user:uid:LMhash:NThash:comment:homedir: */
```

Briefly tested, seems to work fine.

magnumripper:
Thanks. I was also pondering adding logic for not accepting mixed-case hex. Perhaps we could make a function for that in common.c or even amend the current one, with no performance regression. On the other hand, at these lengths there should be virtually no false positives.

PR #5330, opened 2023-07-02
apex2john.py: Refactored and enhanced APEX hash processing script

tsvenbla:
The Oracle APEX hash files processing script has been significantly refactored to improve its structure, error handling, and readability. The changes include:

1. Introduced an `ArgumentParser` to handle command-line arguments: Instead of using `sys.argv` directly, the script now uses the `argparse` library to handle command-line arguments, which provides more flexibility and better error messages.

2. Introduced the `HashFileProcessor` class, which encapsulates the logic for reading and processing a file. It includes methods for reading the file, validating the content, and printing the formatted output. This makes the code more organized and easier to maintain.

3. Improved error handling: The original code ignored lines that could not be processed correctly, while the new code prints an error message to stderr for such lines. This can help diagnose problems with the input files.

4. Removed leading and trailing white space from username, apexhash, and sgid using `str.strip()`: This was previously done using `rstrip()` and `lstrip()` separately for each value.

5. The output is formatted using `str.format()`, instead of the old-style string formatting using `%`.

6. Replaced a general `except:` block with a more specific `except ValueError as e:` to catch exceptions that occur when processing a line, providing more detailed error information.

7. Added extensive docstrings to functions and the `HashFileProcessor` class to improve code documentation.

8. Ensured PEP8 compliance, with each line of code limited to 79 characters and each docstring/comment limited to 72 characters for better readability.

### DIFF
--- a/run/apex2john.py
+++ b/run/apex2john.py
@@ -1,27 +1,126 @@
 #!/usr/bin/env python
+# -*- coding: utf-8 -*-
 
+"""
+A script for processing Oracle APEX password hashes in a specific format.
+
+Each line in the file should have three comma-separated values: username,
+apexhash, and sgid. Each of these values are stripped of leading and trailing
+white space. The script will write to stdout in the format
+"$dynamic_1$<apexhash><sgid><username>".
+
+Lines that do not adhere to this format will be skipped, and an error message
+indicating the line number and content will be written to stderr.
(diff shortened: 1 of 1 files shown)


tsvenbla:
Forgot to add some files, will open a new one.

PR #5359, opened 2023-09-22
Minimum viable nix-shell with opencl support.

lambdajack:
Adding minimum viable `nix-shell` configuration for use in [Nix/Nixos](https://nixos.org/) with opencl support. The shell supports both AMD and Intel opencl implementations

NOTE: John is available directly through the nixos repository (https://github.com/NixOS/nixpkgs/blob/nixos-23.05/pkgs/tools/security/john/default.nix#L85) however that implementation does not support opencl and limits custom building.

The following optional libraries/features are enabled in this shell:

```
Memory map (share/page large files) ................ yes
Fork support ....................................... yes
OpenMP support ..................................... yes (not for fast formats)
OpenCL support ..................................... yes
Generic crypt(3) format ............................ yes
OpenSSL (many additional formats) .................. yes
libgmp (PRINCE mode and faster SRP formats) ........ yes
128-bit integer (faster PRINCE mode) ............... yes
libz (7z, pkzip and some other formats) ............ yes
libbz2 (7z and gpg2john bz2 support) ............... yes
libpcap (vncpcap2john and SIPdump) ................. yes
Non-free unrar code (complete RAR support) ......... yes
librexgen (regex mode, see doc/README.librexgen) ... no
OpenMPI support (default disabled) ................. no
Experimental code (default disabled) ............... no
ZTEX USB-FPGA module 1.15y support ................. no

```

`./src/configure` and `make` tested in a `--pure` nix-shell, meaning it should be reproducible on any nixos installation. make compiles with no errors.

`./run/john --test` runs with no errors and expected opencl speeds.

This should serve as a 'good-enough' base for most requirements, while being a simple starting point for more advanced configuration when building in a nix environment.

`shell.nix` placed in the root directory as recommended by this nix wiki.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,57 @@
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted.
+# 
+# There's ABSOLUTELY NO WARRANTY, express or implied.
+# -------------------------------------------------------------------------------
+# Nix shells are used create development environments which provide the necessary 
+# tools/dependencies to develop/build software. The environment is typically 
+# declared in a shell.nix file in the softwares root directory. 
+# 
+# To create the nix-shell, run: `nix-shell ./shell.nix`.
+# 
+# This will place you in a development shell with the minimum required dependencies 
+# to build openwall/john with opencl support.
+#
+# NOTE: the 'Official OpenCL runtime for Intel CPUs' is unfree software and
+# therefore, when invoking this shell, nix will require the $NIXPKGS_ALLOW_UNFREE=1
+# environment variable to be set. If you wish to create your shell with free software 
+# only, then you can remove 'intel-ocl' entry from 'nativeBuildInputs' below. Please
+# note however if you are running and intel based system and remove intel-ocl then
+# you will not have opencl support when building openwall/john. For the avoidance of 
+# doubt, this file does not contain any unfree software.
+#
+# More information about Nixos: https://nixos.org/
+# More information about nix-shell: https://nixos.org/manual/nix/stable/command-ref/nix-shell.html
+#
+# Copyright Jack Bizzell (lambdajack) 2023
+
+{ pkgs ? import <nixpkgs> {} }:
+let
+	perlEnv = pkgs.perl.withPackages (p: with p; [
+		CompressRawLzma
+		DigestMD4 
+		DigestSHA1 
+		GetoptLong
+		perlldap
+	]);
+	pythonEnv = pkgs.python3.withPackages(p: with p; [
+		dpkt 
+		scapy 
+		lxml
+		wrapPython
+	]);
+in
+pkgs.mkShell {
+	nativeBuildInputs = with pkgs.buildPackages; 
+		[ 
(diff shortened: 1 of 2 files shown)


shell.nix
```diff
@@ -0,0 +1,29 @@
...
+	nativeBuildInputs = with pkgs.buildPackages; 
+		[ 
+			openssl libzip rocm-opencl-runtime opencl-headers 
+			bzip2 libpcap libgmpris libxcrypt gmp intel-ocl
+			gcc zlib nss nspr libkrb5 re2 makeWrapper
+			perlEnv pythonEnv
+		];
+	shellHook = ''
+     		export AS=$CC
+			export LD=$CC
```
solardiz: Shouldn't the two `export` lines be at same indentation?

shell.nix
```diff
@@ -0,0 +1,29 @@
...
+  	'';
+}
```
solardiz: Apparently, missing linefeed after the last line - add it?

solardiz (changes requested):
Hi. Thanks for the contribution! I'm not familiar with Nix, but I'll make some common sense comments.

For the commit message, can you please reformat it (amend and force-push the one commit) so that it has a title line of at most 72 chars, blank line, and then subsequent paragraphs are wrapped at 75 (except for lines that are naturally longer - e.g., long URLs or copy-paste from elsewhere)? Similar style is used on Linux kernel commit messages.

Please also add a `NEWS` file entry. Currently it would be right after this one:

```
- Added krb5tgs-sha1[-opencl] formats for etypes 17 and 18.  [magnum; 2023]
```

Finally, are comments allowed inside the `shell.nix` file? If so, perhaps it should start with a comment explaining what it is, as well as its copyright statement (that you're the copyright holder?) and license - ideally, our usual 0-clause BSD:

```
Redistribution and use in source and binary forms, with or without
modification, are permitted.

There's ABSOLUTELY NO WARRANTY, express or implied.
```

I understand this may be too minor to be copyrightable, but explicitly having it under a permissive license like that shouldn't hurt.

solardiz:
> Update indentation in shell.nix

Such fix-ups within a PR should ideally be made as amends and force-pushes of the commit you're fixing, not as more commits added. If you keep adding commits, we'll need to use GitHub's "squash and merge" on this PR, which is non-ideal.

lambdajack:
I've added the copyright, license and fixed indentation onto just the one commit.

lambdajack:
I'm finished force pushing typos now I think........

solardiz (approved):
I still notice minor things, like this:

```
license, copywrite, info added.
```

which doesn't belong to the commit message (it's part of the temporary history within this PR, and makes little sense in one and only final commit message).

But I don't mind merging this as-is.

Nebucatnetzer:
To make it work with Nvidia add this part to mkShell worked for me.

```
LD_LIBRARY_PATH= pkgs.lib.makeLibraryPath [
    pkgs.ocl-icd
];
```

PR #5364, opened 2023-10-02
Update INSTALL for Mac OS Big Sur

squio:
Update INSTALL for
-  Mac OS Big Sur `v11.7.10 (20G1427)`
- Homebrew `v4.1.14`
- Platform `X86_64`

### DIFF
--- a/doc/INSTALL
+++ b/doc/INSTALL
@@ -176,9 +176,9 @@ that probably work fine but these instructions are for "Homebrew":
   5. Configure, possibly adding a CC option for pointing to a specific gcc and
      using whatever LDFLAGS and CPPFLAGS was recommended when you installed
      Homebrew's OpenSSL:
-        ./configure CC="/opt/homebrew/bin/gcc-10" \
-          LDFLAGS="-L/opt/homebrew/opt/openssl/lib" \
-          CPPFLAGS="-I/opt/homebrew/opt/openssl/include"
+        ./configure CC="/usr/local/bin/gcc-13" \
+          LDFLAGS="-L/usr/local/opt/openssl/lib " \
+          CPPFLAGS="-I/usr/local/opt/openssl/include"
   6. Clean old files and make:
         make -s clean && make -sj4
 


solardiz (approved):
Thanks. I guess this will keep getting out of date (e.g., the specific gcc version reference), but I don't mind accepting this edit. Also, I guess the added space character after `lib` is a typo, but it should be harmless and I suppose it's not easy for you to amend the commit if you've made it via GitHub UI (judging by the `patch-1` branch name). Just something for us to fix in a further cleanup.

squio:
Hi, this will definitely be a moving target, and I'm already behind (old Intel macbook, sorry!).
Also sorry for being really lazy and doing the quick GitHub UI fork / online edit / PR thing - I really should have used my IDE instead with its linting settings...

Many thanks!!!

doc/INSTALL
```diff
@@ -176,9 +176,9 @@ that probably work fine but these instructions are for "Homebrew":
   5. Configure, possibly adding a CC option for pointing to a specific gcc and
      using whatever LDFLAGS and CPPFLAGS was recommended when you installed
      Homebrew's OpenSSL:
-        ./configure CC="/opt/homebrew/bin/gcc-10" \
-          LDFLAGS="-L/opt/homebrew/opt/openssl/lib" \
-          CPPFLAGS="-I/opt/homebrew/opt/openssl/include"
+        ./configure CC="/usr/local/bin/gcc-13" \
+          LDFLAGS="-L/usr/local/opt/openssl/lib " \
+          CPPFLAGS="-I/usr/local/opt/openssl/include"
```
magnumripper, line 181: I think I recall noticing that my MacBook has it one of the ways (i.e. the `/opt` vs. `/usr/local`) while the M1 I can borrow sporadically has the other. Both have Homebrew installed more or less the same way. Either it's because mine had HomeBrew for two years longer, or it's somehow caused by the fact the latter is M1 while I still have an intel.

PR #5370, opened 2023-10-23
Add overhauled version of pdf2john.py

### LINKED ISSUE #5369: pdf2john major overhaul
https://github.com/benjamin-awd/pdf2john/issues

@benjamin-awd will you honor us a PR once you're done with it please? And we/you can try it for [our open issues](https://github.com/openwall/john/issues?q=is%3Aopen+pdf2john) as well as [hashcat's](https://github.com/hashcat/hashcat/issues?q=is%3Aopen+pdf2john).

benjamin-awd:
Closes https://github.com/openwall/john/issues/5369

This implementation of pdf2john.py is from here: https://github.com/benjamin-awd/pdf2john

### DIFF
--- a/doc/NEWS
+++ b/doc/NEWS
@@ -307,6 +307,8 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
 
 - Added shell.nix with opencl support on AMD and Intel. [lambdajack; 2023]
 
+- Added modern, overhauled version of pdf2john.py. [benjamin-awd; 2023]
+
 Major changes from 1.8.0-jumbo-1 (December 2014) to 1.9.0-jumbo-1 (May 2019):
 
 - Updated to 1.9.0 core, which brought the following relevant major changes:
--- a/doc/README-PDF
+++ b/doc/README-PDF
@@ -1,10 +1,12 @@
 Cracking PDF files with JtR
 ---------------------------
 
-1. Run pdf2john.pl on the .pdf file(s).
+You can use pdf2john.py (Python) or pdf2john.pl (Perl)
 
-E.g. $ ../run/pdf2john.pl test.pdf > hashes
+1. Run pdf2john.py on the .pdf file(s).
 
-2. Run john on the output of pdf2john.pl program.
+E.g. $ ../run/pdf2john.py test.pdf > hashes
+
+2. Run john on the output of pdf2john.py program.
 
 E.g. $ ../run/john hashes
--- a/run/pdf2john.py
+++ b/run/pdf2john.py
@@ -1,0 +1,149 @@
+#!/usr/bin/env python3
+
+# This software is Copyright (c) 2023 Benjamin Dornel <benjamindornel@gmail.com>
+# and it is hereby released to the general public under the following terms:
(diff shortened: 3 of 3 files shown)


solardiz:
Thank you @benjamin-awd!

What does this change for the users, besides the different dependencies? Like specific kinds of PDFs that were not supported before, but now are, or vice versa?

Can you please add a `doc/NEWS` entry?

Is the script Python 3 only, or supporting both Python 2 and 3? If it's 3-only, then please use `python3` on the shebang line - this is not perfect, but it's our current (perhaps temporary) convention here.

Let's keep the Perl script as well, at least because its dependencies differ (starting with Perl vs. Python itself, but also modules). It could be easier to get one or the other working in a particular environment. Maybe mention this in the documentation?

If you make changes per this feedback (or otherwise) within this PR, please amend the commit and force-push. Do not add more commits.

magnumripper:
I'm planning some testing to see where this pdf2john is better (or worse). Soon-ish...

benjamin-awd:
> What does this change for the users, besides the different dependencies? Like specific kinds of PDFs that were not supported before, but now are, or vice versa?

Hmm I think the general points are:
- Moves closer to the single responsibility principle by delegating PDF parsing and handling to pyhanko. This means that we can worry less about general PDF weirdness e.g. the encryption dictionary having multiple definitions for the same key. For example, if your encrpytion dictionary has two definitions for the /Filter key, pyhanko will raise a warning like `Multiple definitions in dictionary at byte 0x1eb for key /Filter`
- The side effect of this is that pdf2john now has much less code complexity, which improves readability/maintability. This should make it easier to adapt this for future PDF specifications.
- An optional debugging flag that shows the encryption dictionary of a PDF.
- A test suite (though this only exists on https://github.com/benjamin-awd/pdf2john currently -- I'm not sure how to migrate it over here)

In terms of specifics, I realized the old pdf2john.py had some limitations, especially around the hardcoded escape sequence map, where some characters like `\\0` were not included. It was also quite hard to read, and update.

I think both this and the perl version are capable of handling the latest PDF specification with an r6 security handler

benjamin-awd:
> I'm planning some testing to see where this pdf2john is better (or worse). Soon-ish...

SGTM -- I grabbed a couple of the old PDF files I could find on the openwall website, but might not have covered all edge cases. If you find anything that doesn't work, let me know and I can fix that and add to the current test suite.

Besides the test cases, this version of pdf2john is definitely slower compared to perl, and the original pdf2john due to the greater level of parsing and validation that pyhanko does.

benjamin-awd:
> Can you please add a doc/NEWS entry?

I've added the doc/NEWS entry.

> Is the script Python 3 only, or supporting both Python 2 and 3? If it's 3-only, then please use python3 on the shebang line - this is not perfect, but it's our current (perhaps temporary) convention here.

I've updated the shebang line.

> Let's keep the Perl script as well, at least because its dependencies differ (starting with Perl vs. Python itself, but also modules). It could be easier to get one or the other working in a particular environment. Maybe mention this in the documentation?

👍 ✅

solardiz (approved):
I was hoping we'd also see some testing results by magnum, but the changes here look good to me given that we also keep the Perl script now.

Issue #5362, opened 2023-09-28
bitlocker2john: VMK not encrypted with AES-CCM


s2mael:
Hello, 

I'm trying find a hash from disk but I get a error:

![Screenshot_1](https://github.com/openwall/john/assets/140153594/fbabfd8f-876c-43e4-93a5-0ec859a68aba)

[dev_sda2_first_256k.zip](https://github.com/openwall/john/files/12749496/dev_sda2_first_256k.zip)

solardiz:
Hi @s2mael. What version/build of JtR are you getting this with?

Overall, it looks to me that `bitlocker2john` doesn't find what it supports, and instead finds headers that look very much like what it supports but are not quite it. Note how the reported `Salt` values do not look random, so something is probably wrong already by that point. OTOH, it could be helpful to compare this against output of a successful run (against another disk image).

Thank you for attaching the first 256k, but unfortunately these almost-supported headers are way outside of that range. You can possibly provide an excerpt starting at 0xc1880000:

```
dd if=/dev/sda2 bs=64k skip=49544 count=4 of=0xc1880000_256k.bin
```

solardiz:
@s2mael Also, I notice you're running `bitlocker2john` on Windows against a 1.5 TB image file, right? But then the attached filename suggests you also access the disk on Linux? I suggest you run the whole thing on Linux.

Also, going through 1.5 TB of data should be taking a very long while... but I guess for you the search is aborting somewhere within the first 4 GiB because of a bug we had in older versions, fixed in #5099.

How long does the search take for you?

s2mael:
@solardiz You're right, I run it on Windows, I also runed Linux, but it was version 7 of bitlocker2john, I'll try to run version 8 next week and let you know, also attach the file.  About Windows, it takes a couple of hours, maybe 4-5 or more.

s2mael:
@solardiz You're right, I run it on Windows, I also runed Linux, but it was version 7 of bitlocker2john, I'll try to run version 8 next week and let you know. About Windows, it takes a couple of hours, maybe 4-5 or more.

solardiz:
> it was version 7 of bitlocker2john, I'll try to run version 8 next week

This makes no sense to me - there isn't a `bitlocker2john` version number at all. I asked you about JtR version numbers and specific builds of it that you use (as you get `bitlocker2john` from a JtR tree, right?)

[...]

PR #5376, opened 2023-11-20
Assorted minor fixes

solardiz:
For bitlocker2john, bench.c, relbench, NEWS.


solardiz:
I tested the `bitlocker2john` changes here against `john-samples/BitLocker/*.vdi` and the second file attached in #5362 ([0x1880000_256b.zip](https://github.com/openwall/john/files/12788375/0x1880000_256b.zip)), then diff'ed the output (stdout+stderr) against that produced by the version from just prior to my changes here - only saw expected differences.

Issue #3191, opened 2018-03-07
 JtR on Windows: OpenCL DLL usage


dwendt:
This bug initially started with `clGetPlatformIDs` returning `-1001` which is  `CL_PLATFORM_NOT_FOUND_KHR` and john dies which is wrong/bad. 

hashcat on the same system detects the platform fine using literally the same function. The only conclusion I can come to is that hashcat is doing something correctly in their build, and john is not. 

### Steps to reproduce
1. grab the first build of bleeding-jumbo you can find on google http://openwall.info/wiki/_media/john/johntheripper-v1.8.0.12-jumbo-1-bleeding-e6214ceab-2018-02-07-win-x64.7z
2. install a graphics card whose platform is opencl 1.2 
3. run `john.exe --list=opencl-devices` with a card that's restricted to OCL 1.2 platform (this should work fine with 2.1/2.2 headers...)

### System configuration
windows enterprise 10.0.16299.19, 64-bit
```
H:\john180j1w\bleeding>.\run\john.exe --verbosity=5 --list=opencl-devices
Warning: '/dev/shm' does not exists or is not a directory.

POSIX shared memory objects require the existance of this directory.
Create the directory '/dev/shm' and set the permissions to 01777.
For instance on the command line: mkdir -m 01777 /dev/shm
initUnicode(UNICODE, UTF-8/ISO-8859-1)
UTF-8 -> UTF-8 -> UTF-8
Error: No OpenCL-capable platforms were detected by the installed OpenCL driver.
Throw clError: clGetPlatformIDs() = -1001
Error: No OpenCL-capable devices were detected by the installed OpenCL driver.


H:\john180j1w\bleeding>.\run\john.exe --list=build-info
Warning: '/dev/shm' does not exists or is not a directory.

POSIX shared memory objects require the existance of this directory.
Create the directory '/dev/shm' and set the permissions to 01777.
For instance on the command line: mkdir -m 01777 /dev/shm
Version: 1.8.0.12-jumbo-1-bleeding-e6214ceab 2018-02-07 17:56:37 +0100
Build: cygwin 64-bit x86_64 SSE4.2 AC MPI + OMP
SIMD: SSE4.1, interleaving: MD4:3 MD5:3 SHA1:1 SHA256:1 SHA512:1
CPU tests: SSE4.2
$JOHN is ./run/
Format interface version: 14
Max. number of reported tunable costs: 4
Rec file version: REC4
Charset file version: CHR3
[...]

PR #5386, opened 2023-12-12
OpenCL: Dynamically load OpenCL library

alainesp:
A PR to support loading of OpenCL at execution time. No modification is done to the configure script, so it stills checks for OpenCL headers files and statically links to the library if it is found.

### DIFF
--- a/src/generate_opencl_dynamic_loader.py
+++ b/src/generate_opencl_dynamic_loader.py
@@ -1,0 +1,176 @@
+#-------------------------------------------------------------------------------------
+# Dynamic OpenCL library loader.
+#
+# This software is copyright (c) 2023, Alain Espinosa <alainesp at gmail.com> and it
+# is hereby released to the general public under the following terms:
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted.
+#-------------------------------------------------------------------------------------
+
+# Import modules
+import re
+
+# OpenCL version supported
+CL_TARGET_OPENCL_VERSION = 120
+
+opencl_header = open("cl.h")
+header_text: str = opencl_header.read()
+opencl_header.close()
+
+# Write C file
+dynamic_loader = open("opencl_dynamic_loader_plug.c", "w")
+dynamic_loader.write(
+'''//-------------------------------------------------------------------------------------
+// Dynamic OpenCL library loader. Automatically generated.
+//
+// This software is copyright (c) 2023, Alain Espinosa <alainesp at gmail.com> and it
+// is hereby released to the general public under the following terms:
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted.
+//-------------------------------------------------------------------------------------
+#ifdef HAVE_OPENCL
+
+#ifndef CL_TARGET_OPENCL_VERSION
+''')
+dynamic_loader.write(f'#define CL_TARGET_OPENCL_VERSION {CL_TARGET_OPENCL_VERSION}')
+dynamic_loader.write(
+'''
+#endif
+
+#ifdef __APPLE__
+#include <OpenCL/opencl.h>
+#include <OpenCL/cl_ext.h>
+#else
+#include <CL/cl.h>
+#include <CL/cl_ext.h>
+#endif
+
+// DLL handle
+static void* opencl_dll = NULL;
+static void load_opencl_dll();
+
+''')
+
+REGEX_C_ID = '[_a-zA-Z][_a-zA-Z0-9]*'
+REGEX_ANY_SPACES = '[ \t\n\r]*'
+REGEX_SPACES_ONE_OR_MORE = '[ \t\n\r]+'
+
+# Search for all function definitions
+funtions = re.findall(
+    f"extern CL_API_ENTRY (CL_API_PREFIX__VERSION_[1-9]_[0-9]_DEPRECATED )?({REGEX_C_ID}[ \*]*) CL_API_CALL{REGEX_SPACES_ONE_OR_MORE}({REGEX_C_ID})\(([_a-zA-Z0-9, \t\r\n\*\(\)]*)\) CL_API_SUFFIX__VERSION_([0-9_]+)(_DEPRECATED)?;", 
+    header_text)
+
+# Handle this special case
+special_function_name = "clGetPlatformIDs"
+special_function_code = '''load_opencl_dll();
+
+        if (!opencl_dll)
+        {
+                // Our implementation
+                if ((num_entries == 0 && platforms) || (!num_platforms && !platforms))
+                        return CL_INVALID_VALUE;
+
+                if (num_platforms)
+                        *num_platforms = 0;
+                        
+                return CL_SUCCESS;
+        }
+\t'''
+
+# Declare funtions and pointer to functions
+# return_type, function_name, list_params, api_version
+for x in funtions:
+    if len(x) == 6:
+        function_return = x[1]
+        function_name = x[2]
+        function_params = " ".join(x[3].split()) # Better param definition
+        api_version = int(x[4].replace('_', '')) * 10
+        if api_version > CL_TARGET_OPENCL_VERSION:
+            print(f"Function '{function_name}' skipped give api={x[4]}")
+            continue
+        
+        # Begin function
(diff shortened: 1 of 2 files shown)


solardiz:
Thank you, Alain!

> No modification is done to the configure script, so it stills checks for OpenCL headers files and statically links to the library if it is found.

Of course, you don't literally mean "statically" - the linking is dynamic; what I think you mean is that we still have a hard dependency on having an OpenCL dynamic library (of proper filename and soname) installed. There's no regression in that compared to what we had before this PR, so it's not a blocker, but it's something for us to improve now that we can use OpenCL where available without such build-time linking.

Anyway, I think this PR as-is is sufficient to address #3191 because we're currently including Cygwin's OpenCL DLL with Windows binary builds that we make and distribute. So the dynamic linking dependency would be satisfied, like it is now, even on systems that lack OpenCL. And on those that have it, the dynamic loading introduced in this PR should work to provide actual OpenCL support.

alainesp:
An explanation of how the **configure** script now works relating to OpenCL support:

1- JtR configure checks for CL.h existence. 
2-  Then JtR configure  tries to compile and run small OpenCL programs linking to a list of OpenCL libraries (_-lOpenCL -lCL -lclparser -lpocl_), 
3 - Then if all works JtR configure adds a linking option to OpenCL. 

All that configuration/checking is unnecessary, IMO. We can always support OpenCL formats. If no OpenCL library is found the code will return 0 OpenCL platform and that's all. We also already had the option **--disable-opencl** to disable OpenCL formats if the user wants to reduce the size of the JtR executable and doesn't need OpenCL.

What we need to do to reach that place is to add the OpenCL headers to JtR source files and eliminate all the configuration/checking on the _configure_  script.

solardiz:
> add the OpenCL headers to JtR source files

That's #3918, perhaps something you'll address with a next PR. Thanks!

ghost:
> eliminate all the configuration/checking on the _configure_ script.

I like the idea of keep checking for people who are building from source (why add all the OpenCL formats plus tools in the final binary for non-OpenCL users?).

I prefer to remove only the static linking, as seens at:
- [opencl.patch](https://github.com/openwall/john/files/13674372/opencl.txt)

solardiz:
> I like the idea of keep checking for people who are building from source (why add all the OpenCL formats plus tools in the final binary for non-OpenCL users?).

I have mixed feelings about this. We somehow (I don't recall exactly why, need to re-read) wanted to bundle `CL.h` anyway, which hashcat already does. When we do that, it does feel sort of silly to also be checking for the system providing it and/or the library in order to enable OpenCL at compile-time by default. That said, it also feels sort of silly to be including OpenCL support in every build since we still have more non-OpenCL formats and since so many systems that JtR runs on lack OpenCL. Where by "systems" I mean both individual machines/VMs/containers and operating system kinds/archs.

> I prefer to remove only the static linking, as seens at:
> 
>     * [opencl.patch](https://github.com/openwall/john/files/13674372/opencl.txt)

@claudioandre-br Please contribute this change via a separate PR. I think the commit title should focus on what the change does - "Stop linking against the OpenCL library" - and what you currently have in there ("start using dynamic binding") should be moved into the description (keeping in there also the "Continue building a non-OpenCL binary" part). Thank you!

solardiz (approved):
Good enough to merge, although we could want to discuss/correct a few things later - such as the script's location. We used to put files unused during a normal build in the `unused` subdirectory, but this too may be wrong for files that we do intend to use, just not during a normal build.

src/generate_opencl_dynamic_loader.py
```diff
@@ -0,0 +1,176 @@
...
+#include <stdio.h>
+static void load_opencl_dll()
+{
+        int i;
+        if (opencl_dll)
+            return;
+
+        // Names to try to load
+        const char* opencl_names[] = {
+            "libOpenCL.so",      // Linux/others
```
solardiz, line 130: Just to document what we're doing here: loading the `.so` file, as opposed to a `.so.1` or such, is strictly speaking wrong. That file/symlink is intended for use at compile time and on distros that split packages into sub-packages it normally only comes from a development sub-package. However, if such a development sub-package happens to be installed, using the `.so` file for loading the library is a working hack to avoid having to know the exact soname.

src/generate_opencl_dynamic_loader.py
```diff
@@ -0,0 +1,176 @@
...
+            "OpenCL",            // _WIN
+            "/System/Library/Frameworks/OpenCL.framework/OpenCL", // __APPLE__
```
solardiz, line 132: I've already merged this as-is (and I don't regret yet ;-) but @magnumripper and others could want to test that the path given here actually works on a Mac, or/and if some of the other library names listed here work on a Mac. I don't know whether @alainesp tested on a Mac - probably not?

src/generate_opencl_dynamic_loader.py
```diff
@@ -0,0 +1,176 @@
...
+        dynamic_loader.write(f'\t\tif (!ptr_{function_name})\n')
+        dynamic_loader.write('\t\t{\n')
+        dynamic_loader.write(f'\t\t\tall_functions_loaded = 0;\n')
+        dynamic_loader.write(f'\t\t\tprintf("Cannot load {function_name} function\\n");\n')
+        dynamic_loader.write('\t\t}\n')
+
+dynamic_loader.write('''
+            if (!all_functions_loaded)
+            {
+                dlclose(opencl_dll);
```
solardiz, line 165: Here we insist on finding all functions for our target OpenCL version. I suppose someone could want to get JtR working on a partial OpenCL implementation, which does not yet have all the functions, but maybe happens to have all those that we actually use. We could want to re-add a way to do that - somehow lookup/require only functions that we actually use or/and re-add a way (e.g., a `configure` option) to link against OpenCL like we did before.

src/generate_opencl_dynamic_loader.py
alainesp: No, I don't have access to a Mac. That was the path used by Hashcat.

solardiz:
FWIW, when running `generate_opencl_dynamic_loader.py` on https://github.com/KhronosGroup/OpenCL-Headers/blob/main/CL/cl.h as of now, I am getting exactly the same `opencl_dynamic_loader.c` as committed here, except with Unix linefeeds. That's good.

And we'll need to fix the linefeeds for both the script and the generated file. Also fix the weird mix of spaces and tabs in the generated file, I guess to be tabs alone.

src/generate_opencl_dynamic_loader.py
alainesp: We can add a check on the Python script generator to only generate code/checks for functions we use, but all contributors need to know if they will use in the future some non-common OpenCL function, JtR will not compile. Do you want that I make these changes?

solardiz:
This PR's added files are now cleaned up via #5394.

src/generate_opencl_dynamic_loader.py
solardiz: > We can add a check on the Python script generator to only generate code/checks for functions we use, but all contributors need to know if they will use in the future some non-common OpenCL function, JtR will not compile. Do you want that I make these changes?

I went with somewhat different approach in #5395, please help test and review it. Thank you!

solardiz:
In `configure`, we try these: `-lOpenCL -lCL -lclparser -lpocl` (also with varying library paths). With this PR's dynamic loading, we don't try the `CL`, `clparser`, and `pocl` names - should we be trying any of these? What OpenCL implementations use the `CL` and `clparser` names? Does `pocl` (still?) need to be linked/loaded under that name, rather than a generic `-lOpenCL`?

Can someone please try our current code on a system with only `pocl` available?

ghost:
> Can someone please try our current code on a system with only `pocl` available?

Now it uses the -lopencl:

```
configure:14024: checking for OpenCL library
configure:14062: gcc -o conftest -g -O2 -I/usr/local/include   -I/usr/local/include -DCL_SILENCE_DEPRECATION  -L/usr/local/lib -L/usr/lib64 -L/lib64 conftest.c -lOpenCL  -lm  >&5
[...]
configure:14062: $? = 0
configure:14143: result: -lOpenCL
```

```
$ run/john --list=opencl-devices
Platform #0 name: Portable Computing Language, version: OpenCL 2.0 pocl 1.8  Linux, None+Asserts, RELOC, LLVM 11.1.0, SLEEF, DISTRO, POCL_DEBUG
```

PR #5382, opened 2023-12-04
Support argon2 on GPUs through OpenCL

alainesp:

PR #5387, opened 2023-12-13
Argon2-OpenCL Optimization

alainesp:
Many optimizations/quality of live updates to the argon2-opencl format


solardiz:
Somehow our `ci/circle-ci: encoding-cpu` test failed:

```
opencl_argon2_fmt_plug.c: In function 'autotune':
opencl_argon2_fmt_plug.c:480:13: error: invalid storage class for function 'reset'
  480 | static void reset(struct db_main *db)
      |             ^~~~~
opencl_argon2_fmt_plug.c: In function 'reset':
opencl_argon2_fmt_plug.c:509:17: warning: this 'if' clause does not guard... [-Wmisleading-indentation]
  509 |                 if (min_salt_lanes > salt->lanes)
      |                 ^~
opencl_argon2_fmt_plug.c:511:4: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the 'if'
  511 |    if (max_memory_size < memory_size)
      |    ^~
opencl_argon2_fmt_plug.c: In function 'autotune':
opencl_argon2_fmt_plug.c:713:13: error: invalid storage class for function 'ctx_init'
  713 | static void ctx_init(argon2_context *ctx)
      |             ^~~~~~~~
[... many lines omitted ...]
opencl_argon2_fmt_plug.c:967:21: error: invalid storage class for function 'tunable_cost_p'
  967 | static unsigned int tunable_cost_p(void *_salt)
      |                     ^~~~~~~~~~~~~~
opencl_argon2_fmt_plug.c:974:21: error: invalid storage class for function 'tunable_cost_type'
  974 | static unsigned int tunable_cost_type(void *_salt)
      |                     ^~~~~~~~~~~~~~~~~
opencl_argon2_fmt_plug.c:1046:1: error: expected declaration or statement at end of input
 1046 | };
      | ^
opencl_argon2_fmt_plug.c:982:17: warning: unused variable 'fmt_opencl_argon2' [-Wunused-variable]
  982 | struct fmt_main fmt_opencl_argon2 = {
      |                 ^~~~~~~~~~~~~~~~~
```

The above also includes a warning about misleading code indentation.

alainesp:
> Somehow our `ci/circle-ci: encoding-cpu` test failed:
> 
> The above also includes a warning about misleading code indentation.

When I rebase all the commits, I did something wrong, I guess, and the code doesn't work. I am solving it.

alainesp:
> When I rebase all the commits, I did something wrong, I guess, and the code doesn't work. I am solving it.

Now it should work. The problem now is that some of the commits in the history are in a bad state (compilation may fail at that commit). I am not sure how to solve it or if it's needed.

solardiz:
> The problem now is that some of the commits in the history are in a bad state (compilation may fail at that commit). I am not sure how to solve it or if it's needed.

I have mixed feelings about this. The history we already have in jumbo is in poor quality, but I am trying to make my recent contributions of greater quality anyway. OTOH, it's not productive use of your time on revising these commits over and over, and it is unlikely to result in a state that I would be entirely happy with. So I think let's cut the costs here, and accept this PR as-is. The history will unfortunately be somewhat broken.

solardiz:
For others reading this, I've been testing and commenting on Alain's code via e-mail as he worked towards this PR. So this passes my testing now. The speedup achieved is typically small, but is nevertheless desirable. It is on the order of 10% for our `--test` benchmarks, which use 4 MiB x 3 iterations hashes, and typically just a few percent for my actual cracking test with a 16 MiB x 3 iterations hash I use for actual cracking test.

Exception: on the old Titan Kepler, the speedup is almost 40% in actual cracking. Perhaps newer devices were already mostly memory-bound, and this one was compute-bound (it has relatively less compute power while having similar memory bandwidth as GPUs from several years later).

There's still some further speedup achievable by manual GWS tuning (higher GWS than the code auto-tunes to). On NVIDIA, this additional speedup is now tiny. On Vega 64, it's almost 2x in my testing - so we probably need to improve the auto-tuning.

The manual GWS values to use are now several times smaller than those I used in #5382 comments - e.g., for 1080 it's now `-gws=15360` to achieve similar behavior to what was in https://github.com/openwall/john/pull/5382#issuecomment-1839167284 and for Vega 64 it's `-gws=8000` to achieve similar to what we had in https://github.com/openwall/john/pull/5382#issuecomment-1839208723 - I don't know why nor which commit here made this difference.

solardiz:
> The manual GWS values to use are now several times smaller than those I used in #5382 comments - e.g., for 1080 it's now `-gws=15360` to achieve similar behavior to what was in [#5382 (comment)](https://github.com/openwall/john/pull/5382#issuecomment-1839167284)

And it achieves exactly the same `MAX_KEYS_PER_CRYPT = 480` as the 3x higher GWS achieved before.

> and for Vega 64 it's `-gws=8000` to achieve similar to what we had in [#5382 (comment)](https://github.com/openwall/john/pull/5382#issuecomment-1839208723)

This achieves similar memory usage, but at 2x lower `MAX_KEYS_PER_CRYPT` than we had before.

solardiz:
> > for Vega 64 it's `-gws=8000` to achieve similar to what we had in [#5382 (comment)](https://github.com/openwall/john/pull/5382#issuecomment-1839208723)
> 
> This achieves similar memory usage, but at 2x lower `MAX_KEYS_PER_CRYPT` than we had before.

I think I figured this part out: the previous version's mkpc reporting was erroneous, the value was printed for potential memory usage that was then determined to be above the device's maximum allocation size, so it ended up halved, but no new mkpc was printed. The new code prints the correct value.

Still no answer on why the GWS are now smaller.

solardiz:
> Still no answer on why the GWS are now smaller.

Anyway, the current values look correct to me, and the old ones are puzzling. Maybe in the previous code version our test vectors' p=3 was somehow getting in the way even when later running on a hash with p=1. Just a guess.

solardiz (approved):
I think this is good enough to merge, although I do see many things I'd want to clean up later.

PR #5388, opened 2023-12-14
autotools: make OpenCL library loading a non-plug

ghost:
In theory, we can delete all *_plug files without loss of functionality. However, if we delete the OpenCL library loader file (opencl_dynamic_loader_plug.c) we will have compilation problems occurring in some situations. Renaming the file is desirable.

Within the newly added files, there are some issues related to `space` versus `tab` and CRLF line terminators;
- I haven't changed that.


solardiz (approved):
Makes sense to me.

PR #5389, opened 2023-12-14
autotools: stop linking against the OpenCL library

ghost:
Start using dynamic OpenCL binding. And continue building a non-OpenCL binary for users who don't have OpenCL installed.

Needs #5388


solardiz (approved):
This is as discussed/expected. Thanks!

PR #5392, opened 2023-12-19
argon2-opencl: Remove unnecessary instructions.

alainesp:
Remove unnecessary instructions. Less memory on the host and faster self-testing.

### DIFF
--- a/run/opencl/argon2_kernel.cl
+++ b/run/opencl/argon2_kernel.cl
@@ -225,19 +225,18 @@ void shuffle_block(struct block_th *block, uint thread, __local ulong *buf)
     g(block);
 
     // shuffle_shift2(block, thread, buf);
-    uint thread_src0 = apply_shuffle_shift2(thread, 0);
     thread_src1 = apply_shuffle_shift2(thread, 1);
     thread_src2 = apply_shuffle_shift2(thread, 2);
     thread_src3 = apply_shuffle_shift2(thread, 3);
     // TODO: Try to optimize 'apply_shuffle_shift2' with LUT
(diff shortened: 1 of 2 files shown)


solardiz:
In my testing, these changes appear to have caused a very slight slowdown on 1080, but I am not sure, and I did not try to figure out whether kernel or host side changes caused this. Nevertheless, I asked Alain to send us this PR, so that we know his code is fully merged and can proceed with cleanups from there.

solardiz (approved):

solardiz:
> these changes appear to have caused a very slight slowdown on 1080, but I am not sure, and I did not try to figure out whether kernel or host side changes caused this.

I am sure now, and the slowdown is caused by the removal of instructions from the kernel, and out of super's devices is only seen on the 1080. Unfortunately, we don't readily have a macro to check for Pascal (we have for Maxwell, and "Maxwell or above").

solardiz:
> Unfortunately, we don't readily have a macro to check for Pascal (we have for Maxwell, and "Maxwell or above").

This works to detect "Pascal or above":

```c
#if nvidia_sm_5plus(DEVICE_INFO) && !nvidia_sm_5x(DEVICE_INFO)
```

I don't know whether we want to be checking for that, or for "Pascal exactly" (for which we'd need more changes). Someone would need to benchmark on newer cards.

solardiz:
> we don't readily have a macro to check for Pascal

Turns out we did have it until 8fa21b5a913d9013ed5fcc45706afbd322ac6033, but anyway I managed without it for now.

PR #5394, opened 2023-12-19
Cleanup OpenCL library wrapper file generation

solardiz:
Mostly coding style fixes.


solardiz:
Editing pieces of C embedded in Python turned out to be tricky - `vim` defaults for Python were getting in the way. I had to temporarily put `let g:python_recommended_style = 0` into my `~/.vimrc` to get around this (as suggested in https://vim-jp.org/vimdoc-en/filetype.html#ft-python-plugin). As an alternative, we could consistently use `\t` in place of literal TABs when generating C code from Python. Right now (with the changes in this PR), the script has a mix of `\t` and literal TABs, using the latter in longer embedded C code blocks.

PR #5395, opened 2023-12-20
OpenCL dynamic loader: allow functions to be missing

solardiz:
This PR has additional code cleanups followed by a functional change: when we fail to lookup a function, just print a warning and set the pointer to a wrapper that returns `CL_INVALID_OPERATION`. Not ideal, but maybe right for our project.

Also skip functions with `Image` in their names. Maybe we could skip those with `Rect` too, but we do have exactly one use of one of those now.


solardiz:
I should note that this is currently untested (other than the Python script running and producing a reasonably-looking C file, which is indeed also included here). Let's see what CI says, and let's test this manually as well.

solardiz:
This built on CI without warnings and passed a few OpenCL tests, so I think is good enough. If there's anything wrong, we'll adjust separately.

PR #5396, opened 2023-12-20
Add a local copy of the OpenCL headers

alainesp:
This includes a local copy of the Khronos Group OpenCL header files and use it instead of the system ones. No modification is done to the `configure` script.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -110,3 +110,6 @@ dict/
 /.vs/JohnTheRipper/v16/.suo
 /.vs
 src/.vscode/
+/src/.vs
+/src/CppProperties.json
+/src/launch.vs.json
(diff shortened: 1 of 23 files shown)


solardiz (changes requested):
Thank you @alainesp! Please rebase this on what's currently in the repo, as I've made changes to your Python script and its generated file that you also patch here.

solardiz:
Also, I didn't check, but are the added files possibly with CRLF line endings? Can you easily avoid that?

solardiz:
Oh, this probably also needs a `NEWS` entry.

solardiz:
You got lots of changed lines in the OpenCL dynamic loader files (Python and C) now, perhaps the line endings again? Also you're reverting the Python script's file permissions to non-executable. Please avoid making these changes/reverts.

solardiz:
Also, looks like you rebased on something other than the latest. Please rebase on the latest.

solardiz:
Looks better now, but you're still reverting the Python script's file permissions to non-executable. And please add a `NEWS` item. Thank you!

solardiz (approved):
Looks good to me now, thanks!

solardiz:
Unfortunately, this PR got us some CRLF linefeeds again, which I've now fixed in a separate commit I pushed directly to bleeding-jumbo.

PR #5398, opened 2023-12-20
autotools: stop checking for OpenCL headers

ghost:
We no longer need to check for the existence of OpenCL headers.
****

The reasoning used up to the current commit is:
- if the build environment supports OpenCL, enable it.
- if the build environment does not support OpenCL, create a binary without OpenCL support.

### DIFF
--- a/src/autoconfig.h.in
+++ b/src/autoconfig.h.in
@@ -39,9 +39,6 @@
 /* Define to 1 if you have the `atoll' function. */
 #undef HAVE_ATOLL
 
-/* Define to 1 if you have the <CL/cl.h> header file. */
-#undef HAVE_CL_CL_H
-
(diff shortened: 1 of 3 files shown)


solardiz:
> The reasoning used up to the current commit is:
> 
>     * if the build environment supports OpenCL, enable it.
> 
>     * if the build environment does not support OpenCL, create a binary without OpenCL support.

As I understand, with this PR we still apply the same reasoning, but our requirement for "the build environment supports OpenCL" is reduced to being able to test-link against an OpenCL library at `configure` time.

solardiz (approved):

ghost:
> As I understand, with this PR we still apply the same reasoning, but our requirement for "the build environment supports OpenCL" is reduced to being able to test-link against an OpenCL library at configure time.

Yes. Nothing has really changed, just the `configure` checks.

PR #5399, opened 2023-12-21
OpenCL: print platform info even when no devices

ghost:
At least my cloud mac M2 machine is like this. The platform is installed but no device is available.

****

The output would be:
```
Platform #0 name: Apple, version: OpenCL 1.2 (Jun 23 2023 20:24:12)
Error: No OpenCL-capable devices were detected by the installed OpenCL driver.
```

### DIFF
--- a/src/opencl_common.c
+++ b/src/opencl_common.c
@@ -2937,6 +2937,22 @@ void opencl_list_devices(void)
 	}
 
 	if (!available_devices) {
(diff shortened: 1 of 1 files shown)


solardiz (approved):
Looks OK, except that the indentation/alignment of the continuation lines after each `clGetPlatformInfo` is weird.

Issue #5402, opened 2023-12-27
Improve argon2-opencl (auto-)tuning for high m_cost


solardiz:
The current argon2-opencl format has usability and efficiency issues:

For Argon2 hashes with somewhat high `m` (32 MiB or more) on smaller devices (8 GiB or less), it fails with default settings:

```
Cost 2 (m) is 32768 for all loaded hashes
Cost 3 (p) is 1 for all loaded hashes

Trying to use 4096 MB / 8119 MB GPU memory, status 0.
[...]
LWS=[256-1024] GWS=[8192-24576] ([8-96] blocks) => Mode: WARP_SHUFFLE
\
Trying to use 8192 MB / 8119 MB GPU memory, status 0.
0: OpenCL CL_MEM_OBJECT_ALLOCATION_FAILURE (-4) error in opencl_argon2_fmt_plug.c:332 - Error on kernel
```

In other words, `clCreateBuffer` succeeds even for a size beyond `get_global_memory_size(gpu_id)`, but then the kernel invocation fails later. (The `status 0` is my temporary addition to print `ret_code` from `clCreateBuffer`.)

For testing, I further patched it as:

```c
                memory_buffer = NULL;
                ret_code = CL_OUT_OF_RESOURCES;
                if (max_memory_size <= get_global_memory_size(gpu_id))
                        memory_buffer = clCreateBuffer(context[gpu_id], CL_MEM_READ_WRITE, max_memory_size, NULL, &ret_code);
```

but I don't know if that would catch all such cases or maybe not.

With this, things mostly work for me, but when I tried a 256 MiB hash on our 1080, I got:

```
[solar@super run]$ ./john -w=w pw-argon2-256m -dev=4 -form=opencl 
Device 4: GeForce GTX 1080
Using default input encoding: UTF-8
Loaded 1 password hash (argon2-opencl [Blake2 OpenCL])
Cost 1 (t) is 1 for all loaded hashes
Cost 2 (m) is 262144 for all loaded hashes
Cost 3 (p) is 1 for all loaded hashes
Cost 4 (type [0:Argon2d 1:Argon2i 2:Argon2id]) is 0 for all loaded hashes

Trying to use 4096 MB / 8119 MB GPU memory.
\
Trying to use 65536 MB / 8119 MB GPU memory.

Trying to use 32768 MB / 8119 MB GPU memory.

Trying to use 16384 MB / 8119 MB GPU memory.

Trying to use 8192 MB / 8119 MB GPU memory.

[...]

PR #5405, opened 2023-12-29
Further argon2-opencl GWS tuning

solardiz:
This completes #5402 and while at it improves some messages that the format prints.

There's also a global change to set `benchmark_running` before calling `reset()`, so that formats can know when they're being reset for benchmarking. In here, this is used to disregard `p > 1` seen on some test vectors when choosing optimal GWS, because the test vectors we use for benchmarking are `p = 1`.

### DIFF
--- a/run/opencl/argon2_kernel.cl
+++ b/run/opencl/argon2_kernel.cl
@@ -415,9 +415,9 @@ __kernel void KERNEL_NAME(ARGON2_TYPE)(__global struct block_g* memory, uint pas
 	/* select job's memory region: */
 	memory += (size_t)job_id * lanes * lane_blocks;
 
+#if ARGON2_TYPE == ARGON2_I || ARGON2_TYPE == ARGON2_ID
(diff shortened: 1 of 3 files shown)


solardiz:
Got this to work with the ancient Catalyst driver, on our ancient 7990, 7790, and 6770 cards. Turns out this needed a compiler memory barrier for the local memory accesses. It may in fact be our bug that we're trying without it on newer drivers/devices, and on NVIDIA - being unaware of the data dependency, the compiler is free to re-order write vs. read instructions, @alainesp. So we may need to introduce something like this unconditionally, maybe the `asm` or maybe `volatile`.

Using an actual local memory barrier also works, but is significantly slower.

solardiz:
> So we may need to introduce something like this unconditionally, maybe the `asm` or maybe `volatile`.

The not-so-ancient AMD driver on "super" doesn't accept `asm` (nor `__asm__`). Luckily, with that driver this kernel works without that.

PR #5407, opened 2023-12-30
CI: Add (revised) GitHub Actions from passwdqc

solardiz:
This is untested. Theoretically, it should give us many build tests via GitHub Actions. It should also fail the trailing whitespace test badly, because we have lots of that.

### DIFF
--- a/.ci/install-dependencies.sh
+++ b/.ci/install-dependencies.sh
@@ -1,0 +1,86 @@
+#!/bin/sh -ex
+#
+# Copyright (c) 2018-2020 The strace developers.
+# All rights reserved.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+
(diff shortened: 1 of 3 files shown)


solardiz:
> Theoretically, it should give us many build tests via GitHub Actions. It should also fail the trailing whitespace test badly, because we have lots of that.

In my fork, those many builds all succeeded, and the whitespace test failed as expected. I'll need to add whitespace fixes before we merge this PR, after which point I guess these extra tests will also run in here.

solardiz:
> the whitespace test failed as expected. I'll need to add whitespace fixes before we merge this PR

Looking at those thousands of errors, this felt unrealistic for now. Some of them are kind of on purpose. So I commented out this test for now, even though I really would like to have it for new contributions.

PR #5416, opened 2024-01-05
CI: add a build for ARM architecture

ghost:
Remove some unnecessary tasks and add a build for ARM.

****

Using CircleCI SSH we will be able to reproduce and debug the SunMD5 issue.
```
Testing: SunMD5 [MD5 128/128 ASIMD 4x2]... (2xOMP) *** stack smashing detected ***: terminated
.circleci/circle-ci.sh: line 102:  9032 Aborted                 (core dumped) $JTR --test=0
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,7 +33,7 @@ jobs:
   no-plugs:
     <<: *common_linux
     docker:
-      - image: claudioandre/john:opencl18
+      - image: claudioandre/john:opencl
         environment:
           OMP_NUM_THREADS: 2
           PLUGS: "none"
@@ -43,7 +43,7 @@ jobs:
   encoding-cpu:
     <<: *common_linux
     docker:
-      - image: ghcr.io/openwall/john-ci:fedora.32
+      - image: ghcr.io/claudioandre-br/john-ci:fedora.latest
         environment:
           OMP_NUM_THREADS: 2
           ENCODING_TEST: "cpu"
@@ -59,7 +59,7 @@ jobs:
   asan:
     <<: *common_linux
     docker:
-      - image: ghcr.io/openwall/john-ci:fedora.32
+      - image: ghcr.io/claudioandre-br/john-ci:fedora.latest
         environment:
           OMP_NUM_THREADS: 2
           ASAN_TEST: "yes"
@@ -68,7 +68,7 @@ jobs:
   test-full:
     <<: *common_linux
     docker:
-      - image: ghcr.io/openwall/john-ci:fedora.32
+      - image: ghcr.io/claudioandre-br/john-ci:fedora.latest
         environment:
           ASAN_TEST: "yes"
           FULL_TEST: "yes"
@@ -77,7 +77,7 @@ jobs:
   fast-formats-omp:
     <<: *common_linux
     docker:
-      - image: ghcr.io/openwall/john-ci:fedora.32
+      - image: ghcr.io/claudioandre-br/john-ci:fedora.latest
         environment:
           OMP_NUM_THREADS: 2
           ASAN_TEST: "yes"
@@ -87,7 +87,7 @@ jobs:
   fast-formats-omp-x-non-SIMD:
     <<: *common_linux
     docker:
-      - image: ghcr.io/openwall/john-ci:fedora.32
+      - image: ghcr.io/claudioandre-br/john-ci:fedora.latest
         environment:
           OMP_NUM_THREADS: 2
           ASAN_TEST: "yes"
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -11,7 +11,6 @@ jobs:
       - run:
           name: Testing JtR
           command: .circleci/circle-ci.sh TEST
-          no_output_timeout: 20m
```
ghost, line 14: We (probably) don't need this.

solardiz:
> Using CircleCI SSH we will be able to reproduce and debug the SunMD5 issue.
> 
> ```
> Testing: SunMD5 [MD5 128/128 ASIMD 4x2]... (2xOMP) *** stack smashing detected ***: terminated
> .circleci/circle-ci.sh: line 102:  9032 Aborted                 (core dumped) $JTR --test=0
> ```

This actually failed just like that. So I guess this PR is not intended for merging yet? @claudioandre-br Are you going to debug and fix this issue now? Or on the contrary, exclude SunMD5 from this test for now? Thank you!

solardiz:
GitHub Actions is going to also have Arm support soon: https://github.blog/changelog/2023-10-30-accelerate-your-ci-cd-with-arm-based-hosted-runners-in-github-actions/ "will be entering private beta in January 2024."

ghost:
> So I guess this PR is not intended for merging yet?

It's easy to reproduce the error and get information about it, but isn't it a bug in the ARM omp implementation itself? I'll try disabling simd just to see what happens. but the `for` itself looks good.

If so, it is beyond our hands.

solardiz:
> isn't it a bug in the ARM omp implementation itself?

I think we don't know.

solardiz:
> exclude SunMD5 from this test for now

Maybe we need to do that, so that we test the rest of the formats with Arm ASIMD, OpenMP, and ASan? We haven't run such tests beyond this format, have we?

ghost:
> > exclude SunMD5 from this test for now
> 
> Maybe we need to do that, so that we test the rest of the formats with Arm ASIMD, OpenMP, and ASan? We haven't run such tests beyond this format, have we?

All other formats (that are not disabled because of something else [1]) are tested in Mac and by this PR.

[1] more or less 20 are deactivated for macOS.
```
[...]
Testing: dynamic_2014 [md5($s.md5($p).$s) (PW > 55 or salt > 11 bytes) 128/128 ASIMD 4x2]... PASS
Testing: dummy [N/A]... PASS
Testing: crypt, generic crypt(3) [?/64]... (4xOMP) PASS
All 407 formats passed self-tests!
 Ok: -test-full=0 --format=cpu (macOS ASIMD)
```

solardiz:
> All other formats (that are not disabled because of something else [1]) are tested in Mac and by this PR.

But that's without ASan, right? So maybe we need to exclude SunMD5 from the test with ASan, to let it proceed to test further formats.

> [1] more or less 20 are deactivated for macOS.

That's a lot. What are they, and where are they deactivated? Perhaps try re-enabling them?

ghost:
> > All other formats (that are not disabled because of something else [1]) are tested in Mac and by this PR.
> 
> But that's without ASan, right? So maybe we need to exclude SunMD5 from the test with ASan, to let it proceed to test further formats.

In fact, since we don't have much hardware, _**there is no**_ CI with ASAN in non-X86 environments. You're thinking of ASAN for some reason, but it's safe to assume that it's disabled (I mean, not tested) on anything other than Linux X86_64.


> > [1] more or less 20 are deactivated for macOS.
> 
> That's a lot. What are they, and where are they deactivated? Perhaps try re-enabling them?

I disable formats when they fail and I know no one will fix them. For example:
```sh
echo 'raw-BLAKE2 = Y' >> ../run/john-local.conf  #BLAKE2
echo 'timeroast = Y' >> ../run/john-local.conf # fails on BE

# Only viable if we use our own NVIDIA hardware;
echo 'pgpdisk-opencl = Y' >> ../run/john-local.conf #FAILED (cmp_all(49)) Intel OpenCL CPU
echo 'bitlocker-opencl = Y' >> ../run/john-local.conf # Very slow format
echo 'argon2-opencl = Y' >> ../run/john-local.conf # Very slow format
```

Everything disabled is documented (disabling is documenting the format fails for me). Since I know it fails, I stop testing it. But, e.g., I will re-enable the recently fixed VMX format to check.

ghost:
For example, the CI that runs in this repository does not have any CPU formats disabled, but it is easy to see that we are not testing (on purpose) many OpenCL formats. At:
* https://github.com/openwall/john/blob/bleeding-jumbo/.ci/disable_formats.sh

ghost:
I disabled the `sunmd5` testing. So, CI should pass now.

solardiz (approved):

solardiz:
Somehow `ci/circleci: non-SIMD` doesn't run. @claudioandre-br Would you like me to merge this PR now?

ghost:
I removed it (because GitHub Actions has one identical build job).

PR #5422, opened 2024-01-09
Armory wallet support

solardiz:
This adds Armory wallet support. The input is btcrecover data extracts. Thanks to @gurnec of btcrecover for providing 2 of the test vectors.

This does not add an extract script of our own, and a self-contained one can't easily be created - btcrecover's uses code from Armory itself, which is expected to be separately installed on the system. But at least the JtR format is kind of self-contained (only uses what we already had in the tree), which I think is the first implementation for this wallet type that does not rely on the wallet software.

Not implemented yet:
1. Tunable costs reporting.
2. SIMD.
3. Interleaving.
4. Prefetch (only possible with SIMD or/and interleaving).
5. OpenCL format.
6. Own `*2john` script.

5 and 6 are not currently planned, 1 to 4 are.


solardiz:
> 1. Tunable costs reporting.

Added. The rest of the to-do items I listed are beyond scope of this PR.

solardiz:
> [Armory format: Support only lookup table sizes that are powers of 2](https://github.com/openwall/john/pull/5422/commits/0bdfbf09aef78ca7fb3b8ea442a2598a530dc7b0)

This is per my review of `BitcoinArmory/cppForSwig/EncryptionUtils.cpp: KdfRomix::computeKdfParams()`. It looks unable to choose anything lower than 1024 bytes or not a power of 2.

solardiz:
Added 2 more test vectors kindly provided by @gurnec, also standardizing on 8 MiB, 2 iterations for our benchmarks. These appear to be somewhat-likely settings that Armory would use by default given CPU speeds back when it was popular.

PR #5424, opened 2024-01-12
Armory format: Pre-check for cracks within parallel section

solardiz:
This is a controversial change for a slow format like this, but OTOH I don't want my decision not to make this format salt-only (where we'd have to check for cracks in `crypt_all`) to have any obvious performance cost, whereas the number of comparisons per `crypt_all` can be significant when there are many threads and SIMD vectors are wide (e.g., `192*8` in my recent test means 1536 comparisons, which when done sequentially out of `crypt_all` can have a measurable cost).

### DIFF
--- a/src/armory_fmt_plug.c
+++ b/src/armory_fmt_plug.c
@@ -27,6 +27,7 @@ john_register_one(&fmt_armory);
 
 #include "base64_convert.h"
 #include "formats.h"
+#include "loader.h"
 #include "memory.h"
 #include "crc32.h"
 #include "sha2.h"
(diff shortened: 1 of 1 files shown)


solardiz:
Weird, this segfaulted on `ci/circleci: arm`, whereas the previous revision passed tests there. I wonder if I introduced or exposed a bug here, or if it's something intermittent - e.g., maybe we simply don't always have the same amount of RAM available there?

solardiz:
> I wonder if I introduced or exposed a bug here

Indeed, I did. The`salt` argument to `crypt_all` is allowed to be NULL, which I didn't handle. I could also reproduce the crash on Solaris/sparc32. Now I fixed that. However, I'm puzzled why this didn't show up on x86(-64) - is our self-test/benchmark on x86(-64) different, not ever setting `salt = NULL` there?

PR #5425, opened 2024-01-14
Specialize SIMD SHA-512 for half-length input, iterated hashing

solardiz:
This introduces specialized versions of SIMDSHA512body() and makes use of them in a few formats, including in the recently added Armory, where it provides a speedup of 4% on i7-4770K and some smaller speedups on some other machines I tested.

There are 3 specialized functions now, but I mostly care about the case of `SSEi_HALF_IN|SSEi_OUTPUT_AS_INP_FMT` (as that's what's used in some iterated formats now) and the original case from prior to my changes here (not to introduce performance regressions via increased code size). The third case of having `SSEi_HALF_IN` along with other or no flags is specialized merely to avoid code size increase in the original case. We could alternatively unsupport `SSEi_HALF_IN` except in `SSEi_HALF_IN|SSEi_OUTPUT_AS_INP_FMT`, but for now I opted to have it supported separately as well.

Further optimization potential is doing similar for SHA-256 (should be easy now) and maybe for others (it'd be different there), and introducing function specializations also for the previously existing flags (to reduce code size in the functions that many/most formats use).

### DIFF
--- a/run/relbench
+++ b/run/relbench
@@ -42,19 +42,18 @@ sub parse
 		return;
 	}
 
-	if (/^(All|\d+ out of) \d+ (tests|formats)( have)? (passed self-tests|FAILED)/) {
-		($total) = /^All (\d+) (tests|formats)( have)? passed self-tests/;
-		if (!defined($total)) {
+	if (/^(|All |\d+ out of )\d+ (tests|formats)( have)? (benchmarked|passed self-tests|FAILED)/) {
+		($total) = /^(|All )(\d+) (tests|formats)( have)? (benchmarked|passed self-tests)/;
+		if (defined($total)) {
+			$failed = 0;
+		} else {
 			($failed, $total) =
 			    /^(\d+) out of (\d+) (tests|formats)( have)? FAILED/;
 		}
 		return;
 	}
-	return if /Benchmarking: .*\.\.\. FAILED/;
+	return if /Benchmarking: .*\.\.\. FAILED|^Speed for cost |^Warning: /;
 
-	if (/^Speed for cost /) {
-		return;
-	}
 	my $ok = 0;
 	if (defined($name)) {
 		($kind, $real, $reals, $virtual, $virtuals) =
@@ -71,9 +70,12 @@ sub parse
 			$id = $name . ':' . $kind;
 			$real *= 1000 if ($reals eq 'K');
 			$real *= 1000000 if ($reals eq 'M');
+			$real *= 1000000000 if ($reals eq 'G');
+			$real = 0.001 if ($real eq "0.0" || $real eq "0.00");
 			$virtual *= 1000 if ($virtuals eq 'K');
 			$virtual *= 1000000 if ($virtuals eq 'M');
-			$virtual = 0.001 if $virtual eq "0.0";
+			$virtual *= 1000000000 if ($virtuals eq 'G');
+			$virtual = 0.001 if ($virtual eq "0.0" || $virtual eq "0.00");
 			return;
 		}
 	} else {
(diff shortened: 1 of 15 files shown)


solardiz:
The [SIMDSHA512body(): Build 3 specialized versions of the function](https://github.com/openwall/john/pull/5425/commits/d9ced9865438c67082f3d84b3a305a69db7b5950) commit here reduces the need to have manually optimized the `STEP` macro in the first commit here, if we also add a `memset(&w[k][9], 0, 6 * sizeof(vtype));`, as that lets the compiler figure out that certain values are always zero in certain builds of the function.

solardiz:
> [SIMDSHA512body() and calls: Fix SIMD_PARA_SHA512 > 1 support](https://github.com/openwall/john/pull/5425/commits/3e893337c6da3955136716ab87f518103f1751bd)

Testing `SIMD_PARA_SHA512 = 2` after this commit, everything works on up to AVX2, including all of our GitHub Actions (https://github.com/solardiz/john/actions/runs/7534465934), however in my own testing on AVX-512 there are failures:

```
$ OMP_NUM_THREADS=1 ../run/john -test -form=office
Benchmarking: Office, 2007/2010/2013 [SHA1 512/512 AVX512BW 16x / SHA512 512/512 AVX512BW 8x2 AES]... FAILED (get_key(3))
$ OMP_NUM_THREADS=2 ../run/john -test -form=office
Will run 2 OpenMP threads
Benchmarking: Office, 2007/2010/2013 [SHA1 512/512 AVX512BW 16x / SHA512 512/512 AVX512BW 8x2 AES]... (2xOMP) FAILED (cmp_all(6))
$ OMP_NUM_THREADS=3 ../run/john -test -form=office
Will run 3 OpenMP threads
Benchmarking: Office, 2007/2010/2013 [SHA1 512/512 AVX512BW 16x / SHA512 512/512 AVX512BW 8x2 AES]... (3xOMP) FAILED (get_key(87))
$ OMP_NUM_THREADS=4 ../run/john -test -form=office
Will run 4 OpenMP threads
Benchmarking: Office, 2007/2010/2013 [SHA1 512/512 AVX512BW 16x / SHA512 512/512 AVX512BW 8x2 AES]... (4xOMP) FAILED (get_key(87))
```

The above is in a build with ASan. Exact same test errors without ASan. And exact same faulty indices remain across runs.

3 other formats fail tests as well, I haven't looked at them yet.

solardiz:
> 3 other formats fail tests as well, I haven't looked at them yet.

They are:

```
Testing: Blackberry-ES10 (101x) [SHA-512 512/512 AVX512BW 8x2]... (4xOMP) FAILED (cmp_all(2))
Testing: eCryptfs (65536 iterations) [SHA512 512/512 AVX512BW 8x2]... (4xOMP) FAILED (cmp_all(2))
Testing: Drupal7, $S$ (x16385) [SHA512 512/512 AVX512BW 8x2]... (4xOMP) FAILED (cmp_all(2))
```

Out of these, Blackberry-ES10 and eCryptfs are affected by changes in this PR (and got good speedup when they work), but Drupal7 shouldn't be. I have no idea whether they worked with `SIMD_PARA_SHA512 = 2` on AVX-512 before this PR; we don't normally set this on any arch currently.

solardiz:
> Testing `SIMD_PARA_SHA512 = 2` [...] on AVX-512 there are failures:

Apparently, the AVX-512 specific portion of [SIMDSHA512body(): Optimize SSEi_FLAT_OUT](https://github.com/openwall/john/pull/5425/commits/f64a8b93f966d0116309753650d5121a29125ade) here is buggy for that case. Reverting this commit makes these 4 formats pass the same tests. I'll review and revise it.

Issue #5431, opened 2024-01-30
mozilla2john errors with python 3.11.6


Munto:
mozilla2john is incompatible with python 3.11.6.
Here is the error I got:
```
Traceback (most recent call last):
  File "/usr/bin/mozilla2john", line 73, in <module>
    process_file(sys.argv[i])
  File "/usr/bin/mozilla2john", line 33, in process_file
    idx = data.find("global-salt")
          ^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: argument should be integer or bytes-like object, not 'str'
```
I had to convert string to bytestring in the following lines to make it works:
https://github.com/openwall/john/blob/bleeding-jumbo/run/mozilla2john.py#L33
https://github.com/openwall/john/blob/bleeding-jumbo/run/mozilla2john.py#L43
https://github.com/openwall/john/blob/bleeding-jumbo/run/mozilla2john.py#L58

I don't know if you need to maintain a compatibility with previous python version so I'm just opening an issue.
I use Arch, btw.

solardiz:
Thank you. Confirmed. So basically, we need this? -

```diff
+++ b/run/mozilla2john.py
@@ -30,7 +30,7 @@ def process_file(filename):
 
     data = open(filename, "rb").read()
 
-    idx = data.find("global-salt")
+    idx = data.find(b"global-salt")
     if idx < 0:
         fail(filename)
         return
@@ -40,7 +40,7 @@ def process_file(filename):
         return
     global_salt = data[idx-20:][0:20]  # is the salt always 20 bytes?
 
-    idx = data.find("password-check")
+    idx = data.find(b"password-check")
     if idx < 0:
         fail(filename)
         return
@@ -55,7 +55,7 @@ def process_file(filename):
     version = 3  # fake (this should be improved)
     nnLen = 1  # fake
     oidLen = 11  # fake
-    oidData = "\x00" * oidLen  # fake
+    oidData = b"\x00" * oidLen  # fake
     sys.stdout.write("%s:$mozilla$*%s*%s*%s*" % (os.path.basename(filename), version, len(entry_salt), nnLen))
     sys.stdout.write(hexlify(entry_salt).decode("ascii"))
     sys.stdout.write("*%s*%s*%s*" % (oidLen, hexlify(oidData).decode("ascii"), len(verifier)))
```

> I don't know if you need to maintain a compatibility with previous python version

[...]

PR #5434, opened 2024-02-02
Fix TypeError with python 3.11+ for mozilla2john #5431

Munto:


solardiz (approved):

solardiz:
Oh, I should have suggested a specific commit message. I think this affects Python 3 earlier than 3.11 as well, and we generally prefix commit titles by subsystem names (so `mozilla2john.py: ...`). Nevermind, I'll merge this anyway.

PR #5441, opened 2024-02-25
Update INSTALL-WINDOWS

David-Domenech-Pereira:
I updated it to work in a 2024 machine with the last release of CygWin

### DIFF
--- a/doc/INSTALL-WINDOWS
+++ b/doc/INSTALL-WINDOWS
@@ -16,11 +16,12 @@ system.
 
     - Open the Command Prompt in Windows (see [1])
 
-    - Install required CygWin packages.
-      - C:\cygwin64\setup-x86_64.exe -q -P gcc-core -P libgcc1 -P make -P perl
-      - C:\cygwin64\setup-x86_64.exe -q -P libssl-devel -P libbz2-devel
-      - C:\cygwin64\setup-x86_64.exe -q -P libgmp-devel -P zlib-devel
-      - C:\cygwin64\setup-x86_64.exe -q -P libOpenCL-devel -P libcrypt-devel
+    - Install required CygWin packages from where you have the CygWin setup:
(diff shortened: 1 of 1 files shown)


doc/INSTALL-WINDOWS
```diff
@@ -16,11 +16,12 @@ system.
 
     - Open the Command Prompt in Windows (see [1])
 
-    - Install required CygWin packages.
-      - C:\cygwin64\setup-x86_64.exe -q -P gcc-core -P libgcc1 -P make -P perl
-      - C:\cygwin64\setup-x86_64.exe -q -P libssl-devel -P libbz2-devel
-      - C:\cygwin64\setup-x86_64.exe -q -P libgmp-devel -P zlib-devel
-      - C:\cygwin64\setup-x86_64.exe -q -P libOpenCL-devel -P libcrypt-devel
+    - Install required CygWin packages from where you have the CygWin setup:
+      - cd C:\Users\Name\Downloads...
```
solardiz, line 20: I'm unhappy about suggesting that people run the setup program from the downloads directory, because this is vulnerable to past drive-by DLL download attacks.

doc/INSTALL-WINDOWS
```diff
@@ -32,23 +33,24 @@ system.
     Unzip it to your preferred directory
 
 ==== Build
+    - Copy your Jonh the Ripper source code inside C:\cygwin64
```
solardiz, line 36: Typo in `Jonh`, and would need to add empty line before this one. Also, this is redundant with "Go to the folder you unzipped John the Ripper source code." that follows - so either don't add this or make more edits to avoid the redundancy.

solardiz (commented):
Thanks. I made some comments.

Ideally, any changes to this PR would be made by amending the one commit and force-pushing, but I understand that if you only use GitHub web UI (rather than use `git` directly), you might not be able to.

PR #5447, opened 2024-03-09
Workaround for broken CircleCI Ubuntu VM image

ghost:
Since the base image is out of date, we need to do something to fix it (or the build).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -97,12 +97,16 @@ jobs:
   arm:
     machine:
       image: ubuntu-2204:current
+      environment:
+        export DEBIAN_FRONTEND="noninteractive"
     resource_class: arm.medium
     steps:
       - checkout
       - run:
           name: Preparing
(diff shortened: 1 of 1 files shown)


solardiz (approved):
Thank you, @claudioandre-br! This seems to have worked, although I am puzzled by the commented out `sudo apt ...` line - didn't we need to install those packages?

ghost:
We will need to reverse this line later. Now some post-installation script is presenting an "OK" window (probably because of the OpenSSL update).

solardiz:
How does the build succeed with the `apt` line commented out at all? `libssl-dev` is a required dependency unless we use `--without-openssl`. Is this package somehow installed anyway?

Anyhow, I don't mind just merging this without knowing more, since it's just CI and now it works.

ghost:
> Is this package somehow installed anyway?

Yes, there is an old version installed.

PR #5449, opened 2024-03-12
Migrate `pgpwde2john` to Python3 and make it work with John again

iyesin:


solardiz:
Thanks. After these changes, does the script also still work with Python 2? (Ideally, it would.)

solardiz:
We don't seem to have any sample inputs to test this script with. Can you please add some (test only, suitable for publication) via a pull request to https://github.com/openwall/john-samples ?

iyesin:
Hey @solardiz,

Thanks for a review.
With a change I made in a first line this script, unfortunately, it wouldn't work with python2.
From my point of view it pretty uncomfortable and actually unneeded to fiddle around with two branches of same language in one script.
Also, it could be a good staring point for migration of outdated code to new runtime (python.org does not even mention python2 on downloads page, btw).

Regarding test sample: I really not sure if I would be able to craft an image. Again, it old technology and it needs some experimentation.

solardiz:
I wonder if @kholia sees this and maybe still has a test image we could use for retesting?

> With a change I made in a first line this script, unfortunately, it wouldn't work with python2.

Do you mean the shebang line, which you then reverted in the second commit here?

iyesin:
Hey @solardiz,

> I wonder if @kholia sees this and maybe still has a test image we could use for retesting?

I managed to find winxp_x32 backup image with pgp_desktop 9 installed. Let's see if I can make it work under QEMU. 

> Do you mean the shebang line, which you then reverted in the second commit here?

Oh, right. I reverted it. Then, it needs good testing before merging.

solardiz:
> I managed to find winxp_x32 backup image with pgp_desktop 9 installed. Let's see if I can make it work under QEMU.

That's great news. Thank you very much!

iyesin:
@solardiz I uploaded pgpwde sample image in https://github.com/openwall/john-samples/pull/23. Please, validate.

iyesin:
@solardiz I found a way to test my changes with python2 and it does not work. Do you think it worth trying to make it work with both python versions?

solardiz:
> Do you think it worth trying to make it work with both python versions?

If can be done easily, yes. If not, then we need to update the shebang line to say `python3`.

You may want to grep our other scripts for `PY3` (uppercase) to see how we handle Python version checks/compatibility.

solardiz:
> Do you think it worth trying to make it work with both python versions?

I took care of this in my copy now, will commit shortly.

PR #5452, opened 2024-03-20
CI: use default version of gcc on Ubuntu VM

ghost:
asan is failing a lot (and seems random). 

Well, it is probably not a good idea to use gcc-13 + ASAN on Ubuntu-latest (the default gcc is 11).
****

It looks better now, using the PR, but we can't be sure without more random testing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -98,7 +98,7 @@ jobs:
     machine:
       image: ubuntu-2204:current
       environment:
-        export DEBIAN_FRONTEND="noninteractive"
+        DEBIAN_FRONTEND: "noninteractive"
     resource_class: arm.medium
     steps:
       - checkout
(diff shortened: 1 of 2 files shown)


ghost:
CI is clean. Forcing another run for testing.

solardiz (approved):

solardiz:
> asan is failing a lot (and seems random).

My guess was that we simply exceeded some kind of GitHub Actions resources for our org, with my recent multiple quick pushes to bleeding-jumbo, which resulted in several sets of CI jobs running concurrently. I've since rerun the failed jobs on a couple of commits, and they all succeeded.

I think it made sense for us to use newer gcc's ASan, but I also don't mind your changes here. So I'll merge.

ghost:
> My guess was that we simply exceeded some kind of GitHub Actions resources for our org,

I saw this ASAN `AddressSanitizer:DEADLYSIGNAL` error a week ago [1] (and nothing was running). But since it's random, I can't say much about it. 

The org in the situation was `me`.

[1] https://github.com/claudioandre-br/JohnTheRipper/actions/runs/8248714971 versus, e.g., https://github.com/openwall/john/actions/runs/8351346753/job/22859525043

PR #5457, opened 2024-03-31
build-info: add Cygwin version information

ghost:
The recent version of Cygwin removed support for many old Windows versions/platforms. Adding this information to the output can help us detect if end users are complaining about unsupported things.

```
Version: 1.9.0-jumbo-1+bleeding-03c6e07808 2024-03-31 12:50:58 -0300
Build: cygwin 64-bit x86_64 AVX2 AC OMP OPENCL
[...]
gcc version: 7.4.0
OpenCL headers version: 1.2
Crypto library: OpenSSL
OpenSSL library version: 0300000d0
OpenSSL 3.0.13 30 Jan 2024
GMP library version: 6.3.0	(loaded: 6.1.2)
[...]
Parsed terminal locale: UTF-8
Cygwin version: 3.1.2-340.x86_64, 2019-12-21 15:25 UTC

```

```
Version: 1.9.0-jumbo-1+bleeding-03c6e07808 2024-03-31 12:50:58 -0300
[...]
gcc version: 11.4.0
Crypto library: OpenSSL
OpenSSL library version: 0300000d0
OpenSSL 3.0.13 30 Jan 2024
GMP library version: 6.3.0
[...]
Parsed terminal locale: UTF-8
Cygwin version: 3.5.1-1.x86_64, 2024-02-27 11:54 UTC
```

### DIFF
--- a/src/listconf.c
+++ b/src/listconf.c
@@ -100,6 +100,10 @@ extern char CPU_req_name[];
 
 #define SINGLE_MAX_WORDS(len) MIN(SINGLE_IDX_MAX, SINGLE_BUF_MAX / len + 1)
 
+#ifdef __CYGWIN__
+#include <sys/utsname.h>
+#endif
+
 /*
  * FIXME: Should all the listconf_list_*() functions get an additional stream
  * parameter, so that they can write to stderr instead of stdout in case fo an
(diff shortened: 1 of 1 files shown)


src/listconf.c
```diff
@@ -345,6 +349,18 @@ static void listconf_list_build_info(void)
 	printf("Terminal locale string: %s\n", john_terminal_locale);
 	printf("Parsed terminal locale: %s\n", cp_id2name(options.terminal_enc));
 
+#if defined (__CYGWIN__)
+	{
+		struct utsname buffer;
+
+		errno = 0;
+		if (uname(&buffer) != 0)
+			printf("Cygwin version detection error");
```
solardiz: Add `\n` or use `puts`.

Also, why reset `errno` if we don't check it later? Even if we did use `errno`, per POSIX it should be safe to do so after `uname` returns failure - but success is indicated by `a non-negative value`, not just 0. Perhaps Cygwin has its own documentation making this more specific.

POSIX:

```
RETURN VALUE
       Upon  successful completion, a non-negative value shall be returned.
       Otherwise, -1 shall be returned and errno set to indicate the error.
```

solardiz (changes requested):
Thanks. I added a review comment. Also, I'd simply use `#ifdef` (in two places).

solardiz (approved):

PR #5478, opened 2024-05-15
python: recommend pycryptodome package over PyCrypto

exploide:
The PyCrypto package is old and deprecated, pycryptodome works as a drop-in replacement with the same API.


solardiz:
Thanks. I see `DPAPImk2john.py` is currently claimed to be both Python 2 and 3 compatible, which is great. I wonder if switching it to the new module dependency possibly excludes Python 2 support - in other words, is there pycryptodome for Python 2?

exploide:
Valid point. I just checked and it is available for Python 2, too. So no problem here.

PR #5479, opened 2024-05-17
Assorted fixes 2024/05/17

solardiz:
This reimplements @AlbertVeli's logical CPU count detection in `configure` and fixes 3 assorted bugs.


solardiz:
> This reimplements @AlbertVeli's logical CPU count detection in `configure`

Now revised after testing on a few shells (sh, ksh, zsh) on Solaris and AIX. Need to run the maybe-missing command in a subshell to be able to hide the (sub)shell's error message about it.

PR #5480, opened 2024-05-18
External modes test and minor optimization

solardiz:
This tests almost all external modes on `make check`, which we run on CI.

This also adds proper compile-time handling of negative integer constants, which previously emitted the unary `-` instruction. Unfortunately, the code for this is a hack. It was cleaner not to handle this case specially.

My additional test case (not included here) for this optimization is:

```
[List.External:Test]
void generate()
{
	if (word[2] == 666) {
		word = 0;
		return;
	}
	word[0] =-1 + 'a' + 1 - -111 -111 - 222 + 222 -111-111+222;
	word[0] -=-22;
	word[0] -= 22;
	word[0]-=-22;
	word[0]-=22;
	word[1] = -1 + 1;
	word[2] = 666;
}
```

This has a mix of `-` in different meanings, which the compiler has to distinguish properly. The expected output is `a`.

### DIFF
--- a/src/Makefile.in
+++ b/src/Makefile.in
@@ -758,6 +758,9 @@ check: default
 	$(MAKE) unit-tests
 	../run/john --test=0 --verbosity=2 --format=cpu
 	../run/john --test=0 --verbosity=2 --format=+dynamic,all
+	if [ -x /usr/bin/md5sum ]; then \
+		JOHN=../run/john tests/test_externals.sh | md5sum -c tests/test_externals.md5; \
+	fi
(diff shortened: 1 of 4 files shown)


solardiz:
> This also adds proper compile-time handling of negative integer constants, which previously emitted the unary `-` instruction. Unfortunately, the code for this is a hack. It was cleaner not to handle this case specially.

We should be able to drop this hack later with no performance loss if we instead implement generic compile-time evaluation of constant subexpressions.

PR #5490, opened 2024-05-30
New version of Radius2john.py

k4amos:
`radius2john.py` didn't seem to be able to work properly : 

- The attack method is described  [here](http://www.untruth.org/~josh/security/radius/radius-auth.html). The previous version of `radius2john.py` implemented the _3.1 Response Authenticator Based Shared Secret Attack_. The attacker can pre-compute the MD5 state for (Code+ID+Length+RequestAuth+Attributes). However, the value of the attributes was forgotten in the code, the calculated hash could not be broken if there were attributes.
- The _3.3 User-Password Attribute Based Shared Secret Attack_ was not implemented

For these two reasons, I propose a new version of radius2john.py much closer to the perl version _radius2john.pl_

image

### DIFF
--- a/run/radius2john.py
+++ b/run/radius2john.py
@@ -24,95 +24,92 @@
 # "3.1 Response Authenticator Based Shared Secret Attack"
 
 # For attack 3.3 :
-# we try authentications using a known password, and sniff the radius packets to a pcpap file.
+# We try authentications using a known password, and sniff the radius packets to a pcap file.
 # This script reads access-request in the pcap file, and dumps the md5(RA+secret) and RA, in a john-friendly format.
-# The password must be always the same, be less then 16 bytes long, and entered in the $PASSWORD variable below.
-# The user names used during this attack must be entered in @LOGINS below.
 
 # For attack 3.1:
-# we don't need to try authentications. Just sniff the radius packets in a pcap file.
+# We don't need to try authentications. Just sniff the radius packets in a pcap file.
 # This script reads the pcap file, matches radius responses with the corresponding all_requests,
 # and dumps md5 and salt as needed.
 
-import scapy.all as scapy
 import binascii
 import sys
+import argparse
 
+try:
+    import scapy.all as scapy
+except ImportError:
+    print(
+        "Scapy seems to be missing, run 'pip install --user scapy' to install it"
+    )
+    sys.exit(1)
 
-# Global variables
-PASSWORD = b"1"  # The user password used for the 3.3 attack
-LOGINS = ["max"]  # The user logins used for the 3.3 attack
-UNIQUE = 1  # Set to 0 to disable unicity of client IPs in the output file
-VALID_LOGIN = {login: True for login in LOGINS}
-
-all_requests = {}
-dumped_ips = {}
-
-
-def read_file(filename):
+def read_file(args, filename):
     packets = scapy.rdpcap(filename)
     for packet in packets:
-        process_packet(packet)
+        process_packet(args, packet)
 
 
-def process_packet(packet):
+def process_packet(args, packet):
     if packet.haslayer(scapy.IP) and packet.haslayer(scapy.UDP):
         ip_layer = packet[scapy.IP]
         udp_layer = packet[scapy.UDP]
 
         if udp_layer.dport in [1812, 1813] or udp_layer.sport in [1812, 1813]: 
-            radius_data = bytes(udp_layer.payload)
-            process_radius(ip_layer, radius_data, bytes(udp_layer.payload))
+            process_radius(args, ip_layer, bytes(udp_layer.payload))
 
 
-def process_radius(ip, radius_data, udpdata):
-    radius_packet = scapy.Radius(radius_data)
+def process_radius(args, ip, udpdata):
+    radius_packet = scapy.Radius(udpdata)
 
     if radius_packet.code in [1, 4]:  # Access-Request, Accounting-Request
 
-        user_name, user_password = None, None
-        for _ in range(len(radius_packet.attributes)):
-            if radius_packet.attributes[_].name == "User-Name":
-                user_name = radius_packet.attributes[_].value.decode("utf-8")
+        if args['password'] is not None:
 
-            if radius_packet.attributes[_].name == "User-Password":
-                user_password = radius_packet.attributes[_].value.decode("utf-8")
+            user_name, user_hash = None, None
 
-        if user_name in VALID_LOGIN and user_password:
-            dump_access_request(
-                ip.src, user_name, radius_packet.authenticator, user_password
-            )
+            for attr in radius_packet.attributes:
+                if attr.name == "User-Name":
+                    user_name = attr.value.decode("utf-8")
+
+                if attr.name == "User-Password":
+                    user_hash = attr.value
+
+            if user_hash is not None:
+                if (args['login'] is not None and user_name == args['login']) or args['login'] is None:
+                    dump_access_request(
+                        args, ip.src, radius_packet.authenticator, user_hash
+                    )
 
         all_requests[f"{ip.src}-{radius_packet.id}"] = radius_packet.authenticator
 
     elif radius_packet.code in [2, 11, 3, 5]:  # Access-Accept, Access-Challenge, Access-Reject, Accounting-Response
         key = f"{ip.dst}-{radius_packet.id}"
         if key in all_requests:
-            dump_response(ip.dst, all_requests[key], radius_packet, udpdata)
+            dump_response(args, ip.dst, all_requests[key], radius_packet.authenticator, udpdata)
 
 
-def dump_response(ip, req_ra, radius_packet, udpdata):  # 3.1 attack
-    if UNIQUE and ip in dumped_ips:
+def dump_response(args, ip, req_ra, ra, udpdata):  # 3.1 attack
+    if args["single"] and ip in dumped_ips:
         return
-
-    hash_val = radius_packet.authenticator
 
     salt = bytearray(udpdata)
     salt[4:20] = req_ra  # Replace Response Authenticator with the Request Authenticator
 
     response_type = "1009" if len(salt) <= 16 else "1017"
     print(
-        f"{ip}:$dynamic_{response_type}${binascii.hexlify(hash_val).decode()}$HEX${binascii.hexlify(salt).decode('utf-8')}"
+        f"{ip}:$dynamic_{response_type}${binascii.hexlify(ra).decode()}$HEX${binascii.hexlify(salt).decode('utf-8')}"
     )
 
     dumped_ips[ip] = "reply"
 
 
-def dump_access_request(ip, login, ra, hashed):  # 3.3 attack
-    if UNIQUE and ip in dumped_ips and dumped_ips[ip] == "request":
+def dump_access_request(args, ip, ra, hashed):  # 3.3 attack
+    if args["single"] and ip in dumped_ips and dumped_ips[ip] == "request":
         return
 
-    xor_result = bytes(a ^ b for a, b in zip(hashed, PASSWORD))
+    xor_result = bytes(a ^ b for a, b in zip(hashed, args['password'][:16].encode('utf-8').ljust(16, b'\x00')))
+
     print(
         f"{ip}:$dynamic_1008${binascii.hexlify(xor_result).decode()}$HEX${binascii.hexlify(ra).decode('utf-8')}"
     )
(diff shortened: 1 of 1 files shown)


solardiz:
Thank you @k4amos! For reference, the previous version was contributed via #3621.

I'd appreciate review by @MaximeGoyette @exploide @kholia.

Is this revised script for Python 3 and/or 2? The current shebang lines implies it supports both, which is great if so, but I'm not sure is the case.

k4amos:
Hello @solardiz,

I added the python3 shebang and a try / except block to display a message if scapy is not installed.

RADIUS accounting messages are on port 1813, I had forgotten to consider them, but I corrected that. The shared-secret can be cracked using these methods, as demonstrated by running the code on the first sample provided by Maxime Goyette https://github.com/openwall/john/pull/3621#issuecomment-458752330 

I realise that the perl version has also forgotten to consider accounting messages, do I create an issue for the perl version ?

run/radius2john.py
```diff
@@ -1,68 +1,139 @@
...
-    print("scapy is missing, run 'pip install --user scapy' to install it!")
-    exit(1)
+# ---
+
+# Application of two  methods described in http://www.untruth.org/~josh/security/radius/radius-auth.html :
+# "3.3 User-Password Attribute Based Shared Secret Attack"
+# "3.1 Response Authenticator Based Shared Secret Attack"
+
+# For attack 3.3 :
+# we try authentications using a known password, and sniff the radius packets to a pcpap file.
```
exploide: typo pcap

run/radius2john.py
```diff
@@ -1,68 +1,139 @@
...
+# This script reads access-request in the pcap file, and dumps the md5(RA+secret) and RA, in a john-friendly format.
+# The password must be always the same, be less then 16 bytes long, and entered in the $PASSWORD variable below.
+# The user names used during this attack must be entered in @LOGINS below.
+
+# For attack 3.1:
+# we don't need to try authentications. Just sniff the radius packets in a pcap file.
+# This script reads the pcap file, matches radius responses with the corresponding all_requests,
+# and dumps md5 and salt as needed.
+
+import scapy.all as scapy
```
exploide: There is a scapy import here as well as in the `if __name__ == "__main__"` block. The latter is guarded with a try-catch, but this first import here will crash the script way before it hits the main block. Please move the try-except import up here.

run/radius2john.py
```diff
@@ -1,68 +1,139 @@
...
+# ---
+
+# Application of two  methods described in http://www.untruth.org/~josh/security/radius/radius-auth.html :
+# "3.3 User-Password Attribute Based Shared Secret Attack"
+# "3.1 Response Authenticator Based Shared Secret Attack"
+
+# For attack 3.3 :
+# we try authentications using a known password, and sniff the radius packets to a pcpap file.
+# This script reads access-request in the pcap file, and dumps the md5(RA+secret) and RA, in a john-friendly format.
+# The password must be always the same, be less then 16 bytes long, and entered in the $PASSWORD variable below.
```
exploide: As far as I understood, users need to touch the code for attack 3.3 in order to set the `PASSWORD` and `LOGINS` variables. I think it would be much nicer to have commandline options for that. The [argparse](https://docs.python.org/3/library/argparse.html) module comes in handy here. I can help with this if needed.

run/radius2john.py
```diff
@@ -1,68 +1,139 @@
...
+
+
+def process_packet(packet):
+    if packet.haslayer(scapy.IP) and packet.haslayer(scapy.UDP):
+        ip_layer = packet[scapy.IP]
+        udp_layer = packet[scapy.UDP]
+
+        if udp_layer.dport in [1812, 1813] or udp_layer.sport in [1812, 1813]: 
+            radius_data = bytes(udp_layer.payload)
+            process_radius(ip_layer, radius_data, bytes(udp_layer.payload))
```
exploide: you set `radius_data = bytes(udp_layer.payload)`, then you pass `radius_data` as well as `bytes(udp_layer.payload)` as parameters to the `process_radius` function. This is two times the same data.

run/radius2john.py
```diff
@@ -1,68 +1,139 @@
...
+
+
+def process_radius(ip, radius_data, udpdata):
+    radius_packet = scapy.Radius(radius_data)
+
+    if radius_packet.code in [1, 4]:  # Access-Request, Accounting-Request
+
+        user_name, user_password = None, None
+        for _ in range(len(radius_packet.attributes)):
```
exploide: This can be simplified to `for attr in radius_packet.attributes:` and then directly accessing `attr.name` and `attr.value`. In Python `for` loops are usually meant as for-each loops. Iterating over indices is only done when really necessary.

run/radius2john.py
```diff
@@ -1,68 +1,139 @@
...
+            if radius_packet.attributes[_].name == "User-Name":
+                user_name = radius_packet.attributes[_].value.decode("utf-8")
+
+            if radius_packet.attributes[_].name == "User-Password":
+                user_password = radius_packet.attributes[_].value.decode("utf-8")
+
+        if user_name in VALID_LOGIN and user_password:
```
exploide: Is it possible that the extracted `user_password` is an empty string and would that still be a valid case? If yes, then the condition needs to be `and user_password is not None` because the empty string also evaluates to `False` in an `if`.

run/radius2john.py
```diff
@@ -1,68 +1,139 @@
...
-    accounting_request_packets = {}
-    accounting_response_packets = {}
+    dumped_ips[ip] = "reply"
 
-    for packet in capture_file:
 
-        if Radius in packet:
-            src_ip = packet[IP].src
-            dst_ip = packet[IP].dst
+def dump_access_request(ip, login, ra, hashed):  # 3.3 attack
```
exploide: unused argument `login`

run/radius2john.py
```diff
@@ -1,68 +1,139 @@
...
+if __name__ == "__main__":
 
-            file_path = os.path.abspath(filename)
+    try:
+        import scapy.all as scapy
+    except ImportError:
+        print(
+            "Scapy seems to be missing, run 'pip install --user scapy' to install it"
+        )
+        exit(1)
```
exploide: using `sys.exit(1)` is recommended over `exit(1)`

exploide (changes requested):
I had a look on the script. Thanks for the contribution. :)

I left a few comments and suggestions. I only read and linted the script for know, I haven't actually executed it yet. I assume we only have the test data available in https://github.com/openwall/john/pull/3621#issuecomment-458752330 ? Nothing in the john-samples repo unfortunately.

If further questions arise or when I could help with code changes, please note that I'm away over the weekend.

solardiz:
> I added the python3 shebang

In other words, you're saying the script is now Python 3 only? Was the previous one working with either Python version? If so, that's a minor regression (but not a blocker).

> I realise that the perl version has also forgotten to consider accounting messages, do I create an issue for the perl version ?

Yes, or ideally a pull request right away. Thank you!

solardiz:
> Nothing in the john-samples repo unfortunately.

BTW, @k4amos @MaximeGoyette we welcome contributions via pull requests to https://github.com/openwall/john-samples so that it's easier and more reliable to regression-test future versions.

k4amos:
Hi @solardiz and @solardiz, I hope I've corrected everything correctly. With the samples, you can test the code : the previous version of the code only worked with the first 2 samples, whereas this new version works with all 5 : 

previous version of the code : 
image

new version of the code : (by default, I no longer return a single hash)
image

You can test the 3.3 attack for example with the sample n°5 :
image
image

k4amos:
The previous version of the code seems to work well with both python2 and python3 @solardiz , this is a regression on this point, I have no solution to keep this compatibility 🫤

solardiz:
> Hi @solardiz and @solardiz, I hope I've corrected everything correctly.

Hi @k4amos and @k4amos. Thank you. For fix-up commits within a PR, we prefer that the original commit be amended and force-pushed, instead of adding more commits. Also, the commit titles should start with `radius2john.py: `, although if the only commit is `New version of radius2john.py` that is also fine.

In this case, you may squash the commits, or you may just tell me to use GitHub's squash-and-merge (which we normally have disabled to avoid inadvertent use, but I can briefly re-enable it for merging this one PR).

I haven't reviewed your actual changes. I hope @exploide will.

Also thank you for the sample files and for the test examples. It would be even better if you post such examples as plain text copy-paste, rather than as screenshots, so they would also be in email.

solardiz:
> The previous version of the code seems to work well with both python2 and python3 @solardiz , this is a regression on this point, I have no solution to keep this compatibility 🫤

OK. Does the Perl script provide the same functionality? If so, I suppose it would be the one to use on older systems that don't have Python 3 installed.

Edit: Oh, I suppose this (from an earlier comment) implies the Perl version isn't yet as functional:

> I realise that the perl version has also forgotten to consider accounting messages

run/radius2john.py
```diff
@@ -1,68 +1,163 @@
...
+
+            for attr in radius_packet.attributes:
+                if attr.name == "User-Name":
+                    user_name = attr.value.decode("utf-8")
+
+                if attr.name == "User-Password":
+                    user_hash = attr.value
+
+            if user_hash is not None:
+                if (args['login'] is not None and user_name == args['login']) or args['login'] is None:
```
exploide: I think this can be simplified to `if args['login'] is None or user_name == args['login']`

run/radius2john.py
```diff
@@ -1,68 +1,163 @@
...
+    We don't need to try authentications. Just sniff the radius packets in a pcap file.
+    This script reads the pcap file, matches radius responses with the corresponding all_requests,
+    and dumps md5 and salt as needed.
+    		""")
 
-            if packet[Radius].code == 4 and not (src_ip, dst_ip) in accounting_request_packets:
-                accounting_request_packets[(src_ip, dst_ip)] = packet
-            elif packet[Radius].code == 5 and not (dst_ip, src_ip) in accounting_response_packets:
-                accounting_response_packets[(dst_ip, src_ip)] = packet
+    parser.add_argument('-f', '--file', type=str, required=True, nargs='+')
```
exploide, line 143: the `-f` option conflicts with the your hand-written usage message (at multiple places) that the basic usage is `./radius2john.py <pcap files>`. Either adapt the usage message, or better, use a positional argument without option flag for the files.

run/radius2john.py
```diff
@@ -1,68 +1,163 @@
...
-    for ips, packet in accounting_response_packets.items():
-        if ips in accounting_request_packets:
-            code = hex(packet[Radius].code)[2:].zfill(2)
-            identifier = hex(packet[Radius].id)[2:].zfill(2)
-            length = hex(packet[Radius].len)[2:].zfill(4)
-            authenticator = hexlify(accounting_request_packets[ips][Radius].authenticator).decode('utf-8')
 
-            salt = code + identifier + length + authenticator
+    parsed_args = parser.parse_args()
+    args = vars(parsed_args)
```
exploide, line 149: I don't understand why this is necessary. You could just access something like `parsed_args.login`. But it's not important.

run/radius2john.py
```diff
@@ -1,68 +1,163 @@
...
 
-            hash_content = hexlify(packet[Radius].authenticator).decode('utf-8')
-            hash_type = 1009 if len(salt) <= 16 else 1017
+    if args["login"] is not None and args["password"] is None:
```
exploide: Did I understand correctly, that you always need both, login and password for attack 3.3? If yes, then also a check for password is not None but login is would be required, wouldn't it?

exploide (changes requested):
Thanks for improving the PR. I did another review round. Much less comments now, and less important. Please have a look.

Concerning python2 compatibility: Personally I don't care tbh. If we require python2 compatibility than we cannot use modern language features, like f-strings, need multiple switches and workarounds and increase the maintenance overhead in general. A user however, it not forced to execute the script on an ancient system but can just copy the file to a modern platform. I think lowering the developer and maintenance effort is more important. Python2 is end of life for years. But @solardiz has to decide this. (If you ask me, I would even dump the perl script since this one here has all features the perl script has plus more.)

run/radius2john.py
k4amos: I don't necessarily need the login for the 3.3 attack: if there's only one client, there's no point in checking that it's the right one (because that's necessarily the case), but in all cases, I need the password. That's why if I have a login but no password, I generate an error. 

```bash
user@computer:~/Documents$ python3 john/run/radius2john.py -f john-samples/Radius/radius-sample-5.pcap -l bob 
You must specify the password used by the client for the '3.3 User-Password Attribute Based Shared Secret Attack'
Basic Usage: ./radius2john.py -f <pcap files>
You can specify '-h' to display help
```


On the other hand, if I just have just the password but no login, the code can work : 
```bash
user@computer:~/Documents$ python3 john/run/radius2john.py -f john-samples/Radius/radius-sample-5.pcap -p ThisIsThePassword
194.247.5.25:$dynamic_1008$9df88bcafd6b2550fcf2c62c87dd8619$HEX$1da7522f1df80483e73573c72b7d9114
194.247.5.25:$dynamic_1017$4fb969d9dc9d7aee6f58132e0b9662cd$HEX$021300201da7522f1df80483e73573c72b7d9114120c48656c6c6f2c20626f62
```

k4amos: I'm going to add this explanation to the code

k4amos:
If it's possible to make a GitHub's squash-and-merge @solardiz instead of editing the commits already made, that would be perfect 😃

solardiz:
> If it's possible to make a GitHub's squash-and-merge @solardiz instead of editing the commits already made, that would be perfect 😃

OK, I'll plan on doing that. Does this look ready, @exploide?

exploide (approved):
Yes, I guess this is ready now.

solardiz:
This is now merged. However, I made the mistake of not letting CI run on this before merging - my thinking was we don't test Python scripts anyway - but we do have the whitespace errors test for all text files, and this test failed. I've now made a follow-up commit fixing the whitespace errors.

@k4amos Please take a look at how this looks in the tree now. Thank you!

Issue #4600, opened 2021-03-08
c9825e6 fails on armhf


ghost:
> Bus error               (core dumped)

- It builds, I see no warnings;
- A `--list=build-info` works;
- But a simple `$JTR_BIN --stdout --mask='[0-2]password[A-C]'` fails;
- arm64 **works**.

*****
https://launchpadlibrarian.net/526869316/buildlog_snap_ubuntu_bionic_armhf_john-the-ripper_BUILDING.txt.gz

```
Target CPU ................................. armv7l NEON, 32-bit LE
AES-NI support ............................. no
Target OS .................................. linux-gnueabihf
Cross compiling ............................ no
Legacy arch header ......................... arm32le.h

Optional libraries/features found:
Memory map (share/page large files) ........ yes
Fork support ............................... yes
OpenMP support ............................. yes (not for fast formats)
OpenCL support ............................. no
Generic crypt(3) format .................... yes
libgmp (PRINCE mode and faster SRP formats)  yes
128-bit integer (faster PRINCE mode) ....... no
libz (pkzip and some other formats) ........ yes
libbz2 (gpg2john extra decompression logic)  yes
libpcap (vncpcap2john and SIPdump) ......... yes
OpenMPI support (default disabled) ......... no
ZTEX USB-FPGA module 1.15y support ......... no

Install missing libraries to get any needed features that were omitted.

Configure finished.  Now "make -s clean && make -sj4" to compile.
stribog_fmt_plug.c:472:2: warning: #warning Stribog-256 and Stribog-512 formats require SSE 4.1, formats disabled [-Wcpp]
 #warning Stribog-256 and Stribog-512 formats require SSE 4.1, formats disabled
  ^~~~~~~
ar: creating aes.a
ar: creating poly1305-donna.a
ar: creating ed25519-donna.a
ar: creating secp256k1.a

Make process completed.
```

```
Version: 1.9.0-jumbo-1+bleeding-c9825e6 2021-03-08 11:40:52 +0100
Build: linux-gnueabihf 32-bit armv7l NEON AC OMP
SIMD: NEON, interleaving: MD4:2 MD5:2 SHA1:1 SHA256:1 SHA512:1
System-wide exec: /snap/john-the-ripper/current/run
System-wide home: /snap/john-the-ripper/current/run
Private home: ~/.john
OMP fallback binary: john-non-omp
$JOHN is /snap/john-the-ripper/current/run/
Format interface version: 14
Max. number of reported tunable costs: 4
[...]

PR #5493, opened 2024-06-10
RAR format: Allow to use UBSAN in tests on AMD64

ghost:
It seems ok to me as this change will only impact someone when/if UBSAN (a "Development Option") is enabled.

Works fine locally:
```
$ OMP_NUM_THREADS=3 ../run/john --test=0 --format=rar
Will run 3 OpenMP threads
Testing: rar, RAR3 (length 5) [SHA1 256/256 AVX2 8x AES]... (3xOMP) PASS
```

```
$ ../run/john --list=build-info
Version: 1.9.0-jumbo-1+bleeding-82d9e7f383 2024-06-10 10:16:32 -0300
Build: linux-gnu 64-bit x86_64 AVX2 AC OMP UbSan
[...]
Parsed terminal locale: UTF-8
Built with these debugging options
        UbSan (Undefined Behavior Sanitizer debugging)
```

### DIFF
--- a/src/rar_fmt_plug.c
+++ b/src/rar_fmt_plug.c
@@ -45,7 +45,7 @@
 #endif
 #include "arch.h"
 
-#if ARCH_ALLOWS_UNALIGNED || __ARM_FEATURE_UNALIGNED
+#if ARCH_ALLOWS_UNALIGNED || __ARM_FEATURE_UNALIGNED || (WITH_UBSAN && (__amd64__ || __x86_64 || __x86_64__))
 
 #if FMT_EXTERNS_H
 extern struct fmt_main fmt_rar;


solardiz:
Hmm, it's surprising this format works with UbSan on any platform, given that it's said to require unaligned accesses and UbSan is supposed to catch those.

Nitpick: you check `__x86_64__` twice.

ghost:
It is (both are defined in my local gcc):
```
__x86_64  VERSUS
__x86_64__
```

I can test using clang before we merge it. gcc 11.4, Ubuntu 22 it works fine. Double checked now.

solardiz:
For some context behind us disabling the RAR format on archs without unaligned access, see https://github.com/openwall/john/issues/2868#issuecomment-349657708 and #4600

The thing with requiring alignment is that passing tests do not guarantee the code is correct and will work on another build. We may just happen to have alignment in that build/run, which isn't unusual - e.g. an array of chars on the stack would most often have greater alignment, yet without a guarantee.

So I'm tempted not to merge this PR without proper understanding as to why this passes UbSan, and whether that's consistent.

solardiz:
> So I'm tempted not to merge this PR without proper understanding as to why this passes UbSan, and whether that's consistent.

And if the reason is there's no unaligned access possibility, including not in actual usage of the format, then we'd need to remove the check - just have this format available on all archs. If the reason is we're lacking a test vector that would expose the issue, we should identify and add such test vector.

ghost:
That is ok. I would like the format to be enabled for testing, but I can ignore the format or patch it for testing.
****
It also works on clang 18 (AVX2, no OMP).

```
[...]
Testing: RAKP, IPMI 2.0 RAKP (RMCP+) [HMAC-SHA1 256/256 AVX2 8x2]... PASS
Testing: rar, RAR3 (length 5) [SHA1 256/256 AVX2 8x2 AES]... PASS
Testing: RAR5 [PBKDF2-SHA256 256/256 AVX2 8x]... PASS
Testing: Raw-SHA512 [SHA512 256/256 AVX2 4x]... PASS
[...]
Testing: dynamic_2014 [md5($s.md5($p).$s) (PW > 55 or salt > 11 bytes) 256/256 AVX2 8x5]... PASS
Testing: dummy [N/A]... PASS
Testing: crypt, generic crypt(3) [?/64]... PASS
All 425 formats passed self-tests
```

```
+ git am 5493.patch
Applying: RAR format: Allow to use UBSAN in tests on AMD64
```

ghost:
Thinking a little more about the problem, it still seems to me that only the test that fails should be disabled using:

E.g., probably: `__attribute__((no_sanitize("alignment")))`. 

Are we not testing it because in some other place, condition or architecture it might fail? Bad strategy.
UBSAN is not just about alignment.

PR #5508, opened 2024-07-07
pcap2john.py: improvements for htdigest, gadu, cleanup

exploide:
Some enhancements for `pcap2john.py`.

- Rewrote `pcap_parser_htdigest` function to use scapy for parsing. Prior to this change, it was crashing on pcapng files (as opposed to pcap). Furthermore, it was not compatible with Python 3, now it is. While I was at it, I decided to print Basic Auth credentials directly to stderr. Could be interesting too, but no cracking required. Renamed the function to `pcap_parser_http_authorization` in this course. Since there was no example file for testing I created https://github.com/openwall/john-samples/pull/26.

- A cleanup commit removes an annoying call to `sleep` and adds a few comments.

- Accidentally, I discovered `pcap_parser_gadu` was not compatible with Python 3, now it is. I refactored the function a bit to make it more conformant to the remainder of the script. Tested with the [available sample files](https://github.com/openwall/john-samples/tree/main/Gadu-Gadu).

### DIFF
--- a/run/pcap2john.py
+++ b/run/pcap2john.py
@@ -896,54 +896,43 @@ def pcap_parser_wlccp(fname):
     f.close()
 
 
-def endian(s):
-    ret = ""
-    for i in range(0, len(s), 2):
-        ret += s[i + 1] + s[i]
-    return ret
-
-
-def process_hash(uid, nonce, sha1):
-    if len(nonce) == 0:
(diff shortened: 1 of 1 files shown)


solardiz:
Thank you very much for your contributions, @exploide!

PR #5512, opened 2024-07-11
Update DYNAMIC doc with new formats

SamuraiOcto:


solardiz (approved):
Oh wow, apparently we were missing many more of these from there. Thank you for adding them! I did not verify each and every line here, but overall this looks right.

solardiz:
Separately, we could want to reformat the table so that the columns are aligned again. I welcome such contribution via a separate PR.

PR #5532, opened 2024-09-11
Bugfix for PDF, and a couple of minor tweaks

magnumripper:
* PDF format: Bug fixes for R3 with RC4-40
* pbkdf2-hmac-sha256: Support MIME Base64
* loader.c: Fix a bug which garbled encodings for --show=formats

### DIFF
--- a/src/pbkdf2_hmac_common_plug.c
+++ b/src/pbkdf2_hmac_common_plug.c
@@ -116,12 +116,13 @@ char *pbkdf2_hmac_md4_split(char *ciphertext, int index, struct fmt_main *self)
 {
 	static char out[PBKDF2_MD4_MAX_CIPHERTEXT_LENGTH + 1];
 	char *cp;
+
 	strnzcpylwr(out, ciphertext, sizeof(out));
-	cp = strchr(out, '.');
-	while (cp) {
+
+	cp = out;
+	while ((cp = strchr(cp, '.')))
 		*cp = '$';
-		cp = strchr(cp, '.');
-	}
+
 	return out;
 }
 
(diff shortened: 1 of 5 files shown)


src/pbkdf2_hmac_common_plug.c
```diff
@@ -638,6 +638,22 @@ char *pbkdf2_hmac_sha256_prepare(char *fields[10], struct fmt_main *self) {
...
 	char tmp[43+1], *cp;
 
+	// MIME Base64, optionally with escaped slashes. Trailing '=' is also optional.
+	if (!strncmp(fields[1], PBKDF2_SHA256_FORMAT_TAG, PBKDF2_SHA256_TAG_LEN)) {
+		if (strchr(fields[1], '\\') || strchr(fields[1], '=') || strchr(fields[1], '+')) {
+			strcpy(Buf, fields[1]);
+			while ((cp = strchr(Buf, '\\')))
+				memmove(cp, cp + 1, strlen(cp));
+			while ((cp = strchr(Buf, '=')))
+				memmove(cp, cp + 1, strlen(cp));
```
solardiz, line 654: I'd speed these up a bit by doing a `cp = Buf` before the loops and then `while ((cp = strchr(cp, '\\')))`, so that only the remaining portion of string is searched.

src/pbkdf2_hmac_common_plug.c
```diff
@@ -638,6 +638,22 @@ char *pbkdf2_hmac_sha256_prepare(char *fields[10], struct fmt_main *self) {
...
+			cp = strchr(Buf, '+');
+			while (cp) {
+				*cp = '.';
+				cp = strchr(cp, '+');
+			}
```
solardiz: I'd simplify this by doing a cp = Buf before the loop and then `while ((cp = strchr(cp, '+')))`, so that the `strchr()` call is in one place.

solardiz (approved):
It's weird that we have optimized scalar MD5 here - I guess we could have so much more speedup by using SIMD. But indeed we already had that and if it was wrong for some inputs, we should fix that first - like you do.

src/pbkdf2_hmac_common_plug.c
magnumripper: They were based on nearby copy/paste... I fixed all similar cases in this file now.

solardiz: Looks good to me, thanks!

PR #5533, opened 2024-09-11
Astra Linux crypt variants using GOST R 34.11-94 or GOST R 34.11-2012

magnumripper:
The former is 256-bit. The latter is 256 or 512-bit and also known as Streebog. The formats exactly match Drepper's sha256crypt and sha512crypt, with the hash primitive replaced with one of the GOST alternatives.

gost94crypt      (`$gost94hash$`)
streebog256crypt (`$gost12256hash$`)
streebog512crypt (`$gost12512hash$`)


run/opencl/gost12256hash_kernel.cl
```diff
@@ -0,0 +1,302 @@
...
+
+	for (uint i = lid; i < 256; i += ls) {
+		for (uint j = 0; j < 8; j++)
+			loc_buf->Ax[j][i] = Ax[j][i];
+	}
+	barrier(CLK_LOCAL_MEM_FENCE);
+#endif
+
+	/* Repeatedly run the collected hash value through Streebog to burn CPU cycles.  */
+#pragma unroll HASH_LOOPS
```
solardiz, line 204: I guess this unroll produces truly a lot of code. We should be able to optimize this without so much code later (not in this PR).

solardiz (approved):
OK, let's get these in, although I doubt all 3 are in active use.

magnumripper:
What's with the circleci stall? Should I ignore it and merge? I can't seem to click it for looking into it.

> We should be able to optimize this without so much code later (not in this PR)

Yeah, best have a version with readable code in history before going crazy with it. Actually, I based these CPU formats on pristine versions of sha256crypt and sha512crypt formats from our Git history, that was handy for sure (before that, I tried basing it on the current versions with Jim's tweaks but something failed and I lost interest).

solardiz:
> What's with the circleci stall? Should I ignore it and merge? I can't seem to click it for looking into it.

Same here. I think yes, you should ignore it and merge. Thank you!

ghost:
> What's with the circleci stall?

The "stall"  means that an administrator must review the (branch protection) rules. As seen in:

> Repository administrators can add constraints like this to branches using branch protection rules. For more information, see "[About protected branches](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches)."


The `ci/circleci: non-SIMD` should no longer be required.

solardiz:
> The `ci/circleci: non-SIMD` should no longer be required.

I've just removed it from required checks, and somehow this resulted in it disappearing from pending checks completely. I have no idea why it was getting stalled lately nor why it disappeared completely when not required.

ghost:
Just to let you know in case you get stuck and bored in an airport:

```
Testing: bitwarden-opencl, Bitwarden Password Manager [PBKDF2-SHA256 AES OpenCL]... PASS
Testing: blockchain-opencl, blockchain My Wallet (v2 x5000) [PBKDF2-SHA1 AES OpenCL]... PASS
Testing: cloudkeychain-opencl, 1Password Cloud Keychain [PBKDF2-SHA512 AES OpenCL]... PASS
##[warning]Free memory is lower than 5%; Currently used: 97.07%
##[warning]Free memory is lower than 5%; Currently used: 97.07%
##[warning]Free memory is lower than 5%; Currently used: 97.07%
##[warning]Free memory is lower than 5%; Currently used: 97.07%
Testing: streebog256crypt-opencl, Astra Linux $gost12256hash$ (rounds=5000) [GOST R 34.11-2012 OpenCL]... run_tests.sh: line 304:  6476 Killed                  "$JTR_BIN" -test-full=0 --format=opencl
```

Various formats work there:
```
[...]
Testing: zed-opencl, Prim'X Zed! encrypted archives [PKCS#12 PBE (SHA1/SHA256) OpenCL]... PASS
Testing: ZIP-opencl, WinZip [PBKDF2-SHA1 OpenCL]... PASS
All 72 formats passed self-tests
 Ok: -test-full=0 --format=opencl
```

It is Intel CPU driver:
```
Platform #0 name: Intel(R) OpenCL, version: OpenCL 3.0 LINUX
    Device #0 (1) name:     Intel(R) Xeon(R) Platinum 8171M CPU @ 2.60GHz
    Device vendor:          Intel(R) Corporation
    Device type:            CPU (LE)
    Device version:         OpenCL 3.0 (Build 0)
    OpenCL version support: OpenCL C 3.0 
    Driver version:         2023.16.10.0.17_160000 

```

run/opencl/gost12256hash_kernel.cl
solardiz: @magnumripper Maybe we should simply drop these `#pragma unroll HASH_LOOPS` lines from here? Were they meant to be included in the pushed code? There's even a comment in `gost12512hash_kernel.cl` that says:

```
 * Without unroll: Build time: 45.125 s, binary size 2542391, 10402 c/s
 * Pragma unroll:  Build time: 21 min 14.788 s, binary size 11502092, 10051 c/s
 * Manual unroll:  Build time: 4 min 26.142 s, binary size 5961490, 9959 c/s
```

So are you saying you had better build times and better runtime speeds without the unroll, yet kept the unroll? Weird.

solardiz:
@magnumripper Is the placement of braces here right? -

```c
#define LOOP_CALLS              (5000 + (HASH_LOOPS - 1) / HASH_LOOPS)
```

Maybe it was meant to be `((5000 + (HASH_LOOPS - 1)) / HASH_LOOPS)`?

I think we can skip invocation of the final kernel when `loops` is a multiple of `HASH_LOOPS`, right? But this needs a test vector that would trigger this condition.

run/opencl/gost12256hash_kernel.cl
solardiz: Testing on various ancient systems, this full unroll results in various "out of resources" errors. Reducing the unroll to 42 also gives such errors, but quicker. Removing the unroll makes the kernel build and invocation succeed, but the the formats commonly fail self-test on `cmp_all`.

magnumripper:
```c
/*
 * Without unroll: Build time: 45.125 s, binary size 2542391, 10402 c/s
 * Pragma unroll:  Build time: 21 min 14.788 s, binary size 11502092, 10051 c/s
 * Manual unroll:  Build time: 4 min 26.142 s, binary size 5961490, 9959 c/s
 */
#define STREEBOG_UNROLL         0
#define STREEBOG_MANUAL_UNROLL  0
#include "opencl_streebog.h"
```
Those comments are about unrolling within opencl_streebog.h (controlled by the macros that are indeed set to 0 before sourcing it). 

The `#pragma unroll HASH_LOOPS` in gost12512hash_kernel.cl is a different beast. I assume I got better results with it on whatever nvidia I tested with at the time.

solardiz: Also unroll 21 could be reasonable (leaves only the trivial `& 1` checks in the loop) and it even passes my build/run tests (where 42 fails), only failing self-test (wrong compute as usual for these ancient systems). Build time and code size e.g. for `gost94crypt-opencl` with old Catalyst targeting Tahiti is 11 seconds and 1 MB without unroll, 110 seconds and 5 MB with unroll 21.

All of these code sizes are too large. I doubt that even runtime `% 3` and `% 7` checks make much of a difference when code doesn't fit in cache anyway. And I don't know why it's so large without unrolling?

Anyway, instead of unrolling we could use one of the tricks we already use elsewhere. Like maintaining counter(s) that are reset to 0 upon reaching 3, 7, 21, or 42 (and use an offsets map like `md5crypt_long_fmt.c` does). Crazy unrolling is really not the way.

solardiz:
Also, other formats call this not `LOOP_CALLS`, but `LOOP_COUNT`, and have it consider the actual iteration count (of one of the loaded hashes? not sure).

magnumripper:
> Maybe it was meant to be `((5000 + (HASH_LOOPS - 1)) / HASH_LOOPS)`?

Yes, or `((5000 + HASH_LOOPS - 1) / HASH_LOOPS)` as I probably thought I was typing. Definitely a bug.

> I think we can skip invocation of the final kernel when `loops` is a multiple of `HASH_LOOPS`, right? But this needs a test vector that would trigger this condition.

I haven't looked now but even if we can, it shouldn't matter at these speeds.

PR #5549, opened 2024-11-01
PDF-opencl

magnumripper:
Adds pdf-opencl with device-side mask. Also bugfix in CPU format for rev. 3 documents with RC4-40 as well as added support for rev 2-4 documents with key length other than 40 or 128 bits (spec. allows anything inbetween in multiples of 4).

<s>First</s> Second commit (bugs/workarounds to cryptosafe-opencl) is kind of unrelated but I used that kernel as template so fixed it while at it.
Other separate commits enhances shared RC4 (support for arbitrary length [and unrolled 40-bit] key) and SHA-2 (adds SHA-384) and make corresponding changes to callers (oldoffice-opencl, krb5pa-md5-opencl and krb5tgs-opencl for RC4 and tezos-opencl, pbkdf2-hmac-sha512-opencl for a SHA-512 macro rename).

### DIFF
--- a/run/opencl/opencl_misc.h
+++ b/run/opencl/opencl_misc.h
@@ -202,6 +202,9 @@ inline uint funnel_shift_right_imm(uint hi, uint lo, uint s)
 #endif
 #endif
 
+#define block_swap32(W, len)	for (uint i = 0; i < len; i++) W[i] = SWAP32(W[i])
+#define block_swap64(W, len)	for (uint i = 0; i < len; i++) W[i] = SWAP64(W[i])
+
 inline ushort SWAP16(ushort x)
 {
 	return ((x << 8) + (x >> 8));
@@ -348,6 +351,8 @@ inline MAYBE_VECTOR_UINT VSWAP32(MAYBE_VECTOR_UINT x)
 #define GETCHAR_BE(buf, index) (((uchar*)(buf))[(index) ^ 3])
 #define GETCHAR_MC(buf, index) (((MAYBE_CONSTANT uchar*)(buf))[(index)])
 #define LASTCHAR_BE(buf, index, val) (buf)[(index)>>2] = ((buf)[(index)>>2] & (0xffffff00U << ((((index) & 3) ^ 3) << 3))) + ((val) << ((((index) & 3) ^ 3) << 3))
+#define LASTCHAR_BE64(buf, index, val) (buf)[(index)>>3] = ((buf)[(index)>>3] & (0xffffffffffffff00UL << ((((index) & 7) ^ 7) << 3))) + ((ulong)(val) << ((((index) & 7) ^ 7) << 3))
+#define LASTCHAR(buf, index, val) (buf)[(index)>>2] = ((buf)[(index)>>2] & (0xffffff00U << (((index) & 3) << 3))) + ((val) << (((index) & 3) << 3))
 
 #if no_byte_addressable(DEVICE_INFO) || !SCALAR || (gpu_amd(DEVICE_INFO) && defined(AMD_PUTCHAR_NOCAST))
 /* 32-bit stores */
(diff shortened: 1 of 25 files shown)


ghost (approved):
Good stuff(tm)

magnumripper:
> Also bugfix in CPU format for rev. 3 documents with RC4-40 as well as added support for rev 2-4 documents with key length other than 40 or 128 bits (spec. allows anything inbetween in multiples of 4).

I see now I already pushed that fix to the existing CPU format in 07aef97d. It's now also checked in `valid()` though (key length 40..256 in multiples of 4) and we've got a proper test vector for RC4-40.

src/opencl_pdf_fmt_plug.c
```diff
@@ -0,0 +1,509 @@
+/*
+ * This software is Copyright (c) 2024 magnum and it is hereby released to the
+ * general public under the following terms:  Redistribution and use in source
+ * and binary forms, with or without modification, are permitted.
+ */
+
+#define FORMAT_STRUCT fmt_pdf_opencl
```
solardiz, line 7: Why do you need the `FORMAT_STRUCT` macro?

src/pdf_fmt_plug.c
```diff
@@ -16,369 +16,43 @@ extern struct fmt_main fmt_pdf;
...
 	omp_autotune(self, OMP_SCALE);
 
-	saved_key = mem_calloc(sizeof(*saved_key), self->params.max_keys_per_crypt);
-	any_cracked = 0;
-	cracked_size = sizeof(*cracked) * self->params.max_keys_per_crypt;
-	cracked = mem_calloc(sizeof(*cracked), self->params.max_keys_per_crypt);
+	pdf_key = mem_calloc(sizeof(*pdf_key), self->params.max_keys_per_crypt);
+	any_pdf_cracked = 0;
+	pdf_cracked_size = sizeof(*pdf_cracked) * self->params.max_keys_per_crypt;
+	pdf_cracked = mem_calloc(sizeof(*pdf_cracked), self->params.max_keys_per_crypt);
```
solardiz, line 47: Why these renames to `pdf_`?

solardiz (approved):
This doesn't look wrong. ;-)

run/opencl/opencl_misc.h
```diff
@@ -591,75 +596,42 @@ inline int memmem_pc(const void *haystack, size_t haystack_len,
...
-/* requires char/uchar */
-#define dump_stuff8_msg(msg, x, size) do {	  \
-		uint ii; \
 		printf("%s : ", msg); \
-		for (ii = 0; ii < (uint)size; ii++) { \
-			printf("%02x", (x)[ii]); \
-			if (ii % 4 == 3) \
+		for (uint xedni_ = 0; xedni_ < (uint)size; xedni_++) { \
+			printf("%02x", (x)[xedni_]); \
+			if (xedni_ % 4 == 3) \
```
solardiz, line 612: Why the name `xedni_`? Edit: Oh, I see, it's `_index` backwards. Presumably not to clash with any macro name.

src/opencl_pdf_fmt_plug.c
magnumripper: It was in whatever I used as template - probably something predating our current use/storing of `self`.

run/opencl/opencl_misc.h
magnumripper: I assume I wouldn't be protected from someone calling with eg. `buffer[ii]` as `x` ending up as `printf("%02x", (buffer[ii])[ii]);`. Or would I?

solardiz: You're right, you need a weird name here also for that reason.

src/opencl_pdf_fmt_plug.c
solardiz: Oh, I now see we do have this `FORMAT_STRUCT` thing in a couple of formats already. This will be the third.

src/pdf_fmt_plug.c
magnumripper: Most of them are now shared with the OpenCL format so not static, but I see now that `pdf_key` should be made static - and then doesn't need that prefix.

run/opencl/pdf_kernel.cl
```diff
@@ -0,0 +1,695 @@
...
+		uint *WP = W;
+		for (uint left = md5len; left > 55; left -= 64, WP +=16)
+			md5_block(uint, WP, digest);
+		md5_block(uint, WP, digest);
+
+		if (pdf_salt->key_length == 40) {
+			rc4_40_set_key(&rc4_ctx, key);
+			rc4(&rc4_ctx, digest, output, 16);
+			for (uint x = 0x01010101; x <= 0x13131313; x += 0x01010101) {
+				uint xor[2];
```
solardiz, line 345: Surprisingly, this array named `xor` fails build on ancient NVIDIA driver on our "bull". Renaming to `rox` makes it build and pass test.

solardiz: The name `key_xor_x` also works, so I think let's use that.

magnumripper: > Surprisingly, this array named xor fails

Interesting bug. Did it fail with some descriptive error?

solardiz: Yes, IIRC it said it was expecting a `(` after `xor`, so probably had it as a function or macro.

PR #5550, opened 2024-11-01
Tezos-opencl: Workaround for buggy nvidia driver

### LINKED ISSUE #5546: tezos-opencl kernel build takes longer than 10 hours with recent nvidia drivers
This was during a `--test=0 --format=opencl` (edit: and a 2080ti) with driver 550.107.02. I killed it after 45 minutes with 100% CPU. Did we see this before?

Also tried `-test=0 -form:tezos-opencl` by itself, same thing but I interrupted sooner. Will try other drivers (as long as they are readily available as Ubuntu packages).

magnumripper:
Closes #5546

### DIFF
--- a/run/opencl/ed25519-donna/ed25519-donna-impl-base.h
+++ b/run/opencl/ed25519-donna/ed25519-donna-impl-base.h
@@ -121,6 +121,9 @@ ge25519_scalarmult_base_choose_niels(ge25519_niels *t, uint32_t pos, signed char
 
 /* computes [s]basepoint */
 static void
+#if gpu_nvidia(DEVICE_INFO) && DEV_VER_MAJOR > 525 && DEV_VER_MAJOR < 560
+__attribute__((noinline))
+#endif
 ge25519_scalarmult_base_niels(ge25519 *r, const bignum256modm s) {
 	signed char b[64];
 	uint32_t i;


solardiz (approved):

run/opencl/ed25519-donna/ed25519-donna-impl-base.h
```diff
@@ -121,6 +121,9 @@ ge25519_scalarmult_base_choose_niels(ge25519_niels *t, uint32_t pos, signed char
 
 /* computes [s]basepoint */
 static void
+#if gpu_nvidia(DEVICE_INFO) && DEV_VER_MAJOR > 500
```
solardiz: The strict `> 500` is a bit weird, meaning 501 or higher. Perhaps you meant `>= 500`.

magnumripper: I *think* I had it working with 525.xx so maybe we should use that. Also I now confirmed the 560 or higher CUDA drivers work fine without the noinline, which is great! I'll amend it.

run/opencl/ed25519-donna/ed25519-donna-impl-base.h
```diff
@@ -121,6 +121,9 @@ ge25519_scalarmult_base_choose_niels(ge25519_niels *t, uint32_t pos, signed char
 
 /* computes [s]basepoint */
 static void
+#if gpu_nvidia(DEVICE_INFO) && DEV_VER_MAJOR > 525 && DEV_VER_MAJOR < 560
```
solardiz, line 124: Again, you probably mean `>= 525`, not `> 525`.

solardiz: Oh, I see you wrote you saw the issue with 535 and 550. But apparently not with 525, nor 560. If so, the check is right.

ghost: Generic, but very appropriate to the situation: it's documenting that anything after 525 and before 560 is or could be bad.

PR #5551, opened 2024-11-01
Streebog*crypt formats: Disable for BE builds

### LINKED ISSUE #5535: Astra Linux variants (streebog) fail on BE
We can fix or disable them on BE.

```
Testing: streebog256crypt, Astra Linux $gost12256hash$ (rounds=5000) [GOST R 34.11-2012 64/64]... (4xOMP) FAILED (cmp_all(64))
Testing: streebog512crypt, Astra Linux $gost12512hash$ (rounds=5000) [GOST R 34.11-2012 64/64]... (4xOMP) FAILED (cmp_all(64))
Testing: gost94crypt, Astra Linux $gost94hash$ (rounds=5000) [GOST R 34.11-94 32/64]... (4xOMP) PASS
[...]
Testing: crypt, generic crypt(3) [?/64]... (4xOMP) PASS
2 out of 410 tests have FAILED
 FAILED: -test-full=0 --format=cpu
```

```
Target CPU ......................................... s390x, 64-bit BE
Legacy arch header ................................. autoconf_arch.h
```

```
Version: 1.9.0-jumbo-1+bleeding-07aef97db6 2024-09-13 18:57:07 +0200
Build: linux-gnu 64-bit s390x  AC OMP
System-wide exec: /snap/john-the-ripper/current/bin
System-wide home: /snap/john-the-ripper/current/bin
Private home: ~/.john
OMP fallback binary: john-s390x
$JOHN is /snap/john-the-ripper/current/bin/
```

magnumripper:
The problem is not in these formats but in our shared Streebog code. It has clauses for BE but apparently they don't cut it.

Closes #5535

### DIFF
--- a/doc/NEWS
+++ b/doc/NEWS
@@ -328,6 +328,9 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
 
 - Added support for SM3.  [SamuraiOcto; 2024]
 
+- Added support for Astra Linux crypt variants using GOST R 34.11-94 or GOST R
+  34.11-2012.  [magnum; 2024]
+
 - More optimal bitslice DES S-box expressions for NVIDIA GPUs and AVX-512.
   [Sovyn Y., Solar; 2024]
 
--- a/src/gost12256hash_fmt_plug.c
+++ b/src/gost12256hash_fmt_plug.c
@@ -9,6 +9,8 @@
  * modification, are permitted.
  */
 
+#if ARCH_LITTLE_ENDIAN
+
 #if FMT_EXTERNS_H
 extern struct fmt_main fmt_gost12256hash;
 #elif FMT_REGISTERS_H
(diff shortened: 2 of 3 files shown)


solardiz (approved):
Looks good to me and is consistent with how we do it in other formats, but separately (after this is merged) we could want to clean them all up not to use this `#if !defined(FMT_EXTERNS_H) && !defined(FMT_REGISTERS_H)` construct, instead moving the little-endian check and warning to inside the existing main code block before `#endif /* plugin stanza */`.

solardiz:
> separately (after this is merged)

... or as a second commit here, I don't mind.

magnumripper:
> Looks good to me and is consistent with how we do it in other formats, but separately (after this is merged) we could want to clean them all up not to use this `#if !defined(FMT_EXTERNS_H) && !defined(FMT_REGISTERS_H)` construct, instead moving the little-endian check and warning to inside the existing main code block before `#endif /* plugin stanza */`.

I can't figure out what you mean. For big endian, we need to define away the whole format including the plugin stuff. So we need the check at the very start of the file.

solardiz:
> For big endian, we need to define away the whole format including the plugin stuff. So we need the check at the very start of the file.

Oh, you're right, and I was wrong.

PR #5554, opened 2024-11-02
OpenCL: remove the “AMD driver name”

ghost:
These days the printed information is wrong. 

I think it's best to just remove everything.

### DIFF
--- a/src/opencl_common.c
+++ b/src/opencl_common.c
@@ -296,8 +296,7 @@ static char *remove_spaces(char *str) {
 
 static char *opencl_driver_info(int sequential_id)
 {
-	static char buf[64 + MAX_OCLINFO_STRING_LEN];
-	char dname[MAX_OCLINFO_STRING_LEN], tmp[sizeof(buf)], set[64];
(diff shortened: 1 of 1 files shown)


magnumripper (approved):

solardiz (changes requested):
I agree in principle, and these changes look fine, but I think we should clean this up some more - also the `buf` array becomes unneeded, so we should remove it completely, and where we had `name = buf;` I think we should have `name = "";`

In fact, I think it was a bug that the contents of `buf` could be used uninitialized (via `name`) in the obscure case that one (or both) of the `if` conditions before the `do` is false, yet we see an AMD device later.

ghost:
I improvised a test here and everything seemed fine:
```
    Driver version:         5.0+debian - 2024.11(Claudio) [supported]
```
***
For some reason, I thought it considered everything given by NVIDIA to be OK and recommended. 
Anyway, something like 560 or 560+ is missing. It's the “current” software.

solardiz (approved):

Issue #4721, opened 2021-05-10
Support PuTTY 0.75+'s PPK3 private key file format, which uses Argon2


solardiz:
PuTTY 0.75 [release announcement](https://lists.tartarus.org/pipermail/putty-announce/2021/000031.html) says:

```
 - PuTTY's format for private key files has been updated. The new
   format, PPK3, does not depend on SHA-1, and also, it uses the
   Argon2 password hash function to make it more difficult for an
   encrypted key file stolen by an attacker to be brute-forced.
```

PR #5557, opened 2024-11-04
Sync Argon2 and Blake2 with upstream

### LINKED ISSUE #2738: Sync Argon2 implementation with latest upstream
Since we got the upstream Argon2 code into jumbo, there's been a contribution to upstream adding AVX2 support, which we lack:

https://github.com/P-H-C/phc-winner-argon2/pull/208/files

We should update, primarily to get this. (I expect that AVX-512 might follow, needing another sync.)

When we update, we should also try dropping the workaround we applied for #2650. I expect that the issue had been fixed upstream, and we must check this, and if not fixed (unlikely) then notify upstream that the issue is still unfixed.

We should also try to make it easier to sync to latest upstream. Can we possibly have all of Argon2 upstream code, even if with our changes, in a subdirectory, and keep the upstream's filenames? I understand this may be tricky to combine with our use of `*_plug.c` filenames. Maybe we need a generic alternative to those? Like also checking for `*_plug.a` static archives, which could be created from sets of object files built in subdirectories. Or we can treat Argon2 specially.

magnumripper:
This primarily adds Argon2id.  The Argon2 format got support it, with trivial parsing changes.  A test vector was also added to it. The reference Blake2 code included with Argon2 was replaced with the separate, more optimized one (just like we had before).

There are now optimizations for up to AVX512F in both (we previously had SSSE3 in Argon and XOP/AVX in Blake2). Some boost (3-4%) seen in raw-blake2 and argon2 formats, tested on AVX2.

Upstream https://github.com/P-H-C/phc-winner-argon2.git as of 4 Nov 2024 (Jun 25 10:21:15 2021 +0200).
Upstream https://github.com/BLAKE2/BLAKE2.git as of 4 Nov 2024 (Sat Jul 4 00:11:10 2020 +0100)

Closes #2738.
See also #2650 (the workaround we introduced was removed now). See also #4721.


magnumripper:
Some fix needed in Blake2 code for the case of JOHN_NO_SIMD or non-intel build. **EDIT:** Multiple ones in Argon2. Very tricky stuff that. It would have been harder without the bots.

ghost:
So far, it's passing all tests: Windows, macOS, BSD, other archs, ...

There are only a few more to go.

solardiz:
I wonder if it'd make sense to split this into more commits, like Blake2 in one and Argon2 in another, and/or upstream as-is first and our changes next (that's up to 4 commits for what I suggested here). I did the latter for yescrypt in:

```
commit c798c3f39215f6e08c67677eb9b79f65cfe08e40
Author: Solar <solar@openwall.com>
Date:   Mon Jul 17 22:02:50 2023 +0200

    yescrypt: Reintroduce relevant JtR hacks

commit 715be88adfcb842ba6b841c0307c6235168c98ce
Author: Solar <solar@openwall.com>
Date:   Mon Jul 17 21:58:45 2023 +0200

    yescrypt: Update to latest upstream code as-is
```

The upstream code as-is commit may be non-working or even non-buildable (this can be said in the commit description if so), but it could help figure things out later (such as in the next upstream sync or when figuring out where a bug came from).

magnumripper:
> like Blake2 in one and Argon2 in another

Yes. There are pros and cons: At first I had the updated Argon2 (with its Blake2 overwriting our old faster Blake2) in one commit and the updated faster Blake2 in another but that meant "history" would be lost (breaking things like `git log -p --follow blake2.h` [_edit: perhaps not that file, but files that only exist in standalone Blake2_]).

Another "and/or" option (and probably even better) would be to move Argon2 to its own subdirectory and the separate Blake2 to another. I considered that as well but it would've been more work - this whole thing was quite a lot of work as it were, in order to get right.

solardiz:
> this whole thing was quite a lot of work as it were, in order to get right.

Sure. OK, let's not create more work for you and merge this as-is.

solardiz (approved):
Not a full review. I intend to test this more right after merging.

solardiz:
I built this for AVX512BW, and the benchmarks say they use AVX512F, but for Argon2 speeds became a lot lower (by 40% maybe) and for Blake2 very slightly faster (by 3% maybe). Checking what code actually runs in `gdb`, I see some plain AVX code for Blake2 and scalar code for Argon2.

solardiz:
With this, Argon2 becomes actually a lot faster than before these changes (2x for 1 thread, 1.5x for 4 threads):

```diff
+++ b/src/argon2_opt_plug.c
@@ -19,7 +19,7 @@
  * Non-SIMD builds have these functions in argon2_ref_plug.c
  */
 #include "arch.h"
-#if defined(__SSE2__) && defined(JOHN_NO_SIMD)
+#if defined(__SSE2__) && !defined(JOHN_NO_SIMD)
 
 #include <stdint.h>
 #include <string.h>
diff --git a/src/argon2_ref_plug.c b/src/argon2_ref_plug.c
index b982e2b6e..8b13b4c3b 100644
--- a/src/argon2_ref_plug.c
+++ b/src/argon2_ref_plug.c
@@ -19,7 +19,7 @@
  * SIMD builds have these functions optimised in argon2_opt_plug.c
  */
 #include "arch.h"
-#if !defined(__SSE2__) || !defined(JOHN_NO_SIMD)
+#if !defined(__SSE2__) || defined(JOHN_NO_SIMD)
 
 #include <stdint.h>
 #include <string.h>
```

solardiz:
> There are now optimizations for up to AVX512F in both (we previously had SSSE3 in Argon and XOP/AVX in Blake2).

I see nothing beyond XOP/AVX for Blake2 here. There is up to AVX512F for BlaMka, which is a Blake2-derived construction used in Argon2, but not for Blake2 proper. I think we should revert the changes claiming AVX2+ for Blake2 from here:

```diff
+++ b/src/rawBLAKE2_512_fmt_plug.c
@@ -26,12 +26,14 @@ john_register_one(&fmt_rawBLAKE2);
 
 #define FORMAT_LABEL            "Raw-Blake2"
 #define FORMAT_NAME             ""
-#if !defined(JOHN_NO_SIMD) && defined(__AVX__)
-#define ALGORITHM_NAME          "128/128 AVX"
+#if !defined(JOHN_NO_SIMD) && defined(__AVX512F__)
+#define ALGORITHM_NAME          "512/512 AVX512F"
+#elif !defined(JOHN_NO_SIMD) && defined(__AVX2__)
+#define ALGORITHM_NAME          "256/256 AVX2"
 #elif !defined(JOHN_NO_SIMD) && defined(__XOP__)
 #define ALGORITHM_NAME          "128/128 XOP"
-#elif !defined(JOHN_NO_SIMD) && defined(__SSE4_1__)
-#define ALGORITHM_NAME          "128/128 SSE4.1"
+#elif !defined(JOHN_NO_SIMD) && defined(__AVX__)
+#define ALGORITHM_NAME          "128/128 AVX"
 #elif !defined(JOHN_NO_SIMD) && defined(__SSSE3__)
 #define ALGORITHM_NAME          "128/128 SSSE3"
 #elif !defined(JOHN_NO_SIMD) && defined(__SSE2__)
```

Also the removal of SSE4.1 reporting looks wrong. There's logic to use that.

solardiz:
I am preparing and testing commits fixing these issues.

solardiz:
I took care of the above two fixes. Pushed them directly to bleeding-jumbo after CI-testing on my fork of the repo.

Another major issue is that with this PR we've lost the memory allocation optimization. We were using my `init_region` / `alloc_region` extracted from early yescrypt. This let our modified Argon2 code reuse the memory allocation across calls (or replace it with a new one if the size would need to increase). The new upstream code we got probably allocates and frees on every call. We're still getting a speedup on my tests now perhaps due to AVX-512 and the test vectors being for a rather small allocation, but we probably do also incur slowdown compared to what was possible. And maybe there's a slowdown on below-AVX2 now - need to check.

solardiz:
> The new upstream code we got probably allocates and frees on every call.

Confirmed. Testing with XOP on our ancient "bull", somehow the speed is unchanged for 1 thread (maybe glibc manages to optimize such repeated allocations), but there's a 20% slowdown with 8 threads.

magnumripper:
Duh, I'm sorry, apparently we shouldn't have merged this yet.

> With this, Argon2 becomes actually a lot faster than before these changes (2x for 1 thread, 1.5x for 4 threads)

So a brain fart during my battle with the build bots. I should have re-tested speeds after juggling with the JOHN_NO_SIMD macro.

> I see nothing beyond XOP/AVX for Blake2 here. There is up to AVX512F for BlaMka, which is a Blake2-derived construction used in Argon2, but not for Blake2 proper. 

Oh, I thought blamka was part of blake2. Too many moving parts involved here. Thank you for fixing these issues.

solardiz:
> Duh, I'm sorry, apparently we shouldn't have merged this yet.

Not your fault, it was my decision to merge before my own performance testing.

solardiz:
> > The new upstream code we got probably allocates and frees on every call.
> 
> Confirmed. Testing with XOP on our ancient "bull", somehow the speed is unchanged for 1 thread (maybe glibc manages to optimize such repeated allocations), but there's a 20% slowdown with 8 threads.

Now that this is fixed via #5566, speeds on "bull" are almost same as before both of these PRs - about 2% faster for 1 thread, but about 3% slower for 8 threads.

solardiz:
> about 3% slower for 8 threads.

... and e.g. `-tune=16` doesn't help much, so it's not an effect of reducing MKPC and OMP_SCALE to 1. But on the bright side, it's not always 3%, but sometimes just 1% slower than before these PRs.

magnumripper:
I'm puzzled there are any difference at all considering there are no small allocations/frees at all but a single one per [relatively very slow] hash calculation, of eg. the full 64 MB for m=65536. Maybe it gets more relevant for huge m and/or with NUMA, in case alloc_region can somehow allocate "better"?

solardiz:
> I'm puzzled there are any difference at all considering there are no small allocations/frees at all but a single one per [relatively very slow] hash calculation

You mean any difference between reusing allocations (like before this PR or after #5566) vs. redoing them (like we had with this PR)? That's no surprise to me at all. When libc optimizes a large `malloc` into `mmap`, this also generally means that its `free` would be `munmap`. And the kernel has to zeroize memory before handing it over to userspace - for security, as well as because it's defined that memory is all-zeros on allocation from the kernel. This doesn't mean the kernel has to `memset` the memory before return from `mmap` - it may instead allocate pages lazily on access (page fault) and `memset` them at that time. But as far as userspace is concerned, the memory must read as zero. Even if the userspace itself starts by writing data into the entire page before ever reading anything (so zeroization wouldn't matter for the computation result), when a page fault is triggered it's only for a smaller write (one SIMD vector at most) and the kernel has to zeroize the entire page before handing it over to userspace (as it cannot reliably predict there would be no reads).

In some cases where we see almost no slowdown, libc may have preserved the allocation from the kernel. This happens for small allocations, although I'd have expected 4 MiB not to be small enough. Another possibility is that the zeroization was happening in the kernel between allocations or page accesses by another otherwise-idle CPU core, so when pages were to be handed over to userspace the kernel always readily had a pre-zeroized one.

PR #5563, opened 2024-11-05
CI: Add gcc-14 and clang-18/clang-17/clang-16 jobs

ldv-alt:
I also took the liberty of updating x86 jobs to use the latest gcc, hope that's OK.


solardiz (approved):
Thank you very much @ldv-alt! This looks good to me.

PR #5564, opened 2024-11-05
Added bitlocker2john.py

holly-o:
Added Python script to extract hashes for BitLocker-encrypted volumes

### DIFF
--- a/doc/README.BitLocker
+++ b/doc/README.BitLocker
@@ -5,9 +5,11 @@ Step 1: Extract the hash
 ------------------------
 
 In order to use the BitLocker-OpenCL format, you must produce a well-formatted
-hash from your BitLocker encrypted image. Use the bitlocker2john tool to
-extract hashes from password protected BitLocker encrypted volumes. It returns
-four output hashes with different prefixes:
+hash from your BitLocker encrypted image. Use the tool run/bitlocker2john.py to
+extract hashes from password protected BitLocker encrypted volumes.
+Usage: python3 bitlocker2john.py <bitlocker_image> [-o <bitlocker_partition_offset>]
+
+It returns four output hashes with different prefixes:
 
 * If the device was encrypted using the User Password authentication method,
   bitlocker2john prints these two hashes:
@@ -19,6 +21,40 @@ four output hashes with different prefixes:
   * $bitlocker$3$... : it starts the Recovery Password attack mode with MAC verification (slower execution, no false positives)
 
 Hash extraction example,
+
+$ python3 bitlocker2john.py bitlocker_image -o 4194304
+
+[+] BitLocker signature found: -FVE-FS-
+[+] Identified volume GUID: 4967D63B-2E29-4AD8-8399-F6A339E3D001 = BitLocker
+[+] FVE metadata info found at offsets ['0x2500000', '0x42500000', '0x82500000']
+
+Parsing FVE block...
+
+Parsing FVE metadata header...
+Metadata size: 820
+Volume GUID: F23E2E52-12F0-4066-94B6-F91C9A1A1D91
+Encryption method: 0x80028002
+
+Parsing FVE metadata entry...
+Entry size: 68
+Entry type: 0x7 = Computer description
+Value type: 0x2 = UTF-16 string
+
+Parsing description...
+Info: DESKTOP-J5HJVN9 E: 12/08/2022
+
+...
+
+The following hashes were found:
+$bitlocker$0$16$4a67bc123abedc43d60b3ece78ec6d1e$1048558$12$a015f77b68aed80103000000$60$2dbacf4710d3d42aa4f7baeedff85d72fc892f8f3457271901c0d2eccc3de890f081b3335740a5b5f1473892569ec0455d1aa2fd0075ac073a5f7b2a
+$bitlocker$1$16$4a67bc123abedc43d60b3ece78ec6d1e$1048558$12$a015f77b68aed80103000000$60$2dbacf4710d3d42aa4f7baeedff85d72fc892f8f3457271901c0d2eccc3de890f081b3335740a5b5f1473892569ec0455d1aa2fd0075ac073a5f7b2a
+$bitlocker$2$16$4b10ca85ab17a7419990d92f75abc848$1048558$12$a015f77b68aed80106000000$60$11e39cfd4dc9f647cef46b843347a3677c0706d3653f3477d44c72c8e36e8e02e010744dc384a419ff487a0190b42da0a29229d57a0bc3c6a7193f7
+$bitlocker$3$16$4b10ca85ab17a7419990d92f75abc848$1048558$12$a015f77b68aed80106000000$60$11e39cfd4dc9f647cef46b843347a3677c0706d3653f3477d44c72c8e36e8e02e010744dc384a419ff487a0190b42da0a29229d57a0bc3c6a7193f7
+
+
+Alternatively, run the compiled program bitlocker2john, which is normally
+built from source along with the rest of John the Ripper and is included
+pre-built in John the Ripper binary releases.
 
 $ ../run/bitlocker2john minimalistic.raw  # operate on a disk image
 Signature found at 0x00010003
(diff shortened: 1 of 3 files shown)


solardiz:
Thank you @holly-o.

@exploide You might want to help us review this, although since it's a new script we can as well merge first - it shouldn't hurt.

run/bitlocker2john.py
```diff
@@ -0,0 +1,237 @@
...
+    print("Salt:", salt.hex())
+    current_pos = 0
+    aes_ccm_data = data[20:]
+    current_pos, data, value_type = parse_fve_metadata_entry(current_pos, aes_ccm_data)
+    nonce, mac, enc_data = parse_aes_ccm_encrypted_key(data)
+
+    return salt, nonce, mac, enc_data
+
+def generate_hashcat_hash(salt, nonce, mac, enc_data, protection_type):
+    print("\nFound hashcat hash!")
```
solardiz: I understand this script was first submitted to the hashcat project, which is why it focuses on usage with hashcat and that's fine, but at least the printed messages (and ideally also code comments) shouldn't give the impression that the hashes are only usable with hashcat.

run/bitlocker2john.py
```diff
@@ -0,0 +1,237 @@
...
+    if protection_type == 0x2000:
+        versions = BITLOCKER_PASSWORD_HASH_VERSIONS
+    if protection_type == 0x800:
+        versions = BITLOCKER_RECOVERY_HASH_VERSIONS
```
solardiz, line 88: The comment at start of script says this won't extract recovery key VMKs, but then we do have this in code - what exactly does that mean? Will the script actually extract such data, or is this functionality in some way incomplete?

solardiz (changes requested):
The script is missing a shebang line. Is it Python 3 only or is it also compatible with Python 2? Depending on this, please start it with either `#!/usr/bin/env python3` or `#!/usr/bin/env python`. For this sort of updates to the main script within this PR, please amend the one existing commit and force-push.

You could also want to update `doc/README.BitLocker` to mention this script as (the recommended?) alternative to our compiled `bitlocker2john` program (making that program a secondary choice?)

doc/README.BitLocker
```diff
@@ -20,6 +22,38 @@ four output hashes with different prefixes:
...
+...
+
+The following hashes were found:
+$bitlocker$0$16$4a67bc123abedc43d60b3ece78ec6d1e$1048558$12$a015f77b68aed80103000000$60$2dbacf4710d3d42aa4f7baeedff85d72fc892f8f3457271901c0d2eccc3de890f081b3335740a5b5f1473892569ec0455d1aa2fd0075ac073a5f7b2a
+$bitlocker$1$16$4a67bc123abedc43d60b3ece78ec6d1e$1048558$12$a015f77b68aed80103000000$60$2dbacf4710d3d42aa4f7baeedff85d72fc892f8f3457271901c0d2eccc3de890f081b3335740a5b5f1473892569ec0455d1aa2fd0075ac073a5f7b2a
+$bitlocker$2$16$4b10ca85ab17a7419990d92f75abc848$1048558$12$a015f77b68aed80106000000$60$11e39cfd4dc9f647cef46b843347a3677c0706d3653f3477d44c72c8e36e8e02e010744dc384a419ff487a0190b42da0a29229d57a0bc3c6a7193f7
+$bitlocker$3$16$4b10ca85ab17a7419990d92f75abc848$1048558$12$a015f77b68aed80106000000$60$11e39cfd4dc9f647cef46b843347a3677c0706d3653f3477d44c72c8e36e8e02e010744dc384a419ff487a0190b42da0a29229d57a0bc3c6a7193f7
+
+
+Alternatively, run compiled script bitlocker2john. Found at tools/john/src/bitlocker2john.c
```
solardiz: The C program isn't a `script` and we do not have a `tools/john` directory and people are not expected to compile just one C file manually. So I'd replace this line with:

```
Alternatively, run the compiled program bitlocker2john, which is normally
built from source along with the rest of John the Ripper and is included
pre-built in John the Ripper binary releases.
```

run/bitlocker2john.py
```diff
@@ -0,0 +1,240 @@
+#!/usr/bin/python3
```
solardiz: Please use the `env` trick here for consistency with our other scripts and not to rely on specific location of `python3`. So the line should be:

```
#!/usr/bin/env python3
```

solardiz (changes requested):
Thank you for these updates, they mostly look good. I suggest a couple of minor edits, can you please incorporate those as well?

For such fixups within a PR, we normally amend the existing commit and force-push, not add more commits. But if you're not used to work like that (or if the git client software you use doesn't let you), it's OK if you add commits (I will then need to temporarily allow squash-and-merge for this repo and use that on this specific PR).

Thanks again!

solardiz:
This failed our CI whitespace-errors check, please also fix this minor detail:

```
Run git diff-index --check --cached b1b622f691d40196815939e4736a5da71befd206
doc/README.BitLocker:9: trailing whitespace.
+extract hashes from password protected BitLocker encrypted volumes. 
run/bitlocker2john.py:4: trailing whitespace.
+# Supported modes: 
run/bitlocker2john.py:100: trailing whitespace.
+    print("\nParsing description...")    
run/bitlocker2john.py:105: trailing whitespace.
+    print("\nParsing volume header block...")        
run/bitlocker2john.py:150: trailing whitespace.
+        
run/bitlocker2john.py:158: trailing whitespace.
+    entry_size = uint_to_int(block[0:2]) 
run/bitlocker2john.py:160: trailing whitespace.
+    value_type = uint_to_int(block[4:6]) 
run/bitlocker2john.py:219: trailing whitespace.
+        
run/bitlocker2john.py:226: trailing whitespace.
+        
```

solardiz:
Oh, we also need to add a `doc/NEWS` entry mentioning this contribution.

solardiz (approved):

doc/README.BitLocker
```diff
@@ -20,6 +22,40 @@ four output hashes with different prefixes:
...
+Parsing description...
+Info: DESKTOP-J5HJVN9 E: 12/08/2022
+
+...
+
+The following hashes were found:
+$bitlocker$0$16$4a67bc123abedc43d60b3ece78ec6d1e$1048558$12$a015f77b68aed80103000000$60$2dbacf4710d3d42aa4f7baeedff85d72fc892f8f3457271901c0d2eccc3de890f081b3335740a5b5f1473892569ec0455d1aa2fd0075ac073a5f7b2a
+$bitlocker$1$16$4a67bc123abedc43d60b3ece78ec6d1e$1048558$12$a015f77b68aed80103000000$60$2dbacf4710d3d42aa4f7baeedff85d72fc892f8f3457271901c0d2eccc3de890f081b3335740a5b5f1473892569ec0455d1aa2fd0075ac073a5f7b2a
+$bitlocker$2$16$4b10ca85ab17a7419990d92f75abc848$1048558$12$a015f77b68aed80106000000$60$11e39cfd4dc9f647cef46b843347a3677c0706d3653f3477d44c72c8e36e8e02e010744dc384a419ff487a0190b42da0a29229d57a0bc3c6a7193f7
+$bitlocker$3$16$4b10ca85ab17a7419990d92f75abc848$1048558$12$a015f77b68aed80106000000$60$11e39cfd4dc9f647cef46b843347a3677c0706d3653f3477d44c72c8e36e8e02e010744dc384a419ff487a0190b42da0a29229d57a0bc3c6a7193f7
```
solardiz, line 52: I've just merged this PR (thanks!) but now I notice that the two example Recovery Password "hashes" given here are somehow one character too short - last field is 119 instead of 120 chars. Do we maybe have a bug where a leading zero is omitted? In my testing of the script, I got all strings of the same length, but maybe I just didn't trigger that bug.

solardiz: I'm not seeing such bug in the current script, so maybe it was in some older revision? I think I'll "fix" these "hashes" by inserting a `0` after `$60$`, just so that they're a correct illustration. But I'd appreciate @holly-o's comments here.

exploide:
> @exploide You might want to help us review this, although since it's a new script we can as well merge first - it shouldn't hurt.

Sorry for being late. I had a quick look, no deep dive or actual testing but I think it looks good. Only some formatting things or minor idiomatic points. Noting that needs a fix now and it's probably fine being merged.

doc/README.BitLocker
holly-o: I edited one of my known hashes to add as an example, and accidentally removed a character. The script should always output len 120.
Thanks for your fix!

PR #5565, opened 2024-11-06
Removed divides in Argon2

### LINKED ISSUE #5385: Get rid of modulo division in Argon2
See https://github.com/P-H-C/phc-winner-argon2/pull/306

We need this for both CPU and OpenCL formats. For CPU, perhaps do it after #2738.

Sc00bz:
Fixes #5385

### DIFF
--- a/src/argon2_opt_plug.c
+++ b/src/argon2_opt_plug.c
@@ -183,7 +183,7 @@ void argon2_fill_segment(const argon2_instance_t *instance,
     block address_block, input_block;
     uint64_t pseudo_rand, ref_index, ref_lane;
     uint32_t prev_offset, curr_offset;
-    uint32_t starting_index, i, i_address;
+    uint32_t starting_index, i;
     uint32_t lanes_reciprocal = 0;
     uint32_t lanes = instance->lanes;
 #if defined(__AVX512F__)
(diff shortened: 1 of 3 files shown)


solardiz:
Thank you, @Sc00bz! I wonder if this makes a measurable performance difference - have you checked?

For the commit message, can you please prefix it with `Argon2: `? Amend and force-push.

Meanwhile, I've just let the CI jobs run on this PR.

src/argon2_opt_plug.c
```diff
@@ -228,36 +230,47 @@ void argon2_fill_segment(const argon2_instance_t *instance,
...
-        if (curr_offset % instance->lane_length == 1) {
-            prev_offset = curr_offset - 1;
-        }
-
+         ++i, ++curr_offset) {
         /* 1.2 Computing the index of the reference block */
         /* 1.2.1 Taking pseudo-random value from the previous block */
         if (data_independent_addressing) {
-            if (i % ARGON2_ADDRESSES_IN_BLOCK == 0) {
+            if (i_address == 0) {
```
solardiz: I'm not sure we want this set of changes for `% ARGON2_ADDRESSES_IN_BLOCK`, which was likely just an AND. I suggest looking at the generated code with/without these changes.

solardiz:
> Thank you, @Sc00bz! I wonder if this makes a measurable performance difference - have you checked?

We've just merged #5566, which will affect performance a lot. Ideally, the changes here would be benchmarked relative to and on top of that newer code.

Sc00bz:
Fixed commit messages and undid the `i % ARGON2_ADDRESSES_IN_BLOCK` -> `i_address` change. I was not thinking.

solardiz (approved):
Thanks. I'm going to merge this, then experiment with it and maybe revise further.

solardiz:
With this PR's changes as merged, there's slight increase in compiled code size (which is expected) and maybe a 1% or even 2% speedup at 1 thread on the old AMD FX-8120, same speed at 8 threads (where we probably bump into memory bandwidth anyway, although I didn't do the math to confirm). I'm actually a bit surprised there's measurable speedup - I don't know if it's a genuine effect of these changes or just some random fluctuation in how the old vs. new `john` binary is cached in memory (order of pages could matter, especially when cache associativity is low). But at least I saw no regression yet.

PR #5566, opened 2024-11-07
Argon2 format: Custom malloc/free using callbacks

### LINKED ISSUE #5558: Argon2: Reintroduce the memory (de)allocation optimization
With #5557, we switched from preserving and reusing Argon2's memory allocation (unless it needs to increase in size) to allocating and freeing for every hash computed. We need to reintroduce our optimization.

As I wrote in https://github.com/openwall/john/issues/2738#issuecomment-343739923:

Regarding "use of preallocated memory", I realized this can be implemented on top of current upstream code as-is using its memory (de)allocation callbacks, which we'll override. In our deallocator, we'll save the pointer and size and mark them as free in our own per-thread variables instead of calling `free()`. In our allocator, we'll check if the size is same as or smaller than the saved size and if our previous allocation is marked free (in our own Boolean variable) and if so reuse the old allocation, or `realloc()` as necessary, or error out if the previous allocation is not marked free (shouldn't happen). Better yet, we should reuse (y)escrypt's `alloc_region()` in there, simplifying some of those checks (`alloc_region()` already has similar checks) and taking advantage of its explicit huge pages support. We'd only call `free_region()` from the format's `done()`.

magnumripper:
We never free memory except in format's done(), instead we reuse the allocations per thread, and realloc if needed.

We also got rid of a small malloc/free per call for a temporary buffer, by moving away from argon2_hash() in favor of argon2_ctx().

Closes #5558

### DIFF
--- a/src/argon2_fmt_plug.c
+++ b/src/argon2_fmt_plug.c
@@ -62,9 +62,9 @@ john_register_one(&fmt_argon2);
 #define SALT_ALIGN              sizeof(uint32_t)
 
 #define MIN_KEYS_PER_CRYPT      1
-#define MAX_KEYS_PER_CRYPT      2
+#define MAX_KEYS_PER_CRYPT      1
 
-#define OMP_SCALE               8 // tuned w/ MKPC for core i7m
+#define OMP_SCALE               1
 
 #ifdef _OPENMP
 #define THREAD_NUMBER omp_get_thread_num()
--- a/src/argon2_fmt_plug.c
+++ b/src/argon2_fmt_plug.c
@@ -65,6 +65,14 @@ john_register_one(&fmt_argon2);
 
 #define OMP_SCALE               8 // tuned w/ MKPC for core i7m
 
+#ifdef _OPENMP
+#define THREAD_NUMBER omp_get_thread_num()
+#define NUM_THREADS   omp_get_max_threads()
+#else
+#define THREAD_NUMBER 0
+#define NUM_THREADS   1
+#endif
+
 /*
  * Argon2 is a KDF, using a hash function built upon Blake2b (Blake2b_long,
  * variable length and capable of generating up to 2^32 bytes digests).
@@ -116,7 +124,14 @@ struct argon2_salt {
 	argon2_type type;
 };
 
+struct argon2_memory {
+	int used;
+	size_t size;
+	void *ptr;
+};
+
 static struct argon2_salt saved_salt;
+static struct argon2_memory *thread_mem;
 
 static char (*saved_key)[PLAINTEXT_LENGTH + 1];
 static int *saved_len;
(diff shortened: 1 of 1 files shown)


src/argon2_fmt_plug.c
```diff
@@ -245,6 +266,35 @@ static void set_salt(void *salt)
...
+
+static void deallocate(uint8_t *memory, size_t size)
+{
+	if (!thread_mem[THREAD_NUMBER].used)
+		error_msg("%s(): thread %u: Freed memory not in use\n", __FUNCTION__, THREAD_NUMBER);
+	if (thread_mem[THREAD_NUMBER].size < size)
+		error_msg("%s(): thread %u: incorrect size %zu, was %zu\n", __FUNCTION__, THREAD_NUMBER, size, thread_mem[THREAD_NUMBER].size);
+
+	thread_mem[THREAD_NUMBER].used = 0;
+}
```
magnumripper: Not sure why they supply a size to `deallocate()` but I used it for an extra check, bc why not.

magnumripper:
Well that crashed and burned. I'll be trying to understand what ASan wants to explain to me.

Edit: OK that was a fencepost error for non-OMP. Thread was hard coded as 1 instead of 0.

magnumripper:
@solardiz if this PR is correct, I can go ahead and switch to `alloc_region_t()` instead - either as a separate commit or as a fixup here.

src/argon2_fmt_plug.c
solardiz: Maybe can even insist on `== size` in that check?

src/argon2_fmt_plug.c
```diff
@@ -253,14 +303,30 @@ static int crypt_all(int *pcount, struct db_salt *salt)
...
+		context.ad = NULL;
+		context.adlen = 0;
+		context.t_cost = saved_salt.t_cost;
+		context.m_cost = saved_salt.m_cost;
+		context.lanes = saved_salt.lanes;
+		context.threads = saved_salt.lanes;
+		context.allocate_cbk = &allocate;
+		context.free_cbk = &deallocate;
+		context.flags = ARGON2_DEFAULT_FLAGS;
+		context.version = ARGON2_VERSION_NUMBER;
```
solardiz: Maybe use struct initialized with named fields here, which is a C feature that we already use in the tree anyway (and for a related purpose). Like this:

```c
argon2_context context = {
        .out = (uint8_t*)crypted[i],
and so on
};
```

omitting the fields where we initialize to NULL or 0. That way, we'll be sure to initialize the entire struct even if more fields are added later.

Our existing usage is e.g.:

```c
yescrypt_params_t params = {.N = cur_salt->N, .r = cur_salt->r, .p = cur_salt->p};
```

solardiz (approved):
Looks good to me. I've made a couple of suggestions for a fixup of the existing commit.

Please do also switch to using the "regions" functions, but in a separate commit. Thank you!

src/argon2_fmt_plug.c
magnumripper: No, it might be smaller after we reused the existing alloc for a smaller request.

solardiz:
The switch to regions looks good to me, except that our `size` field became redundant with what's inside the region struct - maybe switch to reusing the latter?

solardiz:
e.g. yescrypt does:

```c
                if (local->aligned_size < need) {
                        if (free_region(local))
                                return -1;
                        if (!alloc_region(local, need))
                                return -1;
                }
```

without needing to maintain a separate size field.

solardiz:
and in the Armory format I did:

```c
        lut_item *lut = memory->aligned;
        size_t bytes_reqd = (size_t)n * sizeof(*lut);
        if (!lut || memory->aligned_size < bytes_reqd) {
                free_region_t(memory);
                if (!(lut = alloc_region_t(memory, bytes_reqd)))
                        return -1;
        }
```

magnumripper:
> The switch to regions looks good to me, except that our `size` field became redundant with what's inside the region struct - maybe switch to reusing the latter?

Ah, yes. The fact we can hand out an oversized existing alloc had me confused but right now they are indeed always the same value.

I can throw in a commit resetting MKPC and OMP_SCALE to 1 as well, while at it.

solardiz:
> I can throw in a commit resetting MKPC and OMP_SCALE to 1 as well, while at it.

That's OK with me.

solardiz:
We were ignoring possible error returns from Argon2 code before this PR and made it even worse with this PR. Memory allocation errors go undetected, resulting in "high speeds". I'm going to fix that.

Issue #5524, opened 2024-08-15
Allow --dupe-suppression in batch mode


solardiz:
Dupe suppression is enabled by default in batch mode (because it includes wordlist+rules), but the command-line option to control it is not allowed (technically because no other option specifies the use of rules). Ideally, we'd allow this option to make it easier to explicitly disable the dupe suppressor or adjust its memory usage. Right now, the only ways to do that are editing the configuration file or overriding the configuration file name.

This may be a bit tricky to implement (which is why it hasn't been done right away).

solardiz:
This was tricky to figure out, but looks easy if we accept the side-effects as desirable:

```diff
commit cd8d511847dca14d0cbadabc2d6587ae99cf5352 (HEAD -> fixes-20240923)
Author: Solar <solar@openwall.com>
Date:   Mon Sep 23 02:50:30 2024 +0200

    Options: Allow --rules and/or --dupe-suppression in batch mode
    
    Fixes #5524

diff --git a/src/options.h b/src/options.h
index d0a71b01e..09d82605e 100644
--- a/src/options.h
+++ b/src/options.h
@@ -69,7 +69,7 @@
        (FLG_EXTERNAL_CHK | FLG_ACTION | FLG_CRACKING_SUP | FLG_PWD_SUP)
 /* Batch cracker */
 #define FLG_BATCH_CHK                  0x0000000000004000ULL
-#define FLG_BATCH_SET                  (FLG_BATCH_CHK | FLG_CRACKING_SET)
+#define FLG_BATCH_SET                  (FLG_BATCH_CHK | FLG_CRACKING_SET | FLG_RULES_ALLOW)
 /* Stdout mode */
 #define FLG_STDOUT                     0x0000000000008000ULL
 /* Restoring an interrupted session */
```

This allows not only `--dupe-suppression`, but also `--rules` with and without parameter (in the latter case, it's a no-op since our default rules are enabled in batch mode by default) and even `--rules-stack`.

However, specifying only `--rules-stack` (without `--rules`) somehow results in weird behavior where the process almost locks up or runs very slowly once it reaches pass 2 (wordlist + rules), but only with some rulesets (e.g., with `--rules-stack=o1`).

magnumripper:
Are you saying it somehow doesn't behave like `-w -ru -rules-stack=o1`?

solardiz:
> Are you saying it somehow doesn't behave like `-w -ru -rules-stack=o1`?

[...]

PR #5569, opened 2024-11-07
Wordlist: Move stacked rules logging to after it's initialized

magnumripper:
It wasn't reporting stacked rules correcly.  See #5524


solardiz (approved):
Note that this is affected by possible zeroing of `do_lmloop` just above the new location of this code. I hope this is as intended.

magnumripper:
Yes, I think that's just fine. Let me test this a bit before we merge though.

magnumripper:
I couldn't see any regression, so merging.

PR #5574, opened 2024-11-12
Keepass KDBX4 and Argon2 support

magnumripper:
Including KDBX4 updates to keepass2john.

Keepass-opencl now supports KDBX4 but only with AES. A new separate keepass-argon2-opencl format now lives within the argon2-opencl source file.

Keepass default costs are now hard. The new KeePass test vectors are made with lower-than-default costs (even for KDBX4/AES, which now has to AES-encrypt about a gigabyte with default settings).

### DIFF
--- a/src/keepass_fmt_plug.c
+++ b/src/keepass_fmt_plug.c
@@ -9,10 +9,10 @@
  * added by Fist0urs <eddy.maaalou at gmail.com>
  *
  * This software is
- * Copyright (c) 2012 Dhiru Kholia <dhiru.kholia at gmail.com>,
- * Copyright (c) 2014 m3g9tr0n (Spiros Fraganastasis),
+ * Copyright (c) 2017-2024 magnum,
  * Copyright (c) 2016 Fist0urs <eddy.maaalou at gmail.com>, and
- * Copyright (c) 2017 magnum,
+ * Copyright (c) 2014 m3g9tr0n (Spiros Fraganastasis),
+ * Copyright (c) 2012 Dhiru Kholia <dhiru.kholia at gmail.com>,
  * and it is hereby released to the general public under the following terms:
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted.
@@ -29,6 +29,11 @@ john_register_one(&fmt_KeePass);
 
 #ifdef _OPENMP
 #include <omp.h>
+#define THREAD_NUMBER omp_get_thread_num()
+#define NUM_THREADS   omp_get_max_threads()
+#else
+#define THREAD_NUMBER 0
+#define NUM_THREADS   1
 #endif
 
 #include "arch.h"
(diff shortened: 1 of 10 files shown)


magnumripper:
Somehow when I load a bunch of keepass samples without specifying a format...
```
Warning: detected hash type "KeePass", but the string is also recognized as "KeePass-opencl"
Use the "--format=KeePass-opencl" option to force loading these as that type instead
```
...it doesn't suggest also trying keepass-argon2-opencl. I assume this problem is unrelated to this PR?

Edit: If I specify `--format=opencl`, I get the other variant of the warning, and it's correct. Not sure how they differ in logic.
```
Warning: only loading hashes of type "KeePass-opencl", but also saw type "KeePass-Argon2-opencl"
Use the "--format=KeePass-Argon2-opencl" option to force loading hashes of that type instead
```

ghost:
>  problem is unrelated to this PR?

The most recent one was added on https://github.com/openwall/john/commit/6dcfceb88e3b50d551e0f510ce75236b24af9e15

Both, side by side: https://github.com/openwall/john/commit/047ab688b8c224cfafb67e48dd00533bbf3fbbbb

magnumripper:
"keepass-opencl" can only do AES-based hashes. "keepass-argon2-opencl" can only do Argon2-based ones. The "Keepass" CPU format can do both.

The `string is also recognized` sounds like it's describing a particular AES based hash, and not mentioning keepass-argon2-opencl is indeed correct - for that very hash. Not sure why later lines with Argon2-based hashes doesn't emit a note about the other format.

The `only loading hashes of type` indicates a wider view of different hashes in the input file (and it does report correctly). It's unclear to me what dictates which of the messages we get.

EDIT: I opened #5577 for this

src/keepass_fmt_plug.c
```diff
@@ -37,115 +42,240 @@ john_register_one(&fmt_KeePass);
...
+#elif !defined (JOHN_NO_SIMD) && defined(__SSE2__)
+#define ALGORITHM_NAME          "AES/Argon2 128/128 SSE2"
+#else
+#define ALGORITHM_NAME          "AES/Argon2 32/" ARCH_BITS_STR
+#endif
+
+struct argon2_memory {
+	int used;
+	region_t region;
+};
```
solardiz, line 79: In the Argon2 format, I've since revised this struct to:

```c
struct argon2_memory {
        region_t region;
        int used;
        char padding[MEM_ALIGN_CACHE - sizeof(region_t) - sizeof(int)];
};
```

src/keepass_fmt_plug.c
```diff
@@ -37,115 +42,240 @@ john_register_one(&fmt_KeePass);
...
+			.allocate_cbk = &allocate,
+			.free_cbk = &deallocate,
+			.flags = ARGON2_DEFAULT_FLAGS,
+			.version = ARGON2_VERSION_13
+		};
+		error_code = argon2_ctx(&context, cur_salt->type);
+		if (error_code != ARGON2_OK) {
+			ret = -1;
+			fprintf(stderr, "Keepass: Error: Argon2 failed: %s\n",
+			        argon2_error_message(error_code));
```
solardiz, line 247: Other error messages in this source file (in the allocation functions) start with `Error: KeePass: `, so let's make this one similar.

src/keepass_fmt_plug.c
```diff
@@ -158,103 +288,135 @@ static int crypt_all(int *pcount, struct db_salt *salt)
...
+#pragma omp atomic
+#endif
+						any_cracked |= 1;
+					}
+				}
+				MEM_FREE(decrypted_content);
+			} else {
+				// KeePass version 2 with Twofish is TODO. Twofish support under KeePass version 2
+				// requires a third-party plugin. See http://keepass.info/plugins.html for details.
+				error_msg("KeePass v2 w/ Twofish not supported yet\n");
```
solardiz: Do we really detect this only so late? It's not great to load hashes and only then have all OpenMP threads exit with each printing this error.

src/keepass_fmt_plug.c
```diff
@@ -158,103 +288,135 @@ static int crypt_all(int *pcount, struct db_salt *salt)
...
-			MEM_FREE(decrypted_content);
-		} else {
-			// KeePass version 2 with Twofish is TODO. Twofish support under KeePass version 2
-			// requires a third-party plugin. See http://keepass.info/plugins.html for details.
-			error_msg("KeePass v2 w/ Twofish not supported yet");
 		}
 	}
+	if (failed) {
+#ifdef _OPENMP
+		fprintf(stderr, "Keepass: Error: Argon2 failed in some threads\n");
```
solardiz: Other error messages in this source file (in the allocation functions) start with `Error: KeePass: `, so let's make this one similar.

solardiz (changes requested):
This is complicated stuff, so I can't properly review without diving into it for real, but I've made some nitpick comments. Please also add this to `NEWS`.

BTW, I see we use the `WAIT_*` stuff here and note that we don't have that in the `argon2-opencl` format - we could want to add to there, too.

src/keepass_fmt_plug.c
magnumripper: Oh right, I saw that but forgot.

src/keepass_fmt_plug.c
magnumripper: I believe they are rejected in valid so this is just redundant. I'll verify that and remove this

magnumripper: Note to self: we do 4,x 1,0 2,0 2,2 1,1 in crypt_all() so not 1,2 & 2,1 and they're both rejected in valid

magnumripper:
> BTW, I see we use the `WAIT_*` stuff here and note that we don't have that in the `argon2-opencl` format - we could want to add to there, too.

I tried this out now but it ended up with massive performance regression - up to 50%. I have no idea why.
```diff
commit 43393ae3843986dc941311839e44f6a710325ed4
Author: magnum <magnumripper@hushmail.com>
Date:   Thu Nov 14 00:28:38 2024 +0100

    Argon2 OpenCL: Avoid nvidia busy-waits

diff --git a/src/opencl_argon2_fmt_plug.c b/src/opencl_argon2_fmt_plug.c
index dd537d690..bd6fe1612 100644
--- a/src/opencl_argon2_fmt_plug.c
+++ b/src/opencl_argon2_fmt_plug.c
@@ -183,7 +183,13 @@ static int run_kernel_on_gpu(uint32_t count)
 	cl_uint num_keys = MAX_KEYS_PER_CRYPT;
 	HANDLE_CLERROR(clSetKernelArg(pre_processing_kernel, 2, sizeof(buffer_row_pitch), &buffer_row_pitch), "Error setting pre-processing kernel argument");
 	HANDLE_CLERROR(clSetKernelArg(pre_processing_kernel, 3, sizeof(num_keys), &num_keys), "Error setting pre-processing kernel argument");
+
+	WAIT_INIT(*gws_pre_processing)
 	BENCH_CLERROR(clEnqueueNDRangeKernel(queue[gpu_id], pre_processing_kernel, 2, NULL, gws_pre_processing, lws_pre_processing, 0, NULL, NULL), "Run pre-processing kernel");
+	WAIT_SLEEP
+	BENCH_CLERROR(clFinish(queue[gpu_id]), "clFinish");
+	WAIT_UPDATE
+	WAIT_DONE
 
 	// Set parameters and execute kernel
 	assert(saved_salt.type >= 0 && saved_salt.type < (ARGON2_NUM_TYPES + 1) && kernels[Argon2_d] && kernels[Argon2_i]);
@@ -225,8 +231,14 @@ static int run_kernel_on_gpu(uint32_t count)
 			}
 			HANDLE_CLERROR(clSetKernelArg(kernels[selected_type], 4, sizeof(pass), &pass), "Error setting kernel argument");
 			HANDLE_CLERROR(clSetKernelArg(kernels[selected_type], 5, sizeof(slice), &slice), "Error setting kernel argument");
+
+			WAIT_INIT(*global_range)
 			BENCH_CLERROR(clEnqueueNDRangeKernel(queue[gpu_id], kernels[selected_type], 2, NULL, global_range, local_range, 0, NULL, NULL), "Run loop kernel");
-			HANDLE_CLERROR(clFlush(queue[gpu_id]), "clFlush");
+			WAIT_SLEEP
+			BENCH_CLERROR(clFinish(queue[gpu_id]), "clFinish");
+			WAIT_UPDATE
+			WAIT_DONE
+			opencl_process_event();
 		}
 	} else { // Argon2_d || Argon2_i
 		// Find the autotune params
@@ -253,8 +265,14 @@ static int run_kernel_on_gpu(uint32_t count)
 		for (slice = 0; slice < ARGON2_SYNC_POINTS; slice++) {
 			HANDLE_CLERROR(clSetKernelArg(kernels[argon2_type], 4, sizeof(pass), &pass), "Error setting kernel argument");
 			HANDLE_CLERROR(clSetKernelArg(kernels[argon2_type], 5, sizeof(slice), &slice), "Error setting kernel argument");
+
+			WAIT_INIT(*global_range)
 			BENCH_CLERROR(clEnqueueNDRangeKernel(queue[gpu_id], kernels[argon2_type], 2, NULL, global_range, local_range, 0, NULL, NULL), "Run loop kernel");
-			HANDLE_CLERROR(clFlush(queue[gpu_id]), "clFlush");
+			WAIT_SLEEP
+			BENCH_CLERROR(clFinish(queue[gpu_id]), "clFinish");
+			WAIT_UPDATE
+			WAIT_DONE
+			opencl_process_event();
 		}
 	}
 
```
BTW there's a mix of HANDLE_CLERROR and BENCH_CLERROR in argon2-opencl. Not sure it matters at all as it has its own autotune stuff.

solardiz:
Is this PR ready to merge?

> > BTW, I see we use the `WAIT_*` stuff here and note that we don't have that in the `argon2-opencl` format - we could want to add to there, too.
> 
> I tried this out now but it ended up with massive performance regression - up to 50%. I have no idea why.

OK, please open a separate issue for this. Thank you!

solardiz (approved):

solardiz:
Perhaps unrelated to the changes here, but I think it's bad that we mix different costs in default benchmark:

```
Benchmarking: KeePass [AES/Argon2 512/512 AVX512F]... (4xOMP) DONE
Speed for cost 1 (t (rounds)) of 50000 and 6000, cost 2 (m) of 0, cost 3 (p) of 0, cost 4 (KDF [0=Argon2d 2=Argon2id 3=AES]) of 3
Raw:	888 c/s real, 236 c/s virtual
```

I think we should choose the first 2 test vectors such that costs are the same.

solardiz:
More related to this PR, `KeePass-Argon2-opencl` default benchmark also uses a mix of costs, and it also should not:

```
Device 1: GeForce GTX 570
Benchmarking: KeePass-Argon2-opencl [BlaMka OpenCL]... |
Trying to compute 90 hashes at a time using 360 of 1279 MiB device memory
LWS=[32-64] GWS=[2880-5760] ([45-180] blocks) => Mode: LOCAL_MEMORY
DONE
Speed for cost 1 (t) of 3 and 2, cost 2 (m) of 4096 and 1024, cost 3 (p) of 1 and 2, cost 4 (KDF [0=Argon2d 2=Argon2id]) of 0 and 2
Raw:    2430 c/s real, 2417 c/s virtual
```

(and yes, it works on this ancient GPU)

solardiz:
After these changes, the size of `.bss` in the two KeePass format files increases to 16+ MiB each, so 32 MIB for `john` as a whole. That's not great.

I think it's because `keepass_get_salt` has a static salt struct, which contains a `KEEPASS_MAX_CONTENT_SIZE` field.

solardiz:
FWIW, while our `argon2-opencl` format works fine on the ancient AMD Tahiti, the `keepass-opencl` and `keepass-argon2-opencl` formats fail:

```
[solar@well src]$ ../run/john -te -form=keepass-opencl -dev=4
Device 4: Tahiti [AMD Radeon HD 7900 Series]
Benchmarking: KeePass-opencl [AES OpenCL]... FAILED (cmp_all(20))
[solar@well src]$ ../run/john -te -form=keepass-argon2-opencl -dev=4
Device 4: Tahiti [AMD Radeon HD 7900 Series]
Benchmarking: KeePass-Argon2-opencl [BlaMka OpenCL]... FAILED (cmp_one(1))
[solar@well src]$ ../run/john -te -form=argon2-opencl -dev=4
Device 4: Tahiti [AMD Radeon HD 7900 Series]
Benchmarking: argon2-opencl, Argon2 [BlaMka OpenCL]... |
Trying to compute 128 hashes at a time using 2048 of 3014 MiB device memory
LWS=[192-256] GWS=[4096-12288] ([16-64] blocks) => Mode: LOCAL_MEMORY
DONE
Speed for cost 1 (t) of 3, cost 2 (m) of 4096, cost 3 (p) of 1, cost 4 (type [0:Argon2d 1:Argon2i 2:Argon2id]) of 0 and 1
Raw:    1267 c/s real, 64000 c/s virtual
```

magnumripper:
> After these changes, the size of `.bss` in the two KeePass format files increases to 16+ MiB each, so 32 MIB for `john` as a whole. That's not great.
> 
> I think it's because `keepass_get_salt` has a static salt struct, which contains a `KEEPASS_MAX_CONTENT_SIZE` field.

I guess we could lower that size to 64K or even lower for now (it will warn correctly in valid if exceeding it, and besides, who has a 16 MB password database?!) but the real fix is to use dynamic salt.

BTW all three keepass formats should be affected, so 48 MiB I guess...

magnumripper:
> who has a 16 MB password database

OK it comes from https://github.com/openwall/john/issues/3169#issuecomment-369235965 where the OP needed 1.5 MB but we used a value with significant margin. This again suggests we should use dynamic salt. I'll open an issue.

PR #5578, opened 2024-11-13
RAR5 formats: Canonicalize ciphertext for pot entries

magnumripper:
The hashes come with an IV field but we don't use it. Without this canonicalization (we set it to all zeros), john can't recognize same hash with different IV from the pot file as "same".


solardiz (approved):
I didn't verify `CIPHERTEXT_LENGTH`. ;-)

PR #5580, opened 2024-11-14
Update our Unicode and legacy codepage support to Unicode 16.0.0

### LINKED ISSUE #5537: Upgrade to new Unicode spec 16.0.0+
Our last update seems here:

```
commit 74e896be8d7dd59c744eda3c6a6bc22ab9d29e15
Date:   Tue Aug 25 16:03:26 2020 +0200

    Upgrade Unicode stuff from version 11.0.0 to 13.0.0.
```

Current is apparently 16.0.0: https://x.com/thorsheim/status/1835612002237972490

```
Per Thorsheim
@thorsheim
Unicode 16.0.0 adds 5185 new symbols, including 7 complete alphabets from India, Albania and West Africa. Unicoe 16.0.0 is now 154998 symbols in total.

Can't wait to see what @CrackMeIfYouCan will do with this next year! :D

@hashcat @solardiz @yiannistox
```

magnumripper:
We missed a couple of releases so this adds 11,131 new characters compared to our previous version 13.0.0.  These are mostly obscure alphabets and emojis (I'm just guessing) that are not super likely to be included in any password.

There were also a few minor changes to definitions of character classes such as punctuation and specials - we always used the Unicode Database's definition of them and apparently they changed their mind about a few.

Closes #5537


magnumripper:
> There were also a few minor changes to definitions of character classes

The build bots seem upset with something - maybe because of this?

solardiz:
We're now testing that all external modes produce the same output as before. As I understand, this PR is supposed to change that, so the hash should be updated accordingly.

ghost:
Pre-merge, as seen at `src/tests/test_externals.md5`:

```
$ JOHN=./john $HOME/test_externals.sh | md5sum 
[...]
69e3af53cbe9ab14095b3a57c4a2e7f3  -
```
The tool is useful for testing the “before” and “after”.
As some, such as `datetime` were not impacted, you can use this information to test the PR.

Or simply run the command above with the new source code and update the hash.

solardiz:
@magnumripper Thank you! I am puzzled by the changes to 8-bit charsets in `encoding_data.h` - why would they change between Unicode versions?

Other than that, this looks good to me. So please update the expected external modes hash, and this is good to go.

solardiz:
> I am puzzled by the changes to 8-bit charsets in `encoding_data.h` - why would they change between Unicode versions?

OK, I guess this is answered by:

> There were also a few minor changes to definitions of character classes such as punctuation and specials - we always used the Unicode Database's definition of them and apparently they changed their mind about a few.

PR #5585, opened 2024-11-26
Correct the names of shared region functions.

### LINKED ISSUE #5571: Clean up the memory regions API
We now have two copies of yescrypt-derived init/alloc/free region code - one in the yescrypt tree, and one copied into `memory.c`. The latter uses weird function names with `_t` in them, named after the type. We should at least rename the latter functions, or ideally switch to using shared functions.

magnumripper:
The type is region_t but the functions are init/alloc/free_region().

Closes #5571 but copies of the functions still remain in yescrypt directory, not sure if they are identical or not.

### DIFF
--- a/src/argon2_fmt_plug.c
+++ b/src/argon2_fmt_plug.c
@@ -153,7 +153,7 @@ static void init(struct fmt_main *self)
 	thread_mem = mem_calloc(NUM_THREADS, sizeof(struct argon2_memory));
 
 	for (i = 0; i < NUM_THREADS; i++)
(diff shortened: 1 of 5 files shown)


solardiz (approved):
Looks good to me.

The copies under `yescrypt` are all `static`, so should not clash with these at link time, and not at compile time either because `yescrypt` files do not inlcude our `memory.*`, and our formats that use `yescrypt` only include `yescrypt.h` and not any of the `.c` files from there.

PR #5587, opened 2024-11-27
KeePass formats: Dynamic size salt

### LINKED ISSUE #5579: KeePass formats should use dynamic size salt
https://github.com/openwall/john/pull/5574#issuecomment-2475385160
_After these changes, the size of `.bss` in the two KeePass format files increases to 16+ MiB each, so 32 MIB for `john` as a whole. That's not great._

_I think it's because `keepass_get_salt` has a static salt struct, which contains a `KEEPASS_MAX_CONTENT_SIZE` field._

Current size comes from #3169

### LINKED ISSUE #5581: KeePass misaligned pointer
This is probably easy to fix:

```
Testing: iwork, Apple iWork '09 or newer [PBKDF2-SHA1 AES 256/256 AVX2 8x2]... PASS
Testing: KeePass [AES/Argon2 256/256 AVX2]... Warning: salt() returned misaligned pointer
PASS
Testing: keplr, Keplr Wallet (131072, 8, 1) [scrypt Salsa20/8 128/128 AVX, SHA256 32/64]... PASS
```

UBSAN + clang:` --test=0`

magnumripper:
Closes #5581
Closes #5579

### DIFF
--- a/run/opencl/keepass_kernel.cl
+++ b/run/opencl/keepass_kernel.cl
@@ -42,6 +42,7 @@ typedef enum {
 #define ARGON2_SALT_SIZE            64
 
 typedef struct {
+	dyna_salt dsalt;
 	uint32_t t_cost, m_cost, lanes;
 	uint32_t hash_size;
 	uint32_t salt_length;
@@ -52,7 +53,6 @@ typedef struct {
 	int kdbx_ver;
 	uint32_t kdf; // 0=AES, 1=Argon2
 	uint32_t cipher; // 0=AES, 1=TwoFish, 2=ChaCha
-	uint32_t key_transf_rounds;
 	uint8_t enc_iv[16];   // KDBX3 and earlier
 	union {
 		uint8_t final_randomseed[32]; // KDBX3 and earlier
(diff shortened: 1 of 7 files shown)


solardiz (approved):
I skimmed and didn't see anything certainly wrong. I'm not familiar with our dynamic salts.

I'm not sure the separation into `contents` vs. `header` union fields helps at all - maybe get rid of it (and the union)?

I don't get what the `+ 31` are for and why this number - add comments?

Separately from (and after) this PR, I think we could want to renumber the tunable costs so that the KDF type is cost 0 and there's no gap in used tunable cost numbers for any KDF.

magnumripper:
> I'm not sure the separation into `contents` vs. `header` union fields helps at all - maybe get rid of it (and the union)?

I did consider that - fixed now. The separate names were handy while figuring out the differences in KDBX 4 (using their vocabulary for the fields) but it got messy with this commit.

> I don't get what the `+ 31` are for and why this number - add comments?

Fixed. The minimum content size is 32, I now write it as `+ 32 - 1` where the `- 1` is because we use `contents[1]` for the variable size array. I can't remember why we don't declare them as eg. `contents[0]`, I think there were some problems with portability.

> Separately from (and after) this PR, I think we could want to renumber the tunable costs so that the KDF type is cost 0 and there's no gap in used tunable cost numbers for any KDF.

I think changing the KDF cost values would just add to the mess: They were set by argon2-opencl but KeePass never use Argon2i.
As for what to use for cost 0, keepass-opencl [only supporting AES] only has `t` so changing that [for the other ones] would make cost 0 inconsistent between CPU and OpenCL.

PR #5589, opened 2024-11-27
Dynamic salt formats: Drop a redundant and very confusing pointer copy

magnumripper:
Similar (yet different, and relevant) code is used in other formats where we don't copy just a pointer but a complete salt.


solardiz (approved):
I skimmed this with `?w=1` added to the URL (as stuff inside a `get_salt` was unindented because the code now assumes the string had passed `valid`, as we generally do in these functions). I didn't spot anything wrong.

PR #5600, opened 2024-12-02
Autoconf: Use -std=gnu99 if supported

magnumripper:
We first ensure it's supported by the compiler, so this "can't" break anything [famous last words, right?].

Also includes a commit that refactors confusing struct, union and member names for dyna_salt.

<s>If this is merged, it should be after merging #5599 or that one will be included too.</s> **Now rebased**.


magnumripper:
BTW my testing included on super, which is where I stumbled upon the problem (it has gcc 4.4.7). I will test a few more (on top of build bots compilers and versions).

magnumripper:
clang 16.0.0 "(clang-1600.0.26.4)" new warnings with this (but does accept the gnu99 option, and builds fine)
```
gpg2john.c:1632:20: warning: passing arguments to a function without a prototype is deprecated in all versions of C and is not supported in C2x [-Wdeprecated-non-prototype]
 1632 |                         (*tag_func[tag])(len, 1, partial, hash);        // first packet (possibly only one if partial is false).
      |                                         ^
```

magnumripper:
Just a note
```
checking whether we are using the GNU C compiler... yes
checking whether gcc accepts -g... yes
checking for gcc option to accept ISO C89... none needed
(...)
checking if gcc supports -Wall... yes
checking if gcc supports -std=gnu99... yes
checking if gcc supports -Wno-stringop-truncation... no
```
That "ISO C89" test seems to be something built-in to autoconf, nothing added to makefiles.

solardiz:
I think this PR is not ready. Let's merge #5599 first, then discuss here.

Anonymous structs/unions is a useful feature, but I've been avoiding it in this tree so far. What uses of it do we currently have?

magnumripper:
> Anonymous structs/unions is a useful feature, but I've been avoiding it in this tree so far. What uses of it do we currently have?

We've had them for over a decade.
The bitfields in dyna_salt.h and all uses of them (many). Sure we can change them but it would make code fugly.
KeePass, for things that were renamed in KDBX4 spec. so unions for readability. Convenient for following the code but not super important now that it works.
slow_hash_plug.c, uaf2john.c and wpapcap2john.c, presumably structs copied from "original" code.

magnumripper:
I'm not quite sure about dyna_salt: It's either that the bitfield is anonymous, or that the allocation unit is `size_t` (which we do need), or both.

magnumripper:
From a quick look at wpapcap2john.h it seems the bitfields are not anonymous, but they do use `uint16_t`as allocation unit - and an odd number of them. Strict standard only allows variations of `int`. It would be hard to change this.

magnumripper:
> It would be hard to change this.

Looking at the uses of it, maybe not, in that case. We could probably use `int` for them.

magnumripper:
What is the (specific) reason for us to switch to C99 again? Perhaps we could enable whatever feature is needed - but that would likely be even less portable. Maybe it's mostly things like `for (int i; ...)`.

Also, we could obviously rewrite any such code, but for eg. princeprocessor and mbedtls that would make syncing with upstream much harder.

solardiz (approved):
OK, I don't mind `gnu99` for now.

solardiz:
This broke build (or rather the final linking) on our "well" as follows:

```
dynamic_fmt.o: In function `DynamicFunc__crypt_md5':
/home/solar/j/bleeding-jumbo-20241211/src/dynamic_fmt.c:5274: undefined reference to `MD5_body_for_thread'
dynamic_fmt.o: In function `DynamicFunc__POCrypt':
/home/solar/j/bleeding-jumbo-20241211/src/dynamic_fmt.c:5373: undefined reference to `MD5_body_for_thread'
dynamic_fmt.o: In function `DynamicFunc__crypt2_md5':
/home/solar/j/bleeding-jumbo-20241211/src/dynamic_fmt.c:5416: undefined reference to `MD5_body_for_thread'
dynamic_fmt.o: In function `DynamicFunc__crypt_md5_in1_to_out2':
/home/solar/j/bleeding-jumbo-20241211/src/dynamic_fmt.c:5500: undefined reference to `MD5_body_for_thread'
dynamic_fmt.o: In function `DynamicFunc__crypt_md5_in2_to_out1':
/home/solar/j/bleeding-jumbo-20241211/src/dynamic_fmt.c:5586: undefined reference to `MD5_body_for_thread'
dynamic_fmt.o:/home/solar/j/bleeding-jumbo-20241211/src/dynamic_fmt.c:5674: more undefined references to `MD5_body_for_thread' follow
```

I've tried adding `-std=gnu99` also at link time, but this made no difference. I'll open an issue.

Issue #5594, opened 2024-11-30
Faster (by a magnitude) OpenCL AES code


magnumripper:
https://github.com/openwall/john/pull/5591#issuecomment-2507661775

The default KeePass benchmark includes old low iteration test vectors. To see modern, real-life cost, edit opencl_keepass_fmt_plug.c, change the define of KEEPASS_REAL_COST_TEST_VECTORS to 1, rebuild and benchmark with `-cost=31250000` (it will take a looong time).

Each KDF-AES round is 32 bytes of AES-256 so at this cost it will encrypt exactly 1 GB with AES-256 per candidate - great for testing AES performance. Our shared OpenCL AES comes in two flavors. Using the "bitsliced AES" (default for GPU - I kinda doubt bitsliced is a correct description but it does _some_ things in parallel), Super's 1080 does 20.5 GB/s which I believe [could be improved](https://eprint.iacr.org/2021/646) with <s>at least 10x</s> over 5x, (and forcing it to run our older, even worse, vanilla AES it's down to 1.15 GB/s which is essentially useless for anything but decrypting a short verifier).

The faster GPU code (CUDA) is at https://www.github.com/cihangirtezcan/CUDA_AES and there's also a [freely available PDF](https://eprint.iacr.org/2021/646.pdf).

magnumripper:
> Our shared OpenCL AES comes in two flavors. Using the "bitsliced AES" (default for GPU - I kinda doubt bitsliced is a correct description

I had a closer look at our current code now. It really is bitsliced, and here's how:
```c
/*
 * This constant-time implementation is "bitsliced": the 128-bit state is
 * split over eight 32-bit words q* in the following way:
 *
 * -- Input block consists in 16 bytes:
 *    a00 a10 a20 a30 a01 a11 a21 a31 a02 a12 a22 a32 a03 a13 a23 a33
 * In the terminology of FIPS 197, this is a 4x4 matrix which is read
 * column by column.
 *
 * -- Each byte is split into eight bits which are distributed over the
 * eight words, at the same rank. Thus, for a byte x at rank k, bit 0
 * (least significant) of x will be at rank k in q0 (if that bit is b,
 * then it contributes "b << k" to the value of q0), bit 1 of x will be
 * at rank k in q1, and so on.
 *
 * -- Ranks given to bits are in "row order" and are either all even, or
 * all odd. Two independent AES states are thus interleaved, one using
 * the even ranks, the other the odd ranks. Row order means:
[...]

PR #5606, opened 2024-12-04
OpenCL: Double the performance of shared AES-CBC-decrypt

magnumripper:
Ensure AES_cbc_decrypt() utilizes the parallel capability of the bitsliced AES code when more than one block is decrypted at once.  For encryption this is not possible as the IV for block n is unknown until block n-1 has been encrypted.

The potential boost is great (over 2x seen with AES-ECB) but none of our current formats use AES-CBC enough for this to make a significant difference, at least not with their (often small) test vectors.

See #5594.

Also a few commits that gets rid of OpenCL build warnings.


solardiz (approved):
Doesn't look wrong ;-)

PR #5608, opened 2024-12-06
OpenCL: Show ptxas info from verbosity 4 (was 5)

magnumripper:
Level 5 would also show autotune stats, which is not always what I want. Also, we now get compiler warnings and ptxas at the same verbosity level which is logical (to me, at least).

### DIFF
--- a/src/opencl_common.c
+++ b/src/opencl_common.c
@@ -1142,7 +1142,7 @@ static char *get_build_opts(int sequential_id, const char *opts)
 #ifdef __APPLE__
 	        "-D__OS_X__ ",
 #else
-	        (options.verbosity >= VERB_MAX &&
+	        (options.verbosity >= VERB_LEGACY &&
 	         gpu_nvidia(device_info[sequential_id])) ?
 	         "-cl-nv-verbose " : "",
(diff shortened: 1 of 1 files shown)


solardiz (approved):
I don't mind, although maybe I'd want to see auto-tuning before I see build logs (so levels 4 and 5 swapped in this respect?)

Also, the names of these `VERB_*` constants are weird. But not to fix here.

magnumripper:
Yes they are weird for hysterical reasons.

I'd love to change verbosity to a bitfield and have verbosity only as "low or high" but settable for specific contexts. So in your case something like `-verbosity=+autotune,-nv-verbose` (overriding any conf settings - they would be supported for each bit/context). It could even have some kind of backward compatibility so `-verbosity=5` would set and unset default bits to a result similar to today. But I'm not sure I'll ever have time to implement any of this (and with "love to implement" I really mean "love to have implemented by someone else" lol.

Another thing I've pondered is to add something like `-conf-override=xx` so we don't need specific options for overriding config values -  would make it a one-for-all canonical code. Like ssh `-o` option.

PR #5613, opened 2024-12-16
OpenCL: Default to table based AES, now backed in local memory

### LINKED ISSUE #5594: Faster (by a magnitude) OpenCL AES code

magnumripper:
This revised version pushes 1.4 Tbps of AES-128 decryption (axcrypt) or 914 Gbps of AES-256 encryption (keepass) on a 4070ti.

The bitsliced code we defaulted to before is really good but it's register hungry. It has some merits when two or more blocks are encrypted/decrypted at once (does two in parallel) but still is slower than table based now. We can still opt in to use it.

Closes #5594


magnumripper:
Up to 6.6x boost seen on super's AMD, and a mere 3x on the nvidias. I was hoping to achieve 10x but it turns out the bitsliced AES we used was too good as baseline for that to happen :) Anyway we seem to be on par with hashcat now.

solardiz (approved):
Cool stuff, but impossible to review for real without diving into it.

As a minor suggestion, maybe the moving of tables to a separate file can be a separate commit?

magnumripper:
> impossible to review for real without diving into it.

The important changes are to opencl_aes_plain.h and are surprisingly few due to macros.

> As a minor suggestion, maybe the moving of tables to a separate file can be a separate commit?

I fail to see the point of that, but did so now. BTW the bitlocker format was also changed to use that table file but was otherwise not changed - it has its own copy of more or less identical (afaics) table based AES but did not gain anything from using local memory as the AES part of it is insignificant. So I did not commit any such changes to it.

src/opencl_keepass_fmt_plug.c
```diff
@@ -53,10 +53,13 @@ typedef struct {
 	uint32_t cracked;
 } result;
 
+#define AES_MAXNR   14
```
solardiz, line 56: FWIW, this addition of `AES_MAXNR` to `opencl_keepass_fmt_plug.c` looks unused.

solardiz:
> > As a minor suggestion, maybe the moving of tables to a separate file can be a separate commit?
> 
> I fail to see the point of that, but did so now.

Thank you for doing it. Looks cleaner to me that way, and makes the actual changes (in other commits) stand out.

src/opencl_keepass_fmt_plug.c
magnumripper: Good catch, that's a remnant from the older state struct.

run/opencl/7z_kernel.cl
```diff
@@ -162,7 +163,7 @@ __kernel void sevenzip_aes(__constant sevenzip_salt *salt,
 	/* Early rejection if possible (only decrypt last 16 bytes) */
 	if (pad > 0 && salt->length >= 32) {
 		uint8_t buf[16];
-		AES_KEY akey;
+		AES_KEY akey; akey.lt = &lt;
```
magnumripper, line 166: Oh BTW I first had this as `AES_KEY akey = { .lt = &lt };` everywhere, but some device did not like that. Unfortunately I forget which. Good to know, perhaps I should start a wiki page listing knowledge like that. Another example would be the `static` vs `inline` vs `static inline` that we had to add a [workaround for in opencl_misc.h](https://github.com/openwall/john/blob/18eae8da9d095bd4909d3a1f32294c9d0d4bfa08/run/opencl/opencl_misc.h#L54-L72).

Br0kenUK:
Hi, can you use inverse sbox to speed up the invert key mix columns. I got 40% speed up doing it in hashcats code on aes256

magnumripper:
> Hi, can you use inverse sbox to speed up the invert key. I got 40% speed up doing it in hashcats code on aes256

Are we not doing that already?

Br0kenUK:
> > Hi, can you use inverse sbox to speed up the invert key. I got 40% speed up doing it in hashcats code on aes256
> 
> Are we not doing that already?

I don't think so? Especially if your speed is only on par with hashcat

magnumripper:
I'm trying to context switch into this. My current guess is you mean we don't have the four T-tables for decryption, only a plain reverse sbox table - right?

Is your code in hashcat already or are you planning a PR? Feel free to add a PR here as well!

Br0kenUK:
> > Hi, can you use inverse sbox to speed up the invert key. I got 40% speed up doing it in hashcats code on aes256
> 
> Are we not doing that already?

https://pastebin.com/b9CBfXRS

Above is roughly what I'm doing, I have left some random things out unrelated like hashcats decrypt function. This was mostly chatgpts idea so take it with a grain of salt, regardless it got me 40% increase in speed on a certain hashcat mode which also has additional md5 steps( and it works perfect). The only thing that has actually changed is the invert key step and the tables. And no I havn't PR'd it there, from what I understand they dont care much about improving AES.

magnumripper:
You are very welcome to make a PR for us, for fun and fame 😉 

If not, I will look into it! And anyway thanks a lot for this suggestion!

Br0kenUK:
I do not care for fame on things that was not my idea. I just randomly came across this issue and suggested it as it helped me. I won't be PRing it as I dont have the time to figure out your code aswell. GL.

PR #5614, opened 2024-12-17
WPAPSK CPU formats: Remove dependency on OpenSSL for 802.11w support

magnumripper:
Use the same simple CMAC-AES-128 implementation we have in OpenCL.

### DIFF
--- a/src/cmac_plug.c
+++ b/src/cmac_plug.c
@@ -1,0 +1,113 @@
+/*
+ * This code implements the CMAC (Cipher-based Message Authentication)
+ * algorithm described in FIPS SP800-38B using the AES-128 cipher.
+ *
+ * Copyright (c) 2017, magnum.
+ * Copyright (c) 2008 Damien Bergamini <damien.bergamini@free.fr>
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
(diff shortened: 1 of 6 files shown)


magnumripper:
I should have fixed this back in 2017 when I added the CMAC code for OpenCL. As expected, I see no change in performance - the AES-CMAC shouldn't be significant.

src/cmac_plug.c
```diff
@@ -0,0 +1,113 @@
...
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#include "cmac.h"
+#include "common.h"
```
solardiz, line 22: What do we use from `common.h` here?

solardiz (approved):
I didn't review the CMAC code itself, but if the bots are happy it should be right.

src/cmac_plug.c
magnumripper: Just the MIN() macro.

PR #5623, opened 2024-12-20
Create fvde2john.py

holly-o:
Added fvde2john.py for extracting FileVault hashes. Easier and more reliable hash extraction. Also will give user info such as username/password hint

### DIFF
--- a/run/fvde2john.py
+++ b/run/fvde2john.py
@@ -1,0 +1,246 @@
+#!/usr/bin/env python3
+
+# Usage: python3 fvde2john.py <image_file>
+# The partition table is parsed to find the boot volume, often named 'Recovery HD'. The boot volume can be identified by its type GUID: 426F6F74-0000-11AA-AA11-00306543ECAC.
+# The boot volume contains a file called `EncryptedRoot.plist.wipekey`. This is stored on the volume at `/com.apple.boot.X/System/Library/Caches/com.apple.corestorage/EncryptedRoot.plist.wipekey`, where `X` is variable but is often `P` or `R`. This plist file is encrypted with AES-XTS; the key is found in the CoreStorage volume header, and the tweak is b'\x00' * 16.
+# The decrypted plist contains information relating to the user(s). This includes the salt, kek and iterations required to construct the hash as well as information such as username and password hints (if present).
+
+import plistlib
+import os
+import argparse
+import sys
+
+try:
+    from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
+except ImportError:
+    sys.stderr.write("cryptography is missing, run 'pip install --user cryptography' to install it!")
+    sys.exit(1)
+try:
+    import pytsk3
+except ImportError:
+    sys.stderr.write("pytsk3 is missing, run 'pip install --user pytsk3' to install it!")
+    sys.exit(1)
+
+HEX_CORE_STORAGE_TYPE_GUID = '53746F72-6167-11AA-AA11-00306543ECAC'
+HEX_APPLE_BOOT_STORAGE_TYPE_GUID = '426F6F74-0000-11AA-AA11-00306543ECAC'
+LOCAL_USER_TYPE_ID = 0x10060002
+
+def uint_to_int(b):
+    return int(b[::-1].hex(), 16)
+
+def guid_to_hex(guid):
+    guid_parts = guid.split('-')
+
+    hex_str  = ''.join([guid_parts[0][i:i+2] for i in range(0, len(guid_parts[0]), 2)][::-1])
+    hex_str += ''.join([guid_parts[1][i:i+2] for i in range(0, len(guid_parts[1]), 2)][::-1])
+    hex_str += ''.join([guid_parts[2][i:i+2] for i in range(0, len(guid_parts[2]), 2)][::-1])
+    hex_str += guid_parts[3]
+    hex_str += guid_parts[4]
+
+    return hex_str.lower()
+
+# call in place of fp.read(), to stop reading out of bounds of file
+def try_read_fp(fp, bytes_to_read):
+    out = fp.read(bytes_to_read)
+    if len(out) != bytes_to_read:
+        sys.stderr.write("Error reading out of bounds of file, exiting.")
+        sys.exit(1)
+
+    return out
+
+def parse_partition_table(fp):
+    # determine whether sector size is 0x200 or 0x1000
+    sector_size = 0x0
+
+    # look for EFI PART at start of sector 1
+    fp.seek(0x200)
+    signature = try_read_fp(fp, 0x8)
+    if signature == b'EFI PART':
+        sector_size = 0x200
+
+    else:
+        fp.seek(0x1000)
+        signature = try_read_fp(fp, 0x8)
+        if signature == b'EFI PART':
+            sector_size = 0x1000
+
+    if not sector_size:
+        sys.stderr.write(f"[!] Invalid sector size {sector_size} (not 512 or 4096 bytes). Exiting.")
+        sys.exit(1)
+
+    fp.seek(2 * sector_size) # go to sector 2
+    partitions = []
+    partition_entry = b'1'
+    while any(partition_entry):
+        partition_entry = try_read_fp(fp, 0x80)
+        if any(partition_entry):
+            partitions.append(partition_entry)
+
+    partition_dict = {}
+    for p in partitions:
+        part_GUID, type_GUID, start, partition_name = parse_partition_entry(p)
+        sp = uint_to_int(start) * sector_size
+        partition_dict[part_GUID.hex()] = {'start':sp, 'partition_type':type_GUID.hex(), 'partition_name':partition_name.decode('utf-16').strip('\x00')}
+
+    return partition_dict
+
+def findall(p, s):
+    i = s.find(p)
+    while i != -1:
+        yield i
+        i = s.find(p, i+1)
+
+def parse_partition_entry(partition_entry):
+    type_GUID = partition_entry[0:0x10]
+    part_GUID = partition_entry[0x10:0x20]
+    start_LBA = partition_entry[0x20:0x28]
+    partition_name = partition_entry[0x38:0x80]
+    return part_GUID, type_GUID, start_LBA, partition_name
+
+def parse_corestorage_header(fp, start_pos):
+    fp.seek(start_pos + 176)
+    aes_key = try_read_fp(fp, 0x10)
+    return aes_key
+
+def AES_XTS_decrypt(aes_key, tweak, ct):
+    decryptor = Cipher(
+        algorithms.AES(key=aes_key + b'\x00' * 16),
+        modes.XTS(tweak=tweak),
+    ).decryptor()
+    pt = decryptor.update(ct)
+    return pt
+
+def parse_keybag_entry(uuid, pt):
+    uuid_iterator = findall(uuid, pt)
+    for sp in uuid_iterator:
+        ke_uuid, ke_tag, ke_keylen = pt[sp:sp+16], uint_to_int(pt[sp + 16:sp + 18]), uint_to_int(pt[sp + 18:sp + 20])
+        padding = pt[sp + 20:sp + 24]
+        keydata = pt[sp + 24: sp + 24 + ke_keylen]
+
+        # only tag 3 is needed for constructing the hash
+        if ke_tag == 3:
+            assert padding == b'\x00\x00\x00\x00'
+            return keydata
+
+    return None
+
+def get_all_partitions_of_type(partition_dict, part_type):
+    return [partition_dict[p]['start'] for p in partition_dict if partition_dict[p]['partition_type'] == guid_to_hex(part_type)]
+
+def load_plist_dict(pt):
+    # resultant pt has one extra malformed line in the xml, so we remove this.
+    plist_str = b''.join(pt.split(b'\n')[:-1]).decode()
+    d = plistlib.loads(plist_str)
+    return d
+
+# Recursive traversal - Recovery HD partition does not contain a lot of files, and so this approach is fine
+def traverse_filesystem(fs_object, target_file, path='/'):
+    for entry in fs_object.open_dir(path):
+        if entry.info.name.name in [b'.', b'..']:
+            continue
+
+        file_path = os.path.join(path, entry.info.name.name.decode('utf-8'))
+
+        if entry.info.meta and entry.info.meta.type == pytsk3.TSK_FS_META_TYPE_REG:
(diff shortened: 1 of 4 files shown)


solardiz:
Thank you @holly-o!

Please also add a `doc/NEWS` entry. Maybe some other documentation additions are also relevant?

For the username and password hints, perhaps the script should produce an /etc/passwd-format line, putting the username into the first field, hash into the second, and password hints into the GECOS field (but first replace any potential embedded `:` characters in the username/hints). Then our "single crack" mode will pick those up and base candidate passwords on them.

I see there's a little bit of try/except, but perhaps it's not exhaustive for potential file read errors - in particular, the seeks look unchecked, so if a seek offset is ever beyond file size that will go unnoticed? Perhaps room for improved robustness there. But I did not review this carefully, only skimmed.

@exploide Any review from you as well?

solardiz:
> For the username and password hints, perhaps the script should produce an /etc/passwd-format line

Are these "hashes" also supported by hashcat, though? If so, I guess hashcat would insist on a bare hash? And we could need to support both output formats to be friendly to hashcat.

solardiz:
> Maybe some other documentation additions are also relevant?

Yes, this should also be added to `doc/README.FileVault2`, which currently only refers to @kholia's external project https://github.com/kholia/fvde2john, which is an archived repo so I assume is not currently maintained. We should probably keep that reference as a secondary option and make this new script the primary.

exploide:
> @exploide Any review from you as well?

Will do a review tomorrow.

magnumripper:
> > For the username and password hints, perhaps the script should produce an /etc/passwd-format line
> 
> Are these "hashes" also supported by hashcat, though? If so, I guess hashcat would insist on a bare hash? And we could need to support both output formats to be friendly to hashcat.

FWIW hashcat does have a `--username` option that makes it understand a login field. Unfortunately it still chokes on fields after the hash though (at least last time I tried).

run/fvde2john.py
```diff
@@ -0,0 +1,220 @@
...
+# Usage: python3 fvde2john.py <image_file>
+# The partition table is parsed to find the boot volume, often named 'Recovery HD'. The boot volume can be identified by its type GUID: 426F6F74-0000-11AA-AA11-00306543ECAC.
+# The boot volume contains a file called `EncryptedRoot.plist.wipekey`. This is stored on the volume at `/com.apple.boot.X/System/Library/Caches/com.apple.corestorage/EncryptedRoot.plist.wipekey`, where `X` is variable but is often `P` or `R`. This plist file is encrypted with AES-XTS; the key is found in the CoreStorage volume header, and the tweak is b'\x00' * 16.
+# The decrypted plist contains information relating to the user(s). This includes the salt, kek and iterations required to construct the hash as well as information such as username and password hints (if present).
+
+from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
+import plistlib
+import os
+import argparse
+import pytsk3    # pip install pytsk3
```
exploide: For third party dependencies, we usually add a try-except block in order to guide the user towards the installation of the correct package.
I also moved the `cryptography` import below the stdlib imports in my suggestion.

```suggestion
import argparse
import os
import plistlib
import sys

try:
    from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
except ImportError:
    print("cryptography is missing, run 'pip install --user cryptography' to install it!")
    sys.exit(1)
try:
    import pytsk3
except ImportError:
    print("pytsk3 is missing, run 'pip install --user pytsk3' to install it!")
    sys.exit(1)
```

run/fvde2john.py
```diff
@@ -0,0 +1,220 @@
...
+    else:
+        fp.seek(0x1000)
+        signature = fp.read(0x8)
+        if signature == b'EFI PART':
+            sector_size = 0x1000
+
+    print("[+] Identified sector size:", sector_size)
+
+    if not sector_size:
+        print(f"[!] Invalid sector size {sector_size} (not 512 or 4096 bytes). Exiting.")
```
exploide: It says it will exit but it doesn't. A `sys.exit(1)` is missing.

```suggestion
        print(f"[!] Invalid sector size {sector_size} (not 512 or 4096 bytes). Exiting.")
        sys.exit(1)
```

run/fvde2john.py
```diff
@@ -0,0 +1,220 @@
...
+        ke_uuid, ke_tag, ke_keylen = pt[sp:sp+16], uint_to_int(pt[sp + 16:sp + 18]), uint_to_int(pt[sp + 18:sp + 20])
+        padding = pt[sp + 20:sp + 24]
+        keydata = pt[sp + 24: sp + 24 + ke_keylen]
+
+        # only tag 3 is needed for constructing the hash
+        if ke_tag == 3:
+            assert padding == b'\x00\x00\x00\x00'
+            volume_unlock_record = keydata
+
+            return volume_unlock_record
```
exploide: This can be simplified.

```suggestion
            return keydata
```

run/fvde2john.py
```diff
@@ -0,0 +1,220 @@
...
+                    return file_data
+
+            # Traverse lower layer if entry is a dir
+            elif entry.info.meta and entry.info.meta.type == pytsk3.TSK_FS_META_TYPE_DIR:
+                file_data = traverse_filesystem(fs_object, target_file, file_path)
+
+                if file_data:
+                    return file_data
+
+        except Exception as e:
```
exploide: Catching a bare `Exception` always has the drawback of silencing other unexpected errors which could maybe be handled in a better way. Is there any more *specific* exception type suitable to be caught here instead?
Besides, should the program exit here, or is this a soft failure?

run/fvde2john.py
```diff
@@ -0,0 +1,220 @@
...
+def recover_file(fs_object, file_path):
+    try:
+        file_obj = fs_object.open(file_path)
+        size = file_obj.info.meta.size
+        offset = 0
+        data = file_obj.read_random(offset, size)
+
+        return data
+
+    except Exception as e:
```
exploide: Same as above.

run/fvde2john.py
```diff
@@ -0,0 +1,220 @@
...
+        partition_dict = parse_partition_table(fp)
+
+        core_storage_volumes = get_all_partitions_of_type(partition_dict, HEX_CORE_STORAGE_TYPE_GUID)
+        boot_volumes = get_all_partitions_of_type(partition_dict, HEX_APPLE_BOOT_STORAGE_TYPE_GUID)
+
+        # Unlikely to have more than one boot volume, but loop anyway
+        for boot_start_pos in boot_volumes:
+            EncryptedRoot_data = get_EncryptedRoot_plist_wipekey(image_file, boot_start_pos)
+
+            if core_storage_volumes == []:
```
exploide: Since this condition is already clear before entering the for loop, I suggest moving it before the loop.
Besides, the `exit()` should be a `sys.exit(1)`.

exploide (changes requested):
Left a few review comments and suggestions. :)

Looks fine overall. I just generally recommend running a code linter and formatter but no big deal here.

I haven't put much thought into the parsing logic since I'm not familiar with the FileVault data structures. And of course I did not test it in the absence of test data.

exploide:
> Are these "hashes" also supported by hashcat, though? If so, I guess hashcat would insist on a bare hash?

Yes it seems so.

```console
$ hashcat -m 16700 --hash-info 
hashcat (v6.2.6) starting in hash-info mode

Hash Info:
==========

Hash mode #16700
  Name................: FileVault 2
  Category............: Full-Disk Encryption (FDE)
  Slow.Hash...........: Yes
  Password.Len.Min....: 0
  Password.Len.Max....: 256
  Salt.Type...........: Embedded
  Salt.Len.Min........: 0
  Salt.Len.Max........: 256
  Kernel.Type(s)......: pure
  Example.Hash.Format.: plain
  Example.Hash........: $fvde$1$16$84286044060108438487434858307513$20000$f1620ab93192112f0a23eea89b5d4df065661f974b704191
  Example.Pass........: hashcat
  Benchmark.Mask......: ?b?b?b?b?b?b?b
  Autodetect.Enabled..: Yes
  Self.Test.Enabled...: Yes
  Potfile.Enabled.....: Yes
  Custom.Plugin.......: No
  Plaintext.Encoding..: ASCII, HEX
```

run/fvde2john.py
magnumripper: I didn't check, but maybe this redundancy adds an oz of self documentation in line with some spec?

run/fvde2john.py
```diff
@@ -0,0 +1,220 @@
...
+                        print("Passphrase hint:", PassphraseHint)
+
+                        PassphraseWrappedKEKStruct = d['CryptoUsers'][userIndex].get('PassphraseWrappedKEKStruct')
+                        print("[+] Found PassphraseWrappedKEKStruct in decrypted plist")
+
+                        salt = PassphraseWrappedKEKStruct[8:24]
+                        kek  = PassphraseWrappedKEKStruct[32:56]
+                        iterations = uint_to_int(PassphraseWrappedKEKStruct[168:172])
+
+                        print(f"[+] Successfully extracted hash: $fvde$1${len(salt)}${salt.hex()}${iterations}${kek.hex()}")
```
magnumripper: I'm a python n00b but to me this looks like it's literally printing something like
```
[+] Successfully extracted hash: $fvde$1$(...)
```
If so, maybe the "[+] Successfully extracted hash" should go to stderr, if kept at all?

exploide: Yes, all messages in the script currently go to stdout. This is in contrast to most other scripts, where only the hash is printed to stdout and everything else to stderr.
I would have recommended to adapt to this behavior. Though the difference to other scripts is that this prints additional information like username and password hint.
But yes, at the moment it is untypically verbose.

solardiz: I think one of the script's output lines (on stdout) should be an /etc/passwd-format line without any sort of `[+] Successfully extracted hash: ` prefix or such. So that if output is redirected to a file, that file is directly and optimally usable with `john`, without any need to manually select a portion of output and copy-paste.

solardiz:
> FWIW hashcat does have a `--username` option that makes it understand a login field. Unfortunately it still chokes on fields after the hash though (at least last time I tried).

Well, that should probably be fixed in hashcat. Maybe we should check if it still chokes on such input and send a PR fixing that.

run/fvde2john.py
holly-o: Yes this was from following the documentation, but will simplify anyway

holly-o:
> 
> For the username and password hints, perhaps the script should produce an /etc/passwd-format line, putting the username into the first field, hash into the second, and password hints into the GECOS field (but first replace any potential embedded `:` characters in the username/hints). Then our "single crack" mode will pick those up and base candidate passwords on them.

For the cases where there is any potential embedded `:` characters or any spaces in the username/hints, is it best to remove these or replace with another character e.g. underscore, dash..?

solardiz:
> For the cases where there is any potential embedded `:` characters or any spaces in the username/hints, is it best to remove these or replace with another character e.g. underscore, dash..?

Spaces should be no problem, so keep those. And I suggest replace `:` with spaces.

Also any embedded newline characters (LF or CR) would be problematic. I suggest replace them with spaces too. I think our other `*2john` programs aren't careful about that, but it may be time to start to improve.

magnumripper:
> I think our other `*2john` programs aren't careful about that, but it may be time to start to improve.

It varies wildly. Some are careful, most are not. I think they've been fixed on a case by case basis.

solardiz:
> Spaces should be no problem, so keep those. And I suggest replace `:` with spaces.

I've just checked, we already have precedent - commit 4ef1ef28f91034cebce396f81456c540f00355ab implements just that kind of replacement in some `*2john` tools. In some others, where pathnames were used, it simply takes the basename portion (omits the directory path, which is what on Windows would have the colon after drive letter), so may still have the problem if the filename contains a colon.

As to CR/LF, I only found one example where we take care of them:

```
1b7ac0fdf2 run/odf2john.py         (ShaneQful    2013-07-02 20:35:04 +0100 145)         gecos = gecos.replace("\n","").replace("\r","").replace(":","")
```

This is now `libreoffice2john.py`. This script also takes `os.path.basename(filename)` without care for possible "bad" characters in there, but for the GECOS field it's careful like the above (but it removes the characters rather than replacing with a space - we could want to change that).

`libreoffice2john.py` is also usable as example for how to format the line to include hints in the GECOS field. It uses:

```python
        sys.stdout.write("%s:$odf$*%s*%s*%s*%s*%s*%d*%s*%d*%s*%d*%s:::%s::%s\n" % \
                (os.path.basename(filename), algorithm_type, checksum_type,
                iteration_count, key_size, checksum, len(iv) / 2, iv,
                len(salt) / 2, salt, 0, binascii.hexlify(content).decode("ascii"),
                gecos, filename))
```

Oh, I now see it also puts the full filename into the shell field. I think we only parse that field for the shells filter option:

```
--shells=[-]SHELL[,..]     Load users with[out] this (these) shell(s) only
```

I doubt anyone would know to use that. So we could want to instead include the full filename into the GECOS field (perhaps after an added space separator) prior to the bad characters replacements.

I didn't mean to review that other script in here - it's just an example that is in some ways right and in others not.

exploide (approved):
Looks good :) Thanks for incorporating the changes!

run/fvde2john.py
```diff
@@ -0,0 +1,246 @@
...
+
+                        full_name_info = format_hash_str(full_name_info)
+                        username_info  = format_hash_str(username_info)
+                        passphrase_hint = format_hash_str(passphrase_hint)
+
+                        # Hash info stored in the PassphraseWrappedKEKStruct in decrypted plist
+                        PassphraseWrappedKEKStruct = d['CryptoUsers'][user_index].get('PassphraseWrappedKEKStruct')
+                        fvde_hash = construct_fvde_hash(PassphraseWrappedKEKStruct)
+
+                        sys.stdout.write(f"{username_info}:{fvde_hash}:{full_name_info}:{passphrase_hint}\n")
```
solardiz, line 240: This doesn't appear to skip the UID/GID fields, yet puts `{full_name_info}:{passphrase_hint}` into two separate fields where there aren't two arbitrary-format fields like that. I suggest that we fit the usual 7 fields of `/etc/passwd`:

```python
sys.stdout.write(f"{username_info}:{fvde_hash}:::{full_name_info} {passphrase_hint}::\n")
```

solardiz: I just took care of this with a separate commit, but it's completely untested. I'd appreciate re-testing of the code currently in bleeding-jumbo.

holly-o: I have re-tested this code on example data, and it outputs the hash in the desired format, and cracks correctly.

PR #5627, opened 2024-12-21
OpenCL gostcrypt/streebogcrypt formats bugfixes

### LINKED ISSUE #5626: Excessive unrolling and buggy auto-tuning in gost94crypt-opencl, streebog256crypt-opencl, streebog512crypt-opencl
As discussed in #5533:
1. `#pragma unroll HASH_LOOPS` may take forever to build and produces excessive code size. This unroll should be dropped or at least reduced to 21. Anyway, instead of unrolling we could use one of the tricks we already use elsewhere. Like maintaining counter(s) that are reset to 0 upon reaching 3, 7, 21, or 42 (and use an offsets map like `md5crypt_long_fmt.c` does). Crazy unrolling is really not the way.
2. `LOOP_CALLS` has misplaced braces (and may better be called `LOOP_COUNT` for consistency with other formats).

magnumripper:
Stop unrolling the Drepper crypt loop and fix weird autotune bugs that led to excessive work sizes being tried for no good.

Closes #5626 (I'm assuming the mod operations are insignificant until someone optimizes the basic GOST code a lot).

### DIFF
--- a/run/opencl/gost12256hash_kernel.cl
+++ b/run/opencl/gost12256hash_kernel.cl
@@ -201,7 +201,6 @@ __kernel void gost12256loop(__global inbuf *in,
 #endif
 
 	/* Repeatedly run the collected hash value through Streebog to burn CPU cycles.  */
-#pragma unroll HASH_LOOPS
 	for (cnt = 0; cnt < HASH_LOOPS; ++cnt) {
 		/* New context. */
 		GOST34112012Init(&ctx, 256);
--- a/run/opencl/gost12256hash_kernel.cl
+++ b/run/opencl/gost12256hash_kernel.cl
@@ -12,7 +12,6 @@
 #define STREEBOG_LOCAL_AX       1
 #define STREEBOG_VECTOR         1
 #define STREEBOG_UNROLL         0
-#define STREEBOG_MANUAL_UNROLL  0
 #include "opencl_streebog.h"
 
 #define SALT_LENGTH                 16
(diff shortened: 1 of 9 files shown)


magnumripper:
With this, Tahiti on 'well' goes from build failure with
```
Error:E013:Insufficient Private Resources! 
```
to the only marginally better self-test failure
```
FAILED (cmp_all(1))
```
Same on all three formats

magnumripper:
On AMD Vega, still `FAILED (cmp_all(1))` with gost94crypt-opencl, and unchanged speeds for the Streebog formats.
GTX 1080 also unchanged speeds.  I didn't bother comparing build time (still very long).

magnumripper:
I'm investigating a problem with the code paths used when not using local memory

magnumripper:
> I'm investigating a problem with the code paths used when not using local memory

Fixed.

magnumripper:
I tried explicitly not inlining the GOST (primitives) update/final functions and it worked better than expected: Build times went from 20 seconds (GOST-94) and 135 seconds (Streebog) to 3 and 4 seconds respectively. Speeds more or less unaffected. Perhaps inlining with 21 would now be sane but the speeds are so low I don't think it's worth it.

Edit: Build time and speeds was on GTX 1080. For AMD Vega, similar results (eg. from 35 to 3 seconds build time) and as a bonus gost94crypt stopped failing.

magnumripper:
> For AMD Vega, similar results (eg. from 35 to 3 seconds build time) and as a bonus gost94crypt stopped failing.

Tahiti on 'well' still fails with all three formats. I'm not looking into that, the runtime is a disgrace for such a fine GPU.

PR #5628, opened 2024-12-22
PDF format: Update reference link for version 5 revision 6

lnguyen639:
The old link is no longer accessible. Update to link from Web Archive.


magnumripper (approved):
Thank you!

Issue #5620, opened 2024-12-18
Consider revising the use of inline in eg. unicode.c


magnumripper:
See https://github.com/openwall/john/issues/5611#issuecomment-2551654652

Many functions in eg. unicode.c are tiny and performance critical so should better be inlined to the caller, but as we compile a file at a time that can't happen - unless we _define_ those functions in unicode.h instead, and just _declare_ them `external` in unicode.c.

Other functions that come to mind are eg. `strnzcpy()`.

magnumripper:
Made some tests with code paths that could benefit from this (essentially `strnzcpy()`, targetting `rules.c` for a test), and (not surprisingly) there's no clear benefit. Maybe a 0.5% boost from `-w -ru`, or maybe that was just line noise.

magnumripper:
Also, trying to do this with a few functions in unicode.[ch], it quickly went downhill due to dependency of other functions and variables that are currently static.

magnumripper:
I think I'll do this to just a few self-contained functions, such as `valid_utf8()`

PR #5629, opened 2024-12-22
Inline strnzcpy() and strnzcpyn()

magnumripper:
I thought we could try this out for a while, verifying it doesn't break anywhere, then maybe add a few more. @solardiz do you think it's worth it at all? Function calls may be cheap but in some cases I reckon stuff can be optimized away when inlined.

See #5620

### DIFF
--- a/src/cprepair.c
+++ b/src/cprepair.c
@@ -168,7 +168,7 @@ static int process_file(char *name)
 							inv_cp;
 					initUnicode(0);
 
-					enc_to_utf16(u16, sizeof(u16), (UTF8*)convin, len);
+					enc_to_utf16(u16, LINE_BUFFER_SIZE, (UTF8*)convin, len);
 					out = (char*)utf16_to_utf8_r(u8buf, sizeof(u8buf), u16);
 					if (options.verbosity > VERB_DEFAULT &&
 					    strcmp(convin, out))
@@ -182,7 +182,7 @@ static int process_file(char *name)
 				strcpy (dd, convin);
 				while (*convin) {
 					out = dd;
-					utf8_to_utf16(u16, sizeof(u16),
+					utf8_to_utf16(u16, LINE_BUFFER_SIZE,
 					              (UTF8*)dd, len);
 					if (!valid_ansi(u16))
 						break;
--- a/src/dmg2john.c
+++ b/src/dmg2john.c
@@ -39,7 +39,6 @@
 
 #include "arch.h"
 #include "filevault.h"
-#include "misc.h"
 #include "jumbo.h"
 #include "memory.h"
 #include "johnswap.h"
@@ -104,7 +103,7 @@ static void print_hex(unsigned char *str, int len)
 // to link dmg2john.o with misc.o results in linking errors, as parts of misc.o
 // depends on other parts of JtR jumbo. These linking problems are not easily
 // solvable without refactoring misc.c file.
-int strnzcpyn(char *dst, const char *src, int size)
+static int strnzcpyn(char *dst, const char *src, int size)
 {
 	char *dptr;
 	if (!size) return 0;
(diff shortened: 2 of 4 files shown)


solardiz:
> @solardiz do you think it's worth it at all?

For `strnzcpy`, yes, and I had this thought for a while. For other functions, I'm not sure - would need to consider specific cases.

Also, I think you may partially misunderstand the `extern line` in `.c` file thing. I think it was only needed for `MD5_body_for_thread` because `dynamic_types.h` has it as `extern` without including a header file that would have it. I think this is never the case for the two functions you're currently turning into inline in this PR. I also think this `extern inline` thing is never needed as a defense against the compiler just choosing not to inline despite of the `inline` keyword - that will work just fine anyhow, when the right header file with the function definition is included where the function is needed. So let's exclude the `extern inline` from the `.c` files.

magnumripper:
> Also, I think you may partially misunderstand the `extern line` in `.c` file thing. I think it was only needed for `MD5_body_for_thread` because `dynamic_types.h` has it as `extern` without including a header file that would have it. I think this is never the case for the two functions you're currently turning into inline in this PR. I also think this `extern inline` thing is never needed as a defense against the compiler just choosing not to inline despite of the `inline` keyword - that will work just fine anyhow, when the right header file with the function definition is included where the function is needed. So let's exclude the `extern inline` from the `.c` files.

I need to read up on that again and try to understand what you are saying. Several texts mention specificallly the "define inline in .h and declare extern in .c". Such as https://www.greenend.org.uk/rjk/tech/inline.html or https://stackoverflow.com/questions/6312597/is-inline-without-static-or-extern-ever-useful-in-c99/6312854#6312854.

One thing is unclear: Some texts says the header should have a function definition using `inline` and the source file should have a declaration using `extern` - others say the latter should be `extern inline`. Maybe that detail doesn't matter.

solardiz:
> Several texts mention specificallly the "define inline in .h and declare extern in .c". Such as https://www.greenend.org.uk/rjk/tech/inline.html

Oh, I now realize what I missed/forgot. Yes, "the compiler just choosing not to inline despite of the inline keyword [...] will work just fine anyhow", but if this happens more than once (in different translation units that include the header file), then there will be duplicate instances of the function body in the compiled program (no clash, just wasteful). The approach with `extern in .c` under "Strategies for using inline functions" / "A C99 model." over that link you referenced (very helpful) should avoid that.

So you're right, we need those `extern` things for commonly used functions like these. I was wrong.

solardiz:
I took a look at the uses of `valid_utf8`, and I think we shouldn't be inlining this not-so-tiny function into those. So let's limit this PR's experiment to `strnzcpy`, then see whether we have any issues on any previously supported platform.

solardiz:
> let's limit this PR's experiment to `strnzcpy`

... and `strnzcpyn`, which is also used a lot in similar places.

solardiz (approved):

solardiz:
With this PR, we got a regression for `--enable-ztex` builds:

```
ztex/pkt_comm/cmp_config.o: In function `strnzcpy':
/home/user/j/bleeding-jumbo-20241229/src/ztex/pkt_comm/../../misc.h:121: multiple definition of `strnzcpy'
misc.o:/home/user/j/bleeding-jumbo-20241229/src/misc.h:118: first defined here
ztex/pkt_comm/cmp_config.o: In function `strnzcpyn':
/home/user/j/bleeding-jumbo-20241229/src/ztex/pkt_comm/../../misc.h:146: multiple definition of `strnzcpyn'
misc.o:/home/user/j/bleeding-jumbo-20241229/src/misc.h:142: first defined here
```

solardiz:
> With this PR, we got a regression for --enable-ztex builds

The problem is f88688e273185bfec79d8d5e11cd419ffe41ed12 missed a deeper-level `Makefile.in`.

solardiz:
> > With this PR, we got a regression for --enable-ztex builds
> 
> The problem is [f88688e](https://github.com/openwall/john/commit/f88688e273185bfec79d8d5e11cd419ffe41ed12) missed a deeper-level `Makefile.in`.

I've just fixed this in 41101537f6d75422976c51235aeb0a2e117001e3.

PR #5630, opened 2024-12-22
OpenCL: Minor tweaks of format sort order

magnumripper:
Sort streebogcrypt after gostcrypt, and eg. PBKDF2-HMAC-SHA256-opencl after PBKDF2-HMAC-SHA1-opencl


magnumripper:
Most crypt(3) formats were called eg. (...)cryptmd5 instead of (...)md5crypt so they were grouped together but sorted within the group. I now included descrypt and bcrypt to that order.

Issue #5631, opened 2024-12-22
When no SIMD is found, configure must deactivate SIMD support


ghost:
When no SIMD is found, configure should behave as if the user had set `--disable-simd`:
- set that no SIMD is avaiblable;
- define JOHN_NO_SIMD;
- ...

That's not what's happening on Android.

The lines below will include the header even when SIMD is not supported (in my example).

https://github.com/openwall/john/blob/f3517213a576867423cd9de55e6f720b0339c6cd/src/pseudo_intrinsics.h#L47-L48

```
checking special compiler flags... ARM
checking for NEON... no
checking for ASIMD... no
```

magnumripper:
So what's the output of `gcc -dM -E -x c /dev/null | grep -E 'NEON|SIMD'`? Are the build tools generic ones from third party?

magnumripper:
I think this is basically a bug in [configuration of] the build tools (although we could add some logic to mitigate the problem). The natural workaround is to just pass --disable-simd to configure.

magnumripper:
Actually if they configured a default `-march=armv8-a+simd` I'm not sure we can disable it (unless `-march=native` turns off the relevant functions). There doesn't seem to be any `-no-march=` or `-march=no-(...)`

magnumripper:
Apparently for some ARM archs you can disable simd with an march such as `-march=armv7-a+nosimd`.

It doesn't appear to work on my armv8.3-a macbook M1:
```
$ gcc -dM -E -x c /dev/null -march=armv8.3-a+nosimd | grep -iE 'NEON|SIMD'
#define __AARCH64_SIMD__ 1
#define __ARM_NEON__ 1
```

ghost:
Full log (config.log and Makefile included) when arm32le.h is selected:
[arm32le.h.txt](https://github.com/user-attachments/files/18223820/arm32le.h.txt)

Full log (idem) when arm64le.h is selected (error at line 1179):
[arm64le.txt](https://github.com/user-attachments/files/18223844/arm64le.txt)


I can't see what's causing it:

ghost:
This is bad:

https://github.com/openwall/john/blob/2d585e0f8a796bc0f89f26552e3c0d923a0d498a/src/arm64le.h#L45-L52

And wrong.
****

**[EDITED]**

In my example probably `configure` is doing the right thing and the `#ifndef` above is surely doing something wrong.

magnumripper:
[...]

PR #5632, opened 2024-12-22
autoconf: fix arch.h selection for arm*

ghost:
As it is now, it ends up picking arm32le.h on Android.

```
Target CPU ......................................... arm, 64-bit LE
Target OS .......................................... linux-gnueabi
Legacy arch header ................................. arm32le.h
```

Tested (because of #5631, I have to force `SIMD disabled`):
```
Target CPU ......................................... arm (SIMD disabled), 64-bit LE
Target OS .......................................... linux-gnueabi
Legacy arch header ................................. arm64le.h
```

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -9375,7 +9375,12 @@ ac_compiler_gnu=$ac_cv_c_compiler_gnu
 
 fi
     ;;
-  arm*) ARCH_LINK=arm32le.h endian=little
+  arm*) endian=little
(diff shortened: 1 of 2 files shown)


magnumripper (approved):
This looks right. I fixed this same problem in d6c8792 but apparently only for SIMD.

magnumripper:
Tested this on Macbook, problem triggered when using --disable-simd, and it was fixed with this commit

PR #5638, opened 2024-12-24
Opencl INLINE

### LINKED ISSUE #5618: OpenCL Add a DECLSPEC or INLINE macro
We've had this in opencl_misc.h for ages:
```c
/*
 * Some runtimes/drivers breaks on using inline, others breaks on lack of it,
 * yet others require use of static as well.
 */
#if __MESA__
#define inline	// empty!
#elif __POCL__
// Do nothing (POCL complains if we redefine)
#elif gpu_amd(DEVICE_INFO) // We really target ROCM here
#define inline	static inline
#else
// Do nothing
#endif
```
This is suboptimal because 1) it's not obvious in affected functions that `inline` may be changed, and 2) it would change `static inline` to `static static inline`. Also, the whole thing might be a long gone problem - we do have code that use neither or just `static` and there's no reports of it breaking anything. On the other hand, just maybe #5456 is affected by what syntax (or none) is used  - I recall a few formats do not emit those warnings, while most do.

FWIW hashcat has this:
```c
/**
 * function declarations can have a large influence depending on the opencl runtime
 * fast but pure kernels on rocm is a good example
 */

#ifdef NO_INLINE
#define HC_INLINE
#else
#define HC_INLINE inline static
#endif

#if defined IS_AMD && defined IS_GPU
#define DECLSPEC HC_INLINE
#elif defined IS_HIP
#define DECLSPEC __device__ HC_INLINE
#else
#define DECLSPEC
#endif
```
...then all their functions start with `DECLSPEC` unless they're already ifdef'ed for a specific platform.

magnumripper:
Define NOINLINE as `__attribute__((noinline))` which is has been seen working. Replace all uses of the latter with the macro. Drop the questionable "inline" macro and instead define INLINE as `static inline`, which should be the right thing, for most targets. Use this only for inlines that were replaced by the old inline macro (so not to make an uncontrollable mess at once).

Closes #5618

### DIFF
--- a/run/opencl/7z_kernel.cl
+++ b/run/opencl/7z_kernel.cl
@@ -46,7 +46,7 @@
 		} \
 	}
 
-inline void sha256_zerofinal(uint *W, uint *output, const uint tot_len)
+INLINE void sha256_zerofinal(uint *W, uint *output, const uint tot_len)
 {
 	uint len = ((tot_len & 63) >> 2) + 1;
 
--- a/run/opencl/axcrypt_kernel.cl
+++ b/run/opencl/axcrypt_kernel.cl
@@ -43,7 +43,7 @@ typedef struct {
                 (cp)[2] ^= (uchar)((value) >> 16), \
                 (cp)[3] ^= (uchar)((value) >> 24 ) )
 
-inline int axcrypt_decrypt(__global const axcrypt_password *inbuffer, uint gid, __constant axcrypt_salt *cur_salt, __global axcrypt_out *output, __local aes_local_t *lt)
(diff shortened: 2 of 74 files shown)


solardiz (approved):

solardiz:
> Use this only for inlines that were replaced by the old inline macro (so not to make an uncontrollable mess at once).

Yet maybe we should, seeing how #5778 shows a failure on a symbol declared `inline` without `static`:

```
opencl/androidbackup_kernel.cl:inline int ab_decrypt(__global uchar *key, MAYBE_CONSTANT ab_salt *salt, __local aes_local_t *lt)
```

magnumripper:
Yeah, definitely. I just couldn't picture how to test that on enough devices to be fully comfortable with it.

PR #5639, opened 2024-12-25
Fix BINARY_ALIGN in a few OpenCL formats

magnumripper:
It's amazing we still have this problem 12 years after introducing the alignment warnings.

### DIFF
--- a/src/opencl_rawmd4_fmt_plug.c
+++ b/src/opencl_rawmd4_fmt_plug.c
@@ -291,12 +291,12 @@ static char *split(char *ciphertext, int index, struct fmt_main *self)
 
 static void *get_binary(char *ciphertext)
 {
-	static unsigned char out[DIGEST_SIZE];
+	static uint32_t out[DIGEST_SIZE];
 	char *p;
 	int i;
 	p = ciphertext + TAG_LENGTH;
 	for (i = 0; i < sizeof(out); i++) {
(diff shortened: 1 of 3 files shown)


solardiz (approved):
FWIW, as I recall from PHC list discussions, only `char` and `unsigned char` are allowed to alias any type. `uint8_t` _isn't_. Of course, in practice it's usually handled just the same as `unsigned char`, but theoretically a compiler is free to optimize things away when `uint8_t *` aliases something larger. So I suggest changing those two instances here to `unsigned char *`.

solardiz:
> It's amazing we still have this problem 12 years after introducing the alignment warnings.

I think the reasons for this are:
1. Just because greater expected alignment is not guaranteed does not mean it won't be there in a specific build/run - quite often it will be.
2. I think our CI runs don't check for those warnings at all - they're just buried in the logs. Maybe we should fix that.
3. Our CI setup tests only a handful of OpenCL formats (as part of `ci/circleci: encoding-opencl`, which is mostly an encoding test). We should add explicit OpenCL tests.

magnumripper:
> only char and unsigned char are allowed to alias any type. uint8_t isn't.

So the standard says "a character type". That begs the question whether `wchar_t` is "a character type" and consequently can alias anything. I expect the answer to that is "no" which makes me really hope OpenCL's `uchar` is "a character type".

I'll fix it, to increase the chance of remembering this.

Issue #5610, opened 2024-12-08
Oracle Authentification with a long password (>16 characters)


k4amos:
In an authentication with an Oracle database, the size of AUTH_PASSWORD will change if the password length exceeds 16 characters, as explained [here](https://github.com/hashcat/hashcat/issues/371#issuecomment-226456822). 
@magnumripper had proposed in this issue https://github.com/openwall/john/issues/2155#issuecomment-226731965 to add a field with the size of Auth_PASSWORD at the end of the hash, but unfortunately this option doesn't seem to be implemented.

Is there a way to break these hashes with john?

solardiz:
I skimmed those referenced issues now, and I think this should just work. Have you tried? Unfortunately, we don't seem to have a test vector with a long password - it would be good to add one - maybe you'd contribute that?

> add a field with the size of Auth_PASSWORD at the end of the hash, but unfortunately this option doesn't seem to be implemented.

As I understand, this was determined to be unnecessary and redundant - the length is determined from the hex encoding length of the existing field.

k4amos:
Hi solardiz, 
I have provided an example of o5logon authentication with a password longer than 16 characters `password1234567890` in this pull request: https://github.com/openwall/john-samples/pull/28
Here is the expected format of the hash to crack with john :

```
$o5logon$ <server's AUTH_SESSKEY> * <AUTH_VFR_DATA> * <AUTH_PASSWORD> * <client's AUTH_SESSKEY>
```

Here is what we have in the pcap sample : 
```
<server's AUTH_SESSKEY> --> 4D04DBD23D103F05D9B57EB6EC14D83A0A468AB906EAC907D3A8C796573E5F34BC15F0ECBC9EAC0350A38A663A368233
<AUTH_VFR_DATA> --> 442192E518F6F43D7CF7
<AUTH_PASSWORD> --> D3963B6AAED39C231BD5C92A10C0F146CA4784D1503A9B97598B31D33406390B7CA4F8B3EE5406A54C1842E4E63D1220
<client's AUTH_SESSKEY> --> 192AB7C9BA21C883824CF3D5BA073AC1129FD841E0AF6DF522C7EBDC52783CB8B97B792BFB6D9D743C7F4376FF0E7F93
```
Here is the obtained hash that we should be able to crack with john (but it doesn't work):

```
[...]

PR #5643, opened 2024-12-30
Fix of the bug with long Oracle passwords and creation of oracle2john.py

k4amos:
https://github.com/openwall/john/issues/5610


solardiz:
Thanks @k4amos. Can you please separate the two changes here (added tool vs. patched code) into separate commits?

Also, I am really not sure the trivial fix I suggested is correct. This needs further analysis. I wouldn't be surprised if we see ASan failures after you fix the syntax error (add missing comma after the added test vector). But I don't mind you giving this a try.

solardiz:
Oh, also we have a convention to prefix commit titles with subsystem names, e.g. `o5logon format: Fix the bug with long Oracle passwords`. For newly added subsystems, no need to prefix with their names, but you can use just `Add oracle2john.py`.

@exploide As usual, we'd appreciate your review of the Python script. Thank you! (And we'll need to remember to credit you prominently for all the Python work when we make the next numbered release.)

run/oracle2john.py
```diff
@@ -0,0 +1,133 @@
...
+
+import sys
+import argparse
+import re
+
+try:
+    import scapy.all as scapy
+except ImportError:
+    print(
+        "\033[91m[Error] Scapy seems to be missing, run 'pip install --user scapy' to install it\033[0m"
```
exploide: This should go to stderr, i.e. add the parameter `file=sys.stderr` to `print`.

run/oracle2john.py
```diff
@@ -0,0 +1,133 @@
...
+        "client_auth_sesskey": None,
+    }
+
+    packets = scapy.rdpcap(filename)
+    for packet in packets:
+        auth_data = process_packet(args, packet, auth_data)
+
+    if auth_data["server_auth_sesskey"] is None or auth_data["auth_vfr_data"] is None:
+        print(
+            f"\033[91m[Error] The PCAP file {filename} does not contain the required fields of o5logon authentification\033[0m"
```
exploide: Should also go to stderr to avoid having this in the hash output file when redirection is used. I would also consider this being a warning (or omit it altogether) because it might be a valid use case to throw this script on a bunch of pcap files containing arbitrary network traffic.

run/oracle2john.py
```diff
@@ -0,0 +1,133 @@
...
+        )
+    elif None not in list(auth_data.values()):
+        # Format of the hash : $o5logon$ <server's AUTH_SESSKEY> * <AUTH_VFR_DATA> * <AUTH_PASSWORD> * <client's AUTH_SESSKEY>
```
exploide, line 49: I'm not familiar with the o5logon protocol and don't know why the client values can be missing. Would it make sense to add a comment explaining the two cases?

run/oracle2john.py
```diff
@@ -0,0 +1,133 @@
...
+def process_packet(args, packet, auth_data):
+    """
+    Processes a packet and updates the auth_data dictionary with the extracted values.
+    """
+    raw_data = bytes(packet)
+
+    pbkdf2_match = re.search(rb"PBKDF2", raw_data)
+    if pbkdf2_match:
+        print(
+            "\033[91m[Error] Sorry, PBKDF2 is not (yet) supported by this utility (only o5logon is)\033[0m"
```
exploide: stderr again. But I'm wondering, it is that fatal that the program should exit? What if another connection is contained within the pcap? And the script seems to support multiple files as well. If the first cannot be processed, the latter will never be tried?

run/oracle2john.py
```diff
@@ -0,0 +1,133 @@
...
+
+def select_hexa(raw_string):
+    """
+    Extracts the first valid hexadecimal string from the raw data.
+    """
+    match_hexa = re.search(
+        "([A-Fa-f0-9]+)", raw_string.decode("ascii", errors="ignore").replace(" ", "")
+    )
+    if match_hexa:
+        return match_hexa.group(1)
```
exploide, line 72: explicitly `return None` otherwise to make the intention clear

run/oracle2john.py
```diff
@@ -0,0 +1,133 @@
...
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        epilog="""
+    ### Utility to obtain a hash of ORACLE authentication (o5logon) that can be cracked with John
+        Written by k4amos
+
+    Usage: ./oracle2john.py -f <pcap files>
+    """,
+    )
+
+    parser.add_argument("-f", "--file", type=str, required=True, nargs="+")
```
exploide: Since files are a mandatory and the only parameter the script expects, I would personally omit the flag-style option and just make it a positional argument instead. Most other *2john scripts also do it this way (but not all), so this is a matter of preferences.

run/oracle2john.py
```diff
@@ -0,0 +1,133 @@
...
+
+    parsed_args = parser.parse_args()
+    args = vars(parsed_args)
+
+    for filename in args["file"]:
+        auth_data = read_file(args, filename)
```
exploide: `read_file` does not return any `auth_data`, it looks rather like a `process_file`, prints the hash directly and returns nothing.

exploide (changes requested):
Just added a few comments which concern mostly possible minor improvements.

However, I have two major things for discussion.

1. There already is `pcap2john.py` which iterates through pcap files and outputs various hashes it can extract. In principle, o5logon support could have been added as a new function within `pcap2john.py`. But I see that there is something different, i.e. o5logon seems to assemble information contained within multiple different packets, if I understand the code correctly? Most (but not all) functions in `pcap2john.py` process single packet authentications, hence they don't need to keep state when parsing the next packet. So it might be fine to have this as a separate script?

2. I think the script is not very robust when parsing larger captures which contain more than just the o5logon procedure. Especially the regex searching within the raw data might be prone to false positives and negatives. Imagine a plain HTTP packet where the word PBKDF2 is contained on a web page. The script would immediately exit in its current form. Is o5logon always contained in TNS packets? I don't see scapy having support for this as a layer which would be nice for filter reasons. If we can't find a way to filter relevant packets in a better way, we may at least need to document that the pcap file should only contain the o5logon/TNS stuff for best results.

solardiz:
One thing I forgot to note - this is significant enough addition that we should have a `doc/NEWS` entry for it. Adding that can be part of this same PR.

magnumripper:
> 1. There already is `pcap2john.py` which iterates through pcap files and outputs various hashes it can extract. In principle, o5logon support could have been added as a new function within `pcap2john.py`. But I see that there is something different, i.e. o5logon seems to assemble information contained within multiple different packets, if I understand the code correctly? Most (but not all) functions in `pcap2john.py` process single packet authentications, hence they don't need to keep state when parsing the next packet. So it might be fine to have this as a separate script?

Yeah I think it's fine. If we had more manpower we should rewrite pcap2john.py to be more consistent about what external modules are used (it's currently a merge of several disparate utilities) and ideally support multi packet stuff as well.

src/o5logon_fmt_plug.c
```diff
@@ -131,7 +132,7 @@ static int valid(char *ciphertext, struct fmt_main *self)
 	if ((p = strtokm(NULL, "*"))) {	/* client's encrypted password */
 		int len = hexlenu(p, &extra);
 
-		if (extra || len < 64 || len % 32 || len > 2 * PLAINTEXT_LENGTH + 16)
+		if (extra || len < 64 || len % 32 || len > 3 * PLAINTEXT_LENGTH)
```
magnumripper: I didn't pay attention at first but I now realized I wrote the original line, lol.

I'm pretty sure the `+ 16` was meant to accommodate a full block of padding after a password length of exact multiple of block size. So `+ 16` should be `+ 32` (presuming I just overlooked that hex doubles the length) or better yet there should be parens as in `2 * (PLAINTEXT_LENGTH + 16)`. If we vary `PLAINTEXT_LENGTH` the `len > 2 * (PLAINTEXT_LENGTH + 16)` is more correct than `len > 3 * PLAINTEXT_LENGTH`.

<s>I had a quick glance at the code for possible places we should bump some array or such, and I believe there is:
```diff
diff --git a/src/o5logon_fmt_plug.c b/src/o5logon_fmt_plug.c
index 498f2d6b1..b5af3e887 100644
--- a/src/o5logon_fmt_plug.c
+++ b/src/o5logon_fmt_plug.c
@@ -83,7 +83,7 @@ static struct custom_salt {
 	unsigned char salt[SALT_LENGTH];         /* AUTH_VFR_DATA */
 	unsigned char ct[CIPHERTEXT_LENGTH];     /* Server's AUTH_SESSKEY */
 	unsigned char csk[CIPHERTEXT_LENGTH];    /* Client's AUTH_SESSKEY */
-	unsigned char pw[16 + PLAINTEXT_LENGTH]; /* Client's AUTH_PASSWORD */
+	unsigned char pw[2 * (PLAINTEXT_LENGTH + 16)]; /* Client's AUTH_PASSWORD */
 	int pw_len;                              /* AUTH_PASSWORD length (blocks) */
 } *cur_salt;
 
```
</s>

EDIT: Scrapped the last comment, that part is correct in current code (sizes are after hex decode).

magnumripper (changes requested):

src/o5logon_fmt_plug.c
magnumripper: <s>
> I had a quick glance at the code for possible places we should bump some array or such, and I believe there is:

...and BTW after that fix, the `valid()` line might be best written as `if (extra || len < 64 || len % 32 || len > sizeof(((struct custom_salt*)0)->pw))`.
</s>

We could make it `if (extra || len < 64 || len % 32 || len > 2 * sizeof(((struct custom_salt*)0)->pw))` but since we already take PLAINTEXT_LENGTH into account, I think it just makes the code less readable.

magnumripper (changes requested):
Oh, and you should `chmod +x oracle2john.py`.

Also, thanks a lot for your contribution! While Ettercap is a fine tool, we better not depend on it.

run/oracle2john.py
magnumripper: Agreed, the `-f` option should be dropped. Our convention is `whatever2john <file(s)>`, hashes to stdout and anything else to stderr.

> Most other *2john scripts also do it this way (but not all)

Which do not? We should open an issue for that!

exploide: According to my quick search, at least `radius2john.py` and `aix2john.py`.
Should I change these?

magnumripper: That would be terrific 👍

k4amos:
Hi everyone,

Thank you for your careful review @exploide  and for your comments @magnumripper

O5logon indeed assembles information from different packets. I agree with you that the regex are not super clean, but I couldn't do better with Scapy and with Pyshark, the problem would be the same.

I hope everything is good this time

run/oracle2john.py
```diff
@@ -0,0 +1,132 @@
...
+
+# Utility to obtain a hash of ORACLE authentication (o5logon) that can be cracked with John
+# This code is not compatible with the PBKDF2 key derivation function
+#
+# Usage: ./oracle2john.py <pcap files>
+#
+# This script depends on Scapy (https://scapy.net)
+# To install: pip install --user scapy
+#
+# Make the script executable: chmod +x oracle2john.py
```
exploide: I think what @magnumripper meant was to make this script executable before commiting it, so the comment should no longer be necessary after that. (The executable bit is the only file permission attribute tracked by git.) If you take a look in the `run/` folder, you see the other 2john scripts already have that bit set right after checkout.

exploide (approved):
Looks good, thanks for incorporating the requested changes :)

The only minor point is the executable bit.

magnumripper (approved):

PR #5649, opened 2025-01-06
Use positional argument for input files in 2john scripts

exploide:
As requested in https://github.com/openwall/john/pull/5643#discussion_r1903171219 this PR turns the input file arguments for `radius2john.py` and `aix2john.py` into positional arguments.

Plus a bit of cleanup while I was at it.

### DIFF
--- a/run/aix2john.py
+++ b/run/aix2john.py
@@ -17,7 +17,6 @@
 
 
 import argparse
-import binascii
 import sys
 import re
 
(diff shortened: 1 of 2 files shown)


marko608 (approved):

magnumripper (approved):
Thanks! I see radius2john also takes pcap so presumedly the reason for it not being a part of pcap2john is the potential need for `-l` and `-p`.

I tested radius2john with our samples, we don't seem to have any samples for AIX.

exploide:
> we don't seem to have any samples for AIX.

The only bit I found was in the comment at the beginning of the script. I copied that to a file and the script produced the same output on it.

PR #5654, opened 2025-01-27
Add robustness checks in RAR3 formats

### LINKED ISSUE #5653: Problem with $RAR3$*1* hash. John can't find known password. Very strange situation.
Hello.
I encountered an extremely strange but at the same time interesting situation.
I have a rather old RAR archive (2013, judging by the file properties, created in version RAR 4.*). 
The file names in the RAR archive are not encrypted.
Using John I got the hash of this archive. 
I apologize in advance, I am not sure that the rules allow publishing full hashes, but when studying other topics I saw similar examples. 
I believe that without a full indication of the hash the description of the problem will be incomplete and useless.
So, I got the following hash:

`$RAR3$*1*2597b58e3fafb7d9*99875cb5*816*2036*1*45b*****************************3051f46e9ed0*33`

Using this hash in Hashcat I got the password: 861582585

I used this password on the RAR archive, however, when using it I was able to decrypt only one file (out of a total of about 20 files). The remaining files remained encrypted.
As far as I understand, this archive is a structure in which files were added one by one with different passwords.
Next, I tried to attack this hash using John, using a dictionary containing an already known password and by brute force. 
John does not find it. 
To exclude the possibility of a build error, I tried it on three different systems (MAC, Kali, Debian - build from repo, my main system is MAC M1), the result is the same. 
The most interesting thing is that I created a test RAR, where I created a similar structure, adding several files with different passwords (I also used version RAR 4.*). After that, I also took a hash from it using John. According to this hash, the password (the password of the last added file) was found by both John and Hashcat.

This situation confused me and I did the following.

I deleted this one file from the archive, for which I had the password, and made the RAR hash again using John.
I found that after deleting this one file with a known password, the hash of the overall RAP file changed.

`$RAR3$*1*2597b58e3fafb7d9*0731a89e*1728*3992*1*978554*********************b55f8dfd193a99100*33`

I used this new hash again in Hashcat and got another password: ivanmendezl

This password again only unlocked one file, the rest of the files remained encrypted.
The situation repeated itself completely, John could not find this password.
Later, when working with the remaining files manually, by gradually deleting files, I found that this archive has 8 different passwords, 2 of which I already know. 
I am currently trying to get them using Hashcat (except for two hashes that are too large and are not accepted by Hashcat, accordingly - my only chance to get passwords from them is to use John).
As far as I understand, with this structure, John can get the hash of the last file added to the RAR with a password. He does not see the rest, but when deleting that very last file and accordingly changing the general structure of the RAR archive, John sees the hash of the previous file added and so on chronologically.
To sum it up, I encountered such a situation for the first time and I can't understand why John can't get the password from the hashes. 
I exclude a false positive of Hashcat because 2 different files were successfully decrypted. 
In addition, John's successful work on test archives with the conditions met makes me think. 
It is quite possible that I am doing something wrong. 
I really ask for help from the community and would be grateful for any advice.

magnumripper:
Fixes #5653

### DIFF
--- a/src/rar_common.c
+++ b/src/rar_common.c
@@ -1,6 +1,6 @@
 /*
  * This software is Copyright (c) 2011, Dhiru Kholia <dhiru.kholia at gmail.com>
- * and Copyright (c) 2012-2020, magnum
+ * and Copyright (c) 2012-2025, magnum
  * and it is hereby released to the general public under the following terms:
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted.
@@ -37,6 +37,7 @@ static unsigned char *aes_iv;
 #define FORMAT_TAG_LEN      (sizeof(FORMAT_TAG)-1)
 
 #define YEL	"\x1b[0;33m"
+#define RED	"\x1b[0;31m"
 #define NRM	"\x1b[0m"
 
 static struct fmt_tests cpu_tests[] = {
@@ -302,7 +303,7 @@ static void *get_binary(char *ciphertext)
 		} else if (file->type == 0)
 			fprintf(stderr, YEL "%.32s(...) solid\n" NRM, ciphertext);
 		else
-			fprintf(stderr, YEL "%.32s(...) 0x%02x size %"PRIu64"\n" NRM, ciphertext, file->method, file->pack_size);
+			fprintf(stderr, YEL "%.32s(...) 0x%02x size %"PRIu64" unp_size %"PRIu64"\n" NRM, ciphertext, file->method, file->pack_size, file->unp_size);
 	}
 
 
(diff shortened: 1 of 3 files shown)


solardiz:
> I synced our code with the very last plain C code that was used in ClamAV for now (very small changes). I then made the effective [retval] changes after comparing with the newer C++ code.

This could be two commits.

magnumripper:
> This could be two commits.

And is now. BTW I need to test this a bit more before merging.

solardiz (approved):

magnumripper:
3b92a406 is not likely a fix we really want (although it doesn't hurt much really) as I suspect the issue is false positives and/or a corrupted archive. That story continues in #5653.

This PR will instead end up with some other goodies found in the process, stay tuned.

magnumripper:
So #5653 wasn't a bug (except in hashcat it was) - but the extra tests added here just might add a little more early rejection 👍  and I'm sure ASan will be slightly less furious over the unrar code now.

And in the end, this made me NOT want to replace the old ClamAV code with official unrar C++ code. The latter, as seen in hashcat, doesn't return a true or false like our current ClamAV code did to start with - and now with many more false (reject) conditions added by (the royal) us.

Issue #5652, opened 2025-01-22
Add support for encrypted LibreOffice 24.2+ documents


qw42:
libreoffice2john.py  output is "foo.otd is not an encrypted OpenOffice file!"

Version: 24.8.4.2 (X86_64) / LibreOffice Community
Build ID: bb3cfa12c7b1bf994ecc5649a80400d06cd71002
CPU threads: 4; OS: Windows 10 X86_64 (10.0 build 20348); UI render: Skia/Raster; VCL: win
Locale: en-US (en_US); UI: en-US
Calc: threaded

solardiz:
Thank you for reporting this @qw42!

Can you share a sample file (without security-sensitive info in it, and with a known password you'd also share)? Ideally, via a pull request to https://github.com/openwall/john-samples adding the new file(s) somewhere under `Office/`.

Is this problem seen with the latest `libreoffice2john.py` as found in this repo, or are you possibly using some release/archive/package of John the Ripper (if so, which one and what version)? What version of Python are you running this with?

exploide:
Just stumbled upon this: https://wiki.documentfoundation.org/ReleaseNotes/24.8#ODF_Wholesome_Encryption
Seems like the encryption mechanism was completely overhauled.

qw42:
Attaching the file which caused the failure. It was created using LibreOffice v 24.8.

[pwd_123.odt](https://github.com/user-attachments/files/18509598/pwd_123.odt)

solardiz:
> Just stumbled upon this: https://wiki.documentfoundation.org/ReleaseNotes/24.8#ODF_Wholesome_Encryption

```
ODF Wholesome Encryption

A new mode of password-based ODF encryption has been implemented, with the following advantages:

    more performant due to deriving a key only once per package
    more tamper-resistant with authenticated encryption (AES-GCM)
    better hiding of metadata to reduce information leaks
    higher resistance to brute forcing using memory-hard Argon2id key derivation function (Michael Stahl, allotropia) [tdf#105844](https://bugs.documentfoundation.org/show_bug.cgi?id=105844)

Warning:
You will need LibreOffice 24.2 or newer to open these encrypted files.
```

ghost:
We can easily warn about this inability:

```diff
From 824d6f59829f41a58a927641b35b84ecd5d0dfb1 Mon Sep 17 00:00:00 2001
From: =?UTF-8?q?Claudio=20Andr=C3=A9?= <dev@claudioandre.slmail.me>
[...]

PR #5655, opened 2025-01-31
libreoffice2john.py: error in encrypted 24.2+ files

ghost:
Warn that LibreOffice 24.2+ encrypted documents are not supported.

See: #5652.

### DIFF
--- a/run/libreoffice2john.py
+++ b/run/libreoffice2john.py
@@ -46,8 +46,26 @@ def process_file(filename):
         elements = list(r.getiterator())
 
     is_encrypted = False
+    is_supported = True
     key_size = 16
     start_key_generation_name = "SHA1"
+
+    for element in elements:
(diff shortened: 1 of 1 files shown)


run/libreoffice2john.py
```diff
@@ -46,8 +46,27 @@ def process_file(filename):
         elements = list(r.getiterator())
 
     is_encrypted = False
+    is_supported = True
     key_size = 16
     start_key_generation_name = "SHA1"
+
+    for i in range(0, len(elements)):
+        element = elements[i]
```
exploide: Minor stylistic comment: `for` loops in Python are for-each loops. So a bit simpler and more idiomatic without the need for indices would be the following.

```suggestion
    for element in elements:
```

solardiz (approved):

run/libreoffice2john.py
solardiz: Thank you for helping review this PR @exploide!

PR #5665, opened 2025-02-10
logger.c: Retry file lock also after ENOLCK

magnumripper:
This means "No locks available".  We now handle it just like EAGAIN, retrying and hoping for recover.

Also adds diagnostic output of the lock delay when this happens.

### DIFF
--- a/src/logger.c
+++ b/src/logger.c
@@ -58,6 +58,7 @@
 #include "cracker.h"
 #include "signals.h"
 #include "logger.h"
+#include "timer.h"
 
 static int cfg_beep;
 static int cfg_log_passwords;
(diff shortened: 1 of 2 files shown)


src/logger.c
```diff
@@ -136,24 +137,37 @@ int log_lock(int fd, int cmd, int type, const char *name,
...
-				fprintf(stderr, "Node %d: File locking apparently exhausted, check ulimits and any NFS server limits. This is recoverable but will harm performance (muting further of these messages from same node)\n", NODE);
+			if (!(warned & warn_bit)) {
+				if (!lock_start)
+					lock_start = john_get_nano();
+
+				log_event("- Got %s despite F_SETLKW on %s, %s() %s:%d", errno == EAGAIN ? "EAGAIN" : "ENOLCK", name, function, source_file, line);
+				if (options.node_count)
+					fprintf(stderr, "%u: ", NODE);
+				fprintf(stderr, "File locking apparently exhausted (%s), retrying. Check ulimits and any NFS server limits\n", strerror(errno));
 				srand(NODE);
```
solardiz: I suggest dropping the `srand(NODE);` line from here. I think it actually hurts. Previously, it was only reached once because the `warned` variable was static - but now that you make the warning bits per-call, allowing for repeated log messages from subsequent calls, you'd also end up resetting the PRNG to the same state multiple times.

src/logger.c
```diff
@@ -136,24 +137,37 @@ int log_lock(int fd, int cmd, int type, const char *name,
...
+
+				warned |= warn_bit;
 			}
 
 			/* Sleep for a random time of max. ~260 ms */
 			t.tv_sec = 0; t.tv_usec = (rand() & 1023) << 8;
```
solardiz, line 166: Oh, I see by using `srand(NODE);` this code tries to have different delays for different nodes to avoid clashes. However, besides the issue I just pointed out, this is also not great because this PRNG is shared with other uses we have in the tree:

```
./logger.c:			t.tv_sec = 0; t.tv_usec = (rand() & 1023) << 8;
./dynamic_compiler.c:		int j = rand() % 16;
./oaes_lib_plug.c:		_key->data[_i] = (uint8_t) rand( _ctx->rctx );
./oaes_lib_plug.c:		_key->data[_i] = (uint8_t) rand();
./oaes_lib_plug.c:					_ctx->iv[_i] = (uint8_t) rand( _ctx->rctx );
./oaes_lib_plug.c:					_ctx->iv[_i] = (uint8_t) rand();
```

```
./logger.c:				srand(NODE);
./oaes_lib_plug.c:	srand(0); // XXX
```

We could also want to check those other uses and maybe get rid of (some of) them.

solardiz: > We could also want to check those other uses and maybe get rid of (some of) them.

Got rid of those in `oaes_lib_plug.c` since they didn't affect anything in our usage anyway.

src/logger.c
solardiz: Well, maybe limit the `srand(NODE);` to be done just once per process like it was before?

magnumripper:
So I guess I'll just ensure they are executed only once like you said. Perhaps even put a `srand(NODE)` (or something better than NODE) in john_init or something like that?

The use of prng in dynamic_compiler is just for building test vector passwords.

solardiz:
> Perhaps even put a `srand(NODE)` (or something better than NODE) in john_init or something like that?

I don't mind.

magnumripper:
Today I could confirm IRL that this works as intended, and that a retry on ENOLCK can actually recover.

PR #5682, opened 2025-03-04
Oubliette: expanded supported database version to 3.x (Oubliette v1.5)

davidedg:
Hi.
Minor patch to support latest (last) version of Oubliette.

### DIFF
--- a/src/oubliette_blowfish_fmt_plug.c
+++ b/src/oubliette_blowfish_fmt_plug.c
@@ -107,7 +107,7 @@ static int valid(char *ciphertext, struct fmt_main *self)
 		goto err;
 	if (sscanf(p, "%d.%d", &version, &minor) != 2)
 		goto err;
-	if (version != 1 || minor != 0)
+	if (version < 1 || version > 3)
 		goto err;
 
 	if ((p = strtokm(NULL, "$")) == NULL)	/* hash */
--- a/src/oubliette_idea_fmt_plug.c
+++ b/src/oubliette_idea_fmt_plug.c
@@ -105,7 +105,7 @@ static int valid(char *ciphertext, struct fmt_main *self)
 		goto err;
 	if (sscanf(p, "%d.%d", &version, &minor) != 2)
 		goto err;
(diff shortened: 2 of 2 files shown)


solardiz (approved):
The _change_ looks good to me, but reviewing it I realize we have a problem here:

In this `valid`, we don't fully validate the string's syntax (`sscanf` will just stop at a character it can't match against the format string, plus we don't check there's nothing after what we think is the last field), yet in `get_binary` we rely on proper syntax (length of the last field) having been checked. We need to fix this separately.

solardiz:
The behavior is a bit weird now. When run on all the samples at once, this cracks all 3 "hashes" (for the current format) including both `$oubliette-blowfish$1.0$82ea2c652362c380349d6ac0d5a06d2ae9b164becc5a75103c2744b4d27fe35d` and `$oubliette-blowfish$3.0$82ea2c652362c380349d6ac0d5a06d2ae9b164becc5a75103c2744b4d27fe35d`, but only the former one of these two is written to `john.pot` at first. It takes a second invocation, with that `john.pot` already in place, for the second "hash" to also be written to there (and thus for `--show` to work on it).

This must be because the version isn't decoded in `get_binary` output, so "binary hashes" look the same regardless of version. But then I'm surprised our loader doesn't appear to de-duplicate these "hashes" - I'd expect it to load only 2 and not 3.

Anyway, the fix may be either to introduce a `split` that would unify the hash encodings (make them all look like version 1 for writing into `john.pot`) or to exclude version info from what the Python script outputs (and change the code not to expect version info). If we're supporting literally all versions of Oubliette now and no new versions are expected, then perhaps we can move the version check into the Python script?

davidedg:
> or to exclude version info from what the Python script outputs (and change the code not to expect version info). If we're supporting literally all versions of Oubliette now and no new versions are expected, then perhaps we can move the version check into the Python script?

I agree, the original program is long since dead anyway so no new versions are to be expected.
I am testing a new version with no dependencies on version number and should be creating a PR soon.

PR #5689, opened 2025-03-05
Update fvde2john.py

holly-o:
Fix path error on Windows when locating EncryptedRoot.plist.wipekey - instead of recursive traversal of filesystem, navigate to EncryptedRoot.plist.wipekey directly.

### DIFF
--- a/run/fvde2john.py
+++ b/run/fvde2john.py
@@ -9,6 +9,7 @@ import plistlib
 import os
 import argparse
 import sys
+import re
 
 try:
     from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
@@ -24,6 +25,7 @@ except ImportError:
 HEX_CORE_STORAGE_TYPE_GUID = '53746F72-6167-11AA-AA11-00306543ECAC'
 HEX_APPLE_BOOT_STORAGE_TYPE_GUID = '426F6F74-0000-11AA-AA11-00306543ECAC'
 LOCAL_USER_TYPE_ID = 0x10060002
+BOOT_DIR_REGEX = re.compile(r'com.apple.boot.(?P<boot_letter>[A-Z])')
 
 def uint_to_int(b):
     return int(b[::-1].hex(), 16)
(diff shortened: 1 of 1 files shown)


solardiz:
Thank you, @holly-o!
@exploide Can you please take a look?

> Fix path error on Windows when locating EncryptedRoot.plist.wipekey - instead of recursive traversal of filesystem, navigate to EncryptedRoot.plist.wipekey directly.

What was the error? This otherwise looks like an optimization, but the word "error" suggests the file previously could not be located on Windows?

There's a comment saying:

> The boot volume contains a file called `EncryptedRoot.plist.wipekey`. This is stored on the volume at `/com.apple.boot.X/System/Library/Caches/com.apple.corestorage/EncryptedRoot.plist.wipekey`, where `X` is variable but is often `P` or `R`.

but here you add a regexp that seems to check for `P` only (not `R`)?

And what's the deal about `<boot_letter>[A-Z]`?

Also, why would this be Windows-specific when we're parsing a filesystem image that's expected to already have these names in it, regardless of what OS it's being parsed on?

exploide:
> @exploide Can you please take a look?

I have not done a practical check and do not really know the format but form reading the code it looks fine.

`(?P<boot_letter>[A-Z])` has nothing to do with the drive letter P but is Python regular expression syntax.

> `(?P<name>...)`  Similar to regular parentheses, but the substring matched by the group is accessible via the symbolic group name name.

solardiz:
> `(?P<boot_letter>[A-Z])` has nothing to do with the drive letter P but is Python regular expression syntax.
> 
> > `(?P<name>...)`  Similar to regular parentheses, but the substring matched by the group is accessible via the symbolic group name name.

Thanks! It looks like this code isn't actually using this feature for real - I see no other mentions of `boot_letter`. So is it used as a way to insert a comment into the regexp? Do we want that?

Anyway, I don't mind.

holly-o:
Yes just a way of labelling the regex - it could be removed I don't mind.

I had errors on Windows due to the reading the path '/' as part of the local folder structure and not the image. I believe this update will be better anyway as it directly reads the file from its location rather than traversing the filesystem.

exploide (approved):
Yes, the `boot_letter` name is not really used but it doesn't hurt either. I don't really care.

PR #5694, opened 2025-03-10
Enable AVX512VL + AVX512DQ

ghost:
Let's hear bots.

I'll remove `runstatedir` after testing.
****

```sh
checking for AVX2... yes
checking for AVX512BW + AVX512VL + AVX512DQ... yes
checking if gcc supports -maes -mpclmul... yes
```
OR:

```sh
checking for AVX2... yes
checking for AVX512BW + AVX512VL + AVX512DQ... no
checking for AVX512F... no
checking if gcc supports -maes -mpclmul... yes
```

```sh
Target CPU ......................................... x86_64 AVX512BW, 64-bit LE
```

```sh
Version: 1.9.0-jumbo-1+bleeding-60f3614a06 2025-03-10 07:27:48 -0300
Build: linux-gnu 64-bit x86_64 AVX512(BW+VL+DQ) AC OMP
SIMD: AVX512BW, interleaving: MD4:3 MD5:3 SHA1:1 SHA256:1 SHA512:1
AES hardware acceleration: AES-NI
CPU tests: AVX512(BW+VL+DQ)
$JOHN is ../run/
```

****
```
Sorry, AVX512(BW+VL+DQ) is required for this build


John the Ripper 1.9.0-jumbo-1+bleeding-7146e4c827 2025-03-12 05:14:17 +0100 OMP [linux-gnu 64-bit x86_64 AVX2 AC]
Copyright (c) 1996-2025 by Solar Designer and others
Homepage: https://www.openwall.com/john/
Usage: john-avx2-omp [OPTIONS] [PASSWORD-FILES]
Use --help to list all available options.
```

### DIFF
--- a/doc/NEWS
+++ b/doc/NEWS
@@ -420,6 +420,10 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
 - Add Oubliette Password Manager support (two formats and oubliette2john.py).
   [DavideDG; 2025]
 
+- Use AVX512VL XOP-like bit rotates for scrypt's Salsa20.  [Solar; 2025]
+
+- When we use AVX512BW, also enable usage of AVX512VL and AVX512DQ.  [Claudio André; 2025]
+
 
 Major changes from 1.8.0-jumbo-1 (December 2014) to 1.9.0-jumbo-1 (May 2019):
 
--- a/src/configure
+++ b/src/configure
@@ -9224,7 +9224,62 @@ ac_compiler_gnu=$ac_cv_c_compiler_gnu
 
     SIMD_NAME="Altivec2"
     ;;
-    mmx|sse*|ssse3|avx*|xop*)
+    avx512|avx512bw)
+    SIMD_NAME="avx512bw"
+    { $as_echo "$as_me:${as_lineno-$LINENO}: Trying to force $SIMD_NAME using default method (--enable-simd=$simd)." >&5
+$as_echo "$as_me: Trying to force $SIMD_NAME using default method (--enable-simd=$simd)." >&6;}
+    if test 2 -gt 0; then :
+  { $as_echo "$as_me:${as_lineno-$LINENO}: checking if $CC supports -mavx512bw -mavx512vl -mavx512dq w/ linking" >&5
+$as_echo_n "checking if $CC supports -mavx512bw -mavx512vl -mavx512dq w/ linking... " >&6; }
+fi
+  ac_ext=c
+ac_cpp='$CPP $CPPFLAGS'
+ac_compile='$CC -c $CFLAGS $CPPFLAGS conftest.$ac_ext >&5'
+ac_link='$CC -o conftest$ac_exeext $CFLAGS $CPPFLAGS $LDFLAGS conftest.$ac_ext $LIBS >&5'
+ac_compiler_gnu=$ac_cv_c_compiler_gnu
+
+  ac_saved_cflags="$CFLAGS"
+  CFLAGS="-Werror -mavx512bw -mavx512vl -mavx512dq"
+  cat confdefs.h - <<_ACEOF >conftest.$ac_ext
+/* end confdefs.h.  */
+
+int
+main ()
+{
+
+  ;
+  return 0;
+}
+_ACEOF
+if ac_fn_c_try_link "$LINENO"; then :
+  if test "2" -gt 0; then :
+  { $as_echo "$as_me:${as_lineno-$LINENO}: result: yes" >&5
+$as_echo "yes" >&6; }
+fi
+      CFLAGS_EX="$CFLAGS_EX -mavx512bw -mavx512vl -mavx512dq"
+
+else
+  if test 2 -gt 0; then :
+  { $as_echo "$as_me:${as_lineno-$LINENO}: result: no" >&5
+$as_echo "no" >&6; }
+fi
+    if test "2" = 2; then :
+  as_fn_error $? "Not supported by compiler/linker" "$LINENO" 5
+fi
+
+fi
+rm -f core conftest.err conftest.$ac_objext \
+    conftest$ac_exeext conftest.$ac_ext
+  CFLAGS="$ac_saved_cflags"
+  ac_ext=c
+ac_cpp='$CPP $CPPFLAGS'
+ac_compile='$CC -c $CFLAGS $CPPFLAGS conftest.$ac_ext >&5'
+ac_link='$CC -o conftest$ac_exeext $CFLAGS $CPPFLAGS $LDFLAGS conftest.$ac_ext $LIBS >&5'
+ac_compiler_gnu=$ac_cv_c_compiler_gnu
+
+
+    ;;
+  mmx|sse*|ssse3|avx*|xop*)
     SIMD_NAME=`echo $simd | tr a-z A-Z`
     { $as_echo "$as_me:${as_lineno-$LINENO}: Trying to force $SIMD_NAME using default method (--enable-simd=$simd)." >&5
 $as_echo "$as_me: Trying to force $SIMD_NAME using default method (--enable-simd=$simd)." >&6;}
@@ -11485,10 +11540,10 @@ fi
 
   if test "x$CPU_NOTFOUND" = x0; then :
 
-  CFLAGS="$CFLAGS_BACKUP -mavx512bw -P $EXTRA_AS_FLAGS $CPPFLAGS $CFLAGS_EXTRA $CPUID_ASM"
+  CFLAGS="$CFLAGS_BACKUP -mavx512bw -mavx512vl -mavx512dq -P $EXTRA_AS_FLAGS $CPPFLAGS $CFLAGS_EXTRA $CPUID_ASM"
 
-  { $as_echo "$as_me:${as_lineno-$LINENO}: checking for AVX512BW" >&5
-$as_echo_n "checking for AVX512BW... " >&6; }
+  { $as_echo "$as_me:${as_lineno-$LINENO}: checking for AVX512BW + AVX512VL + AVX512DQ" >&5
+$as_echo_n "checking for AVX512BW + AVX512VL + AVX512DQ... " >&6; }
   if test "$cross_compiling" = yes; then :
   { { $as_echo "$as_me:${as_lineno-$LINENO}: error: in \`$ac_pwd':" >&5
 $as_echo "$as_me: error: in \`$ac_pwd':" >&2;}
@@ -11503,7 +11558,7 @@ extern int CPU_detect(void); extern char CPU_req_name[];
 
 _ACEOF
 if ac_fn_c_try_run "$LINENO"; then :
-  CPU_BEST_FLAGS="-mavx512bw"
+  CPU_BEST_FLAGS="-mavx512bw -mavx512vl -mavx512dq"
      SIMD_NAME="AVX512BW"
      { $as_echo "$as_me:${as_lineno-$LINENO}: result: yes" >&5
 $as_echo "yes" >&6; }
@@ -11810,20 +11865,21 @@ fi
 
   if test "x$CPU_NOTFOUND" = x0; then :
 
-  { $as_echo "$as_me:${as_lineno-$LINENO}: checking for AVX512BW" >&5
-$as_echo_n "checking for AVX512BW... " >&6; }
+  { $as_echo "$as_me:${as_lineno-$LINENO}: checking for AVX512BW + AVX512VL + AVX512DQ" >&5
+$as_echo_n "checking for AVX512BW + AVX512VL + AVX512DQ... " >&6; }
   cat confdefs.h - <<_ACEOF >conftest.$ac_ext
 /* end confdefs.h.  */
 
     #include <immintrin.h>
         #include <stdio.h>
         extern void exit(int);
-        int main(){__m512i t=_mm512_slli_epi16(_mm512_set1_epi16(1),1);exit(!(_mm_cvtsi128_si64x(_mm512_extracti32x4_epi32(t,0))==0x2000200020002ULL));}
+        int main(){__m512i t=_mm512_slli_epi16(_mm512_set1_epi16(1),1);__int64_t ret_value=(!(_mm_cvtsi128_si64x(_mm512_extracti32x4_epi32(t,0))==0x2000200020002ULL));
+                   __m128i u, t1;*((long long*)&u)=1;t1=u;u=_mm_rol_epi32(t1,1);if((*(long long*)&t)==88)printf(".");exit(ret_value);}
 
 
 _ACEOF
 if ac_fn_c_try_link "$LINENO"; then :
-  CPU_BEST_FLAGS="-mavx512bw"
+  CPU_BEST_FLAGS="-mavx512bw -mavx512vl -mavx512dq"
      SIMD_NAME="AVX512BW"
      { $as_echo "$as_me:${as_lineno-$LINENO}: result: yes" >&5
 $as_echo "yes" >&6; }
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -439,6 +439,11 @@ case "$simd" in
     SIMD_NAME="Altivec2"
     ;;
   dnl Handle known cases of --enable-simd=foo --> -mfoo
+  avx512|avx512bw)
+    SIMD_NAME="avx512bw"
+    AC_MSG_NOTICE([Trying to force $SIMD_NAME using default method (--enable-simd=$simd).])
+    JTR_FLAG_CHECK_LINK([-mavx512bw -mavx512vl -mavx512dq], 2)
+    ;;
   mmx|sse*|ssse3|avx*|xop*)
     SIMD_NAME=`echo $simd | tr a-z A-Z`
     AC_MSG_NOTICE([Trying to force $SIMD_NAME using default method (--enable-simd=$simd).])
--- a/src/m4/jtr_x86_logic.m4
+++ b/src/m4/jtr_x86_logic.m4
@@ -268,15 +268,15 @@ if test "x$simd" != xno; then
 
   AS_IF([test "x$CPU_NOTFOUND" = x0],
   [
-  CFLAGS="$CFLAGS_BACKUP -mavx512bw -P $EXTRA_AS_FLAGS $CPPFLAGS $CFLAGS_EXTRA $CPUID_ASM"
+  CFLAGS="$CFLAGS_BACKUP -mavx512bw -mavx512vl -mavx512dq -P $EXTRA_AS_FLAGS $CPPFLAGS $CFLAGS_EXTRA $CPUID_ASM"
 
-  AC_MSG_CHECKING([for AVX512BW])
+  AC_MSG_CHECKING([for AVX512BW + AVX512VL + AVX512DQ])
   AC_RUN_IFELSE([AC_LANG_SOURCE(
     [[extern int CPU_detect(void); extern char CPU_req_name[];
       unsigned int nt_buffer8x[4], output8x[4];
       int main(int argc, char **argv) { return !CPU_detect(); }
     ]])],
-     [CPU_BEST_FLAGS="-mavx512bw"]
+     [CPU_BEST_FLAGS="-mavx512bw -mavx512vl -mavx512dq"]
      [SIMD_NAME="AVX512BW"]
      [AC_MSG_RESULT([yes])]
     ,[AC_MSG_RESULT([no])]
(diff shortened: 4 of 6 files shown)


src/m4/jtr_x86_logic.m4
```diff
@@ -444,42 +445,44 @@ dnl ======================================================================
...
+        int main(){__m512i t, t1;*((long long*)&t)=1;t1=t;t=_mm512_mul_epi32(t1,t);if((*(long long*)&t)==88)printf(".");exit(0);}]] // TODO: still missing
     )]
-    ,[CPU_BEST_FLAGS="-mavx512f"]
-     [SIMD_NAME="AVX512F"]
+    ,[CPU_BEST_FLAGS="-mavx512bw -mavx512vl -mavx512dq"]
+     [SIMD_NAME="AVX512(BW+VL+DQ)"]
      [AC_MSG_RESULT([yes])]
-    ,[CPU_NOTFOUND=1]
-     [AC_MSG_RESULT([no])]
+     [CPU_NOTFOUND=1]
```
claudioandre-br, line 460: This works like a `CPU_FOUND=1`.

ghost:
CI is happy. Nothing bad so far.

ghost:
It seems it tests CPU support twice (as seen below):

```sh
configure: Trying to force avx512bw using default method (--enable-simd=avx512bw).
checking if gcc supports -mavx512bw -mavx512vl -mavx512dq w/ linking... yes
checking for X32 ABI... no
checking for MMX... yes
checking for SSE2... yes
checking for SSSE3... yes
checking for SSE4.1... yes
checking for SSE4.2... yes
checking for AVX... yes
checking for XOP... no
checking for AVX2... yes
checking for AVX512BW + AVX512VL + AVX512DQ... yes
checking if gcc supports -maes -mpclmul... yes
```

It doesn't hurt.

doc/NEWS
```diff
@@ -420,6 +420,7 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
 - Add Oubliette Password Manager support (two formats and oubliette2john.py).
   [DavideDG; 2025]
 
+- Turn AVX512 into AVX512BW + AVX512VL + AVX512DQ. [Claudio André; 2025]
```
solardiz: That's confusing. I suggest:

```
- Use AVX512VL XOP-like bit rotates for scrypt's Salsa20.  [Solar; 2025]

- When we use AVX512BW, also enable usage of AVX512VL and AVX512DQ.  [Claudio André; 2025]

```

src/configure.ac
```diff
@@ -902,7 +907,16 @@ AC_CHECK_FUNCS([posix_memalign aligned_alloc memalign __mingw_aligned_malloc _al
...
+  if echo ${SIMD_NAME} | ${GREP} -q "AVX512("; then
+    types=$(echo "${SIMD_NAME}" | ${GREP} -o '(.*)' | tr -d "()" | tr "+" "\n")
+
+    for ext in $types
+    do
+      CPU_BEST_FLAGS_MAIN+="-DJOHN_$(echo AVX512$ext) "
+    done
+  else
+    CPU_BEST_FLAGS_MAIN=-DJOHN_$(echo ${SIMD_NAME} | tr .a-z _A-Z)
+  fi
```
solardiz: I doubt we need this complication. Can't we just continue with `JOHN_AVX512BW` alone, but understand that it implies VL and DQ? I also don't know whether the `+=` syntax works with other shells.

src/configure.ac
```diff
@@ -1163,7 +1177,11 @@ else
 fi
 
 if test -n "$SIMD_NAME"; then
-   CPU_NAME="$host_cpu $SIMD_NAME"
+   if echo ${SIMD_NAME} | ${GREP} -q "AVX512("; then
+      CPU_NAME="$host_cpu AVX512BW"
+   else
+      CPU_NAME="$host_cpu $SIMD_NAME"
+   fi
```
solardiz: Looks like unneeded complication as well.

src/m4/jtr_x86_logic.m4
```diff
@@ -249,37 +249,38 @@ if test "x$simd" != xno; then
 
   AS_IF([test "x$CPU_NOTFOUND" = x0],
   [
-  CFLAGS="$CFLAGS_BACKUP -mavx512f -P $EXTRA_AS_FLAGS $CPPFLAGS $CFLAGS_EXTRA $CPUID_ASM"
+  CFLAGS="$CFLAGS_BACKUP -mavx512bw -mavx512vl -mavx512dq -P $EXTRA_AS_FLAGS $CPPFLAGS $CFLAGS_EXTRA $CPUID_ASM"
```
solardiz: If we're not implementing the full reverse order of checks + optimization, then maybe let's not reorder F vs. BW here? If we were checking F first, then continue to check it first. This PR's changes would be smaller then.

src/m4/jtr_x86_logic.m4
```diff
@@ -249,37 +249,38 @@ if test "x$simd" != xno; then
...
+  AC_MSG_CHECKING([for AVX512BW + AVX512VL + AVX512DQ])
   AC_RUN_IFELSE([AC_LANG_SOURCE(
     [[extern int CPU_detect(void); extern char CPU_req_name[];
       unsigned int nt_buffer8x[4], output8x[4];
       int main(int argc, char **argv) { return !CPU_detect(); }
     ]])],
-     [CPU_BEST_FLAGS="-mavx512f"]
-     [SIMD_NAME="AVX512F"]
+     [CPU_BEST_FLAGS="-mavx512bw -mavx512vl -mavx512dq"]
+     [SIMD_NAME="AVX512(BW+VL+DQ)"]
```
solardiz: Maybe continue to say just `AVX512BW` here.

src/m4/jtr_x86_logic.m4
```diff
@@ -444,42 +445,44 @@ dnl ======================================================================
...
-  AC_MSG_CHECKING([for AVX512F])
+  AC_MSG_CHECKING([for AVX512BW + AVX512VL + AVX512DQ])
   AC_LINK_IFELSE(
     [
     AC_LANG_SOURCE(
       [[#include <immintrin.h>
         #include <stdio.h>
         extern void exit(int);
-        int main(){__m512i t, t1;*((long long*)&t)=1;t1=t;t=_mm512_mul_epi32(t1,t);if((*(long long*)&t)==88)printf(".");exit(0);}]]
+        int main(){__m128i t, t1;*((long long*)&t)=1;t1=t;t=_mm_rol_epi32(t1,1);if((*(long long*)&t)==88)printf(".");exit(0);}]]
```
solardiz: I did suggest using the same intrinsic we actually use, but I didn't mean to test it instead of testing any 512-bit BW intrinsic. I think we should either revert this change entirely or test both `_mm_rol_epi32` and `_mm512_mul_epi32`.

While there are no current nor planned CPUs that have BW without VL nor vice versa, there may be future CPUs supporting AVX10/256 where the 128-bit VL intrinsic would compile and run yet this wouldn't imply support for 512-bit BW. Such future CPUs wouldn't set the CPUID bit corresponding to VL, but here we're not checking CPUID at all.

src/x86-64.h
```diff
@@ -74,11 +76,13 @@
...
 
-#ifdef JOHN_AVX512BW
+#if defined(JOHN_AVX512BW) || defined(JOHN_AVX512VL) || defined(JOHN_AVX512DQ)
 #define CPU_DETECT			1
 #define CPU_REQ				1
 #define CPU_REQ_AVX512BW		1
-#define CPU_NAME			"AVX512BW"
+#define CPU_REQ_AVX512VL		1
+#define CPU_REQ_AVX512DQ		1
+#define CPU_NAME			"AVX512(BW+VL+DQ)"
```
solardiz, line 83: We can keep all 3 mentioned in `CPU_NAME`, for reporting in the "Sorry" line. (No further change is needed here.)

solardiz (changes requested):

src/m4/jtr_x86_logic.m4
solardiz: Oh, I didn't realize you previously got the `_mm512_mul_epi32` from the section for F, not for BW. Then revert to what we were checking for BW, please.

src/x86-64.h
ghost: The addition of `CPU_REQ_AVX512VL` (+DQ) is also required. At least desired.

src/configure.ac
ghost: This is possible if everyone understands that BW implies the rest.

doc/NEWS
```diff
@@ -420,6 +420,9 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
 - Add Oubliette Password Manager support (two formats and oubliette2john.py).
   [DavideDG; 2025]
 
+- Use AVX512VL XOP-like bit rotates for scrypt's Salsa20.  [Solar; 2025]
+
+- When we use AVX512BW, also enable usage of AVX512VL and AVX512DQ.  [Claudio André; 2025]
```
solardiz, line 425: I'd re-add the empty line here (so we have two empty lines before old version's list).

src/configure
```diff
@@ -724,6 +724,7 @@ infodir
 docdir
 oldincludedir
 includedir
+runstatedir
```
solardiz: IIRC, you said `runstatedir` was temporary and you'd drop it when finalizing this PR? Perhaps now is the time?

src/m4/jtr_x86_logic.m4
```diff
@@ -268,15 +268,15 @@ if test "x$simd" != xno; then
 
   AS_IF([test "x$CPU_NOTFOUND" = x0],
   [
-  CFLAGS="$CFLAGS_BACKUP -mavx512bw -P $EXTRA_AS_FLAGS $CPPFLAGS $CFLAGS_EXTRA $CPUID_ASM"
+  CFLAGS="$CFLAGS_BACKUP -mavx512bw -mavx512vl -mavx512dq -P $EXTRA_AS_FLAGS $CPPFLAGS $CFLAGS_EXTRA $CPUID_ASM"
 
-  AC_MSG_CHECKING([for AVX512BW])
+  AC_MSG_CHECKING([for AVX512BW + AVX512VL + AVX512DQ])
```
solardiz, line 273: Strictly speaking, the test program we run only checks BW and VL, and then we assume DQ is implied. So we could want to make it just `for AVX512BW + AVX512VL` here.

src/m4/jtr_x86_logic.m4
```diff
@@ -464,16 +464,17 @@ dnl ======================================================================
 
   AS_IF([test "x$CPU_NOTFOUND" = x0],
   [
-  AC_MSG_CHECKING([for AVX512BW])
+  AC_MSG_CHECKING([for AVX512BW + AVX512VL + AVX512DQ])
```
solardiz, line 467: ... and here.

(I don't get why we have this in two places.)

src/x86-64.S
```diff
@@ -1651,9 +1651,15 @@ DES_bs_crypt_plain_loop:
...
 /* Leaf 7 */
+// 	0x00010000	/* AVX512 foundation */
+// 	0x00020000	/* AVX512 Doubleword and Quadword Instructions */
+// 	0x40000000	/* AVX512BW */
+// 	0x80000000	/* AVX512VL */
 #define C7_AVX2                 $0x00000020 /* AVX2 */
 #define C7_AVX512F              $0x00010000
 #define C7_AVX512BW             $0x40010000 /* AVX512BW + AVX512F */
+#define C7_AVX512VL             $0xC0010000 /* AVX512BW + AVX512VL + AVX512F */
+#define C7_AVX512DQ             $0xC0030000 /* AVX512BW + AVX512DQ + AVX512VL + AVX512F */
```
solardiz, line 1662: (I didn't review the specific bitmasks against the documentation. I just hope they're correct.)

solardiz (approved):
This looks almost good enough to merge, with only trivial cleanups maybe left. Thank you, @claudioandre-br!

src/m4/jtr_x86_logic.m4
solardiz: > there may be future CPUs supporting AVX10/256 where the 128-bit VL intrinsic would compile and run yet this wouldn't imply support for 512-bit BW. Such future CPUs wouldn't set the CPUID bit corresponding to VL

Upon a second thought, actually maybe they would set that CPUID bit. It's no problem, and no reason to change anything in this PR - I am just correcting what I wrote for the sake of it. We may want to add AVX10/256 support later, with a separate PR, and maybe when such CPUs actually appear and can be tested. As a guess, maybe we'll be checking for VL alone as a separate configure test from BW+VL+DQ, and would need to treat it differently in code (in many ways, including CPUID check and non-usage of 512-bit vectors).

src/m4/jtr_x86_logic.m4
ghost: Now it gets confusing.

The `else` part (runs when `--enable-simd=avx512bw`) has a test program that only tests `AVX512BW + AVX512VL`.

The `if` part (runs when `--native-tests=true`) does not use a test program. It uses `CPU_detect()`.

1. In any case, **both use the `-mavx512dq` flag.**
2. `CPU_detect` without setting a value for CPU_REQ_* seems wrong to me. It should be like this:

I added a #define CPU_REQ_AVX512BW		1
```C
#define CPU_REQ_AVX512BW		1
extern int CPU_detect(void); extern char CPU_req_name[];
      unsigned int nt_buffer8x[4], output8x[4];
      int main(int argc, char **argv) { return !CPU_detect(); }
```

Anyway, should I remove DQ string ???? Is a new commit with a fix for cpu_detection required?

solardiz: Oh. I think it's OK to leave this as you have it for this PR, no further change needed. Thank you!

ghost:
So far, everything seems to be fine.

```powershell
Version: 1.9.0-jumbo-1+bleeding-7146e4c827 2025-03-12 05:14:17 +0100
Build: cygwin 64-bit x86_64 AVX512BW AC OMP OPENCL
SIMD: AVX512BW, interleaving: MD4:3 MD5:3 SHA1:1 SHA256:1 SHA512:1
AES hardware acceleration: AES-NI
CPU tests: AVX512(BW+VL+DQ)
CPU fallback binary: john-avx2-omp
OMP fallback binary: john-avx512bw
[...]
Cygwin version: 3.5.7-1.x86_64, 2025-01-29 19:46 UTC
```

```powershell
Will run 2 OpenMP threads
Testing: descrypt, traditional crypt(3) [DES 512/512 AVX512F]... (2xOMP) PASS
Testing: bsdicrypt, BSDI crypt(3) ("_J9..", 725 iterations) [DES 512/512 AVX512F]... (2xOMP) PASS
Testing: md5crypt, crypt(3) $1$ (and variants) [MD5 512/512 AVX512BW 16x3]... (2xOMP) PASS
Testing: md5crypt-long, crypt(3) $1$ (and variants) [MD5 32/64]... (2xOMP) PASS
Testing: bcrypt ("$2a$05", 32 iterations) [Blowfish 32/64 X3]... (2xOMP) PASS
Testing: scrypt (16384, 8, 1) [Salsa20/8 128/128 AVX512VL]... (2xOMP) PASS
[...]
```

src/m4/jtr_x86_logic.m4
magnumripper: > I don't get why we have this in two places.

The first half of `m4/jtr_x86_logic.m4` checks what the build host supports (using `cpuid`), unless cross compiling. The second half of it is (only) for cross compiling [eg. fallbacks], so it just checks what the toolchain can do.

magnumripper:
> It seems it tests CPU support twice (as seen below):
> 
> ```shell
> configure: Trying to force avx512bw using default method (--enable-simd=avx512bw).
> checking if gcc supports -mavx512bw -mavx512vl -mavx512dq w/ linking... yes
> (...)
> configure: Testing tool-chain's CPU support with given options
> (...)
> checking for AVX512BW + AVX512VL + AVX512DQ... yes
> ```

That's because you forced AVX512BW on the command line... I guess the second test is redundant then. That forcing stuff was added later.

magnumripper:
So I'm now seeing this:
```
$ ./john -list=build-info
Version: 1.9.0-jumbo-1-internal+bleeding-ff4c3b5cc3 2025-03-24 16:50:20 +0100
Build: linux-gnu 64-bit x86_64 AVX512BW AC OMP OPENCL
SIMD: AVX512BW, interleaving: MD4:3 MD5:3 SHA1:1 SHA256:1 SHA512:1
AES hardware acceleration: AES-NI
CPU tests: AVX512(BW+VL+DQ)
```
It says "AVX512BW" twice (which is oddly specific, so I was lead to believe it was literally just that), but then AVX512(BW+VL+DQ) in the "CPU tests" line. I was worried I had ended up with a Frankenstein build that didn't actually have VL or DQ instructions but the scrypt format does say VL so I guess all is fine. This output is confusing but I'm not sure how to make it better.

Another problem is that I should apparently cross compile using `--enable-simd=avx512bw` (will translate to BW+VL+DQ by an easter egg in `./configure`, plus a gcc-implied AVX512F) but if a user tried `--enable-simd=avx512vl` she would end up with only that (plus the gcc-implied AVX512F) and no BW or DQ. And similarly for using `--enable-simd=avx512dq`.

This is not a problem for me because I recalled seeing this PR, but how would a user or even a package maintainer know that the only correct way of writing it is `--enable-simd=avx512bw`?

I'm not sure I have any suggestion for this problem either, other than maybe we should parse `--enable-simd=avx512bw` without any magic (so would mean `-mavx512bw`) and instead trigger the easter egg with just `--enable-simd=avx512` (which doesn't map directly to anything used by gcc - there's no `-mavx512` option).

And if that last idea holds, maybe that leads to an answer for the first problem. It could say:
```
$ ./john -list=build-info
Version: 1.9.0-jumbo-1-internal+bleeding-ff4c3b5cc3 2025-03-24 16:50:20 +0100
Build: linux-gnu 64-bit x86_64 AVX512 AC OMP OPENCL
SIMD: AVX512, interleaving: MD4:3 MD5:3 SHA1:1 SHA256:1 SHA512:1
AES hardware acceleration: AES-NI
CPU tests: AVX512(BW+VL+DQ)
```
This output would be much less confusing.

ghost:
> ```
> $ ./john -list=build-info
> Version: 1.9.0-jumbo-1-internal+bleeding-ff4c3b5cc3 2025-03-24 16:50:20 +0100
> Build: linux-gnu 64-bit x86_64 AVX512 AC OMP OPENCL
> ```

Printing only `AVX512` seems like a good option to me. 

The `--enable-simd=avx512` should work now (I remember testing it).

Regarding other issues, the good thing is that people experimenting should know what they are doing or avoid doing it on production systems.

ghost:
On second thought, forcing `--enable-simd=avx512dq` to produce a binary (BW+VL+DQ) is also easy (just add new conditions to the case statement). The problem is: should we do this? Or should we allow people to experiment?

magnumripper:
> The `--enable-simd=avx512` should work now (I remember testing it).

Oh, indeed it does! I must have made a typo when I tried that.

> On second thought, forcing `--enable-simd=avx512dq` to produce a binary (BW+VL+DQ) is also easy (just add new conditions to the case statement). The problem is: should we do this? 

We could, but I think we should instead stop `--enable-simd=avx512bw` from doing so: We should support `--enable-simd=avx512` as a recipe for bw+vl+dq (and possibly more in the future) but stop calling it "default method" because such recipe is not the default method! The default method is that `--enable-simd=foo` gets translated to `-mfoo`, period, end of story. The `--enable-simd=avx512` is a recipe just like eg. `--enable-simd=altivec`, and neither are the default method.

I didn't test this yet, but I think something like this addresses what I mean:
```diff
diff --git a/src/configure.ac b/src/configure.ac
index ba480c409..b27e03d11 100644
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -438,12 +438,11 @@ case "$simd" in
     JTR_FLAG_CHECK_LINK([-mpower8vector], 2)
     SIMD_NAME="Altivec2"
     ;;
-  dnl Handle known cases of --enable-simd=foo --> -mfoo
-  avx512|avx512bw)
-    SIMD_NAME="AVX512BW"
-    AC_MSG_NOTICE([Trying to force $SIMD_NAME using default method (--enable-simd=$simd).])
+  avx512)
     JTR_FLAG_CHECK_LINK([-mavx512bw -mavx512vl -mavx512dq], 2)
+    SIMD_NAME="AVX512"
     ;;
+  dnl Handle known cases of --enable-simd=foo --> -mfoo
   mmx|sse*|ssse3|avx*|xop*)
     SIMD_NAME=`echo $simd | tr a-z A-Z`
     AC_MSG_NOTICE([Trying to force $SIMD_NAME using default method (--enable-simd=$simd).])
```

> Or should we allow people to experiment?

I'm all for allowing people to experiment but we don't want them to struggle.

ghost:
```sh
Target CPU ......................................... x86_64 AVX512, 64-bit LE
```

```sh
Version: 1.9.0-jumbo-1+bleeding-aa93adfa59 2025-03-27 09:33:47 -0300
Build: linux-gnu 64-bit x86_64 AVX512 AC OMP
[...]
AES hardware acceleration: AES-NI
CPU tests: AVX512(BW+VL+DQ)
```

See also https://github.com/openwall/john-packages/pull/778.

PR #5696, opened 2025-03-14
autotools: skip CPU support tests if already run them

ghost:
It is not necessary to run the second (redundant) test.

E.g.:
```
configure: Trying to force avx512bw using default method (--enable-simd=avx512bw).
checking if gcc supports -mavx512bw -mavx512vl -mavx512dq w/ linking... yes
(...)
(...)
checking for AVX512BW + AVX512VL + AVX512DQ... yes
```


magnumripper (approved):
This looks right to me

Issue #5038, opened 2022-01-31
Add Pantagrule


solardiz:
https://www.kitploit.com/2021/08/pantagrule-large-hashcat-rulesets.html
https://github.com/rarecoil/pantagrule

I haven't looked into this yet, and don't expect to have time for that soon, so someone else might.

magnumripper:
Interesting. And I see our friend @roycewilliams had a part in it!

roycewilliams:
Only in passing - just briefly suggested also running against a different corpus (Hashes.org founds). Somehow this translated into getting my name on it. 😬

magnumripper:
Fame is hard to get on purpose - easier to get if talented 😉

solardiz:
Anyway, maybe this means @roycewilliams is the one to contribute this to JtR? ;-)

solardiz:
@roycewilliams Is this something we should add? If so, would you contribute it via a pull request? Thanks!

solardiz:
I guess we'll also need to include these rules from somewhere in `All`. Need review and experiments to determine where exactly in `All` would be best.

PR #5697, opened 2025-03-16
rules: updated OneRuleToRuleThemStill from 1.1 to 1.3 from upstream

exploide:
Turns out `OneRuleToRuleThemStill.rule` got an update some time ago. This PR updates the vendored copy to latest [upstream](https://github.com/stealthsploit/OneRuleToRuleThemStill).

### DIFF
--- a/run/rules/OneRuleToRuleThemStill.rule
+++ b/run/rules/OneRuleToRuleThemStill.rule
@@ -21,49234 +21,48419 @@
 #    Mastodon: @Stealthsploit@infosec.exchange 			 #
 #    Web:      https://in.security                      	 #
 #                                                      		 #
-# ***                   v1.1 10/01/23                        *** #
+# ***                   v1.3 22/12/23                        *** #
 ##################################################################
 :
-,7]
-,8$9
-,6$1
-,7-4
-,7+6
-,8-7
-,6*45
-,7$0
-,6D3
-,6Es]v
-,9+5{
-,6*30
-,6$1$2$3
-,7,8
-,6Z1
-.6$2
-,6$8
-,6u
-,6o1o
-,8i6s
-Z1cD7
-,7+6Y2
-,5o41
-,5,6,7
-.6c
-Y5'8
-,4K
-,7E$3
-,2,1*02
-.6-9-5
-,2,6
-Y4x26
-suo$1
-.8s26
-,3+3
-,3.5
-,5k
-ss7$7
-,7.5
-,2*01
-T7$1
-,5*52
+,7 ]
+,8 $9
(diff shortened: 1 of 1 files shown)


solardiz:
Thank you @exploide! You might also want to consider #5038.

solardiz:
FWIW, here's how this affects the numbers for `All`:

```
0:00:00:00 - 9526968 preprocessed word mangling rules were reduced by dropping 152906 rules
0:00:00:00 - 9374062 preprocessed word mangling rules
```

```
0:00:00:00 - 9526153 preprocessed word mangling rules were reduced by dropping 152286 rules
0:00:00:00 - 9373867 preprocessed word mangling rules
```

Only a slight change, so looks good to me.

The reason I went to check this is to ensure that despite of the introduction of space separators between rule commands in this new revision of `OneRuleToRuleThemStill`, our code is still able to identify and remove duplicates between this ruleset and the `hashcat` ruleset, which is also part of `All`. Most rules in `OneRuleToRuleThemStill` are expected to also be in `hashcat` because of this optimized ruleset's heritage.

I also find this introduction of space separators curious - while we squeeze them out at startup, I thought hashcat didn't do that? Don't they still have performance impact for hashcat?

exploide:
Numbers look reasonable. AFAIK OneRuleToRuleThemStill has been reduced by about 800 duplicate rules.

I'm not familiar with the rule syntax. Didn't test yet whether formatting with spaces could have a performance impact.

solardiz:
> Didn't test yet whether formatting with spaces could have a performance impact.

There's no need to test this with `john` - we know we squeeze them out at startup, so there should be no further runtime performance impact. We also actively use them for clarity outselves. It's just unusual to see them in a ruleset intended for hashcat, because I think there is performance impact for hashcat.

exploide:
> It's just unusual to see them in a ruleset intended for hashcat, because I think there is performance impact for hashcat.

This made me curious and I performed a test but I'm not able to reproduce a performance regression.

Test setup:

- hashcat v6.2.6
- 1x RTX 3070 Ti + 2x RTX 2080 SUPER
- 300 NT hashes
- 15GB crackstation.txt wordlist
- OneRuleToRuleThemStill with commits https://github.com/stealthsploit/OneRuleToRuleThemStill/commit/dc416883f6f03c20af1d534d17e605f9dba5a1da and https://github.com/stealthsploit/OneRuleToRuleThemStill/commit/b40ceac8b78121b2276c3a567fa3a370cfdfdb8d (right before and after the introduction of spaces)

Both times, the cracking attempt took 24 minutes with around 42700 MH/s.

PR #5700, opened 2025-03-18
Count/report hashes removed by --salts or --cost options

### LINKED ISSUE #5699: Missing "Remaining" line when using --cost and/or --salt
The "Remaining..." message is sometimes entirely missing when excluding hashes with `--cost`. I believe some of the problems has been forever but the "Remaining" used to be correct. The behavior changed slightly with 6d7b3f50, for better and worse.

The following examples use all 7z samples.

### Current Bleeding, none cracked and 1 filtered with `--cost`:
```
Loaded 8 password hashes with 8 different salts (7z, 7-Zip archive encryption [SHA256 256/256 AVX2 8x AES])
Loaded hashes with cost 1 (iteration count) varying from 262144 to 524288
Loaded hashes with cost 2 (padding size) varying from 4 to 14
```
The "Remaining" line is missing entirely, so this doesn't report the fact that `--cost` removed one hash at all (you will see it if calling detailed status). We should ideally add a line saying something like "Removed 1 hash due to cost constraint(s)" and (more important to me) a "Remaining" line should be emitted.

### Current Bleeding, 3 already cracked and 1 filtered with `--cost`:
```
Loaded 8 password hashes with 8 different salts (7z, 7-Zip archive encryption [SHA256 256/256 AVX2 8x AES])
Cracked 3 password hashes (are in ../run/john.pot), use "--show"
Remaining 4 password hashes with 4 different salts
Remaining hashes with cost 1 (iteration count) varying from 262144 to 524288
Remaining hashes with cost 2 (padding size) varying from 11 to 14
```
The "Cracked" and "Remaining" lines are correct here, it just doesn't explain to the user what happened to the eighth hash except the "Loaded hashes with cost" lines very subtly changed to "Remaining hashes with cost" (I only noticed now) but that also happens with some cracked and none filtered.

### Before 6d7b3f50, none cracked and 1 filtered with `--cost`:
```
Loaded 8 password hashes with 8 different salts (7z, 7-Zip archive encryption [SHA256 256/256 AVX2 8x AES])
Cracked 1 password hash (is in ../run/john.pot), use "--show"
Remaining 7 password hashes with 7 different salts
Remaining hashes with cost 1 (iteration count) varying from 262144 to 524288
Remaining hashes with cost 2 (padding size) varying from 4 to 14
```
This older version had a correct "Remaining" line for this case, although it listed the filtered hash as cracked instead of filtered.

### Before 6d7b3f50, 3 already cracked and 1 filtered with `--cost`:
```
Loaded 8 password hashes with 8 different salts (7z, 7-Zip archive encryption [SHA256 256/256 AVX2 8x AES])
Cracked 4 password hashes (are in ../run/john.pot), use "--show"
Remaining 4 password hashes with 4 different salts
Remaining hashes with cost 1 (iteration count) varying from 262144 to 524288
Remaining hashes with cost 2 (padding size) varying from 11 to 14
```
Basically the same problem here.

magnumripper:
Closes #5699

### DIFF
--- a/src/john.c
+++ b/src/john.c
@@ -1084,7 +1084,7 @@ static void john_load(void)
 	}
 
 	if (options.flags & FLG_PASSWD) {
-		int total;
+		int total, cracked;
 		int i = 0;
 
 		if (options.flags & FLG_SHOW_CHK) {
@@ -1241,16 +1241,16 @@ static void john_load(void)
  */
 		load_extra_pots(&database, &ldr_load_pot_file);
 
-		total = ldr_fix_database(&database);
+		total = database.password_count;
+		cracked = ldr_fix_database(&database);
 
 		if (database.password_count && options.regen_lost_salts)
 			build_fake_salts_for_regen_lost(&database);
 
-		if (john_main_process && database.password_count < total) {
-			int count = total - database.password_count;
+		if (john_main_process && cracked) {
 			printf("Cracked %d password hash%s%s%s%s, use \"--show\"\n",
-			    count, count != 1 ? "es" : "",
-			    loaded_extra_pots ? "" : (count != 1 ? " (are in " : " (is in "),
+			    cracked, cracked != 1 ? "es" : "",
+			    loaded_extra_pots ? "" : (cracked != 1 ? " (are in " : " (is in "),
 			    loaded_extra_pots ? "" : path_expand(options.activepot),
 			    loaded_extra_pots ? "" : ")");
 		}
--- a/src/loader.c
+++ b/src/loader.c
@@ -2038,6 +2038,7 @@ static void ldr_fill_user_words(struct db_main *db)
 int ldr_fix_database(struct db_main *db)
 {
 	int total = db->password_count;
+	int cracked = 0;
 
 	ldr_init_salts(db);
 	MEM_FREE(db->password_hash);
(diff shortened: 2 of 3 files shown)


magnumripper:
Tested with `--cost`, `--salts`, none or both, with or without some hashes cracked.

magnumripper:
Example output:
```
$ ../run/john -v:3 all_tests.in -form:sha512crypt -mask:a -cost=5000 --salts=-2
Loaded 33 password hashes with 15 different salts (2.2x same-salt boost) (sha512crypt, crypt(3) $6$ [SHA512 256/256 AVX2 4x])
Removed 20 password hashes due to --salts
Removed 1 password hash due to --cost
Cracked 2 password hashes (are in ../run/john.pot), use "--show"
Remaining 10 password hashes with 10 different salts
Cost 1 (iteration count) is 5000 for all remaining hashes
```
I could easily add how many salts each of the filter options removed, but I think it's overkill?

solardiz (changes requested):
I only skimmed. I think you also need to update the comment in `loader.h` on the return value of `ldr_fix_database`.

Other than that, this is good stuff, but it'd take a long while until I get around to reviewing this for real. So maybe just fix the comment and merge, and we'll rely on your own testing for now.

Issue #5126, opened 2022-05-02
`mem_alloc(): Cannot allocate memory` with huge wordlist and high number of forks


AlekseyCherepanov:
I got access to an unused machine with 512Gb of memory and 96 threads. I have non-openmp build. When I use `--fork=96` with rules and a huge wordlist (like HIBP cracks), I get messages like the following:
```
mem_alloc(): Cannot allocate memory trying to allocate 8000000008 bytes
```

Wordlist without rules works well.

Preparations:
```
$ echo '$MD4$aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa' > md4.pw
$ seq 1000000000 > s1g.txt
$ ls -sh s1g.txt
9.3G s1g.txt
$ free -h
              total        used        free      shared  buff/cache   available
Mem:          503Gi       1.1Gi       502Gi       3.0Mi       138Mi       500Gi
Swap:         974Mi        63Mi       911Mi
```

Example of failure. There were 8 `mem_alloc()` lines, 29 `DONE` lines:
```
$ ./john/run/john md4.pw --fork=96 --wordlist=s1g.txt --rules=':'
Using default input encoding: UTF-8
Loaded 1 password hash (Raw-MD4 [MD4 512/512 AVX512BW 16x3])
Node numbers 1-96 of 96 (fork)
Each node loaded the whole wordfile to memory
mem_alloc(): Cannot allocate memory trying to allocate 8000000008 bytes
[...]
mem_alloc(): Cannot allocate memory trying to allocate 8000000008 bytes
Press 'q' or Ctrl-C to abort, 'h' for help, almost any other key for status
Enabling duplicate candidate password suppressor
67: posix_memalign (268435456 bytes): Cannot allocate memory
94 0g 0:00:03:57 DONE (2022-05-02 17:51) 0g/s 43929p/s 43929c/s 43929C/s 999899230..999999934
[...]
Waiting for 95 children to terminate
41 0g 0:00:04:09 DONE (2022-05-02 17:52) 0g/s 41701p/s 41701c/s 41701C/s 999899177..999999977
Main process session completed, but some child processes failed
```

`--dupe-suppression=0` does not help.

`--fork=65` works well. `--fork=66` fails. 66 prints this message unlike 65:
```
Each node loaded the whole wordfile to memory
```

[...]

PR #5704, opened 2025-03-21
mgetl.h: Fix a buffer over-read

magnumripper:
It's unclear to me how this bug never surfaced (until now) since introduced in 2021.

### DIFF
--- a/src/mgetl.h
+++ b/src/mgetl.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020-2021, magnum
+ * Copyright (c) 2020-2025, magnum
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted.
@@ -58,11 +58,24 @@ static MAYBE_INLINE char *mgetl(char *line)
 #endif
 
 	/* One char at a time */
-	while (*map_pos != '\n' && map_pos < map_end && pos < line + LINE_BUFFER_SIZE - 1)
-		*pos++ = *map_pos++;
+	if (map_end - map_pos >= LINE_BUFFER_SIZE - (pos - line)) {
+		/* Need only check write */
+		while (*map_pos != '\n' && pos < line + LINE_BUFFER_SIZE - 1)
+			*pos++ = *map_pos++;
+
+		/* Line may be longer than our buffer, discard extra. */
+		if (pos >= line + LINE_BUFFER_SIZE - 1) {
+			while (map_pos < map_end && *map_pos != '\n')
+				map_pos++;
+		}
+	} else
+		/* Need only check read */
+		while (map_pos < map_end && *map_pos != '\n')
+			*pos++ = *map_pos++;
+
 	map_pos++;
 
-	/* Replace LF with NULL */
+	/* Replace LF (or missing last line LF) with NULL */
 	*pos = 0;
 
 	/* Handle CRLF too */
--- a/src/wordlist.c
+++ b/src/wordlist.c
@@ -615,7 +615,7 @@ void do_wordlist_crack(struct db_main *db, const char *name, int rules)
 #ifdef HAVE_MMAP
 		if (mmap_max && mmap_max >= (file_len >> 20)) {
 			if (john_main_process)
-				log_event("- memory mapping wordlist (%"PRId64" bytes)",
+				log_event("- Memory mapping wordlist (%"PRId64" bytes)",
 				          (int64_t)file_len);
 #if (SIZEOF_SIZE_T < 8)
 /*
@@ -634,7 +634,7 @@ void do_wordlist_crack(struct db_main *db, const char *name, int rules)
 				fprintf(stderr, "wordlist: memory mapping failed (%s) (non-fatal)\n",
 				        strerror(errno));
 #endif
-				log_event("- memory mapping failed (%s) - but we'll do fine without it.",
+				log_event("- Memory mapping failed (%s) - but we'll do fine without it.",
 				          strerror(errno));
 			} else {
 				map_pos = mem_map;
@@ -728,15 +728,15 @@ void do_wordlist_crack(struct db_main *db, const char *name, int rules)
 				if (nWordFileLines != myWordFileLines)
 					fprintf(stderr, "Warning: wordlist changed as"
 					        " we read it\n");
-				log_event("- loaded this node's share of "
-				          "wordfile %s into memory "
+				log_event("- Loaded this node's share of "
+				          "wordlist %s into memory "
 				          "(%"PRIu64" bytes of %"PRId64", max_size="Zu
 				          " avg/node)", name, my_size,
 				          (int64_t)file_len,
 				          options.max_wordfile_memory);
 				if (john_main_process)
 				fprintf(stderr,"Each node loaded 1/%d "
-				        "of wordfile to memory (about "
+				        "of wordlist to memory (about "
 				        "%"PRIu64" %s/node)\n",
 				        options.node_count,
 				        my_size > 1<<23 ?
@@ -746,13 +746,13 @@ void do_wordlist_crack(struct db_main *db, const char *name, int rules)
 			}
 			else {
 				if (john_main_process) {
-					log_event("- loading wordfile %s into memory "
+					log_event("- Loading wordlist %s into memory "
 					          "(%"PRId64" bytes, max_size="Zu")",
 					          name, (int64_t)file_len,
 					          options.max_wordfile_memory);
 					if (options.node_count > 1)
 						fprintf(stderr,"Each node loaded the whole "
-						        "wordfile to memory\n");
+						        "wordlist to memory\n");
 				}
 				word_file_str =
 					mem_alloc_tiny((size_t)file_len +
@@ -788,7 +788,7 @@ void do_wordlist_crack(struct db_main *db, const char *name, int rules)
 			if (aep[-1] != csearch)
 				++nWordFileLines;
 			words = mem_alloc((nWordFileLines + 1) * sizeof(char*));
-			log_event("- wordfile had %"PRId64" lines and required %"PRId64
+			log_event("- Wordlist had %"PRId64" lines and required %"PRId64
 			          " bytes for index.",
 			          (int64_t)nWordFileLines,
 			          (int64_t)(nWordFileLines * sizeof(char*)));
@@ -806,7 +806,7 @@ void do_wordlist_crack(struct db_main *db, const char *name, int rules)
 					hash_log++;
 				hash_size = (1 << hash_log);
 				hash_mask = (hash_size - 1);
-				log_event("- dupe suppression: hash size %u, "
+				log_event("- Dupe suppression: hash size %u, "
 					"temporarily allocating %"PRId64" bytes",
 					hash_size,
 					(hash_size * sizeof(unsigned int)) +
@@ -876,7 +876,7 @@ skip:
 				if (ec == '\n' && *cp == '\r') cp++;
 			} while (cp < aep);
 			if ((int64_t)nWordFileLines - i > 0)
-				log_event("- suppressed %"PRId64" duplicate lines "
+				log_event("- Suppressed %"PRId64" duplicate lines "
 				          "and/or comments from wordlist.",
 				          (int64_t)nWordFileLines - i);
 			MEM_FREE(buffer.hash);
--- a/src/wordlist.c
+++ b/src/wordlist.c
@@ -1,8 +1,8 @@
 /*
  * This file is part of John the Ripper password cracker,
  * Copyright (c) 1996-99,2003,2004,2006,2009,2013,2017 by Solar Designer
- *
- * Heavily modified by JimF, magnum and maybe by others.
+ * Copyright (c) 2009-2025, magnum
+ * Copyright (c) 2009-2018, JimF
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted.
@@ -456,6 +456,49 @@ static MAYBE_INLINE int wbuf_unique(char *line)
 	return 1;
 }
 
+#ifdef HAVE_MMAP
+static int mmap_init(int64_t file_len)
+{
+	if (mem_map)
+		return 1;
+
+	int mmap_max = cfg_get_int(SECTION_OPTIONS, NULL, "WordlistMemoryMapMaxSize");
+
+	if (mmap_max == -1)
+		mmap_max = 1 << 10;
+
+	if (mmap_max && mmap_max >= (file_len >> 20)) {
+		if (john_main_process)
+			log_event("- Memory mapping wordlist (%"PRIu64" bytes)",
+			          (uint64_t)file_len);
+#if (SIZEOF_SIZE_T < 8)
+/*
+ * Now even though we are 64 bit file size, we must still deal with some
+ * 32 bit functions ;)
+ */
+		mem_map = MAP_FAILED;
+		if (file_len < ((1LL)<<32))
+#endif
+			mem_map = mmap(NULL, file_len,
+			               PROT_READ, MAP_SHARED,
+			               fileno(word_file), 0);
+		if (mem_map == MAP_FAILED) {
+			mem_map = NULL;
+			log_event("- Memory mapping failed (%s) - but we'll do fine without it.",
+			          strerror(errno));
+		} else {
+			map_pos = mem_map;
+			map_end = mem_map + file_len;
+#if MGETL_HAS_SIMD
+			map_scan_end = map_end - VSCANSZ;
+#endif
+		}
+	}
+
+	return mem_map != NULL;
+}
+#endif
+
 void do_wordlist_crack(struct db_main *db, const char *name, int rules)
 {
 	union {
@@ -594,14 +637,6 @@ void do_wordlist_crack(struct db_main *db, const char *name, int rules)
 	if (name && !file_is_fifo) {
 		char *cp, csearch;
 		int64_t ourshare = 0;
-#ifdef HAVE_MMAP
-		int mmap_max =
-			cfg_get_int(SECTION_OPTIONS, NULL,
-			            "WordlistMemoryMapMaxSize");
-
-		if (mmap_max == -1)
-			mmap_max = 1 << 10;
-#endif
 		jtr_fseek64(word_file, 0, SEEK_END);
 		if ((file_len = jtr_ftell64(word_file)) == -1)
 			pexit(STR_MACRO(jtr_ftell64));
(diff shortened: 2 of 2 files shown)


solardiz:
I'd call this "over-read" in the commit message.

How did you find this?

solardiz:
I guess this fix has slight performance impact? Because at end of line, the `\n` condition would be detected first, and now is only reached after the length check.

Also, I guess the bug would only be triggered when there's no `\n` after the last line in a file _and_ it happens to be at page boundary. So maybe we can avoid it (if not already) by over-allocating 1 byte (rounding up to extra page allocation when it's on page boundary), without performance impact.

magnumripper:
I've had some weird (and very rare) segfaults in automated scenarios using temporary wordlists. They were/are hard to reproduce because the wordlist would not exist anymore, or not be the same (so just resuming wouldn't crash again, as it "should"). The addresses printed by the segfault always pointed to your dupe suppression code but I couldn't narrow it down to anything sensible. Today "it" happened again but this time I had all artefacts needed to reproduce with a debug build. Turns out it wasn't dupe suppression code this time.

I'm kinda hoping the dupe suppression crashes were also _somehow_ caused by this bug, just surviving slightly longer. Not sure how that would be possible at all though. But It is possible that I gave the wrong `-e` argument to `addr2line` because CPU/OMP fallback ended up with some other binary than I thought was running, so I got mislead into the dupe suppression code. I hope so 🤣 but time will tell.

I don't have the energy to delve into why this bug triggered so very rarely but maybe it's something like the memory mapped wordlist needed to end exactly at a page boundary, or something like that. That would definitely raise the odds.

magnumripper:
> I guess this fix has slight performance impact?

I think very slight. I'll do some testing tomorrow. Maybe that LF test can be made more clever, or we can over-allocate like you said.

solardiz:
BTW, looks like in `wordlist.c` we're only using this code when running with mode than 1 node - why is that limitation? Is this not also useful with 1 node?

magnumripper:
> BTW, looks like in `wordlist.c` we're only using this code when running with mode than 1 node - why is that limitation? Is this not also useful with 1 node?

I got, and got rid of, the segfaults in single node runs, so that can't be correct. Maybe in some cases (like above some size threshold?) it only does mmap for multi node runs. In that case, I guess the rationale could be that copying a multi gigabyte wordlist to memory is very wasty (having forked _before_ wordlist init, we copy the same data to a huge identical buffer per node), compared to just mem-mapping it which reuses the same real memory.

Like we've said many times, wordlist.c should be rewritten from scratch.

solardiz:
> > BTW, looks like in `wordlist.c` we're only using this code when running with mode than 1 node - why is that limitation? Is this not also useful with 1 node?
> 
> I got, and got rid of, the segfaults in single node runs, so that can't be correct.

I ran into this the other day when trying to reproduce this very issue - only to find out that the `mgetl` code isn't even reached in a single node run. Well, at least not with the wordlists I was using, but in `wordlist.c` both uses of `mgetl` are inside `if (ourshare < file_len)`, so I don't see how they would be reached in single-node. Do you?

solardiz:
> in `wordlist.c` both uses of `mgetl` are inside `if (ourshare < file_len)`

This specific check comes from my commit f723538fe20ba36509a1cef2c3c3d93361f81849 referencing #5126, but prior to that commit the check also excluded single-node:

```diff
                        // Load only this node's share of words to memory
-                       if (mem_map && options.node_count > 1 &&
-                           (file_len > options.node_count * (length * 100))) {
+                       if (ourshare < file_len) {
```

solardiz:
> check comes from my commit [f723538](https://github.com/openwall/john/commit/f723538fe20ba36509a1cef2c3c3d93361f81849) referencing #5126, but prior to that commit the check also excluded single-node

The previous version of this check dates back all the way to your introduction of memory-mapped wordlists via 1f827c7aee3207498a098b40d3f6e6b3195b3294 in 2014. Somehow you simply added `mem_map && ` to pre-existing (even older) `options.node_count > 1` check. In our git commits and source code comments, there doesn't appear to be any explanation of why you chose to have this limitation.

solardiz:
Oh, I see what I missed. There was an extra explicit use of `mgetl` in `wordlist.c` in the original commit from 2014. That use still exists, but is now implicit via the `GET_LINE` macro. So maybe we did and still do use memory-mapping for single-node, too. Then I don't know why it didn't appear to be triggered in my tests.

magnumripper:
> I'm kinda hoping the dupe suppression crashes were also _somehow_ caused by this bug, just surviving slightly longer. Not sure how that would be possible at all though.

Indeed that was an unrelated bug: That's now #5714. 

I'll revistit this one soon.

magnumripper:
> Then I don't know why it didn't appear to be triggered in my tests.

It's usually very hard to know: There's far too much hidden engineering and undocumented thresholds in wordlist.c. Out of the plethora of obscure features, what actually kicks in depend on wordlist size (or if it's not even a file but stdin, a FIFO or a stdin "pipe" that we buffer), number of forks or MPI nodes, rules or not, loopback or not, encoding conversions needed or not and probably more. And many of the decisions can be tweaked by configuration alternatives and/or command line options.

Even looking in the source doesn't help much, it's a proper mess. We have two disparate notions of memory mapping and three different notions of dupe suppression. Also, in my great wisdom as of 2012 I decided to piggy-back the loopback mode onto all this, instead of writing a separate file for it. Of all dubious source files in our tree, wordlist.c has got to be the winner, lol.

magnumripper:
> I guess this fix has slight performance impact?

It seems to be ~2% faster with this fix when using rules. Without rules, it appears to be 13% slower!? I can't believe that.

magnumripper:
I think I'll wait with more testing or profiling until #5716 is merged, with the original consecutive dupe suppression dropped.

magnumripper:
> So maybe we can avoid it (if not already) by over-allocating 1 byte (rounding up to extra page allocation when it's on page boundary), without performance impact.

Maybe this is the way to go. Trivial too.
```diff
diff --git a/src/wordlist.c b/src/wordlist.c
index cbd5cd745..cc8e8dc48 100644
--- a/src/wordlist.c
+++ b/src/wordlist.c
@@ -626,7 +626,7 @@ void do_wordlist_crack(struct db_main *db, const char *name, int rules)
                        mem_map = MAP_FAILED;
                        if (file_len < ((1LL)<<32))
 #endif
-                       mem_map = mmap(NULL, file_len,
+                       mem_map = mmap(NULL, file_len + 1,
                                       PROT_READ, MAP_SHARED,
                                       fileno(word_file), 0);
                        if (mem_map == MAP_FAILED) {
```

solardiz:
> It seems to be ~2% faster with this fix when using rules.

That's probably random luck.

> Without rules, it appears to be 13% slower!? I can't believe that.

It can be for real, or almost: it's one extra check per line, and the lines may be of just 7 characters long on average.

solardiz:
> ```diff
> -                       mem_map = mmap(NULL, file_len,
> +                       mem_map = mmap(NULL, file_len + 1,
> ```

This could be the best way to do it, but I worry it could result in `mmap` sometimes failing - maybe precisely when it'd have needed an extra page beyond file size? Or is it guaranteed to work for mappings larger than current file size?

magnumripper:
> > ```diff
> > -                       mem_map = mmap(NULL, file_len,
> > +                       mem_map = mmap(NULL, file_len + 1,
> > ```
> 
> This could be the best way to do it, but I worry it could result in `mmap` sometimes failing - maybe precisely when it'd have needed an extra page beyond file size? Or is it guaranteed to work for mappings larger than current file size?

Macos manpage says "_The mmap() system call causes the pages starting at addr and continuing for at most len bytes to be mapped_". I guess we can't rely on "at most" - even if I would test it and it worked fine here, it may not elsewhere.

POSIX says "_The range of bytes starting at off and continuing for len bytes shall be legitimate for the possible (not necessarily current) offsets in the file_". That does sound promising though. Or maybe not.

magnumripper:
Both say "_Any extension beyond the end of the mapped object will be zero-filled_".

magnumripper:
> Also, I guess the bug would only be triggered when there's no `\n` after the last line in a file _and_ it happens to be at page boundary.

The test case I had, using artefacts from the real-world segfault, really did have a LF for the very last line (I think). Just from staring at the code, I think it doesn't matter at all - the problem is when file size is a multiple of page size, regardless.

magnumripper:
POSIX also says "_References within the address range starting at pa and continuing for len bytes to whole pages following the end of an object shall result in delivery of a SIGBUS signal._". If I understand that correctly, using `+ 1` will not work.

magnumripper:
Here's a variant
```c
	if (map_end - map_pos > LINE_BUFFER_SIZE)
		while (*map_pos != '\n' && pos < line + LINE_BUFFER_SIZE - 1)
			*pos++ = *map_pos++;
	else
		while (map_pos < map_end && *map_pos != '\n')
			*pos++ = *map_pos++;
```

magnumripper:
> > Also, I guess the bug would only be triggered when there's no `\n` after the last line in a file _and_ it happens to be at page boundary.
> 
> The test case I had, using artefacts from the real-world segfault, really did have a LF for the very last line. Just from staring at the code, I think it doesn't matter at all - the problem is when file size is a multiple of page size, regardless.

Yet I can't reproduce that (I don't have access to the artefacts right now). I can produce a file not ending with a LF that segfaults with the original code. But with the LF, be it at the very end of a page or a byte before/after it, there's no segfault. I can't understand that from dry-running the code.

magnumripper:
So a file with a size of exactly eg. 8388608 bytes and no LF in the ending line triggers the segfault, as expected. But if adding rules, it doesn't segfault. For the life of me I can't see why?

magnumripper:
> POSIX also says "_References within the address range starting at pa and continuing for len bytes to whole pages following the end of an object shall result in delivery of a SIGBUS signal._". If I understand that correctly, using `+ 1` will not work.

Confirmed: Macos gave me a `Bus error: 10` when adding to the size, so this is not a way to go.

magnumripper:
> Without rules, it appears to be 13% slower!? I can't believe that.

That result can mostly be ignored. I used a file misnamed "bycount.txt" which was actually _with_ count, so much longer lines.

> Here's a variant
> 
> ```c
> 	if (map_end - map_pos > LINE_BUFFER_SIZE)
> 		while (*map_pos != '\n' && pos < line + LINE_BUFFER_SIZE - 1)
> 			*pos++ = *map_pos++;
> 	else
> 		while (map_pos < map_end && *map_pos != '\n')
> 			*pos++ = *map_pos++;
> ```
Apparently this is a winner. Comparing the best result of five to the worst result of five (_to this version's disfavor_), this one is 0.8% faster than the original fix and 6.1% slower than the unsafe code.

Comparing to this version's favor instead, it's 20% faster than the original fix and exactly as fast as the unsafe code.

magnumripper:
BTW here's a thought: If we did run the SIMD code prior to this, we know that the remainder can't blow `LINE_BUFFER_SIZE` so we don't have to bother tracking that.

magnumripper:
> BTW here's a thought: If we did run the SIMD code prior to this, we know that the remainder can't blow `LINE_BUFFER_SIZE` so we don't have to bother tracking that.

No, that was a brainfart. It would only apply for the last word in the file, and end up much slower up to that.

magnumripper:
Other observations re-made while looking into this: The mmap is a fine performance boost, like 50%. Without the SIMD scanning it's more like 40% (the SIMD scan loves longer lines, so better results from eg. loopback mode).

magnumripper:
> So a file with a size of exactly eg. 8388608 bytes and no LF in the ending line triggers the segfault, as expected. But if adding rules, it doesn't segfault. For the life of me I can't see why?

This is a bug: In this case, we do memory map the file but then never use it. Instead, we use the pre-mmap way of buffering and addng an array of pointers so we can address each word without scanning. Separate issue, #5721 .

magnumripper:
> This is a bug: In this case, we do memory map the file but then never use it. Instead, we use the pre-mmap way of buffering and addng an array of pointers so we can address each word without scanning. Separate issue.

I added a fix for that (just don't mmap when we're not going to use it) to this PR, as they're related. Also threw in some consistency/cosmetic changes to log messages.

solardiz:
> a file with a size of exactly eg. 8388608 bytes and no LF in the ending line triggers the segfault

Thanks, I'm finally able to reproduce this with a file produced by `tail -n +99 password.lst | dd bs=1k count=8k of=w8`.

magnumripper:
I searched for the artefacts from the original segfault today but unfortunately they were deleted. I must be mistaken about it not missing an ending LF - I probably confused that case with the suppressor segfault case.

This is ready to merge!

src/mgetl.h
```diff
@@ -58,8 +58,12 @@ static MAYBE_INLINE char *mgetl(char *line)
 #endif
 
 	/* One char at a time */
-	while (*map_pos != '\n' && map_pos < map_end && pos < line + LINE_BUFFER_SIZE - 1)
-		*pos++ = *map_pos++;
+	if (map_end - map_pos > LINE_BUFFER_SIZE)
```
solardiz: Shouldn't this use `>=`? So that the loop in `else` would implicitly satisfy the `- 1` in its omitted `pos < line + LINE_BUFFER_SIZE - 1`.

src/wordlist.c
```diff
@@ -456,6 +456,44 @@ static MAYBE_INLINE int wbuf_unique(char *line)
 	return 1;
 }
 
+#ifdef HAVE_MMAP
+static int mmap_init(int64_t file_len, int mmap_max)
+{
+	if (mem_map)
+		return 1;
+
+	if (mmap_max && mmap_max >= (file_len >> 20)) {
```
solardiz: Looks like we allow almost an extra MiB on top of the configured limit here. Not a new bug, but we could want to rewrite this check (if you do, then do in a separate commit, please).

solardiz (commented):
I can't really review the third commit here - as you say, `wordlist.c` is too complicated.

src/mgetl.h
magnumripper:
```c
	if (map_end - map_pos > LINE_BUFFER_SIZE)
		while (*map_pos != '\n' && pos < line + LINE_BUFFER_SIZE - 1)
			*pos++ = *map_pos++;
	else
		while (map_pos < map_end && *map_pos != '\n')
			*pos++ = *map_pos++;
	map_pos++;

	/* Replace LF with NULL */
	*pos = 0;
```
> Shouldn't this use `>=`? So that the loop in `else` would implicitly satisfy the `- 1` in its omitted `pos < line + LINE_BUFFER_SIZE - 1`.

I was being defensive. But I see now the `else` part is probably always dangerous for overlong lines, isn't it? I need to add a bunch of test files that also contain lines being 126-129 bytes with or without LF. This as last line as well as some non-last lines.

I see now pp.c has similar code.
```c
/* Sort-of fgets() but for a memory-mapped file. Updates len, returns pointer to string */
static MAYBE_INLINE char *mgets(int *len)
{
  char *pos = map_pos;
  char *end = MIN(map_end, pos + BUFSIZ);

  if (map_pos >= map_end)
    return NULL;

  while (map_pos < end && *map_pos != '\n' && *map_pos != '\r')
    map_pos++;

  *len = map_pos - pos;

  while (map_pos < end && (*map_pos == '\n' || *map_pos == '\r'))
    map_pos++;

  return pos;
}
```
This MIN() might be more clever but this code has similar bugs plus it's own ones. I need to test and fix this as well (yes, separate commit, maybe separate PR too). BTW in pp.c the performance is less critical, the mmap is just for decreasing memory usage when multiple sessions use the same wordlist.

Back to the drawing board :)

src/wordlist.c
magnumripper: I assume you mean `if (((uint64_t)mmap_max << 20) >= file_len)`. I was instinctively avoiding overflow but you are right. The cast fixes that.

src/mgetl.h
magnumripper: > This MIN() might be more clever

No, that wouldn't apply to or improve mgetl.h due to us wanting the LF check first.

magnumripper: > But I see now the `else` part is probably always dangerous for overlong lines, isn't it? I need to add a bunch of test files that also contain lines being 126-129 bytes with or without LF. This as last line as well as some non-last lines.

No that shouldn't be a problem (but I'll add tests anyway). This is tricky.

But wait a minute, why are we even using LINE_BUFFER_SIZE here as opposed to PLAINTEXT_BUFFER_SIZE? This comes from john proper. I'd need to also test lines of 1022-1025 bytes I guess, unless we change this.

magnumripper:
OK the code as-is (since yesterday 20:37Z, should be the one you reviewed) passes long line tests. Any input lines of 1024 bytes or longer end up as 1024 bytes/1023 characters. No changes needed.

Except `fgetl()` also discards extra characters and `mgetl()` does not.
```c
		while (c != EOF && c != '\n')
			/* Line was longer than our buffer. discard extra */
			c = getc(stream);
```
Does it matter? Fixing that should (I hope) not result in a significant performance penalty.

magnumripper:
> OK the code as-is (since yesterday 20:37Z, should be the one you reviewed) passes long line tests.

Amended it now just for the below plus minor comments in the code:

> Shouldn't this use `>=`? So that the loop in `else` would implicitly satisfy the `- 1` in its omitted `pos < line + LINE_BUFFER_SIZE - 1`.

```diff
diff --git a/src/mgetl.h b/src/mgetl.h
index c2e394e83..f36b625a1 100644
--- a/src/mgetl.h
+++ b/src/mgetl.h
@@ -58,15 +58,17 @@ static MAYBE_INLINE char *mgetl(char *line)
 #endif
 
        /* One char at a time */
-       if (map_end - map_pos > LINE_BUFFER_SIZE)
+       if (map_end - map_pos >= LINE_BUFFER_SIZE)
+               // Performance critial
                while (*map_pos != '\n' && pos < line + LINE_BUFFER_SIZE - 1)
                        *pos++ = *map_pos++;
        else
+               // Less performance critical
                while (map_pos < map_end && *map_pos != '\n')
                        *pos++ = *map_pos++;
        map_pos++;
 
-       /* Replace LF with NULL */
+       /* Replace LF (or missing last line LF) with NULL */
        *pos = 0;
 
        /* Handle CRLF too */
```

src/mgetl.h
magnumripper: I'll be fixing pp.c in a separate PR

magnumripper:
> Except `fgetl()` also discards extra characters and `mgetl()` does not.

Or so I thought. Somehow the current code (here) actually does discard the rest of lines exceeding 1024 characters (I have yet to understand how/why?) instead of continuing with them at next read - but for the case of a 1024 character word with a LF, it ends up as-is plus an extraneous empty line with just the LF.

magnumripper:
Oh, right. We still need this:
```diff
	/* One char at a time */
	if (map_end - map_pos >= LINE_BUFFER_SIZE)
		// Performance critial
		while (*map_pos != '\n' && pos < line + LINE_BUFFER_SIZE - 1)
                        *pos++ = *map_pos++;
        else
                // Less performance critical
-               while (map_pos < map_end && *map_pos != '\n')
+               while (map_pos < map_end && *map_pos != '\n' && pos < line + LINE_BUFFER_SIZE - 1)
                        *pos++ = *map_pos++;
        map_pos++;
```
But that part has less of a performance requirement.

How could that elude my tests with debugging prints!? I need to retry them.

magnumripper:
OK with that committed (not pushed), we're back with the problem that the next read will get the remainder of overlong lines.

magnumripper:
> Oh, right. We still need this:
> 
> ```diff
> 	/* One char at a time */
> 	if (map_end - map_pos >= LINE_BUFFER_SIZE)
> 		// Performance critial
> 		while (*map_pos != '\n' && pos < line + LINE_BUFFER_SIZE - 1)
>                         *pos++ = *map_pos++;
>         else
>                 // Less performance critical
> -               while (map_pos < map_end && *map_pos != '\n')
> +               while (map_pos < map_end && *map_pos != '\n' && pos < line + LINE_BUFFER_SIZE - 1)
>                         *pos++ = *map_pos++;
>         map_pos++;
> ```

This really doesn't make sense. The `if (...` should mean we don't need that check there. But it did make a difference. I probably need a break.

magnumripper:
Now I get it. The SIMD code quite obviously reads most of the line, only at most 31 characters are left to the code I'm struggling with - so I can't just calculate from `line`! I need to look at `pos` after SIMD and calculate from that. I'm a dork.

magnumripper:
Now we're talking. This version passes all tests I can think of:
```c
	/* One char at a time */
	if (map_end - map_pos >= LINE_BUFFER_SIZE - (pos - line)) {
		/* Need only check write */
		while (*map_pos != '\n' && pos < line + LINE_BUFFER_SIZE - 1)
			*pos++ = *map_pos++;

		/* Line may be longer than our buffer, discard extra. */
		if (pos >= line + LINE_BUFFER_SIZE - 1) {
			while (map_pos < map_end && *map_pos != '\n')
				map_pos++;
		}
	} else
		/* Need only check read */
		while (map_pos < map_end && *map_pos != '\n')
			*pos++ = *map_pos++;
```

magnumripper:
> This version passes all tests I can think of

Performance is also fine, and output using fgetl or mgetl is identical (with very long lines input, etc).
And I double checked we're still immune to page boundary problems. Having lost that in the end would be comical.

src/mgetl.h
magnumripper: > But wait a minute, why are we even using LINE_BUFFER_SIZE here as opposed to PLAINTEXT_BUFFER_SIZE? This comes from john proper. I'd need to also test lines of 1022-1025 bytes I guess, unless we change this.

Never mind this. The longer one is needed for loopback mode, for one.

solardiz (approved):

Issue #5702, opened 2025-03-21
LM hash format: Add test vectors with 8-bit chars


solardiz:
Our current test vectors in `LM_fmt.c` don't test `FINALIZE_NEXT_KEY_BIT_7` (except that it advances the key bit pointer), so any bugs/miscompile in there may go unnoticed. We should add test vectors to ensure proper code coverage by tests.

exaghost:
Hi!
Try this from an old Windows XP system. `C6A3396B0AEF1EF9AAD3B435B51404EE` 
 cp852 or ?b?b?b?b.

magnumripper:
Thank you @exaghost, I'll submit a PR with that one.

solardiz:
Thank you @exaghost and @magnumripper!

Unfortunately, there are two issues:

1. This broke `--test-full=0`:
```
Testing: LM [DES 512/512 AVX512F]... FAILED (LM doesn't set FMT_CASE but at least one test-vector is case-sensitive)
```

2. It appears that we need many more test vectors for LM hashes to reach full "finalize key" code coverage. With this added test vector, the below hack of `FINALIZE_NEXT_KEY_BIT_7` is now detected (self-test fails):

```diff
+++ b/src/DES_bs_b.c
@@ -549,7 +549,7 @@ typedef unsigned ARCH_WORD kvtype;
        kvtype m = mask80, va, vb, tmp; \
        kvand_shr(va, v0, m, 7); \
        kvand_shr(vb, v1, m, 6); \
-       kvand_shr_or(va, v2, m, 5); \
+       kvand_shr_or(va, vzero, m, 5); \
        kvand_shr_or(vb, v3, m, 4); \
        kvand_shr_or(va, v4, m, 3); \
        kvand_shr_or(vb, v5, m, 2); \
```

However, similar hacks of `FINALIZE_NEXT_KEY_BIT_7` for any other `v` (0, 1, or 3 to 7) go undetected (self-test passes). I now suspect we might not have 100% coverage for 7-bit chars, either.

solardiz:
> 1. This broke `--test-full=0`:

Oops, I'm wrong about this one. This is actually a result of my hack to `FINALIZE_NEXT_KEY_BIT_7`, for `v0` in this case, which went undetected by `--test`, but was detected by `--test-full=0` in this weird way.

solardiz:
[...]

PR #5705, opened 2025-03-22
Bitslice DES key setup optimization for AVX-512, related self-test improvements, suppressor auto-disable change

solardiz:
See #5702 and #5707. I only tested on AVX-512. I did not test OpenCL.


solardiz:
> I did not test OpenCL.

I did now. `lm-opencl` passes self-test (with the added test vectors here) on all 6 devices in `super`. `descrypt-opencl` passes on 5/6 (fails on MIC, like it already did prior to this PR).

Issue #5709, opened 2025-03-23
Workaround many OpenCL format self-test failures on Vega 64


solardiz:
```
Device 1: gfx900 [Radeon RX Vega]
Testing: ZIP-opencl, WinZip [PBKDF2-SHA1 OpenCL]... FAILED (cmp_one(1))
```

Adding some `printf` calls, I tracked this down to the very last `hmac_sha1` in `zip_final`. Inputs to this `hmac_sha1` look the same as on NVIDIA, but the output value is different. This happens on the very first pass through this code, where `comp_len` is 0. On the second pass and on (different test vectors?), `comp_len` would be non-zero, but those aren't reached on the failing run.

Looks like this might not be too hard to repair/workaround, so creating issue.

solardiz:
Not the issue here, but while we're at it - `salt->comp_len` is `uint64_t`, but then we assign it to `uint` in the kernel, and I don't see anything in host code to ensure it'd fit in 32-bit - maybe we should add such checks.

solardiz:
`encfs-opencl` and `pfx-opencl` also failing on this device may potentially also be caused by the same underlying miscompile of `hmac_sha1`.

solardiz:
This repairs `zip-opencl` and `encfs-opencl`, but not `pfx-opencl`:

```diff
+++ b/run/opencl/opencl_hmac_sha1.h
@@ -94,6 +94,10 @@ INLINE void hmac_sha1(HMAC_KEY_TYPE void *_key, uint key_len,
        SHA1_Final(local_digest, &ctx);
        for (i = 0; i < 16; i++)
                u.pW[i] ^= (0x36363636 ^ 0x5c5c5c5c);
+#if gpu_amd(DEVICE_INFO)
+       SHA_CTX ctx2;
+#define ctx ctx2
+#endif
        SHA1_Init(&ctx);
        SHA1_Update(&ctx, u.buf, 64);
        SHA1_Update(&ctx, local_digest, 20);
```

magnumripper:
> Not the issue here, but while we're at it - `salt->comp_len` is `uint64_t`, but then we assign it to `uint` in the kernel, and I don't see anything in host code to ensure it'd fit in 32-bit - maybe we should add such checks.

Or trivially fix the kernal? It looks like a bug to me.

magnumripper:
[...]

PR #5710, opened 2025-03-24
Workaround many OpenCL format self-test failures, and general OpenCL fixes

solardiz:
See #5709.

I also included a change to `pbkdf2_ripemd160_kernel.cl` that doesn't make this kernel/format work on this device (it still works on 5 others in "super"), but I think is the right correctness fix anyway.

### DIFF
--- a/run/opencl/ansible_kernel.cl
+++ b/run/opencl/ansible_kernel.cl
@@ -35,5 +35,10 @@ __kernel void ansible_final(__global crack_t *out,
 
 	pbkdf2_sha256_final(out, &salt->pbkdf2, state);
 
+#if gpu_amd(DEVICE_INFO)
+/* Workaround miscompile with AMD-APP 2766.4 */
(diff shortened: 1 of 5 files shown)


magnumripper (approved):

magnumripper:
FWIW it didn't seem make any difference on macos with their crappy runtime... until it crashes midway (did before too) and doesn't test the rest of the formats 🙄 

I ran `./john -test=0 -form:@PBKDF2-SHA,@RIPE,+opencl`

solardiz:
> FWIW it didn't seem make any difference on macos with their crappy runtime...

I'm concerned that there could be performance regressions for kernels that worked fine as-is on certain AMD drivers/GPUs previously, but are now impacted by these workarounds. So maybe you'd run some benchmarks for before/after?

magnumripper:
I'll see what I can do but I don't trust the macos drivers for, well, anything.

PR #5713, opened 2025-03-24
Various MongoDB auditing related improvements (2025-1)

kholia:
This is currently a work-in-progress. It works fine but isn't optimized.

### DIFF
--- a/doc/NEWS
+++ b/doc/NEWS
@@ -424,6 +424,8 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
 
 - When we use AVX512BW, also enable usage of AVX512VL and AVX512DQ.  [Claudio André; 2025]
 
+- Add support for auditing SCRAM PBKDF2-SHA256 hashes.  [Dhiru; 2025]
+
 
 Major changes from 1.8.0-jumbo-1 (December 2014) to 1.9.0-jumbo-1 (May 2019):
 
--- a/src/scram_sha256_fmt_plug.c
+++ b/src/scram_sha256_fmt_plug.c
@@ -1,14 +1,17 @@
 /*
- * This software is Copyright (c) 2016, Dhiru Kholia <dhiru.kholia at gmail.com>,
+ * This software is Copyright (c) 2025, Dhiru Kholia <dhiru.kholia at gmail.com>,
  * and it is hereby released to the general public under the following terms:
  * Redistribution and use in source and binary forms, with or without modification,
  * are permitted.
+ *
+ * https://github.com/mongodb/specifications/blob/master/source/auth/auth.md
+ * https://datatracker.ietf.org/doc/html/rfc7677
  */
 
 #if FMT_EXTERNS_H
-extern struct fmt_main fmt_mongodb_scram;
+extern struct fmt_main fmt_mongodb_scram_sha256;
 #elif FMT_REGISTERS_H
-john_register_one(&fmt_mongodb_scram);
+john_register_one(&fmt_mongodb_scram_sha256);
 #else
 
 #include <string.h>
@@ -27,25 +30,24 @@ john_register_one(&fmt_mongodb_scram);
 #include "base64_convert.h"
 #include "hmac_sha.h"
 #include "simd-intrinsics.h"
-#include "pbkdf2_hmac_sha1.h"
-#include "md5.h"
+#include "pbkdf2_hmac_sha256.h"
 
 #if defined SIMD_COEF_32
-#define SIMD_KEYS		(SIMD_COEF_32 * SIMD_PARA_SHA1)
+#define SIMD_KEYS		(SIMD_COEF_32 * SIMD_PARA_SHA256)
 #endif
 
-#define FORMAT_LABEL            "scram"
+#define FORMAT_LABEL            "scram-pbkdf2-sha256"
 #define FORMAT_NAME             ""
-#define ALGORITHM_NAME          "SCRAM PBKDF2-SHA1 " SHA1_ALGORITHM_NAME
+#define ALGORITHM_NAME          "SCRAM PBKDF2-SHA256 " SHA256_ALGORITHM_NAME
 #define PLAINTEXT_LENGTH        125
-#define HASH_LENGTH             28
+#define HASH_LENGTH             44
 #define SALT_SIZE               sizeof(struct custom_salt)
 #define SALT_ALIGN              sizeof(uint32_t)
-#define BINARY_SIZE             20
+#define BINARY_SIZE             32
 #define BINARY_ALIGN            sizeof(uint32_t)
 #define BENCHMARK_COMMENT       ""
 #define BENCHMARK_LENGTH        0x107
-#define FORMAT_TAG              "$scram$"
+#define FORMAT_TAG              "$scram-pbkdf2-sha256$"
 #define FORMAT_TAG_LENGTH       (sizeof(FORMAT_TAG) - 1)
 #define MAX_USERNAME_LENGTH     128
 
@@ -62,8 +64,9 @@ john_register_one(&fmt_mongodb_scram);
 #endif
 
 static struct fmt_tests tests[] = {
-	{"$scram$someadmin$10000$wf42AF7JaU1NSeBaSmkKzw==$H6A5RF0qz6DrcWNNX4xe+wIeVEw=", "secret"},
-	{"$scram$admin$10000$ouQdw5om9Uc5gxulO9F/8w==$DSnATYsgoE8InL5Petfjp8MWGh4=", "test@12345"},
+	/* MongoDB 8.0.6 hashes */
+	{"$scram-pbkdf2-sha256$accountAdmin16$15000$OxAPvANwV/ZQXqgiW6s6o2+wPM+gfZNthjpUjw==$MenSdE9VmSij4sIKMKfRs+bHy9vkareAopWM8MB+364=", "openwall"},
+	{"$scram-pbkdf2-sha256$user$15000$0sWhCP4Z0gjI7KY6WJ7z/Hs3SEcxC+PUSkv4og==$qnttlssP81IuwtcoZ4TV/M0SMCajePUhPP3mRrnv0aA=", "openwall@1234567890"},
 	{NULL}
 };
 
@@ -71,7 +74,7 @@ static struct custom_salt {
 	int saltlen;
 	int iterations;
 	char username[MAX_USERNAME_LENGTH + 1];
-	unsigned char salt[18 + 1]; /* base64 decoding, 24 / 4 * 3 = 18 */
+	unsigned char salt[28 + 4 + 1]; // 4 bytes for 'startKey'
 } *cur_salt;
 
 static char (*saved_key)[PLAINTEXT_LENGTH + 1];
@@ -96,6 +99,7 @@ static void done(void)
 static int valid(char *ciphertext, struct fmt_main *self)
 {
 	char *ctcopy, *keeptr, *p;
+
 	if (strncmp(ciphertext, FORMAT_TAG, FORMAT_TAG_LENGTH) != 0)
 		return 0;
 	ctcopy = xstrdup(ciphertext);
@@ -111,7 +115,7 @@ static int valid(char *ciphertext, struct fmt_main *self)
 		goto err;
 	if ((p = strtokm(NULL, "$")) == NULL)	/* salt */
 		goto err;
-	if (strlen(p)-2 != base64_valid_length(p, e_b64_mime, flg_Base64_MIME_TRAIL_EQ, 0) || strlen(p) > 24)
+	if (strlen(p)-2 != base64_valid_length(p, e_b64_mime, flg_Base64_MIME_TRAIL_EQ, 0) || strlen(p) > 40)
 		goto err;
 	if ((p = strtokm(NULL, "")) == NULL)	/* hash */
 		goto err;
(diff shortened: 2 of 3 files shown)


magnumripper:
Kholia! 🎉  ❤️

src/mongodb_scram_sha256_fmt_plug.c
```diff
@@ -0,0 +1,303 @@
...
+	const int count = *pcount;
+
+#ifdef _OPENMP
+#pragma omp parallel for
+#endif
+	for (index = 0; index < count; index += MIN_KEYS_PER_CRYPT) {
+#if !defined (SIMD_COEF_32)
+		unsigned char intermediate[BINARY_SIZE];
+		unsigned char output[BINARY_SIZE];
+		SHA256_CTX ctx;
```
ghost: This variable is not used (compilers will complain).

ghost:
In fact, I'm testing it after applying:

```diff
diff --git a/src/mongodb_scram_sha256_fmt_plug.c b/src/mongodb_scram_sha256_fmt_plug.c
index 8f1c57d3d..eaa53011e 100644
--- a/src/mongodb_scram_sha256_fmt_plug.c
+++ b/src/mongodb_scram_sha256_fmt_plug.c
@@ -32,7 +32,6 @@ john_register_one(&fmt_mongodb_scram_sha256);
 
 #undef SIMD_COEF_32 // FIXME
 #include "simd-intrinsics.h"
-#include "pbkdf2_hmac_sha256.h"
 
 #if defined SIMD_COEF_32
 #define SIMD_KEYS              (SIMD_COEF_32 * SIMD_PARA_SHA256)
@@ -191,7 +190,6 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 #if !defined (SIMD_COEF_32)
                unsigned char intermediate[BINARY_SIZE];
                unsigned char output[BINARY_SIZE];
-               SHA256_CTX ctx;
 
                // U1 = HMAC(input, salt + 0001)
                hmac_sha256((unsigned char *)saved_key[index], strlen(saved_key[index]), cur_salt->salt, 28 + 4, output, BINARY_SIZE);

```

src/mongodb_scram_sha256_fmt_plug.c
```diff
@@ -0,0 +1,303 @@
...
+#include "simd-intrinsics.h"
+#include "pbkdf2_hmac_sha256.h"
+
+#if defined SIMD_COEF_32
+#define SIMD_KEYS		(SIMD_COEF_32 * SIMD_PARA_SHA256)
+#endif
+
+#define FORMAT_LABEL            "scram-sha256"
+#define FORMAT_NAME             ""
+#define ALGORITHM_NAME          "MongoDB ServerKey SCRAM PBKDF2-SHA256 " SHA256_ALGORITHM_NAME
```
solardiz: Let's drop `MongoDB ServerKey ` part from this string. This looks like a generic SCRAM format, so probably usable not only for MongoDB. The SHA-1 kind of this format doesn't have that in its algorithm name.

A `Server Key` string is part of the SCRAM spec, no need to make it into the algorithm name string.

Oh, I see that the other format uses `Client Key`. Both are part of the spec, but to be used by client vs. server. Why is the difference here? Maybe we want to make both formats more generic by including this client vs. server bit in the hash strings?

Separately, we may want to rename these format files removing the `mongodb_` part. Or we could make them into one shared format with SHA-1 vs. SHA-256 becoming a tunable cost.

solardiz (approved):
Happy to see you back, @kholia!

We can accept this as first draft (CI bots are happy) and then revise it with subsequent commits.

Please add a `NEWS` entry. Thank you!

src/mongodb_scram_sha256_fmt_plug.c
kholia: Attacking `Server Key` instead of `Client Key -> Stored Key` saves on one (final) SHA1 / SHA256 operation.

I only realized this last night.

This can be seen in the code link below:

https://github.com/mongodb/mongo/blob/0a68308f0d39a928ed551f285ba72ca560c38576/src/mongo/crypto/mechanism_scram.h#L214

src/mongodb_scram_sha256_fmt_plug.c
kholia: Ack - thank you! I have cleaned up the code now.

kholia:
I was able to simplify the code (use existing crypto primitives) and also add SIMD support.

Thanks to @philsmd for the superb work @ https://github.com/hashcat/hashcat/issues/2551#issuecomment-699510706.

doc/NEWS
```diff
@@ -424,6 +424,7 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
 
 - When we use AVX512BW, also enable usage of AVX512VL and AVX512DQ.  [Claudio André; 2025]
 
+- Add support for auditing SCRAM PBKDF2-SHA256 hashes.  [kholia; 2025]
```
solardiz: You're `Dhiru` in all other applicable `NEWS` entries, so maybe make it consistent?  Also, please retain the double empty line before previous version's list of changes.

src/mongodb_scram_sha256_fmt_plug.c
```diff
@@ -0,0 +1,305 @@
...
+#define ALGORITHM_NAME          "SCRAM PBKDF2-SHA256 " SHA256_ALGORITHM_NAME
+#define PLAINTEXT_LENGTH        125
+#define HASH_LENGTH             44
+#define SALT_SIZE               sizeof(struct custom_salt)
+#define SALT_ALIGN              sizeof(uint32_t)
+#define BINARY_SIZE             32
+#define BINARY_ALIGN            sizeof(uint32_t)
+#define BENCHMARK_COMMENT       ""
+#define BENCHMARK_LENGTH        0x107
+#define FORMAT_TAG              "$scram-sha256$"
```
solardiz: Should we possibly use `scram-pbkdf2-sha256` in the format label and tag? And ditto in the other format (with SHA-1), but maybe only if we make it consistent also in other ways.

solardiz (dismissed):
Great to have SIMD support, but CI is failing now - please look into that and fix before we can merge.

src/mongodb_scram_sha256_fmt_plug.c
solardiz: OK, per that link MongoDB stores both `Stored Key` (for authenticating clients) and `Server Key` (to let clients authenticate server), and you choose to attack the latter because it's slightly quicker. Yet some other service may use SCRAM for client authentication only and wouldn't have/store `Server Key`. So we could support both modes maybe?

Either way (supporting both or not), we could do it in the same way for both formats (SHA-1 and SHA-256), right?

solardiz: A separate later project could be to add support for attacking SCRAM challenge/response messages - also either/both kinds (`ClientProof` or `ServerSignature`). We may want to keep this future addition in mind when we name things, but I don't know how this affects the naming. One way could be to include the names `ClientKey`, `ServerKey`, `ClientProof`, or `ServerSignature` as a field in the hash strings, or we can use numeric values for them (0 to 3). For challenge/responses, we'd also need an extra field `Auth`. https://en.wikipedia.org/wiki/Salted_Challenge_Response_Authentication_Mechanism#Proofs

solardiz: Oh, I just realized that the older format is MongoDB-specific - it includes preprocessing with MD5 and a ":mongo:" string. So we can't just make them consistent. Since it's unlikely that new uses of SCRAM with SHA-1 will appear now, maybe we can keep the older format MongoDB-specific, but maybe we should rename its label and format tag to reflect that then. You could also want to switch it to using `ServerKey` if that's available in those older versions of MongoDB too?

kholia: > A separate later project could be to add support for attacking SCRAM challenge/response messages - also either/both kinds (ClientProof or ServerSignature).

This is on my mind. I have the necessary pcaps - just need to figure out the underlying crypto a bit more.

solardiz: We also need to see what hashcat supports, and use compatible hash strings where possible. I haven't checked yet. The above was just me thinking out loud - we really need to see how anything we choose fits in before we decide on it.

solardiz:
Please also add new samples to https://github.com/openwall/john-samples/tree/main/MongoDB

I see you keep adding commits instead of force-pushing - should we plan on using GitHub's squash-and-merge on this one PR as an exception (we normally only do rebase merges, without squashing)? Or are you planning to squash the commits and force-push before we merge?

kholia:
> Or are you planning to squash the commits and force-push before we merge?

Yep - I will squash the commits manually and force-push once everything works.

kholia:
The following scripts were helpful in my debugging work:

```JS
% cat create_user.js
db.createUser({
    user: "user7",
    pwd: passwordPrompt(), // Or  "<cleartext password>"
    customData: {
        employeeId: 12345
    },
    roles: [{
            role: "clusterAdmin",
            db: "admin"
        },
        {
            role: "readAnyDatabase",
            db: "admin"
        },
        "readWrite"
    ],
    // mechanisms: [ "SCRAM-SHA-1"],
    mechanisms: ["SCRAM-SHA-256"],
}, {
    w: "majority",
    wtimeout: 5000
})
```

```Python
$ cat mongod-auth-sniffer.py
#!/usr/bin/gdb -x
# This script is a sniffer for the MongoDB's auth stuff (2025).
#
# Based on https://github.com/crossbowerbt/GDB-Python-Utils

import gdb
import binascii
import traceback
import struct


def usage():
    print("Usage:")
    print("\tsudo ./mongod-auth-sniffer -p <pid>")
    gdb.execute('quit')

class Sniffer3Breakpoint(gdb.Breakpoint):
    # Initialize the breakpoint
    def __init__(self):
        super(Sniffer3Breakpoint, self).__init__('HMAC_CTX_reset')

    # Called when the breakpoint is hit
    def stop(self):
        print("HMAC_CTX_reset")
        # return False to continue the execution of the program
        return False

class Sniffer2Breakpoint(gdb.Breakpoint):

    # Initialize the breakpoint
    def __init__(self):
        # int HMAC_Update(HMAC_CTX *ctx, const unsigned char *data, int len);
        super(Sniffer2Breakpoint, self).__init__('HMAC_Update')

    # Called when the breakpoint is hit
    def stop(self):
        # man 2 syscall
        registers = ["$rsi"]
        for register in registers:
            try:
                length = int(gdb.parse_and_eval('$rdx'))  # third argument
                address = gdb.parse_and_eval(register)
                address = int(address)
            except:
                traceback.print_exc()
                return True

            try:
                data = gdb.inferiors()[0].read_memory(address, length)
                print("HMAC_Update", register, binascii.hexlify(data), length)
                out.write("%s %s %s\n" % (register, binascii.hexlify(data), length))
                out.flush()
            except gdb.MemoryError:
                traceback.print_exc()
                return True
            except:
                traceback.print_exc()
                return True

        # return False to continue the execution of the program
        return False


class SnifferBreakpoint(gdb.Breakpoint):

    # Initialize the breakpoint
    def __init__(self):
        super(SnifferBreakpoint, self).__init__('SHA256_Update')

    # Called when the breakpoint is hit
    def stop(self):
        # man 2 syscall
        registers = ["$rsi"]
        for register in registers:
            try:
                length = int(gdb.parse_and_eval('$rdx'))  # third argument
                address = gdb.parse_and_eval(register)
                address = int(address)
            except:
                traceback.print_exc()
                return True

            try:
                data = gdb.inferiors()[0].read_memory(address, length)
                print(register, binascii.hexlify(data), length)
                out.write("%s %s %s\n" % (register, binascii.hexlify(data), length))
                out.flush()
            except gdb.MemoryError:
                traceback.print_exc()
                return True
            except:
                traceback.print_exc()
                return True

        # return False to continue the execution of the program
        return False

out = open("log.txt", "a")

# GDB setup
gdb.execute("set print repeats unlimited")
gdb.execute("set print elements unlimited")
gdb.execute("set pagination off")
gdb.execute("handle SIGUSR1 nostop")
gdb.execute("handle SIGUSR1 noprint")

# generate sniffer breakpoint
SnifferBreakpoint()

Sniffer2Breakpoint()

Sniffer3Breakpoint()

# run and sniff
gdb.execute('continue')

# gdb.execute('detach')
# gdb.execute('quit')
```

src/mongodb_scram_sha256_fmt_plug.c
kholia: Done - thanks!

ghost (approved):
Since this meets solar's criteria and since Dhiru may need this merged for use, merging.

solardiz:
Thank you @kholia for your contribution!

One pending task is to see how this compares against what hashcat has. If they also support these things, then we'd want to use the same string encodings as them.

> The following scripts were helpful in my debugging work:

Maybe you can add copyright+license statements to them? At least in a comment here, but maybe also add them to the tree - but I don't know exactly where?

> $ cat mongod-auth-sniffer.py
> #!/usr/bin/gdb -x

Perhaps the shebang line here is wrong?

src/scram_sha256_fmt_plug.c
```diff
@@ -0,0 +1,306 @@
...
+#define BINARY_SIZE             32
+#define BINARY_ALIGN            sizeof(uint32_t)
+#define BENCHMARK_COMMENT       ""
+#define BENCHMARK_LENGTH        0x107
+#define FORMAT_TAG              "$scram-pbkdf2-sha256$"
+#define FORMAT_TAG_LENGTH       (sizeof(FORMAT_TAG) - 1)
+#define MAX_USERNAME_LENGTH     128
+
+#ifndef OMP_SCALE
+#define OMP_SCALE               1 // MKPC and scale tuned for i7
```
solardiz, line 55: Did you in fact do any tuning or is this a copy-paste leftover from the other format? Either way, these values don't look like they're optimal for either format.

solardiz: Oh, I see the previous values came from:

```
commit be422692c66478e59d0b614b34f2ee6cb84ed0de
Date:   Tue Apr 9 22:35:14 2019 +0200

    OMP_SCALE & MKPC tuning, many formats, Closes #3091
```

So there was in fact some tuning of the related format, and it may not be that bad.

PR #5715, opened 2025-03-24
ZIP formats: Honor 64-bit data length

### LINKED ISSUE #5712: zip-opencl comp_len truncation from 64 to 32 bits
https://github.com/openwall/john/issues/5709#issuecomment-2746621611

>>> Not the issue here, but while we're at it - salt->comp_len is uint64_t, but then we assign it to uint in the kernel, and I don't see anything in host code to ensure it'd fit in 32-bit - maybe we should add such checks.

>> Or trivially fix the kernal? It looks like a bug to me.

> It might have been a quick (dumb) way out of the problem that the shared HMAC-SHA1 code has it as uint. But that in turn should also be trivial to fix.

magnumripper:
This includes changing shared OpenCL HMAC-SHA1 to support this.

Closes #5712

### DIFF
--- a/run/opencl/opencl_hmac_sha1.h
+++ b/run/opencl/opencl_hmac_sha1.h
@@ -24,8 +24,12 @@
 #define HMAC_OUT_TYPE
 #endif
 
+#ifndef SHA1_DATA_LENGTH_TYPE
+#define SHA1_DATA_LENGTH_TYPE	uint
+#endif
+
 INLINE void hmac_sha1(HMAC_KEY_TYPE void *_key, uint key_len,
-                      HMAC_MSG_TYPE void *_data, uint data_len,
+                      HMAC_MSG_TYPE void *_data, SHA1_DATA_LENGTH_TYPE data_len,
                       HMAC_OUT_TYPE void *_digest, uint digest_len)
 {
 	HMAC_KEY_TYPE uchar *key = _key;
@@ -74,7 +78,7 @@ INLINE void hmac_sha1(HMAC_KEY_TYPE void *_key, uint key_len,
 	SHA1_Update(&ctx, u.buf, 64);
 #ifdef USE_DATA_BUF
 	HMAC_MSG_TYPE uint *data32 = (HMAC_MSG_TYPE uint*)_data;
-	uint blocks = data_len / 64;
+	SHA1_DATA_LENGTH_TYPE blocks = data_len / 64;
 	data_len -= 64 * blocks;
 	data += 64 * blocks;
 	ctx.total += 64 * blocks;
--- a/run/opencl/opencl_sha1_ctx.h
+++ b/run/opencl/opencl_sha1_ctx.h
@@ -40,12 +40,16 @@
 #include "opencl_misc.h"
 #include "opencl_sha1.h"
 
+#ifndef SHA1_DATA_LENGTH_TYPE
+#define SHA1_DATA_LENGTH_TYPE	uint
+#endif
+
 /*
  * SHA-1 context setup
  */
 
 typedef struct {
-	uint total;        /* number of bytes processed  */
+	SHA1_DATA_LENGTH_TYPE total;    /* number of bytes processed  */
 	uint state[5];     /* intermediate digest state  */
 	uchar buffer[64];  /* data block being processed */
 } SHA_CTX;
@@ -125,7 +129,7 @@ INLINE void _sha1_process(SHA_CTX *ctx, const uchar data[64])
 /*
  * SHA-1 process buffer
  */
-INLINE void SHA1_Update(SHA_CTX *ctx, const uchar *input, uint ilen)
+INLINE void SHA1_Update(SHA_CTX *ctx, const uchar *input, SHA1_DATA_LENGTH_TYPE ilen)
 {
 	uint fill;
 	uint left;
--- a/run/opencl/zip_kernel.cl
+++ b/run/opencl/zip_kernel.cl
@@ -8,6 +8,7 @@
 #include "opencl_misc.h"
 #define HMAC_MSG_TYPE __global const
 #define HMAC_OUT_TYPE __global
+#define SHA1_DATA_LENGTH_TYPE uint64_t
 #include "opencl_hmac_sha1.h"
 #include "opencl_sha1.h"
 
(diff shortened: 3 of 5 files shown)


solardiz (approved):
I was worried we could be introducing performance impact for other uses of the HMAC - but probably not since it'd get inlined and the compiler would see that the actually passed value is a `uint`? Still, have you run any benchmarks or looked at kernel binary size changes or register number changes in builds of other kernels that use this HMAC?

solardiz (changes requested):
Oh, this is not so easy at all. `hmac_sha1` passes `data_len` into `SHA1_Update`, which currently also only accepts a `uint`. And there's the `USE_DATA_BUF` code version which also has `uint blocks`.

magnumripper:
Back to the drawing board

magnumripper:
So I fixed the opencl_sha1_ctx.h as well. Tests on nvidia, compared to current bleeding:

ZIP itself: Binary size grew with 289 bytes. Kernel used one (1) more register. Performance varies a bit but probably didn't change.

DMG uses the HMAC-SHA1. Binary size grew from 1577956 to 1671130 (!?). A stack frame grew from 9360 to 9376 bytes but other than that I see no difference in ptxas output. Performance didn't change.

GPG uses only the affected SHA1-CTX. Binary size grew from 1340541 to 1434672 (!?). The kernel uses one (1) more register. Performance didn't change.

Those sizes growing with 100KB are puzzling.

Next is to create a test archive and see if this is worthwhile at all.

EDIT:
Tried encfs as well. Binary size grew from 1979962 to 2126108, even more than the others. One of its stack frames _shrinked_ by 32 bytes, no other difference in ptxas output. Speed varies but doesn't seem affected.

Also, no warnings or problems with an all-formats test.

magnumripper:
> Next is to create a test archive and see if this is worthwhile at all.

That was harder than expected, I thought I could do it with 7z but apparently it can only produce pkzip archives. I'll see if I can find some Windows luser willing to produce a sample file for me.

ghost:
> > Next is to create a test archive and see if this is worthwhile at all.
> 
> That was harder than expected, I thought I could do it with 7z but apparently it can only produce pkzip archives. I'll see if I can find some Windows luser willing to produce a sample file for me.

Looks like I managed to create such a file using Ubuntu (if I've understood all the requirements), but john itself can't load the file (it is too much to my poor hardware):

```sh
$ head -c 128 teste.hash 
myfile.zip/myfile:$zip2$*0*3*0*96e664ad0e2d714ea7246303831fd088*3f4e*14018fd26*4ddf34229f02c2760fe80f89727fd2beb0af378ea06eaa3c9

$ john teste.hash 
Morto # ==> (it means killed, I guess)

$ file myfile.zip 
myfile.zip: Zip archive data, at least v4.5 to extract, compression method=AES Encrypted
```

Even after allocating 17 G (VIRT field from `top`).

magnumripper:
> Looks like I managed to create such a file using Ubuntu (if I've understood all the requirements), but john itself can't load the file (it is too much to my poor hardware):

Thanks! From the look of it I'd suspect the OOM killer got you. How did you create the file?

ghost:
> > Looks like I managed to create such a file using Ubuntu (if I've understood all the requirements), but john itself can't load the file (it is too much to my poor hardware):
> 
> Thanks! From the look of it I'd suspect the OOM killer got you. How did you create the file?

Straightforward. Using the Nautilus GUI.

```sh
head -c 5G </dev/urandom >myfile
# GUI, right click, ...
# 2john
```

magnumripper:
The OpenCL format, with this fix, works. Curiously enough the CPU format doesn't but that's a separate issue.

solardiz:
> Those sizes growing with 100KB are puzzling.

To avoid this, how about we use the same approach as we use with `HMAC_MSG_TYPE`, etc. - `SHA1_DATA_LENGTH_TYPE` that would default to `uint` and be overridden to `uint64_t` only in formats where we need that?

magnumripper:
> > Those sizes growing with 100KB are puzzling.
> 
> To avoid this, how about we use the same approach as we use with `HMAC_MSG_TYPE`, etc. - `SHA1_DATA_LENGTH_TYPE` that would default to `uint` and be overridden to `uint64_t` only in formats where we need that?

Good idea, did that. 

I also added support for 64-bit length in CPU side shared HMAC_SHA1 and HMAC_SHA-2. That shouldn't lead to any problem or performance regression but I did an all-formats `-test=0` for good measure.

Both ZIP formats now crack a real > 4GB archive. This is ready to review.

solardiz (approved):

solardiz:
> I also added support for 64-bit length in CPU side shared HMAC_SHA1 and HMAC_SHA-2. That shouldn't lead to any problem or performance regression

There may be a performance regression in 32-bit builds, but I guess that's fine these days. Or we could use `unsigned long` or `ARCH_WORD` there, relying on data sizes larger than what fits in 32 bits not fitting in the address space on 32-bit anyway.

magnumripper:
> There may be a performance regression in 32-bit builds, but I guess that's fine these days. Or we could use `unsigned long` or `ARCH_WORD` there, relying on data sizes larger than what fits in 32 bits not fitting in the address space on 32-bit anyway.

I'd rather change it to just `size_t`. I must still be exhausted because I can't even find the scalar SHA1 in our tree. Are we somehow still using openssl? Sourcing `sha.h` does a `#define SHA1_Update sph_sha1` and in turns sources `sph_sha1.h` which actually has it as `size_t`, in `void sph_sha1(void *cc, const void *data, size_t len);`. But that latter function is not in our tree. What's up with that?

magnumripper:
> But that latter function is not in our tree. What's up with that?

Apparently I'm actually running libcrypto. That is openssl, right? Why do we have sph_sha1.h at all?

magnumripper:
> There may be a performance regression in 32-bit builds, but I guess that's fine these days

BTW the underlying SHA1, whichever one we use, likely has it as 64-bit anyway - at least until we implement our own scalar SHA1 for CPU (I thought we had one). So I doubt there can be any measurable performance regression.

solardiz:
> I'd rather change it to just size_t.

I agree, `size_t` would be appropriate.

> Sourcing `sha.h` does a `#define SHA1_Update sph_sha1` and in turns sources `sph_sha1.h` which actually has it as `size_t`, in `void sph_sha1(void *cc, const void *data, size_t len);`. But that latter function is not in our tree. What's up with that?

There's some preprocessor magic to define `sph_sha1`, from `sha1.c`:

```c
#define RFUN   sha1_round
#define HASH   sha1
#define BE32   1
#include "md_helper.c"
```

Issue #5688, opened 2025-03-05
7z password hash can not be matched.


linghaolu:
use 7z2john got the hash:
[...]

PR #5717, opened 2025-03-25
7z formats: Amend debug stuff

magnumripper:
For researching issues like #5688. I need to merge this before I lose it.


magnumripper:
Note that nothing is changed unless `-DDEBUG`

solardiz (approved):

PR #5724, opened 2025-03-27
autotools: simplify the SIMD name for AVX512(*)

ghost:
The AVX512BW binary is translated into BW+VL+DQ plus an implied AVX512F. Report it as just AVX-512.

----

Currently:
```sh
Version: 1.9.0-jumbo-1+bleeding-9d8ac611fc 2025-03-28 09:05:58 -0300
Build: linux-gnu 64-bit x86_64 AVX-512 AC OMP OPENCL
SIMD: AVX-512, interleaving: MD4:3 MD5:3 SHA1:1 SHA256:1 SHA512:1
AES hardware acceleration: AES-NI
CPU tests: AVX512(BW+VL+DQ)
```

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -9224,10 +9224,7 @@ ac_compiler_gnu=$ac_cv_c_compiler_gnu
 
     SIMD_NAME="Altivec2"
     ;;
-    avx512|avx512bw)
-    SIMD_NAME="AVX512BW"
-    { $as_echo "$as_me:${as_lineno-$LINENO}: Trying to force $SIMD_NAME using default method (--enable-simd=$simd)." >&5
-$as_echo "$as_me: Trying to force $SIMD_NAME using default method (--enable-simd=$simd)." >&6;}
+  avx512)
     if test 2 -gt 0; then :
   { $as_echo "$as_me:${as_lineno-$LINENO}: checking if $CC supports -mavx512bw -mavx512vl -mavx512dq w/ linking" >&5
 $as_echo_n "checking if $CC supports -mavx512bw -mavx512vl -mavx512dq w/ linking... " >&6; }
@@ -9278,8 +9275,9 @@ ac_link='$CC -o conftest$ac_exeext $CFLAGS $CPPFLAGS $LDFLAGS conftest.$ac_ext $
 ac_compiler_gnu=$ac_cv_c_compiler_gnu
 
 
+    SIMD_NAME="AVX-512"
     ;;
-  mmx|sse*|ssse3|avx*|xop*)
+    mmx|sse*|ssse3|avx*|xop*)
     SIMD_NAME=`echo $simd | tr a-z A-Z`
     { $as_echo "$as_me:${as_lineno-$LINENO}: Trying to force $SIMD_NAME using default method (--enable-simd=$simd)." >&5
 $as_echo "$as_me: Trying to force $SIMD_NAME using default method (--enable-simd=$simd)." >&6;}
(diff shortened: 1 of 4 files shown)


magnumripper:
Looks good!

> The change in `simd-intrinsics.h` is still missing.
> 
> ```c
> diff --git a/src/simd-intrinsics.h b/src/simd-intrinsics.h
> index 53814208f..b5dbaedd9 100644
> --- a/src/simd-intrinsics.h
> +++ b/src/simd-intrinsics.h
> @@ -42,7 +42,7 @@
>  #define SIMD_TYPE            "MIC"
>  #elif __AVX512BW__
>  #undef SIMD_TYPE
> -#define SIMD_TYPE            "AVX512BW"
> +#define SIMD_TYPE            "AVX512(*)" // whatever it may be
>  #elif __AVX512F__
>  #undef SIMD_TYPE
>  #define SIMD_TYPE            "AVX512F"
> ```

IMHO it too should be just `#define SIMD_TYPE            "AVX512"`, intentionally unspecific.

ghost:
There will be a side effect at, e.g.:

```C
#define ALGORITHM_NAME		BITS " " SIMD_TYPE  " " STRINGIZE(SIMD_COEF_32)
```

Anyway, I made the change locally.

magnumripper:
> There will be a side effect at, e.g.:
> 
> ```c
> #define ALGORITHM_NAME		BITS " " SIMD_TYPE  " " STRINGIZE(SIMD_COEF_32)
> ```
> 
> Anyway, I made the change locally.

OK let's hear @solardiz before we do something we'll regret.

solardiz:
Maybe we should use the name `AVX-512` (with dash, as Intel also calls it) to refer to the typical kind of it that we have on CPUs, including in `ALGORITHM_NAME`. We can continue to use `AVX512F` to refer to the cut-down version on Xeon Phi 2nd+ gen.

As to my current mention of `AVX512VL` in `ALGORITHM_NAME` for scrypt, we can keep it as-is because it clarifies why we're reporting 128/128 and not 128/512 nor 512/512.

ghost:
So:

```sh
Version: 1.9.0-jumbo-1+bleeding-9d8ac611fc 2025-03-28 09:05:58 -0300
Build: linux-gnu 64-bit x86_64 AVX-512 AC OMP OPENCL
SIMD: AVX-512, interleaving: MD4:3 MD5:3 SHA1:1 SHA256:1 SHA512:1
AES hardware acceleration: AES-NI
CPU tests: AVX512(BW+VL+DQ)
```

```sh
Testing: scrypt (16384, 8, 1) [Salsa20/8 128/128 AVX512VL]... (2xOMP) PASS
Testing: LM [DES 512/512 AVX512F]... (2xOMP) PASS
[...]
Testing: mscash, MS Cache Hash (DCC) [MD4 32/64, UTF-8]... (2xOMP) PASS
Testing: mscash2, MS Cache Hash 2 (DCC2) [PBKDF2-SHA1 512/512 AVX-512 16x, UTF-8]... (2xOMP) PASS
Testing: MSCHAPv2, C/R [MD4 DES (ESS MD5) 512/512 AVX-512 16x3, UTF-8]... PASS
```

But something bad is happening with the dash. We need to investigate.

magnumripper:
> But something bad is happening with the dash. We need to investigate.

What bad? I think everything looks correct.

ghost:
> > But something bad is happening with the dash. We need to investigate.
> 
> What bad? I think everything looks correct.

I've already force-pushed a fix.

magnumripper (approved):

Issue #5722, opened 2025-03-27
--rules-stack broken (now fixed but may still need some TLC)


solardiz:
Looks like we (perhaps I) broke it in the recent week. Before recent changes:

```
$ echo password > w
$ ./john- -w=w --rules-stack=':[lu]' -stdo
Using default input encoding: UTF-8
password
PASSWORD
2p 0:00:00:00 100.00% (2025-03-27 04:13) 40.00p/s PASSWORD
```

Now:

```
$ ./john -w=w --rules-stack=':[lu]' -stdo | head
Using default input encoding: UTF-8
Press 'q' or Ctrl-C to abort, 'h' for help, almost any other key for status
password
password
password
password
password
password
password
password
password
password
```

solardiz:
Reverting d265b3cf097510c26e80bd0d18b5779605b34c8e repairs it.

magnumripper:
Happens with PRINCE mode too (indeed any mode).

solardiz:
This hack repairs it (just to narrow down where the problem is):

```diff
+++ b/src/rules.c
@@ -1863,8 +1863,11 @@ char *rules_process_stack_all(char *key, rule_stack *ctx)
        rules_stacked_after = 0;
 
        while (ctx->rule) {
-               if ((word = rules_apply(key, ctx->rule->data, -1)))
+               static char last[99];
+               if ((word = rules_apply(key, ctx->rule->data, -1)) && strcmp(word, last)) {
+                       strcpy(last, word);
                        return word;
+               }
                if ((ctx->rule = ctx->rule->next)) {
                        rules_stacked_number++;
                        if (!stack_rules_mute)
```

solardiz:
So _somehow_ `rules_process_stack_all` depended on this check against last word, which `rules_apply` no longer has. We should remove this dependency. @magnumripper I'm very tired now and I think stacked rules is your code, so maybe you could take over?

magnumripper:
I'll look into it right away

solardiz:
Thank you! I'm sorry I didn't test this feature before merging - I should have, as it was obviously touched by the changes.

solardiz:
This almost works:

```diff
+++ b/src/rules.c
@@ -1863,9 +1863,11 @@ char *rules_process_stack_all(char *key, rule_stack *ctx)
        rules_stacked_after = 0;
 
        while (ctx->rule) {
-               if ((word = rules_apply(key, ctx->rule->data, -1)))
[...]

PR #5725, opened 2025-03-27
rules.c: Simplify the latest bugfix to stacked rules

magnumripper:
Just drop a couple of identical `log_event()` and move the last one so it covers all three

Also as a separate commit, add a log call just before session end that says how many candidates were processed.

See #5722

### DIFF
--- a/src/john.c
+++ b/src/john.c
@@ -1,8 +1,9 @@
 /*
  * This file is part of John the Ripper password cracker,
  * Copyright (c) 1996-2024 by Solar Designer
- *
- * ...with changes in jumbo by JimF, magnum, Claudio, and several others
+ * Copyright (c) 2009-2025, magnum
+ * Copyright (c) 2021, Claudio
+ * Copyright (c) 2009-2018, JimF
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted.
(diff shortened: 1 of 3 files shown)


magnumripper:
Also added a commit requiring debug verbosity in order to log all stacked rules (otherwise only the first run of the stacked rules are logged, then a "_Some rule logging suppressed. Re-enable with --verbosity=6 or greater_"). Without this, we can very easily end up with petabytes worth of log file, and performance should plummet.

solardiz (approved):
Looks good to me assuming you've tested it, but there's a lot more room for simplification and optimization.

src/john.c
```diff
@@ -1947,6 +1948,7 @@ static void john_done(void)
 			/* We already printed to stderr from signals.c */
 			log_event("%s", abort_msg);
 		} else if (children_ok) {
+			log_event("Candidates tried: %"PRIu64"p", status.cands);
```
solardiz, line 1951: I'd also log suppressor's hits and misses, which is info that is otherwise impossible to access at session end (no way to press `s` at exactly the right moment).

magnumripper: That would be a new function in status.c, right? You could create a ticket for it.

solardiz: > That would be a new function in status.c, right?

No, it's `status.suppressor_hit` and `status.suppressor_miss` just like `status.cands` that you use.

Issue #1908, opened 2015-11-23
LM/DEScrypt OpenCL performance


magnumripper:
We should try to match the performance seen with Meriken's tripcode engine, and now oclHashcat, for our LM and DEScrypt implementations. Something like 20Gc/s for LM and 800Mc/s for DEScrypt on Titan X.

The S-boxes for LOP3.LUT are already implemented. The bitselect ones might need a review too though.

Some hints from "DeepLearningJohnDoe": 
http://www.openwall.com/lists/john-users/2015/10/10/3
https://github.com/DeepLearningJohnDoe/merikens-tripcode-engine
https://github.com/DeepLearningJohnDoe/SLUT
https://github.com/DeepLearningJohnDoe/SBOXDiscovery

Hint from Atom: _There's only just one thing in NV that's important. One need to call s1-s8 inside a loop (i=0;i<2;i++) and #pragma unroll 1_

magnumripper:
Oh BTW oclHashcat doesn't use a single goto in any of its kernels, including descrypt. Other than checking that (with grep) I haven't looked at the descrypt kernel differences at all. But it's fair game to do so now.

magnumripper:
FWIW lm-opencl is almost **twice as fast** on Titan X (also a little faster on Intel CPU) if changing opencl_lm_b_plug.c so it doesn't pass `-D WORK_GROUP_SIZE` to the kernel build.

magnumripper:
Atom now managed to get another 8-14% boost in https://github.com/hashcat/oclHashcat/commit/245301c9b444616aa083ef8fda83decbc690f4c1#commitcomment-15457781 - now 12G c/s for LM and 405M c/s for descrypt.

meriken:
[Meriken's Tripcode Engine](https://github.com/meriken/merikens-tripcode-engine) recently got significant boosts with new kernels written in GCN assembly with [CLRadeonExtender](http://clrx.nativeboinc.org/).

AMD Radeon HD 7990 **1022MH/s** (descrypt; 1250mV +20% 1180MHz)
NVIDIA GeForce GTX 980 Ti **996MH/s** (descrypt; 110% +250MHz)
AMD Radeon HD 290X **647MH/s** (descrypt; +100mV +50% 1074MHz)

 I highly recommend CLRadeonExtender as it is cross-platform and very reliable.

magnumripper:
Thanks @meriken for the heads-up. I presume the nvidia speedup came from other improvements?

[...]

Issue #5707, opened 2025-03-23
Use LUT3 for bitslice DES key setup


solardiz:
We can use our familiar LUT3 instructions for the key setup. We currently use a lot of AND-shift-OR sequences. We can instead move the shift out, so that we can do the AND-OR in one LUT3. A drawback is that we'd need many AND masks per output bit. We already have those masks prepared, but within processing of one bit we may currently preload just one mask into a register. With 32 AVX-512 registers, the compiler would probably keep all 8 preloaded in registers anyway, so we can reasonably try this change.

PoC for 1 bit on CPU:

```diff
+++ b/src/DES_bs_b.c
@@ -451,12 +451,18 @@ typedef unsigned ARCH_WORD kvtype;
        kvtype m = mask01, va, vb, tmp; \
        kvand(va, v0, m); \
        kvand_shl1(vb, v1, m); \
-       kvand_shl_or(va, v2, m, 2); \
-       kvand_shl_or(vb, v3, m, 3); \
-       kvand_shl_or(va, v4, m, 4); \
-       kvand_shl_or(vb, v5, m, 5); \
-       kvand_shl_or(va, v6, m, 6); \
-       kvand_shl_or(vb, v7, m, 7); \
+       kvshl(tmp, v2, 2); \
+       va = vlut3(va, tmp, mask04, 0xf8); \
+       kvshl(tmp, v2, 3); \
+       vb = vlut3(vb, tmp, mask08, 0xf8); \
+       kvshl(tmp, v2, 4); \
+       va = vlut3(va, tmp, mask10, 0xf8); \
+       kvshl(tmp, v2, 5); \
+       vb = vlut3(vb, tmp, mask20, 0xf8); \
+       kvshl(tmp, v2, 6); \
+       va = vlut3(va, tmp, mask40, 0xf8); \
+       kvshl(tmp, v2, 7); \
+       vb = vlut3(vb, tmp, mask80, 0xf8); \
        kvor(*(kvtype *)kp, va, vb); \
        kp++; \
 }
```

Works, code size is smaller, but will fail to build on archs without a LUT3, so real changes will need to make this conditional. Also, in the generated code the masks are loaded from memory operands. Maybe this would change when we do it for all bits, so the compiler would see greater constant demand for those masks.

We'd probably also want this in OpenCL.

solardiz:
> real changes will need to make this conditional.

I implemented this now and pushed into my current PR.

[...]

PR #5726, opened 2025-03-27
OpenCL Bitslice DES key setup: Use LUT3 on nvidias

magnumripper:
There's no change in performance.

The non-mask (self test) kernel uses 2 more registers (36 vs 34) with LUT3, and binary size grew with 928 bytes from 6828 to 7756.
The mask kernel stays at 64 registers, but binary size grew with 922 bytes from 9891 to 10813.

Should I leave LUT3 enabled here? I could just add an `#undef HAVE_LUT3` somewhere, with a comment.

#5707

### DIFF
--- a/run/opencl/DES_bs_finalize_keys_kernel.cl
+++ b/run/opencl/DES_bs_finalize_keys_kernel.cl
@@ -1,5 +1,7 @@
 /*
- * This software is Copyright (c) 2015 Sayantan Datta <std2048 at gmail dot com>
+ * This software is Copyright 2011-2025 Solar Designer,
+ * Copyright (c) 2015 Sayantan Datta <std2048 at gmail dot com>,
+ * Copyright (c) 2025 magnum
  * and it is hereby released to the general public under the following terms:
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted.
@@ -7,6 +9,7 @@
  */
 #include "opencl_DES_kernel_params.h"
 #include "opencl_mask.h"
+#include "opencl_misc.h"
 
 #if 1
 #define MAYBE_GLOBAL __global
@@ -20,6 +23,7 @@
 #define kvshl1 vshl1
 #define kvshl vshl
 #define kvshr vshr
+#define vlut3 lut3
 
 #define mask01 0x01010101
 #define mask02 0x02020202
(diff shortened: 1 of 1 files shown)


solardiz:
> There's no change in performance.
> 
> The non-mask (self test) kernel uses 2 more registers (36 vs 34) with LUT3, and binary size grew with 928 bytes from 6828 to 7756. The mask kernel stays at 64 registers, but binary size grew with 922 bytes from 9891 to 10813.

What driver/device is this on? Maybe test on more?

> Should I leave LUT3 enabled here? I could just add an `#undef HAVE_LUT3` somewhere, with a comment.

If this only gives code size increase in our testing so far, then let's "add an `#undef HAVE_LUT3` somewhere, with a comment" as you suggest.

Also, the host code verbosity check change should be a separate commit.

Oh, and CircleCI is somehow failing here - hopefully, intermittently, and will work when you push updates.

run/opencl/DES_bs_finalize_keys_kernel.cl
```diff
@@ -20,6 +22,7 @@
 #define kvshl1 vshl1
 #define kvshl vshl
 #define kvshr vshr
+#define vlut3 lut3
```
solardiz, line 26: I don't think defining `vlut3` clarifies anything here - we could as well use `lut3` in one place below directly.

solardiz:
> There's no change in performance.

Also, for which formats? For `descrypt-opencl`, this is probably too little of a change to measure. For `lm-opencl`, it may be significant.

run/opencl/DES_bs_finalize_keys_kernel.cl
magnumripper: I like it to be very near identical with your CPU code.

magnumripper:
> Also, for which formats? For `descrypt-opencl`, this is probably too little of a change to measure. For `lm-opencl`, it may be significant.

Good call, I was going to test LM after DEScrypt but then I forgot about it. Will test LM now, let's hope for better results!

solardiz:
Also, our speeds at DES at least on older NVIDIA are poor because we never fixed #1908. So maybe also benchmark on AMD? Or are you getting good speeds on your NVIDIA (e.g., on par with hashcat's)?

magnumripper:
Instructions unclear... are you saying I should benchmark LUT3 on AMD?

magnumripper:
For LM, I got the same ptxas output, same binary size (for a given LWS, see below) and seemingly the same speed, although that one varies a tad as usual. Best speed of five runs is 18362M for both (2080ti 565.57.01).

The LWS is told to the kernel with eg. `-D WORK_GROUP_SIZE=64` for firm-coded adaptions, so during autotune there are many builds with varying binary sizes. If that define isn't made at all, the code ends up different and local memory isn't used. But the speed is virtually unaffected by this (in #1908 I claimed this to give almost 2x boost on Titan X).

Even weirder, if auto-tune comes up with LWS=128 GWS=131072 and I reuse those figures, the speed consistently drops by some 600-800Mp/s. I never noticed that before. Must be something very odd with the non-standard autotune.

magnumripper:
I added the `#undef LUT3` in a separate commit. I'm done with this for now.

solardiz:
> Instructions unclear... are you saying I should benchmark LUT3 on AMD?

That's what I thought, but I forgot we only have LUT3 on NVIDIA. So there's nothing to do on this.

> I added the #undef LUT3 in a separate commit. I'm done with this for now.

OK to merge this. CircleCI is still failing, but perhaps it's your quota with them or something like that.

run/opencl/DES_bs_finalize_keys_kernel.cl
```diff
@@ -30,31 +33,54 @@
...
 
-#define kvand_shl1_or(dst, src, mask) 			\
-	kvand(tmp, src, mask); 				\
-	kvshl1(tmp, tmp); 				\
-	kvor(dst, dst, tmp)
+/*
+ * LUT3 did not do any good in our testing (2025-03) and the binary size
+ * for DEScrypt kernel grew a little, so this is defined out for now.
+ */
+#undef LUT3
```
solardiz: You must have meant `HAVE_LUT3` here, not just `LUT3`.

run/opencl/DES_bs_finalize_keys_kernel.cl
```diff
@@ -1,12 +1,14 @@
 /*
  * This software is Copyright (c) 2015 Sayantan Datta <std2048 at gmail dot com>
+ * and Copyright (c) 2025 magnum
```
solardiz: Probably should add my copyright as well, 2011-2025 I think.

solardiz (changes requested):

solardiz:
> > I added the #undef LUT3 in a separate commit. I'm done with this for now.
> 
> OK to merge this.

Sorry, didn't notice the typo in there at first. Please fix that before merging.

run/opencl/DES_bs_finalize_keys_kernel.cl
magnumripper: Oops

PR #5731, opened 2025-03-28
john.c: At session exit, log important numbers

ghost:
You guys are missing this or `11da3ab` moved it to another place?
****
Log the number of candidates processed and suppressor's hits and misses.

Among other things, this is great for debugging.


magnumripper:
This should be covered by 11da3ab52, did you base this on latest HEAD? I think there was never a PR for it so was easy to miss. This commit doesn't seem to have 11da3ab52.

PR #5738, opened 2025-03-29
Added support for EC SSH keys encrypted with AES-256-CBC

ghost:
We can merge this as a first draft and then revise it with subsequent commits.

To supersede 5642 (it've been waiting too long).


solardiz:
Thank you @claudioandre-br. This _may_ be reasonable to merge, but re-reading the comments in #5642 it looks like @peshev thought those heuristics you found are more of a coincidence and may not be reliable and the salt length check does nothing?

What SSH keys did you test this on? How were they generated? Would similar keys be generated by tools/procedures people commonly use?

We should probably also add a test vector for this new functionality to the `tests` array.

@peshev @Nothing4You You mentioned "a CTF-style challenge, part of 38C3 - [TELNET-KLARTEXT-REDEN](https://events.ccc.de/congress/2024/hub/en/assembly/TELNET-KLARTEXT-REDEN/)". That link seems broken now. Anyhow, maybe you can share the SSH keys + cracked passwords from that challenge, to use as our test vectors? Thank you!

solardiz:
I think it'd be best for more SSH key samples to be added via a PR to https://github.com/openwall/john-samples/tree/main/SSH

ghost:
I've added a test vector provided to us by https://github.com/openwall/john/pull/5641#issuecomment-2564723781

ghost:
> What SSH keys did you test this on? How were they generated? Would similar keys be generated by tools/procedures people commonly use?

From https://github.com/openwall/john/pull/5642#issuecomment-2607192448


```sh
echo "" > solucao.txt

for i in {1..32}; do
    echo "-- Iteration: $i"
    echo $string >> solucao.txt
    echo $string > johnpass
done

# Extracting hashes

# Testing john
rm -f ../run/john.pot; ../run/john --wordlist=solucao.txt ../ssh-hashes.txt
```

peshev:
I just tested it, and it works with my original test case. I'm happy with this implementation - I'm closing my original PR.

solardiz (approved):

solardiz:
So I merged this, but we already have a regression (opened issue) and I'm also concerned the heuristics may not catch all EC keys, which would result in false negatives, right? Maybe in `ssh2john.py`, we need to pre-check for this same condition, and if it's not met for a `ktype == 3` key, then don't output a "hash" as it wouldn't be handled correctly.

We may also want to look at how exactly hashcat handles these - does it also use heuristics or not? The same check or not? This could affect what we do.

solardiz:
> the heuristics may not catch all EC keys, which would result in false negatives, right?

As an option, we could relax the check from "exactly 224" to "below 300" or something like that, but ideally we'd first have better understanding of where the number 224 comes from. (This many _bytes_ is many times higher than the _bit_ sizes of EC keys.)

solardiz:
> We may also want to look at how exactly hashcat handles these - does it also use heuristics or not?

I took a look. It appears not to have this sort of check. Perhaps we can instead revise our ASN.1 checks such that they're usable for non-EC keys as well.

ghost:
BTW:

The GitHub GUI did the right thing:
- It kept the two authors plus the person responsible for the merge;

It discreetly removed my signature, but the rebase would have done it anyway.

solardiz:
> Perhaps we can instead revise our ASN.1 checks such that they're usable for non-EC keys as well.

Omitting this check works:

```c
        // OCTET STRING
        if (asn1_get_next(pos, end - pos, &hdr) < 0 ||
                        hdr.class != ASN1_CLASS_UNIVERSAL ||
                        hdr.tag != ASN1_TAG_OCTETSTRING) {
              goto bad;
```

but then I don't know if we have enough checking left to avoid false positives?

hashcat doesn't use a real ASN.1 parser, instead doing:

```c
DECLSPEC int asn1_detect (PRIVATE_AS const u32 *buf, const int len)
{
  if (len < 128)
  {
    if ((buf[0] & 0x00ff80ff) != 0x00020030) return 0;
  }
  else if (len < 256)
  {
    if ((buf[0] & 0xff00ffff) != 0x02008130) return 0;
  }
  else if (len < 65536)
  {
    if ((buf[0] & 0x0000ffff) != 0x00008230) return 0;
    if ((buf[1] & 0x000000ff) != 0x00000002) return 0;
  }

  if (len < 128)
  {
    const int lenb = ((buf[0] & 0x00007f00) >>  8);

    if ((lenb + 2) != len) return 0;
  }
  else if (len < 256)
  {
    const int lenb = ((buf[0] & 0x00ff0000) >> 16);

    if ((lenb + 3) != len) return 0;
  }
  else if (len < 65536)
  {
    const int lenb = ((buf[0] & 0xff000000) >> 24)
                   | ((buf[0] & 0x00ff0000) >>  8);

    if ((lenb + 4) != len) return 0;
  }

  return 1;
}
```

PR #5740, opened 2025-03-29
Change shared CPU HMAC code to use size_t for datalen instead of uint64_t

magnumripper:
This change means 32-bit hosts won't support ZIP files of 4 GiB or larger. They will be reject with a warning and hints.

Also changes ZIP cost to report size in KiB, instead of overflowing the unsigned int at 4 GiB or larger.

### DIFF
--- a/src/pkzip.c
+++ b/src/pkzip.c
@@ -248,3 +248,10 @@ void *winzip_common_binary(char *ciphertext) {
 	}
 	return bin;
 }
+
+unsigned int winzip_common_cost_hmac_len(void *salt)
+{
+	winzip_salt *s = *((winzip_salt**)salt);
+
+	return (s->comp_len + 1023) / 1024;
+}
--- a/src/hmac_sha.h
+++ b/src/hmac_sha.h
@@ -16,19 +16,19 @@
 
 #ifndef _JTR_HMAC_SHA_H
 
-extern void JTR_hmac_sha1(const unsigned char *key, int key_len, const unsigned char *data, uint64_t data_len, unsigned char *digest, int digest_len);
+extern void JTR_hmac_sha1(const unsigned char *key, int key_len, const unsigned char *data, size_t data_len, unsigned char *digest, int digest_len);
 #define hmac_sha1(key,keylen,data,datalen,dgst,dgstlen) JTR_hmac_sha1(key,keylen,data,datalen,dgst,dgstlen)
 
-extern void JTR_hmac_sha256(const unsigned char *key, int key_len, const unsigned char *data, uint64_t data_len, unsigned char *digest, int digest_len);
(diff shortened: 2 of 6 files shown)


src/pkzip.c
```diff
@@ -120,6 +123,21 @@ int winzip_common_valid(char *ciphertext, struct fmt_main *self)
 	}
 	sscanf((const char*)cp, "%"PRIx64, &val);
 
+	if (self == &fmt_zip && sizeof(size_t) < 8 && val > UINT_MAX) {
+		static int warned;
+
+		if (!warned) {
+			if (john_main_process && !ldr_in_pot) {
+				fprintf(stderr, "%s: Warning: Not loading hashes larger than 4 GiB. Use a 64-bit build or zip-opencl\n", self->params.label);
```
solardiz: This is probably dead code on reasonable systems where `size_t` reflects the address space size, because we'd only reach this when we already have 8+ GiB worth of hex in the address space, so are not on 32-bit.

Also, why the nested `if`'s - perhaps one `if` for all conditions would be enough.

src/pkzip.c
```diff
@@ -248,3 +266,10 @@ void *winzip_common_binary(char *ciphertext) {
 	}
 	return bin;
 }
+
+unsigned int winzip_common_cost_hmac_len(void *salt)
+{
+	winzip_salt *s = *((winzip_salt**)salt);
+
+	return (s->comp_len + 1023) / 1024;
```
solardiz, line 256: This will silently overflow above 4 TiB. Perhaps pre-check for that in `valid` and refuse to load, for 64-bit hosts with 8+ TiB of RAM.

solardiz (approved):
I've suggested a couple of things, but this is also OK to merge as-is.

src/pkzip.c
magnumripper: > This is probably dead code on reasonable systems where size_t reflects the address space size, because we'd only reach this when we already have 8+ GiB worth of hex in the address space, so are not on 32-bit.

You are right, I should drop this. Perhaps we need a catch-all length check in fgetl. If its first read detect no LF (or EOF) so exceeds LINE_BUFFER_SIZE, before continuing it could scan the line size by calling a separate function that doesn't copy anything and just return a uint64_t line length. With that info, fgetl could have a catch-all "reject" for overlong lines (for host, memory size or otherwise unreasonable). Warning or not, fgetl would return an empty line to the caller.

> why the nested if's - perhaps one if for all conditions would be enough

True but the compiler would emit the same result.

src/pkzip.c
magnumripper: > This will silently overflow above 4 TiB. Perhaps pre-check for that in `valid` and refuse to load, for 64-bit hosts with 8+ TiB of RAM.

So 8+ TB of input... Even if that could theoretically reach this code, I think this is overkill. And it would take so very long time to get here that any user would cancel the run.

PR #5743, opened 2025-03-30
fix: ssh-opencl cannot handle EC AES-256-CBC test vector

### LINKED ISSUE #5742: ssh-opencl fails self-test after addition of EC AES-256-CBC test vector
After merging #5738, ssh-opencl fails self-test because we didn't add support for this EC key type in there yet added the test vector for both formats. We should either add support also to ssh-opencl or have its `valid` reject such inputs. cc: @peshev @Nothing4You

ghost:
Prior to this commit, ssh-opencl was failing self-test after adding the EC AES-256-CBC test vector.

Fix: #5742.

IMO: this is a better approach (no need to mess with the OpenCL format).


solardiz:
Wouldn't this let people load those hashes into that format and have 100% false negatives? If so, this is wrong, and the self-test failure was right and shouldn't be merely silenced.

ghost:
> Wouldn't this let people load those hashes into that format and have 100% false negatives? If so, this is wrong, and the self-test failure was right and shouldn't be merely silenced.

But this is another error. In fact, is it an error in the code that has always existed in the code or that has existed since the 2john tool was able to produce such hashes?

solardiz:
> But this is another error. In fact, is it an error in the code that has always existed in the code or that has existed since the 2john tool was able to produce such hashes?

OK, I agree. We can fix the self-test failure first to let people use the format like before. Then fix the format separately.

solardiz (approved):

PR #5745, opened 2025-03-30
Rework the SSH CPU format

solardiz:
This integrates the EC-specific ASN.1 checks we had into the function shared with non-EC, so that it would detect either kind of key without prior knowledge. Besides simplification, this should reduce the risk of false negatives, at a benign increase of risk of false positives. cc: @peshev @Nothing4You

I might add a few other commits to this PR, and will edit its title if so.


solardiz:
In fact, I think the length check and the `strict_mode` checks are excessive. By the time we reach them, we seem to have checked at least ~55 bits. We may want to drop them in order to reduce risk of false negatives in case any valid key differs from our strict expectations.

solardiz:
@kholia I am removing a function you added in 609a5d84b96a6ab1f30a8b1c7ee84776c559b589 and I'm contemplating removing some more checks you had added. You might want to review. Any specific reasons why you felt `strict_mode` was needed?

ghost:
A change to `ssh_valid()` is still required after this PR. #5746 is useless after this change.

solardiz:
> A change to `ssh_valid()` is still required after this PR.

Yes, I'm not fixing the OpenCL format to reject or properly handle anything extra in this PR.

solardiz:
> I think the length check and the `strict_mode` checks are excessive. By the time we reach them, we seem to have checked at least ~55 bits. We may want to drop them in order to reduce risk of false negatives in case any valid key differs from our strict expectations.

Added a commit to drop them, which (along with allowing ASN1_TAG_SEQUENCE for the last field we do check) allows us to enable checking most of the previously-disabled hashcat test vectors (except for one, which is for type 6 that we don't currently support and I don't yet know what it is).

ghost:
I copied these changes to the OpenCL format and then it passed the self-test with EC AES-256-CBC enabled [1] and my random test [2].

[1] test vector provided to us by https://github.com/openwall/john/pull/5641#issuecomment-2564723781
```sh
../run/john-opencl --test -v=5 --format=ssh-opencl
Benchmarking: ssh-opencl, SSH private key [RSA/DSA/EC 3DES/AES OpenCL]... 
Loaded 15 hashes with 15 different salts to test db from test vectors
```
[2]
```sh
echo "" > solucao.txt

for i in {1..32}; do
    echo "-- Iteration: $i"
    echo $string >> solucao.txt
    echo $string > johnpass
done

# Extracting hashes

# Testing john
rm -f ../run/john.pot; ../run/john-opencl --wordlist=solucao.txt ../ssh-hashes.txt
```

The change itself is simple (basically a copy+paste). The problem is that I have no idea how much it does (or doesn't) affect performance on a real GPU.

If the format for the CPU is correct, the OpenCL format should also be correct.

src/ssh_variable_code.h
```diff
@@ -3,10 +3,12 @@
...
...0b52449251cb35e183f5b309a09071535154c6f1013b58f305b544f3589c9eb0e9ac4267a84374a3eab49c53aa9bedbf97f8f19ebc212d8db74ee03554a3514140667fa4ce8e06aad3f32d1b00015be0e8979fe66736018589beee06d6f318851dbe8d9689e70202185d71fc5e5a3d2996ddb8ae1d7718c49855c6f8c43301e0915f324f30d0d9c6a8504a91ad5a7179aafb87ede58598394949910874850994abe815817359152ff6a7c8cc6f19524dfc5e50ddfd038a2275bf809e3c8f05ed3e3137ebd62d91cd3578533787c3847e3c5e07e5a891480e5ceabcf6c344e7bec8b640ab9a03e90b846b35d2f46ba150accef32d2597b064810b15fd54fca6d2b146feabcd05c0b51617ae95e36f6817a62c3ff42c5c2f6f1d20a8a1fd334d3b7d3f83bba057b79d9b5508bb0cb706ba00acb0ab797401fdcfac80b5b6e38e51aec0b38f33ff4690425ca28d88a2e876591521230150b4e20a4a82e50061cee9c0705100bfe5fdbd8ef27aec20387cf32455ef305bce2a91ae6da91fc41376b97149e9b41c901b24811df9272ff09718923b8d94e8e459a164a22b0eca47653f3efcbf08188c5da78cd9fb9eda1761094f9d8bc3d479e9f40c7d79ebaaba2a5c632329f20a9962040ff8f512b42c5f32a8460d87b8e93c6f980a1562c436eea1c8994fbf671dda3c4ccd3c142acfcdde2ab61227289ad408213ac8e22d9ef487f36925f5ba3b8e7e913d25c4a8592c861d13f03b615bc2760aabc61d68db80d35296a3312fdf4b56c0fbee5ab3fea1cf9caf3960a564046939e8002d2dd909db446d85aeae9dd42a33fe28684f722172e6", "hashcat"},
```
ghost, line 9: This comment is inaccurate. 
Note that the vector is of type 6 and yet looks like type 0 above. You can find a "real" type 6 at:

https://github.com/openwall/john/blob/17ea3ab69eca091cf5fe6cde9931ccd4050db5c0/src/ssh_variable_code.h#L40

It has two extra fields "$16$183"

rounds  and  ciphertext_begin_offset. I have no idea what it really is and how to deal with it.

solardiz: Yes, I figured out the same and already experimented with this briefly. Still there's more to figure out.

solardiz: Per hashcat code, type 6 looks like it's single DES without a KDF (not even MD5). I wonder where that was used.

solardiz: Looks like a clash between hashcat and us. hashcat seems to have added type 6 on Aug 3, 2020 and we on Aug 31, 2020, but with different meaning. Maybe we should renumber ours to 7.

solardiz: > Per hashcat code, type 6 looks like it's single DES without a KDF (not even MD5).

I misread. It is single DES, but _with_ MD5 as KDF. And I got it to work in JtR.

solardiz: Pushed these changes - add single DES, renumber our 6 to 7. The `ssh2john.py` addition of single DES is untested - need an ancient enough SSH key file, I guess.

solardiz:
There's a lot of stuff in this PR now (15 commits). There's still a lot more room for improvement, but I think I'll stop here, unless there are regressions or major issues to fix. Then we'll need to bring the OpenCL format to parity with the CPU format, which @claudioandre-br has already started successfully experimenting with.

I'd appreciate review and testing, especially by @peshev @Nothing4You @vkhromov and indeed by @kholia @claudioandre-br @magnumripper. Thanks!

solardiz:
This changes default benchmark speeds because the hashcat test vectors are now at the start of the array. We need to decide on which two we want to use for benchmarking, ideally two of the same type.

solardiz:
> This changes default benchmark speeds because the hashcat test vectors are now at the start of the array. We need to decide on which two we want to use for benchmarking, ideally two of the same type.

Actually, I think it's just one when we benchmark as Raw, and I've just moved a test vector to make things similar to what we had before, yet for length 7 for consistency with hashcat benchmarks.

kholia (approved):
LGTM.

Nice work!

I don't remember why I chose to check more bits when I originally wrote this a long time back. Yep, checking ~55 bits does seem excessive to me now.

solardiz:
Hmm, no, this is not ready - I am in fact getting false positives, so I must have been wrong about how much we check. We'll need to figure this out.

solardiz:
Using our previous first test vector:

```
$ ./john pw-ssh -mask='strongpa?l?l?l?l?l?l'
Warning: detected hash type "SSH", but the string is also recognized as "ssh-opencl"
Use the "--format=ssh-opencl" option to force loading these as that type instead
Loaded 1 password hash (SSH, SSH private key [MD5/bcrypt-pbkdf/[3]DES/AES 32/64])
Cost 1 (KDF/cipher [0:MD5/AES 1:MD5/[3]DES 2:bcrypt-pbkdf/AES]) is 0 for all loaded hashes
Cost 2 (iteration count) is 1 for all loaded hashes
Will run 8 OpenMP threads
strongpassword   (?)     
1g 0:00:00:20 DONE (2025-03-31 22:20) 0.04796g/s 6381Kp/s 6381Kc/s 6381KC/s strongpaxnword..strongpamsword
Session completed. 
$ rm john.pot john.log john.rec
rm: cannot remove `john.rec': No such file or directory
$ ./john pw-ssh -mask='?l?lrongpass?l?l?l?l'
Warning: detected hash type "SSH", but the string is also recognized as "ssh-opencl"
Use the "--format=ssh-opencl" option to force loading these as that type instead
Loaded 1 password hash (SSH, SSH private key [MD5/bcrypt-pbkdf/[3]DES/AES 32/64])
Cost 1 (KDF/cipher [0:MD5/AES 1:MD5/[3]DES 2:bcrypt-pbkdf/AES]) is 0 for all loaded hashes
Cost 2 (iteration count) is 1 for all loaded hashes
Will run 8 OpenMP threads
uirongpassccto   (?)     
1g 0:00:00:06 DONE (2025-03-31 22:20) 0.1585g/s 6257Kp/s 6257Kc/s 6257KC/s uerongpassccto..qorongpassccto
```

but this appears to be a lot of poor luck (or good luck for us to find out before merging this PR) to hit a false positive so quickly. On another machine, I have this running with 8-character mask `-mask='strong?l?l?l?l?l?l?l?l'` for 5 minutes and still nothing.

solardiz:
`-mask='strongp?l?l?l?l?l?l?l'` finds another one: `strongpvlxjaor`. So it wasn't just poor luck - they are in fact quite frequent.

solardiz:
> `-mask='strongp?l?l?l?l?l?l?l'` finds another one: `strongpvlxjaor`. So it wasn't just poor luck - they are in fact quite frequent.

There's more to this. This one is repeatedly "crackable" when running 8 threads, but not when running 4. Suggests we're processing uninitialized (stale) data somewhere.

solardiz:
> we're processing uninitialized (stale) data somewhere.

On one hand, we're indeed processing uninitialized (stale) data when we do partial decryption, yet let `asn1_get_next` potentially scan the full buffer. We may want to deal with that to speed things up and reduce false positives.

On the other hand, for `strongpvlxjaor` this only affects `.hdr.length` after the last `asn1_get_next`, which we don't use, so e.g. zeroizing the buffer prior to decryption or zeroizing the portion from `SAFETY_FACTOR` after `check_pkcs_pad` or doing full decryption don't affect this false positive crack. Yet thread count and mask variations do.

So we must be processing uninitialized (stale) data somewhere else as well. I haven't yet found where.

ghost:
I have a doubt:
- Does the "hash" produced by the script is still also usable with hashcat? I mean, we recently complained about a new code.

Apart from that, I didn't find anything bad.

solardiz:
> * Does the "hash" produced by the script is still also usable with hashcat?

For the types they support, yes, they should be usable. For the types they don't, of course not, but we're not making things worse. It could be good to inform them of us allocating type 7, so they don't allocate it differently.

solardiz:
hashcat suffers from false positives, but apparently only for short (EC) keys, not for long (RSA/DSA) keys. This is consistent with the kind of ASN.1 checks I see in its code.

```
./hashcat -d 1,2,3 -O -a 3 -m 22951 pw-ssh5 '?a?a?a?a?l?l?l?l'
./hashcat -d 1,2,3 -O -a 3 -m 22931 pw-ssh3 '?a?a?a?a?a?a?a?a'
./hashcat -d 1,2,3 -O -a 3 -m 22931 pw-ssh '?l?lrongp?l?l?l?l?l?l?l'
cat hashcat.potfile
```

```
$sshng$5$16$8f1f99cc6f27c8b23478a3de6fcec7d7$224$7b10854dd55b7a38408952d479eab4833d03e61646bb0aadb418f95891862def3869c0443058267fbab4378286ca3bbe30d85cda12934095098dc106443578006226760a615f955d13e742f06da762d682982a5422bff5d4bcdc321d269f52937d3077db8313f70fd66238fe90d165f5fd53f839554274b81e3b3d2d3bb0ee0ad92402f794e0cb20aa4ee3ba2aaa26150b0847786dd4f20421eade809a068886769538de7fc8bbb73fdab2af4f26cf0f0e6ab7074c50defbc70114ca4c1b3ee83e4fa72fa2f3c75d12031179d1f333c2d9e0d270121313af6bd2423bb814d873:`dsufbou
$sshng$3$16$00b535fba963402f20c12648a59d7258$128$dfa09369ff38f33c9789d33760d16fdd47730311b41b51a0c7b1dd1dec850c5c2ff523710af12839f25a709f0076cdd3e3643fab2ea1d17c6fae52a797b55e752b71a1fdd46d5bd889b51ddc2a01922340e5be914a67dabf666aff1c88275bd8ec3529e26386279adeb480446ab869dc27c160bd8fe469d5f993b90aaffef8ce:UERm@lll
$sshng$1$16$570f498f6ff732775ee38648130f600d$1200$1777f12047d4ebab06d052d52946e5e0e73b41d5077b20e1ffe1c97ef9459b8c6844fecc24fdf63314c8889398fa140026339c85336278600e299c0f4c236648ca684f0c122e66d3e860e19eab8b46a564eb101def1c6a38f2f1800040c6b59a66e7b86e145e180f8a126e46544be1e17dd32e4e72f735c9e6b0ca4bbbb32ccf34ba0a7827858b0be32f9e53f13466e2ac78c3fecdf2a51cd7871286a3a91f9c71ae9e857a74bcc06071af6f60d827f7e13ccf6c1be722246c0796f509744c2b1b1452315ea6f86a1c8765d1f0c1d795349b4ea1ba229318b392fe505292cd0c6b4e3e9b2acc13b96943d92fa5635e05b7795989906274b0fb1894102d07facdd8f2122299960e1490823d62bbd5bf6d6c92ed26e68cc2edc93fbffec557a5d187fffe085ded9408ac63293851a684ca10d6e9a4ee9b5c552c827caee1f1c41870fe2d0e79bc4a0b85478fa82a58f947d345122c8ac7c80ba2ae8452b093dda70e2a4329fce70af9cf98e19477a622083664d1e62393a01b20371fc5be9390059f1c4af75d5448a2fbe1aaa46701c696afec927c67d15c046036531d9252faa08bbf9ea0e019ea574e6af94edd7ec17c83c0f87e34c7456e19bc53b2de04dafa83267694c1f61d038e0fc5f8f1b8ce573da470e6db6d38c0e8f7141ad9e9609ea408e3823271e987766039d484bc88f23f2f2a1175636ece950c7d82f43726287fef37da945ec6ad6adc04cb59f66087f68a3e84e8cc39c578bcbce3aaf67f1325d3d20dbd5872cc88ab72fc0bda05bf969eca08f8cafb306424a1597ba5d612e155b4723c2c1bee9a8e3d195be3b798ea417008a2340a919e23ac899ea4dbc4ef05af2cf6b12293eeb293584b37d3f8465e36a62d65b21f68725603e11dc14acf4e3855e25980387a34a34919fdd49844ed888e37199bb26df1bbbc303e895615fcbb0aa9ddc8a2aa685da942a1e68dc3a355d27236f74220d404d25e0ac64ae9203bb04296b4d67481a4f516fd22e47092073c9c44fa098670d736c5c509e55d6b40d3bf346ea5bb0007e32e9d8290c2633621fd84c2f5f428a5649ff3a16d00fec21381543202f2ee12078ddea8a371935f2ffa15aafa644e111a29c1c4703bf8e9cf1397356e296c5484558b96639b9cf3703aabff0cf42864dab91b1e09c6439159bc95374da7a5d416402286390e76cb766cd94e7a002596e8862b8d7e46c1fc6f7bdd0b93c73d2dc3cf58ea31bc549086209f450bb7460d5e9ba0d0f7b80337651f45bf83bef1783c3a15631c82428bfe167dc0692402d7f15144fff01ad8596970439ce8a2df0107c85a23ef93edd19f62de499ab58ada581886494c3e52dd5ec53c191f6d62729729a252c2c0d8024950d1637cfd7c61a4fe64ce41cde76fe00fa2607af66a44d3b4b8836820f40c03669f08b4e986f4d03c09e3c026a910f83be623d7f68ff80d81662f020f433f7a896e10134a278cd9a8517d3bcd77c5287f7d41bc52d2f8db79b5f8f9ed6d6f45a482b13cb91ecdef43ebe38f5ad71836185ae6faf1dd11c50cc1759e4834fcab2b3523d4224a32d2eaba224a2c950dac7524afc74f02f17b511f3b22d577a6928d40909bed64f6ed27096dff591a8fbee3f32733fd2b36c0c4708a5224f165af000d93832e211ae52465f680e7a4fd66bb5eb210c4402eb58f6ebfde:strongpassword
```

This is two false positives and one correct result.

solardiz:
OK, by my count and tests (with some checks ignored for debugging), the code currently pushed to this PR checks ~7 bytes for long keys, ~6 bytes for short keys. Edit: except for bcrypt-pbkdf keys, where we still check only 4 bytes, but this is partially compensated for by the slowness of bcrypt-pbkdf.

It also implements some "early accepts" as soon as the combined checks of padding and ASN.1 reach or exceed the same ~7 bytes threshold. This should avoid most false negatives in case our ASN.1 checks turn out to be too strict for future key files. However, maybe it'd be better to run the checks to completion and print a warning asking to contact us and report the issue, so that we'd fix the code to also avoid false negatives for unlucky keys of future types like this? Unfortunately, implementing that would complicate the code.

solardiz:
For bcrypt-pbkdf keys (unsupported in hashcat?), where we still check only 4 bytes, there's also room for improvement. For our 4 test vectors, the decrypted data starts with:

```
c38767f8 c38767f8 00000013 65636473
e2e9392d e2e9392d 0000000b 7373682d
3eb37c71 3eb37c71 00000013 65636472
c785799f c785799f 0000000b 73736832
```

Per OpenSSH's `PROTOCOL.key`, the two 32-bit `checkint` fields are immediately followed by `privatekey1`, "where each private key is encoded using the same rules as used for SSH agent." So we may need to find those rules. The above looks like it's length field followed by ASCII key type id (`ssh2` or `ecdr`), but it's better to know for sure per spec.

solardiz:
I checked related hashcat GitHub issues. Many people complaining about hashcat not supporting our type 6, with samples posted. Also in a linked hashcat forum thread. So I now think we may want to handle the clash differently - rather than renumber, we can distinguish their single-DES 6 vs. our AES 6 by salt length 8 vs. 16. That way, we'll avoid any breakage for anyone.

solardiz:
> handle the clash differently - rather than renumber, we can distinguish their single-DES 6 vs. our AES 6 by salt length 8 vs. 16.

And by lack of extra two fields. This is now implemented. We internally renumber their 6 to -1 in `get_salt`, but this isn't user-visible - it's just for convenience of further handling in the code.

magnumripper (approved):
I skimmed the changes, for what that's worth. More importantly I have many assorted private samples and archived personal keys that I can test this with in a few days.

solardiz:
> I have many assorted private samples and archived personal keys that I can test this with in a few days.

That would be great. I haven't tested the single-DES extraction at all - I only have the hashcat "hash" string for it, but no key file of this type. I still have no idea what they used to obtain that string - from what atom wrote in a hashcat issue comment, he thought our ssh2john produced such strings before, but I could not find any traces of that.

solardiz:
> It also implements some "early accepts" as soon as the combined checks of padding and ASN.1 reach or exceed the same ~7 bytes threshold. This should avoid most false negatives in case our ASN.1 checks turn out to be too strict for future key files. However, maybe it'd be better to run the checks to completion and print a warning asking to contact us and report the issue, so that we'd fix the code to also avoid false negatives for unlucky keys of future types like this? Unfortunately, implementing that would complicate the code.

I'm also considering an approach opposite to the above (and to its further enhancement mentioned above): decrypt only the first AES block first, not the padding. The ASN.1 BER checks of first AES block (but two DES blocks) let us reject based on ~5 to ~6 bytes, whereas the padding check rejects based on ~1 byte. This difference is insignificant for scalar AES code on CPU because even with ~1 byte we proceed to decrypt the other block only ~1/255 of the time, which is just good enough considering it's only a slight slowdown in those cases. However, on a GPU (when we port these changes over to the OpenCL format) we'd typically have a group of 64 work-items proceed to the second decryption if _any_ of them passed the padding check, which per my math would happen ~22.2% of the time. Also, extra AES block decryption on a GPU is relatively more work than on a CPU with AES-NI.

magnumripper:
> I have many assorted private samples and archived personal keys that I can test this with in a few days.

I had fewer of those than I thought, and none were of any new/older and/or interesting kind. The output from ssh2john did not change, and they cracked just the same both before and after these changes.

I also tried digesting man pages for creating custom sample keys but ended up no wiser.

magnumripper:
> > However, maybe it'd be better to run the checks to completion and print a warning asking to contact us and report the issue, so that we'd fix the code to also avoid false negatives for unlucky keys of future types like this?

I'd vote for such warning. Would it possibly be of any benefit to run such "extra" checks in eg. `cmp_one()`? Perhaps not considering all different types we handle here.

> on a GPU we'd typically have a group of 64 work-items proceed to the second decryption if _any_ of them passed the padding check, which per my math would happen ~22.2% of the time. Also, extra AES block decryption on a GPU is relatively more work than on a CPU with AES-NI.

Good thinking. We might also want to try opting in the bitsliced AES, which does the second block for free.

solardiz:
> > > However, maybe it'd be better to run the checks to completion and print a warning asking to contact us and report the issue, so that we'd fix the code to also avoid false negatives for unlucky keys of future types like this?
> 
> I'd vote for such warning.

OK, but this PR is taking too long to complete already, so I don't intend to get this in here. Also, our immediate task after merging this one should be to update the OpenCL format as well. Then we can revisit.

> Would it possibly be of any benefit to run such "extra" checks in eg. `cmp_one()`?

No. We'd need logic to reach `cmp_one` in those cases first, which means we had already checked the condition.

solardiz:
> > `-mask='strongp?l?l?l?l?l?l?l'` finds another one: `strongpvlxjaor`. So it wasn't just poor luck - they are in fact quite frequent.
> 
> There's more to this. This one is repeatedly "crackable" when running 8 threads, but not when running 4. Suggests we're processing uninitialized (stale) data somewhere.

It bothers me a bit that we never figured out why this false positive was unstable. Of course, we're not getting the false positive with the current set of checks. When I insert a `return 0;` after the first check in our current `check_structure_asn1`, I am able to get this false positive again (by directly specifying it as "mask"), and it's stable across different thread counts on the same system where I had previously observed it (as unstable back then). So hopefully whatever stability bug there was is now gone.

Tested with:

```shell
for N in `seq 1 9`; do OMP_NUM_THREADS=$N ./john pw-ssh -mask='strongpvlxja?l?l'; rm john.pot john.log john.rec; done
```

All of these crack `strongpvlxjaor` (with the code hack I mentioned).

PR #5746, opened 2025-03-30
fix: ssh-opencl must reject EC_KEY_WITH_AES256 hashes

ghost:
Heuristics deserve more attention. Probably inside the 2john tool.


ghost:
The change in `ssh_valid()` is still required.

Issue #5456, opened 2024-03-31
NVIDIA driver 551.86 new "Warning: Function [...] is a kernel, so overriding noinline attribute"


illsk1lls:
Getting an error displaying on screen on the latest Win11 with jumbo when working on 7z, zip (tested)

Win11 - 10.0.22631.3296

Error text: `Build log: (): Warning: Function zip is a kernel, so overriding noinline attribute. The function may be inlined when called.
(): Warning: Function zip_final is a kernel, so overriding noinline attribute. The function may be inlined when called.`

Example:
```Device 1: NVIDIA GeForce RTX 2070
Using default input encoding: UTF-8
Loaded 1 password hash (ZIP-opencl, WinZip [PBKDF2-SHA1 OpenCL])
Cost 1 (HMAC size) is 4 for all loaded hashes
Build log: (): Warning: Function zip is a kernel, so overriding noinline attribute. The function may be inlined when called.
(): Warning: Function zip_final is a kernel, so overriding noinline attribute. The function may be inlined when called.


Build log: (): Warning: Function zip is a kernel, so overriding noinline attribute. The function may be inlined when called.
(): Warning: Function zip_final is a kernel, so overriding noinline attribute. The function may be inlined when called.


LWS=32 GWS=73728 (2304 blocks)
Press 'q' or Ctrl-C to abort, 'h' for help, almost any other key for status
Enabling duplicate candidate password suppressor
testpassword#    (winzip-AES-256-testpassword#.zip/test.txt)
1g 0:00:01:03 DONE (2024-03-30 23:16) 0.01568g/s 1677Kp/s 1677Kc/s 1677KC/s 129456#..villena#
Use the "--show" option to display all of the cracked passwords reliably
Session completed.
```

Since the output for john.exe is using stderr instead of stdout for job status info, this information cannot be suppressed.

```
Version: 1.9.0-jumbo-1+bleeding-76cc11476a 2024-03-20 18:57:49 +0100
Build: cygwin 64-bit x86_64 AVX2 AC OMP OPENCL
SIMD: AVX2, interleaving: MD4:3 MD5:3 SHA1:1 SHA256:1 SHA512:1
CPU tests: AVX2
CPU fallback binary: john-avx-omp
OMP fallback binary: john-avx2
$JOHN is
Format interface version: 14
Max. number of reported tunable costs: 4
Rec file version: REC4
Charset file version: CHR3
CHARSET_MIN: 1 (0x01)
[...]

PR #5749, opened 2025-04-04
o5logon-opencl: Support Oracle 12

### LINKED ISSUE #5648: Update o5logon-opencl to support Oracle 12 (the CPU format does already)
This is trivial but we should merge #5643 first.

We should also move common stuff to `o5logon_common_plug.c` and `o5logon_common.h` like most other formats does. The CPU format currently lacks test vectors for Oracle 12. Here are two (from john-samples):
```c
	{"$o5logon$475FEF5BD6E8BCD9972935CE0A857B0B928438B0E4662588384F164801E59373490674B58039AEA2FDBB814C2496EA35*3D14D54520BC9E6511F4*BD0B6E2FD58320CA4B0516A21EF2AF70B61E5ACFC3EE4D5D6E5C492A7932D09C*80A1EADECBD79D1F70221AD8081FD201B92699151735499F42AD36A028A6A836D46AA81F0E27AC7E68CC66619F1F4D56", "openwall"},
	{"$o5logon$8D032D2C3AAB74E34ADABE9D4E62A3B19029DB37054665A004435DAF0B6C571A16DFE69C232923A1685414307DC01D88*681D0B32B396420D5914*EA974D3A9D7ECDC7316E03772847A9960E7DE5D86D416781072A36A6B7DBB430*0B8297E48756AA562F369651C18B73A5FEAA89151787298A4116E66090B6B72535154AE28B5D29DFA8740706D74FC49A", "password"},
```
We should also add one with length 17 or longer as well, and ideally one of eg. length 32 (the current maximum - perhaps an actual max. set by Oracle?).

magnumripper:
Also implements internal mask, for a 3-4x boost (due to "packed" key buffer it was pretty good already).
    
Closes #5648

<s>The "Mute stupid warnings" commit (#5456) will be dropped before merging (I could make a separate PR after adding a conf option for it). It strips the idiot warnings while retaining any relevant warnings (mutes the warning entirely if there were no relevant ones).</s>

<s>As of now, there's a bug in here. I have stared at the code for days, can't find it. Things behave weird - segfaults and such, on any device (including super's AMD that brings down the entire host). Debugging sometimes show super weird stuff that I can't draw any conclusions from.

Switching to the alternative (bitsliced) AES code sometimes move the problem around, or nearly hides it.

Running at LWS=1 sometimes helps a little or a lot. Everything points to a buffer overflow, except I can't find any, ASan can't either, and under ASan there's no nvidia device to be seen. A backtrace nearly always points to opaque OpenCL runtime stuff only.</s>

The o5logon stuff should ideally have been two commits, one for separating shared code and one for internal mask. But it's all too tangled to be fixed in retrospect.

### DIFF
--- a/run/john.conf
+++ b/run/john.conf
@@ -423,6 +423,17 @@ SleepOnTemperature = 1
 AvoidBusyWait = Y
 
 [Options:OpenCL]
+# Mute buggy nvidia warnings about kernel overriding noinline
+# attribute. Even with this set, they will show at "debug verbosity"
+# as in --verbose:6.
+MuteBogusWarnings = Y
+
+# Add ptxas info (-cl-nv-verbose) to build options for nvidia
+NvidiaShowPtxas = Y
+
+# Show runtime build warnings regardless of verbosity.
+AlwaysShowBuildWarnings = N
+
 # Set default OpenCL device(s). Command line option will override this.
 # If not set, we will search for a GPU or fall-back to the most
 # powerful device.  Syntax is same as --device option.
--- a/run/opencl/o5logon_kernel.cl
+++ b/run/opencl/o5logon_kernel.cl
@@ -1,89 +1,213 @@
 /*
-   This code was largely inspired by
-   pyrit opencl kernel sha1 routines, royger's sha1 sample,
-   and md5_opencl_kernel.cl inside jtr.
-   Copyright 2011 by Samuele Giovanni Tonon
-   samu at linuxasylum dot net
-   and Copyright (c) 2012-2017, magnum
-   This program comes with ABSOLUTELY NO WARRANTY; express or
-   implied .
-   This is free software, and you are welcome to redistribute it
-   under certain conditions; as expressed here
-   http://www.gnu.org/licenses/gpl-2.0.html
-*/
-
-/*
- * Modifications (c) 2014 Harrison Neal.
- * Licensed GPLv2
+ * Copyright (c) 2012-2025, magnum
+ * Copyright (c) 2014 Harrison Neal
+ * Copyright (c) 2011 by Samuele Giovanni Tonon, samu at linuxasylum dot net
+ *
+ * This program comes with ABSOLUTELY NO WARRANTY; express or
+ * implied .
+ * This is free software, and you are welcome to redistribute it
+ * under certain conditions; as expressed here
+ * http://www.gnu.org/licenses/gpl-2.0.html
  */
 
 #include "opencl_device_info.h"
 #include "opencl_misc.h"
 #include "opencl_sha1.h"
-#define AES_SRC_TYPE __constant
+#include "opencl_md5_ctx.h"
+#include "opencl_mask.h"
+
+// Workaround for UHD Graphics 630 version "1.2(Jan 10 2025 21:16:54)"
+#if (__OS_X__ && gpu_intel(DEVICE_INFO))
+#define AES_BITSLICE 1
+#endif
 #include "opencl_aes.h"
 
 typedef struct {
-	uint salt[3]; // ((SALT_LENGTH + 3)/4)
-	uchar ct[48]; // CIPHERTEXT_LENGTH
-} salt_t;
+	uint pw_len;                          /* AUTH_PASSWORD length (blocks) */
+	uint salt[(SALT_LENGTH + 1 + 3) / 4]; /* AUTH_VFR_DATA */
+	uchar ct[CIPHERTEXT_LENGTH];          /* Server's AUTH_SESSKEY */
+	uchar csk[CIPHERTEXT_LENGTH];         /* Client's AUTH_SESSKEY */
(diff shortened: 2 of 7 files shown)


magnumripper:
> The "Mute stupid warnings" commit (#5456) will be dropped before merging (I could make a separate PR after adding a conf option for it). It strips the idiot warnings while retaining any relevant warnings (mutes the warning entirely if there were no relevant ones).

I added conf options. IMO it can now be kept in this PR.
```
# Mute buggy nvidia warnings about kernel overriding noinline
# attribute. Even with this set, they will show at "debug verbosity"
# as in --verbose:6.
MuteBogusNvidiaWarnings = Y
```

Also added these in same commit:
```
# Show runtime build warnings regardless of verbosity.
AlwaysShowBuildWarnings = N

# Add ptxas info (-cl-nv-verbose) for nvidia to build log
NvidiaShowPtxas = Y
```
Perhaps the ptax info should default to `N` - as developers we can enable it in john-local.conf.

magnumripper:
Pro-tip: For POCL, you get much better crash info after adding `-g` to conf's `GlobalBuildOpts`.

magnumripper:
> As of now, there's a bug in here

I think I squashed it. Need to test more, and also check if other formats with that keybuffer packing method have the same problem, never surfaced.

magnumripper:
> Need to test more, and also check if other formats with that keybuffer packing method have the same problem, never surfaced.

It seems fine now. Other formats may or may not have been affected (it should often be triggered in self tests, which apparently didn't happen) but I added safety checks in #5750 anyway.

magnumripper:
> Pro-tip: For POCL, you get much better crash info after adding `-g` to conf's `GlobalBuildOpts`.

This breaks build on some other platforms though, such as nvidia, so need to be used selectively.

ghost:
> > Pro-tip: For POCL, you get much better crash info after adding `-g` to conf's `GlobalBuildOpts`.
> 
> This breaks build on some other platforms though, such as nvidia, so need to be used selectively.

So I think we can add it automatically inside `get_build_opts()` if the verbosity is > XX (only if pocl and ...).

```diff
diff --git a/src/opencl_common.c b/src/opencl_common.c
index 87dea222b..004a64f7a 100644
--- a/src/opencl_common.c
+++ b/src/opencl_common.c
@@ -1136,8 +1136,11 @@ static char *get_build_opts(int sequential_id, const char *opts)
                        global_opts = OPENCLBUILDOPTIONS;
 
        snprintf(build_opts, LINE_BUFFER_SIZE,
-                "-I opencl %s %s%s%s%s%s%d %s%d %s -D_OPENCL_COMPILER %s",
+                "-I opencl %s %s%s%s%s%s%s%d %s%d %s -D_OPENCL_COMPILER %s",
                global_opts,
+                       options.verbosity >= VERB_LEGACY &&
+                       get_platform_vendor_id(get_platform_id(sequential_id)) ==
+                        PLATFORM_POCL ? "-g " : "",
                get_device_version(sequential_id) >= 200 ? "-cl-std=CL1.2 " : "",
 #ifdef __APPLE__
                "-D__OS_X__ ",
```
[0001-OpenCL-add-g-for-POCL-to-get-better-crach-info.PATCH](https://github.com/user-attachments/files/19608857/0001-OpenCL-add-g-for-POCL-to-get-better-crach-info.PATCH)

magnumripper:
> So I think we can add it automatically inside `get_build_opts()` if the verbosity is > XX (only if pocl and ...).

True, like for -v:6 (debug) and above. Theoretically we could also auto-detect support for it, autoconf style 🤣

magnumripper:
It now fails self-test on Super's Vega, but at least it doesn't bring the entire machine down now 👀 
Switched to the alternative AES, no difference. Running with `-lws=1 -gws=1` (for clues) did not help.

The format worked fine before this PR. would have been nicer the other way around. Had 250927K c/s before, now (running with `-skip`) 330752K.

Tried several other little tricks to no avail. Giving up on that.

solardiz (approved):
I skimmed and saw nothing obviously wrong, but then there are some things here that I'm not familiar with for a proper review.

Regarding `MuteBogusNvidiaWarnings`, maybe rename it to `MuteBogusWarnings`, so that we may later use this same setting for other bogus OpenCL warnings? This rename would also justify putting it in the OpenCL rather than GPU section.

solardiz:
@hantwister This PR is heavily reworking code some of which you had kindly contributed. Looking at your commit from 2014, I think you chose GPLv2 because you reused pieces of code already under that license. We've since (re)moved those. Would you be OK with us re-licensing the remaining pieces of your code (in the OpenCL kernel here) under our preferred cut-down BSD license? (Not yet done in this PR, but is something I suggest to do.) Thank you!

magnumripper:
> Regarding `MuteBogusNvidiaWarnings`, maybe rename it to `MuteBogusWarnings`, so that we may later use this same setting for other bogus OpenCL warnings?

OK. I'll keep the comment about nvidia in john.conf for now though, we can change that if/when needed.

PR #5750, opened 2025-04-04
Some OpenCL packed-buffer formats: Add safety for count < GWS

magnumripper:
The problem, if any, would occur during self-tests, single mode or at the last batch of a session, with stale data past currently used part of buffer. Actual problem seen with upcoming revised o5logon-opencl.

Can be benign if triggered but it may lead to a kernel crash.

### DIFF
--- a/src/opencl_cryptosafe_fmt_plug.c
+++ b/src/opencl_cryptosafe_fmt_plug.c
@@ -346,6 +346,10 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 		if (idx_offset > 4 * (gws + 1))
 			idx_offset = 0;
 
+		/* Safety for when count < GWS */
+		for (int i = count; i <= gws; i++)
+			saved_idx[i] = key_idx;
+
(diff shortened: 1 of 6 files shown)


magnumripper:
Oh BTW it's not actually about "count < GWS", but "count < LWS multiple", due to an optimization we have in all formats (afaik) where GWS is temporary lowered to such multiple when applicable.

solardiz:
> Oh BTW it's not actually about "count < GWS", but "count < LWS multiple",

Right.

> due to an optimization we have in all formats (afaik) where GWS is temporary lowered to such multiple when applicable.

You mean temporarily increased (not lowered) to next LWS multiple, and not as an optimization but to satisfy the constraint? At least that's what I recall.

solardiz (approved):

magnumripper:
> > due to an optimization we have in all formats (afaik) where GWS is temporary lowered to such multiple when applicable.
> 
> You mean temporarily increased (not lowered) to next LWS multiple

No I mean lowered/decreased. If GWS is 10240, LWS is 32 and count is 23, the GWS will be temporarily decreased to 32. One could say that `count` is effectively increased to 32 though.

>  and not as an optimization but to satisfy the constraint? At least that's what I recall.

As it's decreased (with my definition) it's an optimization: In our early implementations, the GWS stayed at the higher number, so execution could be longer for no good.

Edit: At least that is what _I_ recall, your recollection of it may be more accurate 😆 or perhaps it's some mix of the two versions.

PR #5752, opened 2025-04-04
Tweak the "excessive partial hash collisions" warnings

### LINKED ISSUE #5736: Drop the "Warning: excessive partial hash collisions detected" for salt-only and blob-binary formats
Due to how FMT_BLOB formats (office for one) are implemented, they can't have `binary_hash` functions. Either mute that warning for them (very easy) or fix the problem (very hard, I think. Can't remember the details but I think I've tried and gave up).

magnumripper:
For salt-only formats, bump the warning threshold to <s>10x `LDR_HASH_COLLISIONS_MAX`.</s> `LDR_HASH_COLLISIONS_SALT_ONLY` which is currently 10x `LDR_HASH_COLLISIONS_MAX` in params.h.

Add a self-test for detecting misuse of default binary hash functions with FMT_BLOB formats.

Fix such bug in Office formats.

Closes #5736

### DIFF
--- a/src/formats.c
+++ b/src/formats.c
@@ -1,6 +1,8 @@
 /*
  * This file is part of John the Ripper password cracker,
  * Copyright (c) 1996-2001,2006,2008,2010-2013,2015 by Solar Designer
+ * Copyright (c) 2011-2025, magnum
+ * Copyright (c) 2009-2018, JimF
  */
 
 #if AC_BUILT
@@ -902,6 +904,30 @@ static char *fmt_self_test_body(struct fmt_main *format,
 			    format->methods.get_hash[size] !=
 			    fmt_default_get_hash)
 				return "get_hash method not allowed for FMT_HUGE_INPUT";
+		}
+	}
+
+	if (format->params.flags & FMT_BLOB) {
+		/*
+		 * BLOB formats can't use the default binary_hash_[0-6] functions.
+		 */
+		if (format->methods.binary_hash[0] == fmt_default_binary_hash_0)
+			return "default binary_hash_[0-6] method not allowed for FMT_BLOB";
+	}
+
+	if (format->params.binary_size == 0) {
+		for (size = 0; size < PASSWORD_HASH_SIZES; size++) {
+			/*
+			 * Salt-only formats can't have binary hash functions.
+			 */
+			if (format->methods.binary_hash[size] &&
+			    format->methods.binary_hash[size] !=
+			    fmt_default_binary_hash)
+				return "binary_hash method not allowed for salt-only formats";
+			if (format->methods.get_hash[size] &&
+			    format->methods.get_hash[size] !=
+			    fmt_default_get_hash)
+				return "get_hash method not allowed for salt-only formats";
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


src/formats.c
```diff
@@ -905,6 +907,32 @@ static char *fmt_self_test_body(struct fmt_main *format,
 		}
 	}
 
+	if (format->params.flags & FMT_BLOB) {
+		/*
+		 * BLOB formats can't use the default binary_hash_[0-6] functions.
+		 */
+		if (format->methods.binary_hash[0] &&
+		    format->methods.binary_hash[0] ==
+		    fmt_default_binary_hash_0)
```
solardiz: No need for the `format->methods.binary_hash[0] &&` pre-check if you then check for equality to a specific non-NULL address.

src/loader.c
```diff
@@ -996,6 +996,11 @@ static void ldr_load_pw_line(struct db_main *db, char *line)
 
 		if (!(db->options->flags & DB_WORDS) && dupe_checking) {
 			int collisions = 0;
+			int hash_collisions_max = LDR_HASH_COLLISIONS_MAX;
+
+			if (!format->params.binary_size)
+				hash_collisions_max *= 10;
```
solardiz: I think it'd be cleaner to have this x10 multiplier or alternative max value in `params.h` near `LDR_HASH_COLLISIONS_MAX`. Also, why was 1000 not sufficient but 10000 expected to be sufficient? 1000 was kind of arbitrary, so if only a 10x increase is needed we can simply bump it to 10000.

src/office_common_plug.c
```diff
@@ -145,3 +147,87 @@ unsigned int ms_office_common_version(void *salt)
...
+{
+	fmt_data *blob = binary;
+	ms_office_binary_blob *verifiers = blob->blob;
+	uint8_t *encryptedVerifier = verifiers->encryptedVerifier;
+	uint32_t hash;
+
+	memcpy(&hash, encryptedVerifier, sizeof(uint32_t));
+
+	return hash & PH_MASK_0;
+}
```
solardiz, line 161: Maybe avoid code duplication by defining/using a macro or a `static inline` function.

solardiz (approved):
I suggested some changes, but I'm also fine merging this as-is.

src/loader.c
magnumripper: > I think it'd be cleaner to have this x10 multiplier or alternative max value in `params.h` near `LDR_HASH_COLLISIONS_MAX`. Also, why was 1000 not sufficient but 10000 expected to be sufficient? 1000 was kind of arbitrary, so if only a 10x increase is needed we can simply bump it to 10000.

Perhaps just muting them with no limit would be more logical? The formats can't be fixed anyway.

If not, I agree just bumping LDR_HASH_COLLISIONS_MAX for now is better bc KISS.

src/formats.c
magnumripper: > No need for the `format->methods.binary_hash[0] &&` pre-check

That was from a copy/paste, I'll fix the originals as well. (EDIT: No, the others had `!=` so not applicable)

src/loader.c
magnumripper: > Perhaps just muting them with no limit would be more logical? The formats can't be fixed anyway.

Maybe a problem with that is the user won't get an explanation about load time (which I presume increases non linearly).

PR #5753, opened 2025-04-04
Random improvements in OpenCL(?)

ghost:
A PR to discuss small tweaks to OpenCL.

### DIFF
--- a/src/opencl_autotune.c
+++ b/src/opencl_autotune.c
@@ -12,6 +12,9 @@
 #ifdef HAVE_OPENCL
 
 #include "opencl_common.h"
+#ifdef _OPENMP
+#include <omp.h>
+#endif
 
 /* Allow the developer to select configurable step size for gws. */
 int autotune_get_next_gws_size(size_t num, int step, int startup,
@@ -51,13 +54,32 @@ size_t autotune_get_task_max_work_group_size(int use_local_memory,
                                            cl_kernel crypt_kernel)
 {
 
-	size_t max_available;
+	size_t max_available, compute_units;
 
 	if (use_local_memory)
 		max_available = get_local_memory_size(gpu_id) /
 			(local_memory_size);
 	else
 		max_available = get_device_max_lws(gpu_id);
+
+	/* Still allows the end user to specify a value for 'LWS=' (high or not)
+	   using the command line or a configuration file; so the local_work_size
+	   variable is (can be) initialized by the end user. */
+	if (cpu(device_info[gpu_id]) && !get_lws_set_by_user()) {
+		compute_units = get_max_compute_units(gpu_id);
+
+		if (compute_units <= 1)
+#ifdef _OPENMP
+			compute_units = omp_get_max_threads();
+#else
+			compute_units = 32;
+#endif
+		get_power_of_two(compute_units);
+
+		/* If we can restrict max_available we should do so */
+		if (max_available > (compute_units * 4))
+			max_available = compute_units * 4;
+	}
 
 	if (max_available > get_kernel_max_lws(gpu_id, crypt_kernel))
 		return get_kernel_max_lws(gpu_id, crypt_kernel);
--- a/src/opencl_common.c
+++ b/src/opencl_common.c
@@ -1136,8 +1136,11 @@ static char *get_build_opts(int sequential_id, const char *opts)
 			global_opts = OPENCLBUILDOPTIONS;
 
 	snprintf(build_opts, LINE_BUFFER_SIZE,
-	         "-I opencl %s %s%s%s%s%s%d %s%d %s -D_OPENCL_COMPILER %s",
+	         "-I opencl %s %s%s%s%s%s%s%d %s%d %s -D_OPENCL_COMPILER %s",
 	        global_opts,
+	        options.verbosity >= VERB_DEBUG &&
+	        get_platform_vendor_id(get_platform_id(sequential_id)) ==
+	         PLATFORM_POCL ? "-g " : "",
 	        get_device_version(sequential_id) >= 200 ? "-cl-std=CL1.2 " : "",
 #ifdef __APPLE__
 	        "-D__OS_X__ ",
--- a/src/opencl_common.c
+++ b/src/opencl_common.c
@@ -119,6 +119,7 @@ cl_kernel crypt_kernel;
 size_t local_work_size;
 size_t global_work_size;
 size_t max_group_size;
+int lws_set_by_user = 0;
 unsigned int ocl_v_width = 1;
 unsigned long long global_speed;
 
@@ -154,6 +155,11 @@ void opencl_process_event(void)
 			event_pending = (event_abort || event_poll_files || event_reload);
 		}
 	}
+}
+
+int get_lws_set_by_user()
+{
+	return (lws_set_by_user != 0);
 }
 
 int get_number_of_available_platforms()
@@ -815,6 +821,7 @@ void opencl_load_environment(void)
 		int i;
 		const char *cmdline_devices[MAX_GPU_DEVICES];
 
+		local_work_size = global_work_size = duration_time = 0;
 		nvidia_probe();
 		amd_probe();
 
@@ -1036,6 +1043,11 @@ void opencl_get_user_preferences(const char *format)
 		local_work_size = options.lws;
 	else if ((tmp_value = getenv("LWS")))
 		local_work_size = atoi(tmp_value);
+
+	if (local_work_size)
+		lws_set_by_user = 1;
+	else
+		lws_set_by_user = 0;
 
 	if (format && (tmp_value = (char*)cfg_get_param(SECTION_OPTIONS, SUBSECTION_OPENCL,
 			opencl_get_config_name(fmt_base_name, GWS_CONFIG_NAME))))
--- a/src/opencl_common.h
+++ b/src/opencl_common.h
@@ -139,6 +139,7 @@ extern cl_kernel crypt_kernel;
 extern size_t local_work_size;
 extern size_t global_work_size;
 extern size_t max_group_size;
+extern int lws_set_by_user;
 extern unsigned int ocl_v_width;
 extern unsigned long long global_speed;
 
@@ -166,6 +167,10 @@ void opencl_done(void);
  * affect the return value.
  */
 unsigned int opencl_get_vector_width(int sequential_id, int size);
+
+/* Returns true if the end user has specified a value for local_work_size
+ * using 'LWS=', '-lws', or the configuration file. */
+int get_lws_set_by_user(void);
 
 /* Returns number of selected devices */
 int get_number_of_devices_in_use(void);


src/opencl_common.c
```diff
@@ -1136,8 +1136,11 @@ static char *get_build_opts(int sequential_id, const char *opts)
 			global_opts = OPENCLBUILDOPTIONS;
 
 	snprintf(build_opts, LINE_BUFFER_SIZE,
-	         "-I opencl %s %s%s%s%s%s%d %s%d %s -D_OPENCL_COMPILER %s",
+	         "-I opencl %s %s%s%s%s%s%s%d %s%d %s -D_OPENCL_COMPILER %s",
 	        global_opts,
+	        options.verbosity > VERB_LEGACY &&
```
magnumripper: I think I've been trying to write things like this as `>= VERB_DEBUG` instead, which is the same thing but needs less work if we'd add a verbosity level between LEGACY and DEBUG (for example).

ghost: VERB_LEGACY is 4. Do you want VERB_DEBUG or VERB_MAX?
```C
#define VERB_DEBUG			6
#define VERB_MAX			5
#define VERB_LEGACY			4
```

src/opencl_autotune.c
```diff
@@ -59,6 +59,12 @@ size_t autotune_get_task_max_work_group_size(int use_local_memory,
 	else
 		max_available = get_device_max_lws(gpu_id);
 
+	if (cpu(device_info[gpu_id])) {
+#include <omp.h>
+		max_available = omp_get_max_threads();
+		get_power_of_two(max_available);
+	}
```
magnumripper: I guess we should try to research this. I'll google around a little.

ghost: We can add a multiplication, such as num_threads x 2 or num_threads x 4. However, nothing beyond this makes sense.

src/opencl_common.c
magnumripper: Oh, right. Well since `-g` is a debugging option, VERB_DEBUG is the logical one, so `>= VERB_DEBUG`.

It's peculiar that max is not the max. It's a little less max than debug o.O

...or max is highest but debug is highester. Right now it's highestest, even.

src/opencl_autotune.c
magnumripper: Super has 32 cores, and auto-tunes LWS to anywhere between 128..1024 with NT-opencl, ending up in more or less the same performance. It doesn't seem to want lower than 128 but if I peg it to 16 (so equal to only half the number of cores), it performs as good anyway.
Lower than 4 is really bad, 4 is OK and >=8 is full performance.

So no answers from that really.

magnumripper: My old 4-core/8-thread machine tuned to LWS=8 for wpapsk and 16 for nt. Those numbers are "logical" to me - one OpenCL thread per CPU thread, or two (somewhat like `OMP_SCALE 2`) for a fast format.

That's with POCL though, as opposed to Intel's runtime on super. They might be very differently designed.

magnumripper: > if I peg it to 16 (so equal to only half the number of cores), it performs as good anyway.

Perhaps that actually ends up as two groups of 16 running concurrently on 32 cores? Precise grouping is mostly about memory access and barriers, much more delicate on a GPU.

ghost: IMO: a multiplier (perhaps 4) combined with the fact that the autotune is still doing its job (up to the new max) should be enough to ensure that the patch is correct.

magnumripper: x4 sounds about right from what we know right now. But I'll check a couple machines out on Monday.

magnumripper: Also, if `get_max_compute_units()` returns anything _higher than 1_ (super's runtime stupidly say 1 even for 32 cores), that figure should be used and this OpenMP check can be skipped. But if we get 1 from that, the OpenMP check should be tried instead. It has other flaws, such as OpenMP not always available, or `OMP_NUM_THREADS` limited.

BTW the omp.h header include should be moved up to where it belongs, and must be guarded with `#ifdef _OPENMP` like elsewhere. Then this whole new (`else`) block of code should also be inside `#ifdef _OPENMP`.

If all else fails, we should set it to something sensible, I guess 16 at the very least.

magnumripper (changes requested):

src/opencl_autotune.c
magnumripper: Also, if `get_max_compute_units()` returns anything _higher than 1_ (super's runtime stupidly say 1 even for 32 cores), that figure should be used and this OpenMP check can be skipped. But if we get 1 from that, the OpenMP check should be tried instead. It has other flaws, such as OpenMP not always available, or `OMP_NUM_THREADS` limited.

BTW the omp.h header include should be moved up to where it belongs, and must be guarded with `#ifdef _OPENMP` like elsewhere. Then this whole new (`else`) block of code should also be inside `#ifdef _OPENMP`.

If all else fails, we should set it to something sensible, I guess 16 at the very least.

magnumripper (changes requested):

src/opencl_autotune.c
```diff
@@ -51,14 +54,30 @@ size_t autotune_get_task_max_work_group_size(int use_local_memory,
...
 {
 
-	size_t max_available;
+	size_t max_available, compute_units;
 
 	if (use_local_memory)
 		max_available = get_local_memory_size(gpu_id) /
 			(local_memory_size);
 	else
 		max_available = get_device_max_lws(gpu_id);
```
ghost, line 64: To allow the end user to specify a high(er) value for `LWS=x`.
```diff
diff --git a/src/opencl_autotune.c b/src/opencl_autotune.c
index 19c844822..00ab0fae2 100644
--- a/src/opencl_autotune.c
+++ b/src/opencl_autotune.c
@@ -62,7 +62,9 @@ size_t autotune_get_task_max_work_group_size(int use_local_memory,
        else
                max_available = get_device_max_lws(gpu_id);
 
-       if (cpu(device_info[gpu_id])) {
+       /* Still allows the end user to specify a hight value for 'LWS='
+          on the command line */
+       if (cpu(device_info[gpu_id]) && (local_work_size <= 8)) { // local_work_size is using the "default" value.
                compute_units = get_max_compute_units(gpu_id);
 
                if (compute_units <= 1)

```

magnumripper: Not sure I understand where 8 comes from. What does "the "default" value" mean here?

Any requested value of `LWS=` or `-lws=` (lower or higher) should override this new logic. If I say 1024, it should be `MIN(1024, get_kernel_max_lws(gpu_id, crypt_kernel)`, like our current code does it iirc. And if I say 1 it has to be 1.

claudioandre-br:
```diff
diff --git a/src/opencl_autotune.c b/src/opencl_autotune.c
index 91013f390..634a6ef84 100644
--- a/src/opencl_autotune.c
+++ b/src/opencl_autotune.c
@@ -65,7 +65,7 @@ size_t autotune_get_task_max_work_group_size(int use_local_memory,
	/* Still allows the end user to specify a value for 'LWS=' (high or not)
 	   using the command line or a configuration file; so the local_work_size
	   variable is (can be) initialized by the end user. */
-	if (cpu(device_info[gpu_id])) {
+	if (cpu(device_info[gpu_id]) && !get_lws_is_set()) {
 		compute_units = get_max_compute_units(gpu_id);
 
 		if (compute_units <= 1)
diff --git a/src/opencl_common.c b/src/opencl_common.c
index b66667cc4..9983e6062 100644
--- a/src/opencl_common.c
+++ b/src/opencl_common.c
@@ -119,6 +119,7 @@ cl_kernel crypt_kernel;
 size_t local_work_size;
 size_t global_work_size;
 size_t max_group_size;
+int lws_is_set = 0;
 unsigned int ocl_v_width = 1;
 unsigned long long global_speed;
 
@@ -156,6 +157,11 @@ void opencl_process_event(void)
 	}
 }
 
+int get_lws_is_set()
+{
+	return (lws_is_set != 0);
+}
+
 int get_number_of_available_platforms()
 {
 	int i = 0;
@@ -1037,6 +1043,9 @@ void opencl_get_user_preferences(const char *format)
 	else if ((tmp_value = getenv("LWS")))
 		local_work_size = atoi(tmp_value);
 
+	if (local_work_size)
+		lws_is_set = 1;
+
 	if (format && (tmp_value = (char*)cfg_get_param(SECTION_OPTIONS, SUBSECTION_OPENCL,
 			opencl_get_config_name(fmt_base_name, GWS_CONFIG_NAME))))
 		global_work_size = atoi(tmp_value);
diff --git a/src/opencl_common.h b/src/opencl_common.h
index ac0281576..190f51847 100644
--- a/src/opencl_common.h
+++ b/src/opencl_common.h
@@ -139,6 +139,7 @@ extern cl_kernel crypt_kernel;
 extern size_t local_work_size;
 extern size_t global_work_size;
 extern size_t max_group_size;
+extern int lws_is_set;
 extern unsigned int ocl_v_width;
 extern unsigned long long global_speed;
 
@@ -167,6 +168,10 @@ void opencl_done(void);
  */
 unsigned int opencl_get_vector_width(int sequential_id, int size);
 
+/* Returns true if the end user has specified a value for local_work_size using
+ * 'LWS=', '-lws', or the configuration file. */
+int get_lws_is_set(void);
+
 /* Returns number of selected devices */
 int get_number_of_devices_in_use(void);
 

```

magnumripper:
> super's runtime stupidly say 1 even for 32 cores

I have yet to find another driver (even intel) that doesn't report the number of CPU cores (actually threads). This weirdness on super's runtime also has the result that its "Speed Index" is guesstimated as 5200 instead of 166400 though that is mostly cosmetical.

Anyway, I think the scheme we have now with fallback to OpenMP (if available) is sensible.

magnumripper (approved):
I think this looks good now, although I didn't test it yet

magnumripper:
> super's runtime stupidly say 1 even for 32 cores

Oddly enough, in https://github.com/openwall/john/issues/3639 the same machine, driver and version did report 32 cores. I wonder what's the difference? The only thing I could think of is `OMP_NUM_THREADS` but that was not it (and it shouldn't).

ghost:
I've seen [1] happen more than once, but I can't link it to PR:
- Only on the first run does the “defect” occur;
- A second run goes as expected;
- maybe it happens because the computer is busy and the tune fails |(exceeds 200 ms)|, then everything that follows is compromised(?).

[1]
```bash
LWS=128 run/john --test --format=nt-opencl -v:5
initUnicode(UNICODE, RAW/RAW)
RAW -> RAW -> RAW
Benchmarking: NT-opencl [MD4 OpenCL/mask accel]... Loaded 44 hashes with 1 different salts to test db from test vectors
Build time: 7.363 ms
Internal mask not utilized (target: 224)
Calculating best GWS for LWS=128; max. 200 ms single kernel invocation.
Tuning for password length 7
key xfer: 862 ns, idx xfer: 131 ns, crypt: 321.294 ms* (exceeds 200 ms)
key xfer: 571 ns, idx xfer: 712 ns, crypt: 213.837 ms* (exceeds 200 ms)
LWS=64 GWS=128 (2 blocks) PASS, 
Test mask: ?a?a?l?u?d?d?s
Options used: -I opencl -cl-mad-enable -cl-std=CL1.2 -D__POCL__ -D__CPU__ -DDEVICE_INFO=9 -D__SIZEOF_HOST_SIZE_T__=8 -DDEV_VER_MAJOR=5 -DDEV_VER_MINOR=0 -D_OPENCL_COMPILER -D OFFSET_TABLE_SIZE=28 -D HASH_TABLE_SIZE=45 -DUCS_2 -D NUM_LOADED_HASHES=44 -D NUM_INT_KEYS=260 -D SELECT_CMP_STEPS=2 -D BITMAP_MASK=0x3ffffU -D USE_LOCAL_BITMAPS=0 -D IS_STATIC_GPU_MASK=1 -D CONST_CACHE_SIZE=524288 -DENC_RAW -DENC_RAW -DPLAINTEXT_LENGTH=27 -D LOC_0=2 -D LOC_1=4 -D LOC_2=-1 -D LOC_3=-1 $JOHN/opencl/nt_kernel.cl
Build time: 1.645 s
binary size 70466
Internal mask, multiplier: 260 (target: 224)
Calculating best GWS for LWS=128; max. 200 ms single kernel invocation.
key xfer: 772 ns, idx xfer: 371 ns, crypt: 259.998 ms* (exceeds 200 ms)
key xfer: 551 ns, idx xfer: 260 ns, crypt: 170.136 ms*, res xfer: 921 ns
gws:        64    97802 c/s       97802 rounds/s  170.138 ms per crypt_all()!!
key xfer: 842 ns, idx xfer: 561 ns, crypt: 167.153 ms*, res xfer: 16.461 us
gws:        32    49769 c/s       49769 rounds/s  167.171 ms per crypt_all()-
LWS=32 GWS=64 (2 blocks) x260 DONE
Raw:	13112K c/s real, 6919K c/s virtual
```

solardiz:
I wonder if LWS on CPU devices could also be affected by (auto-)vectorization, and so by SIMD width? If so, the max multiplier of 4 over OpenMP thread count could be too small - e.g., couldn't it need to be 16 for AVX-512?

magnumripper:
> I wonder if LWS on CPU devices could also be affected by (auto-)vectorization, and so by SIMD width? If so, the max multiplier of 4 over OpenMP thread count could be too small - e.g., couldn't it need to be 16 for AVX-512?

IIRC they do their best to make that transparent to host code (perhaps multiplying what we give them?) but we could make it 16 or even 32 just to be sure.

PR #5756, opened 2025-04-19
Added support for PostgreSQL SCRAM-SHA-256

ghost:
I have added an implementation for PostgreSQL SCRAM-SHA-256 authentication. It is based on the SCRAM-SHA-256 as specified in RFC 7677. Another implementation is added but it is for MongoDB. 
This PR introduces initial handling support for the PostgreSQL SCRAM-SHA-256$ password format, as used in the pg_authid catalog. The format looks like this:

`SCRAM-SHA-256$<iterations>:<salt(base64)>$<storedKey(base64)>:<serverKey(base64)>`

This is a step toward enabling John the Ripper to crack PostgreSQL SCRAM-based password hashes.
PostgreSQL 10+ uses SCRAM-SHA-256 for password storage by default. These hashes are becoming increasingly common, and currently, John the Ripper does not support them. This addition lays the groundwork for parsing and potentially cracking these hashes in future updates.

### DIFF
--- a/src/scram_postgres.c
+++ b/src/scram_postgres.c
@@ -1,0 +1,299 @@
+/*
+ * This file is part of John The Ripper
+ * Copyright (c) 2025 Pranjal Prasad <prasadpranjal213@gmail.com>
+ *
+ * PostgreSQL SCRAM-SHA-256 authentication implementation.
+ * Based on SCRAM-SHA-256 as specified in RFC 7677
+ *
+ * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


solardiz:
I don't know why you chose to open and close this within 12 minutes. That's weird.

Anyway, something like this, but not quite it, would be a welcome contribution. However, it should probably not be a new format, but rather an enhancement to the recently added implementation for MongoDB - unless you give a very good reason for it to be a separate format? Also, we'd prefer our cut-down BSD license.

PR #5757, opened 2025-04-24
Several Kerberos formats: Lowercase non-ASCII input

### LINKED ISSUE #5651: krb5tgs formats fails to lowercase non-ascii salt
The salt is the user name but lowercased. The current code only lowercases ascii. If given non-ascii hashes not already lowercased, this leads to false negatives.

Other krb5 formats possibly have the same problem.

magnumripper:
They would only lowercase ASCII until now, resulting in false negatives unless the tool that made the hash did it properly for us to start with.

This also includes modifications to shared enc_lc() and enc_uc() functions, getting rid of hardcoded length limits.

Closes #5651

### DIFF
--- a/src/krb5_asrep_common_plug.c
+++ b/src/krb5_asrep_common_plug.c
@@ -17,9 +17,6 @@
 char *krb5_asrep_split(char *ciphertext, int index, struct fmt_main *self)
 {
 	static char *ptr, *keeptr;
-	unsigned char etype = 0;
-	char *p = ciphertext;
-	int i;
 
 	if (strnlen(ciphertext, LINE_BUFFER_SIZE) < LINE_BUFFER_SIZE &&
 	    strstr(ciphertext, "$SOURCE_HASH$"))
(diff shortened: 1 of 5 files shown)


src/krb5_asrep_common_plug.c
```diff
@@ -47,15 +46,14 @@ char *krb5_asrep_split(char *ciphertext, int index, struct fmt_main *self)
 			p = strchr(ciphertext + FORMAT_TAG_LEN + ETYPE_TAG_LEN + 1, '$') + 1;
 			for (i = 0; i < p - ciphertext; i++)
 				ptr[i] = ARCH_INDEX(ciphertext[i]);
```
solardiz: The above loop (not touched by the changes here yet) looks like a weird way to do `memcpy` - so perhaps simplify it? You'd also remove the variable `i` entirely, and replace your two `+=` lines below with assignments of the correct pointer values.

solardiz (approved):
Not a real review as I'm unfamiliar with this encoding stuff.

PR #5759, opened 2025-04-25
SSH format: Fix bogus "warning: array subscript is above array bounds", optimize

### LINKED ISSUE #5758: SSH format build warning
I get this while building (on super only, w/o any `scl enable`) as of HEAD @88e3eeb92:
```
ssh_fmt_plug.c: In function 'common_crypt_code':
ssh_fmt_plug.c:127: warning: array subscript is above array bounds
```
The line number is bogus.

All kinds of self-tests passes just fine.

solardiz:
which was seen with RHEL6'ish gcc 4.4.7.

Amends 71bb5a74a718c8d9b20ceece373804152c39aca7
Fixes #5758

### DIFF
--- a/src/ssh_common_plug.c
+++ b/src/ssh_common_plug.c
@@ -127,7 +127,7 @@ void *ssh_get_salt(char *ciphertext)
 	if (cs.cipher == 2 || cs.cipher == 6) {
 		p = strtokm(NULL, "$");
 		if (!p && cs.cipher == 6 && cs.sl == 8) {
-			cs.cipher = -1;
+			cs.cipher = 7;
 		} else {
 			cs.rounds = atoi(p);
 			p = strtokm(NULL, "$");
@@ -144,7 +144,7 @@ unsigned int ssh_iteration_count(void *salt)
 	struct custom_salt *cur_salt = salt;
 
 	switch (cur_salt->cipher) {
-	case -1:
+	case 7:
 		return 1; // generate 8 bytes of key + DES
 	case 0:
 		return 2; // generate 24 bytes of key + 3DES
@@ -166,7 +166,7 @@ unsigned int ssh_kdf(void *salt)
 
 	switch (cur_salt->cipher) {
 	case 0:
-	case -1:
+	case 7:
 		return 1; // MD5 KDF + 3DES or DES
 	case 2:
 	case 6:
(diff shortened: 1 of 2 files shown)


solardiz:
I've reproduced the warning, and confirmed that it's gone with these changes. They result in a slight code change on "super" (where the warning was seen):

```diff
 0000000000000210 <common_crypt_code>:
      210:      55                      push   %rbp
      211:      48 89 e5                mov    %rsp,%rbp
      214:      41 57                   push   %r15
      216:      41 56                   push   %r14
      218:      41 55                   push   %r13
      21a:      41 54                   push   %r12
      21c:      49 89 fc                mov    %rdi,%r12
      21f:      53                      push   %rbx
      220:      48 89 f3                mov    %rsi,%rbx
      223:      48 81 ec 98 05 00 00    sub    $0x598,%rsp
      22a:      48 8b 15 00 00 00 00    mov    0x0(%rip),%rdx        # 231 <common_crypt_code+0x21>
      231:      8b 8a 10 20 00 00       mov    0x2010(%rdx),%ecx
      237:      8d 41 01                lea    0x1(%rcx),%eax
      23a:      83 f8 07                cmp    $0x7,%eax
      23d:      76 11                   jbe    250 <common_crypt_code+0x40>
-     23f:      be 67 01 00 00          mov    $0x167,%esi
+     23f:      be 6d 01 00 00          mov    $0x16d,%esi
      244:      bf 00 00 00 00          mov    $0x0,%edi
      249:      e8 00 00 00 00          callq  24e <common_crypt_code+0x3e>
      24e:      66 90                   xchg   %ax,%ax
      250:      89 c0                   mov    %eax,%eax
      252:      ff 24 c5 00 00 00 00    jmpq   *0x0(,%rax,8)
```

which corresponds to the line number of `error();` here:

```c
        default:
                error();
```

So no real change to code.

solardiz:
> ```diff
>       237:      8d 41 01                lea    0x1(%rcx),%eax
> ```

Incidentally, this tells me my idea to internally renumber single DES specifically to -1 may not have been the best. A small positive number would probably have avoided needing the above instruction. But then we'd possibly have to renumber it again later.

solardiz:
> my idea to internally renumber single DES specifically to -1 may not have been the best

Same extra instruction seen on x86_64 with recent gcc. So I renumber to 7 now, which avoids it.

PR #5766, opened 2025-04-29
OpenCL: Fallback to libnvidia-ml.so.1 for NVML

magnumripper:
We've been fine for a well over decade with just dlopening `libnvidia-ml.so` but recently it stopped working on my Ubuntu 24.04.  Turns out we need to fallback to `libnvidia-ml.so.1`.  Not sure if there's a better fix but this is how hashcat does it as well.

### DIFF
--- a/src/gpu_common.c
+++ b/src/gpu_common.c
@@ -126,7 +126,9 @@ void nvidia_probe(void)
 	if (nvml_lib)
 		return;
 
-	if (!(nvml_lib = dlopen("libnvidia-ml.so", RTLD_LAZY|RTLD_GLOBAL)))
+	if (!(nvml_lib = dlopen("libnvidia-ml.so.1", RTLD_LAZY|RTLD_GLOBAL)) &&
+	    !(nvml_lib = dlopen("libnvidia-ml.so", RTLD_LAZY|RTLD_GLOBAL)) &&
+	    !(nvml_lib = dlopen("nvml.dll", RTLD_LAZY|RTLD_GLOBAL)))
 		return;
 
 	nvmlInit = (NVMLINIT) dlsym(nvml_lib, "nvmlInit");
@@ -160,14 +162,10 @@ void amd_probe(void)
 	if (adl_lib)
 		return;
 
-#if HAVE_WINDOWS_H
-	if (!(adl_lib = dlopen("atiadlxx.dll", RTLD_LAZY|RTLD_GLOBAL)) &&
+	if (!(adl_lib = dlopen("libatiadlxx.so", RTLD_LAZY|RTLD_GLOBAL)) &&
+	    !(adl_lib = dlopen("atiadlxx.dll", RTLD_LAZY|RTLD_GLOBAL)) &&
 	    !(adl_lib = dlopen("atiadlxy.dll", RTLD_LAZY|RTLD_GLOBAL)))
 		return;
-#else
-	if (!(adl_lib = dlopen("libatiadlxx.so", RTLD_LAZY|RTLD_GLOBAL)))
-		return;
-#endif
 
 	env = getenv("COMPUTE");
 	if (env && *env)
--- a/src/opencl_dynamic_loader.c
+++ b/src/opencl_dynamic_loader.c
@@ -591,12 +591,12 @@ static void load_opencl_dll(void)
 
 	/* Names to try to load */
 	const char * const opencl_names[] = {
+		"libOpenCL.so.1",	/* Linux/others */
 		"libOpenCL.so",		/* Linux/others, hack via "development" sub-package's symlink */
 		"OpenCL",		/* _WIN */
 		"/System/Library/Frameworks/OpenCL.framework/OpenCL", /* __APPLE__ */
 		"opencl.dll",		/* __CYGWIN__ */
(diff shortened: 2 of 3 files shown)


magnumripper:
While at it, I added a commit that tries to load NVML on Windows, using the name `nvml.dll`. Hashcat also has a fallback that looks in the registry for a path but I'm not going there with no means or incentive to test it.

It's unclear to me if a cygwin build will `HAVE_WINDOWS_H` - if not, we might want to amend those #ifdefs (for AMD ADL as well).

magnumripper:
> It's unclear to me if a cygwin build will `HAVE_WINDOWS_H` - if not, we might want to amend those #ifdefs (for AMD ADL as well).

Perhaps they should be `#if _MSC_VER || __MINGW32__ || __MINGW64__ || __CYGWIN__ || HAVE_WINDOWS_H` as seen in some other corners of our tree?

Actually we could drop the ifdefs and simply try all known names - similar to what Alain did in a156f534b.

magnumripper:
> Actually we could drop the ifdefs and simply try all known names

Did so now, reducing this to a single commit.

magnumripper:
BTW this should not affect https://www.openwall.com/lists/john-users/2025/04/29/1 as far as I can see, although that conversation was what got me to investigate why I no longer saw temperatures from my nvidia/linux since a few weeks.

solardiz:
> dlopening `libnvidia-ml.so` but recently it stopped working on my Ubuntu 24.04. Turns out we need to fallback to `libnvidia-ml.so.1`.

I don't know about this specific library, but in general `.so` is considered a development/build-time library/symlink, and `.so.1` (or whatever number) is the runtime library. So it's a bug to be using `.so` at runtime. And it's correct for a distro not to install `.so` if a development/build sub-package is not installed.

But I suppose using `.so` as a dirty hack fallback if we can't find the right-numbered `.so.1` (or whatever we check for) could work, if we're OK taking the risk of loading a potentially incompatible library version, and having this fallback fail on non-development installs.

src/gpu_common.c
```diff
@@ -126,7 +126,9 @@ void nvidia_probe(void)
 	if (nvml_lib)
 		return;
 
-	if (!(nvml_lib = dlopen("libnvidia-ml.so", RTLD_LAZY|RTLD_GLOBAL)))
+	if (!(nvml_lib = dlopen("libnvidia-ml.so", RTLD_LAZY|RTLD_GLOBAL)) &&
+	    !(nvml_lib = dlopen("libnvidia-ml.so.1", RTLD_LAZY|RTLD_GLOBAL)) &&
```
solardiz: I suggest `.so.1` first, fallback to `.so` next.

src/gpu_common.c
```diff
@@ -160,14 +162,10 @@ void amd_probe(void)
 	if (adl_lib)
 		return;
 
-#if HAVE_WINDOWS_H
-	if (!(adl_lib = dlopen("atiadlxx.dll", RTLD_LAZY|RTLD_GLOBAL)) &&
+	if (!(adl_lib = dlopen("libatiadlxx.so", RTLD_LAZY|RTLD_GLOBAL)) &&
```
solardiz, line 165: Can we identify the right `.so.N` to try first?

magnumripper:
> > dlopening `libnvidia-ml.so` but recently it stopped working on my Ubuntu 24.04. Turns out we need to fallback to `libnvidia-ml.so.1`.
> 
> I don't know about this specific library, but in general `.so` is considered a development/build-time library/symlink, and `.so.1` (or whatever number) is the runtime library. So it's a bug to be using `.so` at runtime. And it's correct for a distro not to install `.so` if a development/build sub-package is not installed.

Yeah I read up on [Program-Library-HOWTO](https://tldp.org/HOWTO/Program-Library-HOWTO/shared-libraries.html) again now, "_Programs, when they internally list the shared libraries they need, should only list the soname they need_", which confirms what you say. When I read it yesterday I got confused by the naming convention of "soname".

Back when I introduced this code, every single example I saw (for nvidia and AMD) used the "linker names" (ending with only `.so`). Hashcat does so too (and still only tries `libatiadlxx.so` for AMD, while falling back to `libnvidia-ml.so.1` if `libnvidia-ml.so` fails). Perhaps due to the same examples.

src/gpu_common.c
magnumripper: Alain's code for dynamically loading libOpenCL.so tries just that first, but does fall back to libOpenCL.so.1. Maybe I should fix that too then.

src/gpu_common.c
magnumripper: I tried peeking at "super" but it doesn't have that lib. Then I tried `dpkg -S libatiadlxx.so` on my Linux (also with .1) to no avail. So I give up.

src/gpu_common.c
magnumripper: Did so

src/gpu_common.c
magnumripper: In fact "libatiadlxx.so.1" doesn't get any google hits!?

magnumripper:
I'm merging this. I can't find any sign that anyone used a proper soname for AMD. Even worse, I saw indications they changed the lib and API two times (!) after this one, each breaking backward compatibility. How stupid is that.

PR #5769, opened 2025-05-16
nt-opencl: Revert 946d7bfa, which did not apply to this format

magnumripper:
While the safety added in that commit was valid for the other formats, it wasn't (and isn't needed in any form) for nt-opencl which is different.

### DIFF
--- a/src/opencl_nt_fmt_plug.c
+++ b/src/opencl_nt_fmt_plug.c
@@ -636,10 +636,6 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 	if (key_idx)
 		BENCH_CLERROR(clEnqueueWriteBuffer(queue[gpu_id], buffer_keys, CL_FALSE, key_offset, 4 * key_idx - key_offset, saved_plain + key_offset / 4, 0, NULL, multi_profilingEvent[0]), "failed in clEnqueueWriteBuffer buffer_keys.");
 
-	/* Safety for when count < GWS */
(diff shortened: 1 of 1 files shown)


solardiz:
Maybe add an explanation of how exactly nt-opencl is different such that this is not needed?

magnumripper:
I could have added all this in the commit description but it feels like overkill: Basically we have two different schemes of packed buffers. Both get rid of not only trailing waste of the password string but also the ending NUL.

One allocates `GWS + 1` offset elements. Plaintext _n_ starts at index `saved_idx[n]` and ends at `saved_idx[n + 1]` (its length is `saved_idx[n + 1] - saved_idx[n]`). The keys are obviously unaligned in this case. This scheme needs the safety added in 946d7bfa because stale data may result in `saved_idx[n + 1] - saved_idx[n]` ending up negative or astronomical.

The other allocates just `GWS` elements. It packs the length in a few of the lower bits of `saved_idx[n]` and the buffer offset in the rest of it. NT uses this scheme but for some reason I misdetected it and the `for (int i = count; i <= gws; i++)` would end up past buffer end. Changing it to `i < gws` is not needed either - this scheme can't suffer from stale data as long as the offset buffer was zeroed at alloc.

We have (I think) variances of the latter scheme: The offset may be aligned to 4 (trailing unaligned bytes zeroed on host side) so not optimally packed but instead quicker processed by the GPU. The index is probably counting 32-bit ints instead of chars in that case. There are probably more variants.

PR #5770, opened 2025-05-17
bitcoin2john.py: SQLite3 format support

mmcx:
This PR add support  SQLite3 format for wallet.dat
see: https://github.com/bitcoin/bitcoin/pull/19077

### DIFF
--- a/doc/NEWS
+++ b/doc/NEWS
@@ -450,6 +450,8 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
   project in August 2020 by supporting (and distinguishing) both our meaning
   (bcrypt-pbkdf + AES-256-CTR) and theirs (MD5 + single DES).  [Solar; 2025]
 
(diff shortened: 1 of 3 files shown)


solardiz:
Thank you for the contribution!

Can you please also add your copyright statement, and an update to `doc/NEWS`?

Then as a separate change (at least separate commit, and probably separate PR) you could move the `import` of the two database modules (or at least of one of them, perhaps BDB given the way you currently check for SQLite via the module) to be done conditionally only upon detecting that database type.

Please also provide a sample wallet file (with no coins in the wallet) with a known password, ideally via a PR to https://github.com/openwall/john-samples

mmcx:
* copyright statement added
* doc/NEWS updated
* conditional import done
* sample wallet here: https://github.com/openwall/john-samples/pull/38
* сompatibility with python2 and python3 is preserved and tested

solardiz:
Thanks again for your contribution. I've now tested the updated script with the sample SQLite3 wallet you provided on a system with Python 3, and with our older BDB wallet sample on a system with Python 2.

PR #5773, opened 2025-05-19
Update and improve our huge pages support

solardiz:
This syncs and combines the two code versions and adds reporting.


solardiz:
I thought we could possibly want to suppress reporting on huge page (non-)usage when it's solely by the dupe suppressor (since this is usually unimportant), but I didn't arrive at a clean way to implement this. So we're reporting on all huge page (non-)usage for now. This is a reason why I don't prefix any of this with "Warning: " and made the messages such that most of them don't make it sound that anything is wrong (except for the one about partial huge page usage as it could indicate behavior unexpected by the user who presumably had made a custom configuration).

src/yescrypt/yescrypt-platform.c
```diff
@@ -25,22 +25,76 @@
...
+/*
+ * JtR hack: on old systems like RHEL6, do use huge pages if available even at
+ * the risk of failing on munmap() if the kernel is reconfigured to use a
+ * non-default huge page size.  In upstream yescrypt, this tradeoff is resolved
+ * the other way around because failure of a critical system library like
+ * libxcrypt is deemed unacceptable.
+ */
+#ifndef MAP_HUGE_2MB
+#define MAP_HUGE_2MB 0
+#endif
```
magnumripper, line 37: This is potentially very confusing but I guess it keeps the diff against upstream smaller? Edit: I don't really object to it, just commenting.

solardiz: Yes, and it keeps not only the diff but also the code smaller. But I agree it's dirty, and not something I would do in upstream yescrypt.

solardiz:
This also passed testing on Alpine Linux in a Docker container on Arch Linux, and in a Podman container on Rocky Linux 9.5.

However, it required `apk add linux-headers` for our added `#include <linux/mman.h> /* for MAP_HUGE_2MB */`. That's rather unfortunate. It got me thinking of defining `MAP_HUGE_2MB` ourselves, but then we'd try to use it even on old systems that lack it, so would fail at runtime on RHEL6. Which also reminds me that the same failure would happen if a `john` binary is moved from a newer to an older Linux system (if it's otherwise compatible enough to run).

Maybe the right way is to try (the numeric value of) `MAP_HUGE_2MB` and fallback to not using it at runtime. Maybe that would even be the right way for upstream yescrypt as well, eliminating the difference.

solardiz:
> Maybe the right way is to try (the numeric value of) MAP_HUGE_2MB and fallback to not using it at runtime. Maybe that would even be the right way for upstream yescrypt

Created https://github.com/openwall/yescrypt/issues/8

PR #5774, opened 2025-05-19
Stacked rules: Fix rule number in logs, in Single mode

### LINKED ISSUE #5729: --rules-stack with --single logs "Stacked Rule #1" twice
This is a discrepancy in logging the number of stacked rule. Those two log lines actually refer to different rules (would be number 1 and 2 if numbered correctly).

magnumripper:
Closes #5729

Tested in all combos I could imagine: Single+stack, wordlist with rules+stack or only stack, prince with rules+stack or only stack. Then all of them again with stopped/resumed session. Is there anything more we should test?

I refrained from doing anything more (which would actually be #5722) to avoid untimely regression but I did put some debug stuff in there, trying to verify we don't have more crazy bugs such as the one fixed by 3e75e1bd. I saw nothing suspicious.

### DIFF
--- a/src/rules.c
+++ b/src/rules.c
@@ -1804,9 +1804,10 @@ out_ERROR_UNALLOWED:
 }
 
 /*
- * Advance stacked rules. We iterate main rules first and only then we
- * advance the stacked rules (and rewind the main rules). Repeat until
- * main rules are done with the last stacked rule.
+ * Advance stacked rules. We run all words through each main rule,
+ * rewinding the wordlist inbetween them. Each result from that is run
+ * through each stacked rule, rewinding the latter before advancing to
+ * next word (or next main rule).
  */
 int rules_advance_stack(rule_stack *ctx, int quiet)
 {
--- a/src/rules.c
+++ b/src/rules.c
@@ -1816,7 +1816,7 @@ int rules_advance_stack(rule_stack *ctx, int quiet)
 		rules_stacked_number++;
 		if (!quiet)
 			log_event("+ Stacked Rule #%u: '%.100s' accepted",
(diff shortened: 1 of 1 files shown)


magnumripper:
I see now I should add a fix in this PR for this incorrect and/or confusing comment (close to the fix and same function):
```c
/*
 * Advance stacked rules. We iterate main rules first and only then we
 * advance the stacked rules (and rewind the main rules). Repeat until
 * main rules are done with the last stacked rule.
 */
```
Not sure how to put it concisely.

For words alpha and beta, rules prepend-1 and prepend-2 and stacked rules append-a and append-b, the result is
```
1alphaa
1alphab
1betaa
1betab
2alphaa
2alphab
2betaa
2betab
```
So maybe
```c
/*
 * Advance stacked rules. We run all words through each main rule,
 * rewinding the wordlist inbetween them. Each result from that is run
 * through each stacked rule, rewinding the latter before advancing to
 * next main rule.
 */
```

magnumripper:
> ```c
> /*
>  * Advance stacked rules. We run all words through each main rule,
>  * rewinding the wordlist inbetween them. Each result from that is run
>  * through each stacked rule, rewinding the latter before advancing to
>  * next main rule.
>  */
> ```

Not quite. More like:
```c
/*
 * Advance stacked rules. We run all words through each main rule,
 * rewinding the wordlist inbetween them. Each result from that is run
 * through each stacked rule, rewinding the latter before advancing to
 * next word (or next main rule).
 */
```

solardiz (approved):
Looks good to me. I had concerns at first (such as the patched line never logging a rule number less than 2), but reviewing this in more context eliminated those.

PR #5776, opened 2025-05-23
Several formats' prepare(): Fix bugs when checking fields

magnumripper:
fields[n] is never NULL, and the string fields[0] points to is only a nullstring if there is an explicit field saying so (i.e. line starts with a ":")

For a missing login field (no ":", bare hash) fields[0] is set to "?" in loader.
    
This bug made some formats make up uncrackable hashes from thin air instead of rejecting when salt is unknown - leading to false negatives. Those formats are in a separate commit.

This was tested manually and with Test Suite.

### DIFF
--- a/src/MSCHAPv2_bs_fmt_plug.c
+++ b/src/MSCHAPv2_bs_fmt_plug.c
@@ -281,13 +281,11 @@ static char *prepare(char *split_fields[10], struct fmt_main *pFmt)
 			}
 		}
 	}
-	else if (split_fields[0] && split_fields[3] && split_fields[4] && split_fields[5] &&
-	        strlen(split_fields[3]) == CHALLENGE_LENGTH/2 &&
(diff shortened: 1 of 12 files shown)


solardiz (approved):
The checks for `?` are not ideal - they'd also be triggered on literal username `?`. OTOH, such a username would also be confusing in our output.

Anyway, I agree these changes are what we need right now.

magnumripper:
> The checks for `?` are not ideal - they'd also be triggered on literal username `?`. OTOH, such a username would also be confusing in our output.

I've been thinking instead of adding a pretty much useless "?" we could print the hash (or part of it) for such hashes, perhaps with a notice what is happening. That way the user can correlate the output with the input without resorting to the potfile. And that would be similar to hashcat (which, I think prints the whole hash even if it's several KB or MB).

PR #2510, opened 2017-04-21
racf2john update to dump KDFAES hashes

bigendiansmalls:
### Summary

racf2john currently dumps DES RACF hashes, this update
allows it to also dump the newer type (2014) KDFAES hashes.

The new password algorithm announcement can be found here:
https://www-01.ibm.com/support/docview.wss?uid=isg1OA43999

### The new output looks like:
```
sh-4.4$ run/racf2john
Usage: racf2john [RACF binary files]

sh-4.4$ run/racf2john ./testdb.bin
BBBBBBB:$racf$*BBBBBBB*932C1BBBB12A8274
C123456:$racf$*C123456*50A4B524A1854E38
A123456:$racf$*A123456*E7D7E66D00018000000800320010001041E09CD903BD84255D547CC162DCC1EC135D5D0AF2961EE5DD01B5FF489D2776
B123456:$racf$*B123456*E7D7E66D0001800000080032001000103D743EDF9DB18E2762E9C3F801672954A787F176DCF675EBE303B463FC81775A
```
The short hashes are DES and are currently supported by John for cracking.
The longer hashes are not yet supported by John, but do represent the KDFAES format.

PR #5777, opened 2025-05-25
RACF-KDFAES format fixes

solardiz:
This PR is to make the format a little bit more correct and reliable, maybe throw in some easy optimizations too (not done yet).


solardiz:
See also: #3486 #2511 #2510
cc: @bigendiansmalls

bigendiansmalls:
Happy to test the changes and make sure nothing broke :) Will let you know. Holiday weekend so might take me a day.

solardiz:
There's finally one commit here (building upon preceding commits) that makes this ~30% faster, but it's still very slow indeed, even when using SHA-NI via OpenSSL (I've confirmed that it does for me).

A further step could be to implement SIMD for the nested loop, which shouldn't be too hard and would cover most of the running time. But I probably won't bother now.

solardiz:
> There's finally one commit here (building upon preceding commits) that makes this ~30% faster,

I had missed part of the optimization there. Corrected. Now this one commit makes the format ~75% faster. Along with the rest of the changes here, we're ~80% faster.

solardiz:
Having worked on these changes, I now have a better idea of what this KDFAES thing is. It's modeled after scrypt, where there are 3 steps: key expansion, memory-hard loop, and calculation of digest of the final state.

1. In scrypt, key expansion is done with PBKDF2 at 1 iteration but a derived key length of like 1 KiB. It produces just the initial block B. In KDFAES, it's done with an algorithm extremely similar to but not quite PBKDF2, the iteration count is high, and it produces the full memory content, which in scrypt would be V.

2. The memory-hard loop is similar to scrypt SMix's second loop, except that besides random reads (similar to scrypt's) it writes over memory (scrypt's only reads), which it does sequentially. The first loop (which would fill V given B) is missing - or alternatively we can say that it's replaced by the PBKDF2-alike's outer loop (over derived key blocks).

3. The final state digest calculation in scrypt is done with PBKDF2 at 1 iteration over the final block. Here it's done over the full memory content, and again with high iteration count. The algorithm _may_ be literally PBKDF2 (I didn't verify, but I'm also not aware of this one deviating from proper PBKDF2).

The initial key expansion in scrypt is quick, but in KDFAES it takes almost all its running time because the complexity is quadratic. The would-be memory-hard portion only goes over memory once (fine on its own) and unlike the key expansion it doesn't use iterations, making it _relatively_ quick. So this isn't truly a memory-hard algorithm. The final digest calculation doesn't suffer from quadratic complexity because the large input (full memory content) is hashed first, so even despite of the high iteration count it's _relatively_ quick.

So we may try calling into our shared PBKDF2 code for step 3, but this isn't going to affect overall performance much. What will make more of a difference is calling into our SIMD SHA-256 for step 1.

solardiz:
> Happy to test the changes and make sure nothing broke :) Will let you know. Holiday weekend so might take me a day.

I've just merged, but of course please do test and let us know when you can. Also I suspect that adding support for length > 8 may require a revert of some of the HMAC code specialization here. That's fine.

solardiz:
Just because I was documenting my findings in this PR, I'll also add in here today's Kaspersky SecureList blog post on the topic:

https://securelist.com/zos-mainframe-pentesting-resource-access-control-facility/116873/

There's a lot, but one piece I'd like to quote is:

> Encrypting a password phrase hash with KDFAES shares many similarities with encrypting a password hash. According to public sources, the primary difference lies in the key used during the second stage. For passwords, data derived from DES-encrypting the username was used, while for a password phrase, its SHA256 hash is used. During our analysis, we could not determine the exact password phrase hashing process – specifically, whether padding is involved, if a secret key is used, and so on.
> 
> Additionally, when using a password phrase, the PHRASE and PHRASEX fields instead of PASSWORD and PWDX, respectively, store the final hash, with the PHRASEX value having a similar structure.

So to support passwords (phrases) longer than 8, apparently the following would be needed:
1. An extension to `racf2john` to extract more fields.
2. Figure out the algorithm and implement it.

solardiz:
There's also a hashcat PR implementing this algorithm (for length up to 8 like we have) https://github.com/hashcat/hashcat/pull/3940 by @c0rs

PR #5779, opened 2025-05-27
HMAC: Optimize opad XOR

solardiz:
There are several possible ways to construct HMAC's ipad and opad:
1. Initialize them (or one buffer to be reused as both) to key with NUL padding, then XOR the constants over that.
2. Initialize them to the constants, then XOR the key over the initial portions (until key length).
3. Compute the initial portions right away, then pass the rest of the constant data into separate *_Update() calls.
4. Combinations of the above and other variations (e.g., partial memset(), doing it out of a loop, etc.)

This commit optimizes existing implementations of 1 and 3 by XOR'ing the one reused pad buffer with XOR of the two constants instead of re-reading the key and XOR'ing that with the second constant.  Some implementations already had this optimization, and so are not touched by this commit.

Explicit usage of larger-than-byte XORs may be even more important, but isn't implemented with this minimal-change commit (would be non-trivial with arbitrary sizes and alignment of keys).  Some implementations already have this optimization.

This commit does not touch existing implementations of 2 and 4.  Those may (or may not) actually be as good as or even more optimal than optimized 1, such as due to efficiency of memset() and typically low key lengths.

This commit reduces john .text size by ~400 bytes in a Linux/x86_64 build.

### DIFF
--- a/src/rsvp_fmt_plug.c
+++ b/src/rsvp_fmt_plug.c
@@ -306,7 +306,7 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 				if (len < 64)
 					MD5_Update(&ipad_mctx[index], ipad_constant_block, 64-len);
 				for (i = 0; i < len; ++i) {
-					pad[i] = p[i] ^ 0x5C;
+					pad[i] ^= 0x36 ^ 0x5C;
 				}
 				MD5_Init(&opad_mctx[index]);
 				MD5_Update(&opad_mctx[index], pad, len);
@@ -342,7 +342,7 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 				if (len < 64)
 					SHA1_Update(&ipad_ctx[index], ipad_constant_block, 64-len);
 				for (i = 0; i < len; ++i) {
-					pad[i] = p[i] ^ 0x5C;
+					pad[i] ^= 0x36 ^ 0x5C;
 				}
 				SHA1_Init(&opad_ctx[index]);
 				SHA1_Update(&opad_ctx[index], pad, len);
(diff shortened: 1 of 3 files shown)


solardiz:
Not important and in hindsight wasn't worth my time. Mostly just a cleanup so that we use fewer different approaches, and they're kind of documented in this commit message.

magnumripper:
> XOR'ing the one reused pad buffer with XOR of the two constants instead of re-reading the key and XOR'ing that with the second constant.

Worthwhile or not it's a clever trick, I never saw it or thought of it.

magnumripper:
> Some implementations already had this optimization, and so are not touched by this commit.

Which? I can't find any, neither when grepping for "0x36 ^ 0x5c" nor "0x6a" (and with fuzz).

solardiz:
> > Some implementations already had this optimization, and so are not touched by this commit.
> 
> Which? I can't find any, neither when grepping for "0x36 ^ 0x5c" nor "0x6a" (and with fuzz).

I saw this during my work on #5777 where I copied code from `hmac_sha_plug.c`, which has this already.

> Worthwhile or not it's a clever trick, I never saw it or thought of it.

I also don't recall seeing it before, but there it was right in our tree, just not used consistently.

Clever, yes, but frankly it's one read, one XOR, and one write per element either way. So it mostly optimizes code size on x86-alikes, where we have read-modify-write instructions. There may also be savings in effective address calculation (just one vs. two addresses), and a register and a cache line become unused (free for reuse with no performance impact) a little bit sooner.

solardiz:
Oh, I only reviewed our C code, not OpenCL. We probably should. There may be more instances to patch if we care.

magnumripper:
I did look, and several kernels has this. Not all though.

solardiz:
Actually, this kind of change may be more beneficial on GPUs where the buffer is likely all in registers and the loop unrolled. So it's not a read-modify-write of memory then, but is just of a register (so one instruction), instead of perhaps... the same but with two sets of registers? So it frees up the registers holding the key sooner, and thus may reduce total number of registers needed for a kernel.

solardiz:
> Oh, I only reviewed our C code, not OpenCL.

I've just revised the commit title to reflect that. I probably won't bother with OpenCL now. You may. ;-)

magnumripper:
It's tempting. I'll likely try WPAPSK, bench it and compare sizes for a starter. Oh and all ptxas info.

solardiz:
> It's tempting. I'll likely try WPAPSK, bench it and compare sizes for a starter. Oh and all ptxas info.

Right. Since this would be a separate PR (yours vs. mine), should we perhaps merge this one first?

magnumripper:
Sure, please do.

PR #5780, opened 2025-05-27
OpenCL kernels: Use our INLINE macro consistently

solardiz:
See #5618 #5638 #5778

I re-ran the tests from #5709 - these changes didn't affect them. Also a test on NVIDIA is successful like before.


solardiz:
Tested on super's devices, no changes observed.

PR #5784, opened 2025-06-09
Add H3C/Huawei/HPE hash format

SamuraiOcto:
This adds support for hashes with `$h$6$` prefix that are used in various H3C/Huawei/HPE network devices. The SHA512-based format is described in [this hashcat issue](https://github.com/hashcat/hashcat/issues/3963) and [this PR](https://github.com/hashcat/hashcat/pull/4023).

### DIFF
--- a/src/h3c_fmt_plug.c
+++ b/src/h3c_fmt_plug.c
@@ -38,7 +38,7 @@ john_register_one(&fmt_h3c);
 
 #ifdef SIMD_COEF_64
 #define ALGORITHM_NAME     SHA512_ALGORITHM_NAME
-#define NBKEYS                     (SIMD_COEF_64*SIMD_PARA_SHA512)
+#define NBKEYS             (SIMD_COEF_64*SIMD_PARA_SHA512)
 #else
 #define ALGORITHM_NAME     "64/" ARCH_BITS_STR
 #define NBKEYS             1
@@ -47,7 +47,7 @@ john_register_one(&fmt_h3c);
 #define BENCHMARK_COMMENT  ""
 #define BENCHMARK_LENGTH   0x107
 #ifdef SIMD_COEF_64
-#define PLAINTEXT_LENGTH   ((128-16-8-1)/2-1)
+#define PLAINTEXT_LENGTH   46
 #else
 #define PLAINTEXT_LENGTH   125
 #endif
@@ -85,6 +85,16 @@ static struct fmt_tests h3c_tests[] = {
 		FORMAT_TAG "abcdefghijklmnop$jp3hDbVlf/L1GNDE4n6x4wqvHnFiEr4YrtM6ax1aFXFb6pdu4Nfmpp09pZFOGOH8ID9vw2AOKQA4q8lByhlG4A==",
 		"password"
 	},
+	{
+		FORMAT_TAG "z)`6LE)4lBPKdDEr$MlUS0GBjnwzMOzFEhniVm+F9S9/YxdAKUy04NPCH5eXckRdk6fhH3gEGQmllhtrppRcGymdCxCrs9sBQJj2jbw==",
+		"1234567890123456789012345678901234567890123456"
+	},
+#ifndef SIMD_COEF_64
+	{
+		FORMAT_TAG "aIYeQUH63hnhJ1lK$7CqVJE2q/Es8B8Fqnp5349OQdrdv/kUYHTNiU7Aq+Am8LZG9u+Gawnn5SVEqn8N89n+agG449J0W0wlSk/cuvw==",
+		"12345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345"
+	},
+#endif
 	{ NULL }
 };
 
--- a/src/h3c_fmt_plug.c
+++ b/src/h3c_fmt_plug.c
@@ -1,0 +1,320 @@
+/*
+ * H3C/HPE/Huawei format plugin
+ *
+ * Copyright (c) 2025 SamuraiOcto
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted.
+ *
+ * There's ABSOLUTELY NO WARRANTY, express or implied.
+ */
+
+#if FMT_EXTERNS_H
+extern struct fmt_main fmt_h3c;
+#elif FMT_REGISTERS_H
+john_register_one(&fmt_h3c);
+#else
+
+#include <string.h>
+
+#ifdef _OPENMP
+#include <omp.h>
+#endif
+
+#include "arch.h"
+#include "common.h"
+#include "formats.h"
+#include "params.h"
+#include "options.h"
+#include "johnswap.h"
+#include "base64_convert.h"
+#include "sha2.h"
+#include "simd-intrinsics.h"
+
+#define FORMAT_LABEL       "h3c"
+#define FORMAT_NAME        "H3C/Huawei/HPE"
+#define FORMAT_TAG         "$h$6$"
+#define TAG_LENGTH         (sizeof(FORMAT_TAG)-1)
+
+#ifdef SIMD_COEF_64
+#define ALGORITHM_NAME     SHA512_ALGORITHM_NAME
+#define NBKEYS                     (SIMD_COEF_64*SIMD_PARA_SHA512)
+#else
+#define ALGORITHM_NAME     "64/" ARCH_BITS_STR
+#define NBKEYS             1
+#endif
+
+#define BENCHMARK_COMMENT  ""
+#define BENCHMARK_LENGTH   0x107
+#ifdef SIMD_COEF_64
+#define PLAINTEXT_LENGTH   ((128-16-8-1)/2-1)
+#else
+#define PLAINTEXT_LENGTH   125
+#endif
+#define BINARY_SIZE        8
+#define BINARY_ALIGN       4
+#define SALT_SIZE          16
+#define SALT_ALIGN         1
+#define MIN_KEYS_PER_CRYPT NBKEYS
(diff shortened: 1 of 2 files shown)


solardiz:
Thank you for the contribution @SamuraiOcto!

Can you please add test vectors for the maximum supported password lengths (with and without SIMD)?

I wonder if this hash is implementable as a dynamic format. If so, maybe we'd need to switch this format to a "thin" wrapper around dynamic, as we already have in a few formats.

SamuraiOcto:
Is it possible to include the null bytes with a dynamic format, like `sha512($p."\x00".$s.$p."\x00")`? It doesn't seem to be possible to include null bytes in the wordlist. It is possible to merge the first null byte with the salt and provide the salt as hexadecimal, but that won't work for the second null byte. It is probably possible to create a rule that appends a null byte to all plaintexts, but then it will be annoying if a user wants to use another rule file.

magnumripper:
> I wonder if this hash is implementable as a dynamic format. If so, maybe we'd need to switch this format to a "thin" wrapper around dynamic, as we already have in a few formats.

We could, but why "need"? This one doesn't look too bad. Perhaps we could avoid using scalar buffers for a small boost if we really wanted to.

solardiz:
> We could, but why "need"?

Oh, I used the wrong word. We certainly can accept this as a separate format as well. I just thought we could prefer to reduce code size / duplication at some point. Formats like these have rather simple logic, but much code shared with other formats.

src/h3c_fmt_plug.c
```diff
@@ -0,0 +1,320 @@
...
+
+	if (ciphertext[SALT_SIZE] != '$')
+		return 0;
+
+	ciphertext += SALT_SIZE + 1;
+
+	len = strspn(ciphertext, BASE64_ALPHABET);
+	if (len != CIPHERTEXT_LENGTH)
+		return 0;
+	return 1;
```
solardiz, line 136: I see we have similar logic in `sha512_common_valid_nsldap`, and in both places we don't fully check that the base64 encoding is valid (`=` only used as padding). We might want to enhance both with a test `base64_convert` run and checking its return value. A way to do that may be to have `get_binary` return NULL if `base64_convert` fails to provide the expected length data, call `get_binary` from `valid` and check its return value. Since we would also be fixing the pre-existing code, this (or at least that part) should be a separate commit (may also be separate PR).

src/h3c_fmt_plug.c
```diff
@@ -0,0 +1,320 @@
...
+		uint64_t x;
+	} x;
+	char *realcipher = x.out;
+
+	ciphertext += TAG_LENGTH + SALT_SIZE + 1;
+	base64_convert(ciphertext, e_b64_mime, strlen(ciphertext), realcipher, e_b64_raw, sizeof(x.out), flg_Base64_NO_FLAGS, 0);
+
+#if defined(SIMD_COEF_64) && ARCH_LITTLE_ENDIAN==1
+	alter_endianity_to_BE64(realcipher, DIGEST_SIZE / 8);
+#endif
```
solardiz: It's good that we adjust for endianity on `get_binary`. I am puzzled why we also have such adjustments in `crypt_all` and `cmp_exact`. I think we could have the `binary` in the byte order that is expected by further code and that's it.

src/h3c_fmt_plug.c
```diff
@@ -0,0 +1,320 @@
...
+}
+
+static int cmp_one(void *binary, int index)
+{
+	return *(uint64_t *) binary == crypt_out[index][0];
+}
+
+static int cmp_exact(char *source, int index)
+{
+	uint64_t *binary = get_binary(source);
```
solardiz: I think we don't need a fully implemented `cmp_exact` at all. We could instead compare the full computed hash in `cmp_one`.

solardiz (approved):
I made some comments, but really those can be addressed separately.

solardiz:
> Is it possible to include the null bytes with a dynamic format, like `sha512($p."\x00".$s.$p."\x00")`?

Oh, I don't know. Maybe not. Good point.

solardiz:
> hashes with `$h$6$` prefix that are used in various H3C/Huawei/HPE network devices

Did the two test vectors here come from actual devices?

magnumripper:
> > We could, but why "need"?
> 
> Oh, I used the wrong word. We certainly can accept this as a separate format as well. I just thought we could prefer to reduce code size / duplication at some point. Formats like these have rather simple logic, but much code shared with other formats.

True. Another possibility could be to amend the `SSHA512` format to accept these hashes, but maybe it's different enough it would cause a regression from added logic.

Anyway, format code like this current PR is the nicest start for implementing an OpenCL version. Easy to follow and easy to move valid/binary/salt to shared code. And a big problem with dynamic is it doesn't have an active maintainer.

magnumripper:
I tried doing this in dynamic compiler format just for the hell of it (re-encoding the Base64 into hex manually), but it failed with no clue as to why (possibly the NUL constant - or maybe \xHH isn't even supported but taken as-is, there's very little documentation).

I then looked into doing it with the config style dynamic but the various docs contradict each other so I gave up. In that case \xHH is said to be supported but it doesn't mention \x00 so that is an unknown already.

src/h3c_fmt_plug.c
magnumripper: I just dropped all of them and it still works
```diff
diff --git a/src/h3c_fmt_plug.c b/src/h3c_fmt_plug.c
index d3b1b04fd..dc4664967 100644
--- a/src/h3c_fmt_plug.c
+++ b/src/h3c_fmt_plug.c
@@ -137,9 +137,6 @@ static void *get_binary(char *ciphertext)
        ciphertext += TAG_LENGTH + SALT_SIZE + 1;
        base64_convert(ciphertext, e_b64_mime, strlen(ciphertext), realcipher, e_b64_raw, sizeof(x.out), flg_Base64_NO_FLAGS, 0);
 
-#if defined(SIMD_COEF_64) && ARCH_LITTLE_ENDIAN==1
-       alter_endianity_to_BE64(realcipher, DIGEST_SIZE / 8);
-#endif
        return (void *)realcipher;
 }
 
@@ -192,11 +189,7 @@ static int crypt_all(int *pcount, struct db_salt *salt)
 
                for (int i = 0; i < MIN_KEYS_PER_CRYPT; ++i) {
                        for (int j = 0; j < DIGEST_SIZE / sizeof(uint64_t); ++j) {
-#if ARCH_LITTLE_ENDIAN==1
-                               crypt_out[index + i][j] = JOHNSWAP64(out[i * (DIGEST_SIZE / sizeof(uint64_t)) + j]);
-#else
                                crypt_out[index + i][j] = out[i * (DIGEST_SIZE / sizeof(uint64_t)) + j];
-#endif
                        }
                }
 #else
@@ -252,11 +245,6 @@ static int cmp_exact(char *source, int index)
        SHA512_Update(&ctx, key, strlen(key) + 1);      // include null byte
        SHA512_Final((unsigned char *)crypt_out, &ctx);
 
-#ifdef SIMD_COEF_64
-#if ARCH_LITTLE_ENDIAN==1
-       alter_endianity_to_BE64(crypt_out, DIGEST_SIZE / 8);
-#endif
-#endif
        return !memcmp(binary, crypt_out, DIGEST_SIZE);
 }
 
```
For the BE case this was dead code so provided it ever worked, it still does.

src/h3c_fmt_plug.c
magnumripper: We need this as-is as long as the format use a reduced binary size (8 of 64). That's not needed for this format though, so the fix could be to stop using that, which would also get rid of some copying in SIMD crypt_all. Then cmp_one would be just a full memcmp and cmp_exact would just return 1.

src/h3c_fmt_plug.c
magnumripper: I added this to the PR

src/h3c_fmt_plug.c
magnumripper: I added a commit that drops the idea of reduced binary size and simplifies the rest of the code accordingly. Oh and I also fixed BINARY_ALIGN which was not correct either way.

magnumripper:
I didn't look into an improved Base64 check but I think we can merge this now.

SamuraiOcto:
> Can you please add test vectors for the maximum supported password lengths (with and without SIMD)?

I did and because of that I noticed that `PLAINTEXT_LENGTH` was slightly off, so that's adjusted now.

> Did the two test vectors here come from actual devices?

Only the first one, the others are calculated by Python.

magnumripper:
@SamuraiOcto are you planning to add an OpenCL version as well, or should I put that on my to-do list?

SamuraiOcto:
> @SamuraiOcto are you planning to add an OpenCL version as well, or should I put that on my to-do list?

I'm not very familiar with OpenCL so I wasn't planning on it. So if that's something that you'd want to work on then by all means go ahead.

solardiz:
AIX 7.1 `./configure --disable-pcap --disable-simd` build fails as follows:

```
Benchmarking: h3c, H3C/Huawei/HPE [SHA512 64/32]... FAILED (get_hash[0](0) a7!=56)
```

Also, the `64/` is wrong reporting for a 32-bit build.

solardiz:
> AIX 7.1 `./configure --disable-pcap --disable-simd` build fails as follows:
> 
> ```
> Benchmarking: h3c, H3C/Huawei/HPE [SHA512 64/32]... FAILED (get_hash[0](0) a7!=56)
> ```
> 
> Also, the `64/` is wrong reporting for a 32-bit build.

I have fixes for these pending. And it turns out we have the `64/` error in lots of formats.

solardiz:
Fixes pushed. Also for stricter valid(), as it previously would accept any trailing string, and `=` anywhere.

Issue #5767, opened 2025-05-03
Problem in cracking SL3 for Nokia E7 (RM-626)


kkjangra:
Hi khushwant kumar this side,
I have nokia E7 (RM-626).
IMEI and Hash are
354864040624588:3E4B60A13609477E0996958955497928372F2217

I am cracking it using hashcat brute force but failed once and trying 2nd time with 
\hashcat.exe -a 4500 -a 3 -D 2  .\hash.txt ?dPd?d?d?d?d?d?d?d?d?d?d?d?d?d
Hash.txt contains only hash code.

It will be appreciable if anybody could help to crack it.

Or please give me the cracked code if any of you can.

magnumripper:
Hashcat doesn't support SL3 since many years. To use that hash with our SL3 format you need to lower case it to 354864040624588:3e4b60a13609477e0996958955497928372f2217

You have a "P" in your mask, I assume that is unintentional. Try `-mask=?d -len=15` instead. With average luck it will take ~12 hours using a single decent GPU or 24h at worst.

```
john hash.txt -form:sl3-opencl -mask:?d -len=15
```

magnumripper:
Also see https://github.com/openwall/john/issues/4169#issuecomment-2849051779 for possibly needed transfomation to NCK (whatever that is).

solardiz:
> To use that hash with our SL3 format you need to lower case it

Should we maybe add support for either case, with unification in `split`?

> Also see https://github.com/openwall/john/issues/4169#issuecomment-2849051779 for possibly needed transfomation to NCK (whatever that is).

Should we maybe add this to our documentation?

kkjangra:
> Also see [#4169 (comment)](https://github.com/openwall/john/issues/4169#issuecomment-2849051779) for possibly needed transfomation to NCK (whatever that is).

I have used those links but couldn't succeed.
Yes that P in mask id was mistyped.
Now I am going to try using "OCLhashcat 0.09" on an older computer as it is not supported by newer systems.

Rest if you have any other idea then please let me know. I have gone through your John crack8ng software but i have limitations to use that. Would it crack SL3 ? If yes then please do it for me.

magnumripper:
> I have used those links but couldn't succeed.

[...]

PR #5787, opened 2025-06-11
Add a valid_luhn() function, and use it for SL3

magnumripper:
https://github.com/openwall/john/issues/5767#issuecomment-2862115937

I wanted to base it on https://rosettacode.org/wiki/Luhn_test_of_credit_card_numbers#C which is nice code, but it appears the licensing would be tricky so here's instead my C implementation of the pseudo code on Wikipedia.

There were actually a mockup SL3 test vector featuring an incorrect Luhn so I had to correct that as well.

### DIFF
--- a/src/jumbo.c
+++ b/src/jumbo.c
@@ -470,3 +470,39 @@ char *replace(char *string, char c, char n)
 
 	return string;
 }
+
+/*
+ * This is based on the pseudo-code from wikipedia and not meant to be fast.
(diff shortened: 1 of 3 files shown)


solardiz (approved):
I didn't verify correctness of the Luhn algorithm implementation vs. Wikipedia or whatever, but I skimmed the code and it looks reasonable to me.

Issue #5788, opened 2025-06-17
requirements.txt update


solardiz:
We have a couple of version numbers in our `requirements.txt`:

```
protobuf~=3.20
pycryptodome
scapy
parsimonious
asn1crypto
dpkt
pyhanko==0.20.1
```

Why those specific versions? Should we be updating them?

Dependabot sent us this alert: https://github.com/openwall/john/security/dependabot/1

``` 
  protobuf dependency
  ------------------------------------------------
  Vulnerable versions: < 4.25.8
  Upgrade to: 4.25.8
  Defined in: requirements.txt
  Vulnerabilities:  CVE-2025-4565 (High severity)
```

@exploide As usual for Python stuff, I'd appreciate your guidance on this. Thank you!

exploide:
For protobuf, this was discussed in #5474. From there:

> The error message explicitly says that the current Python code uses features that are deprecated in protobuf>3.20. We'd likely have to regenerate the Python protobuf files, iirc (which is a huge pain).

So this seems to be a current limitation in `run/protobuf/wallet_pb2.py` and `run/multibit2john.py`.
These probably need to be rewritten or at least to be reworked to a certain degree. (I have no experience with protobuf and don't know how much work it is.)

exploide:
The `pyhanko==0.20.1` requirement is from `pdf2john.py`. Don't know why it is constrained. Maybe we could test with a newer version and the pdf test vectors to see if it still works.

solardiz:
The protobuf CVE description says it's infinite recursion DoS. Well, not a big deal if this script can be crashed. Also, I am not sure the vulnerability is exposed in our usage.

PR #5790, opened 2025-06-19
pdf2john.py: fix compatibility with current pyhanko versions

exploide:
See #5788

The actual change for compatibility was easy. But while I was at it, I improved the script's resilience with non-pdf files and unencrypted pdf files.

Produces the same output as before on `john-samples/**.pdf`.

### DIFF
--- a/run/pdf2john.py
+++ b/run/pdf2john.py
@@ -8,13 +8,14 @@
 
 import argparse
 import logging
+import sys
 
 try:
     from pyhanko.pdf_utils.misc import PdfReadError
     from pyhanko.pdf_utils.reader import PdfFileReader
 except ImportError:
-    print("pyhanko is missing, run 'pip install --user pyhanko==0.20.1' to install it!")
-    exit(1)
+    print("pyhanko is missing, run 'pip install --user pyhanko' to install it!", file=sys.stderr)
+    sys.exit(1)
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


run/pdf2john.py
```diff
@@ -8,13 +8,14 @@
...
+import sys
 
 try:
     from pyhanko.pdf_utils.misc import PdfReadError
     from pyhanko.pdf_utils.reader import PdfFileReader
 except ImportError:
-    print("pyhanko is missing, run 'pip install --user pyhanko==0.20.1' to install it!")
-    exit(1)
+    print("pyhanko is missing, run 'pip install --user pyhanko' to install it!", file=sys.stderr)
+    sys.exit(1)
```
solardiz, line 18: BTW, we also have uses of plain `exit` in a few other scripts:

```
apop2john.py:    exit(1)
apop2john.py:    exit(1)
authenticator2john.py:    exit(1)
authenticator2john.py:        exit(1)
bestcryptve2john.py:        exit(0)
bitlocker2john.py:            exit()
kdcdump2john.py:    exit()
mosquitto2john.py:            exit(1)
```

solardiz (approved):

solardiz:
Thank you @exploide!

run/pdf2john.py
exploide: I could fix them. But these are rather minor linting issues and many scripts produce several linting issues. Too many to go over everything in every script. So tackling exactly the `exit()` issue would feel a bit random and not very important.

PR #5791, opened 2025-06-23
Sync ssh-opencl with the CPU format

solardiz:
This brings ssh-opencl almost on par with the CPU format, except for new format keys that use bcrypt-pbkdf.

Successfully tested on POCL and on most devices on "super" (except for Xeon Phi, where it prints an LLVM error and gets stuck), including wordlist-based cracking of all of our test vectors plus john-samples that fit this revised format's constraints. It correctly refuses to load the bcrypt-pbkdf keys, and correctly loads and cracks all others.


solardiz:
This also touches pem-opencl. I've confirmed that it passes tests.

The code in `ssh_kernel.cl` is intentionally kept closer to the C version, not converted to idiomatic OpenCL except where necessary or where other changes are made anyway. So there are quite some uses of `unsigned int` and `unsigned char`, not changed to `uint` and `uchar`.

solardiz:
In the kernel here, I hard-coded `self_test_running` to 0, but this reduces tests coverage. We should also re-test with it set to 1.

We have no easy and readily available way to get the actual variable from host, right?

magnumripper:
> In the kernel here, I hard-coded `self_test_running` to 0, but this reduces tests coverage. We should also re-test with it set to 1.
> 
> We have no easy and readily available way to get the actual variable from host, right?

Easy peasy, I did that in oldoffice-opencl, transferred/updated in `set_salt()`. It's `cl_mem cl_benchmark` in the host code and `__global uint *benchmark` in the kernel parameters.

solardiz:
> Easy peasy, I did that in oldoffice-opencl, transferred/updated in `set_salt()`. It's `cl_mem cl_benchmark` in the host code and `__global uint *benchmark` in the kernel parameters.

I didn't realize we were doing this already. But it's a separate buffer, transferred separately. Not as easy and lightweight as I'd have liked. Maybe it's better to add a field to the salt struct, so that there's no extra buffer to transfer.

solardiz:
> Maybe it's better to add a field to the salt struct, so that there's no extra buffer to transfer.

OK, this was easy. Done. But now it's failing on super's NVIDIAs (but not AMD, nor CPUs), for the single-DES test vector on the very last ASN.1 check. I'll need to investigate.

solardiz:
> now it's failing on super's NVIDIAs (but not AMD, nor CPUs), for the single-DES test vector on the very last ASN.1 check.

Bug found and fixed. All tests are passing again.

PR #5794, opened 2025-07-01
Update bitlocker2john.py

holly-o:
Change reading metadata block so that it reads the exact block length to avoid reading out of bounds. Also using more dictionary values rather than the hex for easy readability

### DIFF
--- a/run/bitlocker2john.py
+++ b/run/bitlocker2john.py
@@ -127,35 +127,37 @@ def parse_VMK(VMK_data):
     print(f"Protection type: {hex(protection_type)} = {PROTECTION_TYPES.get(protection_type)}")
 
     # only try parse properties if correct protection type
-    if protection_type in [0x2000, 0x800]:
+    protection_type_str = PROTECTION_TYPES.get(protection_type)
+    if protection_type_str in ['VMK protected with password', 'VMK protected with recovery password']:
         current_pos = 28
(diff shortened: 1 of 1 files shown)


solardiz:
Thanks! I've just tested this against `john-samples/BitLocker/BitLockerToGo.raw.gz`, after un-gzipping it of course. The only change in output is that the `Parsing FVE block...` message has moved:

```diff
 [+] Identified volume GUID: 4967D63B-2E29-4AD8-8399-F6A339E3D001 = BitLocker
 [+] FVE metadata info found at offsets ['0x2100000', '0x281c800', '0x2f39000']
 
-Parsing FVE block...
-
 Parsing FVE metadata header...
 Metadata size: 806
 Volume GUID: C5417517-54F7-44B8-9303-B3CABBE86754
 Encryption method: 0x80048004
 
+Parsing FVE block...
+
 Parsing FVE metadata entry...
 Entry size: 82
 Entry type: 0x7 = Computer description
```

Maybe unintentional?

Unfortunately, the rest of our sample images appear to be in various other VM formats needing conversion before the tool can be used, so I didn't bother testing.

For commit messages, I currently wrap lines at 75 (except for the title line, which I generally try to keep at most 72). I see this has a long unwrapped line instead. But I went ahead and merged anyway...

PR #5795, opened 2025-07-02
Fix missing fclose calls to prevent file descriptor leaks

lhywk:
### Describe

This PR fixes several instances of missing `fclose()` calls that could lead to file descriptor leaks.

The affected functions are:

- `undrop()`
- `process_KDBX2_database()`
- `do_convert_wholefile()`

Each function previously returned early in some error paths without properly releasing the opened file handles.

### Expected Behavior

- All opened files should be properly closed before any `return`, regardless of success or failure.
- No file descriptor should be left open after the function returns.

### Actual Behavior

- In several paths (e.g., `fgets()` failure, invalid `datasize`, empty input file), the function returns without calling `fclose()`.
- This causes a resource leak and could eventually lead to **file descriptor exhaustion** if such functions are called repeatedly.

This is a **safe and minimal change** with **no behavioral impact** other than proper resource cleanup.

Thanks for reviewing.

### DIFF
--- a/src/base64_convert.c
+++ b/src/base64_convert.c
@@ -1085,8 +1085,10 @@ static void do_convert_wholefile(char *fname, char *outfname, b64_convert_type i
 	fseek(fp, 0, SEEK_END);
 	in_len = ftell(fp);
 	fseek(fp, 0, SEEK_SET);
-	if (in_len == 0)
+	if (in_len == 0) {
+		fclose(fp);
 		return;
+	}
 	in_str = (char*)mem_calloc(1, in_len+4);
 	if (fread(in_str, 1, in_len, fp) != in_len) {
 		MEM_FREE(in_str);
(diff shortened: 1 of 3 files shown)


solardiz (approved):
The changes look good to me.

One one hand, this is silly given how these programs and functions are used, and how much other sloppy code there is nearby, including even when working with these same input files as seen within patch context here. On the other hand, these are fine changes that could become more relevant if the rest of the code of those tools is ever cleaned up.

I did just make a similarly silly commit fixing the unchecked `fopen` across the entire tree, even where plenty of other issues were seen in patch context, just to be done with that type of error. So let's do it.

solardiz:
@lhywk This is failing one of our CI tests:

```
 src/undrop.c:46: trailing whitespace.
+	if (userfile != stdin) fclose(userfile); 
```

Please remove the trailing whitespace and force-push. Thanks!

lhywk:
> @lhywk This is failing one of our CI tests:
> 
> ```
>  src/undrop.c:46: trailing whitespace.
> +	if (userfile != stdin) fclose(userfile); 
> ```
> 
> Please remove the trailing whitespace and force-push. Thanks!

Trailing whitespace removed and force-pushed. Thanks!

solardiz:
Merged, thanks!

Issue #5802, opened 2025-07-15
Password support - Backdrop CMS


guinny65:
Hi, 
Backdrop CMS hash scheme seems currently unsupported. 
I have coded a John module for it.
The doc says one should post an issue before contributing. Should I propose a PR?

solardiz:
Hi. Thanks. Yes, we'd appreciate a PR, especially considering you already have the code. I don't know what this hash scheme is, so we might end up suggesting supporting it in some other way, please do not be discouraged if this happens.

guinny65:
Okay, I’ve created a PR for you to review and see if it’s worth pursuing.

PR #5805, opened 2025-07-19
Backdrop format support

guinny65:
Add support for Backdrop CMS passwords.

Related to issue #5802

### DIFF
--- a/src/backdrop_fmt_plug.c
+++ b/src/backdrop_fmt_plug.c
@@ -1,15 +1,15 @@
 /*
- * Format for cracking BackDrop CMS database passwords
- * Hash logic in python is like:
- * 
- * count = 2**NITER
- * result = hashlib.sha512(salt + password).digest()
- * while count > 0:
- *     result = hashlib.sha512(result + password).digest()
- *     count -= 1
- * return result
- *
- */
+* Format for cracking BackDrop CMS database passwords
+* Hash logic in python is like:
+*
+* count = 2**NITER
+* result = hashlib.sha512(salt + password).digest()
+* while count > 0:
+*     result = hashlib.sha512(result + password).digest()
+*     count -= 1
+* return result
+*
+*/
 
 #if FMT_EXTERNS_H
 extern struct fmt_main fmt_backdrop;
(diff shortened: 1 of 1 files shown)


solardiz:
Thanks. Can you please squash the commits into one and force-push? We normally merge PRs via rebase, so we only want commits that are valuable to keep in our project's git history.

Please also make sure our CI tests pass.

The `cracked` array logic is to be removed - we use it for "non-hashes" where direct hash comparison is not possible, but here you compute normal hashes.

The indentation looks different from ours. We normally use 8-char tabs.

solardiz:
Thanks. The CircleCI ARM job is still failing:

```
Testing: backdrop, BackDrop CMS [SHA512 32/64]... (2xOMP) .circleci/circle-ci.sh: line 105:  8851 Segmentation fault      (core dumped) $JTR --test=0

Exited with code exit status 139
```

BTW, the `32/64` reporting seen here is wrong.

solardiz:
Upon a closer look, these hashes are the same as Drupal7 except they have the extra `$backdrop$` prefix. Simply deleting that part of the hash (making it start with `$S$`) results in it crackable by our Drupal7 format. So let's not introduce this separate format, but instead let's enhance the Drupal7 format to recognize this prefix in `valid` and remove it in `split`.

solardiz:
The `valid` change and `split` addition should go in `drupal7_common.h`. Then `split` usage should be made in `drupal7_fmt_plug.c` and `ztex_drupal7.c`, replacing `fmt_default_split` with `split`.

And yes, we have this supported on CPU and FPGA, but weirdly not in OpenCL.

solardiz:
> Upon a closer look, these hashes are the same as Drupal7 except they have the extra `$backdrop$` prefix.

Where does this prefix come from? Does Backdrop CMS use it, or is it something you maybe invented for this format?

I tried some quick searches on code in https://github.com/backdrop/backdrop and I only find logic similar to Drupal's original, without the additional prefix.

So are any changes in JtR needed at all?

guinny65:
The original backdrop hashes look like : `$S$GqX57SsAfRin5BjWbWU2fVPyn7g6OXhIQNp8F3r53DItd/DH79rm`.
So you are right, Drupal7 implementation can crack these! 
Sorry for not spotting that.

solardiz:
OK, let's close this PR then.

PR #5818, opened 2025-08-18
OpenCL AES: Correction for supporting two or more AES contexts at once

magnumripper:
This was a brain fart:
```c
#if SHARED_MEM_SIZE < (WARP_SIZE * (256*4 + 256) + 2*4 + 4)
```
You can't do math in a macro. The new fix corrects this but there are limits on how "good" we can make it.

There's also a commit just changing axcrypt-opencl's algorithm name to list AES first.

### DIFF
--- a/run/opencl/diskcryptor_aes_kernel.cl
+++ b/run/opencl/diskcryptor_aes_kernel.cl
@@ -9,18 +9,7 @@
 #define AES_SRC_TYPE MAYBE_CONSTANT
 
 #include "pbkdf2_hmac_sha512_kernel.cl"
-
-/*
- * AES_256_XTS uses two AES keys at once so need double the
- * shared memory.
- */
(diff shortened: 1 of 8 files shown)


magnumripper:
I had to revert this in #5820 because there was speed regression and some formats crashed with too much local memory use. But my re-testing confirms it ends up fine? Strangely enough it also showed that it ends up fine without it. I must be senile, I'll revisit this later.

magnumripper:
> This was a brain fart:
> 
> ```c
> #if SHARED_MEM_SIZE < (WARP_SIZE * (256*4 + 256) + 2*4 + 4)
> ```
> 
> You can't do math in a macro. The new fix corrects this but there are limits on how "good" we can make it.

Actually THAT was a brain fart because you can most definitely write a comparison that way. I'll postpone this anyway - both works but the new is better. But I need to concentrate my testing on fewer things at once.

PR #5823, opened 2025-08-19
OpenCL formats: Drop all _attribute__((max_constant_size(n)))

magnumripper:
It was never in any OpenCL standard. It was deprecated ages ago and is allegedly unsupported on any AMD driver since 2018. But it has produced annoying warnings ever since.

As far as I know it was never supported on Nvidia although they just silently ignore it.


solardiz (approved):
I don't really know. I have to trust you on this (or spend time researching, but I have no time). I wonder what this means for older AMD drivers, like we had for Tahiti? Unfortunately, I no longer have a system to test this on.

magnumripper:
> I wonder what this means for older AMD drivers, like we had for Tahiti?

It will work perfectly fine anyway. At best, this could lead to minor optimizations with the old runtimes.

PR #5824, opened 2025-08-19
Colorize self-test fails

magnumripper:
I need this for my sore eyes.

See also #3511 which is a lot more ambitious.

### DIFF
--- a/run/john.conf
+++ b/run/john.conf
@@ -106,23 +106,35 @@ LogDateStderrFormat =
 # parsed as ESC for use in ANSI codes (like in the default)
 TerminalReset = ^[0m
 
+# If this is true, colorize some messages - such as test failures - if stdout
+# is a terminal.
+# The color codes can be redefined here if needed. The ^ character will be
+# parsed as ESC for use in ANSI escape sequences (like in the defaults here,
+# which define red for errors, green for notices and yellow for warnings).
+# Note that you don't strictly need to use ANSI sequences - other things are
+# fine too.
+# Set this to N or comment it out to disable all color stuff.
+UseColors = Y
+ColorError = ^[0;31m
+ColorNotice = ^[0;32m
+ColorWarning = ^[0;33m
+ColorEnd = ^[0m
+
 # This can be used to colorize (on screen) or otherwise emphasize (in log
-# files) output whenever a supposed administrator password gets cracked.
+# files or if stdout is not a terminal) output whenever a supposed
+# administrator password, or some other password, gets cracked.
+#
+# The colors used will be yellow for regular cracks and red for users that we
+# know or guess are admin users (including but not limited to uid 0 or 500,
+# or login 'root' or 'Administrator').
 #
 # Set this to N or comment it out to disable all "MarkAdmin" stuff.
 MarkAdminCracks = Y
+ColorAdmin = ^[0;31m
+ColorCrack = ^[0;33m
 
-# If MarkAdminCracks = Y above, the below will be used (if defined) for
-# terminal output. The default is to change color to red before the username
-# and reset to normal after it. Any "^" characters will be parsed as ESC for
-# use in ANSI codes (like in the defaults).
-# The "MarkOther" entries will make non-admin stuff brown.
-MarkAdminStart = ^[0;31m
-MarkAdminEnd = ^[0m
-MarkOtherStart = ^[0;33m
-MarkOtherEnd = ^[0m
-
-# If MarkAdminCracks = Y above, the below will be used (if defined) for logs.
+# If MarkAdminCracks = Y above, the below will be used (if defined) for logs
+# or otherwise where stdout is not a terminal.
 # This literal string will be printed after the " + Cracked: root" line.
 MarkAdminString = (ADMIN ACCOUNT)
 
--- a/src/color.h
+++ b/src/color.h
@@ -1,0 +1,43 @@
+/*
+ * This software is Copyright (c) 2025 magnum
+ * and is hereby released to the general public under the following terms:
+ * Redistribution and use in source and binary forms, with or without
+ * modifications, are permitted.
+ */
+
+#ifndef _JOHN_COLOR_H
+#define _JOHN_COLOR_H
+
+extern char *parse_esc(const char *string);
+extern void color_init();
+
+/* Color escape sequences as strings */
+extern char *color_error, *color_notice, *color_warning, *color_end;
+
+#define printf_color(color, ...)	fprintf_color(color, stdout, __VA_ARGS__);
+
+#define puts_color(color, string)	  \
+	do { \
+		fputs_color(color, string, stdout); \
+		putchar('\n'); \
+	} while (0)
+
+#define fprintf_color(color, handle, ...)	  \
+	do { \
+		if (isatty(fileno(handle))) \
+			fputs( (color) , handle); \
+		fprintf(handle, __VA_ARGS__); \
+		if (isatty(fileno(handle))) \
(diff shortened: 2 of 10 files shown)


solardiz:
If we put anything per-color in `john.conf` at all, then maybe we should name those settings not by colors but by their purpose or meaning, like `ColorError` or `ColorWarning`, etc.

solardiz:
And I agree we need to make the specific self-test failures stand out. My other idea was/is to add the failing formats to a list, and then print that list at the end of multi-format test. I already added printing of the count in there, which we could extend in this way.

magnumripper:
> If we put anything per-color in `john.conf` at all, then maybe we should name those settings not by colors but by their purpose or meaning, like `ColorError` or `ColorWarning`, etc.

True, I fixed this now. It's still just a hack of course - I've had plans for doing this globally but I'm not sure I'll ever get the time. Anyway with this version the config is ready for that as well.

magnumripper:
> My other idea was/is to add the failing formats to a list, and then print that list at the end of multi-format test.

When the 83rd OpenCL format crashes, you'll never get that list.

On the other hand, when I use `tee` to get output on screen as well as to a file, these colors aren't printed. For those cases I use to `grep --color=yes FAILED file` which gives me both the colors and a list :)

run/john.conf
```diff
@@ -106,6 +106,15 @@ LogDateStderrFormat =
 # parsed as ESC for use in ANSI codes (like in the default)
 TerminalReset = ^[0m
 
+# If this is true, colorize some errors/warnings - such as test failures.
+# The color codes can be redefined here if needed. Any "^" characters will be
+# parsed as ESC for use in ANSI escape sequences (like in the defaults here,
+# which define red for errors and yellow for warnings.
```
solardiz: Missing closing brace.

run/john.conf
```diff
@@ -106,6 +106,15 @@ LogDateStderrFormat =
 # parsed as ESC for use in ANSI codes (like in the default)
 TerminalReset = ^[0m
 
+# If this is true, colorize some errors/warnings - such as test failures.
```
solardiz: Perhaps we'll later change `errors/warnings` to just `messages`, and also have a color for notices such as cracked passwords (or even a separate `ColorCrack`).

run/john.conf
```diff
@@ -106,6 +106,15 @@ LogDateStderrFormat =
...
+# The color codes can be redefined here if needed. Any "^" characters will be
+# parsed as ESC for use in ANSI escape sequences (like in the defaults here,
+# which define red for errors and yellow for warnings.
+UseColors = Y
+ColorError = ^[0;31m
+ColorWarning = ^[0;33m
```
solardiz, line 120: It's a bit weird to define `ColorWarning` without introducing _any_ use of it soon. Maybe it's better to go over the places where we already use colors and add `Color*` categories for those?

magnumripper:
I'll amend this PR soon-ish. I think I'll move stuff around so eg. the `fail_start` and `fail_end` variables currently local to `benchmark_all()` are instead global variables `color_error` and `color_end` (also `color_warning` of course) to be used anywhere needed.

But this whole thing should be used **very** sparsely now - a simplistic `printf` color alternative is better, even if far from the pipe dreams of #3511.

magnumripper:
OK new version, proof of concept.

run/john.conf
```diff
@@ -106,23 +106,35 @@ LogDateStderrFormat =
 # parsed as ESC for use in ANSI codes (like in the default)
 TerminalReset = ^[0m
 
+# If this is true, colorize some messages - such as test failures - if stdout
+# is not a terminal.
```
solardiz: I guess `is not` should be `is` here.

run/john.conf
```diff
@@ -106,23 +106,35 @@ LogDateStderrFormat =
...
+# The color codes can be redefined here if needed. The ^ character will be
+# parsed as ESC for use in ANSI escape sequences (like in the defaults here,
+# which define red for errors, green for notices and yellow for warnings).
+# Note that you don't strictly need to use ANSI sequences - other things are
+# fine too.
+# Set this to N or comment it out to disable all color stuff.
+UseColors = Y
+ColorError = ^[0;31m
+ColorNotice =  ^[0;32m
```
solardiz: Why two spaces before the `ColorNotice` sequence?

run/john.conf
```diff
@@ -106,23 +106,35 @@ LogDateStderrFormat =
...
+ColorEnd = ^[0m
+
 # This can be used to colorize (on screen) or otherwise emphasize (in log
-# files) output whenever a supposed administrator password gets cracked.
+# files or if stdout is not a terminal) output whenever a supposed
+# administrator password, or some other password, gets cracked.
+#
+# The colors used will be yellow for regular cracks and red for users that we
+# know or guess are admin users (including but not limited to  uid 0 or 500,
+# or login 'root' or 'Administrator').
```
solardiz, line 129: Why two spaces before `uid`?

run/john.conf
```diff
@@ -106,23 +106,35 @@ LogDateStderrFormat =
...
 #
 # Set this to N or comment it out to disable all "MarkAdmin" stuff.
 MarkAdminCracks = Y
+ColorAdmin = ^[0;31m
+ColorCrack =  ^[0;33m
```
solardiz: Why two spaces before the `ColorCrack` sequence?

src/color.h
```diff
@@ -0,0 +1,53 @@
...
+#define fprintf_color(color, handle, ...)	  \
+	do { \
+		if (isatty(fileno(handle))) \
+			fputs( (color) , handle); \
+		fprintf(handle, __VA_ARGS__); \
+		if (isatty(fileno(handle))) \
+			fputs(color_end, handle); \
+	} while (0)
+
+#define fputs_color(color, string, handle)	  \
```
solardiz, line 34: Maybe define 3 of these in terms of `fprintf_color`, passing `"%s"` for the `*puts`.

solardiz (approved):

solardiz (changes requested):
Sorry it was a bit weird to Approve when I also point out one error (`is not` in place of `is`). Changed to "Request changes" now. I do approve of these changes in general, but at least this tiny fix is needed.

src/color.h
magnumripper: Perhaps keep the `fputs` macro as is too, but point `puts`to that? Would an optimizer change fprintf to fputs if there is only one format argument?

src/color.h
```diff
@@ -0,0 +1,43 @@
...
+
+/* Color escape sequences as strings */
+extern char *color_error, *color_notice, *color_warning, *color_end;
+
+#define printf_color(color, ...)	fprintf_color(color, stdout, __VA_ARGS__);
+
+#define puts_color(color, string)	  \
+	do { \
+		fputs_color(color, string, stdout); \
+		puts(""); \
```
solardiz: or `putchar('\n');`

src/color.h
solardiz: I don't mind us doing it either way. I think some optimizers might change `fprintf` to `fputs`, but this is certainly not anything to count upon. It's not a big deal if this just stays as `fprintf` in compiled code. Also, we could even have this as a non-inline function rather than a macro - none of this is very performance-critical. Performance only matters a little bit in terms of number of separate `write` syscalls, so if the stream is somehow set to unbuffered, and actually we'd optimize that by using _one_ `fprintf` for the entire string (including colors and the newline character), with multiple maybe-empty `%s`.

solardiz (approved):

PR #5825, opened 2025-08-20
Self test worksizes

### LINKED ISSUE #5822: Note about OpenCL self-tests
During my work with AES on GPU, I realized this:

Formats using local memory may well have bugs only showing up at LWS > warp/wavefront size (or fixed LWS > 32). This means that `--test` will not find it (the self-test runs with much lower LWS, then the benchmark runs with higher LWS but without checking anything). However `--test --lws=128` may find it, as we do honor the LWS option during self-test.

We could improve that eg. by running self-test with LWS at something like 2 x warp/wavefront size and GWS at some multiple of that, instead of the low LWS=7 GWS=49 we use now. Not sure how much it would slow the self-test down but I guess it wouldn't be very noticable.

magnumripper:
Self-test with default LWS at device maximum

Also with a default GWS of 2x that LWS.  The new figures are better at triggering bugs. If a kernel needs a lower LWS than device max. our code already handles that.

We previously had it as LWS=7 GWS=49 for speed and for checking heuristics that could bug out on non-log2 values but that was introduced before our autotune was sped up with orders of magnitude (and the heuristics has been stable for many years).

Like before, the self-test will obey any given lws/gws options or environment variables.

Closes #5822

### DIFF
--- a/src/opencl_common.c
+++ b/src/opencl_common.c
@@ -1099,22 +1099,21 @@ void opencl_get_user_preferences(const char *format)
 
 void opencl_get_sane_lws_gws_values()
 {
-	if (self_test_running) {
-		local_work_size = 7;
-		global_work_size = 49;
-	}
 
-	if (!local_work_size) {
+	if (!local_work_size || self_test_running) {
 		if (cpu(device_info[gpu_id]))
 			local_work_size =
-				get_platform_vendor_id(platform_id) == DEV_INTEL ?
-			8 : 1;
+				get_platform_vendor_id(platform_id) == DEV_INTEL ? 8 : 1;
+		else if (self_test_running)
+			local_work_size = get_device_max_lws(gpu_id);
 		else
-			local_work_size = 64;
+			local_work_size = 2 * get_device_warp_size(gpu_id);
 	}
 
-	if (!global_work_size)
-		global_work_size = 768;
+	if (self_test_running)
+		global_work_size = local_work_size;
+	else if (!global_work_size)
+		global_work_size = 12 * local_work_size;
 }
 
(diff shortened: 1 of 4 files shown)


magnumripper:
So a build bot choked on office-opencl hanging (10 minutes no output) on its Intel CPU device. I'm not quite sure what to do with that.

Edit: I used the same CPU-specific logic that Claudio added nearby.

magnumripper:
Found two format bugs so far because of this change. Fixes are in this PR but as separate commits.

magnumripper:
Ready for merge. Tested a lot, works fine and finds bugs. The only drawback is a tad slower self-test but it's only significant for things like `--test=@opencl`.

src/opencl_common.c
```diff
@@ -1099,22 +1099,21 @@ void opencl_get_user_preferences(const char *format)
...
-			local_work_size = 64;
+			local_work_size = 2 * get_device_warp_size(gpu_id);
 	}
 
-	if (!global_work_size)
-		global_work_size = 768;
+	if (self_test_running)
+		global_work_size = local_work_size;
+	else if (!global_work_size)
+		global_work_size = 12 * local_work_size;
```
solardiz, line 1116: This looks like it'd sometimes affect more than just self-test. Could be worth mentioning in the commit message?

solardiz (approved):

src/opencl_common.c
magnumripper: I'm not sure what this code affects (not shared autotune afaik) so wouldn't know what to write. Anyway I don't think it's worth mentioning, looks like some default start value or worst case fallback.

PR #5826, opened 2025-08-20
OpenCL AES: Correction for supporting two or more AES contexts at once

magnumripper:
This currently applies to AES-XTS, used by diskcryptor and truecrypt.

Also separate Truecrypt kernel from the (potentially) shared PBKDF2-RIPE- MD160 code for good measure.

Nvidia 550.144.03 bugs out on this for truecrypt so we revert that format to bitsliced AES (it doesn't matter for speed).

### DIFF
--- a/run/opencl/diskcryptor_aes_kernel.cl
+++ b/run/opencl/diskcryptor_aes_kernel.cl
@@ -9,18 +9,7 @@
 #define AES_SRC_TYPE MAYBE_CONSTANT
 
 #include "pbkdf2_hmac_sha512_kernel.cl"
-
-/*
- * AES_256_XTS uses two AES keys at once so need double the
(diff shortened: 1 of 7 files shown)


magnumripper:
This was originally #5818. After that one was merged I paniced after seeing regressions and crashes so I reverted it in #5820 but I later realised that there was nothing wrong with it - I was juggling with too many branches and got confused.

Anyway this code here (even sligthly improved since #5818) is what I've been running with for my tests since - it works fine and is even tested with debug output for verifying that work sizes end up correct vs. memory constraints.

magnumripper:
Tested a lot, works fine, ready for merge.

solardiz (approved):
I skimmed, but I mostly trust your testing.

PR #5829, opened 2025-08-21
Include the session name in exit messages.

magnumripper:
If session name given, the name is enclosed in single quotes such as:
```
Session 'rockyou' aborted
```

I've used this for years in a private branch so it's well tested.

### DIFF
--- a/src/john.c
+++ b/src/john.c
@@ -196,7 +196,6 @@ static struct db_main database;
 static int loaded_extra_pots;
 static struct fmt_main dummy_format;
 
-static char *mode_exit_message = "";
 static int exit_status = 0;
 
 static void john_register_one(struct fmt_main *format)
@@ -1833,7 +1832,7 @@ static void john_run(void)
 			event_pending = event_status = 1;
 
 		if (options.flags & FLG_SINGLE_CHK)
-			mode_exit_message = do_single_crack(&database);
+			do_single_crack(&database);
 		else
 		if (options.flags & FLG_WORDLIST_CHK)
 			do_wordlist_crack(&database, options.wordlist,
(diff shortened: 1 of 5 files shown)


src/john.c
```diff
@@ -1956,14 +1963,11 @@ static void john_done(void)
 		} else if (children_ok) {
 			log_event("Session completed");
 			if (john_main_process) {
-				fprintf(stderr, "Session completed. %s\n", mode_exit_message);
+				fprintf(stderr, "Session%s completed. %s\n", john_session_name, mode_exit_message);
```
solardiz: Let's drop the trailing dot here while we're at it. It's been annoying me all the time. ;-)

src/signals.c
```diff
@@ -177,27 +174,36 @@ static void sig_remove_reload(void)
...
 {
-	char *abort_msg = (aborted_by_timer) ?
-		"Session stopped (max run-time reached)\n" :
-		"Session aborted\n";
-
 	if (!event_abort) return;
 
 	tty_done();
 
+	char *condition = " aborted", *cause = "\n";
```
solardiz: Should be able to use `const char *` here.

solardiz (approved):
I suggest two minor changes, but overall this looks good to me. Thanks!

src/john.c
magnumripper: I presume you mean drop unless there's a mode_exit_message? I also presume remove dot from all of the latter.
```
$ tail single.c 
	rule_ctx = NULL; /* Just for good measure */

	if (initial_num_salts > 1 && status.guess_count && !retest_guessed) {
		if (single_disabled_recursion)
			return "Warning: Disabled SingleRetestGuessed due to deep recursion. Consider running '--loopback --rules=none' next.";
		else
			return "Consider running '--loopback --rules=none' next.";
	} else
		return "";
}
```

magnumripper: > I also presume remove dot from all of the latter.

Probably better: Print the exit message first, then a newline, the "Session completed".

BTW only single.c has one:
```c
		if (options.flags & FLG_SINGLE_CHK)
			mode_exit_message = do_single_crack(&database);
		else
```
The other functions are `void`.

So even better: Print that exit message from single.c and drop that from john.c.

magnumripper: > So even better: Print that exit message from single.c and drop that from john.c.

I went with this.

PR #5830, opened 2025-08-21
Change the behavior of SIGUSR1 to emit a *delayed* status

magnumripper:
If a second SIGUSR1 is received, we promote it to immediate.


solardiz (approved):
OK, if you feel that's the preferred behavior (I'm not sure).

magnumripper:
For MPI, it's the only way to verify proper speed early in a run.

Issue #3842, opened 2019-04-18
gcc -MMD allows to find dependencies for Makefile.in


AlekseyCherepanov:
Modifying `Makefile.in` I found it hard to account nested dependencies when .h file includes more .h files.

@rofl0r pointed out to me that it is possible to ask gcc to list dependencies of .c file onto .h files.

For example, my list of dependencies included all .h files included in .c file:
```
showformats.o:	showformats.c showformats.h loader.h options.h config.h dynamic.h
```

```
$ gcc -E -MMD showformats.c > /dev/null
$ cat showformats.d
showformats.o: showformats.c loader.h params.h arch.h list.h formats.h \
 misc.h jumbo.h omp_autotune.h options.h getopt.h common.h memory.h \
 john_mpi.h config.h dynamic.h simd-intrinsics.h pseudo_intrinsics.h \
 aligned.h simd-intrinsics-load-flags.h
```

So dependencies in `Makefile.in` might be improved.

AlekseyCherepanov:
IIRC I saw another solution for includes from .h files: make fake target (IIRC .m file) for every header file that depends onto the header and onto targets of included header files; .c files depend onto fake targets of header files; files are just `touch`ed to build fake targets. I cannot find any standard description for such approach.

magnumripper:
The gcc-generated list lacks a number of entries, such as `os-autoconf.h`
```
$ gcc -E -MMD john.c >/dev/null
$ cat john.d
john.o: john.c os.h params.h arch.h openssl_local_overrides.h misc.h \
 jumbo.h path.h memory.h list.h tty.h options.h loader.h formats.h \
 omp_autotune.h getopt.h common.h john_mpi.h signals.h idle.h dyna_salt.h \
 logger.h status.h recovery.h config.h bench.h charset.h single.h \
 wordlist.h prince.h inc.h mask.h mkv.h mkvlib.h subsets.h external.h \
 compiler.h batch.h dynamic.h simd-intrinsics.h pseudo_intrinsics.h \
 aligned.h simd-intrinsics-load-flags.h dynamic_compiler.h fake_salts.h \
 listconf.h crc32.h regex.h unicode.h gpu_common.h opencl_common.h \
 john_build_rule.h fmt_externs.h fmt_registers.h

$ ./plugin_deps.pl john.c | fold -s
john.o:	john.c autoconfig.h os.h os-autoconf.h jumbo.h arch.h params.h 
[...]

Issue #4271, opened 2020-06-08
gcc 3 warnings & errors


solardiz:
Not sure if we care, but gcc 3.4.6 on x86_64 gives:

```
yescrypt/yescrypt-opt.c: In function `yescrypt_kdf_body':
yescrypt/yescrypt-opt.c:1192: warning: comparison is always false due to limited range of data type
sha2.c:706:7: warning: no newline at end of file
stribog_fmt_plug.c:472:2: warning: #warning Stribog-256 and Stribog-512 formats require SSE 4.1, formats disabled
memory.c: In function `alloc_region_t':
memory.c:625: warning: dereferencing type-punned pointer will break strict-aliasing rules
loader.c: In function `ldr_fill_user_words':
loader.c:1975: warning: comparison is always true due to limited range of data type
loader.c:1975: warning: comparison is always true due to limited range of data type
loader.c:2002: warning: comparison is always true due to limited range of data type
racf2john.c: In function `main':
racf2john.c:100: warning: 'h1_len' might be used uninitialized in this function
racf2john.c:102: warning: 'h2_len' might be used uninitialized in this function
ztex.c: In function `ztex_scan_new_devices':
ztex.c:444: warning: 'result_open' might be used uninitialized in this function
In file included from dmg2john.c:41:
filevault.h:5: warning: malformed '#pragma pack(push[, id], <n>)' - ignored
filevault.h:73: warning: #pragma pack (pop) encountered without matching #pragma pack (push, <n>)
In file included from wpapcap2john.c:37:
jumbo.h:38:1: warning: "__STDC_FORMAT_MACROS" redefined
wpapcap2john.c:29:1: warning: this is the location of the previous definition
ar: creating aes.a
memory.c: In function `alloc_region_t':
memory.c:625: warning: dereferencing type-punned pointer will break strict-aliasing rules
memory.c: In function `alloc_region_t':
memory.c:625: warning: dereferencing type-punned pointer will break strict-aliasing rules
memory.c: In function `alloc_region_t':
memory.c:625: warning: dereferencing type-punned pointer will break strict-aliasing rules
```

Yet the build succeeds.

gcc 3.3.6:

```
In file included from lzma/LzmaDec.h:7,
                 from 7z_common.h:13,
                 from 7z_common_plug.c:31:
lzma/7zTypes.h:66: warning: redefinition of `Byte'
/usr/include/zconf.h:261: warning: `Byte' previously declared here
In file included from lzma/LzmaDec.h:7,
[...]

PR #5834, opened 2025-08-27
Makefile overhaul

magnumripper:
Add real auto-dependency stuff (gcc's own) and drop the incompetent perl script we had for plugins only. The old support for "makedepend" is kept for non-gcc systems.

Fix bugs in format's plugin stanzas.

Have "make distclean" also clear kernel caches (the ones within our tree only).

Move plugin logic back from configure to Makefile. Having them in configure was a major brainfart. Also enhance it so we can actually move plugins (formats or not) into or out of the source tree and just "make" regardless of their timestamps. I made the logic very quick for the case of nothing changed - blindly running the preprocessor over 470 files takes time so we do not want that unless actually needed.
For the same reason, I did not add the plugins themselves as dependencies for the "format register" stuff in john.c. That would mean very slow handling of repeated edit-make with a specific plugin and it's only needed if we do things like change the format struct name.

See #3566 and #3842.


solardiz:
Thanks. This is a lot. I don't expect to have time to review it in the next 2 weeks.

magnumripper:
Notes to self as well as for your consideration, here's what I have tested:
- Configure + build after `make distclean`.
- Configure with OpenCL disabled + build.

With no re-configure:
- make with no changes.
- make after touching a plugin.
- make after adding a plugin.
- make after removing a plugin.
- make after touching `color.h`.
- make after touching `params.h`.
- make after changing git HEAD.
- <s>make emulating a non-gcc-compatible compiler (auto dependency generation are turned off).</s> (obsolete, see below)
- make after deleting ALL plugin files. I had to add a commit promoting sm3_plug.c to a non-plug because the dynamic format needs it present, but that would have been needed even without this entire PR.
- make after putting all the plugin files back.

Instead of simply disabling auto dependency generation for compilers that MAY be incompatible, we could have the configure script detect if the needed options are supported. We do that for several other things. I'll give that a try and possibly add it as a separate commit.

magnumripper:
> Instead of simply disabling auto dependency generation for compilers that MAY be incompatible, we could have the configure script detect if the needed options are supported. We do that for several other things. I'll give that a try and possibly add it as a separate commit.

Done, trivial, works like a champ.

solardiz:
I still have no time for a proper review, but:

I successfully built from this PR's branch in a fresh Docker container with Alpine Linux (so with musl) as a non-root user by `./configure --without-openssl` and `make -sj2`, after having run as root `apk add make gcc git libc-dev linux-headers`.

> Change sm3_plug.c to a non-plugin
>
> It's needed for the dynamic format. We're supposed to be able to delete
> every single plugin and build should still succeed.

Hmm, but doesn't our CircleCI no-plugs job test that? If so, why is it succeeding before these changes?

magnumripper:
> Hmm, but doesn't our CircleCI no-plugs job test that? If so, why is it succeeding before these changes?

I didn't look now but possibly it only deletes plugin _formats_, not other files. Or maybe there's just an exception but then we should have done this earlier.

Now, when I do delete all plugin sources, there will be two "errors" from building (reason being we preprocess using `gcc -E *_plug.c` resulting in "no input files") . While they do stick out they aren't fatal: The build continues and succeeds. I left that as-is but it can be avoided if we want to. OTOH I think such a build is extremely rare, not sure why anyone would do it except for testing the plugin gearing.

magnumripper:
> This is a lot.

It isn't really, especially if you review commit for commit. But take your time.

On another note, I realise we could make the whole dynamic format beast plugin(s) now without much effort (there are lots of `#if DYNAMIC_DISABLED` left from a period when it was defunct and Makefile could define that unless the plugs are present). Could be a separate PR though, if we want it at all.

magnumripper:
Oh BTW I also tried with gcc 3.3.6 but that fails for other reasons than this PR.
```
argon2_core_plug.c: In function `argon2_validate_inputs':
argon2_core_plug.c:478: warning: comparison is always false due to limited range of data type
blake2b_plug.c: In function `blake2b_compress':
blake2b_plug.c:197: warning: dereferencing type-punned pointer will break strict-aliasing rules
gost3411-2012-core_plug.c: In function `g':
gost3411-2012-core_plug.c:126: warning: implicit declaration of function `_mm_cvtsi64_si128'
gost3411-2012-core_plug.c:126: error: incompatible types in assignment
```

EDIT: See #4271

Maybe we should fix that but I'm not sure I'm interested myself. Also, according to ChatGPT gcc has supported the flags needed for dependency generation since 3.0 but it may well be hallucinating.

magnumripper:
Added another commit, for building particular targets (such as unit-tests, found by a buildbot) from a distclean source directory (i.e. with `.deps/` missing).

magnumripper:
Added another little commit that just moves @DEPFLAGS@ later on the command line (they are among the least interesting things to see) and drops the redundant CFLAGSX. Not sure why we ever had the latter, it was just confusing.

Rebased on current bleeding while at it.

magnumripper:
> > This is a lot.
> 
> It isn't really, especially if you review commit for commit.

@solardiz with 2 months lead time for reviewing a PR that doesn't even change any functionality in john, my incentive for contributing plummets. We could merge this and if you find some flaw later we can fix it then.

solardiz:
@magnumripper I'm really sorry. Yes, you may just merge this. I also hope to take a look soon, whether before or after merging. Thank you!

src/configure.ac
```diff
@@ -227,6 +226,9 @@ if test "x$ubsantrap" = xyes ; then
    LDFLAGS=${LDFLAGS_BAK}
 fi
 
+   dnl Enable auto dependency generation if supported by the compiler
+   JTR_FLAG_CHECK([-MMD -MP -MF /tmp/.dep], 1)
+   AC_SUBST([DEPFLAGS],['-MMD -MP -MF $(DEPDIR)/$(notdir $*).d'])
```
solardiz: I think the usage of `/tmp/.dep` in this test is unacceptable - this is basically a vulnerability. Can we use `/dev/null` or a file in the current directory?

Regardless, maybe we should test not only that the flags don't produce an error, but that output is sane?

solardiz (changes requested):
I went through the commits one by one. I like how we're dropping the old Perl-based stuff and over 400 lines. There are many things here I didn't review thoroughly, or would need to read up on GNU make to review fully. But if this works, it works. The only blocker I think is the one misuse of `/tmp`, which we must avoid.

solardiz:
> Oh BTW I also tried with gcc 3.3.6 but that fails for other reasons than this PR.

I actually wonder if we're maybe introducing a dependency on a newer version of GNU Make than we required before.

src/configure.ac
solardiz: Maybe we should revert to or add a check for gcc/clang? Like options are not rejected _and_ the compiler is gcc/clang.

magnumripper: That use of /tmp was embarassing - I now use conftest.d (in cwd, as in src). Somehow it's gone after the tests without me deleting it. Not sure if I should add a `rm -f conftest.d` anyway?

> Regardless, maybe we should test not only that the flags don't produce an error, but that output is sane?

Done. Our "flag check" verifies that unsupported options do fail though, and if not we'll neither test nor blindly use these options.

magnumripper: > Somehow it's gone after the tests without me deleting it. Not sure if I should add a `rm -f conftest.d` anyway?

I see now the configure script takes care of it due to its name. All build tests in autoconf use the name conftest.c and in the end it deletes conftest.* among other known build artefacts. So we're fine as-is.

PR #5838, opened 2025-09-01
Convert NOTE to lowercase for `stderr` messages

gartikis:
@solardiz mentioned in [this PR](https://github.com/openwall/john/pull/5837#issuecomment-3240327509):

> Actually, it is "Note" in many other messages, and "NOTE" in only 3 other messages (but one of them is in zip2john). So we may want to get the rest of them changed to "Note".

So this PR changes the other 3 instances where NOTE is seen in all caps to lowercase for `stderr` messages.


magnumripper (approved):
Thanks!

solardiz (approved):

PR #5840, opened 2025-09-01
OpenCL SHA-512 "free" formats: Fix double race conditions

magnumripper:
Found when reviewing all uses of barriers in our kernels.

I recall we've had problems with intermittent failures from these two formats. The global barrier was added by me in good faith at some point in time. It did help a little but still not safe. Before that, there wasn't even the barrier...

I believe all other kernels' use of barriers are just fine.

### DIFF
--- a/run/opencl/sha512_kernel.cl
+++ b/run/opencl/sha512_kernel.cl
@@ -113,15 +113,10 @@ __kernel void kernel_sha512(
 __kernel void kernel_cmp(
 	__constant uint64_t *binary,
 	__global uint64_t *hash,
-	__global uint32_t *result)
+	__global volatile uint32_t *result)
 {
 	uint32_t idx = get_global_id(0);
 
-	if (idx == 0)
-		*result = 0;
-
-	barrier(CLK_GLOBAL_MEM_FENCE);
-
 	if (*binary == hash[idx])
-		*result = 1;
+		atomic_or(result, 1);
 }
--- a/run/opencl/xsha512_kernel.cl
+++ b/run/opencl/xsha512_kernel.cl
@@ -119,15 +119,10 @@ __kernel void kernel_xsha512(
 __kernel void kernel_cmp(
 	__constant uint64_t *binary,
 	__global uint64_t *hash,
-	__global uint32_t *result)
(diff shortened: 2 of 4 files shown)


solardiz:
Looks like these specific formats have just one `cmp_all`-alike flag, rather than anything per computed hash. If so, should we maybe prefer to reset this flag in `cmp_all` rather than in `crypt_all`? Otherwise (as with this PR's changes), `cmp_all` probably just keeps returning true if there are matches for any of the loaded hashes, whereas it could only return true if called for loaded hashes that match any of the computed hashes.

Edit: OTOH, clearing the flag just once per all `cmp_all` calls is probably quicker when there are no successful cracks by the entire `crypt_all` call.

magnumripper:
> Looks like these specific formats have just one `cmp_all`-alike flag, rather than anything per computed hash.

Yes and that's also the reason for race conditions. Looking at it today I can't understand why it was made this way, but it was very early in our steps towards mastering GPU formats.

> If so, should we maybe prefer to reset this flag in `cmp_all` rather than in `crypt_all`? Otherwise (as with this PR's changes), `cmp_all` probably just keeps returning true if there are matches for any of the loaded hashes, whereas it could only return true if called for loaded hashes that match any of the computed hashes.

How would cmp_all know there aren't more matching binaries?

solardiz:
> How would cmp_all know there aren't more matching binaries?

It wouldn't, but it doesn't need to since it's called again for each loaded hash. That's when bitmap+hash table are not in use (for the current salt if applicable) - when they are, `cmp_all` is skipped.

magnumripper:
Oh, I see what you mean now. In my mind the cmp kernel was called from within crypt_all. The result should obviously be cleared prior to calling the cmp kernel, so in cmp_all. I'll fix that.

solardiz (approved):

solardiz:
BTW, why do you also add `volatile`? Does it matter?

magnumripper:
> BTW, why do you also add `volatile`? Does it matter?

It's required for atomic functions. IIRC there's not always a warning if you forget, and it may even (intermittently) work just fine although perhaps unsafe so could possibly be a bug really hard to spot.

PR #5852, opened 2025-09-15
zip2john should sanitize best file name before printing it

ganzm:
For zip2john, if there is only one candiate file, it is being prepeanded to the hash. This file name should also be sanitized and stripped off ':'

### DIFF
--- a/src/zip2john.c
+++ b/src/zip2john.c
@@ -924,10 +924,18 @@ static void print_and_cleanup(zip_context *ctx)
 	filenames = xstrdup(ctx->best_files[0].file_name);
 	bname = jtr_basename(ctx->archive.fname);
 
-	printf("%s%s%s:$pkzip$%x*%x*", bname,
-			 ctx->num_candidates == 1 ? "/" : "",
-			 ctx->num_candidates == 1 ? ctx->best_files[0].file_name : "",
-			 ctx->num_candidates, ctx->archive.check_bytes);
+	if (ctx->num_candidates == 1) {
+		/* sanitize here, as we don't need the raw file_name any more when exiting this function */
+		replace(ctx->best_files[0].file_name, ':', ' ');
+
+		printf("%s/%s:$pkzip$%x*%x*", bname,
+						ctx->best_files[0].file_name,
+						ctx->num_candidates,
+						ctx->archive.check_bytes);
+	} else {
+		printf("%s:$pkzip$%x*%x*", bname, ctx->num_candidates, ctx->archive.check_bytes);
+	}
+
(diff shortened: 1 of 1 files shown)


src/zip2john.c
```diff
@@ -924,10 +924,18 @@ static void print_and_cleanup(zip_context *ctx)
 	filenames = xstrdup(ctx->best_files[0].file_name);
 	bname = jtr_basename(ctx->archive.fname);
 
-	printf("%s%s%s:$pkzip$%x*%x*", bname,
-			 ctx->num_candidates == 1 ? "/" : "",
-			 ctx->num_candidates == 1 ? ctx->best_files[0].file_name : "",
-			 ctx->num_candidates, ctx->archive.check_bytes);
+	if(ctx->num_candidates == 1) {
+		/* sanitize here, as we don't need the raw file_name any more when exiting this function */
+		replace(ctx->best_files[0].file_name, ':', ' ');
```
magnumripper, line 929: Just a comment (I haven't tested this): I see we already replace with a space in line 957 in current code but IMHO we should replace with some non-space character such as '_' in both places.

solardiz:
@ganzm Thank you! I see you keep this PR as draft, does this mean you intend to update it further and switch it to ready for review before we consider it?

src/zip2john.c
solardiz: @magnumripper I think we should be consistent, and we already replace with spaces in other 2john tools. So let's do the same here for now. If we want to switch to another replacement character, we'll do so with a separate PR later. We also need to consider what will happen in `ldr_split_string`.

src/zip2john.c
```diff
@@ -924,10 +924,18 @@ static void print_and_cleanup(zip_context *ctx)
 	filenames = xstrdup(ctx->best_files[0].file_name);
 	bname = jtr_basename(ctx->archive.fname);
 
-	printf("%s%s%s:$pkzip$%x*%x*", bname,
-			 ctx->num_candidates == 1 ? "/" : "",
-			 ctx->num_candidates == 1 ? ctx->best_files[0].file_name : "",
-			 ctx->num_candidates, ctx->archive.check_bytes);
+	if(ctx->num_candidates == 1) {
```
solardiz: Nitpick: we put a space between `if` and `(`.

solardiz (approved):

solardiz:
For the commit message, we'd prefix the title line with component name, so `zip2john: Sanitize best file name before printing it`, and wrap the description lines at 75.

solardiz:
Thank you for your contribution @ganzm!

PR #5853, opened 2025-09-22
--show=formats: Fix broken json in case of no net output

magnumripper:
This (example) produced a single "]":
```
touch empty && ./john --show=formats empty
```
It now emits "[]" instead, which should be the Right Thing[tm].

### DIFF
--- a/src/john.c
+++ b/src/john.c
@@ -1094,7 +1094,10 @@ static void john_load(void)
 			options.loader.flags |= DB_CRACKED;
 			ldr_init_database(&database, &options.loader);
 
-			if (!options.loader.showformats) {
+			if (options.loader.showformats) {
+				if (!options.loader.showformats_old)
+					fputs("[", stdout);
+			} else {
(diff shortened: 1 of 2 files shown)


src/john.c
```diff
@@ -1094,7 +1094,10 @@ static void john_load(void)
 			options.loader.flags |= DB_CRACKED;
 			ldr_init_database(&database, &options.loader);
 
-			if (!options.loader.showformats) {
+			if (options.loader.showformats) {
+				if (!options.loader.showformats_old)
+					fputs("[", stdout);
```
solardiz, line 1099: FWIW, this could be `putchar`.

magnumripper: True. I got the idea to look for similar cases and the only other one is a single place in uaf_hash.c. I could fix them with almost no effort if you like but it doesn't matter at all, right?

solardiz: This is unimportant indeed. Up to you. We have plenty of more important code cleanups to make anyway.

Issue #5403, opened 2023-12-29
Add/test detection of RTX 4000 series cards


solardiz:
We should probably add a check to `get_compute_capability()` and maybe to `get_processors_count()` in `opencl_common.c`.

solardiz:
Upon a closer look, that code in`get_compute_capability()` may be dead anyway? We check for `if (!major) {` before reaching the "Apple, VCL and some other environments" fallback code that checks device names. However, `major` is a pointer that is supposed to be non-NULL, or if it happens to actually be NULL then we'd crash on trying to assign values to `*major`. If this was broken for years, should we drop it? Also, it would be cleaner for that function itself to initialize `*major` and `*minor` to 0 rather than rely on each caller having done so (and simplify the callers).

We should still verify the logic of `get_processors_count()` for 4000 series cards - maybe it's correct as-is (the `major >= 7` path does what's needed), or maybe we need to add a `major` check.

solardiz:
> Upon a closer look, that code in`get_compute_capability()` may be dead anyway? We check for `if (!major) {` before reaching the "Apple, VCL and some other environments" fallback code that checks device names. However, `major` is a pointer that is supposed to be non-NULL

Looks like a bug introduced in 2021 with 8fa21b5a913d9013ed5fcc45706afbd322ac6033, where @magnumripper added detection of a few more card series (IIRC on my suggestion), but apparently broke this all (and I did not catch that in my review). So now we need to decide between fixing that bug vs. dropping this legacy code.

akirayamaoka32:
Hi, what do i end up doing with my 4060 ti? I have tried adding a value of [Options:OpenCL] Device = 1/2. But john keeps using my CPU instead of GPU. And when adding --devices I get the following: 
Refuse to run with the “--device” option when the format is neither OpenCL nor ZTEX

ghost:
I'm afraid this issue is not related to the problem you are facing now.

> Hi, what do i end up doing with my 4060 ti? I have tried adding a value of [Options:OpenCL] Device = 1/2. But john keeps using my CPU instead of GPU. And when adding --devices I get the following: Refuse to run with the “--device” option when the format is neither OpenCL nor ZTEX

**Does the format you are using exist in OpenCL?** If so.

Could you open a new issue and add the command-line you are using and john's output. In order to see the message you are seeing, you are using a non-OpenCL format. E.g.

```
$ john ~/allTests.in --format=sha512crypt --device=1
Using default input encoding: UTF-8
[...]

PR #5855, opened 2025-09-25
OpenCL: Fix a bug muting nvidia Compute Capability from device list

magnumripper:
In --list=opencl-devices, it would erroneously mute when minor == 0, so eg. 12.0 didn't show.

Parameters were still passed in build as -DSM_MAJOR=12 -DSM_MINOR=0.

### DIFF
--- a/src/opencl_common.c
+++ b/src/opencl_common.c
@@ -2576,17 +2576,19 @@ size_t get_kernel_preferred_multiple(int sequential_id, cl_kernel crypt_kernel)
 void get_compute_capability(int sequential_id, unsigned int *major,
                             unsigned int *minor)
 {
-	clGetDeviceInfo(devices[sequential_id],
-	                CL_DEVICE_COMPUTE_CAPABILITY_MAJOR_NV,
-	                sizeof(cl_uint), major, NULL);
-	clGetDeviceInfo(devices[sequential_id],
-	                CL_DEVICE_COMPUTE_CAPABILITY_MINOR_NV,
-	                sizeof(cl_uint), minor, NULL);
-
-	if (!*major) {
+	ret_code = clGetDeviceInfo(devices[sequential_id],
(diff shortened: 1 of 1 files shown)


magnumripper:
Also added a commit that amends the heuristics for Compute Capability when unknown. Most importantly it adds 40xx Ada and 50xx Blackwell series.

solardiz:
I don't mind, but see also #5403 - apparently, this was dead code for a long while and no one noticed, so do we really need it?

magnumripper:
Thanks, I fixed that issue too now. It went unnoticed because it would only happen on Apple/VCL and other weird platforms, and even then the core count SP multiplier would default to 32 - but that is not always correct. The consequence would be (except for incorrect Core count and Speed index shown in list) that the internal mask target multiplier could end up incorrect.

solardiz:
> I fixed that issue too now.

Huh? I think I had un-dead the code before, in b53e442618d88dc8158c90772af9ae29db85768c.

solardiz (approved):

solardiz:
> > I fixed that issue too now.
> 
> Huh? I think I had un-dead the code before, in [b53e442](https://github.com/openwall/john/commit/b53e442618d88dc8158c90772af9ae29db85768c).

... but your reference to fixing the issue from that larger commit as a whole looks correct to me anyway. So approved.

magnumripper:
> > > I fixed that issue too now.
> > 
> > 
> > Huh? I think I had un-dead the code before, in [b53e442](https://github.com/openwall/john/commit/b53e442618d88dc8158c90772af9ae29db85768c).
> 
> ... but your reference to fixing the issue from that larger commit as a whole looks correct to me anyway. So approved.

I missed that (it was closed as "not planned") but anyway the test for error is more correct. Who knows what a driver not supporting the call could place in `*major` :)

Issue #5860, opened 2025-10-06
Error Read File when using rar2john tool to extract hash , the rar file was built on RAR4.X


mmasoud:
I have installed the latest jumbo release and ran the command rar2john to hash a 13 years old winrar file built on RAR 4.X version , and the result is 
" Error: read failed: No error." , anyone can advise what's the issue ?

<img width="375" height="457" alt="Image" src="https://github.com/user-attachments/assets/37c35ba2-45bd-40fb-89d5-2fa66e27552a" />

solardiz:
> I have installed the latest jumbo release

Please clarify what exactly you mean by this - what archive filename?

If you're on Windows, we suggest you try latest build winX64_1_JtR.7z from https://github.com/openwall/john-packages/releases

mmasoud:
> > I have installed the latest jumbo release
> 
> Please clarify what exactly you mean by this - what archive filename?
> 
> If you're on Windows, we suggest you try latest build winX64_1_JtR.7z from https://github.com/openwall/john-packages/releases

Hi Solardiz,
I mean password protected .rar file I am trying to recover its password using rar2john command to get the hash , hope its clear now, and yes I am using the latest built.

solardiz:
@mmasoud I am just trying to make sure you are _not_ using the latest release off the John the Ripper homepage, which is unfortunately over 6 years old by now. Can you show the output of `john --list=build-info`?

That said, it is quite likely that you found a shortcoming/bug in the `rar2john` program, and it is quite likely that this bug is still unfixed (although some issues in `rar2john` have been fixed since the latest release). We just want it confirmed against a recent build before we spend more time on it.

Are you able to share with us a sample RAR archive (hopefully a much smaller one), with no sensitive info in it and preferably with a known password, which would be usable to reproduce the issue?

Thank you!

mmasoud:
Thanks for the help @solardiz really appreciate it , for sharing a sample RAR file unfortunately I am unable to as they are my wedding pictures 17 years back :) , I tried to create one on that same old version of winrar but I could not find this old version of Winrar , for the build info you can find it below :

Version: 1.9.0-jumbo-1+bleeding-126b2a4814 2025-01-28 23:51:55 +0100
Build: cygwin 64-bit x86_64 AVX2 AC OMP OPENCL
SIMD: AVX2, interleaving: MD4:3 MD5:3 SHA1:1 SHA256:1 SHA512:1
AES hardware acceleration: AES-NI
CPU tests: AVX2
[...]

PR #5862, opened 2025-10-10
rar2john: Distinguish unexpected EOF from other errors

magnumripper:
Check for feof() before parsing errno. Similar fixes should probably be made to a lot of other files throughout our tree.

See #5860


solardiz (approved):

magnumripper:
Added same fixes for keepass2john as well. I expected to find more problems but I only searched for `strerror(errno)` - apparently we often basically just say "failed" which is correct as-is (although could be amended with this ternary for a better description of the problem).

PR #5864, opened 2025-10-12
Fix a "format order" regression from the autodep branch

### LINKED ISSUE #5863: Format order regression after merging the autodep branch
I just noticed the format list sorting is screwed up now - or rather it's "sorted too literally": The OpenCL formats now register between OpenBSD-Softraid and OpenSSL. The worst result of this is we sometimes pick an OpenCL format before the CPU version. One example is rar-opencl.

I failed to notice because only half of the formats come "too late" now. I'm very puzzled this happened - it was definitely not a problem a while ago in that topic branch.

Investigating.

magnumripper:
We're sorting the format list in struct name order, but we want the OpenCL formats last in the list.

Closes #5863

### DIFF
--- a/src/Makefile.in
+++ b/src/Makefile.in
@@ -257,10 +257,11 @@ version.h: version.h.new
 
 # These are SLOW nowadays, lots of plugins.
 fmt_externs.h: .plugin_fmt_list
-	$(CC) -DFMT_EXTERNS_H -E -P $(CFLAGS) $(PLUGFORMATS_SRCS) | grep -F "extern struct fmt_main" | $(SORT) -f > $@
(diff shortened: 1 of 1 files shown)


solardiz:
Thank you for noticing this issue and fixing it.

Why exactly do you also add `|| true` to some commands here?

magnumripper:
> Why exactly do you also add `|| true` to some commands here?

It's for the academical case of no plugins present. The cpp commands will error on "no input files" and without the `|| true` the whole build will fail. Somehow as mentioned in https://github.com/openwall/john/pull/5834#issuecomment-3241143105 this did not happen before these changes.

PR #5865, opened 2025-10-13
kwallet2john.py: compatibility with python3 and python2

### LINKED ISSUE #5811: kwallet2john.py is no longer compatible with Python 3.10+
The run/kwallet2john.py script currently does not run on Python 3.10+ due to usage of deprecated/removed Python features. Specifically:
- **Hex encoding of salt:** The script uses `salt.encode("hex")`, which was removed in Python 3. The correct approach in Python 3 is `salt.hex()`.
- **String/bytes comparison:** The script defines `KWMAGIC` as a string (`"KWALLET\n\r\0\r\n"`) and compares it to bytes read from a file. This will always fail in Python 3, as file reads return bytes. `KWMAGIC` should be a bytes literal (`b"KWALLET\n\r\0\r\n"`).

These issues prevent the script from running successfully on modern Python versions.

I didn't check if this issue applies to other `run/*.py` scripts, but perhaps this is a more general issue.

I'm unsure of the best path forward:
- It may make sense to retain this older code for use with Python <3.10 and make an updated copy of the script for use with Python 3.10+
- Perhaps there is some way to make the script compatible with Python 3.10+ while retaining support for older versions, like one could do with precompiler conditionals (`#ifdef`s) in C?  I'm unsure if such a feature exists in Python though.

I found the following references useful in diagnosing this issue:
- https://stackoverflow.com/questions/2340319/python-3-string-to-hex
- https://www.geeksforgeeks.org/python/python-convert-bytearray-to-hexadecimal-string/

exploide:
I made the script `kwallet2john.py` compatible with both, python3 and python2. Prior to this change, the script was only working with python2.
This fixes #5811.

The old version of the script with python2, the new version of the script with python2 and the new version with python3 all produce the same output when processing the `*.kwl` files from john-samples.

There is a KWallet format which stores a salt in a dedicated `.salt` file and the script was already able to process such files.
However, I found that there is no sample file to test the respective code branch. Therefore, I created a new wallet file in that format. See https://github.com/openwall/john-samples/pull/41

Unfortunately, john is **not able to crack** the salted hash and keeps running indefinitely.
I suspect there is a bug, someone should probably take a look.

### DIFF
--- a/run/kwallet2john.py
+++ b/run/kwallet2john.py
@@ -15,7 +15,7 @@ import os
 import struct
 from binascii import hexlify
 
-KWMAGIC = "KWALLET\n\r\0\r\n"
+KWMAGIC = b"KWALLET\n\r\0\r\n"
 KWMAGIC_LEN = 12
 KWALLET_VERSION_MAJOR = 0
 KWALLET_VERSION_MINOR = 0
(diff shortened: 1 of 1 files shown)


solardiz:
Thank you very much @exploide! I'll take a look and probably merge these shortly. Also it sounds like we'll need to create an issue for the unable-to-crack bug you found.

solardiz (approved):

PR #5872, opened 2025-10-30
Replace assert() with proper error handling in racf2john.c

elcruzo:
As suggested by @solardiz in PR review, assert() is inappropriate for runtime file I/O errors because:
- assert() is compiled out in release builds (NDEBUG)
- File read errors should be handled gracefully, not cause crashes
- Other nearby errors properly use fprintf() + cleanup + exit()

Changes:
- Replaced assert(count == 1) with proper error checking
- Added cleanup (MEM_FREE) before goto cleanup
- Removed unused <assert.h> include
- Matches error handling pattern used elsewhere in the function

### DIFF
--- a/src/racf2john.c
+++ b/src/racf2john.c
@@ -34,7 +34,6 @@
 
 #include <stdio.h>
 #include <string.h>
-#include <assert.h>
 #include <stdlib.h>
 // jumbo.h needs to be above sys/types.h and sys/stat.h for mingw, if -std=c99 used.
 #include "jumbo.h"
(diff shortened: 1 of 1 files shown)


solardiz (changes requested):
We already have the `MEM_FREE` after the `cleanup` label, so adding it here is unneeded - please exclude this one line, amend the commit, and force-push. Thank you!

solardiz (approved):

solardiz:
Thanks. For further contributions, please also note that we tend to prefix commit title lines with component names, so e.g. "racf2john: Replace assert() with proper error handling". But I'll merge this one as-is.

Issue #5873, opened 2025-10-30
Increase maximum supported password length


solardiz:
Our maximum supported password length for any format is currently 125 (with some formats having lower limits for varying reasons). This may be insufficient for cryptocurrency wallet formats using lengthy generated seed phrases (sometimes along with a user-chosen portion), where most of the phrase would be known and only a small portion needing recovery (for full phrase recovery to be feasible).

hashcat's maximum is 256. And it's actually 256, not 255 - the code was adjusted to match what was previously announced: https://github.com/hashcat/hashcat/commit/5da1e4b87266dd893d334ae27852c0c29d802b9d

Choosing exactly 256 as our maximum feels suboptimal as it means buffers of at least 257 where we need to include the trailing NUL. Yet we could choose this anyway to be same as hashcat with respect to this maximum. We could also choose 259 and buffers of size 260, or even 263/264, since 257 would probably occupy at least this much anyway.

We also need to recall why it ended up 125 rather than 127 now. Do we possibly use this extra room somewhere?

We also need to consider performance impact. Maybe arrays of the larger buffers would cause worse cache utilization, perhaps through impact on cache aliasing and thus requiring greater cache associativity for the same size to fit. Maybe a size that isn't close to a power of 2 would perform better, like IIRC some fast hash formats were previously found to perform best at 95 on older CPUs with 32-byte cache lines. Since 191/192 is too low (too little of an increase and way below hashcat's), maybe consider 319/320?

We still have this in `dummy.c`:

```c
/* Max 125, but 95 typically produces fewer L1 data cache tag collisions */
#define PLAINTEXT_LENGTH                95
#define MAX_PLAINTEXT_LENGTH            (PLAINTEXT_BUFFER_SIZE - 3) // 125
```

solardiz:
> We also need to recall why it ended up 125 rather than 127 now. Do we possibly use this extra room somewhere?

I found no intentional uses of the extra room, but I found that we adjust or check for it in a few places:

```
dummy.c:#define MAX_PLAINTEXT_LENGTH		(PLAINTEXT_BUFFER_SIZE - 3) // 125
formats.c:	    format->params.plaintext_length > PLAINTEXT_BUFFER_SIZE - 3)
options.c:		options.length = PLAINTEXT_BUFFER_SIZE - 3;
options.c:	if (options.length < 1 || options.length > PLAINTEXT_BUFFER_SIZE - 3) {
[...]

PR #5884, opened 2025-11-10
Maximum password length increase

solardiz:
Maximum password length increase, see #5873

This is WIP, currently puzzlingly fails with:

```
Testing: PDF, PDF encrypted document [MD5-RC4 / SHA2-AES 32/64]... FAILED (Long test vector ciphertext but no FMT_HUGE_INPUT)
```

Does max supported plaintext length somehow increase test ciphertext lengths for this format, @magnumripper?


solardiz:
> This is WIP, currently puzzlingly fails with:

Oh, I see:

```c
/*
 * Max. ciphertext size that's sure to fit a line when cleartext field
 * is added.
 */
#define MAX_CIPHERTEXT_SIZE     (LINE_BUFFER_SIZE - PLAINTEXT_BUFFER_SIZE)
```

So should we add `FMT_HUGE_INPUT` for this one format or/and increase `LINE_BUFFER_SIZE` a little bit?

solardiz:
The CircleCI encoding-opencl job reported two failures of OpenCL formats. There may be more if we test them all (we should).

```
-- Running $JTR --test-full=0 --format=opencl '--encoding=cp737' --
Device 1: Intel(R) Xeon(R) Platinum 8124M CPU @ 3.00GHz
Testing: diskcryptor-opencl, DiskCryptor [PBKDF2-SHA512 OpenCL, CP737]... (2xOMP) PASS
Testing: diskcryptor-aes-opencl, DiskCryptor AES XTS (only) [PBKDF2-SHA512 AES OpenCL, CP737]... PASS
Testing: krb5pa-md5-opencl, Kerberos 5 AS-REQ Pre-Auth etype 23 [MD4 HMAC-MD5 RC4 OpenCL, CP737]... PASS
Testing: krb5tgs-opencl, Kerberos 5 TGS-REP etype 23 [MD4 HMAC-MD5 RC4 OpenCL, CP737]... PASS
Testing: mscash-opencl, M$ Cache Hash [MD4 OpenCL, CP737]... PASS
Testing: mscash2-opencl, MS Cache Hash 2 (DCC2) [PBKDF2-SHA1 OpenCL, CP737]... FAILED (cmp_all(448))
Testing: NT-opencl [MD4 OpenCL, CP737]... PASS
Testing: ntlmv2-opencl, NTLMv2 C/R [MD4 HMAC-MD5 OpenCL, CP737]... PASS
Testing: NT-long-opencl [MD4 OpenCL, CP737]... -
getkey = 90123456789012345678901234567890123456789012345678901234567890
setkey = 808080809012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678
FAILED (max. length in index 0: wrote 319, got 63 back)
*** Error in `../run/john': free(): invalid next size (normal): 0x0000000032dc7580 ***
```

magnumripper:
> > This is WIP, currently puzzlingly fails with:
> 
> Oh, I see:
> 
> ```c
> /*
>  * Max. ciphertext size that's sure to fit a line when cleartext field
>  * is added.
>  */
> #define MAX_CIPHERTEXT_SIZE     (LINE_BUFFER_SIZE - PLAINTEXT_BUFFER_SIZE)
> ```
> 
> So should we add `FMT_HUGE_INPUT` for this one format or/and increase `LINE_BUFFER_SIZE` a little bit?

Yeah, I guess. This issue points us to a side effect of this PR: Future "truncated hashes" in the pot file may be of a different length than current ones (unless we carefully bump PLAINTEXT_BUFFER_SIZE to no net change of MAX_CIPHERTEXT_SIZE). I think I recall we designed the code to cope with such change (still recognizing older truncated hashes). Or tried to. Also, everything I just wrote is just off the top of my head.

Bottom line is we need to test cracking samples that ends up as truncated pot entries using current bleeding, then (for example) use `--show` with this PR applied. Perhaps try the other way around as well.

solardiz:
> This issue points us to a side effect of this PR: Future "truncated hashes" in the pot file may be of a different length than current ones (unless we carefully bump PLAINTEXT_BUFFER_SIZE to no net change of MAX_CIPHERTEXT_SIZE).

I didn't realize we had such issue/risk, but I happened to adjust this PR for "no net change of MAX_CIPHERTEXT_SIZE":

```c
#define LINE_BUFFER_SIZE		(0x380 + PLAINTEXT_BUFFER_SIZE)
```

solardiz:
> Testing: mscash2-opencl, MS Cache Hash 2 (DCC2) [PBKDF2-SHA1 OpenCL, CP737]... FAILED (cmp_all(448))
> Testing: NT-opencl [MD4 OpenCL, CP737]... PASS
> Testing: ntlmv2-opencl, NTLMv2 C/R [MD4 HMAC-MD5 OpenCL, CP737]... PASS
> Testing: NT-long-opencl [MD4 OpenCL, CP737]... -
> getkey = 90123456789012345678901234567890123456789012345678901234567890
> setkey = 808080809012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678
> FAILED (max. length in index 0: wrote 319, got 63 back)
> *** Error in `../run/john': free(): invalid next size (normal): 0x0000000032dc7580 ***

Looks like mscash2-opencl and NT-long-opencl assume length up to 127 - there are places where `& 127` is used, perhaps to use higher bits for something else? I reverted them to max 125 for now because there may also need to be adjustments for utf8.

solardiz:
Testing with POCL, looks like I also broke this:

```
Testing: keystore-opencl, Java KeyStore [SHA1 OpenCL]... -
getkey = �90123456789012345678901234567890123456789012345678901234567890
setkey = 80808080�9012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678
FAILED (max. length in index 0: wrote 319, got 63 back)
```

solardiz:
There are more of these:

```
$ fgrep 'idx << 7' *.c
opencl_keystore_fmt_plug.c:	saved_idx[index] = (key_idx << 7) | len;
opencl_krb5pa-md5_fmt_plug.c:	saved_idx[index] = (key_idx << 7) | len;
opencl_mscash_fmt_plug.c:	saved_idx[index] = (key_idx << 7) | len;
opencl_nt_fmt_plug.c:	saved_idx[index] = (key_idx << 7) | len;
opencl_ntlmv2_fmt_plug.c:	saved_idx[index] = (key_idx << 7) | len;
opencl_timeroast_fmt_plug.c:	saved_idx[index] = (key_idx << 7) | len;
```

but I'm not sure if it's exhaustive?

Also, do we ensure anywhere that `key_idx << 7` fits the target type, which appears to be `cl_uint`? I think our outer code only limits to 2G-1, so do we risk integer wraparound here?

solardiz:
> There are more of these:
> 
> ```
> $ fgrep 'idx << 7' *.c
> ```

Except for keystore and nt-long, most of these OpenCL formats couldn't even handle 125 anyway, so are exempt from my upgrades now anyway.

magnumripper:
I call them packed buffer formats. I think there may be ones with some other shift than 7. Lets not bump any of them until later. We have a few formats (such as CryptoSafe-opencl) with an other variant of "packing" that doesn't use this overloading (so doesn't limit max. GWS). Perhaps we should change all of them to that.

solardiz:
With the currently pushed changes, `dummy` remains same speed and `plaintext` becomes ~4% faster in my testing.

solardiz:
This passed many tests I ran on different systems/devices, but then I ran `--test-full=0` and this crashes in the PDF format here:

```c
static void
pdf_compute_hardened_hash_r6(unsigned char *password, int pwlen, unsigned char salt[8],
        unsigned char *ownerkey, unsigned char hash[32])
{
        unsigned char data[(128 + 64 + 48) * 64];
        unsigned char block[64];
        int block_size = 32;
        int data_len = 0;
        int i, j, sum;

        SHA256_CTX sha256;
        SHA512_CTX sha384;
        SHA512_CTX sha512;
        AES_KEY aes;

        /* Step 1: calculate initial data block */
        SHA256_Init(&sha256);
        SHA256_Update(&sha256, password, pwlen);
        SHA256_Update(&sha256, salt, 8);
        if (ownerkey)
                SHA256_Update(&sha256, ownerkey, 48);
        SHA256_Final(block, &sha256);

        for (i = 0; i < 64 || i < data[data_len * 64 - 1] + 32; i++)
        {
                /* Step 2: repeat password and data block 64 times */
                memcpy(data, password, pwlen);
                memcpy(data + pwlen, block, block_size);
                // ownerkey is always NULL
                // memcpy(data + pwlen + block_size, ownerkey, ownerkey ? 48 : 0);
                data_len = pwlen + block_size /* + (ownerkey ? 48 : 0) */;
                for (j = 1; j < 64; j++)
                        memcpy(data + j * data_len, data, data_len);
```

Looks like `unsigned char data[(128 + 64 + 48) * 64];` has 128 for the maximum password length there.

solardiz:
The PDF format was easy to fix, but then also the GPG format crashes with `--test-full`. Not so obvious as it uses the macros for buffer sizes.

Maybe we should be using `--test-full` in our `make check` and thus CI.

solardiz:
```
Testing: gpg, OpenPGP / GnuPG Secret Key [32/64]... (4xOMP) =================================================================
==19571==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7ffe009c51f0 at pc 0x76eb1d039a1e bp 0x7ffe009bfe00 sp 0x7ffe009bf5b0
WRITE of size 327 at 0x7ffe009c51f0 thread T0
    #0 0x76eb1d039a1d in __interceptor_memcpy (/lib64/libasan.so.6+0x39a1d)
    #1 0x75b2d0 in S2KItSaltedSHA1Generator /home/user/j/john/src/gpg_common_plug.c:539
    #2 0x764cea in crypt_all._omp_fn.0 /home/user/j/john/src/gpg_fmt_plug.c:133
    #3 0x76eb1ca772d5 in GOMP_parallel (/lib64/libgomp.so.1+0x142d5)
    #4 0x7650e7 in crypt_all /home/user/j/john/src/gpg_fmt_plug.c:127
    #5 0x92afdc in is_key_right /home/user/j/john/src/formats.c:594
    #6 0x933ddb in fmt_self_test_body /home/user/j/john/src/formats.c:1319
    #7 0x9354e5 in fmt_self_test /home/user/j/john/src/formats.c:2079
    #8 0x90f65d in benchmark_all /home/user/j/john/src/bench.c:883
    #9 0x943632 in john_run /home/user/j/john/src/john.c:1695
    #10 0x943632 in main /home/user/j/john/src/john.c:2117
    #11 0x76eb1c89ab74 in __libc_start_main (/lib64/libc.so.6+0x27b74)
    #12 0x4061dd in _start (/home/user/j/john/run/john+0x4061dd)

Address 0x7ffe009c51f0 is located in stack of thread T0 at offset 21408 in frame
    #0 0x75a76a in S2KItSaltedSHA1Generator /home/user/j/john/src/gpg_common_plug.c:466

  This frame has 3 object(s):
    [32, 60) 'incs' (line 522)
    [96, 192) 'ctx' (line 471)
    [224, 21408) 'keybuf' (line 478) <== Memory access at offset 21408 overflows this variable
HINT: this may be a false positive if your program uses some custom stack unwind mechanism, swapcontext or vfork
      (longjmp and C++ exceptions *are* supported)
SUMMARY: AddressSanitizer: stack-buffer-overflow (/lib64/libasan.so.6+0x39a1d) in __interceptor_memcpy
  0x1000401309e0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x1000401309f0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x100040130a00: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x100040130a10: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x100040130a20: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
=>0x100040130a30: 00 00 00 00 00 00 00 00 00 00 00 00 00 00[f3]f3
  0x100040130a40: f3 f3 f3 f3 f3 f3 f3 f3 f3 f3 f3 f3 f3 f3 f3 f3
  0x100040130a50: f3 f3 f3 f3 f3 f3 f3 f3 f3 f3 f3 f3 f3 f3 00 00
  0x100040130a60: 00 00 00 00 00 00 00 00 00 00 f1 f1 f1 f1 00 00
  0x100040130a70: 00 00 00 00 00 00 f3 f3 f3 f3 00 00 00 00 00 00
  0x100040130a80: 00 00 00 00 00 00 00 00 f1 f1 f1 f1 00 f3 f3 f3
  Addressable:           00
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
```

solardiz:
That `keybuf` in the GPG format was previously increased arbitrarily in response to ASan too, in 714779815a9c79d348c28c4a6be86eabd6dfe694. So we can either do that empirically again or come up with a formula for the required size this time.

solardiz:
> Maybe we should be using `--test-full` in our `make check` and thus CI.

Not so easy: it's quite slow with ASan. Also, this GPG bug is only triggerable by `--test-full` when MKPC is high, so with multiple threads or manually increased `--tune`.

solardiz:
I fixed the PDF and GPG bugs seen with `--test-full`.

magnumripper:
I see this with nvidia, won't look into it yet.
```
Testing: pdf-opencl, PDF encrypted document [MD5-RC4 / SHA2-AES OpenCL]... FAILED (crypt_all(5) zero return)
```
Edit: I'm testing d1d2d175c0
Edit: The above was with `--test=0`. With `--test-full=0` I get a similar failure:
```
Testing: pdf-opencl, PDF encrypted document [MD5-RC4 / SHA2-AES OpenCL]... FAILED (crypt_all(1024) zero return)
```

solardiz:
> I see this with nvidia, won't look into it yet.

That's weird, I had tested on all of super's GPUs and all 3 NVIDIAs got 101/101 PASS.

magnumripper:
I now confimed that I can reproduce it, and that current bleeding does not have the problem. I can look into it once I get past the bulk tests.

magnumripper:
Looks like MSSQL segfaults when using ISO-8859-1 (running the test suite)

magnumripper:
> Looks like MSSQL segfaults when using ISO-8859-1 (running the test suite)

Oh but apparently this is not new, so a separate issue!

magnumripper:
I ran all variants of self-tests except encoding stuff so far, seeing no other problem than the mentioned PDF-opencl.

I also managed to nurse the Test Suite through all formats. A number of formats partly "fails" falsely due to limitations in TS itself, and always did. Taking that into account and comparing before/after does NOT show any problem with this PR. It doesn't even trigger any problem with PDF-opencl.

BTW I had a quick look at the PDF kernel and can't see anything obviously wrong with it - it seems to use PLAINTEXT_LENGTH correctly.

Now testing encodings (eg. `--test=0 -enc:iso-8859-1`) and so far I see a strange fail with LM. I can't see why that would be affected by this PR.
```
$ ../run/john --test=0 -enc:iso-8859-1
Will run 24 OpenMP threads
Testing: descrypt, traditional crypt(3) [DES 256/256 AVX2]... (24xOMP) PASS
Testing: bsdicrypt, BSDI crypt(3) ("_J9..", 725 iterations) [DES 256/256 AVX2]... (24xOMP) PASS
Testing: md5crypt, crypt(3) $1$ (and variants) [MD5 256/256 AVX2 8x3]... (24xOMP) PASS
Testing: md5crypt-long, crypt(3) $1$ (and variants) [MD5 32/64]... (24xOMP) PASS
Testing: bcrypt ("$2a$05", 32 iterations) [Blowfish 32/64 X3]... (24xOMP) PASS
Testing: scrypt (16384, 8, 1) [Salsa20/8 128/128 AVX]... (24xOMP) PASS
Testing: LM [DES 256/256 AVX2]... (24xOMP) FAILED (cmp_all(475))
(...)
```
Umm, I hope we don't have any hardcoded lengths in `unicode.c`? It wouldn't surprise me a lot.

magnumripper:
Wait, LM has a max length of 7 anyway 🤣 I wonder what happens here.

magnumripper:
Oh, and even LM-opencl fails with that same test BUT they do fail before this PR! I can't see how this has slipped through our bots.

Separate issue.

solardiz:
I also ran the Test Suite against a build with ASan with this PR's changes. It didn't crash, not even on MSSQL. The summary is:

```
Some tests had Errors. Performed 700 tests.  5 errors  13 errors reprocessing the .POT files  1 runs had non-clean exit
Time used was 15865 seconds
```

I didn't yet check whether it's the same as before this PR or not.

magnumripper:
I skimmed through unicode.[ch] and it uses PLAINTEXT_BUFFER_SIZE and sometimes even LINE_BUFFER_SIZE so it shouldn't break with this PR.

magnumripper:
The PDF-opencl problem is hard to understand (driver bug?) but I pushed a tweak: For versions R2..R4 we copied more than needed and only then truncated the password length to 32. We now truncate earlier so the temporary buffer can be just 32 bytes as well.

This should not change anything (the buffers were correctly sized) but the problem went away, and the code is more efficient. BTW of those versions, only R3 was failing.

Apparently I was allowed to push to here, so it's in!

magnumripper (approved):
I skimmed through all your changes, double checking context of only a couple things that caught my attention. I see no evil.

run/opencl/pdf_kernel.cl
```diff
@@ -484,7 +484,7 @@ void pdf_r5(__global const uchar *pwbuf,
 #endif
 
 		uint gidx = gid * NUM_INT_KEYS + mi;
-		uint W[(PLAINTEXT_LENGTH + 8 + 63 + 9) / 64 * 16]; // Max. 133 bytes, three limbs
+		uint W[(PLAINTEXT_LENGTH + 8 + 63 + 9) / 64 * 16];
```
magnumripper: I see now this instance of PLAINTEXT_LENGTH can be 127 due to that truncation. Doesn't matter much but I'll fix it and force push.

magnumripper: Oh, and I need to protect those shorter buffers from overrun by the device side mask as well.

magnumripper: Done and force pushed. Sadly, decreasing to 127 here made the test failure reappear (but with R5 hashes, which makes sense as the change was in that kernel).

magnumripper:
The ultimate fix for pdf-opencl was adding a `volatile` to a pointer. It doesn't affect performance. Judging from what that could possibly achieve, and from hours of debugging, I am as sure as a man can be that it's a driver bug.

solardiz:
> I also ran the Test Suite against a build with ASan with this PR's changes. It didn't crash, not even on MSSQL. The summary is:
> 
> ```
> Some tests had Errors. Performed 700 tests.  5 errors  13 errors reprocessing the .POT files  1 runs had non-clean exit
> Time used was 15865 seconds
> ```
> 
> I didn't yet check whether it's the same as before this PR or not.

I also ran the TS against clean bleeding-jumbo then. There are no significant differences in output (only commit hash and test durations differ). Both runs are on "super" after `scl enable devtoolset-3 bash` to enable ASan, and for the one OpenCL format that I think TS tests I guess it ran on the Vega.

solardiz:
Oh, regarding the GPG format `keybuf` size, I just realized we have similar code copy-pasted in the OpenCL kernel. Perhaps it was overflowing this buffer all the time... I'll try fixing it.

solardiz:
And indeed, this gets it to work with Intel OpenCL on "super", where it was previously crashing (in this build with ASan):

```diff
[solar@super run]$ git diff
diff --git a/run/opencl/gpg_kernel.cl b/run/opencl/gpg_kernel.cl
index df50dec..3f82500 100644
--- a/run/opencl/gpg_kernel.cl
+++ b/run/opencl/gpg_kernel.cl
@@ -60,7 +60,7 @@ INLINE void S2KItSaltedSHA1Generator(__global const uchar *password,
 #ifdef LEAN
        uchar keybuf[128 + 64+1 + PLAINTEXT_LENGTH + SALT_LENGTH];
 #else
-       uchar keybuf[64 * (PLAINTEXT_LENGTH + SALT_LENGTH)];
+       uchar keybuf[65 * (PLAINTEXT_LENGTH + SALT_LENGTH) + 256];
        uint bs;
 #endif
 
```

Before:

```
[solar@super run]$ time ./john -te=0 -form=gpg-opencl -dev=1
Will run 32 OpenMP threads
Device 1:        Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz
Testing: gpg-opencl, OpenPGP / GnuPG Secret Key [SHA1/SHA2 OpenCL]... (32xOMP) ASAN:SIGSEGV
==25218==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x7ff2b5ee13c3 sp 0x7ff2bf4a6628 bp 0x3231303938373635 T14)
    #0 0x7ff2b5ee13c2 (+0xa3c2)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV ??:0 ??
Thread T14 created by T7 here:
    #0 0x7ff2cff4ec6a in pthread_create (/usr/lib64/libasan.so.1+0x23c6a)
    #1 0x7ff2c92e316c in tbb::internal::rml::private_server::wake_some(int) ../../src/tbb/private_server.cpp:392

Thread T7 created by T3 here:
    #0 0x7ff2cff4ec6a in pthread_create (/usr/lib64/libasan.so.1+0x23c6a)
    #1 0x7ff2c92e316c in tbb::internal::rml::private_server::wake_some(int) ../../src/tbb/private_server.cpp:392

Thread T3 created by T0 here:
    #0 0x7ff2cff4ec6a in pthread_create (/usr/lib64/libasan.so.1+0x23c6a)
    #1 0x7ff2c92e316c in tbb::internal::rml::private_server::wake_some(int) ../../src/tbb/private_server.cpp:392

==25218==ABORTING

real    0m7.030s
user    0m6.047s
sys     0m1.613s
```

After:

```
[solar@super run]$ time ./john -te=0 -form=gpg-opencl -dev=1
Will run 32 OpenMP threads
Device 1:        Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz
Testing: gpg-opencl, OpenPGP / GnuPG Secret Key [SHA1/SHA2 OpenCL]... (32xOMP) PASS

real    0m6.163s
user    0m6.258s
sys     0m1.790s
```

BTW, device numbers appear to be shifted as the GPUs disappear when I run under `scl enable devtoolset-3 bash` (edit: or rather, in builds made that way). So my runs of the TS probably also used Intel OpenCL on the CPUs.

solardiz:
It would also be good to `relbench` these changes across all formats on several devices, but I don't currently have time for that... @magnumripper maybe you?

solardiz:
I've just force-pushed the gpg-opencl fix to here, including it in my previous commits along with changes to the CPU format. Of course, I also preserved @magnumripper's later commit (rebased).

The `+ 256` in those is excessive - it previously unknowingly attempted to account for two distinct things. I don't feel like dedicating time to figuring out the exact required size now.

magnumripper:
> It would also be good to `relbench` these changes across all formats on several devices, but I don't currently have time for that... @magnumripper maybe you?

Sorry, I too have no time. We can deal with it at some later point after merging.

PR #5885, opened 2025-11-11
PKZIP format bugfix

### LINKED ISSUE #5881: pkzip format fails to crack protectedzip.com ZipCrypto archives when the decompressed size is extremely small
`zip2john` fails to parse `protectedzip.com` ZipCrypto archives when the decompressed size is extremely small
Commit: 0e4ebdea43c1eb73a5cc6079d6b9150954ffa2f0

I have isolated a persistent failure in zip2john when processing ZipCrypto-encrypted archives created by the online service protectedzip.com. 
The issue is directly linked to the decompressed size of the file in the archive.
When the decompressed size is extremely small (e.g., 6 bytes), the hash is uncrackable.
Since zip2john uses the first entry in the archive to generate the hash, if that entry is the small file, the crack fails entirely.

**1. Failing Case: Small File from protectedzip.com**

The single file `test.txt` contains the string 'mytest' (6 bytes) --> The crack fails
small_txt_file.zip (Password: 1111)
[small_txt_file.zip](https://github.com/user-attachments/files/23393636/small_txt_file.zip)
```bash
$ run/zip2john small_txt_file.zip > small_txt_file.hash

# zip2john output (CRITICAL: ts=0000 cs=0000):
ver 2.0 small_txt_file.zip/test.txt PKZIP Encr: TS_chk, cmplen=24, decmplen=6, crc=08519E89 ts=0000 cs=0000 type=8

# Cracking attempt (FAILS):
$ run/john --format=pkzip --mask='?d?d?d?d' small_txt_file.hash
Loaded 1 password hash (PKZIP [32/64])
...
0g 0:00:00:00  0g/s 250000p/s 250000c/s 250000C/s 1111..7777
Session completed
```

**2. Working Case (A): Large File from protectedzip.com**

When the file is large, the crack succeeds, proving that zip2john can handle the protectedzip.com headers in general, even though it still reports `ts=0000 cs=0000`. This suggests the necessary verification data is available for large files, but not for small ones.
large_txt_file.zip (Decomplen: 114400 bytes, Password: 1111)
[large_txt_file.zip](https://github.com/user-attachments/files/23393637/large_txt_file.zip)
```bash
$ run/zip2john large_txt_file.zip > large_txt_file.hash

# zip2john output:
ver 2.0 large_txt_file.zip/largetest.txt PKZIP Encr: TS_chk, cmplen=210, decmplen=114400, crc=EC469B4A ts=0000 cs=0000 type=8

# Cracking attempt (SUCCEEDS):
$ run/john --format=pkzip --mask='?d?d?d?d' large_txt_file.hash
1111 (large_txt_file.zip/largetest.txt)
Session completed
```

**3. Working Case (B): Small File from 7-Zip (Control Test)**
To confirm the issue is with protectedzip.com's header writing and not the small file size itself, I created an archive with 7-Zip using the exact same 6-byte file and password. This crack succeeds, and zip2john correctly extracts the check value.
`manually_created_small_txt_file.zip` (Password: 1111)
[manually_created_small_txt_file.zip](https://github.com/user-attachments/files/23394123/manually_created_small_txt_file.zip)
```bash
$ run/zip2john manually_created_small_txt_file.zip > manually_created_small_txt_file.hash

# zip2john output (SUCCESS: Correct ts/cs values):
ver 2.0 manually_created_small_txt_file.zip/test.txt PKZIP Encr: cmplen=18, decmplen=6, crc=08519E89 ts=9888 cs=0851 type=0

# Cracking attempt (SUCCEEDS):
$ run/john --format=pkzip --mask='?d?d?d?d' manually_created_small_txt_file.hash
1111 (manually_created_small_txt_file.zip/test.txt)
Session completed
```

Comparison of ZIP Headers (`7z` vs `protectedzip.com`):
Image

magnumripper:
Also drops lots of dead code (in several separate commits) and adds debug output (not enabled by default).

Closes #5881

### DIFF
--- a/src/pkzip.h
+++ b/src/pkzip.h
@@ -49,7 +49,9 @@ typedef struct zip_hash_type_t {
 	u16 c;
 	u16 c2;
 	u64 datlen;
+#if USE_PKZIP_MAGIC
 	u8 magic;					// This is used as 'magic' signature type. Also, 255 is 'generic text'
+#endif
 	u8 full_zip;
 	u32 compType;				// the type of compression  0 or 8
 #if USE_PKZIP_MAGIC
--- a/src/pkzip.h
+++ b/src/pkzip.h
@@ -34,29 +34,14 @@ u16 fget16LE(FILE *fp);
 #define PWD_VER_LENGTH      0
 #endif
 
-#if USE_PKZIP_MAGIC
-typedef struct zip_magic_signatures_t {
-	u8 *magic_signature[8];
-	u8  magic_sig_len[8];
-	u8  magic_count;
-	u8  max_len;
-} ZIP_SIGS;
-#endif
-
 typedef struct zip_hash_type_t {
 	u8 *h;						// at getsalt time, we leave these null.  Later in setsalt, we 'fix' them
 	u8 type;					// JtR hash version. Version 2 ($pkzip2$) is now the deprecated one.
 	u16 c;
 	u16 c2;
 	u64 datlen;
-#if USE_PKZIP_MAGIC
-	u8 magic;					// This is used as 'magic' signature type. Also, 255 is 'generic text'
-#endif
 	u8 full_zip;
 	u32 compType;				// the type of compression  0 or 8
-#if USE_PKZIP_MAGIC
-	ZIP_SIGS *pSig;
-#endif
 } ZIP_HASH;
 
 typedef struct winzip_salt_t {
--- a/src/pkzip_fmt_plug.c
+++ b/src/pkzip_fmt_plug.c
@@ -1253,7 +1253,6 @@ MAYBE_INLINE static int check_inflate_CODE1(u8 *next, int left)
 				bits += 8;
 			}
 			here = distfix[hold & 0x1F];
-//          dodist:
 			op = (unsigned)(here.bits);
 			hold >>= op;
 			bits -= op;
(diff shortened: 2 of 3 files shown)


magnumripper:
Tested, self-reviewed, ready to merge after peer review.

Edit: I force-pushed a minor change to the zip2john commit description. s/disable/drop/ as I actually dropped the code.

solardiz (approved):
All of these changes look good to me overall. I wonder why we're checking against exactly 24, are we sure 24 is enough and 23 is not? Maybe add a comment explaining this value choice and our confidence.

magnumripper:
> All of these changes look good to me overall. I wonder why we're checking against exactly 24, are we sure 24 is enough and 23 is not? Maybe add a comment explaining this value choice and our confidence.

A comment is and was in there already, but it's in that `check_inflate_CODE1()` function we may end up calling:
```c
/* This function handles inflate CODE type 1. This is a 'fixed' table code.  We set the fixed table, */
/* and then inflate some data (without writing anything.  If we find any BAD lookback data, we can   */
/* return a failure.  We have 24 bytes of inflate data, and this almost always is more than enough   */
/* to turn up an error.  If we find we need more, we will do more than 24                            */
```
What really helped me nailing it was when I found that Hashcat (which cracked the hash just fine) also use this very limit (at first glance it's 36, but it corresponds to 24 here when factoring in the surrounding code).

BTW if this check would end up being too strict, the worst case scenario is that we miss a case of early rejection - it does not affect correctness as opposed to the missing check! But I'm confident we're good now.

PR #5888, opened 2025-11-14
Opencl misc tweaks

magnumripper:
Function-like macro safety, inline PTX tweaks. Oh and some old garbage was removed.

Basically no real change, just adds foundations for future stuff or experiments.

### DIFF
--- a/run/opencl/opencl_misc.h
+++ b/run/opencl/opencl_misc.h
@@ -207,8 +207,11 @@ INLINE uint funnel_shift_right_imm(uint hi, uint lo, uint s)
 #endif
 #endif
 
-#define block_swap32(W, len)	for (uint i = 0; i < len; i++) W[i] = SWAP32(W[i])
-#define block_swap64(W, len)	for (uint i = 0; i < len; i++) W[i] = SWAP64(W[i])
+#define block_swap32(W, LEN)	  \
+	do { for (uint i_ = 0; i_ < (LEN); i_++) (W)[i_] = SWAP32((W)[i_]); } while (0)
+
+#define block_swap64(W, LEN)	  \
+	do { for (uint i_ = 0; i_ < (LEN); i_++) (W)[i_] = SWAP64((W)[i_]); } while (0)
 
 INLINE ushort SWAP16(ushort x)
 {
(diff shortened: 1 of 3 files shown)


run/opencl/opencl_misc.h
```diff
@@ -135,34 +135,56 @@ INLINE uint lut3(uint x, uint y, uint z, uchar m)
...
+#define ASM_VOLATILE
+
+#if 0 //SCALAR && gpu_nvidia(DEVICE_INFO)
+/*
+ * This ensures a ulong lives in a paired 64-bit virtual register.
+ */
+inline ulong mov_b64(ulong x)
+{
+    ulong y;
+    asm ASM_VOLATILE ("mov.b64 %0, %1;" : "=l"(y) : "l"(x));
```
solardiz, line 151: It may be possible to achieve similar effect (or better) by having an empty asm (no instructions), but declaring our variable as input-output for it and of the proper type that implies such register allocation. I did similar things in x86_64 code, it works for me there, so I suspect the same would work here?

solardiz (approved):
I don't know about "i" vs. "n", but I trust you on that. I also made a comment about a possible improvement.

run/opencl/opencl_misc.h
solardiz: In yescrypt:

```c
/* Force use of lower-numbered registers to reduce number of prefixes, relying
 * on out-of-order execution and register renaming. */
#define FORCE_REGALLOC_1 \
        __asm__("" : "=a" (x), "+d" (Smask2reg), "+S" (S0), "+D" (S1));
#define FORCE_REGALLOC_2 \
        __asm__("" : : "c" (lo));
```

```c
#define FORCE_REGALLOC_3 __asm__("" : : "b" (Sw));
```

magnumripper: The mov.b64 instruction (and the inline function) is apparently a common trick. And it appears Hashcat uses mov.b64 in subtly different ways I don't fully understand yet. I will experiment much more with it, so subject to change.

solardiz: Well, I suppose with my suggestion we'd leave it up to the compiler to generate this instruction or not. When you explicitly have it in asm like this, you may sometimes end up with duplicate instructions like this (one compiler-generated and one from your asm), although I suppose with PTX this would get optimized out in the translation to ISA.

PR #5889, opened 2025-11-14
salted-sha1-opencl: Autotune bugfixes

magnumripper:
Problems found while working on something else. Even segfaulted under certain conditions.

### DIFF
--- a/src/opencl_salted_sha_fmt_plug.c
+++ b/src/opencl_salted_sha_fmt_plug.c
@@ -107,6 +107,7 @@ static void release_clobj(void);
 
 static void create_clobj_kpc(size_t kpc)
 {
+	global_work_size = kpc;
 	release_clobj_kpc();
 
 	pinned_saved_keys = clCreateBuffer(context[gpu_id], CL_MEM_READ_ONLY | CL_MEM_ALLOC_HOST_PTR, BUFSIZE * kpc, NULL, &ret_code);
@@ -786,7 +787,6 @@ static void auto_tune(struct db_main *db, long double kernel_run_ms)
 	size_t lws_limit, lws_init;
 
 	struct timeval startc, endc;
-	struct s_salt new_salt;
 	long double time_ms = 0, old_time_ms = 0;
 
 	size_t pcount, count;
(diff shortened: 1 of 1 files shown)


src/opencl_salted_sha_fmt_plug.c
```diff
@@ -846,9 +847,9 @@ static void auto_tune(struct db_main *db, long double kernel_run_ms)
 		tune_gws = 0;
 
 	/* Auto tune start.*/
-	new_salt.len = strlen("Hello");
 	memset(&new_salt, 0, SALT_SIZE);
-	memcpy(new_salt.data.c, "Hello", new_salt.len);
+	strcpy((char*)new_salt.data.c, "Hello");
+	new_salt.len = strlen((char*)new_salt.data.c);
```
magnumripper: Stumbled upon this by coincidence. New_salt.len would always end up as 0, screwing up the autotune.

src/opencl_salted_sha_fmt_plug.c
```diff
@@ -107,6 +107,7 @@ static void release_clobj(void);
 
 static void create_clobj_kpc(size_t kpc)
 {
+	global_work_size = kpc;
```
magnumripper, line 110: This was needed to ensure clear_keys() did not use some outdated value for gws, memsetting past buffer and b00m. I only ever saw it happen under the Test Suite, no idea why.

The ultimate problem is the format has its own autotune, not quite by our "conventions".

src/opencl_salted_sha_fmt_plug.c
solardiz: We generally do want to initialize the entire binary salt structs, or else comparisons for same/different salts don't work reliably. Probably irrelevant here, but maybe it'd be better to swap the old `new_salt.len` assignment and `memset` lines instead of switching to `strcpy`? Or switch to `strncpy` along with a comment that NUL padding is intentional.

solardiz (approved):
Approving, but I also made a comment.

src/opencl_salted_sha_fmt_plug.c
solardiz: Besides salt comparisons, we may also copy salts up to SALT_SIZE, and reading uninitialized data is just not right - may be detected by various sanitizers or even hardware later on.

magnumripper: This isn't in `set_salt()`, it's part of the autotune constructing a salt not from a test vector using its own half assed code. I guess something like `set_salt(get_salt(fmt_opencl_salted_sha1.params.tests[0].ciphertext));` would do it better in one, albeit long line, lol (and something like that is what we do in the shared code, I'm pretty sure).

magnumripper:
> I guess something like `set_salt(get_salt(fmt_opencl_salted_sha1.params.tests[0].ciphertext));` would do it better

I actually tried the above and it worked right off the bat. So I pushed that fix instead!

PR #5894, opened 2025-11-17
Warning and logging tweaks

magnumripper:
Adds color to many warning/errors. This includes shared stuff such as BENCH_CLERROR and error() so we get more visibility to many problems.

Adds shared macros LOG_ONCE, WARN_ONCE and WARN_LOG_ONCE.

Drops obsolete dead code.

### DIFF
--- a/doc/NEWS
+++ b/doc/NEWS
@@ -488,6 +488,8 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
 - Add support to keepass2john for the current format of KeePass XML keyfiles.
   The formats needed no changes.  [magnum; 2025]
 
+- Optionally add colors to many warning/error messages.  [magnum; 2026]
+
 
 Major changes from 1.8.0-jumbo-1 (December 2014) to 1.9.0-jumbo-1 (May 2019):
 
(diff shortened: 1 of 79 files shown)


magnumripper:
I wasn't quite sure where to place WARN_ONCE macro. I put it in misc.h.

Edit: It's now in color.h, which is sourced by misc.h

magnumripper:
I will revert the `log_event` vs. `log_event_ex` thing. Since the callers are changed anyway when switching to the macro, the newline should just be dropped there instead.

magnumripper:
> I will revert the `log_event` vs. `log_event_ex` thing. Since the callers are changed anyway when switching to the macro, the newline should just be dropped there instead.

Did so now, and putc the newline to stream after the fprintf_color of the message.

Also added much more (modes, formats). This PR can stay draft for a while, I'll add more of the same.

magnumripper:
I am sure I missed things that would really benefit from color, we'll just have to add them later.

magnumripper:
@solardiz this is a pain to review but at least "core files", modes, shared OpenCL and formats are different commits and best reviewed one by one.

Also, I'm not only sure you will request some changes right away but also that we'll both end up wanting to change other things later. It's more or less impossible to test all this by just running john - some notes/warnings doesn't pop up too often.

I used color_notice for many messages starting with "Note:" and color_warning for nearly all starting with "Warning:" but there were a few "obvious" exceptions and we'll probably find out more over time.

magnumripper:
I will tweak and test this more. Still a few details to fix.

Lakarije94:
good

PR #5895, opened 2025-11-17
Update fvde2john.py for removable drive support

holly-o:
Added metadata block decryption and parsing to support FileVault on removable drives


solardiz:
Thank you! I cannot properly review this for its actual functionality as I'm not familiar with the problem it solves, but it looks reasonable to me.

Maybe also add a NEWS entry?

@exploide Would you like to take a look as well?

ifechukwuanagwaonye-a11y (approved):

ifechukwuanagwaonye-a11y (approved):

exploide:
> @exploide Would you like to take a look as well?

I'm neither familiar with the file format nor do we have sample files (these would be quite large anyway and difficult to store I guess). But from skimming over the code, I think it's fine.

As long as @holly-o tested the code and also ensured it still works with non-removable media, I think this is a great addition!

solardiz:
> nor do we have sample files (these would be quite large anyway and difficult to store I guess)

Maybe these wouldn't necessarily be too large for inclusion in john-samples, @holly-o? If so, we'd appreciate a PR against that repo. Thanks!

Also, I think a NEWS entry is needed.

solardiz:
@holly-o Should we merge this as-is or are you going to add NEWS (maybe in same commit or separate) and stderr.write newlines (separate commit) first? Please let us know. Thank you!

PR #5896, opened 2025-11-18
Unify MD4 and MD5 raw formats (CPU and OpenCL)

magnumripper:
Unify as in make almost identical - we could also create a mdX_common.h and stuff and share some functions, but that's too much work.

The MD4 ones were initially very close to copies of MD5 with s/MD5/MD4). Over the years they always start to differ unneededly and I use to re-unify them now and then, so a bugfix or improvement in one isn't missing in the other.

Basically I run Meld with the two files and throw things left or right.

This time we got added support for the "{MD4}<base64>" input format as a result, needed or not. Can't hurt.

### DIFF
--- a/src/opencl_rawmd4_fmt_plug.c
+++ b/src/opencl_rawmd4_fmt_plug.c
@@ -1,9 +1,10 @@
 /*
  * MD4 OpenCL code is based on Alain Espinosa's OpenCL patches.
  *
- * This software is Copyright (c) 2010, Dhiru Kholia <dhiru.kholia at gmail.com>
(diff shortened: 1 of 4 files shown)


solardiz (approved):
I see there are very few changes to the MD5 formats here, so I assume the changes to MD4 ones do bring them in sync and so have indirectly been tested via MD5.

magnumripper:
> I assume the changes to MD4 ones do bring them in sync and so have indirectly been tested via MD5.

Exactly. If nothing else, they'll be bug compatible :)

PR #5899, opened 2025-11-18
Iterated sha1

magnumripper:
Add iterated-sha1 format(s). Optionally salted ($s.$p) and optionally iterated. Actually they can even do raw-sha1 if formatted right!

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -225,3 +225,23 @@ int isdecu(const char *q)
 		return 0;
 	return isdec_len(q, "4294967295");
 }
+
+int getdec(const char *string, char separator)
+{
+	int i = 0;
+	char buf[11];
+
+	while (string[i] >= '0' && string[i] <= '9') {
+		buf[i] = string[i];
+		if (++i >= sizeof(buf) - 1)
+			break;
+	}
+	if (i == 0 || string[i] != separator)
+		return -1;
+	buf[i] = 0;
+
+	if (!isdec(buf))
+		return -1;
+
+	return atoi(buf);
+}
--- a/run/opencl/iterated_sha1_kernel.cl
+++ b/run/opencl/iterated_sha1_kernel.cl
@@ -1,0 +1,259 @@
+/*
+ * Copyright (c) 2025, magnum
+ * This software is hereby released to the general public under
+ * the following terms: Redistribution and use in source and binary
+ * forms, with or without modification, are permitted.
+ */
+
+#include "opencl_device_info.h"
+#define AMD_PUTCHAR_NOCAST
+#include "opencl_misc.h"
+#include "opencl_mask.h"
+#include "opencl_sha1.h"
+
+/* This handles an input of 0xffffffffU correctly */
+#define BITMAP_SHIFT ((BITMAP_MASK >> 5) + 1)
+
+INLINE void cmp_final(uint gid,
+		uint iter,
+		uint *hash,
+		__global uint *offset_table,
+		__global uint *hash_table,
+		__global uint *return_hashes,
+		volatile __global uint *output,
+		volatile __global uint *bitmap_dupe) {
+
+	uint t, offset_table_index, hash_table_index;
+	ulong LO, MI, HI;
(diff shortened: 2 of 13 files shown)


src/common.c
```diff
@@ -225,3 +225,23 @@ int isdecu(const char *q)
 		return 0;
 	return isdec_len(q, "4294967295");
 }
+
+int getdec(char *string, char separator)
+{
+	int i = 0;
+	char buf[11];
```
solardiz, line 232: I'd use `const char *` for the argument, and two pointers instead of `i`.

src/common.c
```diff
@@ -225,3 +225,23 @@ int isdecu(const char *q)
...
+
+	while (string[i] && string[i] >= '0' && string[i] <= '9') {
```
solardiz: No need to check `string[i]` for non-zero if we also do the range check.

solardiz (approved):
Not a full review, but I skimmed.

Where are these hashes used? Why specifically 1024 iterations? Do we need default iteration counts at all, or should we require the field?

src/common.c
magnumripper: I don't dare changing to pointers, tested it thoroughly as-is :)

solardiz:
I still would like to know what the motivation was for adding this, and why the 1024 default. We also need a NEWS entry.

magnumripper:
I think it started out in some CTF but the idea now was to have a flexible format (btw the dynamic format can't do iterations).

The 1024 for other lengths was arbitrary. Maybe always default to 1 (not iterated) is more logical? Then it's like raw-sha1 (when no salt) and XSHA (salted) would be supported without any special exception (and it's obsolete anyway I guess).

On another note I toyed with the idea we could have a `--default-iterations=N` option but it would be hard/impossible to warn when it's actually not ending up being used.

solardiz:
> btw the dynamic format can't do iterations

Ideally, we'd have more maintainable code for dynamic format, so that we could reasonably add iterations to there, and also support many different ways iterations can be implemented (e.g., binary vs. hex as input to each iteration). But that's currently unrealistic - that code is too messy as it is.

> The 1024 for other lengths was arbitrary. Maybe always default to 1 (not iterated) is more logical?

I think there should be no default iteration count. Just refuse to work without it specified. I suggest that you revise the code added here to remove any default iteration counts.

> we could have a `--default-iterations=N` option

I think we shouldn't have that. It's more confusing and error-prone than useful.

solardiz:
I think the special case for length 48 makes with 1 iteration makes sense. It's syntax we used in the XSHA format anyway, so it's good the OpenCL format would be suggested as a possible alternative.

On CPU and with auto-detection, will the XSHA or the iterated format be picked up by default for those hashes? Which one is faster? We want the faster one to be the default.

magnumripper:
Turns out the old XSHA format is uppercase (only). And actually it's way faster than this CPU format, as it's more optimized.

> On CPU and with auto-detection, will the XSHA or the iterated format be picked up by default for those hashes?

This new one (if made supporting uppercase), then Tiger and only then XSHA.

A problem with running this OpenCL format for XSHA is that the internal mask will default to not being used (target 0). You need to use `--mask-internal-target=10000` or so to get max. speed. The reason is once we know the max. loaded iteration count, mask mode is initialized already. I've been contemplating some fix for that before (would be great for many other formats as well).

Perhaps we should simply not accept any bare ciphertexts.

solardiz:
Oh, uppercase and Tiger...

> Perhaps we should simply not accept any bare ciphertexts.

Well, or we could _require_ uppercase for those 48 hex strings. Can we set a mask-internal-target upon seeing the first one of those in `valid` unless it had already been set (such as from the command line)?

XSHA on GPU is pretty much the only immediate use case for this PR's additions, I think. So it feels unreasonable not to support it well.

magnumripper:
> we could _require_ uppercase for those

That's not a bad idea. I will look into this. Also check what xsha format pot entries look like.

PR #5900, opened 2025-11-18
androidfde2john.py: Write warning to stderr, not stdout

magnumripper:


solardiz (changes requested):
I think we need to add `\n` to the string

solardiz:
We have some more instances of `sys.stderr.write` that are missing the newline character, including in `fvde2john.py` @holly-o

magnumripper:
> We have some more instances of `sys.stderr.write` that are missing the newline character, including in `fvde2john.py` @holly-o

Yes, many in that one. I'm not touching it as there's a PR affecting it. I can't find any others.

solardiz:
> I can't find any others.

This finds some:

```sh
fgrep -A1 stderr.write *.py | fgrep -v '\n' | grep -B1 '"[) ]'
```

PR #5901, opened 2025-11-18
Format selection: Add an "after" option

magnumripper:
Matches all formats after a specific one. Put in other words, skip all formats up to and including the named one.

This is for situations where you test all formats and it segfaults on eg. foo-opencl.  You can now resume the tests from the next format, using --format=/foo-opencl.

### DIFF
--- a/src/formats.c
+++ b/src/formats.c
@@ -235,6 +235,27 @@ static int exclude_formats(char *rej_format, struct fmt_main **full_fmt_list)
 	return removed;
 }
 
+/* Exclusions. Drop all format(s) up to and including the matching rej_format from full list */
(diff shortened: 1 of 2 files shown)


solardiz (approved):
This is a welcome feature, thanks!

Those `drop` variables look unnecessarily complicated to me, but maybe I'm missing something.

magnumripper:
> Those `drop` variables look unnecessarily complicated to me, but maybe I'm missing something.

Not sure if you're right or not - the logic has many moving gears because of wildcards and other things.

solardiz:
This may be worth a NEWS entry.

magnumripper:
Not in NEWS I think - no normal user would have a need for it anyway. But I could add it to doc/OPTIONS.

Issue #2666, opened 2017-08-14
descrypt-opencl should allow for use of a generic kernel (not force per-salt kernels)


solardiz:
Often it is important to minimize startup time rather than maximize runtime performance. Right now, descrypt-opencl always(?) uses per-salt kernels, which typically take up to ~2 hours to build from source and up to tens of minutes to "build from binary". It should be possible to request use a generic kernel instead - a kernel that would run slower (needing pointer indirection for the salts), but would only be built once. We used to have that, but lost it since. We should reintroduce it as an option (maybe even as the default), and should print a "Note: ..." to the user indicating how to enable the other behavior.

IIRC, previously the HARDCODE_SALT setting controlled this, but now it's at 0 yet we do hard-code salts into the multiple kernels.

Now there's also the USE_BASIC_KERNEL setting, currently only enabled when the device is a CPU, or on OS X. I thought that maybe this was what we needed. I tried setting "--device" to a CPU to test it, but first I got unrealistically good speeds for a CPU (200M+ c/s on a machine that only does under 80M with C+intrinsics) with AMD OpenCL and a segfault with Intel's. Restarting these, I also got an instant segfault with AMD's. So this is actually unreliable at least with those (outdated versions of) OpenCL backends. Trying it on NVIDIA GPU (by forcing "#define USE_BASIC_KERNEL 1" in the source), I got:

```
OpenCL CL_INVALID_KERNEL_ARGS error in opencl_DES_bs_b_plug.c:677 - Enque kernel DES_bs_25 failed.
```

I'd rather leave (re)implementing this to someone familiar with the code. ;-)

magnumripper:
https://github.com/openwall/john/pull/5902#issuecomment-3565101416
> > BTW since some time now, we can use `JOHN_DES_KERNEL=bs_b` to force a kernel that is not as fast, but doesn't use per-salt kernels. For short runs with tons of salts, it's a gem.
> 
> Please add this note to #2666, and can we make this feature more apparent to users? Maybe even the default when there are many salts loaded or many missing per-salt kernels?

[...]

PR #5902, opened 2025-11-18
DEScrypt-opencl: Drop support for parallel kernel build

### LINKED ISSUE #1618: descrypt-opencl: Compile per-salt kernels using OpenMP
This is for HARDCODE_SALT of course.

@Sayantan2048 this should basically be really trivial and would speed up initial building of all salts' kernels a whole lot on multicore systems (eg. super would do it 32x faster). However, you'd probably need to move some code around.

What do you think?

magnumripper:
It was disabled by default but always lead to problems when I experimented with it. Closes #1618

### DIFF
--- a/run/opencl/opencl_DES_hst_dev_shared.h
+++ b/run/opencl/opencl_DES_hst_dev_shared.h
@@ -18,7 +18,6 @@
 #define OVERRIDE_AUTO_CONFIG	0
 #define HARDCODE_SALT 		0
 #define FULL_UNROLL		0
-#define PARALLEL_BUILD		0
 
 #define WORD     		int
 #define DES_bs_vector		WORD
--- a/src/opencl_DES_bs_f_plug.c
+++ b/src/opencl_DES_bs_f_plug.c
@@ -10,10 +10,6 @@
 #include <string.h>
 #include <sys/time.h>
 
-#if _OPENMP
-#include <omp.h>
-#endif
-
 #include "options.h"
 #include "opencl_DES_bs.h"
 #include "../run/opencl/opencl_DES_hst_dev_shared.h"
@@ -701,18 +697,11 @@ static void reset(struct db_main *db)
 		if (num_salts > 10 && !ocl_any_test_running && john_main_process)
 			fprintf(stderr, "Building %d per-salt kernels, one dot per three salts done: ", num_salts);
 
-#if _OPENMP && PARALLEL_BUILD
-#pragma omp parallel for
-#endif
 		for (i = 0; i < num_salts; i++) {
 			init_kernel(salt_list[i], gpu_id, 1, 0, forced_global_keys ? 0 :local_work_size);
 
-#if _OPENMP && PARALLEL_BUILD
-			if (omp_get_thread_num() == 0)
-#endif
-			{
-				opencl_process_event();
-			}
+			opencl_process_event();
+
 			if (num_salts > 10 && (i % 3) == 2 && !ocl_any_test_running && john_main_process)
 				fprintf(stderr, ".");
 		}
--- a/src/opencl_DES_bs_h_plug.c
+++ b/src/opencl_DES_bs_h_plug.c
@@ -10,10 +10,6 @@
 #include <string.h>
 #include <sys/time.h>
 
-#if _OPENMP
-#include <omp.h>
-#endif
-
 #include "options.h"
 #include "opencl_DES_bs.h"
 #include "../run/opencl/opencl_DES_hst_dev_shared.h"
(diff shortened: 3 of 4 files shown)


solardiz (approved):
I'm unsure about dropping this, but I don't mind. Just fix "we'd did" in the comment.

magnumripper:
Oh, we can keep it. But back then we found that the (then current) runtimes seemed to serialize builds despite this! And currently we get a segfault because (perhaps?) of this (opencl_common.c:1270):
```c
 * Saving and restoring of the current directory here is incompatible with
 * concurrent kernel builds by multiple threads, like we'd do with the
 * PARALLEL_BUILD setting in descrypt-opencl (currently disabled and considered
 * unsupported).  We'd probably need to save and restore the directory
 * before/after all kernel builds, not before/after each.
```
I could try to repair it instead. It is a great feature, if it works.

magnumripper:
> I could try to repair it instead. It is a great feature, if it works

You broke it with 753dd9f102, not quite sure how or how to easily fix it.

I can get it running for a while and it doesn't seem any faster (on nvidia) than single thread. I do see several threads but they are far from 100% CPU each.

OTOH it's not many lines of code so we could let it be.

solardiz:
> You broke it with [753dd9f](https://github.com/openwall/john/commit/753dd9f10275c8fb147cf14d7b713e0f03d3778c), not quite sure how or how to easily fix it.

I think when we're in the `run` directory anyway, this `chdir` shouldn't make a difference, so shouldn't have broken parallel builds in that common case.

magnumripper:
OK this is from run directory
```
Building 4096 per-salt kernels, one dot per three salts done: ..........................................................................................................................................................................................................................2 errors generated.
Options used: -I opencl -cl-mad-enable -cl-std=CL1.2 -DSM_MAJOR=12 -DSM_MINOR=0 -D__GPU__ -DDEVICE_INFO=524306 -D__SIZEOF_HOST_SIZE_T__=8  ./opencl/DES_bs_kernel_f.cl
Build time: 6.027 s
Build log: In file included from <kernel>:9:
In file included from opencl/opencl_DES_kernel_params.h:65:
In file included from opencl/opencl_sboxes.h:1:
opencl/opencl_misc.h:193:1: error: unknown type name 'INLINE'
INLINE uint funnel_shift_right(uint hi, uint lo, uint s)
^
opencl/opencl_misc.h:193:12: error: expected ';' after top level declarator
INLINE uint funnel_shift_right(uint hi, uint lo, uint s)
           ^
           ;

Error building kernel ./opencl/DES_bs_kernel_f.cl. DEVICE_INFO=524306
0: OpenCL CL_BUILD_PROGRAM_FAILURE (-11) error in opencl_common.c:1338 - clBuildProgram
```
So it builds over 650 kernels and only then fails on one, in a way that looks like corrupted kernel source. This is with 24 threads.

I'll rebuild it with `-Os` and try it under gdb.

magnumripper:
> I'll rebuild it with `-Os`

There's no point, the crash is in the nvidia runtime.

Meanwhile, I was lucky enough to build all 4096 kernels in parallel with no error.

4096 kernels | initial build wall clock | %CUC | cached build wall clock | %CUC
-|-|-|-|-
single-threaded | 32:47 | ~100% | 6:30 | ~100%
24 threads | 24:27 | ~150% | 12:41 | ~100%

We don't get a good RoI from our 24 threads, just 50% faster. Building JtR itself is 21x faster, that's more like it.

An interesting data point is a parallel build of cached binaries takes _twice as long as with a single thread_. And it runs with %CUC at ~100%. So the threading does absolutely no good in that case, but it's surprising it hits performance that bad!

On a side note I recall 4096 kernels took four hours or so years ago, with earlier hardware. 32 minutes is almost bearable. And cached binaries makes it even more bearable.

On another side note we could want `make kernel-cache-clean` to always move cached DEScrypt binaries to a backup directory instead of deleting them, such as `run/opencl/des-backup`. Just for being able to manually move them back.

Bottom line is I give up on this for a few years or so again. OTOH it's really not important to get rid of this code (it's not a lot anyway) so I'm closing this PR.

BTW since some time now, we can use `JOHN_DES_KERNEL=bs_b` to force a kernel that is not as fast, but doesn't use per-salt kernels. For short runs with tons of salts, it's a gem.

solardiz:
> BTW since some time now, we can use `JOHN_DES_KERNEL=bs_b` to force a kernel that is not as fast, but doesn't use per-salt kernels. For short runs with tons of salts, it's a gem.

Please add this note to #2666, and can we make this feature more apparent to users? Maybe even the default when there are many salts loaded or many missing per-salt kernels?

PR #5903, opened 2025-11-18
LM formats: Fix for 8-bit self-tests

### LINKED ISSUE #5887: LM and LM-opencl fails with --enc:iso8859-1
https://github.com/openwall/john/pull/5884#issuecomment-3523830787

I realize now that is because ISO-8859-1 simply isn't compatible with the LM format, which uses DOS codepages such as CP437.

I assume 96c9f627c is what caused this, and we should add some code to skip that test vector or handle it some other way.

magnumripper:
The test vectors we ended up with in d7a4b6d are only compatible with DOS codepages. We now detect if running with non-DOS codepage or UTF8, and emit a warning. We then also truncate the array of self-tests so they don't fail.

Closes #5887

### DIFF
--- a/src/LM_fmt.c
+++ b/src/LM_fmt.c
@@ -17,6 +17,8 @@
 #include "DES_bs.h"
 #include "common.h"
 #include "formats.h"
+#include "color.h"
+#include "unicode.h"
 
 #define FORMAT_LABEL			"LM"
 #define FORMAT_NAME			""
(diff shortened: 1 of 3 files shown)


src/LM_fmt.c
```diff
@@ -120,6 +122,15 @@ static void init(struct fmt_main *self)
 	fmt_LM.params.min_keys_per_crypt = DES_bs_min_kpc;
 	fmt_LM.params.max_keys_per_crypt = DES_bs_max_kpc;
 #endif
+
+	static int warned;
+	if (!warned && options.target_enc > CP_DOS_HI) {
+		fprintf_color(color_warning, stderr,
+		              "Warning: LM formats incompatible with %s encoding, disabling some tests.\n",
+		              cp_id2name(options.target_enc));
+		tests[16].ciphertext = NULL; // Truncates the array after 16 entries
```
solardiz, line 133: I suggest we at least add a comment within the `tests` array initializer about index/element 16 being special. Also, let's not end messages with a dot except when it's multiple sentences/lines.

solardiz (approved):
Suggested minor changes, but approving overall

src/LM_fmt.c
magnumripper: I also added a test for `sizeof(tests) / sizeof(tests[0]) > 16` before truncating. I often comment out many test vectors when debugging.

PR #5906, opened 2025-11-21
SM3 format: Change tag from $sm3$ to $raw-sm3$

magnumripper:
Apparently libxcrypt implements Drepper style crypt-sm3 using the tag `$sm3$` and the usual B64 encoding with optional `rounds=` defaulting to 5000.

Theoretically we could tell them apart without changing this tag (raw SM3 using hex), but I'm guessing we could end up regretting that.

### DIFF
--- a/src/sm3_fmt_plug.c
+++ b/src/sm3_fmt_plug.c
@@ -31,7 +31,7 @@ john_register_one(&fmt_sm3);
 #include "sm3.h"
 
 #define FORMAT_LABEL       "SM3"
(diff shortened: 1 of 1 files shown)


magnumripper:
For some reason I can't request a review from @SamuraiOcto.

magnumripper:
Oh, and it appears hashcat use no tag for raw-sm3 and this `$sm3$` for sm3crypt already. So we should definitely merge this.

solardiz (approved):

PR #5907, opened 2025-11-22
Iterated-sha1: Don't accept raw ciphertexts except XSHA

magnumripper:
Bare XSHA is exactly 48 chars of uppercase hex. <s>Pot entries will currently be iterated-sha1 canonical, tagged and in lowercase.

I guess we should write XSHA entries to .pot using the raw uppercase format, like the XSHA-specific CPU format does? This will complicate valid a little and need adding a `split()` but it's trivial. Working on it (slowly).</s>

### DIFF
--- a/src/iterated_sha1_common.h
+++ b/src/iterated_sha1_common.h
@@ -37,7 +37,6 @@ typedef struct {
 
 extern struct fmt_tests iterated_sha1_tests[];
 
-extern char *iterated_sha1_prepare(char *fields[10], struct fmt_main *self);
 extern int iterated_sha1_valid(char *ciphertext, struct fmt_main *self);
 extern void* iterated_sha1_get_binary(char* ciphertext);
 extern void* iterated_sha1_get_salt(char* ciphertext);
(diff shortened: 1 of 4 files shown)


magnumripper:
> I guess we should write XSHA entries to .pot using the raw uppercase format, like the XSHA-specific CPU format does? This will complicate valid a little and need adding a split() but it's trivial. Working on it (slowly).

Done. Actually no `split()` needed as we require uppercase for XSHA and lowercase for others. I could even drop `prepare()`.

solardiz (approved):
I agree with the changes here.

However, I think our checks for the format tag should be case-sensitive. I now see we also have similar case-insensitive matching of tags in a few other formats - those are probably also wrong in doing that - but let's at least not make matters worse by replicating this in new formats without a reason.

magnumripper:
> However, I think our checks for the format tag should be case-sensitive. I now see we also have similar case-insensitive matching of tags in a few other formats

Wow, I never noticed. I'll fix this one but we should probably fix all other formats (giving each some thought - there might be some where it's sensible although I can't think of any).

PR #5913, opened 2025-11-25
SM3crypt

magnumripper:
Add SM3crypt formats. These are Drepper-style crypt(3) formats, using Shang Mi 3 hash (SM3) which is a Chinese replacement for SHA-256.
    
They were seen in CMIYC 2024 but are also used in the wild, presumably in Chinese Linux distros.

### DIFF
--- a/run/opencl/opencl_sm3.h
+++ b/run/opencl/opencl_sm3.h
@@ -1,4 +1,9 @@
 /*
+ * This software is Copyright (c) 2025 magnum,
+ * and it is hereby released to the general public under the following terms:
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted.
+ *
  *  Copyright 2014-2023 The GmSSL Project. All Rights Reserved.
  *
  *  Licensed under the Apache License, Version 2.0 (the License); you may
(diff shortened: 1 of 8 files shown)


solardiz (approved):
Not a full review, but looks OK to me. There are shortcoming I notice, but they're the same in our other SHA-crypt'ish files.

Regarding "FF16 is same as MD4 G, SHA-1 G, SHA-2 Maj", maybe include its optimizations for systems that lack LUT3. Can have 4 rather than 5 binary ops, but at expense of increased latency. Can also use bitselect. Perhaps a separate commit.

magnumripper:
> Not a full review, but looks OK to me. There are shortcoming I notice, but they're the same in our other SHA-crypt'ish files.

Yeah, just as with the gostcrypt/streebogcrypt formats I could put sm3crypt together with extremely little effort, using 100% existing code and some search-replace. Meld is a wonderful tool for this, just like with raw-md4 vs. raw-md5 formats.

> Regarding "FF16 is same as MD4 G, SHA-1 G, SHA-2 Maj", maybe include its optimizations for systems that lack LUT3. Can have 4 rather than 5 binary ops, but at expense of increased latency. Can also use bitselect. Perhaps a separate commit.

That's why I jotted that down, I'll improve it later. Also, I'm tempted to add SM3 to simd-intrinsics.c just because it should be easy enough... but I should probably not spend any time on that until we see more use of these formats.

I did code the modulo operations away using separate counters for them, but it ended up slower (!?)  for both CPU and GPU. I guess the compiler is better off without me, as the divisors are constants?

PR #5914, opened 2025-11-26
OpenCL: Fix shared helper macro for creating a pinned buffer with fallback to malloc

magnumripper:
The fallback code path wasn't correct, so when it kicked in the format would crash immediately. Now tested with faked failures and debug prints. Ten or so formats use this shared code so are affected. 

In separate commit, revise cryptosafe-opencl to also use the macros.


solardiz (approved):
I'm not familiar with this, but what the edited comment says about avoiding an extra step makes sense to me.

PR #5918, opened 2025-11-30
Add Windows helper batch script for cracking ZipCrypto-encrypted ZIP archives

KillMuesly:
This pull request adds a simple Windows helper script located at:

    run/zipcrypto-cracker-helper.bat

The script provides an interactive wrapper around zip2john.exe and john.exe
to simplify cracking of classic PKZIP/ZipCrypto-encrypted ZIP archives on
Windows builds of John the Ripper.

Features:
- Verifies that john.exe and zip2john.exe are available in the same directory.
- Prompts the user for the target ZIP filename.
- Automatically extracts the hash via zip2john.
- Automatically runs john to crack the password.
- Captures and displays the line from john’s output containing the cracked password.
- Pure batch script, does not modify any core John code.

The README.md has been updated to include documentation for this helper
script under a new section “Additional Windows helper script (ZipCrypto ZIP cracking)”.

This contribution is optional and intended to improve the experience
for Windows users of JtR jumbo.

### DIFF
--- a/run/RUN_PROGRAM.bat
+++ b/run/RUN_PROGRAM.bat
@@ -1,0 +1,133 @@
+@echo off
+title Crack ZIP (ZipCrypto) with John the Ripper
+setlocal ENABLEDELAYEDEXPANSION
+
+cd /d "%~dp0"
+
+echo ============================================================
+echo  START OF SCRIPT  crack_zip.bat
+echo ============================================================
+echo Current working directory:
+echo   %cd%
+echo.
+
+rem --- CHECK FOR john.exe AND zip2john.exe ---------------------
+
+if not exist "john.exe" (
+    echo ERROR: "john.exe" was not found in this directory.
+    echo Please place this BAT file in the same directory as john.exe and zip2john.exe.
+    goto END
+)
+
+if not exist "zip2john.exe" (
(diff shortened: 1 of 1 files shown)


solardiz:
Thanks, but I think we don't want this. I'll let others weigh in, though.

Some reasons why not:
1. Too much of a special case: just Windows, just one file type, just one file with one password.
2. Easy to use only when the password is simple and it does get cracked quickly. Otherwise it's harder than direct usage of the tools, and deviates from our recommended usage (where the passwords would eventually be extracted with `--show`). If we do introduce an ease-of-use wrapper, it better be more general and correct.
3. Likely AI-written, with the usual kinds of discrepancies seen. Quick to generate first draft like this, more effort to polish and maintain. I accepted some likely AI-generated PRs to passwdqc recently. Didn't want to discourage the contributor. But it's _so_ much waste of time afterwards. I've learned my lesson.
4. For the target users to notice this script is available, it would need to be documented prominently (the PR description claims it already is with this PR, but not the case - one of those AI slops), but then how to avoid cluttering the documentation with mentions of many special-case scripts like this?

magnumripper:
I'd ban the contributor for life, but that's just me.

magnumripper:
BTW I assume not only this contribution is slop but also the GitHub account.

PR #5919, opened 2025-12-04
7z: Change "data length" cost to reflect "data length penalty"

magnumripper:
Make this the default as none of the others are very useful other than for information.

I'm still fond of the other costs so this had to bump FMT_TUNABLE_COSTS to 5. I have wanted to do so before, I think it can't hurt - even though some "costs" are more important for information than for filtering.

### DIFF
--- a/src/7z_common_plug.c
+++ b/src/7z_common_plug.c
@@ -678,10 +678,21 @@ unsigned int sevenzip_compression_type(void *salt)
 	return my_salt->type;
 }
 
-unsigned int sevenzip_data_len(void *salt)
+unsigned int sevenzip_size_penalty(void *salt)
 {
-	sevenzip_salt_t *my_salt;
+	sevenzip_salt_t *my_salt = *((sevenzip_salt_t**)salt);
+	uint32_t pad_size = sevenzip_padding_size(salt);
+	uint64_t e_size = my_salt->packed_size;
 
-	my_salt = *((sevenzip_salt_t**)salt);
-	return my_salt->packed_size;
+	/* Consider early reject */
+	if (sevenzip_trust_padding) {
+		int shift = pad_size * 8;
+
+		if (shift < 64)
+			e_size >>= shift;
+		else
+			e_size = 0;
+	}
+
+	return (unsigned int)MIN(UINT_MAX, e_size);
 }
--- a/src/7z_fmt_plug.c
+++ b/src/7z_fmt_plug.c
@@ -366,7 +366,7 @@ struct fmt_main fmt_sevenzip = {
 			"iteration count",
 			"padding size",
 			"compression type",
-			"data length"
+			"data size penalty"
 		},
 		{ FORMAT_TAG },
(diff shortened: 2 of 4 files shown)


src/7z_common_plug.c
```diff
@@ -685,3 +685,21 @@ unsigned int sevenzip_data_len(void *salt)
...
+{
+	sevenzip_salt_t *my_salt = *((sevenzip_salt_t**)salt);
+	uint32_t pad_size = sevenzip_padding_size(salt);
+	uint64_t cost = my_salt->packed_size;
+
+	/* Consider early reject, if applicable */
+	if (sevenzip_trust_padding)
+		cost >>= (pad_size * 8);
+
+	cost += sevenzip_iteration_count(salt);
```
solardiz: This assumes that packed size (sans early reject) and iterations contribute to the total cost equally, but that's probably a great oversimplification. Maybe there should be an empirically tuned coefficient here at least?

src/7z_fmt_plug.c
```diff
@@ -363,6 +363,7 @@ struct fmt_main fmt_sevenzip = {
 		MAX_KEYS_PER_CRYPT,
 		FMT_CASE | FMT_8_BIT | FMT_OMP | FMT_UNICODE | FMT_ENC | FMT_DYNA_SALT | FMT_HUGE_INPUT,
 		{
+			"cost",
```
solardiz: Maybe call this "estimated cost"? Also in the other format.

solardiz (approved):
I'm unsure this is the right approach, but I don't mind.

src/7z_common_plug.c
magnumripper: I first had it multiplied with iterations and before shifting, which was way worse. Adding iterations after shifting size makes some sense: for many archives, this will correctly end up little more than the iterations count. This all is a simplification anyway because the inflate method matters.

The thing is a huge archive with no or little padding is a huge hit to performance.

Instead of this whole PR, we could change the existing "data length" cost  to be a padding-shifted one, and rename it to "effective size after early reject" or something better (not sure what).

magnumripper: I did consider something like `8 + (pad_size * 8)` though. Thinking again it's probably better and also increases the 4 GB ceiling. I can run some speed tests.

magnumripper:
> Maybe there should be an empirically tuned coefficient here at least?

After some testing I gave up on this. I did find that on CPU, a formula of `(data length >> (6 + (8 * pad_size))) + iterations` would roughly reflect how slow the attack would be. But there are too many other possible variations, and even worse with the OpenCL format.

So I took a much simpler approach: This commit now merely changes the existing "data length" cost to "data size penalty" by simply decreasing the data size according to how much padding we have. The old cost was informative but useless for filtering - the new one reflects that a 4 GB data size is no problem at all as long as we have at least two bytes of padding for early rejection - while a 50 MB size with no padding will hit performance significantly.

PR #5922, opened 2025-12-11
Keepass2john: Support newer format XML keyfile

magnumripper:
Despite comments in code, the existing support appeared to be v1 only. This new clause is tested with a v4 database and keyfile.

### DIFF
--- a/src/keepass2john.c
+++ b/src/keepass2john.c
@@ -133,11 +133,11 @@ static off_t get_file_size(char *filename)
 	return sb.st_size;
 }
 
-static void print_hex(unsigned char *str, int len)
+static void print_hex(void *str, int len)
 {
 	int i;
 	for (i = 0; i < len; ++i)
-		printf("%02x", str[i]);
+		printf("%02x", ((unsigned char*)str)[i]);
 }
 
 static uint64_t BytesToInt(unsigned char *s, const int s_size)
@@ -433,7 +433,7 @@ static void process_database(char* encryptedDatabase)
 	uint64_t Argon2_M = 0;
 
 	/* specific to keyfile handling */
-	unsigned char *buffer;
+	char *buffer;
 	int64_t filesize_keyfile = 0;
 	char *p;
 	char *data;
@@ -847,7 +847,7 @@ static void process_database(char* encryptedDatabase)
 	}
 
 	if (keyfile) {
-		buffer = mem_alloc(filesize_keyfile * sizeof(char));
+		buffer = mem_alloc(filesize_keyfile + 1);
 		printf("*1*64*"); /* inline keyfile content or hash - always 32 bytes */
 		if (fread(buffer, filesize_keyfile, 1, kfp) != 1) {
 			warn("%s: Error: read failed: %s.",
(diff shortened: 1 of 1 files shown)


src/keepass2john.c
```diff
@@ -863,14 +863,31 @@ static void process_database(char* encryptedDatabase)
 		 *  else byte_array = sha256(keyfile_content)
 		 */
 
-		if (!memcmp((char*) buffer, "<?xml", 5) &&
+		if (!memcmp((char*) buffer, "<?xml", 5) && !memcmp((char*) buffer, "<Data>", 6) &&
```
solardiz: Isn't this `&&` condition always false, because only one of the two `memcmp` calls can return zero here?

Also, I think the casts to `char *` on `memcmp` calls are unneeded, or we may want to change them to `strncmp` since we don't seem to reliably know we still fit in `filesize_keyfile` even with these tiny comparisons (or do we?)

src/keepass2john.c
```diff
@@ -863,14 +863,31 @@ static void process_database(char* encryptedDatabase)
...
 		    ((p = strstr((char*) buffer, "<Key>")) != NULL) &&
 		    ((p = strstr(p, "<Data>")) != NULL)) {
 			p += strlen("<Data>");
 			data = p;
 			p = strstr(p, "</Data>");
 			printf ("%s", base64_convert_cp(data, e_b64_mime, p - data, b64_decoded, e_b64_hex, sizeof(b64_decoded), flg_Base64_NO_FLAGS, 0));
```
solardiz: Not new with this PR, but it appears that `p` can become `NULL` here, making `p - data` invalid.

src/keepass2john.c
```diff
@@ -863,14 +863,31 @@ static void process_database(char* encryptedDatabase)
...
 		}
+		else if (!memcmp((char*)buffer, "<?xml", 5) &&
```
solardiz: Same as above about the cast and possibly needing `strncmp`.

src/keepass2john.c
```diff
@@ -863,14 +863,31 @@ static void process_database(char* encryptedDatabase)
...
+		         ((p = strstr((char*)buffer, "<Data Hash=")) != NULL) && (p = strchr(p, '>'))) {
+			char hex[64 + 1];
+			int hidx = 0;
+			p++;
+			while (hidx < 64 && p - (char*)buffer < filesize_keyfile) {
+				if (*p >= '0' && *p <= 'F')
```
solardiz: Why not do a proper two-range check here?

solardiz (changes requested):
I don't know anything about the "newer format", but I made some comments on the code assuming it can end up running on arbitrary/wrong input.

src/keepass2john.c
magnumripper: Yup, saw it and fixed it while testing (with a mockup v1 - I need to create a real one if possible)

src/keepass2john.c
magnumripper: Also fixed now, as well as ensuring the buffer is actually zero terminated (for strstr and such)

magnumripper:
Samples added to [john-samples](https://github.com/openwall/john-samples/commit/d6df64ea).

The new (well it was many many years ago) format for an XML keyfile is like this:
```
<?xml version="1.0" encoding="UTF-8"?>
<KeyFile>
    <Meta>
        <Version>2.0</Version>
    </Meta>
    <Key>
        <Data Hash="50BD7B21">
            4C772D15 A5D58BEC 0C707345 ECEBDFE2
            18F0529A 4882FE3C D48F00E5 4B733073
        </Data>
    </Key>
</KeyFile>
```
The 50BD7B21 is supposedly a CRC32 of the following 32 bytes of key data, not required and not used nor needed in any way for opening (or cracking) a database. If that attribute exists, KeePass would warn if it doesn't match. We don't currently even validate it (I fail to arrive at the CRC above, and I didn't care to figure out why - maybe endianess).

The old format (which we already supported) was more like this (IIRC) for the same 32 byte key:
```
  <?xml version="1.0" encoding="UTF-8"?>
  <KeyFile>
      <Meta>
          <Version>1.0</Version>
      </Meta>
      <Key>
          <Data>THctFaXVi+wMcHNF7Ovf4hjwUppIgv481I8A5UtzMHM=</Data>
      </Key>
  </KeyFile>
```
I think there might be whitespace between `<Data>` and the Base64. We now handle that.

We don't have such sample (the samples we had are "any file used as a key file", not XML). I may try to create one for john-samples later just for completeness.

src/keepass2john.c
magnumripper: I'll change to strncmp.

src/keepass2john.c
magnumripper: Oh, I recalled incorrectly that the uppercase alphabet followed digits. Of course it does not. I'll fix that and I think I'll allow lowercase when at it, because it can't harm.

magnumripper:
Sorry I jumped the gun with requesting a review, I think it's good to go now. Tested with v2 and mockup v1.

magnumripper:
> We don't currently even validate it (I fail to arrive at the CRC above, and I didn't care to figure out why - maybe endianess).

Turns out KeePass 2.x and KeePassXC calculate this CRC differently, making it pretty much useless. I will not add any check, it's just overkill.

src/keepass2john.c
```diff
@@ -857,19 +857,42 @@ static void process_database(char* encryptedDatabase)
...
+		    ((p = strstr((char*) buffer, "<Version>1.0")) != NULL) &&
 		    ((p = strstr((char*) buffer, "<Key>")) != NULL) &&
 		    ((p = strstr(p, "<Data>")) != NULL)) {
 			p += strlen("<Data>");
+			while (*p == ' ' || *p == '\t' || *p == '\r' || *p == '\n')
+				p++;
 			data = p;
-			p = strstr(p, "</Data>");
-			printf ("%s", base64_convert_cp(data, e_b64_mime, p - data, b64_decoded, e_b64_hex, sizeof(b64_decoded), flg_Base64_NO_FLAGS, 0));
+			printf ("%s", base64_convert_cp(data, e_b64_mime, 44, b64_decoded, e_b64_hex, sizeof(b64_decoded), flg_Base64_NO_FLAGS, 0));
```
solardiz, line 882: What if there isn't 44 chars until the end of allocated buffer?

solardiz (changes requested):
Added a comment on one code issue. Also, maybe remove `unsigned` from definition of `buffer` instead of all the casts?

src/keepass2john.c
magnumripper: Yeah base64_convert would not read past buffer but we would silently produce invalid data of course.

We now bail out with a message for that case (in the hex code path as well).

src/keepass2john.c
```diff
@@ -857,19 +857,51 @@ static void process_database(char* encryptedDatabase)
...
+		    ((p = strstr(buffer, "<Version>1.0")) != NULL) &&
+		    ((p = strstr(buffer, "<Key>")) != NULL) &&
 		    ((p = strstr(p, "<Data>")) != NULL)) {
 			p += strlen("<Data>");
+			while (*p == ' ' || *p == '\t' || *p == '\r' || *p == '\n')
+				p++;
 			data = p;
 			p = strstr(p, "</Data>");
-			printf ("%s", base64_convert_cp(data, e_b64_mime, p - data, b64_decoded, e_b64_hex, sizeof(b64_decoded), flg_Base64_NO_FLAGS, 0));
+			if (p == NULL || p - buffer < 44) {
```
solardiz, line 878: Should probably be `!= 44` rather than `< 44` here. That would be consistent with the check for hex.

solardiz (approved):

src/keepass2john.c
magnumripper: Actually no, because there may be whitespace (of any sort) between the Base64 and `</Data>`.

Admittedly we don't know they are 44 actual Base64 characters anyway. I'll see if base64_convert_cp returns something useful. BTW I don't recognize that _cp suffix, need to check what the heck that is.

magnumripper: OK so I can get an output size from base64_convert... but it returns 64 no matter what garbage I feed it. There are also means for an error variable but it always returns "no error". I'm ignoring this for now.

PR #5924, opened 2025-12-12
Python and Perl scripts: Print warnings, errors and usage to stderr

magnumripper:
Also add missing newlines to some python calls to `sys.*.write`.

Usage to stderr isn't always the norm but in this case we're normally running redirected to file and the usage blob may be caused by options/syntax problems.

### DIFF
--- a/run/DPAPImk2john.py
+++ b/run/DPAPImk2john.py
@@ -548,7 +548,7 @@ class MasterKeyFile(DataStruct):
         if context == "domain1607-" or context == "domain":
             self.decryptWithHash(userSID, hashlib.new("md4", pwd.encode('UTF-16LE')).digest())
             if self.decrypted:
-                print("Decrypted succesfully as domain1607-")
+                print("Decrypted succesfully as domain1607-", file=sys.stderr)
                 return
         if context == "domain1607+" or context == "domain":
             SIDenc = userSID.encode("UTF-16LE")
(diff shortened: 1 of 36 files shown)


solardiz (approved):
The changes look good to me, but seem incomplete. At least this is still missing the newlines:
```
fvde2john.py:    sys.stderr.write("cryptography is missing, run 'pip install --user cryptography' to install it!")
fvde2john.py:    sys.stderr.write("pytsk3 is missing, run 'pip install --user pytsk3' to install it!")
fvde2john.py:        sys.stderr.write("Error reading out of bounds of file, exiting.")
fvde2john.py:        sys.stderr.write(f"[!] Invalid sector size {sector_size} (not 512 or 4096 bytes). Exiting.")
fvde2john.py:            sys.stderr.write("[!] No CoreStorage volumes found, exiting.")
```

magnumripper:
Oh, I fixed those but dropped/postponed it because of #5895. I see now I even mentioned it in https://github.com/openwall/john/pull/5900#issuecomment-3551625736.

I'll fix them and add here. I'll also retry some grep-fu.

solardiz (approved):

magnumripper:
Will add yet some to this, I found more.

magnumripper:
I'm still working on this one. It must be rebased onto #5931 once that one is merged.

magnumripper:
> I'm still working on this one. It must be rebased onto #5931 once that one is merged.

Done now. Apparently we can merge this before or after #5931.

PR #5927, opened 2025-12-19
Add a few entries to .gitignore

magnumripper:
Backup files named #NAME#
Cached DEScrypt kernels manually copied to run/opencl/descrypt-cache/
Python .venv directory (venv without the dot existed, matter of taste)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 *.tmp
 *~
 *.old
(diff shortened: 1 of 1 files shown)


solardiz (approved):
I don't know where file/directory names like this would come from, but I don't mind.

PR #5930, opened 2025-12-24
Radius fixes

magnumripper:
Fixes and tweaks for radius2john.py, network2john.lua and the Radius format itself.

### DIFF
--- a/run/network2john.lua
+++ b/run/network2john.lua
@@ -1,6 +1,6 @@
 -- #!/usr/bin/env lua -> shebang line seems to be breaking tshark :-(
 
--- Usage: tshark -q -Xlua_script:network2john.lua -r selected.pcap
+-- Usage: tshark -q -X lua_script:network2john.lua -r selected.pcap
 --
 -- Wireshark in Fedora 25 doesn't have Lua enabled. Use Fedora 27 / Ubuntu or
 -- something else which has Wireshark with Lua support or compile Wireshark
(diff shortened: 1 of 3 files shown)


solardiz (approved):
Not a full review, but I'm OK with these changes. Maybe also add a NEWS entry?

Also, it's a bit confusing when to use radius2john vs. the LUA script - do we have this documented?

magnumripper:
> Maybe also add a NEWS entry?

Not sure it's worthy of a special mention, it's more of a "many other improvements and bugfixes" thing.

> Also, it's a bit confusing when to use radius2john vs. the LUA script - do we have this documented?

I found no documentation, I tried to understand the difference but I'm still confused because they should be essentially the same thing but they aren't.

First of all, the tshark network2john.lua  script produces hashes for the actual radius format while radius2john produces hashes for dynamic 1008, 1009 or 1017 formats, depending on type and salt length. They all mention the same [whitepaper](http://www.untruth.org/~josh/security/radius/radius-auth.html) and its sections 3.1 and 3.3. BTW I guess one difference is you could run tshark "live" while radius2john is pcap only.

What is disturbing is they produce different numbers of hashes from our samples. Not sure why. We should revisit this later and possibly write a README.radius or the like.

Also, there are two versions of radius2john: one in Perl and one in Python. The Perl version is incomplete and outdated and spews out warnings. You even need to edit it for setting required options. I think we should just drop that one.

PR #5931, opened 2025-12-24
Simplify and document Python and Perl requirements for running all our converters

magnumripper:
Many *2john tools just work, but following this virtually all of them work with little effort. This is probably also great for distro package maintainers.

Also a few bug fixes found while at it.


doc/INSTALL-UBUNTU
```diff
@@ -132,3 +132,17 @@ build CPU fallback chains for any x86-64 CPU.
 === Benchmark your build (this also performs self-tests)
 
     ./john --test
+
+=== Install all Python modules required for things like pcap2john.py
+
+    sudo apt-get install python3 python3-venv python3-pip
+    python3 -m venv venv
+    source venv/bin/activate
+    pip3 install `cat requirements.txt`
```
exploide: `pip3 install -r requirements.txt`
No need for command substitution.

magnumripper: Thanks, fixed. I also replaced `source` with `.` as that is more portable.

run/rexgen2rules.pl
```diff
@@ -74,7 +74,7 @@ sub check_parity {
             # Though possible but undefined back references are just skipped.
             # Good: (a)\1, (a\1)
             # Bad: \1(a)
-            if (/\\g{(\d+)}/ && $1 > $k) {
+            if (/\\g\{(\d+)}/ && $1 > $k) {
                 die "wrong back reference $_"
             }
```
magnumripper, line 79: @AlekseyCherepanov Is this correct? Current Perl complains about the old line but I'm not entirely sure I understand the use of curly brackets here.

AlekseyCherepanov: Yes, the fix is right. It might be more stylish to escape both curly brackets with backslashes. OTOH `$token_re` has `\{...}` too (`my $token_re = qr/(\\g\{\s*\d+\s*}|...`). There are backslashes for both braces in other places.

The regular expression in the fix matches substrings like `\g{123}`. The braces were intended to be literal. Braces are magic/active chars in Perl's regular expressions to denote repetition. It was my mistake to forget backslashes there. The magic did not work with `(\d+)` inside and older Perl was ok with that. Newer Perl is more strict parsing this nonsense. Thank you for finding and fixing it!

run/apop2john.py
AlekseyCherepanov, line 1: The first argument of `re.search` function is a string containing regular expression. While there are excessive backslashes, `+` at the beginning should be escaped to be literal. Otherwise there would be an exception:
```python3
>>> re.search(b'+OK', b'+OK')
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.11/re/__init__.py", line 176, in search
[...]
    raise source.error("nothing to repeat",
re.error: nothing to repeat at position 0
```
I think `\+OK` should be kept with its backslash. Other backslashes seem unneeded indeed.

AlekseyCherepanov: `b'\+OK ...'` is the same as `b'\\+OK ...'`. But [howto](https://docs.python.org/3/howto/regex.html#the-backslash-plague) says that the first variant yields a `DeprecationWarning` in the newest Python. So this backslash should be doubled.

magnumripper: Thanks! That is insanely backwards... until you read the howto. I'll fix it.

magnumripper: I guess we could also use raw bytes, as in `re.search(br'\+OK .* <.+>', pkt)`, avoiding one layer of escapes. That is easier to read.

magnumripper:
I now added a commit that drops the hardcoded and outdated copy of 'olefile' module from office2john.py, in favor of a normal import. Also added instructions to INSTALL-UBUNTU on how to install nearly all required modules as Ubuntu packages.

magnumripper:
> Also added instructions to INSTALL-UBUNTU on how to install nearly all required modules as Ubuntu packages.

I will do this for the needed Perl libs as well.

Edit: I dropped that idea, there are many packages needed that doesn't exist as Ubuntu packages.

Edit 2: I ended up adding the few available, for easier use of eg. `7z2john.pl` without hassling with CPAN.

magnumripper:
This should be ready to merge now.

solardiz (approved):

PR #5932, opened 2025-12-25
Disable debug print statements

p-nowodzinski:
This PR removes print statements used for debug output, as noted in existing TODO comments.

It also replaces the magic value 0x7FFFFFFF with None where stream size is unknown, and standardizes None checks.

No functional behavior outside of debug output is changed.

### DIFF
--- a/run/office2john.py
+++ b/run/office2john.py
@@ -334,7 +334,8 @@ else:
 # command line to change it.
 DEBUG_MODE = False
 def debug_print(msg):
-    print(msg)
+    # Debug output currently disabled
+    pass
 def debug_pass(msg):
     pass
 debug = debug_pass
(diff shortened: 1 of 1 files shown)


magnumripper (commented):
All instances of "Debug output currently disabled" and the like should be dropped, as they aren't disabled but removed here.

magnumripper (changes requested):
Looking a bit closer, you changed and/or dropped dead code from an outdated Python module that Dhiru copied into the tool, for some reason or the other. Maybe you should have done it upstream, if even applicable 13 years later.

What would be really usable for _us_ is to drop _all_ that copied code, and instead add whatever is needed to `requirements.txt`.

magnumripper:
Closing in favor of 9c3b5e36 in #5931

PR #5934, opened 2026-01-04
office2john: Emit a user-friendly message about missing olefile module

magnumripper:
I forgot to add this in #5931


magnumripper:
Before:
```
  File "./office2john.py", line 9, in <module>
    from olefile import isOleFile, OleFileIO
ModuleNotFoundError: No module named 'olefile'
```
After:
```
office2john.py: olefile python module is missing, please install your distro's
package, eg. 'sudo apt-get install python3-olefile' if available, otherwise
'pip install --user olefile' (activate your venv if you already did that)
```

solardiz (approved):

PR #5937, opened 2026-01-08
Compressed plist support and better regex

holly-o:
Added support for when encryption context is compressed. Also, broader regex for greater matching.

### DIFF
--- a/run/fvde2john.py
+++ b/run/fvde2john.py
@@ -12,6 +12,7 @@ import argparse
 import sys
 import re
 import base64
+import zlib
 
 try:
     from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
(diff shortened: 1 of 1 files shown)


solardiz (approved):
I don't know if this is right or not, but you're the maintainer of this script. :-)

I just wish you'd use our commit message style, so prefix the title with subsystem (script) name, and wrap further lines at 75. But I don't insist.

run/fvde2john.py
```diff
@@ -298,30 +319,35 @@ def get_CryptoUsers_dict_from_encrypted_metadata(cs_start_pos, offsets, block_si
...
+                CryptoUsers_dict = {}
+                for user_index in range(len(matches)):
+                    CryptoUsers_dict[user_index] = matches[user_index]
+                    # if present convert user type from string to hex
+                    user_type = CryptoUsers_dict[user_index].get('UserType')
+                    if user_type:
+                        CryptoUsers_dict[user_index]['UserType'] = int(CryptoUsers_dict[user_index]['UserType'], 16)
 
-            return CryptoUsers_dict
+                    return CryptoUsers_dict
```
magnumripper: The return line is moved into the loop here, perhaps accidentally? Doesn't look right to me.

magnumripper (changes requested):

run/fvde2john.py
solardiz: Oh, good catch. Maybe it was meant to be moved under `if matches`, but not into the loop, @holly-o?

solardiz: @holly-o We're waiting for you to comment on this and perhaps revise the code. Thank you!

holly-o: Thank you for the comment, yes I will amend the code so the return is outside of the loop.

solardiz (approved):

solardiz:
Thank you @holly-o and @magnumripper! Since the only issue pointed out by magnum has been fixed, I went ahead and merged this.

PR #5939, opened 2026-01-10
Drop yasm from dependencies and installation instructions

solardiz:
We used to need `yasm` for AES-NI, but we no longer do.


magnumripper (approved):
I saw this just the other week while working with #5931, noted it mentally and immediately forgot about it. I was going to say you missed autoconf but we already dropped that.

Issue #5471, opened 2024-05-04
Wrong usage of pcapngepb and pcapngpb in wpapcap2john.c


lucic71:
While running some static code analysis with clang-tidy over the john code-base I saw a dubious use of pcapngepb: https://github.com/openwall/john/blob/1c97519d09500ffcd24fd8d56541c1d3cf10ea87/src/wpapcap2john.c#L1983

It's dubious because pcapngepb is used uninitialized if this branch is taken:
https://github.com/openwall/john/blob/1c97519d09500ffcd24fd8d56541c1d3cf10ea87/src/wpapcap2john.c#L1960

Instead, I think at line 1983 pcapngpb should be used instead of pcapngepb. pcapngepb is only initialized on this branch:
https://github.com/openwall/john/blob/1c97519d09500ffcd24fd8d56541c1d3cf10ea87/src/wpapcap2john.c#L2021

I think naming is the problem here because the two names differ by one letter.

ghost:
Or there is something wrong in the source code and no tests with `pcapngbh.block_type` == 2 were run

Or

There is another possibility:
- if `pcapngbh.block_type == 6` is read from the file before `pcapngbh.block_type == 2`, the data will be filled in correctly (in the correct order) and there is no logic problem in the program (which seems more likely if some real testing for the problematic branch was executed, **see line 1990**).

What can we do:
- silence the warning using initialization for the structure in question;
- find a sample that allows us to properly test the 2john tool. If it's wrong, it's very wrong

lucic71:
I think the easiest is to initialize the struct. If your hypotheses is true (block_type 6 is executed before block_type 2) then at least the struct is 0 initialized and no undefined behavior is triggered.

solardiz:
Reviewing the code, we appear to have the same issue also for `pcapngepb.caplen`, which I guess the static analyzer missed.

All of this code was added in just one commit by @magnumripper in 2018.

Unfortunately, none of the samples we have in https://github.com/openwall/john-samples/tree/main/WPA-PSK trigger these code paths for me. https://wiki.wireshark.org/SampleCaptures has some pcapng files, but not for WPA. Maybe @magnumripper has a sample for us to test this with?

I agree that initializing the struct is the safest bet - will definitely not make things worse - but it may also paper over the bug.

solardiz:
OK, I just did:

```shell
tshark -r wpa-Induction.pcap -w wpa-Induction.pcapng
```

and got with `file`:

```
[...]

PR #5940, opened 2026-01-11
wpapcap2john fixes

### LINKED ISSUE #5938: wpapcap2john reading pcap-ng segfaults on MacOS (only?)
For some reason I only see this on MacOS, doesn't happen on Linux:
```
$ ./wpapcap2john wpa-Induction.pcapng 
File wpa-Induction.pcapng: Radiotap encapsulation
Segmentation fault: 11
```
It only happens with pcap-ng files (any such file, not just this sample). It doesn't happen with the same sample in old pcap format. And it doesn't happen on Linux at all.

While looking into it I stumbled on #5471 that was reported and fixed while I couldn't participate. The fix in 939203c is obviously correct but there were one more copy-paste error (read + seek erroneously done twice) that will be fixed while I'm at this (although that code path is for an obsolete version of pcap-ng that we have no samples for). However, this whole paragraph is unrelated to my current problem.

Back to MacOS:
```
File wpa-Induction.pcapng: Radiotap encapsulation
Process 24178 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x2)
    frame #0: 0x0000000100008406 wpapcap2john`process_packet(link_type=<unavailable>) at wpapcap2john.c:1528:16
   1525		if (link_type == LINKTYPE_RADIOTAP_HDR) {
   1526			if (snap_len < 4)
   1527				return 0;
-> 1528			frame_skip = *(unsigned short*)&packet[2];
   1529	#if !ARCH_LITTLE_ENDIAN
   1530			frame_skip = swap32u(frame_skip);
   1531	#endif
Target 0: (wpapcap2john) stopped.
(lldb) print packet
(uint8_t *) 0x0000000000000000
(lldb) print full_packet
(uint8_t *) 0x0000000000000000
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x2)
  * frame #0: 0x0000000100008406 wpapcap2john`process_packet(link_type=<unavailable>) at wpapcap2john.c:1528:16
    frame #1: 0x000000010000c3e6 wpapcap2john`process_ng(in=0x00007ff851efe9b0) at wpapcap2john.c:2062:9 [inlined]
    frame #2: 0x000000010000c2c2 wpapcap2john`process(in=0x00007ff851efe9b0) at wpapcap2john.c:2132:10 [inlined]
    frame #3: 0x000000010000c2af wpapcap2john`main(argc=2, argv=0x00007ff7bfeff500) at wpapcap2john.c:2427:4
    frame #4: 0x00007ff81015a530 dyld`start + 3056
```
The real WTF is that `packet` is NULL here, yet not under Linux. I mean, it's the exact same code and sample!? We don't use any external pcap lib or anything that could be different versions, it's all in `wpapcap2john.c`. Maybe I should just try building with a different compiler...

magnumripper:
After reading just the IDB, it could try to dump data not yet read from the pcap due to an uninitialized variable and sloppy code. When this happened, it would segfault.

Many other robustness fixes, added some comments, decreased variable scopes.

Now handles (deprecated) PB properly (without the confusing hack leading to #5471 and 939203c, which broke the support for it).
Also adds proper support for (legacy) SPB. They were just skipped prior to this, but I doubt they are used anywhere. I'm not aware of any tool that can produce these two type of blocks at all - any modern tool just go with EPB.

Closes #5938

### DIFF
--- a/src/wpapcap2john.c
+++ b/src/wpapcap2john.c
@@ -51,7 +51,7 @@ static uint8_t *packet;
 static uint8_t *packet_TA, *packet_RA, *packet_SA, *packet_DA, *bssid;
 static uint8_t *new_p;
 static size_t new_p_sz;
-static int swap_needed;
+static int swap_needed = !ARCH_LITTLE_ENDIAN;
 static essid_t *essid_db;   /* alloced/realloced to max_essid */
 static int n_essid;
 static WPA4way_t *apsta_db; /* alloced/realloced to max_state */
@@ -1897,13 +1897,10 @@ static int process_ng(FILE *in)
 	unsigned int res;
 	int aktseek;
 
-	block_header_t pcapngbh;
-	section_header_block_t pcapngshb;
-	interface_description_block_t pcapngidb;
-	packet_block_t pcapngpb;
-	enhanced_packet_block_t pcapngepb;
-
 	while (1) {
+		block_header_t pcapngbh;
+		interface_description_block_t pcapngidb;
+
 		res = fread(&pcapngbh, 1, BH_SIZE, in);
 		if (res == 0) {
 			break;
@@ -1914,56 +1911,49 @@ static int process_ng(FILE *in)
 			break;
 		}
 
+		if (verbosity >= 4)
+			fprintf(stderr, "pcap-ng block type %u\n", pcapngbh.block_type);
+
+		/* SHB, Section Header Block */
 		if (pcapngbh.block_type == PCAPNGBLOCKTYPE) {
+			section_header_block_t pcapngshb;
+
 			res = fread(&pcapngshb, 1, SHB_SIZE, in);
 			if (res != SHB_SIZE) {
 				fprintf(stderr, "%s: Failed to read pcapng SHB: %s\n",
 				       filename, feof(in) ? "EOF" : strerror(errno));
 				break;
 			}
-#if !ARCH_LITTLE_ENDIAN
-			pcapngbh.total_length = swap32u(pcapngbh.total_length);
-			pcapngshb.byte_order_magic	= swap32u(pcapngshb.byte_order_magic);
-			pcapngshb.major_version		= swap16u(pcapngshb.major_version);
-			pcapngshb.minor_version		= swap16u(pcapngshb.minor_version);
-			pcapngshb.section_length	= swap64u(pcapngshb.section_length);
-#endif
-			if (pcapngshb.byte_order_magic == PCAPNGMAGICNUMBERBE) {
-				swap_needed = 1;
+
+			swap_needed = (pcapngshb.byte_order_magic == PCAPNGMAGICNUMBERBE);
+			swap_needed ^= !ARCH_LITTLE_ENDIAN;
+
+			if (swap_needed) {
+				pcapngbh.block_type = swap32u(pcapngbh.block_type);
 				pcapngbh.total_length = swap32u(pcapngbh.total_length);
 				pcapngshb.byte_order_magic	= swap32u(pcapngshb.byte_order_magic);
 				pcapngshb.major_version		= swap16u(pcapngshb.major_version);
 				pcapngshb.minor_version		= swap16u(pcapngshb.minor_version);
 				pcapngshb.section_length	= swap64u(pcapngshb.section_length);
 			}
+
 			aktseek = ftell(in);
 			if (pcapngbh.total_length > (SHB_SIZE + BH_SIZE + 4)) {
 				pcapng_option_walk(in, pcapngbh.total_length);
 			}
 			fseek_chk(in, aktseek + pcapngbh.total_length - BH_SIZE - SHB_SIZE, SEEK_SET);
-			continue;
-		}
-#if !ARCH_LITTLE_ENDIAN
-		pcapngbh.block_type = swap32u(pcapngbh.block_type);
-		pcapngbh.total_length = swap32u(pcapngbh.total_length);
-#endif
-		if (swap_needed == 1) {
-			pcapngbh.block_type = swap32u(pcapngbh.block_type);
-			pcapngbh.total_length = swap32u(pcapngbh.total_length);
+			memset(&pcapngidb, 0, sizeof(pcapngidb));
 		}
 
-		if (pcapngbh.block_type == 1) {
+		/* IDB, Interface Description Block */
+		else if (pcapngbh.block_type == 1) {
 			res = fread(&pcapngidb, 1, IDB_SIZE, in);
 			if (res != IDB_SIZE) {
 				fprintf(stderr, "%s: Failed to get pcapng IDB: %s\n",
 				       filename, feof(in) ? "EOF" : strerror(errno));
 				break;
 			}
-#if !ARCH_LITTLE_ENDIAN
-			pcapngidb.linktype	= swap16u(pcapngidb.linktype);
-			pcapngidb.snaplen	= swap32u(pcapngidb.snaplen);
-#endif
-			if (swap_needed == 1) {
+			if (swap_needed) {
 				pcapngidb.linktype	= swap16u(pcapngidb.linktype);
 				pcapngidb.snaplen	= swap32u(pcapngidb.snaplen);
 			}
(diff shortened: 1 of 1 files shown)


src/wpapcap2john.c
```diff
@@ -1887,182 +1888,198 @@ static int process_ng(FILE *in)
...
-		if (res == 0) {
-			break;
-		}
-		if (res != BH_SIZE) {
-			printf("failed to read pcapng header block\n");
+		block_header_t pcapng_bh;
+		interface_description_block_t pcapng_idb;
+
+		if ((res = fread(&pcapng_bh, 1, BH_SIZE, in)) != BH_SIZE)
 			break;
```
solardiz, line 1911: Previously, we'd print a message when the return is non-zero but also not BH_SIZE, now we'll be silent or print just "End of data" if at high verbosity. Is this right?

solardiz (approved):
I skimmed these commits one by one, then all at once. Saw nothing obviously wrong, except that we're still not checking returns from `fseek`.

Note that the removed duplicated read/seek block wasn't benign - we're actually changing behavior by removing it now, because the second read's offset was different - but looks like the old behavior was more likely wrong?

Sorry that I seem to have broken something with my referenced commit. Re-reading the comments I wrote at the time, I was unsure about much of this, but was somehow sure about which `caplen` to use? So was my fix completely wrong or partially wrong (so partially still needed)?

magnumripper:
> I skimmed these commits one by one, then all at once. Saw nothing obviously wrong, except that we're still not checking returns from `fseek`.

I added checking everywhere with an "fseek_chk" macro, but decided to remove it because I got confused and thought we'd get an error for truncated files (didn't risk changing behavior for that, truncated pcap file are very common). Thinking again and verifying with the man page, this is not the case, right? I'll put it back, it's easy. If nothing else it could reveal bugs.

> Note that the removed duplicated read/seek block wasn't benign - we're actually changing behavior by removing it now, because the second read's offset was different - but looks like the old behavior was more likely wrong?

The old was definitely wrong, it would skip every other packet.

> Sorry that I seem to have broken something with my referenced commit. Re-reading the comments I wrote at the time, I was unsure about much of this, but was somehow sure about which `caplen` to use? So was my fix completely wrong or partially wrong (so partially still needed)?

Due to the dupe read it was broken regardless. Also, looking again I think my intention was fine (piggy-back epb to simplify code) but it looks like it wasn't done properly anyway. Not to mention how confusing that was even for myself now. Anyway it's all good after these fixes I'm pretty sure.

src/wpapcap2john.c
magnumripper: No. I changed it during my confusion between fseek and fread behaviour - got the idea we'd either get 0 or -1 here. I will change this back!

magnumripper:
Done. I did a fix-up for both things into the "right" commit but here's the changes from what you reviewed:
```diff
$ git diff aa9b70d..e34e5da

diff --git a/src/wpapcap2john.c b/src/wpapcap2john.c
index 819ac219b..12b1fa877 100644
--- a/src/wpapcap2john.c
+++ b/src/wpapcap2john.c
@@ -94,6 +94,15 @@ static const char* const ctl_subtype[16] = {
 	"Subtype 15"
 };
 
+#define fseek_chk(s, o, w)	  \
+	do { \
+		if (fseek(s, o, w) == -1) { \
+			fprintf(stderr, "%s: Seek failed: %s, %s:%u\n", \
+			        filename, strerror(errno), __FILE__, __LINE__); \
+			exit(1); \
+		} \
+	} while (0);
+
 #if HAVE___MINGW_ALIGNED_MALLOC
 char *strdup_MSVC(const char *str)
 {
@@ -225,9 +234,9 @@ static int convert_ivs2(FILE *f_in)
 	unsigned char *p, *w;
 	int ess = -1;
 
-	fseek(f_in, 0, SEEK_END);
+	fseek_chk(f_in, 0, SEEK_END);
 	length = ftell(f_in);
-	fseek(f_in, 0, SEEK_SET);
+	fseek_chk(f_in, 0, SEEK_SET);
 
 	safe_malloc(ivs_buf, length);
 
@@ -1878,7 +1887,7 @@ void pcapng_option_walk(FILE *in, uint32_t tl)
 				break;
 		} else {
 			// Just skip unknown options
-			fseek(in, pad_len, SEEK_CUR);
+			fseek_chk(in, pad_len, SEEK_CUR);
 		}
 	}
 }
@@ -1892,8 +1901,15 @@ static int process_ng(FILE *in)
 		block_header_t pcapng_bh;
 		interface_description_block_t pcapng_idb;
 
-		if ((res = fread(&pcapng_bh, 1, BH_SIZE, in)) != BH_SIZE)
+		res = fread(&pcapng_bh, 1, BH_SIZE, in);
+		if (res == 0) {
+			break;
+		}
+		if (res != BH_SIZE) {
+			printf("%s: Failed to read pcap-ng header block: %s\n",
+			       filename, strerror(errno));
 			break;
+		}
 
 		if (verbosity >= 4)
 			fprintf(stderr, "pcap-ng block type %u\n", pcapng_bh.block_type);
@@ -1928,7 +1944,7 @@ static int process_ng(FILE *in)
 
 			memset(&pcapng_idb, 0, sizeof(pcapng_idb));
 
-			fseek(in, aktseek + pcapng_bh.total_length - BH_SIZE - SHB_SIZE, SEEK_SET);
+			fseek_chk(in, aktseek + pcapng_bh.total_length - BH_SIZE - SHB_SIZE, SEEK_SET);
 		}
 
 		/* IDB, Interface Description Block */
@@ -1944,7 +1960,7 @@ static int process_ng(FILE *in)
 				pcapng_idb.snaplen  = swap32u(pcapng_idb.snaplen);
 			}
 
-			fseek(in, pcapng_bh.total_length - BH_SIZE - IDB_SIZE, SEEK_CUR);
+			fseek_chk(in, pcapng_bh.total_length - BH_SIZE - IDB_SIZE, SEEK_CUR);
 		}
 
 		/* PB, Packet Block (deprecated) */
@@ -1980,7 +1996,7 @@ static int process_ng(FILE *in)
 				        filename, feof(in) ? "EOF" : strerror(errno));
 				break;
 			}
-			fseek(in, pcapng_bh.total_length - BH_SIZE - PB_SIZE - pcapng_pb.caplen, SEEK_CUR);
+			fseek_chk(in, pcapng_bh.total_length - BH_SIZE - PB_SIZE - pcapng_pb.caplen, SEEK_CUR);
 
 			if (pcapng_pb.caplen > 0) {
 				snap_len = pcapng_pb.caplen;
@@ -2017,7 +2033,7 @@ static int process_ng(FILE *in)
 				        filename, feof(in) ? "EOF" : strerror(errno));
 				break;
 			}
-			fseek(in, pcapng_bh.total_length - BH_SIZE - SPB_SIZE - pcapng_spb.len, SEEK_CUR);
+			fseek_chk(in, pcapng_bh.total_length - BH_SIZE - SPB_SIZE - pcapng_spb.len, SEEK_CUR);
 
 			if (pcapng_spb.len > 0) {
 				snap_len = pcapng_spb.len;
@@ -2055,7 +2071,7 @@ static int process_ng(FILE *in)
 				        filename, feof(in) ? "EOF" : strerror(errno));
 				break;
 			}
-			fseek(in, pcapng_bh.total_length - BH_SIZE - EPB_SIZE - pcapng_epb.caplen, SEEK_CUR);
+			fseek_chk(in, pcapng_bh.total_length - BH_SIZE - EPB_SIZE - pcapng_epb.caplen, SEEK_CUR);
 
 			if (pcapng_epb.caplen > 0) {
 				snap_len = pcapng_epb.caplen;
@@ -2076,7 +2092,7 @@ static int process_ng(FILE *in)
 			if (verbosity >= 4)
 				fprintf(stderr, "Skipping (unhandled block type)\n");
 
-			fseek(in, pcapng_bh.total_length - BH_SIZE, SEEK_CUR);
+			fseek_chk(in, pcapng_bh.total_length - BH_SIZE, SEEK_CUR);
 		}
 	}
 
@@ -2149,7 +2165,7 @@ static int process(FILE *in)
 	else if (main_hdr.magic_number == 0xd4c3b2a1)
 		swap_needed = 1;
 	else if (main_hdr.magic_number == PCAPNGBLOCKTYPE) {
-		fseek(in, 0, SEEK_SET);
+		fseek_chk(in, 0, SEEK_SET);
 		return process_ng(in);
 	} else {
 		if (convert_ivs2(in)) {
```

src/wpapcap2john.c
solardiz: You'll probably also want to have the error message go to stderr.

magnumripper: Ouch, thanks!

PR #5942, opened 2026-01-16
Detailed status: Have first "new" figures reflect "after restoring"

### LINKED ISSUE #5926: Detailed Status improvement
After resuming a session that had already cracked some hashes, consider that figure when printing the "new" of Succesful guesses the first time (instead of just printing the total once again). Same for other fields where possible.

magnumripper:
Before this, they just repeated the totals on first call.

Closes #5926

### DIFF
--- a/src/status.c
+++ b/src/status.c
@@ -50,6 +50,10 @@ static const char* timeFmt24 = NULL;
 static int showcand, last_count;
 double (*status_get_progress)(void) = NULL;
 
+/* Detailed status previous values */
+static struct status_main prev;
+static double prev_time;
+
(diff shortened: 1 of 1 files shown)


solardiz (approved):
Wow, so simple, and looks right. I assume you've tested that it also works.

magnumripper:
> Wow, so simple, and looks right. I assume you've tested that it also works.

Sure did, but it's indeed so simple I feared I missed something. But if you too think it appears correct, it probably is. Also, worst case scenario should be some weird output on first invocation and then all is fine, right?

PR #5960, opened 2026-03-07
JKS KeyStore - Add support for cracking 'key passwords'

### LINKED ISSUE #5959: keystore2john and the Keystore format should not be limited to files with exactly 1 key
I currently experiment a little with john. When testing Java keystore files I recognized, that the output from the script `keystore2john.py` is not recognized by john. The error message is `No password hashes loaded`, although there is an password hash extracted by the 2john script.

To minimize test setup I created an empty keystore with the password `test`, which is parsed as:
````
test.jks:$keystore$0$12$feedfeed0000000200000000$75db9b45d1619599614130770c02d5917b5f10e0$0$0$:::::test.jks
````
According to the 2john script the `75db9b45d1619599614130770c02d5917b5f10e0` is the password hash.

The output is generated, after fixing the script to support keyfiles without additional keys by init some variables:
```py
def process_file(filename):
    keysize = 0
    protectedPrivKey = b""
```

I have search the web, what output format is expected by john, but I only find a lot of tutorials.
So I am currently unsure, where the problem lies... within the script or john itself.

kholia:
Note: This was done using `gpt-5.3-codex` CLI.

Side-note: I didn't have much luck with Gemini 3.x on this task.

Fixes: https://github.com/openwall/john/issues/5959, https://github.com/openwall/john/issues/2725, https://github.com/openwall/john/issues/3219

Samples:

[KeyStore-Test.jks.txt](https://github.com/user-attachments/files/25812212/KeyStore-Test.jks.txt)
[KeyStore.jks.txt](https://github.com/user-attachments/files/25812211/KeyStore.jks.txt)

Passwords:

```
openwall
12345678
openwall123
testtest
```


solardiz:
Thank you @kholia! I'm afraid I don't have time to seriously get into this and review it properly, but we certainly do want to merge your contribution anyway. You should probably submit the samples via a PR to https://github.com/openwall/john-samples

Maybe @Borim7 @wdormann @frankenstein91 @floyd-fuh who commented on the referenced issues would like to take a look, test, and comment.

Borim7:
I have tested your code with multiple keystore files:
- Empty keystore only with master password does not work, the keystore2john script extract the hash, but john does not find it: no hash loaded
- Keystore with one password works :heavy_check_mark: 
- Keystore with multiple password works :heavy_check_mark: 

So regular use cases works as intended, corner case empty keystore only with master password can be improved ;)

Thanks for the work

kholia:
Awesome and exhaustive testing work @Borim7 - thank you!

With the latest fixes:

```
user@zion:~/repos/john/src$ ../run/john  --format=keystore hashes -w=dict
Loaded 5 password hashes with 5 different salts (keystore, Java KeyStore [SHA1 256/256 AVX2 8x])
Will run 16 OpenMP threads
Warning: Only 7 candidates buffered, minimum 128 needed for performance.
storePassword    (emptyStore.keystore)     
openwall123      (1.jks)     
12345678         (1.jks-mykey)     
openwall         (2.jks)     
testtest         (2.jks-mykey) 
```

https://stackoverflow.com/questions/37994315/how-to-create-an-empty-java-trust-store describes how to create empty KeyStore(s).

floyd-fuh:
Sorry if I'm wrong, but I'm currently just on my phone, but: it looks to me like the implementation just attacks the private key password, while ignoring all the cracking speed improvements that were found. If I'm reading it correctly it xors the entire encrypted private key for each password attempt, which is unnecessary.

This is still a huge improvement: john will never attempt to crack a password that does not protect any data. Yes, jks is so old that they thought creating a "password" (the keystore password) that is used to calculate a hash to proof that the keystore was not corrupted by byte flips so just integrity protection is a good idea. Public keys are always cleartext in there. Private keys are protected by a password that might or might not be the same as for the integrity protection.

However, it's only necessary for a cracking step to calculate a sha1 over 20 bytes plus the password. It looks like this code shoves the entire encrypted private key in the sha1 calculation, is that correct?

So either I haven't read the code correctly or hashcat will still be much much faster.

kholia:
The base implementation can go in first (once reviewed and found to be working) and then you can add your optimization patch.

Borim7:
John works now for all my test files.Thanks

kholia:
GPU testing:

```
user@zion:~/repos/john/src$ ../run/john  --format=keystore-opencl hashes -w=dict
Device 1: NVIDIA GeForce RTX 3060 Laptop GPU
Loaded 5 password hashes with 5 different salts (keystore-opencl, Java KeyStore [SHA1 OpenCL])
Note: Passwords longer than 42 [worst case UTF-8] to 127 [ASCII] rejected
LWS=64 GWS=245760 (3840 blocks) 
Warning: Only 7 candidates buffered, minimum 245760 needed for performance.
storePassword    (emptyStore.keystore)     
openwall         (2.jks)     
openwall123      (1.jks)     
testtest         (2.jks-mykey)     
12345678         (1.jks-mykey) 
```

All good!

kholia:
This is ready and working - let's get it reviewed and merged when we can - thanks!

magnumripper (approved):
I had a look at this and saw no obvious problem. I'm not very good at python though, and a thorough review of the rest would take me too long. Since this is a format and not core code I think we should just merge it.

Issue #5905, opened 2025-11-21
Update hashcat mode with new rules


magnumripper:
_Originally posted by @magnumripper in [#5553](https://github.com/openwall/john/issues/5553#issuecomment-3561093882)_
 _Originally posted by @fukc-gihtub in [#5553](https://github.com/openwall/john/issues/5553#issuecomment-3556093199)_
>> A number of other rules have been added to hashcat in the meantime: `h`, `H`, `S`, `B`, `v`.
>
>> `BNX` and `vNX`.
 
Here's what they do:
 
Hashcat | Description | Comment | JtR (clash or not)
-|-|-|-
h | convert the entire password to lowercase hex | I assume it means hello -> 68656c6c6f | free
H | convert the entire password to uppercase hex | | free
S | shift the case of each char, JtR-like | | Like it says, we have this.
B | add byte value of X at pos N, bytewise. ex: hello0 -> hello` | `BNX`, and `B50` for the example | free
v | insert char X every N chars | `vNX` | `vVNM`: "update `l` (length), then subtract M from N and assign to variable V" 

<s>For BNX and ``"print0 -> hello`"`` my first guess was that `B6N` would add 0x30 (with some `N`) to the 6th char '0' (0x30) resulting in the backtick (0x60). But we can't represent a byte value of 0x30 (48) with our 0-9A-Z (0-36) so we'd need to prepare a variable `n` first. So I presume this is not quite it, or maybe the example shouldn't be ``"print0 -> hello`"`` but some other character in the end of the result string (it's not a very good example). They haven't updated the [Hashcat wiki](https://hashcat.net/wiki/doku.php?id=rule_based_attack) and the file in `docs` is [a header](https://github.com/hashcat/hashcat/blob/master/docs/rules.txt) with no examples so I can't tell.

Like us they support `\xNN` syntax though, so maybe it's simply `B6\x30` ?</s>

magnumripper:
> Like us they support \xNN syntax though, so maybe it's simply B6\x30 

I see I suggested a similar case in https://github.com/hashcat/hashcat/pull/3850#issuecomment-1689475070 and they didn't protest.

EDIT: Oh and the example is correct as of https://github.com/hashcat/hashcat/commit/66a55d22866bde481b094b23fd59b249808a658e
```
#define RULE_OP_MANGLE_CHR_ADD           'B'  // add byte value of X at pos N, bytewise. ex: hello0 -> hello`
```
Edit again:
[...]

Issue #5817, opened 2025-08-17
rar2john works for rar 4.10 multipart rar


sjpotter:
so I downloaded rar 4.10 via archive.org image of rarlabs site, and it seems to be problematic for a multipart rar file

I create a 100MB file of 0s

```
dd if=/dev/zero of=tmp bs=1M count=100
```

I create a multipart rar archive

```
rar a -m0 -hppassword -v10m test.rar tmp
```

I run rar2john on them which generates this hashes file

```
test.part01.rar:$RAR3$*0*a2b2dd9f4894c104*6e0306bbc06f4d2160ee0f979b2a86e7:0::::test.part01.rar
test.part02.rar:$RAR3$*0*82a956ec85b7e110*c50b2aa602b45fb804aad0570c894827:0::::test.part02.rar
test.part03.rar:$RAR3$*0*74404479a24aaf24*d8b5eeeda68881ffed4b87dccf33e05e:0::::test.part03.rar
test.part04.rar:$RAR3$*0*ce8c626a7506965c*91e0398fcbaeb02e2b14e5529c6a5969:0::::test.part04.rar
test.part05.rar:$RAR3$*0*e7d3c84afa2ffb9f*f98f1baffacd94d27f01d1f5479b6f22:0::::test.part05.rar
test.part06.rar:$RAR3$*0*c1d2ccee51bf945f*7c71236a43ab772136845b1501ce042b:0::::test.part06.rar
test.part07.rar:$RAR3$*0*b5cecc223b818e94*665528870a54a1ee5244bf6130496b4f:0::::test.part07.rar
test.part08.rar:$RAR3$*0*06ec0d9bcfde0d8e*ac39db7461924d2e21d9a9a8d601a28c:0::::test.part08.rar
test.part09.rar:$RAR3$*0*22969d6c2105e890*91a23aebe5051db072e4ee5a70c9391e:0::::test.part09.rar
test.part10.rar:$RAR3$*0*d47ded8e0793a07d*1e7bcc27b6525d3b2304d6774be2d4c5:0::::test.part10.rar
test.part11.rar:$RAR3$*0*54fea8823a0a9149*bdbb61b6890d67b05bde2b9337f0aa37:0::::test.part11.rar
```

this is the first clue to me that something isn't right, as the salts are all different, I'd assume they all have to be the same, but I guess it could use the same password for each file with a different salt?

but, when I try to have josh crack it by giving it the password, it fails

```
$ john hashes --mask=password
[...]

PR #5965, opened 2026-03-19
Solve issue Update hashcat mode with new rules #5905

Serax4:
Added the options requested
https://github.com/openwall/john/issues/5905
I did not add option S because we had it already and I wasn't sure about option v so i skipped it as well.

### DIFF
--- a/src/rules.c
+++ b/src/rules.c
@@ -1424,7 +1424,35 @@ char *rules_apply(char *word_in, char *rule, int split)
 			break;
 
 		case 'v': /* assign value to numeric variable */
+			if (hc_logic)
 			{
+				/* HC rule: insert char X every N chars */
+				unsigned int n, i, out_len = 0;
+				char value;
+				char *out;
+				POSITION(n)
+				VALUE(value)
+				if (!n)
+					break;
+				GET_OUT
+				for (i = 0; i < length; i++) 
+				{
+					if (out_len >= RULE_WORD_SIZE - 1)
+						break;
+					out[out_len++] = in[i];
+					if (((i + 1) % n) == 0) {
+						if (out_len >= RULE_WORD_SIZE - 1)
+							break;
+						out[out_len++] = value;
+					}
+				}
+				out[out_len] = 0;
+				in = out;
+				length = out_len;
+				break;
+			} 
+			else 
+			{ /*Original JtR rule */
 				char var;
 				int a, s; /* may be negative */
 				VALUE(var)
@@ -1436,6 +1464,45 @@ char *rules_apply(char *word_in, char *rule, int split)
 				rules_vars[ARCH_INDEX(var)] = a - s; /* may be negative */
 			}
 			break;
+		
+		case 'h': /*convert the entire password to lowercase hex */
+			{
+				char outbuf[RULE_WORD_SIZE * 2];
+				int i;
+				if (length * 2 >= RULE_WORD_SIZE)
+					break;
+				for (i = 0; i < length; i++)
+					sprintf(&outbuf[i * 2], "%02x", (unsigned char)in[i]);
+				strcpy(in, outbuf);
+				length *= 2;
+			}
+			break;
+
+		case 'H': /*convert the entire password to uppercase hex*/
+		{
+			char outbuf[RULE_WORD_SIZE * 2];
+			int i;
+			if (length * 2 >= RULE_WORD_SIZE)
+				break;
+			for (i = 0; i < length; i++)
+				sprintf(&outbuf[i * 2], "%02X", (unsigned char)in[i]);
+			strcpy(in, outbuf);
+			length *= 2;
+		}
+		break;
+
+		case 'B': /* add byte value of X at pos N, bytewise. Format: BNX */
+		{
+			unsigned int pos;
+			unsigned char val;
+			POSITION(pos)
+			VALUE(val)
+			if (pos < length)
+			{
+				in[pos] = (unsigned char)(in[pos] + val);
+			}
+		}
+		break;
 
 /* Additional "single crack" mode rules */
 		case '1':
--- a/src/rules.c
+++ b/src/rules.c
@@ -1464,7 +1464,7 @@ char *rules_apply(char *word_in, char *rule, int split)
 				rules_vars[ARCH_INDEX(var)] = a - s; /* may be negative */
 			}
 			break;
-		
+
 		case 'h': /*convert the entire password to lowercase hex */
 			{
 				char outbuf[RULE_WORD_SIZE * 2];
@@ -1498,9 +1498,7 @@ char *rules_apply(char *word_in, char *rule, int split)
 			POSITION(pos)
 			VALUE(val)
 			if (pos < length)
-			{
 				in[pos] = (unsigned char)(in[pos] + val);
-			}
 		}
 		break;
 


magnumripper (changes requested):
Thank you! I tested each mnemonic and they work fine. However, it seems to me you messed up the indentation. In particular, the old code for '1' and '2' now has an additional tab of indentation. You shouldn't touch existing code. Also, your `case` lines (only that one line for each case) for 'H' and 'B' have one tab too much.

You would be able to implement the 'v' rule if you like (would be much appreciated) if you use the `hc_logic` variable (see the code for 'p' rule, and `doc/RULES-hashcat`).

magnumripper (changes requested):
You can't touch old/unrelated code (some formatting was still changed). I know the old code doesn't comply with current coding standards so what you did was right in a way but things like that should be in some other commit and PR.

Worse, case 'v' doesn't even build. It clashes with the existing one. Could you build and run it?
```
rules.c: In function 'rules_apply':
rules.c:1479:17: error: duplicate case value
 1479 |                 case 'v':
      |                 ^~~~
rules.c:1426:17: note: previously used here
 1426 |                 case 'v': /* assign value to numeric variable */
      |                 ^~~~
make[1]: *** [rules.o] Error 1
make: *** [default] Error 2
```
To make it work, this new code needs to be inside the existing 'v' case (with `if (hc_logic) .. else`). See `case 'p'` for an example.

Serax4:
@magnumripper Sorry this is the first time i work with a public repository I will try to be more caferul and follow the coding guidelines for the next issues (FYI the other issue I am working on is https://github.com/openwall/john/issues/5817). 
I forgot to replace the v case with the new one that's why it didn't compile.
I did implement the `if (hc_logic)... else` but i  see in `case 'p':` 
			`if (hc_logic || (*rule >= '1' && *rule <= '9')) {`
I don't really understand the OR statement here, is the idea here that if the user forgets to enable hc_logic but provides the additional arguments (required for Hashcat mode but not normally) then the case should still be interpreted in hc_logic? Case `v` (hc_logic) is in the format `vNX`. Should i check if a number (N position) and a character (X) are passed?
I just want to make sure that I understood, you said ' what you did was right in a way but things like that should be in some other commit and PR' do you want me to delete this pull request, and make a new one with a nice and tidy commit?

magnumripper:
> Sorry this is the first time i work with a public repository I will try to be more caferul and follow the coding guidelines for the next issues

No problem, you'll get the hang of it. Thanks for contributing!

> I forgot to replace the v case with the new one that's why it didn't compile. I did implement the `if (hc_logic)... else` but i see in `case 'p':` `if (hc_logic || (*rule >= '1' && *rule <= '9')) {` I don't really understand the OR statement here, is the idea here that if the user forgets to enable hc_logic but provides the additional arguments (required for Hashcat mode but not normally) then the case should still be interpreted in hc_logic?

Almost: In the 'pN' case the non-hashcat 'p' is only possible in certain contexts so in other situations we assume the hashcat 'p' even without being in hashcat mode. That hack is not applicable to 'v' so you only need `if (hc_logic)... else`. I'm sorry for giving a poor example, maybe 'x' is a better one.

> Case `v` (hc_logic) is in the format `vNX`. Should i check if a number (N position) and a character (X) are passed? 

You should, and this is similar to 'o' I think. The POSITION macro ensures it's a number while VALUE can be a character. Those macros can be confusing and are poorly documented, I always have a hard time writing new rules.

> I just want to make sure that I understood, you said ' what you did was right in a way but things like that should be in some other commit and PR' do you want me to delete this pull request, and make a new one with a nice and tidy commit?

No need for a new PR, just revert the modifications of the unrelated code and force push to this one. If/when we want to reformat the unrelated code it will be a completely separate issue & PR.

Thanks!
magnum

magnumripper:
> > Case `v` (hc_logic) is in the format `vNX`. Should i check if a number (N position) and a character (X) are passed?
> 
> You should, and this is similar to 'o' I think. The POSITION macro ensures it's a number while VALUE can be a character. Those macros can be confusing and are poorly documented, I always have a hard time writing new rules.

To clarify, in this regard your current version of 'v' looks correct already. You do use the POSITION and VALUE macros correctly as far as I understand and then you ensure the position isn't 0. The `for` takes care of `if (pos < length)` so the latter is not needed.

Serax4:
I think it should work now. It was very satisfying to see only green in the commit XD

src/rules.c
```diff
@@ -1436,6 +1464,45 @@ char *rules_apply(char *word_in, char *rule, int split)
 				rules_vars[ARCH_INDEX(var)] = a - s; /* may be negative */
 			}
 			break;
+
```
magnumripper, line 1467: Almost good now ;-)

The 'v' seems to work correctly now. But you introduced whitespace violations: Lines, including empty lines, are not allowed to have trailing whitespace.

magnumripper (changes requested):

src/rules.c
```diff
@@ -1436,6 +1464,45 @@ char *rules_apply(char *word_in, char *rule, int split)
...
+		case 'B': /* add byte value of X at pos N, bytewise. Format: BNX */
+		{
+			unsigned int pos;
+			unsigned char val;
+			POSITION(pos)
+			VALUE(val)
+			if (pos < length)
+			{
+				in[pos] = (unsigned char)(in[pos] + val);
+			}
```
magnumripper, line 1503: Nitpick: You don't use brackets for single line branches anywhere else, so maybe drop them here as well.

src/rules.c
```diff
@@ -1436,6 +1464,45 @@ char *rules_apply(char *word_in, char *rule, int split)
...
+			break;
+
+		case 'H': /*convert the entire password to uppercase hex*/
+		{
+			char outbuf[RULE_WORD_SIZE * 2];
+			int i;
+			if (length * 2 >= RULE_WORD_SIZE)
+				break;
+			for (i = 0; i < length; i++)
+				sprintf(&outbuf[i * 2], "%02X", (unsigned char)in[i]);
```
magnumripper, line 1488: The first line here (and same for 'h') aborts without converting a single letter to hex. I'm not saying it's wrong (we don't have a spec that detailed) and I don't care - I think RULE_WORD_SIZE is pretty large. Just mentioning it.

Serax4: I also don't think it's a problem, if it's acceptable now should it be merged?

magnumripper: Maybe @solardiz wants to look at this first, I'm not sure. I'm not 100% sure of the cryptic rules.c macros such as GET_OUT, and I haven't had time to delve into it yet. It's unfortunate we have so very long delay in merging stuff right now.

magnumripper (approved):

solardiz:
Thank you @Serax4 for the contribution, and I'm sorry we're slow to process it. I'm just looking at it now.

Somehow GitHub doesn't offer to run our GitHub Actions on this PR, so only CircleCI ran. Normally, there's a button for us to enable the workflows to run for PR by a new contributor, but somehow not this time.

I notice the coding and comments style here is still not exactly the same as that of surrounding code, but I suppose we may fix that after merging. We'll also need to add to `doc/RULES` and `doc/NEWS`.

We'll also need to stress-test/fuzz these commands, and I'd try to avoid `sprintf`.

@magnumripper Have you checked these commands implement exactly what hashcat's do? Per their documentation and/or code?

solardiz:
Oh, also "Remove unnecessary braces and space" as a separate commit (rather than amendment to the original commit) won't do for merging, we'll have to squash-and-merge then.

magnumripper:
> @magnumripper Have you checked these commands implement exactly what hashcat's do? Per their documentation and/or code?

Yes, they meet their examples and specs as seen in a header plus some details/clarification on Github.

PR #5966, opened 2026-03-24
Have --catch-up=SESSION add the new hash files to SESSION

magnumripper:
After the --catch-up=SESSION run finished, add its input files to SESSION's recovery file so that SESSION can simply be resumed for continuing with its old files as well as the new ones.

This can be disabled with a new --no-catch-up-add option.

### DIFF
--- a/doc/OPTIONS
+++ b/doc/OPTIONS
@@ -229,9 +229,11 @@ numbers printed on the status line mean?"
 Limit the session to running only until it reaches the candidates tried
 count of a different, existing and paused session NAME.  This can be
 used for adding more hashes to an existing job:  After the new session
-finishes, just concatenate the new hashes to the older session's hash
-file and resume the old session.  This will end up more or less as if
-both sets of hashes were in the original session to start with.
+finishes, the new hashes (file names) will be added to the older session
+file so you can just resume that session.  This will end up just as if both
+sets of hashes were in the original session to start with, with no wasted
+resources or time (assuming salted hashes).  If you do not wish the original
+session file to be touched, also add the --no-catch-up-add option.
 
 --make-charset=FILE		make a charset, overwriting FILE
 
--- a/src/options.c
+++ b/src/options.c
@@ -1,8 +1,8 @@
 /*
  * This file is part of John the Ripper password cracker,
  * Copyright (c) 1996-2025 by Solar Designer
- *
- * ...with changes in the jumbo patch, by JimF and magnum (and various others?)
+ * Copyright (c) 2009-2026, magnum
+ * Copyright (c) 2009-2018, JimF
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted.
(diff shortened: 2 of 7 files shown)


magnumripper:
This is tested in all relevant ways I could think of, such as with `--fork` and with any combination of single/multiple files.

src/options.c
```diff
@@ -139,6 +139,7 @@ static struct opt_entry opt_list[] = {
 	{"restore", FLG_RESTORE_SET, FLG_RESTORE_CHK, 0, ~FLG_RESTORE_SET & ~GETOPT_FLAGS, OPT_FMT_STR_ALLOC, &options.session},
 	{"session", FLG_SESSION, FLG_SESSION, FLG_CRACKING_SUP, OPT_REQ_PARAM, OPT_FMT_STR_ALLOC, &options.session},
 	{"catch-up", FLG_ONCE, 0, 0, OPT_REQ_PARAM, OPT_FMT_STR_ALLOC, &options.catchup},
+	{"catch-up-add", FLG_ONCE, 0, 0, USUAL_REQ_CLR | OPT_TRISTATE, NULL, &options.catchup_add},
```
solardiz, line 142: Why is this `OPT_TRISTATE`? (Just a question, not implying there's anything wrong.)

solardiz (approved):

doc/OPTIONS
```diff
@@ -229,9 +229,11 @@ numbers printed on the status line mean?"
...
 count of a different, existing and paused session NAME.  This can be
 used for adding more hashes to an existing job:  After the new session
-finishes, just concatenate the new hashes to the older session's hash
-file and resume the old session.  This will end up more or less as if
-both sets of hashes were in the original session to start with.
+finishes, the new hashes (file names) will be added to the older session
+file so you can just resume that session.  This will end up just as if both
+sets of hashes were in the original session to start with, with no wasted
+resources or time (assuming salted hashes).  If you do not wish the original
+session file to be touched, also add the --no-catch-up-add option.
```
solardiz, line 236: Just noting that we don't have the `--no-catch-up-add` option documented in its own section in `OPTIONS`, we just mention it here. This may be fine. Maybe we already do similar for other obscure options?

src/recovery.c
```diff
@@ -763,3 +763,82 @@ uint64_t rec_read_cands(char *session)
 
 	return ret;
 }
+
+void rec_add_files(char *session)
```
solardiz, line 767: This `rec_add_files` is quite some code and quite some duplication of logic (filename suffix, locking) with our existing rec file writing code. As an alternative, maybe we could be adding the extra files to `rec_argv` and `rec_argc`, and then our existing code would just happen to write them into the rec file at next sync point (which we could trigger explicitly if needed).

src/recovery.c
```diff
@@ -763,3 +763,82 @@ uint64_t rec_read_cands(char *session)
...
+	/* Rewrite header */
+	fprintf(catchup_file, "%s\n%d\n", magic, argc + (int)options.passwd->count);
+
+	/* Append new file entries */
+	struct list_entry *current;
+	for (current = options.passwd->head; current; current = current->next)
+		fprintf(catchup_file, "%s\n", current->data);
+
+	/* Append original tail */
+	fprintf(catchup_file, "%s", catchup_data + offset);
```
solardiz, line 830: Maybe use `fwrite`. We read this data with `fread`.

src/recovery.c
```diff
@@ -763,3 +763,82 @@ uint64_t rec_read_cands(char *session)
...
+	fclose(catchup_file);
```
solardiz, line 831: In our usual code, we don't just `fclose`. We check with `ferror` first (to detect any errors of `fprintf`), then `fflush` (with its return value checked), `ftruncate` (ditto), and `fsync` (ditto). OK, `ftruncate` may be unneeded here if we're sure we can only increase the size, but the rest are probably good to have?

src/recovery.c
```diff
@@ -763,3 +763,82 @@ uint64_t rec_read_cands(char *session)
...
+	const char *full_rec_name = path_expand(rec_name);
+	if (unlink(full_rec_name))
+		pexit("unlink: %s", full_rec_name);
+
+	/* Close this session's file */
+	if (rec_file) {
+		if (fclose(rec_file))
+			pexit("fclose");
+		rec_file = NULL;
+	}
```
solardiz, line 843: I'm confused - why do we delete and close this session's file? I don't seem to fully understand the logic this implements. Also, what would these two filenames be in typical usage?

solardiz (commented):
So I thought of just merging this quickly, but then I saw it's complicated. We can still just merge or we can discuss first... which I guess neither of us has much time for now - I certainly don't. What do you suggest?

src/recovery.c
solardiz: Regarding "sure we can only increase the size", I'm confused as to whether we're rewriting the same file we read from, or some other file. If it's a different file, then its previous content may have been unrelated and larger, and then we'd need to truncate.

solardiz: Oh, indeed same file - this code literally seeks to offset 0 again. Then maybe we're fine without truncation.

Issue #3896, opened 2019-05-07
Consistently (don't) split long source file lines


solardiz:
@jfoug's commit d89568a411a8ea2ba5feea944048d9d637f8a0ce split a long test vector that's now in openbsdsoftraid_variable_code.h into multiple lines. There are two issues with that:

1. The split is through use of backslashes inside the string constant, whereas I think it'd be more reliable to close the string on one line and re-open it on the next (and let the compiler concatenate those partial strings) so that the behavior won't depend e.g. on lack of trailing whitespace.

2. We've since got similarly long other lines into the same source file, which are not split, and even longer lines into other source files. So splitting just that one line makes no sense.

```
$ egrep -l '^.{4000}' *.[ch] 
agilekeychain_common_plug.c
bks_fmt_plug.c
diskcryptor_common_plug.c
dmg_common_plug.c
keystore_common_plug.c
openbsdsoftraid_variable_code.h
opencl_diskcryptor_aes_fmt_plug.c
opencl_electrum_modern_fmt_plug.c
opencl_pfx_fmt_plug.c
pfx_fmt_plug.c
```
```
$ egrep -l '^.{5000}' *.[ch] 
dmg_common_plug.c
keystore_common_plug.c
opencl_electrum_modern_fmt_plug.c
opencl_pfx_fmt_plug.c
pfx_fmt_plug.c
```

We should probably either un-split that one line or split all similarly long lines (and do it in the more reliable way I suggest above). It also makes sense to first un-split that one line for consistency, and then split all relevant lines in a separate commit/PR.

magnumripper:
I opened a PR which currently just unsplit them (I also found such lines in pkzip format).

If we're caring to split all very long lines at all, what would be a sensible limit? 1024, 512, or perhaps 160 characters? The softraid split is something like 1475 characters, the pkzip ones merely 181.

solardiz:
> If we're caring to split all very long lines at all, what would be a sensible limit?

Perhaps no lower, or not much lower, than Jim's split in openbsdsoftraid_variable_code.h was. Maybe 1000 or 2000 or 4000, assuming that a compiler's line length limit is likely a power of 2.

magnumripper:
Perhaps we shouldn't close this, but IMHO we can now remove the milestone. Having said that I'll play with some one-liners for doing this automagically.

Josue-Crz:
I'd like to take this on. Is this still open for contribution?

solardiz:
[...]

PR #5967, opened 2026-03-26
Consistently (don't) split long source file lines

Josue-Crz:
Issue #3896 

I built upon magnumripper's previous merged PR #3908 in which I soft split each part of the test vector divided by the delimiter ('$'). Enforcing a soft split per section of the vector test string within a limit of 80 characters per line, increasing readability toward meaning to each delimiter value of the long string vector test. 

As well, I propose a max_line_length of 80 within future header files in .editorconfig to increase readability across long test vectors.

**Testing my changes to openbsd-softraid**:

**~/Documents/Github_Repos/john-JosueFork/run$** ./john --test --format=openbsd-softraid

Output:

**Will run 16 OpenMP threads
Benchmarking: OpenBSD-SoftRAID [PBKDF2-SHA1 256/256 AVX2 8x]... (16xOMP) DONE
Speed for cost 1 (kdf) of 1, cost 2 (iteration count) of 8192
Raw:    9170 c/s real, 588 c/s virtual**


Josue-Crz:
@solardiz Could you please look at this when you have the chance? I just want to know I'm in the right direction when tackling this issue. 

As well, I'm familiar Magnumripper's approach in PR #3908 with removing escape characters in the long test vectors, should I combine Magnumripper's approach to the issue of removing escape characters with the styling I have in my current PR draft?

magnumripper:
Thanks! I have some objections but don't let that discourage you.

> I soft split each part of the test vector divided by the delimiter ('$'). Enforcing a soft split per section of the vector test string within a limit of 80 characters per line

I can see some merit in splitting at the delimiter but that's not in #3896 and we'd want to discuss it first. I'm not sure I like it. Would we also do so with short test vectors that otherwise need no splitting?

Also, we definitely do not want the splits as short as 80 characters, way longer:

#3896 was originally only about replacing backslash continuation with multiple string constants, like in #3908, and split (or not) consistently. Later (in same issue) we decided to split lines at "no lower, or not much lower, than Jim's split in openbsdsoftraid_variable_code.h was. Maybe 1000 or 2000 or 4000,". The original split was ~1480, so I think we should go for ~2000.

Josue-Crz:
@magnumripper Thank you so much for the pointers from my previous commits! Looking back on it is unnecessary to keep soft split at the delimiter especially for shorter test vectors and it'd be a hassle to refactor the smaller vectors when not necessary.

In my latest commit I shot at the ideal ~2000 character per line as previously stated for the openbsd test vector is this ideally what it should look like?

As well, I'll start getting to work on other test vectors and approach it toward the removal of backslashes on string constants such as your work on #3908. Once again thank you very much!

PR #5975, opened 2026-04-13
fix(ssh2john): correct bcrypt rounds assignment and update exception …

dejanazul:
…handling

This commit addresses two issues in the OpenSSH private key extraction process:
1. Exception Handling: Updated the decoding try-except block to catch `binascii.Error` directly, ensuring decode failures are handled gracefully in Python 3 environments.
2. Logic Bug: Fixed a fallback assignment issue for OpenSSH `bcrypt_pbkdf` keys. Replaced an incorrect equality check (`rounds == 16`) with an assignment (`rounds = 16`) so that the iteration count defaults correctly when initially read as 0.

### DIFF
--- a/run/ssh2john.py
+++ b/run/ssh2john.py
@@ -118,7 +118,7 @@ def read_private_key(filename):
         try:
             data = ''.join(lines[start:end]).encode()
             data = base64.b64decode(data)
-        except base64.binascii.Error:
+        except binascii.Error:
             e = sys.exc_info()[1]
             raise Exception('base64 decoding error: ' + str(e))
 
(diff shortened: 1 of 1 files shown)


magnumripper (approved):

dejanazul (commented):
🙌

run/ssh2john.py
```diff
@@ -118,7 +118,7 @@ def read_private_key(filename):
         try:
             data = ''.join(lines[start:end]).encode()
             data = base64.b64decode(data)
-        except base64.binascii.Error:
+        except binascii.Error:
```
solardiz, line 121: I wonder if this was maybe a Python 2 vs. 3 modules thing? I think we had this script try to work with either version, as indicated by the shebang not specifying a version.

solardiz: Oh, you actually wrote "in Python 3 environments." So I wonder what this means for Python 2, then? If this script does not support Python 2 anymore (which would be mildly unfortunate), then we need to update the shebang to python3.

PR #6005, opened 2026-07-05
Add RAKP MD5/SHA256 format support

hdm:
The IPMI RAKP+ protocol supports a variety of ciphers, john supports RAKP+ SHA1 today, and this PR implements MD5 and SHA256, which altogether handle nearly 100% of IPMI implementations.

# Benchmarks

Benchmarks for the four new formats (`RAKP-MD5`, `RAKP-SHA256`, and their `-opencl`
counterparts), alongside the existing `RAKP` (HMAC-SHA1) format for reference.
Suitable for the PR description.

## Test system

| | |
|---|---|
| Machine | Apple M3 Max — 16 cores (12 performance + 4 efficiency), 128 GB RAM |
| OS | macOS 26.5.1 (build 25F80), Darwin 25.5.0, arm64 |
| GPU | Apple M3 Max (integrated), via Apple OpenCL 1.2 |
| John | 1.9.0-jumbo-1 +bleeding (`feat/additional-rakp-formats`), gcc 16.1.0 |
| Build | `64-bit arm ASIMD AC OMP OPENCL`; SIMD interleave MD5:2 SHA1:1 SHA256:1 |

## Methodology

`john --test` per format (self-test + benchmark). All formats **PASS** self-test.
`c/s` = candidate×salt combinations per second. "Many salts" is the salted-format
throughput that matters for real RAKP dumps (each host has a unique salt); "Only
one salt" is the single-target rate. CPU runs use all 16 cores (OpenMP).

### CPU (SIMD, OpenMP 16 threads)

| Format | Algorithm | Many salts (c/s) | One salt (c/s) |
|--------|-----------|-----------------:|---------------:|
| RAKP (existing) | HMAC-SHA1   `128/128 ASIMD 4x`  |  75.1 M | 29.5 M |
| **RAKP-MD5**    | HMAC-MD5    `128/128 ASIMD 4x2` | **103.8 M** | 32.2 M |
| **RAKP-SHA256** | HMAC-SHA256 `128/128 ASIMD 4x`  |  **49.3 M** | 21.6 M |

### OpenCL (Apple M3 Max GPU)

| Format | Algorithm | Many salts (c/s) | One salt (c/s) | vs CPU (many-salts) |
|--------|-----------|-----------------:|---------------:|--------------------:|
| RAKP-opencl (existing) | HMAC-SHA1   | 849.7 M | 153.2 M | 11.3× |
| **RAKP-MD5-opencl**    | HMAC-MD5    | **329.0 M** | 162.7 M | 3.2× |
| **RAKP-SHA256-opencl** | HMAC-SHA256 | **153.2 M** | 97.3 M | 3.1× |

## Notes

- **SIMD speedup (CPU):** the SIMD MD5 path is ~2.8× the scalar fallback
  (103.8 M vs ~37 M measured with `SIMD_COEF_32` undefined on the same host).
- **arm64 SHA256 caveat:** on this Apple-Silicon host the *non-SIMD* SHA256 path
  can actually be faster than the 4-way NEON SIMD path, because Darwin's libc /
  the scalar path uses the ARMv8 SHA-256 crypto instruction (hardware, single
  stream) which beats software 4-way NEON. This mirrors John's existing
  `HMAC-SHA256` format behaviour on arm64; on x86-64 without SHA-NI the SIMD
  path wins, so it is kept.
- **GPU:** OpenCL is a clear win here (MD5 3.2×, SHA1 11×) and should scale
  much further on discrete NVIDIA/AMD GPUs. The kernels are scalar
  one-work-item-per-candidate using the shared `opencl_hmac_{md5,sha256}.h`
  helpers.
- **Loading many hashes:** the formats implement `binary_hash`/`get_hash`
  (RAKP salts are bounded, so `FMT_HUGE_INPUT` is intentionally not set) — this
  avoids the "excessive partial hash collisions" degradation when loading large
  RAKP hash sets (e.g. tens of thousands of unique-salt hashes from a scan).

### DIFF
--- a/doc/NEWS
+++ b/doc/NEWS
@@ -488,6 +488,11 @@ Major changes from 1.9.0-jumbo-1 (May 2019) in this bleeding-edge version:
 - Add support to keepass2john for the current format of KeePass XML keyfiles.
   The formats needed no changes.  [magnum; 2025]
 
+- Added RAKP-MD5 and RAKP-SHA256 CPU and OpenCL formats for IPMI 2.0 RAKP
+  (RMCP+) challenge-response hashes in the existing "$rakp$<salt>$<hmac>"
+  syntax, where digest length selects MD5 (32), SHA1 (40), or SHA256 (64).
+  [HD Moore; 2026]
+
 
 Major changes from 1.8.0-jumbo-1 (December 2014) to 1.9.0-jumbo-1 (May 2019):
 
--- a/run/opencl/rakp_md5_kernel.cl
+++ b/run/opencl/rakp_md5_kernel.cl
@@ -1,0 +1,39 @@
+/*
+ * OpenCL kernel for IPMI 2.0 RAKP (RMCP+) HMAC-MD5 (RAKP auth algorithm 2).
+ *
+ * Copyright (c) 2026 HD Moore, released to the general public under the
+ * following terms:
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted.
+ *
+ * One work-item per candidate. Keys are packed (key_array + index[gid]); the
+ * RAKP salt (raw bytes + length) is the HMAC message. Uses the shared
+ * opencl_hmac_md5 helper so the HMAC construction stays correct and simple.
+ */
+
+#include "opencl_misc.h"
+#define HMAC_KEY_TYPE __global const
+#include "opencl_hmac_md5.h"
+
+__kernel void rakp_md5_kernel(__global const uchar *salt,
+                              const uint salt_len,
+                              __global const uint *key_array,
+                              __global const uint *index,
+                              __global uint *digest)
+{
+	uint gid = get_global_id(0);
+	uint base = index[gid];
+	uint key_len = base & 63;
+	__global const uchar *key = (__global const uchar *)(key_array + (base >> 6));
+	uchar saltbuf[128];
+	uint out[4];
+	uint i;
+
+	for (i = 0; i < salt_len; i++)
+		saltbuf[i] = salt[i];
+
+	hmac_md5(key, key_len, saltbuf, salt_len, out, 16);
+
+	for (i = 0; i < 4; i++)
+		digest[gid * 4 + i] = out[i];
+}
--- a/run/opencl/rakp_sha256_kernel.cl
+++ b/run/opencl/rakp_sha256_kernel.cl
@@ -1,0 +1,39 @@
+/*
+ * OpenCL kernel for IPMI 2.0 RAKP (RMCP+) HMAC-SHA256 (RAKP auth algorithm 3).
+ *
+ * Copyright (c) 2026 HD Moore, released to the general public under the
+ * following terms:
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted.
+ *
+ * One work-item per candidate. Keys are packed (key_array + index[gid]); the
+ * RAKP salt (raw bytes + length) is the HMAC message. Uses the shared
+ * opencl_hmac_sha256 helper so the HMAC construction stays correct and simple.
+ */
+
+#include "opencl_misc.h"
+#define HMAC_KEY_TYPE __global const
+#include "opencl_hmac_sha256.h"
+
+__kernel void rakp_sha256_kernel(__global const uchar *salt,
+                                 const uint salt_len,
+                                 __global const uint *key_array,
+                                 __global const uint *index,
+                                 __global uint *digest)
+{
+	uint gid = get_global_id(0);
+	uint base = index[gid];
+	uint key_len = base & 63;
+	__global const uchar *key = (__global const uchar *)(key_array + (base >> 6));
+	uchar saltbuf[128];
+	uint out[8];
+	uint i;
+
+	for (i = 0; i < salt_len; i++)
+		saltbuf[i] = salt[i];
+
+	hmac_sha256(key, key_len, saltbuf, salt_len, out, 32);
+
+	for (i = 0; i < 8; i++)
+		digest[gid * 8 + i] = out[i];
+}
--- a/src/opencl_rakp_md5_fmt_plug.c
+++ b/src/opencl_rakp_md5_fmt_plug.c
@@ -1,0 +1,384 @@
+/*
+ * OpenCL IPMI 2.0 RAKP (RMCP+) HMAC-MD5 (RAKP auth algorithm 2).
+ *
+ * Host based on opencl_rakp_fmt_plug.c (HMAC-SHA1 RAKP, (C) 2013 Harrison Neal,
+ * magnum, Dhiru Kholia). Copyright (c) 2026 HD Moore. Released under GPLv2 like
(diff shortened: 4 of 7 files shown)


solardiz:
Thank you @hdm, it's great to see you contribute!

Please add a `doc/NEWS` item to the end of the list current list near line 490. Please preserve the double blank line after that sub-list.

We also need your copyright statements, not just abstract "(c) 2026" without specifying who the copyright holders are.

For such minor revisions within a PR, we normally amend the initial PR commit(s) and force-push, not add more commits. We only use separate commits for the history we want to preserve in the repo.

It's a bit puzzling the new MD5 OpenCL format appears slower than the older SHA-1 per your benchmarks. Perhaps the new SHA-256 isn't optimal either. But we can deal with that separately.

It appears that the `FMT_HUGE_INPUT` removal and binary hash function additions would also be helpful for the existing RAKP formats, perhaps as a separate commit. @magnumripper Do you agree?

hdm:
Thank you! I'll clean up a bit and squash/repush.

hdm:
On the usage side, two small things jumped out:
1) The current format names for SHA1 are simply `RAKP`, which can be confusing:
```
 src $ ../run/john ~/Downloads/large.inp
Warning: detected hash type "RAKP-MD5", but the string is also recognized as "RAKP-MD5-opencl"
Use the "--format=RAKP-MD5-opencl" option to force loading these as that type instead
Warning: only loading hashes of type "RAKP-MD5", but also saw type "RAKP"
Use the "--format=RAKP" option to force loading hashes of that type instead
Warning: only loading hashes of type "RAKP-MD5", but also saw type "RAKP-opencl"
Use the "--format=RAKP-opencl" option to force loading hashes of that type instead
Warning: only loading hashes of type "RAKP-MD5", but also saw type "RAKP-SHA256"
Use the "--format=RAKP-SHA256" option to force loading hashes of that type instead
Warning: only loading hashes of type "RAKP-MD5", but also saw type "RAKP-SHA256-opencl"
Use the "--format=RAKP-SHA256-opencl" option to force loading hashes of that type instead
Loaded 16474 password hashes with 16474 different salts (RAKP-MD5, IPMI 2.0 RAKP (RMCP+) [HMAC-MD5 128/128 ASIMD 4x2])
Will run 16 OpenMP threads
Almost done: Processing the remaining buffered candidate passwords, if any.
Enabling duplicate candidate password suppressor using 256 MiB
```

2) This is likely a JtR design decision, but for multi-format/hash inputs, is there a better way to steer users to the most useful implemented hash? MD5 may be that case given that it is a huge portion of the input.

Is it worth aliasing `RAKP` to `RAKP-SHA1` or similar?

solardiz:
> The current format names for SHA1 are simply RAKP, which can be confusing:

We could maybe rename the existing RAKP formats to RAKP-SHA1, or maybe roll this new functionality into them instead of adding separate formats (maybe do this as a separate PR), but then we'd also need to add "tunable cost" reporting for underlying hash type, and benchmarking would become harder (default benchmark runs would include just one hash type, or worse speed for a mix of them - but we can avoid that by placement of test vectors). We're currently inconsistent on whether we add more formats or add alternative code paths into existing formats. I'm not saying this is what we want to do - just sharing the options we have.

> Warning: only loading hashes of type "RAKP-MD5", but also saw type "RAKP"
(and so on)

This means users will get false negatives if they run with the wrong format. Is there any indication anywhere on which hash type is actually used, or are users supported to try all 3?

> This is likely a JtR design decision, but for multi-format/hash inputs, is there a better way to steer users to the most useful implemented hash? MD5 may be that case given that it is a huge portion of the input.

Do you mean RAKP-MD5 is more commonly the right choice than the rest? If so, you happened to make it the default already due to alphabetical order.

Or did you mean raw-MD5 (unrelated to this PR)? If so, I agree we should ideally favor it over the LM hash default inherited from the core tree (which didn't have other 32 hex hashes).

hdm:
>I'm not saying this is what we want to do - just sharing the options we have.

Thank you! Happy to take a shot at any of these if it would help.

>This means users will get false negatives if they run with the wrong format. Is there any indication anywhere on which hash type is actually used, or are users supported to try all 3?

The source tooling will be open source in about a month (BH/DC) and is easy to adjust to produce a better JtR format if a different structure works better. 

At the moment everything I'm working with uses the hash length as the indicator (md5=16,sha1=20,sha256=64), and if the user picks the wrong one, the other types are ignored. Hashcat supports md5/sha1 (but not yet sha256), but requires the correct `-m` flag (7300 vs 7350) to select matching hashes, and ignores non-matching hash types. A specific hash identifier would be easy to implement if that saves work on this side.

>Do you mean RAKP-MD5 is more commonly the right choice than the rest? If so, you happened to make it the default already due to alphabetical order.

Understood - that's a great default in this case (MD5 is more common than other types).

Is there anything that you would like to changed in this PR?

solardiz:
> At the moment everything I'm working with uses the hash length as the indicator (md5=16,sha1=20,sha256=64), and if the user picks the wrong one, the other types are ignored. Hashcat supports md5/sha1 (but not yet sha256), but requires the correct `-m` flag (7300 vs 7350) to select matching hashes, and ignores non-matching hash types. A specific hash identifier would be easy to implement if that saves work on this side.

Identified by hash length should be fine, and I see we do check `len != BINARY_SIZE * 2` in `valid` in these formats - but then I'm puzzled by your sample output showing `but also saw type` for other RAKP formats. If this `valid` works correctly, then I would expect no such detection/suggestions. Let's figure this out.

solardiz:
> I'm puzzled by your sample output showing `but also saw type` for other RAKP formats.

Oh, I think I get it now - you have a mix of many types in the input file, so these messages are produced for other lines. Then everything probably works as it should, and my earlier concern about false negatives does not apply.

solardiz (approved):
LGTM

solardiz:
> The source tooling will be open source in about a month (BH/DC)

Is this something you could contribute later as well, like a rakp2john tool? Maybe also sample input for it to https://github.com/openwall/john-samples?

> Is there anything that you would like to changed in this PR?

I think we can merge it as-is already. We'll need to test the OpenCL formats more thoroughly then.

As to extra changes, maybe rename the old RAKP formats to include SHA1 in the names, and add the binary hash functions to them - but it's good to do this via a separate PR if you feel like it.

Once again, thank you very much for the contribution!

solardiz:
@magnumripper Would you like to review this before we merge?

magnumripper (approved):
I've only had a quick look, but it looks great.

magnumripper:
> It appears that the FMT_HUGE_INPUT removal and binary hash function additions would also be helpful for the existing RAKP formats, perhaps as a separate commit. @magnumripper Do you agree?

Yes it looks like it. While the ciphertext length varies, the limit is like 5x the length of the longest test vector.

solardiz:
Thank you once again @hdm. This is now merged. I've created issued #6017 for possible related cleanup of our old formats.

hdm:
Thank you! I have a TODO for the sample hashes

PR #6007, opened 2026-07-07
listconf.c: Optimize format list output (--list=formats) for pipes

magnumripper:
When scripting or writing "oneliners" the commas are just in the way and requires special handling.  We now skip them and list one format per line if output is not a terminal.
```
$ ../run/john -list=formats -form=*crypt
descrypt, bsdicrypt, md5crypt, bcrypt, scrypt, adxcrypt, AxCrypt, BestCrypt, 
sha1crypt, sha256crypt, sha512crypt, django-scrypt, streebog256crypt, 
streebog512crypt, gost94crypt, Raw-SHA1-AxCrypt, sm3crypt, crypt
18 formats
```
```
$ ../run/john -list=formats -form=*crypt | wc -l
      18
```
```
$ ../run/john -list=formats -form=*crypt | grep ^sha
sha1crypt
sha256crypt
sha512crypt
```

### DIFF
--- a/src/listconf.c
+++ b/src/listconf.c
@@ -592,8 +592,9 @@ void listconf_parse_late(void)
 	if (!strcasecmp(options.listconf, "formats")) {
 		struct fmt_main *format;
 		int column = 0, dynamics = 0;
-		int grp_dyna, total = 0, add_comma = 0;
+		int grp_dyna, total = 0;
 		char *format_option = options.format ? options.format : options.format_list;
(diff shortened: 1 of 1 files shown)


solardiz (approved):
I assume you tested this, so don't review the conditions very closely. I'd rename/invert `stdout_is_pipe` to `stdout_is_tty` (so also invert its checks later). Not a tty does not imply pipe - it can be various non-tty things.

PR #6029, opened 2026-08-13
dmg: support AES-encrypted v2 key blobs

### LINKED ISSUE #6028: dmg2john extracts uncrackable hashes from current DMG files
During the 2026 Crack Me If You Can contest, challenge 4 contained four password-protected DMG files. dmg2john extracts hashes from all four without reporting an error, but the resulting hashes cannot be cracked even when the correct password is included in the wordlist.

Tested with bleeding-jumbo at commit 9a336d800a091bec9650c29282485145f31c9ffc (latest as of this writing).

Challenge files can be downloaded from: https://contest-2026.korelogic.com/downloads/

## Extract challenge 4

```
gpg --batch --passphrase '@u:302wbl:50[;FRiFt2' -d cmiyc_2026_challenge_4.tar.xz.pgp > cmiyc_2026_challenge_4.tar.xz
tar xf cmiyc_2026_challenge_4.tar.xz
```

This produces:

```
its_not_that_hard.dmg
lDK9QqIzhwk.dmg
start_with_this_one.dmg
wow_its_hot.dmg
```

## Extract the hashes and create wordlist

`run/dmg2john *.dmg > dmg.john`

The password **d3fc0n** is known to be correct for at least one of these DMGs. Create a small wordlist containing it and variants:

printf 'defcon\nd3fc0n\n' > defcon.lst
run/john --stdout -w:defcon.lst -rule:single | run/unique defcon-expanded.lst

### Verify d3fc0n is in the expanded list
grep -x d3fc0n defcon-expanded.lst

## Run john
```
$ run/john -pot:defcon.pot -w:defcon-expanded.lst dmg.john
Warning: detected hash type "dmg", but the string is also recognized as "dmg-opencl"
Use the "--format=dmg-opencl" option to force loading these as that type instead
Loaded 4 password hashes with 4 different salts (dmg, Apple DMG [PBKDF2-SHA1 256/256 AVX2 8x 3DES/AES])
Loaded hashes with cost 1 (iteration count) varying from 400000 to 476190
Cost 2 (version) is 2 for all loaded hashes
Will run 12 OpenMP threads
0g 0:00:00:46 DONE (2026-08-13 04:03) 0g/s 43.27p/s 173.1c/s 173.1C/s Defcon1944..D3fc0n1900
Session completed
```

Hans Leininger (hlein) wrote in the post contest discord:

> Minga probably knows more details, but IIRC it's bad in a bad way: it will extract a hash from current DMG formats w/o complaint, but what it spits out is not actually crackable. So you'll think you succeeded, but can't crack a DMG with password d3fc0n

Same results with dmg2john.py

Either these DMG formats should be supported correctly, or dmg2john should detect unsupported/incompatible DMG formats and report an error instead of silently producing uncrackable hashes.

kholia:
Recent macOS releases can wrap v2 DMG key blobs with AES-192-CBC (CSSM_ALGID_AES) instead of 3DES. Teach both extractors to recognize this parameter set and emit a compact $dmg$3 hash, while rejecting unknown blob encryption combinations.

Add CPU and OpenCL verification by decrypting the final key-blob block and checking the CKIE marker and PKCS#7 padding. Include a known-password test vector and retain existing v1/v2 behavior.

Fixes https://github.com/openwall/john/issues/6028


GraceWaddell:
gmail reset

PR #6034, opened 2026-09-08
V4 and V5 encryption key length logic

### LINKED ISSUE #6033: pdf2john.py incorrectly defaults to 40-bit key length in V4/R4 PDF
Hi,
I found a V4/R4 PDF where the encryption dictionary does not contain a top-level /Length, but the applicable crypt filter does:

`qpdf` output for the encryption object:
```
/V 4
/R 4
/CF <<
  /StdCF <<
    /AuthEvent /DocOpen
    /CFM /AESV2
    /Length 16
  >>
>>
/StmF /StdCF
/StrF /StdCF
```

There is no top-level `/Length` entry.

Running:

`pdf2john.py sample.pdf`

produces:

`$pdf$4*4*40*-4*...`

The `40` seems to come from:

`self.length = self.encrypt_dict.get("/Length", 40)`

However, for this V4/R4 PDF the applicable `/StdCF` crypt filter has:

`/Length 16`

which corresponds to a 128-bit encryption key.

For comparison, `qpdf --show-encryption --password='...' sample.pdf` reports:

```
R = 4
stream encryption method: AESv2
string encryption method: AESv2
file encryption method: AESv2
```

The resulting 40 in the `$pdf$` hash causes john to interpret the PDF as having a 40-bit key, even though the actual encryption uses AES-128.
This then results in the hash not being able to be cracked.

Manually chaning the length to 128 instead of 40 fixes the issue and the password was correctly found by john.

PDF metadata:

```
Creator:     Aspose Pty Ltd.
Producer:    PDFsharp 6.2.0 (Original: Aspose.PDF for .NET 26.3.0)
PDF version: 1.7
Encryption:  AES (Revision 4 / V4)
```

The reference also states that `/Length` is optional:
https://opensource.adobe.com/dc-acrobat-sdk-docs/pdfstandards/pdfreference1.7old.pdf - section 3.5, page 117 
> (Optional; PDF 1.4; only if V is 2 or 3) The length of the encryption key, in bits. The value
must be a multiple of 8, in the range 40 to 128. Default value: 40.

**Expected behavior**

For V4/R4 PDFs using a crypt filter, `pdf2john.py` should determine the effective key length from the applicable crypt filter when `/Length` is not present at the top level, rather than unconditionally falling back to the 40-bit default.

KR Alex

alex-barth:
Fix:
pdf2john.py reports wrong key length for V4/V5 PDFs missing top-level `/Length`

Fixes #6033

For V4/V5 crypt-filter encryption, `/Length` at the top level is optional and often omitted — the real key length lives in the crypt filter's own `/Length` instead. The current code unconditionally defaults to 40 bits when it's missing, causing John to derive the wrong key size (e.g. AES-128 reported as 40-bit).

This adds a fallback to the crypt filter's /Length (via `/StmF` / `/StrF`) when the top-level entry is absent, plus a warning if both are present but disagree. V1–V3 behavior is unchanged.

Edit: code generated by claude, thats why the comments are so verbose. Manually checked it afterwards and ran it against the pdfsharp test files mentioned in #6033

### DIFF
--- a/run/pdf2john.py
+++ b/run/pdf2john.py
@@ -71,9 +71,92 @@ class PdfHashExtractor:
                 raise RuntimeError("File not encrypted")
 
             self.algorithm: int = self.encrypt_dict.get("/V")
-            self.length: int = self.encrypt_dict.get("/Length", 40)
+            self.length: int = self._get_key_length()
             self.permissions: int = self.encrypt_dict["/P"]
             self.revision: int = self.encrypt_dict["/R"]
+
+    def _get_key_length(self) -> int:
+        """
+        Determine the effective encryption key length, in bits.
+
+        For V1-V3 encryption, the key length is given directly by the
+        top-level /Length entry (in bits), defaulting to 40 if absent
+        (per PDF 1.7 spec section 7.6.1, Table 20).
+
+        For V4/V5 encryption, PDFs use crypt filters (/CF). The
+        top-level /Length entry is not formally part of the spec for
+        these versions and is frequently omitted -- the *effective*
+        key length is governed by the applicable crypt filter's own
+        /Length entry instead (given in *bytes*, per Table 25).
+        Falling back unconditionally to 40 when /Length is absent
+        silently mis-reports e.g. AES-128 files as 40-bit, which then
+        causes John to derive the wrong key size and fail to crack an
+        otherwise crackable hash.
+
+        See: https://github.com/openwall/john/issues/6033
+
+        Some writers include a top-level /Length on V4/V5 files even
+        though it isn't required. When present, it should agree with
+        the crypt filter's /Length -- if it doesn't, the crypt filter
+        value is treated as authoritative (it's what actually governs
+        the key size for these versions) and a warning is logged, since
+        a mismatch may indicate an unusual or malformed file worth a
+        closer look.
+        """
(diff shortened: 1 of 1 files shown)


solardiz:
Hmm, our `[ci/circleci: arm](https://circleci.com/gh/openwall/john/33376)` job got stuck with:

```
The following packages will be upgraded:
  bzip2 libbz2-1.0 libbz2-dev
3 upgraded, 3 newly installed, 0 to remove and 295 not upgraded.
Need to get 565 kB of archives.
After this operation, 1855 kB of additional disk space will be used.
Get:1 http://us-east-1.ec2.ports.ubuntu.com/ubuntu-ports jammy-updates/main arm64 bzip2 arm64 1.0.8-5ubuntu0.1 [34.5 kB]
Package configuration




                ┌────┤ Daemons using outdated libraries ├─────┐
                │                                             │
                │                                             │
                │ Which services should be restarted?         │
                │                                             │
                │    [ ] networkd-dispatcher.service          │
Too long with no output (exceeded 10m0s): context deadline exceeded
                │    [ ] unattended-upgrades.service          │
                │    [*] xvfb.service                         │
                │                                             │
                │                                             │
                │          <Ok>              <Cancel>         │
                │                                             │
                └─────────────────────────────────────────────┘
```

despite of us using `-y` to `apt install`. This didn't happen before.

PR #6040, opened 2026-09-21
pse2john: port script to Python 3 and handle multiple PSE files

### LINKED ISSUE #102: Update pse2john script in John repo
https://github.com/openwall/john/blob/bleeding-jumbo/run/pse2john.py

martingalloar:
- Port pse2john.py from Python 2 to Python 3
- Preserve John-compatible hash output
- Support multiple input files and optional output files
- Report unsupported PSE files on stderr while continuing processing
- Handle plain PSE files without aborting the batch


solardiz:
Thank you! How has this been tested? Do we have suitable sample files in https://github.com/openwall/john-samples?

martingalloar:
Oh wasn't aware of that repo! I just submitted a separate PR there :)

https://github.com/openwall/john-samples/pull/51

run/pse2john.py
```diff
@@ -19,52 +19,89 @@
...
 # Standard imports
-import os
-import sys
+from argparse import ArgumentParser
 from binascii import hexlify
+from os import path
+from sys import stderr, stdout
 
 # Custom imports
+import pysap
```
exploide: For usability reasons, we have a `try ... except ImportError` around third party imports in some of the other scripts, to emit a message to stderr which tells users what library needs to be installed. However, this is somewhat unusual in the broader Python world and this script is a 1:1 copy of your version in the pysap project. So in my opinion it doesn't really matter.

run/pse2john.py
```diff
@@ -19,52 +19,89 @@
...
+def process_file(filename, output_file=stdout, error_file=stderr):
+    """Process a PSE file and write John the Ripper material if supported."""
+    try:
+        with open(filename, "rb") as fp:
+            data = fp.read()
+
+        pse_file = SAPPSEFile(data)
+
+        if not pse_file.is_encrypted():
+            raise Exception("Unsupported PSE file type")
```
exploide: If the PSE file really is unencrypted and this check is reliable, then there is no work to be done for john, right? So instead of saying the format is unsupported, wouldn't it be better to clearly state the file is already unencrypted?

exploide (approved):
This is clearly an improvement. Thanks!

While I left two comments, I think this would be still acceptable as is.

Side note: This is no longer compatible with python2, which is fine for me but I know that @solardiz was in favor of maintaining compatibility as far as possible. However, since this depends on a third party library that might have diverged after the EOL of Python2 (I don't know), this might be an unreasonable effort.

run/pse2john.py
martingalloar: Just moved to a try/except so at least we're bespoke about that the pysap library need to be installed :)

run/pse2john.py
martingalloar: Yes, nice catch, in those cases the file is not PIN-protected and certs can be exported as is! Updated to a better verbiage now

martingalloar:
> Side note: This is no longer compatible with python2, which is fine for me but I know that @solardiz was in favor of maintaining compatibility as far as possible. However, since this depends on a third party library that might have diverged after the EOL of Python2 (I don't know), this might be an unreasonable effort.

Yes, the core pysap library has moved into Python 3 only, and keeping backwards compatibility is not planned, so it will be a big effort to add it just for this script.

